PR #1434, opened 2026-06-09
fix(F-02): Replace reversible encryption of temp passwords with bcrypt hashing

akilarootcode:
## Summary
Fixes pentest finding **F-02: Reversible Encryption of Temporary Passwords and Admin Disclosure** (CWE-257).

## Problem
Temp passwords were stored using AES-GCM symmetric encryption. The `share-password` endpoint decrypted and returned stored passwords on demand, meaning a compromised admin account could harvest credentials for all onboarding users.

## Solution
- **Temp passwords are now stored as bcrypt hashes** (non-recoverable)
- `sharePassword()` and `resetAndSharePassword()` generate a **fresh** temp password each call, store only the hash, and return plaintext exactly once
- `createNewPassword()` uses `passwordEncoder.matches()` instead of decrypting
- Bulk employee creation sends invitation email inline after transaction commits
- `sendUserInvitationEmail` accepts temp password as a method parameter (no entity/DTO storage)
- Removed `EncryptionDecryptionService` from auth and people services

## Test
- Added `SharePasswordIntegrationTest` verifying the full API-to-DB flow: endpoint returns plaintext, DB stores bcrypt hash, `passwordEncoder.matches()` confirms they correspond

## Files Changed
- `AuthServiceImpl.java` - share/reset password logic
- `PeopleServiceImpl.java` - user creation flows
- `PeopleEmailService.java` / `PeopleEmailServiceImpl.java` - interface accepts temp password param
- `AsyncEmailServiceImpl.java` - updated for new interface
- `EpAuthServiceImpl.java` / `EpPeopleServiceImpl.java` - removed EncryptionDecryptionService
- `SharePasswordIntegrationTest.java` - new integration test

before

after

### DIFF
--- a/backend/src/main/java/com/skapp/community/common/service/impl/AsyncEmailServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/common/service/impl/AsyncEmailServiceImpl.java
@@ -50,7 +50,7 @@ public class AsyncEmailServiceImpl {
 			try {
 				User user = userDao.findByEmail(result.getEmail())
 					.orElseThrow(() -> new IllegalArgumentException("User not found for email: " + result.getEmail()));
-				retrySendEmail(user);
+				retrySendEmail(user, null);
 				log.info("Email sent successfully to: {}", result.getEmail());
 			}
 			catch (Exception exception) {
@@ -59,11 +59,11 @@ public class AsyncEmailServiceImpl {
 		});
 	}
 
-	private void retrySendEmail(User user) throws Exception {
+	private void retrySendEmail(User user, String tempPassword) throws Exception {
 		int retries = 0;
 		while (retries < CommonConstants.MAX_RETRIES_COUNT) {
 			try {
-				peopleEmailService.sendUserInvitationEmail(user);
+				peopleEmailService.sendUserInvitationEmail(user, tempPassword);
 				return;
 			}
 			catch (TooManyRequestsException exception) {
--- a/backend/src/main/java/com/skapp/community/common/service/impl/AsyncEmailServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/common/service/impl/AsyncEmailServiceImpl.java
@@ -50,7 +50,7 @@ public class AsyncEmailServiceImpl {
 			try {
 				User user = userDao.findByEmail(result.getEmail())
 					.orElseThrow(() -> new IllegalArgumentException("User not found for email: " + result.getEmail()));
-				retrySendEmail(user, null);
+				retrySendEmail(user, result.getTempPassword());
 				log.info("Email sent successfully to: {}", result.getEmail());
 			}
 			catch (Exception exception) {
--- a/backend/src/main/java/com/skapp/community/common/service/impl/AuthServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/common/service/impl/AuthServiceImpl.java
@@ -409,7 +409,7 @@ public class AuthServiceImpl implements AuthService {
 		String tempPassword = CommonModuleUtils.generateSecureRandomPassword();
 		user.setTempPassword(passwordEncoder.encode(tempPassword));
 		user.setPassword(passwordEncoder.encode(tempPassword));
-		user.setIsPasswordChangedForTheFirstTime(true);
+		user.setIsPasswordChangedForTheFirstTime(false);
 		userDao.save(user);
 
 		SharePasswordResponseDto sharePasswordResponseDto = getSharePasswordResponseDto(user, user, tempPassword);
@@ -433,7 +433,7 @@ public class AuthServiceImpl implements AuthService {
 		log.info("resetAndSharePassword: Generated new temp password for userEmail={}", user.getEmail());
 		user.setTempPassword(passwordEncoder.encode(tempPassword));
 		user.setPassword(passwordEncoder.encode(tempPassword));
-		user.setIsPasswordChangedForTheFirstTime(true);
+		user.setIsPasswordChangedForTheFirstTime(false);
 		User savedUser = userDao.save(user);
 
 		log.info("resetAndSharePassword: User password reset and saved for  userEmail={}", user.getEmail());
@@ -677,6 +677,7 @@ public class AuthServiceImpl implements AuthService {
 				tempPassword = CommonModuleUtils.generateSecureRandomPassword();
 				user.setTempPassword(passwordEncoder.encode(tempPassword));
 				user.setPassword(passwordEncoder.encode(tempPassword));
+				user.setIsPasswordChangedForTheFirstTime(false);
 			}
 
 			userDao.save(user);
--- a/backend/src/main/java/com/skapp/community/common/service/impl/AuthServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/common/service/impl/AuthServiceImpl.java
@@ -433,7 +433,7 @@ public class AuthServiceImpl implements AuthService {
 		log.info("resetAndSharePassword: Generated new temp password for userEmail={}", user.getEmail());
 		user.setTempPassword(passwordEncoder.encode(tempPassword));
 		user.setPassword(passwordEncoder.encode(tempPassword));
-		user.setIsPasswordChangedForTheFirstTime(false);
+		user.setIsPasswordChangedForTheFirstTime(true);
 		User savedUser = userDao.save(user);
 
 		log.info("resetAndSharePassword: User password reset and saved for  userEmail={}", user.getEmail());
--- a/backend/src/main/java/com/skapp/community/common/service/impl/AuthServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/common/service/impl/AuthServiceImpl.java
@@ -27,7 +27,6 @@ import com.skapp.community.common.repository.OrganizationConfigDao;
 import com.skapp.community.common.repository.UserDao;
 import com.skapp.community.common.service.AuthService;
 import com.skapp.community.common.service.BulkContextService;
-import com.skapp.community.common.service.EncryptionDecryptionService;
 import com.skapp.community.common.service.JwtService;
 import com.skapp.community.common.service.UserService;
 import com.skapp.community.common.type.BulkItemStatus;
@@ -110,8 +109,6 @@ public class AuthServiceImpl implements AuthService {
 	private final PeopleEmailService peopleEmailService;
 
 	private final PeopleNotificationService peopleNotificationService;
-
-	private final EncryptionDecryptionService encryptionDecryptionService;
 
 	private final ProfileActivator profileActivator;
 
@@ -408,9 +405,14 @@ public class AuthServiceImpl implements AuthService {
 		}
 		User user = optionalUser.get();
 
-		log.info("sharePassword: User found for sharePassword: userEmail={}", user.getEmail());
-		SharePasswordResponseDto sharePasswordResponseDto = getSharePasswordResponseDto(user, user,
-				encryptionDecryptionService.decrypt(user.getTempPassword()));
+		log.info("sharePassword: Generating new temp password for userEmail={}", user.getEmail());
+		String tempPassword = CommonModuleUtils.generateSecureRandomPassword();
+		user.setTempPassword(passwordEncoder.encode(tempPassword));
+		user.setPassword(passwordEncoder.encode(tempPassword));
+		user.setIsPasswordChangedForTheFirstTime(true);
+		userDao.save(user);
+
+		SharePasswordResponseDto sharePasswordResponseDto = getSharePasswordResponseDto(user, user, tempPassword);
 
 		log.info("sharePassword: execution ended for  ={}", user.getEmail());
 		return new ResponseEntityDto(false, sharePasswordResponseDto);
@@ -429,7 +431,7 @@ public class AuthServiceImpl implements AuthService {
 
 		String tempPassword = CommonModuleUtils.generateSecureRandomPassword();
 		log.info("resetAndSharePassword: Generated new temp password for userEmail={}", user.getEmail());
-		user.setTempPassword(encryptionDecryptionService.encrypt(tempPassword));
+		user.setTempPassword(passwordEncoder.encode(tempPassword));
 		user.setPassword(passwordEncoder.encode(tempPassword));
 		user.setIsPasswordChangedForTheFirstTime(true);
 		User savedUser = userDao.save(user);
@@ -615,8 +617,8 @@ public class AuthServiceImpl implements AuthService {
 
 	protected void createNewPassword(String newPassword, User user) {
 		log.info("createNewPassword: execution started={}", user.getEmail());
-		String tempPassword = user.getTempPassword();
-		if (tempPassword != null && Objects.equals(encryptionDecryptionService.decrypt(tempPassword), newPassword)) {
+		String tempPasswordHash = user.getTempPassword();
+		if (tempPasswordHash != null && passwordEncoder.matches(newPassword, tempPasswordHash)) {
 			throw new ModuleException(CommonMessageConstant.COMMON_ERROR_CANNOT_USE_PREVIOUS_PASSWORDS);
 		}
 
@@ -670,14 +672,15 @@ public class AuthServiceImpl implements AuthService {
 			LoginMethod loginMethod = firstUser.isPresent() ? firstUser.get().getLoginMethod()
 					: LoginMethod.CREDENTIALS;
 
+			String tempPassword = null;
 			if (loginMethod.equals(LoginMethod.CREDENTIALS)) {
-				String tempPassword = CommonModuleUtils.generateSecureRandomPassword();
-				user.setTempPassword(encryptionDecryptionService.encrypt(tempPassword));
+				tempPassword = CommonModuleUtils.generateSecureRandomPassword();
+				user.setTempPassword(passwordEncoder.encode(tempPassword));
 				user.setPassword(passwordEncoder.encode(tempPassword));
 			}
 
 			userDao.save(user);
-			peopleEmailService.sendUserInvitationEmail(user);
+			peopleEmailService.sendUserInvitationEmail(user, tempPassword);
 			bulkStatusSummary.incrementSuccessCount();
 		}
 		catch (Exception e) {
--- a/backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
@@ -282,6 +282,8 @@ public class PeopleServiceImpl implements PeopleService {
 		applicationEventPublisher.publishEvent(new UserCreatedEvent(this, user));
 		addNewQuickUploadedEmployeeTimeLineRecords(employee, employeeQuickAddDto);
 		if (!isGuestConversion) {
+			// Skip for guest conversion: email is sent via createUserEntity ->
+			// resendInvitationEmail
 			peopleEmailService.sendUserInvitationEmail(user, tempPassword);
 			updateSubscriptionQuantity(1L, true, false);
 		}
--- a/backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
@@ -1192,6 +1192,8 @@ public class PeopleServiceImpl implements PeopleService {
 		List<CompletableFuture<Void>> tasks = createEmployeeTasks(validEmployeeBulkDtoList, executorService, results);
 		waitForTaskCompletion(tasks, executorService);
 
+		asyncEmailServiceImpl.sendEmailsInBackground(results);
+
 		List<EmployeeBulkDto> overflowedEmployeeBulkDtoList = getOverFlowedEmployeeBulkDtoList(employeeBulkDtoList,
 				validEmployeeBulkDtoList);
 
@@ -2072,8 +2074,6 @@ public class PeopleServiceImpl implements PeopleService {
 			user.setTempPassword(encodedTempPassword);
 			user.setPassword(encodedTempPassword);
 			user.setIsPasswordChangedForTheFirstTime(false);
-
-			user.setIsPasswordChangedForTheFirstTime(false);
 			user.setLoginMethod(LoginMethod.CREDENTIALS);
 		}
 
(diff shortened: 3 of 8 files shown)


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR correctly replaces reversible encryption of temporary passwords with bcrypt hashing (a real security improvement) and threads the plaintext temp password through the invitation flow since hashes can't be reversed. The core hashing logic is sound, but moving bulk-invitation email sending inline introduced a result-tracking bug, and the `sharePassword` endpoint silently changed from read-only to credential-resetting. The old async/retry email path is now dead code.

Found **10** new issue(s):

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
```diff
@@ -2475,9 +2488,16 @@ private CompletableFuture<Void> createEmployeeTask(EmployeeBulkDto employeeBulkD
...
-						messageUtil.getMessage(PeopleMessageConstant.PEOPLE_SUCCESS_EMPLOYEE_ADDED), results);
+				String tempPassword = CommonModuleUtils.generateSecureRandomPassword();
+				saveEmployeeInTransaction(employeeBulkDto, transactionTemplate, tempPassword);
+				EmployeeBulkResponseDto bulkResponseDto = createSuccessResponse(employeeBulkDto,
+						messageUtil.getMessage(PeopleMessageConstant.PEOPLE_SUCCESS_EMPLOYEE_ADDED));
+				results.add(bulkResponseDto);
+
+				User user = userDao.findByEmail(employeeBulkDto.getWorkEmail()).orElse(null);
+				if (user != null && user.getLoginMethod() == LoginMethod.CREDENTIALS) {
+					peopleEmailService.sendUserInvitationEmail(user, tempPassword);
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

Bulk create now sends the invitation email inside the same try block, AFTER the employee transaction has already committed AND after the SUCCESS result was added to `results` (line 2495). If `sendUserInvitationEmail` throws (e.g. TooManyRequestsException, SMTP failure), the `catch (Exception)` at line 2506 calls `handleGeneralException`, which adds a second ERROR entry for the same employee. The result is two entries for one employee: one SUCCESS and one ERROR. In `generateBulkErrorResponse` (line 2035), `successCount = totalSize - errorResults.size()`, so a successfully-created-and-committed employee gets reported as a failure, an error log is shown to the admin, and `updateSubscriptionQuantity(successCount,...)` undercounts. Previously emails were decoupled (background), so email failures never affected creation results. Fix: send the email outside the success-tracking try (e.g. after collecting all results), or catch/log email exceptions locally so they don't convert a committed success into a reported failure.

backend/src/main/java/com/skapp/community/common/service/impl/AuthServiceImpl.java
```diff
@@ -408,9 +405,14 @@ public ResponseEntityDto sharePassword(Long userId) {
 		}
 		User user = optionalUser.get();
 
-		log.info("sharePassword: User found for sharePassword: userEmail={}", user.getEmail());
-		SharePasswordResponseDto sharePasswordResponseDto = getSharePasswordResponseDto(user, user,
-				encryptionDecryptionService.decrypt(user.getTempPassword()));
+		log.info("sharePassword: Generating new temp password for userEmail={}", user.getEmail());
+		String tempPassword = CommonModuleUtils.generateSecureRandomPassword();
+		user.setTempPassword(passwordEncoder.encode(tempPassword));
```
ThinuwanW, line 410: 🤖 **Claude** · 💬 **Note**

`sharePassword` changed from a read-only operation (previously just decrypted and returned the existing temp password) into one that regenerates and overwrites BOTH `tempPassword` and `password` and sets isPasswordChangedForTheFirstTime(true) on every call. This is now byte-for-byte identical to `resetAndSharePassword`. Consequence: if an admin calls share-password for a user who has already activated and chosen their own password, that user's current password is silently invalidated and they're forced into first-time-change — a potential lockout/surprise in production. Also note this is a state-mutating GET endpoint (line 96 in AuthController), which is now a side-effecting GET. Recommend either restricting this path to PENDING/not-yet-activated users, or consolidating with resetAndSharePassword and confirming the product intends every 'share' to reset credentials.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
```diff
@@ -1178,7 +1192,6 @@ public ResponseEntityDto addBulkEmployees(List<EmployeeBulkDto> employeeBulkDtoL
 		List<CompletableFuture<Void>> tasks = createEmployeeTasks(validEmployeeBulkDtoList, executorService, results);
```
ThinuwanW, line 1181: 🤖 **Claude** · 💬 **Note**

The `asyncEmailServiceImpl.sendEmailsInBackground(results)` call was removed from addBulkEmployees, leaving the entire AsyncEmailServiceImpl class (sendEmailsInBackground/processBatch/createBatches/retrySendEmail) and the injected `asyncEmailServiceImpl` field (line 194) as dead code. Two regressions follow: (1) the previous exponential-backoff retry on TooManyRequestsException no longer applies to bulk invitation emails — the new inline send in createEmployeeTask has no retry; (2) addBulkEmployees now blocks until every invitation email is sent synchronously (within the thread pool), so large bulk uploads will have much slower API responses and risk request timeouts. Consider keeping the async/retry path (or replicate the retry logic inline) and removing the now-unused class/field if the decoupling is intentionally dropped.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
```diff
@@ -241,13 +239,13 @@ public ResponseEntityDto createEmployee(CreateEmployeeRequestDto requestDto) {
...
 			// resendInvitationEmail
-			peopleEmailService.sendUserInvitationEmail(user);
+			peopleEmailService.sendUserInvitationEmail(user, tempPassword);
 			updateSubscriptionQuantity(1L, true, false);
 		}
 		invalidateUserCache();
 		invalidateUserAuthPicCache();
 
-		return new ResponseEntityDto(false, processCreateEmployeeResponse(user));
+		return new ResponseEntityDto(false, processCreateEmployeeResponse(user, tempPassword));
```
ThinuwanW, line 248: 🤖 **Claude** · 💬 **Note**

Guest-conversion regression in createEmployee/quickAddEmployee: for an existing (guest) user, createUserEntity returns early at line 504 without populating the AtomicReference holder, so `createUserAndReturnTempPassword` returns null. The fresh temp password is instead generated inside `resendInvitationEmail` (line 539) and never propagated back. As a result `processCreateEmployeeResponse(user, null)` sets employeeCredentials.tempPassword to null in the API response for guest conversions with CREDENTIALS login — previously it returned the actual temp password (via decrypt). The email still goes out correctly, but the admin-facing create response no longer contains the credential. If the response credential is relied upon, capture the resendInvitationEmail temp password (e.g. via the holder) and pass it through.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
ThinuwanW: 🤖 **Claude** · 💬 **Note**

Behavioral regression: bulk invitation emails lost their rate-limit/retry handling. Previously emails went through AsyncEmailServiceImpl.sendEmailsInBackground -> retrySendEmail, which batched (100 at a time) and retried on TooManyRequestsException with exponential backoff (MAX_RETRIES_COUNT). That call was removed, and createEmployeeTask now calls peopleEmailService.sendUserInvitationEmail directly with no retry/backoff. For large bulk uploads, the email provider's rate limit will now produce uncaught TooManyRequestsExceptions that fall into handleGeneralException — so successfully-created employees get marked as errors. Fix: route the bulk send through a retry/backoff helper (reuse retrySendEmail logic) or reinstate a batched async send, rather than firing synchronously per task with no throttling.

backend/src/main/java/com/skapp/community/common/service/impl/AuthServiceImpl.java
```diff
@@ -408,9 +405,14 @@ public ResponseEntityDto sharePassword(Long userId) {
...
+		user.setPassword(passwordEncoder.encode(tempPassword));
```
ThinuwanW, line 411: 🤖 **Claude** · 💬 **Note**

Inefficiency/inconsistency: sharePassword (lines 410-411) and resetAndSharePassword (lines 434-435) call passwordEncoder.encode(tempPassword) twice — once for setTempPassword and once for setPassword — producing two distinct bcrypt hashes and doubling the (intentionally expensive) hashing cost. createUserEntity and createNewEmployeeFromBulk correctly encode once into a local (encodedTempPassword) and reuse it. Recommend doing the same here for consistency and to halve the bcrypt work.

backend/src/main/java/com/skapp/community/common/service/impl/AuthServiceImpl.java
```diff
@@ -615,8 +617,8 @@ private UserSettings createNotificationSettings(User user) {
 
 	protected void createNewPassword(String newPassword, User user) {
 		log.info("createNewPassword: execution started={}", user.getEmail());
-		String tempPassword = user.getTempPassword();
-		if (tempPassword != null && Objects.equals(encryptionDecryptionService.decrypt(tempPassword), newPassword)) {
+		String tempPasswordHash = user.getTempPassword();
```
ThinuwanW, line 620: 🤖 **Claude** · 💬 **Note**

Backward-compatibility gap for existing data: createNewPassword now does passwordEncoder.matches(newPassword, tempPasswordHash). Users created before this change have an AES-encrypted (not bcrypt) value in temp_password. BCryptPasswordEncoder.matches against a non-bcrypt string returns false (and logs a 'does not look like BCrypt' warning) rather than matching, so the 'cannot reuse previous temp password' guard silently stops working for those legacy accounts until their temp password is regenerated. Consider a one-off migration that nulls out or re-hashes legacy temp_password values, and confirm no other code still expects the old encrypted format.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
```diff
@@ -2475,9 +2488,16 @@ private CompletableFuture<Void> createEmployeeTask(EmployeeBulkDto employeeBulkD
 		return CompletableFuture.runAsync(() -> {
 			try {
 				bulkContextService.setContext(tenant);
-				saveEmployeeInTransaction(employeeBulkDto, transactionTemplate);
-				handleSuccessResponse(employeeBulkDto,
-						messageUtil.getMessage(PeopleMessageConstant.PEOPLE_SUCCESS_EMPLOYEE_ADDED), results);
+				String tempPassword = CommonModuleUtils.generateSecureRandomPassword();
```
ThinuwanW, line 2480: 🤖 **Claude** · 💬 **Note**

Minor wasted work: createEmployeeTask unconditionally generates a secure random tempPassword, but when the org login method is GOOGLE, createNewEmployeeFromBulk ignores it (the GOOGLE branch never uses tempPassword) and no credentials email is sent. Consider generating the password only when CREDENTIALS is in effect, or document that it is intentionally pre-generated.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
```diff
@@ -2054,10 +2067,10 @@ private void createNewEmployeeFromBulk(EmployeeBulkDto employeeBulkDto) {
...
-			String tempPassword = CommonModuleUtils.generateSecureRandomPassword();
+			String encodedTempPassword = passwordEncoder.encode(tempPassword);
 
-			user.setTempPassword(encryptionDecryptionService.encrypt(tempPassword));
-			user.setPassword(passwordEncoder.encode(tempPassword));
+			user.setTempPassword(encodedTempPassword);
+			user.setPassword(encodedTempPassword);
 			user.setIsPasswordChangedForTheFirstTime(false);
 
 			user.setIsPasswordChangedForTheFirstTime(false);
```
ThinuwanW, line 2065: 🤖 **Claude** · 💬 **Note**

Redundant statement in the touched bulk method: user.setIsPasswordChangedForTheFirstTime(false) is called twice in the CREDENTIALS branch (lines 2074 and 2076). Remove the duplicate while editing this block.

backend/src/main/java/com/skapp/community/common/service/impl/AuthServiceImpl.java
```diff
@@ -408,9 +405,14 @@ public ResponseEntityDto sharePassword(Long userId) {
...
 
-		log.info("sharePassword: User found for sharePassword: userEmail={}", user.getEmail());
-		SharePasswordResponseDto sharePasswordResponseDto = getSharePasswordResponseDto(user, user,
-				encryptionDecryptionService.decrypt(user.getTempPassword()));
+		log.info("sharePassword: Generating new temp password for userEmail={}", user.getEmail());
+		String tempPassword = CommonModuleUtils.generateSecureRandomPassword();
+		user.setTempPassword(passwordEncoder.encode(tempPassword));
+		user.setPassword(passwordEncoder.encode(tempPassword));
+		user.setIsPasswordChangedForTheFirstTime(true);
+		userDao.save(user);
```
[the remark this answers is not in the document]
akilarootcode: Fixed. Set isPasswordChangedForTheFirstTime(false) and encode bcrypt once into a local variable reused for both fields.

backend/src/main/java/com/skapp/community/common/service/impl/AuthServiceImpl.java
```diff
@@ -429,7 +431,7 @@ public ResponseEntityDto resetAndSharePassword(Long userId) {
 
 		String tempPassword = CommonModuleUtils.generateSecureRandomPassword();
 		log.info("resetAndSharePassword: Generated new temp password for userEmail={}", user.getEmail());
-		user.setTempPassword(encryptionDecryptionService.encrypt(tempPassword));
+		user.setTempPassword(passwordEncoder.encode(tempPassword));
 		user.setPassword(passwordEncoder.encode(tempPassword));
 		user.setIsPasswordChangedForTheFirstTime(true);
 		User savedUser = userDao.save(user);
```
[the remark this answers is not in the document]
akilarootcode: Fixed. Same fix as above - set flag to false and encode once.

backend/src/main/java/com/skapp/community/common/service/impl/AuthServiceImpl.java
```diff
@@ -670,14 +672,15 @@ private void validateAndSendReInvitation(Long id, ReInvitationSkippedCountDto re
...
-				user.setTempPassword(encryptionDecryptionService.encrypt(tempPassword));
+				tempPassword = CommonModuleUtils.generateSecureRandomPassword();
+				user.setTempPassword(passwordEncoder.encode(tempPassword));
 				user.setPassword(passwordEncoder.encode(tempPassword));
 			}
 
 			userDao.save(user);
-			peopleEmailService.sendUserInvitationEmail(user);
+			peopleEmailService.sendUserInvitationEmail(user, tempPassword);
 			bulkStatusSummary.incrementSuccessCount();
```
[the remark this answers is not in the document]
akilarootcode: Fixed. Added isPasswordChangedForTheFirstTime(false) and encode once in the re-invitation flow.

backend/src/main/java/com/skapp/community/common/service/impl/AsyncEmailServiceImpl.java
```diff
@@ -59,11 +59,11 @@ private void processBatch(List<EmployeeBulkResponseDto> batch) {
...
 	}
 
-	private void retrySendEmail(User user) throws Exception {
+	private void retrySendEmail(User user, String tempPassword) throws Exception {
 		int retries = 0;
 		while (retries < CommonConstants.MAX_RETRIES_COUNT) {
 			try {
-				peopleEmailService.sendUserInvitationEmail(user);
+				peopleEmailService.sendUserInvitationEmail(user, tempPassword);
 				return;
```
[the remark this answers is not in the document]
akilarootcode: Fixed. AsyncEmailServiceImpl is now deprecated dead code (stripped to a no-op stub). Bulk emails are sent inline with a dedicated retry helper (retrySendInvitationEmail) that has exponential backoff. Email failures are caught/logged separately and never corrupt the success/error result list.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
```diff
@@ -473,6 +472,17 @@ private Employee createEmployeeEntity(Employee employee, CreateEmployeeRequestDt
...
 
 	private User createUserEntity(User user, CreateEmployeeRequestDto requestDto) {
+		return createUserEntity(user, requestDto, null);
+	}
+
+	private String createUserAndReturnTempPassword(User user, CreateEmployeeRequestDto requestDto) {
+		AtomicReference<String> holder = new AtomicReference<>();
+		createUserEntity(user, requestDto, holder);
+		return holder.get();
+	}
```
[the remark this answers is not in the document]
akilarootcode: Fixed. resendInvitationEmail now returns the generated temp password, and createUserEntity propagates it back through the AtomicReference holder. Guest conversions now correctly include the temp password in the API response.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
```diff
@@ -524,10 +537,11 @@ private void resendInvitationEmail(CreateEmployeeRequestDto requestDto, User use
...
 
 		String tempPassword = CommonModuleUtils.generateSecureRandomPassword();
-		CommonModuleUtils.setIfExists(() -> encryptionDecryptionService.encrypt(tempPassword), user::setTempPassword);
-		CommonModuleUtils.setIfExists(() -> passwordEncoder.encode(tempPassword), user::setPassword);
+		String encodedTempPassword = passwordEncoder.encode(tempPassword);
+		CommonModuleUtils.setIfExists(() -> encodedTempPassword, user::setTempPassword);
+		CommonModuleUtils.setIfExists(() -> encodedTempPassword, user::setPassword);
 
-		peopleEmailService.sendUserInvitationEmail(user);
+		peopleEmailService.sendUserInvitationEmail(user, tempPassword);
```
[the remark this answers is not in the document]
akilarootcode: Fixed. resendInvitationEmail now sets isPasswordChangedForTheFirstTime(false) explicitly.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
akilarootcode: Fixed. Email sending is now separated from the success/error tracking. The email is sent in a dedicated try-catch (sendBulkInvitationEmailWithRetry) AFTER the success result is added, so email failures are logged but never convert a committed employee into an error result. Added exponential backoff retry (3 attempts).

backend/src/main/java/com/skapp/community/common/service/impl/AuthServiceImpl.java
akilarootcode: Fixed. sharePassword now checks if the user account status is PENDING before regenerating credentials. If the user has already activated (non-PENDING), it throws an error. This prevents accidental lockout of active users. resetAndSharePassword remains available for the explicit credential-reset use case.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
akilarootcode: Fixed. AsyncEmailServiceImpl is now a deprecated no-op stub (kept only to avoid breaking constructor injection in EpPeopleServiceImpl). The retry/backoff logic has been replicated inline via retrySendInvitationEmail in PeopleServiceImpl with exponential backoff. The addBulkEmployees call to sendEmailsInBackground was already removed.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
akilarootcode: Fixed. resendInvitationEmail now returns the generated temp password, and createUserEntity propagates it back through the AtomicReference holder. Guest conversions now correctly include the credential in both the email and the API response.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
akilarootcode: Fixed. Added retrySendInvitationEmail with exponential backoff (3 retries). Email exceptions are caught separately and logged without affecting the bulk result tracking.

backend/src/main/java/com/skapp/community/common/service/impl/AuthServiceImpl.java
akilarootcode: Fixed. Both sharePassword and resetAndSharePassword now encode once into a local encodedTempPassword variable and reuse it for both setTempPassword and setPassword.

backend/src/main/java/com/skapp/community/common/service/impl/AuthServiceImpl.java
akilarootcode: Fixed. createNewPassword now checks if the stored value starts with the bcrypt prefix before calling passwordEncoder.matches(). Legacy AES-encrypted values are silently skipped (the guard won't fire for non-bcrypt strings). A follow-up DB migration to null out legacy temp_password values is recommended but out of scope for this PR.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
akilarootcode: Fixed. Temp password is now only generated when the org login method is CREDENTIALS (checked via getOrgLoginMethod() before calling saveEmployeeInTransaction).

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
akilarootcode: Fixed. Removed the duplicate line.

shakila-kamalasena (approved):

backend/src/main/java/com/skapp/community/common/service/impl/AuthServiceImpl.java
ThinuwanW, line 410: should save only one

ThinuwanW (approved):

PR #1443, opened 2026-06-09
feat: add reorder deal functionality with fractional indexing

anusham-rc:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds a deal-reorder endpoint using fractional indexing. The controller wiring, DTO, and service follow existing CRM patterns correctly, but the reorder logic has a few robustness gaps: unhandled IllegalArgumentException from generateKeyBetween, no stage handling/validation, and a missing guard when neither neighbour is supplied. The security annotation and DAO usage are consistent with the rest of the module.

Found **9** new issue(s):

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
```diff
@@ -156,4 +157,40 @@ public ResponseEntityDto getBoardInitData() {
...
+		}
+
+		String nextOrderIndex = null;
+		if (requestDto.getNextDealId() != null) {
+			CrmDeal nextDeal = crmDealDao.findByIdAndIsDeletedFalse(requestDto.getNextDealId())
+				.orElseThrow(() -> new ModuleException(CrmMessageConstant.CRM_ERROR_DEAL_NOT_FOUND));
+			nextOrderIndex = nextDeal.getOrderIndex();
+		}
+
+		String newOrderIndex = FractionalIndexUtil.generateKeyBetween(prevOrderIndex, nextOrderIndex);
```
ThinuwanW, line 268: 🤖 **Claude** · 💬 **Note**

FractionalIndexUtil.generateKeyBetween(a, b) throws IllegalArgumentException whenever a >= b (and on other malformed-key conditions). Here `prevOrderIndex` (a) and `nextOrderIndex` (b) come straight from client-supplied neighbour IDs, so stale board state, swapped/equal neighbours, or passing the moved deal itself as previousDealId/nextDealId will all make a >= b. Because IllegalArgumentException is not a ModuleException, it escapes as an HTTP 500 instead of a clean validation error. Validate that prevOrderIndex < nextOrderIndex (and that dealId differs from the neighbour IDs) before calling generateKeyBetween, and throw a ModuleException/bad-request error otherwise — or wrap the call in a try/catch that translates IllegalArgumentException into a domain error.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
```diff
@@ -156,4 +157,40 @@ public ResponseEntityDto getBoardInitData() {
 		return new ResponseEntityDto(false, responseDto);
 	}
 
+	@Override
+	@Transactional
+	public ResponseEntityDto reorderDeal(CrmDealReorderRequestDto requestDto) {
```
ThinuwanW, line 222: 🤖 **Claude** · 💬 **Note**

orderIndex is only meaningful within a single stage — createDeal computes it via findMaxOrderIndexByStageId(stageId), and deals are sorted by stage.orderIndex then deal.orderIndex. reorderDeal neither updates the deal's stage nor verifies that the previous/next deals belong to the same stage as the deal being moved. As a result: (1) moving a deal to a different kanban column cannot change its stage (the request DTO has no stageId), and (2) if a client passes neighbours from another stage, the generated orderIndex is positioned relative to deals in a different column, producing incorrect ordering. Add a check that prevDeal/nextDeal share the moved deal's stage, and if cross-column moves are intended, accept and set a stageId.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
```diff
@@ -156,4 +157,40 @@ public ResponseEntityDto getBoardInitData() {
...
+			prevOrderIndex = prevDeal.getOrderIndex();
+		}
+
+		String nextOrderIndex = null;
+		if (requestDto.getNextDealId() != null) {
+			CrmDeal nextDeal = crmDealDao.findByIdAndIsDeletedFalse(requestDto.getNextDealId())
+				.orElseThrow(() -> new ModuleException(CrmMessageConstant.CRM_ERROR_DEAL_NOT_FOUND));
+			nextOrderIndex = nextDeal.getOrderIndex();
+		}
+
```
ThinuwanW, line 267: 🤖 **Claude** · 💬 **Note**

There is no guard for the case where both previousDealId and nextDealId are null. generateKeyBetween(null, null) always returns the constant "a0", so every such reorder collapses deals to the same index, causing ordering collisions. Require that at least one neighbour is provided (or document/handle the both-null "move to start/end" case explicitly).

backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmBoardController.java
```diff
@@ -29,4 +32,13 @@ public ResponseEntity<ResponseEntityDto> getBoardInitData() {
 		return new ResponseEntity<>(response, HttpStatus.OK);
 	}
 
+	@Operation(summary = "Reorder a deal on the board",
+			description = "Reorders a deal using fractional indexing. Provide prevDealId and/or nextDealId to position the deal between neighbours.")
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

The Operation description says "Provide prevDealId and/or nextDealId", but the DTO field is named previousDealId (the recent rename). Update the Swagger description to match the actual field name to avoid confusing API consumers.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
```diff
@@ -156,4 +157,40 @@ public ResponseEntityDto getBoardInitData() {
 		return new ResponseEntityDto(false, responseDto);
 	}
 
+	@Override
+	@Transactional
+	public ResponseEntityDto reorderDeal(CrmDealReorderRequestDto requestDto) {
+		log.info("reorderDeal: reordering deal with id={}", requestDto.getDealId());
+
+		if (requestDto.getDealId() == null) {
```
ThinuwanW, line 233: 🤖 **Claude** · 💬 **Note**

A null dealId throws CRM_ERROR_DEAL_NOT_FOUND, which is semantically misleading — a missing required field is a bad-request/validation problem, not a not-found one. Consider a dedicated validation error (e.g. via CrmValidations) so the response status/message reflect a malformed request rather than a missing entity.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
```diff
@@ -156,4 +157,40 @@ public ResponseEntityDto getBoardInitData() {
...
+			throw new ModuleException(CrmMessageConstant.CRM_ERROR_DEAL_NOT_FOUND);
+		}
+
+		CrmDeal deal = crmDealDao.findByIdAndIsDeletedFalse(requestDto.getDealId())
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

Authorization gap: reorderDeal performs no ownership/edit-restriction check. Every other CRM mutation that a sales rep can reach guards with CrmValidations.isEditRestricted(currentUser, owner.getEmployeeId()) — see CrmContactServiceImpl.editContact (line 121, throws CRM_ERROR_CONTACT_EDIT_DENIED) and CrmTaskServiceImpl (line 119). Because reorderDeal only loads the deal by id, any ROLE_CRM_SALES_REPRESENTATIVE can reorder deals owned by other reps. Fix: resolve the current user (userService.getCurrentUser()) and, after loading the deal, call `if (CrmValidations.isEditRestricted(currentUser, deal.getOwner().getEmployeeId())) throw new ModuleException(<deal edit denied constant>);` consistent with the contact/task flows.

backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmBoardController.java
```diff
@@ -29,4 +32,13 @@ public ResponseEntity<ResponseEntityDto> getBoardInitData() {
...
+	@PatchMapping("/deal-reorder")
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

REST/placement inconsistency: reorderDeal is a deal mutation but is mounted on CrmBoardController under /v1/crm/board, while all other deal operations (createDeal, getDeals) live in CrmDealController under /v1/crm/deal. This splits deal endpoints across two controllers and makes the API surface harder to discover. Consider moving reorderDeal to CrmDealController (e.g. PATCH /v1/crm/deal/reorder) to keep deal operations cohesive; the board controller already only serves board init data.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
ThinuwanW, line 268: 🤖 **Claude** · 💬 **Note**

Concurrency: two reorder requests targeting the same gap (same previous/next neighbours) will independently call generateKeyBetween with identical bounds and produce the same orderIndex, since there is no unique constraint on order_index and no optimistic locking on CrmDeal. The board query tie-breaks by id (CrmDealRepositoryImpl line 53/80) so it is not fatal, but the resulting order becomes non-deterministic and a subsequent reorder between those two now-equal keys would throw IllegalArgumentException (a >= b) → HTTP 500. Consider an @Version field on CrmDeal or re-reading neighbours within the transaction; at minimum document the single-writer assumption.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
ThinuwanW, line 233: 🤖 **Claude** · 💬 **Note**

No guard against self-referencing neighbours: if previousDealId or nextDealId equals dealId (a stale/buggy client payload), the deal's own current orderIndex is used as a bound, producing a meaningless position; if both previousDealId and nextDealId resolve to the same deal, prevOrderIndex == nextOrderIndex and generateKeyBetween throws IllegalArgumentException → HTTP 500. Suggest rejecting requests where previousDealId or nextDealId equals dealId (or where they equal each other) with a bad-request style ModuleException.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
```diff
@@ -156,4 +157,40 @@ public ResponseEntityDto getBoardInitData() {
...
+		}
+
+		String nextOrderIndex = null;
+		if (requestDto.getNextDealId() != null) {
+			CrmDeal nextDeal = crmDealDao.findByIdAndIsDeletedFalse(requestDto.getNextDealId())
+				.orElseThrow(() -> new ModuleException(CrmMessageConstant.CRM_ERROR_DEAL_NOT_FOUND));
+			nextOrderIndex = nextDeal.getOrderIndex();
+		}
+
+		String newOrderIndex = FractionalIndexUtil.generateKeyBetween(prevOrderIndex, nextOrderIndex);
```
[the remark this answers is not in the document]
anusham-rc: Fixed. Added stage-membership validation for both neighbour deals and a self-reference guard. Both throw a ModuleException with a descriptive message key.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
anusham-rc: Fixed. Added a self-reference guard (throws CRM_ERROR_DEAL_INVALID_NEIGHBOUR) and stage-membership validation (throws CRM_ERROR_DEAL_NEIGHBOUR_STAGE_MISMATCH). The FractionalIndexUtil.generateKeyBetween call is left unwrapped — its IllegalArgumentException path is now unreachable after these guards, and generateKeyBetween already contains the same  >= b check internally.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
anusham-rc: Fixed. Added stage-membership validation — both previousDeal and 
extDeal are checked against the moved deal's stageId, throwing CRM_ERROR_DEAL_NEIGHBOUR_STAGE_MISMATCH on mismatch. Cross-column stage changes are out of scope for this endpoint; the DTO intentionally has no stageId field.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
anusham-rc: Fixed. Added a guard that throws CRM_ERROR_DEAL_NEIGHBOURS_REQUIRED when both previousDealId and 
extDealId are null, preventing generateKeyBetween(null, null) from assigning the constant value 0 to multiple deals.

backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmBoardController.java
anusham-rc: Fixed. The endpoint has been moved to CrmDealController (PATCH /v1/crm/deal/reorder) and the Swagger description now correctly references previousDealId.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
anusham-rc: Fixed. The null dealId check now throws CRM_ERROR_DEAL_ID_REQUIRED — a dedicated validation error — rather than CRM_ERROR_DEAL_NOT_FOUND, which is reserved for the DB lookup failure below it.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
anusham-rc: Fixed. Added userService.getCurrentUser() and CrmValidations.isEditRestricted(currentUser, deal.getOwner().getEmployeeId()) after loading the deal, consistent with the contact and task flows. Throws CRM_ERROR_DEAL_EDIT_DENIED when a sales rep attempts to reorder a deal they don't own.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
anusham-rc: Acknowledged. Concurrent reorders targeting the same gap will produce equal orderIndex values, which the board query tie-breaks by id (non-deterministic but not fatal). A subsequent reorder between two equal keys would throw IllegalArgumentException from generateKeyBetween. Addressing this properly requires either a @Version field on CrmDeal for optimistic locking or re-reading neighbours within the transaction. This is noted as a known limitation to be tracked separately — it is out of scope for this PR.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
anusham-rc: Fixed. Added a self-reference guard before the DB lookups: if previousDealId or 
extDealId equals dealId, it throws CRM_ERROR_DEAL_INVALID_NEIGHBOUR immediately.

backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmBoardController.java
anusham-rc: The endpoint stays in CrmBoardController as PATCH /v1/crm/board/deal-reorder — all board-related APIs are kept together. The Swagger description has been updated to correctly reference previousDealId.

backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmBoardController.java
anusham-rc: Correction to previous reply: the endpoint was NOT moved to CrmDealController. It stays in CrmBoardController as PATCH /v1/crm/board/deal-reorder — all board-related APIs are kept together. The Swagger description has been updated to correctly reference previousDealId.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
```diff
@@ -156,4 +161,62 @@ public ResponseEntityDto getBoardInitData() {
...
+	@Transactional
+	public ResponseEntityDto reorderDeal(CrmDealReorderRequestDto requestDto) {
+		log.info("reorderDeal: reordering deal with id={}", requestDto.getDealId());
+
+		if (requestDto.getDealId() == null) {
+			throw new ModuleException(CrmMessageConstant.CRM_ERROR_DEAL_ID_REQUIRED);
+		}
+
+		if (requestDto.getPreviousDealId() == null && requestDto.getNextDealId() == null) {
+			throw new ModuleException(CrmMessageConstant.CRM_ERROR_DEAL_NEIGHBOURS_REQUIRED);
```
akilarootcode: deal order neighbours

akilarootcode (changes requested):

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
anusham-rc: renamed

akilarootcode (dismissed):

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
```diff
@@ -212,4 +217,62 @@ public ResponseEntityDto getBoardInitData() {
...
+
+		if (requestDto.getDealId().equals(requestDto.getPreviousDealId())
+				|| requestDto.getDealId().equals(requestDto.getNextDealId())) {
+			throw new ModuleException(CrmMessageConstant.CRM_ERROR_DEAL_INVALID_NEIGHBOUR);
+		}
+
+		CrmDeal deal = crmDealDao.findByIdAndIsDeletedFalse(requestDto.getDealId())
+			.orElseThrow(() -> new ModuleException(CrmMessageConstant.CRM_ERROR_DEAL_NOT_FOUND));
+
+		User currentUser = userService.getCurrentUser();
```
ThinuwanW: move the permission up in all the services as a practice

ThinuwanW (changes requested):

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
anusham-rc: moved the user validation to the top

PR #1446, opened 2026-06-10
feat: add CRM access control for Representative role in getDrawerRoutes

dulsara-manakal:
**Title:** `feat: hide CRM nav item when CRM module is disabled`

**Description:**

Added CRM module visibility guard to the sidebar navigation.

Previously, the CRM route had no role-based guard in getDrawerRoutes.ts, causing it to remain visible in the sidebar for all authorized users (including Super Admin) even when the CRM module was disabled from Organization Settings.

**Changes:**
- Imported `RepresentativeTypes` and added it to the `Role` union type
- Added a CRM block that returns `null` when the user does not have `ROLE_CRM_SALES_REPRESENTATIVE`, consistent with how other optional modules (Leave, Timesheet, Invoices, Sign, Projects) are handled

**Behavior:**
- CRM module **enabled** → CRM nav visible for users with any CRM role (`CRM_ADMIN`, `CRM_SALES_MANAGER`, `CRM_SALES_REPRESENTATIVE`)
- CRM module **disabled** → CRM nav hidden for all users (backend strips all CRM roles from the token via `EpAuthorityServiceImpl`)

### DIFF
--- a/frontend/src/community/common/utils/getDrawerRoutes.ts
+++ b/frontend/src/community/common/utils/getDrawerRoutes.ts
@@ -4,6 +4,7 @@ import {
   AdminTypes,
   EmployeeTypes,
   ManagerTypes,
+  RepresentativeTypes,
   SenderTypes,
   SuperAdminType
 } from "~community/common/types/AuthTypes";
@@ -11,7 +12,12 @@ import routes from "~community/common/utils/data/routes";
 import getEnterpriseDrawerRoutes from "~community/common/utils/getEnterpriseDrawerRoutes";
 import { TierEnum } from "~enterprise/common/enums/Common";
 
-type Role = AdminTypes | ManagerTypes | EmployeeTypes | SuperAdminType;
+type Role =
+  | AdminTypes
+  | ManagerTypes
+  | EmployeeTypes
+  | SuperAdminType
+  | RepresentativeTypes;
 
 type RouteWithBadge = { badge?: string };
 
@@ -231,6 +237,16 @@ const getDrawerRoutes = ({
         );
 
         if (!isInvoiceManager) {
+          return null;
+        }
+      }
+
+      if (route?.name === "CRM") {
+        const hasCRMAccess = userRoles?.includes(
+          RepresentativeTypes.CRM_SALES_REPRESENTATIVE
+        );
+
+        if (!hasCRMAccess) {
           return null;
         }
       }
--- a/frontend/src/community/common/utils/getEnterpriseDrawerRoutes.ts
+++ b/frontend/src/community/common/utils/getEnterpriseDrawerRoutes.ts
@@ -3,6 +3,7 @@ import {
   AdminTypes,
   EmployeeTypes,
   ManagerTypes,
+  RepresentativeTypes,
   SuperAdminType
 } from "~community/common/types/AuthTypes";
 
@@ -10,7 +11,12 @@ import { GlobalLoginMethod } from "../enums/CommonEnums";
 import { IconName } from "../types/IconTypes";
 import routes from "./data/routes";
 
-type Role = AdminTypes | ManagerTypes | EmployeeTypes | SuperAdminType;
+type Role =
+  | AdminTypes
+  | ManagerTypes
+  | EmployeeTypes
+  | SuperAdminType
+  | RepresentativeTypes;
 
 interface Props {
   userRoles: Role[] | undefined;


frontend/src/community/common/utils/getDrawerRoutes.ts
```diff
@@ -235,6 +241,16 @@ const getDrawerRoutes = ({
...
 
+      if (route?.name === "CRM") {
+        const hasCRMAccess = userRoles?.includes(
+          RepresentativeTypes.CRM_SALES_REPRESENTATIVE
+        );
+
+        if (!hasCRMAccess) {
+          return null;
+        }
+      }
```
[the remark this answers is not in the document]
dulsara-manakal: Always CRM_SALES_REPRESENTATIVE role is there in every three role types. no need to  do changes.

ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds a CRM-specific access gate in getDrawerRoutes that only shows the CRM drawer item to users with the CRM_SALES_REPRESENTATIVE role. However, the CRM route's requiredAuthLevel (routes.ts) and middleware.ts both grant CRM access to SUPER_ADMIN, CRM_ADMIN, and CRM_SALES_MANAGER as well, so this gate hides the CRM menu item from those three authorized roles — a likely regression. The type addition and import are correct.

Found **3** new issue(s): 🔴 2 important, 🟡 1 suggestion(s)

frontend/src/community/common/utils/getDrawerRoutes.ts
```diff
@@ -235,6 +241,16 @@ const getDrawerRoutes = ({
         }
       }
 
+      if (route?.name === "CRM") {
+        const hasCRMAccess = userRoles?.includes(
```
ThinuwanW, line 245: 🤖 **Claude** · 🔴 **Important**

The new CRM gate only grants drawer visibility to RepresentativeTypes.CRM_SALES_REPRESENTATIVE, but the CRM route's requiredAuthLevel (data/routes.ts:207-212) authorizes SUPER_ADMIN, CRM_ADMIN, CRM_SALES_MANAGER, and CRM_SALES_REPRESENTATIVE — and middleware.ts grants ROUTES.CRM.BASE to all four (adminRoutes CRM_ADMIN, managerRoutes CRM_SALES_MANAGER, superAdminRoutes, employeeRoutes representative). As written, SUPER_ADMIN, CRM_ADMIN, and CRM_SALES_MANAGER will no longer see the CRM item in the drawer even though they can still navigate to the page directly, which is inconsistent and a regression vs. the prior behavior (the generic isAuthorized path showed CRM to all authorized roles). Fix: check membership against all CRM-authorized roles, e.g. const hasCRMAccess = userRoles?.some((role) => [AdminTypes.SUPER_ADMIN, AdminTypes.CRM_ADMIN, ManagerTypes.CRM_SALES_MANAGER, RepresentativeTypes.CRM_SALES_REPRESENTATIVE].includes(role)). If the intent really is to restrict the CRM drawer link to representatives only, confirm that product decision and align middleware/requiredAuthLevel accordingly.

frontend/src/community/common/utils/getDrawerRoutes.ts
```diff
@@ -235,6 +241,16 @@ const getDrawerRoutes = ({
         }
       }
 
+      if (route?.name === "CRM") {
```
ThinuwanW, line 244: 🤖 **Claude** · 🟡 **Suggestion**

This special-case block duplicates authorization logic that the generic `isAuthorized && route?.hasSubTree` path (lines 318-368) already enforces via the route's requiredAuthLevel. Since CRM has a subTree and its requiredAuthLevel already lists exactly the intended roles, the generic path already handles visibility correctly. Unless there is a specific reason the generic path fails for CRM, consider removing this block entirely rather than maintaining a second, narrower copy of the role list that can drift out of sync with requiredAuthLevel.

frontend/src/community/common/utils/getDrawerRoutes.ts
```diff
@@ -4,14 +4,20 @@ import {
...
+  RepresentativeTypes,
   SenderTypes,
   SuperAdminType
 } from "~community/common/types/AuthTypes";
 import routes from "~community/common/utils/data/routes";
 import getEnterpriseDrawerRoutes from "~community/common/utils/getEnterpriseDrawerRoutes";
 import { TierEnum } from "~enterprise/common/enums/Common";
 
-type Role = AdminTypes | ManagerTypes | EmployeeTypes | SuperAdminType;
+type Role =
```
ThinuwanW, line 15: 🤖 **Claude** · 🔴 **Important**

Widening the local `Role` type to include `RepresentativeTypes` makes `userRoles` typed as `(AdminTypes | ManagerTypes | EmployeeTypes | SuperAdminType | RepresentativeTypes)[] | undefined`. This same `userRoles` is passed into `getEnterpriseDrawerRoutes({ userRoles, ... })` at line 51, but that function's `Props.userRoles` is still the narrower `Role[]` (its local `Role` type at getEnterpriseDrawerRoutes.ts:13 does NOT include `RepresentativeTypes`). Since `(...|RepresentativeTypes)[]` is not assignable to `(AdminTypes|ManagerTypes|EmployeeTypes|SuperAdminType)[]`, this is a TS2345 type error that will fail the type-check/build. Fix by adding `RepresentativeTypes` to the `Role` type and `import` in getEnterpriseDrawerRoutes.ts as a paired change (or share a single `Role` type between the two files).

frontend/src/community/common/utils/getDrawerRoutes.ts
dulsara-manakal: CRM_SALES_REPRESENTATIVE always there in the other two as well.

frontend/src/community/common/utils/getDrawerRoutes.ts
dulsara-manakal: Cant do that Super admin level is still exisit.

frontend/src/community/common/utils/getDrawerRoutes.ts
dulsara-manakal: Fixed it

hasalarootcode (approved):

frontend/src/community/common/utils/getDrawerRoutes.ts
ThinuwanW, line 244: check with Hasala

ThinuwanW (changes requested):

ThinuwanW (approved):

PR #1449, opened 2026-06-10
Feat/136 create a deal stage

sanjula-rtc:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds a POST endpoint to create CRM deal stages with manual validation, a duplicate-name check, and a free-tier limit. The core logic is sound and well-tested, but there are three important issues: an NPE risk on null color (returns 500 instead of 400), the new method returns the raw JPA entity (leaking audit/internal fields) instead of a mapped response DTO unlike every other CRM create method, and the free-tier error message text contradicts the configured limit.

Found **13** new issue(s): 🔴 3 important, 🟡 1 suggestion(s), 🟣 3 nit(s)

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealStageServiceImpl.java
```diff
@@ -34,4 +41,39 @@ public ResponseEntityDto getDealStages() {
...
+		}
+
+		if (isFreeTierOpenStageLimitExceeded()) {
+			throw new ModuleException(CrmMessageConstant.CRM_ERROR_DEAL_STAGE_FREE_TIER_LIMIT_EXCEEDED);
+		}
+
+		CrmDealStage stage = new CrmDealStage();
+		stage.setName(requestDto.getName());
+		stage.setDescription(requestDto.getDescription());
+		stage.setColor(requestDto.getColor().toString());
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`requestDto.getColor().toString()` will throw a NullPointerException if `color` is omitted/null in the request body, since there is no @NotNull on the DTO and no manual validation for color. This produces a 500 error instead of a clean 400. Add a validation (e.g. `CrmValidations.validateDealStageColor(requestDto.getColor())` throwing a ModuleException with a dedicated message key) before using it, consistent with the name/description validations. Also prefer `.name()` over `.toString()` for enum-to-string persistence to be explicit.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealStageServiceImpl.java
```diff
@@ -34,4 +41,39 @@ public ResponseEntityDto getDealStages() {
...
+		stage.setStageType(CrmConstants.DEFAULT_DEAL_STAGE_TYPE);
+		stage.setOrderIndex(crmDealStageDao.countByIsDeletedFalse() + 1);
+
+		CrmDealStage saved = crmDealStageDao.save(stage);
+
+		log.info("createDealStage: execution ended, created stage id={}", saved.getId());
+
+		return new ResponseEntityDto(false, saved);
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

This returns the raw `CrmDealStage` JPA entity, whereas every other CRM create method (CrmTaskServiceImpl, CrmContactServiceImpl, CrmDealServiceImpl) returns a mapped response DTO via crmMapper. Serializing the entity leaks internal/audit fields from Auditable (createdBy, createdDate, lastModifiedBy, lastModifiedDate) and isDeleted. A `CrmDealStageResponseDto` already exists (and `description` was just added to it for this purpose). Map the saved entity to it: `return new ResponseEntityDto(false, crmMapper.crmDealStageToCrmDealStageResponseDto(saved));`.

backend/src/main/resources/community/messages/crm-messages.properties
```diff
@@ -59,3 +59,14 @@ api.success.crm.contact-deleted=Contact deleted successfully
...
+########################################
+#### CRM Deal Stage Error Messages ####
+########################################
+api.error.crm.deal-stage.name-required=Deal stage name is required
+api.error.crm.deal-stage.name-length=Deal stage name must be between 2 and 50 characters
+api.error.crm.deal-stage.name-invalid-chars=Deal stage name contains invalid characters
+api.error.crm.deal-stage.name-duplicate=A deal stage with this name already exists
+api.error.crm.deal-stage.description-too-long=Deal stage description must not exceed 250 characters
+api.error.crm.deal-stage.description-invalid-chars=Deal stage description contains invalid characters
+api.error.crm.deal-stage.free-tier-limit-exceeded=Free tier allows a maximum of 3 open deal stages
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The message says "Free tier allows a maximum of 3 open deal stages" but `CrmConstants.FREE_TIER_MAX_OPEN_DEAL_STAGES` is 1, so the user-facing text contradicts the enforced limit. Either fix the constant to 3 (if 3 was intended) or correct the message to match 1. Better still, parameterize the message with the constant value (ModuleException supports args) so the two can never drift.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealStageServiceImpl.java
```diff
@@ -34,4 +41,39 @@ public ResponseEntityDto getDealStages() {
...
+		if (isFreeTierOpenStageLimitExceeded()) {
+			throw new ModuleException(CrmMessageConstant.CRM_ERROR_DEAL_STAGE_FREE_TIER_LIMIT_EXCEEDED);
+		}
+
+		CrmDealStage stage = new CrmDealStage();
+		stage.setName(requestDto.getName());
+		stage.setDescription(requestDto.getDescription());
+		stage.setColor(requestDto.getColor().toString());
+		stage.setStageType(CrmConstants.DEFAULT_DEAL_STAGE_TYPE);
+		stage.setOrderIndex(crmDealStageDao.countByIsDeletedFalse() + 1);
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`orderIndex = countByIsDeletedFalse() + 1` can collide with an existing orderIndex after a soft-delete. E.g. stages with orderIndex 1,2,3 → delete the one with index 2 → count becomes 2 → new stage gets index 3, duplicating the existing index-3 stage and breaking the ordering used by getDealStages. Consider deriving from `max(orderIndex)+1` instead of a count. This is also not concurrency-safe, though that matters less under the free-tier limit.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealStageServiceImpl.java
```diff
@@ -34,4 +41,39 @@ public ResponseEntityDto getDealStages() {
...
+
+		return new ResponseEntityDto(false, saved);
+	}
+
+	protected boolean isFreeTierOpenStageLimitExceeded() {
+		return crmDealStageDao
+			.countByStageTypeAndIsDeletedFalse(CrmDealStageType.OPEN) >= CrmConstants.FREE_TIER_MAX_OPEN_DEAL_STAGES;
+	}
+
+}
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

File is missing a trailing newline ("\ No newline at end of file"). The repo applies formatting (the last commit is "fix: apply format"); add the trailing newline to avoid failing the format/spotless check.

backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmDealStageController.java
```diff
@@ -29,4 +32,12 @@ public ResponseEntity<ResponseEntityDto> getDealStages() {
 		return new ResponseEntity<>(response, HttpStatus.OK);
 	}
 
+	@Operation(summary = "Create a deal stage", description = "Creates a new deal stage.")
+	@PostMapping()
```
ThinuwanW, line 36: 🤖 **Claude** · 🟣 **Nit**

The GET mapping declares `produces = MediaType.APPLICATION_JSON_VALUE` but the new `@PostMapping()` declares neither `produces` nor `consumes`. For consistency and clearer API contracts, consider `@PostMapping(consumes = MediaType.APPLICATION_JSON_VALUE, produces = MediaType.APPLICATION_JSON_VALUE)`.

backend/src/main/java/com/skapp/community/crmplanner/util/CrmValidations.java
```diff
@@ -257,4 +257,29 @@ public static void validateDealBelongsToCompany(CrmDeal deal, CrmCompany company
 		}
 	}
 
+	public static void validateDealStageName(String name) {
```
ThinuwanW, line 261: 🤖 **Claude** · 🟣 **Nit**

The name is persisted as-is without trimming, while the duplicate check uses IgnoreCase but not trim. "Proposal" and " Proposal " would both pass and be stored as distinct stages despite being effectively the same. Consider trimming the name before validation/persistence and comparing the trimmed value for duplicates.

backend/src/main/java/com/skapp/community/crmplanner/constant/CrmConstants.java
```diff
@@ -19,6 +20,20 @@ public class CrmConstants {
 
 	public static final String DEAL_AMOUNT_REGEX = "^\\d+(\\.\\d{1,2})?$";
 
+	public static final int DEAL_STAGE_NAME_MIN_LENGTH = 2;
+
+	public static final int DEAL_STAGE_NAME_MAX_LENGTH = 50;
+
+	public static final String DEAL_STAGE_NAME_REGEX = "^(?=.*\\p{L})[\\p{L}\\p{N} \\-.']+$";
```
ThinuwanW, line 31: 🤖 **Claude** · 💬 **Note**

`FREE_TIER_MAX_OPEN_DEAL_STAGES = 1` doesn't just make the user-facing message wrong (already noted) — it causes the PR's own integration test `createDealStage_ValidPayload_ReturnsCreated` to FAIL. That test seeds one OPEN stage via `savedOpenStage("Existing Stage", 1)` and then expects the POST to return 201 Created with `orderIndex == 2`. With the limit at 1, `isFreeTierOpenStageLimitExceeded()` evaluates `1 >= 1 == true`, so the service throws `CRM_ERROR_DEAL_STAGE_FREE_TIER_LIMIT_EXCEEDED` and returns 400, not 201. The test expectation (seed 1, create a 2nd successfully) plus the message text both imply the constant should be 3. Set `FREE_TIER_MAX_OPEN_DEAL_STAGES = 3` (or fix the test/message) so the constant, message, and tests agree.

backend/src/main/java/com/skapp/community/crmplanner/constant/CrmMessageConstant.java
```diff
@@ -61,7 +61,15 @@ public enum CrmMessageConstant implements MessageConstant {
...
-	CRM_SUCCESS_COMPANY_DELETED("api.success.crm.company-deleted");
+	CRM_SUCCESS_COMPANY_DELETED("api.success.crm.company-deleted"),
+	CRM_ERROR_DEAL_STAGE_NAME_REQUIRED("api.error.crm.deal-stage.name-required"),
+	CRM_ERROR_DEAL_STAGE_NAME_LENGTH("api.error.crm.deal-stage.name-length"),
+	CRM_ERROR_DEAL_STAGE_NAME_INVALID_CHARS("api.error.crm.deal-stage.name-invalid-chars"),
+	CRM_ERROR_DEAL_STAGE_NAME_DUPLICATE("api.error.crm.deal-stage.name-duplicate"),
+	CRM_ERROR_DEAL_STAGE_DESCRIPTION_TOO_LONG("api.error.crm.deal-stage.description-too-long"),
+	CRM_ERROR_DEAL_STAGE_DESCRIPTION_INVALID_CHARS("api.error.crm.deal-stage.description-invalid-chars"),
+	CRM_ERROR_DEAL_STAGE_FREE_TIER_LIMIT_EXCEEDED("api.error.crm.deal-stage.free-tier-limit-exceeded"),
+	CRM_SUCCESS_DEAL_STAGE_CREATED("api.success.crm.deal-stage-created");
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

`CRM_SUCCESS_DEAL_STAGE_CREATED` and its property `api.success.crm.deal-stage-created` are introduced but never referenced anywhere — `createDealStage` returns `new ResponseEntityDto(false, saved)` with no message. This is dead code. Either pass the success message into the response (e.g. `new ResponseEntityDto(messageUtil.getMessage(CRM_SUCCESS_DEAL_STAGE_CREATED), false, ...)`) to match the delete endpoints' pattern, or drop the unused constant and property.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealStageServiceImpl.java
```diff
@@ -34,4 +41,39 @@ public ResponseEntityDto getDealStages() {
...
-}
+	@Override
+	@Transactional
+	public ResponseEntityDto createDealStage(CrmDealStageCreateRequestDto requestDto) {
+		log.info("createDealStage: execution started");
+
+		CrmValidations.validateDealStageName(requestDto.getName());
+		CrmValidations.validateDealStageDescription(requestDto.getDescription());
+
+		if (crmDealStageDao.existsByNameIgnoreCaseAndIsDeletedFalse(requestDto.getName())) {
```
ThinuwanW, line 61: 🤖 **Claude** · 💬 **Note**

TOCTOU race: the duplicate check (`existsByNameIgnoreCaseAndIsDeletedFalse`) and the `orderIndex = countByIsDeletedFalse() + 1` computation are both read-then-write with no DB-level guard. Two concurrent CRM_ADMIN requests can each pass the existence check and both insert the same name, or compute the same `orderIndex`. There is no unique constraint on `crm_deal_stage.name` or `order_index` (see the entity — plain `@Column`s). Consider a partial/functional unique index on active names and catching `DataIntegrityViolationException`, or serializing creation, rather than relying solely on the app-level check.

backend/src/main/java/com/skapp/community/crmplanner/repository/CrmDealStageDao.java
```diff
@@ -14,4 +15,10 @@ public interface CrmDealStageDao extends JpaRepository<CrmDealStage, Long> {
 
 	Optional<CrmDealStage> findByIdAndIsDeletedFalse(Long id);
 
+	boolean existsByNameIgnoreCaseAndIsDeletedFalse(String name);
+
+	long countByStageTypeAndIsDeletedFalse(CrmDealStageType stageType);
+
+	int countByIsDeletedFalse();
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

Return-type inconsistency: `countByIsDeletedFalse()` returns `int` while the sibling `countByStageTypeAndIsDeletedFalse(...)` returns `long`. Spring Data count queries conventionally return `long`. Make `countByIsDeletedFalse()` return `long` for consistency (the `+ 1` assignment to `Integer orderIndex` can stay an int cast at the call site if needed).

backend/src/main/java/com/skapp/community/crmplanner/payload/request/CrmDealStageCreateRequestDto.java
```diff
@@ -0,0 +1,17 @@
...
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+public class CrmDealStageCreateRequestDto {
+
+	private String name;
+
+	private CrmDealStageColors color;
```
ThinuwanW, line 13: 🤖 **Claude** · 💬 **Note**

`color` has no validity check comparable to name/description. The first pass noted the NPE when `color` is null; separately, an unknown color string (e.g. "RED") fails Jackson enum deserialization and produces a generic 400/parse error rather than a CRM-style message. Unlike name and description, there is no `validateDealStageColor` / message constant. Consider validating color (required + valid enum) so all three fields fail with consistent, user-friendly CRM error messages.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealStageServiceImpl.java
```diff
@@ -34,4 +41,39 @@ public ResponseEntityDto getDealStages() {
...
+		stage.setOrderIndex(crmDealStageDao.countByIsDeletedFalse() + 1);
+
+		CrmDealStage saved = crmDealStageDao.save(stage);
+
+		log.info("createDealStage: execution ended, created stage id={}", saved.getId());
+
+		return new ResponseEntityDto(false, saved);
+	}
+
+	protected boolean isFreeTierOpenStageLimitExceeded() {
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

`isFreeTierOpenStageLimitExceeded()` is declared `protected` but is only called internally and the `@Service` class isn't designed for subclassing. Unless this is deliberately exposed for test spying (the tests don't spy it — they use the real DAO), it should be `private` to avoid widening the API surface.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealStageServiceImpl.java
```diff
@@ -34,4 +41,39 @@ public ResponseEntityDto getDealStages() {
...
+		stage.setColor(requestDto.getColor().toString());
+		stage.setStageType(CrmConstants.DEFAULT_DEAL_STAGE_TYPE);
+		stage.setOrderIndex(crmDealStageDao.countByIsDeletedFalse() + 1);
+
+		CrmDealStage saved = crmDealStageDao.save(stage);
+
+		log.info("createDealStage: execution ended, created stage id={}", saved.getId());
+
+		return new ResponseEntityDto(false, saved);
+	}
```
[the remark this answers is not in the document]
sanjula-rtc: Fixed — now maps to `CrmDealStageResponseDto` via `crmMapper.crmDealStageToCrmDealStageResponseDto(saved)`, consistent with other CRM create endpoints.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealStageServiceImpl.java
sanjula-rtc: Fixed — now returns `new ResponseEntityDto(false, crmMapper.crmDealStageToCrmDealStageResponseDto(saved))`, consistent with other CRM create endpoints.

backend/src/main/java/com/skapp/community/crmplanner/util/CrmValidations.java
sanjula-rtc: Trimming is handled globally by a request body advice/interceptor that trims all String fields before they reach the service layer, so this is already covered.

backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmDealStageController.java
sanjula-rtc: Not adding — the existing POST mappings across the CRM controllers don't consistently declare `consumes`/`produces`, so adding it here would be inconsistent with the rest of the codebase. Spring defaults to application/json for @RequestBody anyway.

backend/src/main/java/com/skapp/community/crmplanner/constant/CrmMessageConstant.java
sanjula-rtc: Removed — `CRM_SUCCESS_DEAL_STAGE_CREATED` constant and the `api.success.crm.deal-stage-created` property were dead code since the endpoint doesn't return a success message. Cleaned up both.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealStageServiceImpl.java
sanjula-rtc: Fixed — changed from `protected` to `private`.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealStageServiceImpl.java
```diff
@@ -34,4 +41,39 @@ public ResponseEntityDto getDealStages() {
...
+		CrmValidations.validateDealStageName(requestDto.getName());
+		CrmValidations.validateDealStageDescription(requestDto.getDescription());
+
+		if (crmDealStageDao.existsByNameIgnoreCaseAndIsDeletedFalse(requestDto.getName())) {
+			throw new ModuleException(CrmMessageConstant.CRM_ERROR_DEAL_STAGE_NAME_DUPLICATE);
+		}
+
+		if (isFreeTierOpenStageLimitExceeded()) {
+			throw new ModuleException(CrmMessageConstant.CRM_ERROR_DEAL_STAGE_FREE_TIER_LIMIT_EXCEEDED);
+		}
```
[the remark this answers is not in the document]
sanjula-rtc: removed the tier limitations from this PR.

backend/src/main/resources/community/messages/crm-messages.properties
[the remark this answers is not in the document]
sanjula-rtc: removed the tier limitations from this PR.

sanjula-rtc: removed the tier limitations from this PR.

backend/src/main/java/com/skapp/community/crmplanner/constant/CrmConstants.java
sanjula-rtc: removed the tier limitations from this PR.

backend/src/main/java/com/skapp/community/crmplanner/payload/request/CrmDealStageCreateRequestDto.java
sanjula-rtc: this is acceptable.
this is mapped to consistent enum in frontend and backend 
and allowed to add from a list of colors

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealStageServiceImpl.java
```diff
@@ -34,4 +41,39 @@ public ResponseEntityDto getDealStages() {
...
+
+		CrmDealStage stage = new CrmDealStage();
+		stage.setName(requestDto.getName());
+		stage.setDescription(requestDto.getDescription());
+		stage.setColor(requestDto.getColor().toString());
+		stage.setStageType(CrmConstants.DEFAULT_DEAL_STAGE_TYPE);
```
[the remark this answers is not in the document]
sanjula-rtc: Fixed — added `CrmValidations.validateDealStageColor(requestDto.getColor())` that throws a clean 400 with message key `CRM_ERROR_DEAL_STAGE_COLOR_REQUIRED` when color is null. Also switched from `.toString()` to `.name()` for explicit enum-to-string persistence.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealStageServiceImpl.java
sanjula-rtc: Fixed — added `CrmValidations.validateDealStageColor(requestDto.getColor())` that throws `ModuleException(CRM_ERROR_DEAL_STAGE_COLOR_REQUIRED)` returning a clean 400 when color is null. Also switched to `.name()` instead of `.toString()` for explicit enum persistence.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealStageServiceImpl.java
```diff
@@ -34,4 +41,39 @@ public ResponseEntityDto getDealStages() {
...
+		stage.setOrderIndex(crmDealStageDao.countByIsDeletedFalse() + 1);
+
```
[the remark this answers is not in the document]
sanjula-rtc: Fixed — replaced `countByIsDeletedFalse() + 1` with a `findMaxOrderIndexByIsDeletedFalse()` query. New stages get `max(orderIndex) + 1`, which avoids duplicates after soft-deletes. When no stages exist, it defaults to 1.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealStageServiceImpl.java
sanjula-rtc: Fixed — replaced `countByIsDeletedFalse() + 1` with `findMaxOrderIndexByIsDeletedFalse()` + a private `getNextOrderIndex()` helper that returns `max(orderIndex) + 1` (or 1 if no stages exist). No more collisions after soft-deletes.

backend/src/main/java/com/skapp/community/crmplanner/repository/CrmDealStageDao.java
sanjula-rtc: Fixed — changed `countByIsDeletedFalse()` return type from `int` to `long` for consistency with Spring Data conventions and the sibling count methods.

PR #1451, opened 2026-06-10
Fix/core page loader issue

hasalarootcode:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/frontend/src/community/common/components/organisms/ContentWithDrawer/ContentWithDrawer.tsx
+++ b/frontend/src/community/common/components/organisms/ContentWithDrawer/ContentWithDrawer.tsx
@@ -34,34 +34,44 @@ const ContentWithDrawer = ({ children }: Props) => {
         <Drawer />
         <Stack sx={classes.contentWrapper}>
           <AppBar />
-          <main
-            aria-busy={loading}
-            style={{
-              display: "flex",
-              flexDirection: "column",
-              width: "100%",
-              flex: 1,
-              overflowX: "hidden",
-              position: "relative"
-            }}
-          >
-            {loading && <FullScreenLoader fullPage={false} />}
-            <Stack
-              id="content-with-drawer-main-content"
-              tabIndex={loading ? -1 : 0}
-              role="document"
-              aria-label={translateAria(["contentAreaWithDrawer"])}
-              {...(loading && { inert: "" })}
+          {loading ? (
+            <div
               style={{
+                width: "100%",
+                flex: 1,
+                position: "relative",
+                minHeight: "calc(100vh - 64px)"
+              }}
+            >
+              <FullScreenLoader fullPage={false} />
+            </div>
+          ) : (
+            <main
+              aria-busy={loading}
+              style={{
+                display: "flex",
                 flexDirection: "column",
                 width: "100%",
                 flex: 1,
                 overflowX: "hidden"
               }}
             >
-              {children}
-            </Stack>
-          </main>
+              <Stack
+                id="content-with-drawer-main-content"
+                tabIndex={loading ? -1 : 0}
+                role="document"
+                aria-label={translateAria(["contentAreaWithDrawer"])}
+                style={{
+                  flexDirection: "column",
+                  width: "100%",
+                  flex: 1,
+                  overflowX: "hidden"
+                }}
+              >
+                {children}
+              </Stack>
+            </main>
+          )}
         </Stack>
       </Stack>
       <ToastMessage
--- a/frontend/src/community/common/components/organisms/ContentWithDrawer/ContentWithDrawer.tsx
+++ b/frontend/src/community/common/components/organisms/ContentWithDrawer/ContentWithDrawer.tsx
@@ -36,18 +36,20 @@ const ContentWithDrawer = ({ children }: Props) => {
           <AppBar />
           {loading ? (
             <div
+              role="status"
+              aria-busy={true}
+              aria-live="polite"
               style={{
                 width: "100%",
                 flex: 1,
                 position: "relative",
-                minHeight: "calc(100vh - 64px)"
+                minHeight: "100dvh"
               }}
             >
               <FullScreenLoader fullPage={false} />
             </div>
           ) : (
             <main
-              aria-busy={loading}
               style={{
                 display: "flex",
                 flexDirection: "column",
@@ -58,7 +60,7 @@ const ContentWithDrawer = ({ children }: Props) => {
             >
               <Stack
                 id="content-with-drawer-main-content"
-                tabIndex={loading ? -1 : 0}
+                tabIndex={0}
                 role="document"
                 aria-label={translateAria(["contentAreaWithDrawer"])}
                 style={{
--- a/frontend/src/community/common/components/organisms/ContentWithDrawer/styles.ts
+++ b/frontend/src/community/common/components/organisms/ContentWithDrawer/styles.ts
@@ -2,7 +2,6 @@ const styles = () => ({
   protectedWrapper: {
     flexDirection: "row",
     width: "100%",
-    height: "100vh",
     overflowX: "clip"
   },
   contentWrapper: {


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR refactors the loading state in ContentWithDrawer so that during route loading a plain <div> with a loader is shown instead of overlaying the loader on top of inert <main> content, and removes the fixed height: 100vh from protectedWrapper. The logic is functionally sound, but the change introduces an accessibility regression (the <main> landmark disappears during loading), leaves dead conditional code, hardcodes the AppBar height as a magic number, and alters viewport-height behavior for all protected pages.

Found **9** new issue(s): 🟡 5 suggestion(s), 🟣 4 nit(s)

frontend/src/community/common/components/organisms/ContentWithDrawer/ContentWithDrawer.tsx
```diff
@@ -34,34 +34,44 @@ const ContentWithDrawer = ({ children }: Props) => {
...
-          >
-            {loading && <FullScreenLoader fullPage={false} />}
-            <Stack
-              id="content-with-drawer-main-content"
-              tabIndex={loading ? -1 : 0}
-              role="document"
-              aria-label={translateAria(["contentAreaWithDrawer"])}
-              {...(loading && { inert: "" })}
+          {loading ? (
+            <div
```
ThinuwanW, line 38: 🤖 **Claude** · 🟡 **Suggestion**

Accessibility regression: during loading the content is now wrapped in a plain <div> instead of <main>, so the page loses its <main> landmark and the role="document" region entirely while a route is changing. Previously the <main>/document region stayed mounted (inert) and the loader overlaid it, keeping the landmark present for assistive tech. Consider keeping the <main> element rendered in both branches (toggle inert/aria-busy and swap only the inner content/loader) so the landmark is stable across loading and loaded states.

frontend/src/community/common/components/organisms/ContentWithDrawer/ContentWithDrawer.tsx
```diff
@@ -34,34 +34,44 @@ const ContentWithDrawer = ({ children }: Props) => {
...
+                flex: 1,
+                position: "relative",
+                minHeight: "calc(100vh - 64px)"
+              }}
+            >
+              <FullScreenLoader fullPage={false} />
+            </div>
+          ) : (
+            <main
+              aria-busy={loading}
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Dead/misleading conditionals in the non-loading branch. Since this <main> only renders when !loading, `aria-busy={loading}` is always "false" and `tabIndex={loading ? -1 : 0}` (line 61) is always 0. The aria-busy signal that previously told assistive tech content was loading no longer fires anywhere. Simplify these to constants (or drop aria-busy here and surface busy state on a stable element).

frontend/src/community/common/components/organisms/ContentWithDrawer/ContentWithDrawer.tsx
```diff
@@ -34,34 +34,44 @@ const ContentWithDrawer = ({ children }: Props) => {
...
-              role="document"
-              aria-label={translateAria(["contentAreaWithDrawer"])}
-              {...(loading && { inert: "" })}
+          {loading ? (
+            <div
               style={{
+                width: "100%",
+                flex: 1,
+                position: "relative",
+                minHeight: "calc(100vh - 64px)"
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Hardcoded magic number `calc(100vh - 64px)` assumes the AppBar/PageHeader is exactly 64px tall. The header comes from the external @rootcodelabs/skapp-ui PageHeader and its height isn't guaranteed/derived here, so if it changes the loader container will under/overflow the viewport. Extract the header height to a shared constant/theme token and reference it, or measure it, instead of duplicating 64px.

frontend/src/community/common/components/organisms/ContentWithDrawer/ContentWithDrawer.tsx
```diff
@@ -34,34 +34,44 @@ const ContentWithDrawer = ({ children }: Props) => {
...
+              }}
+            >
+              <FullScreenLoader fullPage={false} />
+            </div>
+          ) : (
```
ThinuwanW, line 51: 🤖 **Claude** · 🟣 **Nit**

Behavioral change worth confirming: the old code kept `children` mounted (inert) during route loading; now the entire content subtree unmounts while `loading` is true and remounts afterward. For Next.js route changes this is usually fine, but any in-progress child state/scroll position during the loading window is discarded. Confirm this is intended.

frontend/src/community/common/components/organisms/ContentWithDrawer/styles.ts
```diff
@@ -2,7 +2,6 @@ const styles = () => ({
   protectedWrapper: {
     flexDirection: "row",
     width: "100%",
```
ThinuwanW, line 4: 🤖 **Claude** · 🟡 **Suggestion**

Removing `height: "100vh"` from protectedWrapper affects all protected pages, not just the loading state. contentWrapper still sets `height: "100%"`, which now resolves against an auto-height parent, so the wrapper/main is no longer guaranteed to fill the viewport when content is short (the new `minHeight: calc(100vh - 64px)` only applies to the loading <div>, not the loaded <main>). Pages/templates relying on full viewport height (e.g. FullPageContainerLayout) may shrink to content height. Verify full-height layouts still render correctly, or add an equivalent min-height to the loaded <main>.

frontend/src/community/common/components/organisms/ContentWithDrawer/ContentWithDrawer.tsx
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Layout-shift / sizing inconsistency between the two branches: the loading <div> forces `minHeight: calc(100vh - 64px)` but the loaded <main> sets no minHeight and relies only on `flex: 1`. Combined with the removal of `height: 100vh` from protectedWrapper (first-pass finding), short-content pages will render the loader at full viewport height and then visibly collapse/jump when loading completes. Either apply the same min-height to the loaded <main> (or to contentWrapper in styles.ts) so both states occupy the same box, or remove the loading-only minHeight and let flex sizing drive both branches consistently.

frontend/src/community/common/components/organisms/ContentWithDrawer/ContentWithDrawer.tsx
```diff
@@ -34,34 +34,44 @@ const ContentWithDrawer = ({ children }: Props) => {
...
-            }}
-          >
-            {loading && <FullScreenLoader fullPage={false} />}
-            <Stack
-              id="content-with-drawer-main-content"
-              tabIndex={loading ? -1 : 0}
-              role="document"
-              aria-label={translateAria(["contentAreaWithDrawer"])}
-              {...(loading && { inert: "" })}
+          {loading ? (
```
ThinuwanW, line 37: 🤖 **Claude** · 🟡 **Suggestion**

Keyboard-focus loss on every route navigation. Because the loading branch fully unmounts the content subtree (the <main>/<Stack> wrapping children) and renders an unrelated, non-focusable <div>, any element the user had focused is destroyed and focus falls back to <body>. When loading completes and the new <main> mounts, focus is not restored, so keyboard/AT users are silently dropped to the top of the page on each navigation. Consider keeping a stable focusable landmark mounted across both states, or programmatically moving focus to the main content (e.g. focus the #content-with-drawer-main-content container) after routeChangeComplete.

frontend/src/community/common/components/organisms/ContentWithDrawer/ContentWithDrawer.tsx
```diff
@@ -34,34 +34,44 @@ const ContentWithDrawer = ({ children }: Props) => {
...
+            <div
               style={{
+                width: "100%",
```
ThinuwanW, line 43: 🤖 **Claude** · 🟣 **Nit**

The new loading <div> omits `overflowX: "hidden"`, which the old <main> wrapper carried and which the loaded <main> still has. The loader itself won't overflow, but dropping it makes the two branches behave differently for horizontal overflow during the loading frame (protectedWrapper's `overflowX: clip` is the only remaining guard). Add `overflowX: "hidden"` to the loading div to match the loaded branch and avoid relying solely on the ancestor.

frontend/src/community/common/components/organisms/ContentWithDrawer/ContentWithDrawer.tsx
ThinuwanW, line 38: 🤖 **Claude** · 🟣 **Nit**

Styling-approach inconsistency: the component already centralizes layout in styles.ts (classes.protectedWrapper / classes.contentWrapper via `sx`), but the newly added loading branch hardcodes a raw inline `style` block (including the 64px magic number). For maintainability and consistency, move the loading-container styles into styles.ts and reference them via `sx`, rather than introducing a new inline style object.

frontend/src/community/common/components/organisms/ContentWithDrawer/ContentWithDrawer.tsx
```diff
@@ -34,34 +34,44 @@ const ContentWithDrawer = ({ children }: Props) => {
...
+                flex: 1,
+                position: "relative",
+                minHeight: "calc(100vh - 64px)"
+              }}
```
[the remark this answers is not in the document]
hasalarootcode: Fixed

frontend/src/community/common/components/organisms/ContentWithDrawer/ContentWithDrawer.tsx
```diff
@@ -34,34 +34,44 @@ const ContentWithDrawer = ({ children }: Props) => {
...
+            >
+              <FullScreenLoader fullPage={false} />
+            </div>
+          ) : (
+            <main
+              aria-busy={loading}
+              style={{
```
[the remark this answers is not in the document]
hasalarootcode: Fixed

frontend/src/community/common/components/organisms/ContentWithDrawer/ContentWithDrawer.tsx
```diff
@@ -34,34 +34,44 @@ const ContentWithDrawer = ({ children }: Props) => {
...
                 overflowX: "hidden"
               }}
             >
-              {children}
-            </Stack>
-          </main>
+              <Stack
+                id="content-with-drawer-main-content"
+                tabIndex={loading ? -1 : 0}
+                role="document"
```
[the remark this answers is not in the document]
hasalarootcode: Fixed

frontend/src/community/common/components/organisms/ContentWithDrawer/ContentWithDrawer.tsx
hasalarootcode: Not needed

frontend/src/community/common/components/organisms/ContentWithDrawer/ContentWithDrawer.tsx
hasalarootcode: Fixed

frontend/src/community/common/components/organisms/ContentWithDrawer/ContentWithDrawer.tsx
hasalarootcode: Fixed

frontend/src/community/common/components/organisms/ContentWithDrawer/ContentWithDrawer.tsx
hasalarootcode: Yes this works fine fully tested

frontend/src/community/common/components/organisms/ContentWithDrawer/styles.ts
hasalarootcode: Not affected for that

frontend/src/community/common/components/organisms/ContentWithDrawer/ContentWithDrawer.tsx
hasalarootcode: It changes to dvh, and it should only be applied to the loader.

frontend/src/community/common/components/organisms/ContentWithDrawer/ContentWithDrawer.tsx
hasalarootcode: That's the current behaviour

frontend/src/community/common/components/organisms/ContentWithDrawer/ContentWithDrawer.tsx
hasalarootcode: Not needed

frontend/src/community/common/components/organisms/ContentWithDrawer/ContentWithDrawer.tsx
hasalarootcode: Removed

shakila-kamalasena (approved):

ThinuwanW (approved):

PR #1453, opened 2026-06-11
fix: guest user BE access removal on get endpoint

akilarootcode:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleReadServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleReadServiceImpl.java
@@ -1,8 +1,8 @@
 package com.skapp.community.peopleplanner.service.impl;
 
 import com.skapp.community.common.constant.AuthConstants;
+import com.skapp.community.common.exception.AuthenticationException;
 import com.skapp.community.common.exception.EntityNotFoundException;
-import com.skapp.community.common.exception.ModuleException;
 import com.skapp.community.common.payload.response.ResponseEntityDto;
 import com.skapp.community.common.service.UserService;
 import com.skapp.community.common.type.Role;
@@ -63,7 +63,7 @@ public class PeopleReadServiceImpl implements PeopleReadService {
 			.orElseThrow(() -> new EntityNotFoundException(PeopleMessageConstant.PEOPLE_ERROR_EMPLOYEE_NOT_FOUND));
 
 		if (CommonModuleUtils.isGuestEmployee(employee)) {
-			throw new ModuleException(PeopleMessageConstant.PEOPLE_ERROR_EMPLOYEE_ACCESS_DENIED);
+			throw new AuthenticationException(PeopleMessageConstant.PEOPLE_ERROR_EMPLOYEE_ACCESS_DENIED);
 		}
 
 		EmployeeProfileViewAccessLevel accessLevel = resolveAccessLevel(employeeId);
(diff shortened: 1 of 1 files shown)


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds a guest-employee access guard to PeopleReadServiceImpl.getEmployeeById, throwing ModuleException with PEOPLE_ERROR_EMPLOYEE_ACCESS_DENIED when the target employee is a PM guest. The check is null-safe, uses an existing constant with a defined message key, and follows the exact same isGuestEmployee + ModuleException pattern already used in LeaveEntitlementServiceImpl. The guard is correctly placed before access-level resolution and DTO mapping. No issues found.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleReadServiceImpl.java
```diff
@@ -60,6 +62,10 @@ public ResponseEntityDto getEmployeeById(Long employeeId) {
 		Employee employee = employeeDao.findById(employeeId)
 			.orElseThrow(() -> new EntityNotFoundException(PeopleMessageConstant.PEOPLE_ERROR_EMPLOYEE_NOT_FOUND));
 
+		if (CommonModuleUtils.isGuestEmployee(employee)) {
+			throw new ModuleException(PeopleMessageConstant.PEOPLE_ERROR_EMPLOYEE_ACCESS_DENIED);
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

ModuleException is mapped to HTTP 400 BAD_REQUEST by GlobalExceptionHandler.handleModuleExceptions, but this is an authorization/access-denied condition for which 403 FORBIDDEN is semantically correct and less confusing for API clients. This is consistent with the existing guest checks in LeaveEntitlementServiceImpl (which also use ModuleException -> 400), so it's a judgment call — but if a dedicated access-denied exception type mapping to 403 exists or could be introduced, prefer it here. Severity left as suggestion since it follows the established pattern.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleReadServiceImpl.java
```diff
@@ -60,6 +62,10 @@ public ResponseEntityDto getEmployeeById(Long employeeId) {
 		Employee employee = employeeDao.findById(employeeId)
 			.orElseThrow(() -> new EntityNotFoundException(PeopleMessageConstant.PEOPLE_ERROR_EMPLOYEE_NOT_FOUND));
 
+		if (CommonModuleUtils.isGuestEmployee(employee)) {
```
ThinuwanW, line 65: 🤖 **Claude** · 🟣 **Nit**

The guard runs before resolveAccessLevel, so it blocks ALL callers — including SUPER_ADMIN, PEOPLE_ADMIN, and PEOPLE_MANAGER who would otherwise get FULL_ACCESS — from viewing a guest employee's profile via this endpoint. Please confirm this is the intended behavior (i.e. admins should also be denied guest profiles). If admins should still be able to view guest profiles, the check would need to be combined with the resolved access level instead of being unconditional.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleReadServiceImpl.java
akilarootcode: moved to Authentication Exception

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleReadServiceImpl.java
akilarootcode: guest user access need to remove from top

yuhanga-rtc (approved):

ThinuwanW (approved):

PR #1458, opened 2026-06-11
fix: update dejavu to NOTO

thusala:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/frontend/pages/_document.tsx
+++ b/frontend/pages/_document.tsx
@@ -31,11 +31,7 @@ export default function Document() {
           crossOrigin="anonymous"
         />
         <link
-          href="https://fonts.googleapis.com/css2?family=Arimo:wght@400;700&family=Carlito:wght@400;700&family=Tinos:wght@400;700&display=swap"
-          rel="stylesheet"
-        />
-        <link
-          href="https://cdn.jsdelivr.net/npm/dejavu-fonts-ttf@2.37.3/fontfaces/DejaVuSans.css"
+          href="https://fonts.googleapis.com/css2?family=Arimo:wght@400;700&family=Carlito:wght@400;700&family=Noto+Sans:wght@400;700&family=Tinos:wght@400;700&display=swap"
           rel="stylesheet"
         />
         <link


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR swaps the separately CDN-hosted DejaVu Sans font for Noto Sans served via the existing Google Fonts request in `_document.tsx`, consolidating two `<link>` tags into one. The frontend change itself is correct and matches the e-sign font config (FONT_FAMILY_CSS_NAMES[NOTO_SANS] = "Noto Sans"), but it completes a frontend migration whose backend counterpart appears not to have landed: the backend still only knows DEJAVU_SANS, which will break PDF generation for any field using this font.

Found **3** new issue(s): 🔴 1 important, 🟣 1 nit(s)

frontend/pages/_document.tsx
```diff
@@ -31,11 +31,7 @@ export default function Document() {
           crossOrigin="anonymous"
         />
         <link
-          href="https://fonts.googleapis.com/css2?family=Arimo:wght@400;700&family=Carlito:wght@400;700&family=Tinos:wght@400;700&display=swap"
-          rel="stylesheet"
-        />
-        <link
-          href="https://cdn.jsdelivr.net/npm/dejavu-fonts-ttf@2.37.3/fontfaces/DejaVuSans.css"
+          href="https://fonts.googleapis.com/css2?family=Arimo:wght@400;700&family=Carlito:wght@400;700&family=Noto+Sans:wght@400;700&family=Tinos:wght@400;700&display=swap"
```
ThinuwanW, line 34: 🤖 **Claude** · 🔴 **Important**

This completes the frontend migration to Noto Sans (FontFamily.NOTO_SANS = "NOTO_SANS", rendered as CSS "Noto Sans"), but the backend was not updated in lockstep. backend/.../esignature/type/EsignFontFamilyType.java still defines DEJAVU_SANS and has NO NOTO_SANS enum constant. In DocumentProcessingServiceImpl.addAdvanceTextField (line 779), the stored font key is resolved via `EsignFontFamilyType.valueOf(fontFamily)`. When a user selects the Verdana/Noto Sans option, the frontend will persist "NOTO_SANS", and `valueOf("NOTO_SANS")` throws an unchecked IllegalArgumentException. The enclosing try only catches IOException (line 851), so the exception propagates and aborts the entire PDF merge/render — a production-breaking regression for any document/template using this font. Before merging, ensure the backend ships together: (1) rename/add NOTO_SANS in EsignFontFamilyType with the correct folderName and the S3 font files uploaded under that folder, (2) handle/migrate any existing rows whose fontFamily column is still "DEJAVU_SANS", and (3) consider guarding the valueOf call (catch IllegalArgumentException → fall back to a default font) so an unknown font value can never crash rendering.

ThinuwanW, line 34: 🤖 **Claude** · 🟣 **Nit**

DejaVu Sans was likely chosen originally for its very broad Unicode/glyph coverage. The Google Fonts 'Noto Sans' family here serves only the default (mainly Latin) subsets via unicode-range, so documents containing characters outside those subsets (e.g. extended Cyrillic, Greek, symbols) may render as fallback/tofu in the live preview. The generated PDF embeds the actual font file from S3 so it is less affected, but preview/PDF parity could drift. Worth confirming the chosen Noto Sans subset covers the character ranges previously supported by DejaVu, or explicitly request the needed subsets.

frontend/pages/_document.tsx
```diff
@@ -31,11 +31,7 @@ export default function Document() {
           crossOrigin="anonymous"
         />
         <link
```
ThinuwanW, line 33: 🤖 **Claude** · 💬 **Note**

Backward-compatibility for existing e-signature content (distinct from the first pass's 'backend not updated' note about new selections). Removing the DejaVu Sans stylesheet means any previously-created/stored field that recorded DejaVu Sans as its font will no longer have that font available in the browser. The renderer resolves fonts via FONT_FAMILY_CSS_NAMES with a `|| styles.fontFamily` fallback (frontend/src/enterprise/sign/utils/textFieldUtils.ts:44,132,180,206), and DEJAVU_SANS was dropped from both the FontFamily enum and that map, so a historical value will fall through to the raw string and render in the browser's default font — a silent visual drift on already-prepared envelopes rather than a crash. Confirm there is a data migration (or backfill) mapping any persisted DEJAVU_SANS values to a still-supported family (e.g. NOTO_SANS) before this ships, or keep the DejaVu stylesheet until such records are migrated.

PR #1463, opened 2026-06-12
fix: company delete reworked to unlink contact

charith-rc:
### Summary

- Redo company delete to unlink contact

### How to test

1.

### Project Checklist

  - [x] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR changes company deletion to unlink associated contacts (set company to null) instead of soft-deleting associated deals. The core code change is correct: the new derived query and the nullable company_id FK make unlinking safe, the method is properly @Transactional, logging is intact, and the integration test was updated to match. The main concern is that deals previously cleaned up on company deletion are now left entirely unhandled, leaving them pointing at a soft-deleted company.

Found **5** new issue(s):

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmCompanyServiceImpl.java
```diff
@@ -133,9 +133,9 @@ public ResponseEntityDto deleteCompany(Long id) {
 		CrmCompany company = crmCompanyDao.findByIdAndIsDeletedFalse(id)
 			.orElseThrow(() -> new ModuleException(CrmMessageConstant.CRM_ERROR_COMPANY_NOT_FOUND));
 
-		List<CrmDeal> deals = crmDealDao.findAllByCompanyIdAndIsDeletedFalse(id);
-		deals.forEach(deal -> deal.setIsDeleted(true));
-		crmDealDao.saveAll(deals);
+		List<CrmContact> contacts = crmContactDao.findAllByCompanyIdAndIsDeletedFalse(id);
+		contacts.forEach(contact -> contact.setCompany(null));
```
ThinuwanW, line 168: 🤖 **Claude** · 💬 **Note**

Behavioral change: the old code soft-deleted all deals linked to the company; this PR removes that and only unlinks contacts. Deals retain their company_id pointing at the now soft-deleted company. CrmDealRepositoryImpl fetches the company via an unconditional LEFT join (CrmDealRepositoryImpl.java:78/154/218) with no isDeleted filter, so deal listings will still display the deleted company's name. Confirm this is intended. If deals should not reference a deleted company, consider also unlinking them here (e.g. crmDealDao.findAllByCompanyIdAndIsDeletedFalse(id) then deal.setCompany(null)) for consistency with the contact handling. If intentional, this is fine as-is — flagging because the previous behavior was explicitly removed.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmCompanyServiceImpl.java
```diff
@@ -133,9 +133,9 @@ public ResponseEntityDto deleteCompany(Long id) {
...
+		crmContactDao.saveAll(contacts);
```
ThinuwanW, line 169: 🤖 **Claude** · 💬 **Note**

Minor: contacts are loaded inside the @Transactional method and are managed entities, so dirty checking on setCompany(null) already flushes the change at commit. The explicit crmContactDao.saveAll(contacts) is redundant (harmless). Consistent with the prior pattern, so optional to remove.

ThinuwanW, line 169: 🤖 **Claude** · 💬 **Note**

Dead code introduced by this change: `CrmDealDao.findAllByCompanyIdAndIsDeletedFalse(Long companyId)` (CrmDealDao.java:17) was only ever called from this method. Now that the call was replaced with `crmContactDao.findAllByCompanyIdAndIsDeletedFalse`, the CrmDealDao method has no remaining callers (verified via grep across main and test sources — it only appears in its own declaration). Remove the now-unused method from CrmDealDao to avoid leaving dead query code behind.

backend/src/test/java/com/skapp/community/crmplanner/controller/v1/CrmCompanyControllerIntegrationTest.java
```diff
@@ -246,45 +235,29 @@ void deleteCompany_AlreadyDeleted_ReturnsBadRequest() throws Exception {
...
+		Long contactId = crmContactDao.save(contact).getId();
 
 		performDeleteRequest(companyId).andExpect(status().isOk())
 			.andExpect(jsonPath(STATUS_PATH).value(STATUS_SUCCESSFUL))
 			.andExpect(jsonPath(RESULTS_0_PATH + MESSAGE_PATH)
 				.value(messageUtil.getMessage(CrmMessageConstant.CRM_SUCCESS_COMPANY_DELETED)));
 
-		CrmDeal deletedDeal = crmDealDao.findById(dealId).orElseThrow();
-		assertTrue(deletedDeal.getIsDeleted());
+		CrmContact unlinkedContact = crmContactDao.findById(contactId).orElseThrow();
```
ThinuwanW, line 290: 🤖 **Claude** · 💬 **Note**

The rewritten test no longer verifies that the unlink is actually persisted. The whole test runs under @Transactional, so the `contact` saved at line 252 stays in the same persistence context; the service's `setCompany(null)` mutates that same managed instance, and `crmContactDao.findById(contactId)` at line 259 returns the first-level-cache copy. The assertion at line 260 would pass even if nothing were flushed to the DB. Note the sibling test deleteCompany_AlreadyDeleted (line 219-220) deliberately uses `TestTransaction.flagForCommit()/end()` to force a real round-trip. Consider flushing and clearing the persistence context (or committing) before re-reading, so the assertion proves the unlink reached the database rather than just the in-memory entity.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmCompanyServiceImpl.java
```diff
@@ -133,9 +133,9 @@ public ResponseEntityDto deleteCompany(Long id) {
 		CrmCompany company = crmCompanyDao.findByIdAndIsDeletedFalse(id)
 			.orElseThrow(() -> new ModuleException(CrmMessageConstant.CRM_ERROR_COMPANY_NOT_FOUND));
 
-		List<CrmDeal> deals = crmDealDao.findAllByCompanyIdAndIsDeletedFalse(id);
-		deals.forEach(deal -> deal.setIsDeleted(true));
-		crmDealDao.saveAll(deals);
+		List<CrmContact> contacts = crmContactDao.findAllByCompanyIdAndIsDeletedFalse(id);
```
ThinuwanW, line 167: 🤖 **Claude** · 💬 **Note**

Behavioral asymmetry with deleteContact worth confirming: when a contact is deleted (CrmContactServiceImpl.deleteContact), its linked deals and tasks are cascaded to soft-deleted. When a company is deleted, this new code only nulls `contact.company` and does nothing for the company's `deals`/`tasks` collections (CrmCompany.java:54-59). The contacts' own deals/tasks therefore remain active but now reference a contact that has been detached from the (soft-deleted) company. If unlinking-without-cascade is intentional, fine; otherwise consider routing through the contact-deletion cascade or explicitly documenting why company deletion intentionally leaves the contacts and their deals/tasks alive.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmCompanyServiceImpl.java
charith-rc: existing pattern

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmCompanyServiceImpl.java
charith-rc: contact can exist without a company. tasks and deals are mapped primarily to a contact by requirement.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmCompanyServiceImpl.java
charith-rc: companies should not delete any deals by requirement. deals fetch should prevent fetching deleted companies.

backend/src/test/java/com/skapp/community/crmplanner/controller/v1/CrmCompanyControllerIntegrationTest.java
charith-rc: fixed https://github.com/SkappHQ/skapp/pull/1463/commits/5e999d40f6521df8b5d2d3eab76141d133ce6d6c

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmCompanyServiceImpl.java
charith-rc: removed https://github.com/SkappHQ/skapp/pull/1463/commits/5e999d40f6521df8b5d2d3eab76141d133ce6d6c

yuhanga-rtc (dismissed):

charith-rc:
corrected and merged in https://github.com/SkappHQ/skapp/pull/1533

PR #1466, opened 2026-06-14
Feat/crm 104 wire create task

charith-rc:
### Summary

- Wires up create task modal and refactors allowing for decoupling of form and logic

### How to test

1.

### Project Checklist

  - [x] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


frontend/src/community/crm/types/FormTypes.ts
```diff
@@ -0,0 +1,64 @@
...
+  taskTypeOptions: {
+    options: ReturnType<typeof useGetTaskTypeOptions>["options"];
+    getCategoryById: (id: number) => CrmTaskCategory | undefined;
+  };
+
+  // Actions
+  onSubmit: () => void;
+  onCancel: () => void;
+
+  translateText: ReturnType<typeof useTranslator>;
```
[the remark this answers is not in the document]
charith-rc: fixed as suggested https://github.com/SkappHQ/skapp/pull/1466/commits/e5c2d03c85efe86417897c24b9cc69c0a022ba5c

frontend/src/community/crm/types/FormTypes.ts
```diff
@@ -0,0 +1,64 @@
...
+  onDealSearchChange: (value: string) => void;
+  onClearDeal: () => void;
+  isDealFetching: boolean;
+
+  // Options
+  priorityOptions: ReturnType<typeof useGetPriorityOptions>;
+  taskTypeOptions: {
+    options: ReturnType<typeof useGetTaskTypeOptions>["options"];
+    getCategoryById: (id: number) => CrmTaskCategory | undefined;
+  };
```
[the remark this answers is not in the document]
charith-rc: fixed as suggested https://github.com/SkappHQ/skapp/pull/1466/commits/e5c2d03c85efe86417897c24b9cc69c0a022ba5c

frontend/src/community/crm/types/FormTypes.ts
```diff
@@ -0,0 +1,64 @@
+import { FormikErrors } from "formik";
+import { ChangeEvent } from "react";
+import { SearchableDropdownItem } from "~community/common/components/molecules/SearchableDropdown/SearchableDropdown";
+import { useTranslator } from "~community/common/hooks/useTranslator";
+import useGetPriorityOptions from "../hooks/useGetPriorityOptions";
+import useGetTaskTypeOptions from "../hooks/useGetTaskTypeOptions";
+import { CrmTaskAddFormTypes, CrmOwner, CrmTaskCategory } from "./CommonTypes";
```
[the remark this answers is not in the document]
charith-rc: fixed as suggested https://github.com/SkappHQ/skapp/pull/1466/commits/e5c2d03c85efe86417897c24b9cc69c0a022ba5c

frontend/src/community/crm/components/molecules/AddTaskModalContent/AddTaskModalContent.tsx
```diff
@@ -140,177 +53,56 @@ const AddTaskModalContent: React.FC = () => {
...
+      typeId: formValues.type?.id ?? undefined,
       dueAt: formValues.dueDate,
       priority: formValues.priority,
-      contactName: formValues.contactName?.trim(),
-      deal: formValues.deal?.trim(),
-      owner: formValues.owner,
+      contactId: formValues.contactId ?? undefined,
+      dealId: formValues.dealId ?? undefined,
+      ownerId: formValues.owner ?? undefined,
       notes: formValues.notes?.trim()
```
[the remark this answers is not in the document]
charith-rc: as per current requirement, a task can be added without deal or contact. BE validation for task targets will be removed if the requirement holds to this condition.

frontend/src/community/crm/components/molecules/AddTaskModalContent/AddTaskModalContent.tsx
```diff
@@ -140,177 +53,56 @@ const AddTaskModalContent: React.FC = () => {
...
     };
 
     createNewTask(payload);
   };
```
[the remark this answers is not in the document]
charith-rc: fixed https://github.com/SkappHQ/skapp/pull/1466/commits/e5c2d03c85efe86417897c24b9cc69c0a022ba5c

ThinuwanW (commented):
## 🤖 Claude Code Review

This PR refactors the Add Task modal by extracting form logic into a `useTaskFormLogic` hook and a presentational `TaskFormTemplate`, wires up real contact/deal/owner lookups, and connects the create-task API. The decomposition is clean and type-safe overall, but the refactor introduces a behavioral regression: `setSubmitting(false)` was dropped from the error handler, so after a failed submission the form's buttons stay permanently disabled. There are also a few i18n/validation-message inconsistencies and an eager-fetch / timezone concern worth addressing.

Found **13** new issue(s): 🔴 1 important, 🟡 9 suggestion(s), 🟣 3 nit(s)

frontend/src/community/crm/components/molecules/AddTaskModalContent/AddTaskModalContent.tsx
```diff
@@ -116,7 +37,6 @@ const AddTaskModalContent: React.FC = () => {
...
 
   const handleError = () => {
```
ThinuwanW, line 39: 🤖 **Claude** · 🔴 **Important**

Behavioral regression: the old `handleError` (and `handleSuccess`) called `setSubmitting(false)`, but the refactor removed it and `useTaskFormLogic` no longer exposes/calls it. Formik's `onSubmit` here (`createTask`) returns `undefined` (synchronous — it just fires the mutation), and Formik does NOT auto-reset `isSubmitting` when `onSubmit` returns undefined (it only dispatches SUBMIT_SUCCESS when a Promise is returned). On success the form is reset via `handleCloseModal`→`resetForm()` so `isSubmitting` clears, but on ERROR nothing resets it. Result: after a failed create, `isSubmitting` stays `true`, so both Save (`disabled={isSubmitting || isPending}`) and Cancel (`disabled={isSubmitting}`) buttons remain permanently disabled and the user is stuck. Fix: expose `setSubmitting` from `useTaskFormLogic` and call `setSubmitting(false)` in `handleError`, OR return the mutation promise from `onSubmit` (use `mutateAsync`), OR reset submitting state inside the hook's error path.

frontend/src/community/crm/components/templates/TaskFormTemplate/TaskFormTemplate.tsx
```diff
@@ -0,0 +1,253 @@
...
+        selectedLabel={selectedContactLabel}
+        searchText={contactSearchText}
+        items={contactDropdownItems}
+        onSelect={onContactSelect}
+        onSearchChange={onContactSearchChange}
+        onClear={onClearContact}
+        isFetching={isContactFetching}
+        emptyMessage={translateText(["emptyStates", "noContacts"])}
+        inputAriaLabel={translateText(["ariaLabels", "contactName"])}
+        clearAriaLabel={translateText(["ariaLabels", "clearContact"])}
```
ThinuwanW, line 190: 🤖 **Claude** · 🟡 **Suggestion**

Missing i18n keys for accessibility labels. `SelectableSearchField` is given `clearAriaLabel={translateText(["ariaLabels", "clearContact"])}` (contact) and `["ariaLabels", "clearDeal"]` (deal), but `crmModule.json` → `tasks.addTaskModal.ariaLabels` only defines `contactName`, `deal`, `taskOwner`, `removeOwner`, etc. — there is no `clearContact` or `clearDeal`. The clear-button `aria-label` will fall back to the raw key string, degrading screen-reader output. Add `clearContact`/`clearDeal` keys under `ariaLabels` in crmModule.json (and other locale files).

frontend/src/community/crm/utils/taskValidations.ts
```diff
@@ -13,7 +13,7 @@ export const addTaskValidations = (translator: TranslatorFunctionType) =>
       .trim()
       .required(translator(["validations", "name"]))
       .max(
-        characterLengths.NAME_LENGTH,
+        characterLengths.TASK_NAME_LENGTH,
```
ThinuwanW, line 16: 🤖 **Claude** · 🟡 **Suggestion**

Stale validation message after the length change. The max was changed from `characterLengths.NAME_LENGTH` (50) to `TASK_NAME_LENGTH` (255), but the translated error message `validations.nameLength` in crmModule.json still reads "Task name must be less than 50 characters". A user typing 60–255 chars now passes validation, yet the message (when they exceed 255) still says 50. Update the `nameLength` translation to reflect 255 (and verify other locales).

frontend/src/community/crm/components/molecules/AddTaskModalContent/AddTaskModalContent.tsx
```diff
@@ -140,177 +53,58 @@ const AddTaskModalContent: React.FC = () => {
   const createTask = (formValues: CrmTaskAddFormTypes) => {
     const payload: CrmTaskCreatePayload = {
       name: formValues.name.trim(),
-      type: formValues.type,
-      dueAt: formValues.dueDate,
+      typeId: formValues.type?.id ?? undefined,
+      dueAt: formValues.dueDate
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Possible timezone day-shift in `dueAt`. `dueDate` is stored via `date.toISOString()` (UTC) in the DatePicker handler, then here `formValues.dueDate.split(".")[0]` strips the milliseconds and the trailing `Z`, sending a naive datetime that is actually the UTC wall-clock. For users east of UTC, local midnight of the selected day converts to the previous day in UTC (e.g. GMT+5:30 picking Jun 14 → `2026-06-13T18:30:00`), so the backend stores the wrong calendar day — while the UI display (`toLocaleDateString`) and the not-backdated Yup validation both use the local date, creating an inconsistency. Consider formatting the local date directly (e.g. date-fns `format(date, "yyyy-MM-dd'T'HH:mm:ss")`) instead of round-tripping through `toISOString()`.

frontend/src/community/crm/hooks/useTaskFormLogic.ts
```diff
@@ -0,0 +1,293 @@
...
+    useGetOwnerLookup(
+      debouncedOwnerSearch,
+      DEFAULT_LOOKUP_PAGE_SIZE,
+      Boolean(isCrmSalesManager)
+    );
+
+  const ownerLookupItems: CrmOwner[] = ownerLookupData?.items ?? [];
+
+  const { data: contactLookupData, isFetching: isContactFetching } =
+    useGetCrmContacts(debouncedContactSearch, DEFAULT_LOOKUP_PAGE_SIZE);
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Contact and deal lookups fire eagerly with an empty search keyword. `useGetCrmContacts(debouncedContactSearch, ...)` and `useGetDealLookup(debouncedDealSearch, ...)` default `enabled=true`, so on modal open (when `debouncedContactSearch`/`debouncedDealSearch` are `""`) they immediately fetch a full unfiltered list. Since `SearchableDropdown` only opens once the user has typed (`value.trim().length > 0`), these initial results are never shown — wasted requests. The owner lookup is correctly gated via its `enabled` flag; consider gating contacts/deals similarly, e.g. `enabled: debouncedContactSearch.length > 0`.

frontend/src/community/crm/api/crmDealApi.ts
```diff
@@ -73,3 +74,21 @@ export const useCreateDeal = (
...
 };
+
+export const useGetDealLookup = (
+  searchKeyword: string,
+  size: number,
+  enabled: boolean = true
+): UseQueryResult<CrmDealListItem[]> => {
+  return useQuery({
+    queryKey: crmDealQueryKeys.DEAL_LOOKUP(searchKeyword, size),
+    queryFn: async (): Promise<CrmDealListItem[]> => {
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Inconsistent lookup return shapes across the three lookups consumed by `useTaskFormLogic`: `useGetCrmContacts` returns `CrmContactLookupResponseType` (object with `.items`) and `useGetOwnerLookup` returns `CrmOwnersResponseType` (object with `.items`), but `useGetDealLookup` unwraps and returns a bare `CrmDealListItem[]`. Each consumer handles its own shape correctly, so this is functional, but aligning the deal lookup to also return the paginated response (or vice-versa) would make the lookup hooks easier to reason about and reuse.

frontend/src/community/crm/api/crmDealApi.ts
```diff
@@ -73,3 +74,21 @@ export const useCreateDeal = (
...
+      const response = await authFetch.get(crmDealEndpoints.GET_DEALS, {
+        params: { page: 0, size, searchKeyword }
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The new deal lookup calls GET_DEALS with only `{ page: 0, size, searchKeyword }`, but every other caller (useGetDealsInfinite) passes a full `CrmDealFilterParams` that requires `sortKey` and `sortOrder`. Because this params object is untyped, TS won't catch the omission. If the backend requires/defaults those sort params, the lookup may 400 or return an unexpectedly-ordered list. Pass explicit `sortKey`/`sortOrder` (or confirm the backend tolerates their absence) for parity with the infinite query.

frontend/src/community/crm/components/templates/TaskFormTemplate/TaskFormTemplate.tsx
```diff
@@ -0,0 +1,253 @@
+import {
```
ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Inconsistent placement of presentation-mapping logic: contact and deal dropdown items are built inside `useTaskFormLogic` (contactDropdownItems / dealDropdownItems), but the owner-to-SearchableDropdownItem mapping (`ownerDropdownItems`) is built here in the template instead. For consistency and testability, move the owner mapping into the hook and pass `ownerDropdownItems` like the others, rather than passing raw `ownerLookupItems`.

ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Naming convention: the file is `TaskFormTemplate.tsx` but the exported component is `TaskFormContent`, and `AddTaskModalContent` imports it under yet another local alias. Component name should match the file name (e.g. rename the component to `TaskFormTemplate`) so the PascalCase file/component convention holds and imports are unambiguous.

frontend/src/community/crm/components/templates/TaskFormTemplate/TaskFormTemplate.tsx
```diff
@@ -0,0 +1,253 @@
...
+  InputField,
+  TextArea
+} from "@rootcodelabs/skapp-ui";
+import { FC } from "react";
+
+import SearchableDropdown, {
+  SearchableDropdownItem
+} from "~community/common/components/molecules/SearchableDropdown/SearchableDropdown";
+import SelectableSearchField from "~community/crm/components/atoms/SelectableSearchField/SelectableSearchField";
+import SelectedOwnerField from "~community/crm/components/molecules/SelectedOwnerField/SelectedOwnerField";
```
ThinuwanW, line 15: 🤖 **Claude** · 🟣 **Nit**

`import OwnerDropdownItem from "../../atoms/OwnerDropdownItem/OwnerDropdownItem"` uses a deep relative path while every other import in this file uses the `~community/...` alias. Switch to `~community/crm/components/atoms/OwnerDropdownItem/OwnerDropdownItem` to follow the path-alias convention and stay consistent within the file.

frontend/src/community/crm/types/FormTypes.ts
```diff
@@ -0,0 +1,64 @@
+import { FormikErrors } from "formik";
```
ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

`TaskFormContentProps` duplicates almost the entire surface of `TaskFormLogicReturn` in useTaskFormLogic.ts (same values/errors/handleChange/setFieldValue and all owner/contact/deal/options fields, just renamed handler keys). This is type duplication that will drift. Consider deriving the props from the hook's return type (e.g. a shared base type, or `Pick`/`Omit` + a small rename mapping) so the two stay in sync.

frontend/src/community/crm/components/atoms/SelectableSearchField/SelectableSearchField.tsx
```diff
@@ -0,0 +1,86 @@
...
+}
+
+const SelectableSearchField: FC<SelectableSearchFieldProps> = ({
+  id,
+  label,
+  placeholder,
+  selectedValue,
+  selectedLabel,
+  searchText,
+  items,
```
ThinuwanW, line 33: 🤖 **Claude** · 🟡 **Suggestion**

`isFetching` is plumbed all the way through but only used to suppress the empty message (`emptyMessage={isFetching ? undefined : ...}`). During an in-flight lookup with no results yet, the dropdown closes entirely (SearchableDropdown's isDropdownOpen requires items or emptyMessage), giving the user no 'Searching…' feedback and causing a flicker between keystroke and results. Render an explicit loading item when `isFetching` instead of `undefined`.

frontend/src/community/crm/hooks/useTaskFormLogic.ts
```diff
@@ -0,0 +1,293 @@
...
+    shouldValidate?: boolean
+  ) => void;
+  submitForm: () => Promise<void>;
+  isSubmitting: boolean;
+  resetForm: () => void;
+
+  // Owner state & handlers
+  selectedOwner: CrmOwner | null;
+  ownerSearchText: string;
+  ownerLookupItems: CrmOwner[];
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`authPic: currentUser.authPic as string | null` uses an `as` assertion to silence the type rather than narrowing it. Strict mode discourages assertions; prefer typing `authPic` correctly on the user-details response (or a runtime check/normalization) so the cast isn't needed.

PR #1468, opened 2026-06-14
feat: add endpoint to check if a deal name exists and implement corre…

anusham-rc:
…sponding service logic

### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmDealController.java
```diff
@@ -24,6 +25,15 @@ public class CrmDealController {
...
 	private final CrmDealService crmDealService;
 
+	@Operation(summary = "Check if a deal name exists",
+			description = "Check if a deal with the given name already exists")
+	@GetMapping("/exists")
+	@PreAuthorize("hasAnyRole('ROLE_CRM_SALES_REPRESENTATIVE')")
+	public ResponseEntity<ResponseEntityDto> checkDealNameExists(@RequestParam String name) {
+		ResponseEntityDto responseDto = crmDealService.checkDealNameExists(name);
+		return new ResponseEntity<>(responseDto, HttpStatus.OK);
+	}
```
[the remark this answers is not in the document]
anusham-rc: integration tests added

ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds a GET /v1/crm/deal/exists endpoint to check whether a deal name already exists, mirroring the established CrmCompany name-exists pattern (controller → service interface → impl → derived Dao query → response DTO) with matching security, transaction, logging, validation, and DTO conventions. The derived query existsByNameIgnoreCaseAndIsDeletedFalse correctly maps to the CrmDeal entity's name/isDeleted fields, and integration test coverage is thorough (found/not-found, case-insensitivity, soft-deleted, blank, and too-long cases). The implementation is consistent and well-tested; the only notable point is that deal creation does not enforce uniqueness server-side the way company creation does.

Found **4** new issue(s):

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
```diff
@@ -70,6 +71,20 @@ public class CrmDealServiceImpl implements CrmDealService {
 
 	private final UserService userService;
 
+	@Override
```
ThinuwanW, line 74: 🤖 **Claude** · 💬 **Note**

Asymmetry with the company flow: createCompany calls checkCompanyExists and throws CRM_ERROR_COMPANY_EXISTS to enforce uniqueness server-side, but createDeal (lines 88+) performs no equivalent name-exists check. This new endpoint is therefore advisory only — two deals with the same name can still be created (including via a race after the frontend check passes). If deal names are intended to be unique, add a checkDealNameExists/existsByNameIgnoreCaseAndIsDeletedFalse guard inside createDeal that throws a ModuleException (with a new CRM_ERROR_DEAL_EXISTS message constant). If duplicate deal names are intentionally allowed, this is fine as-is — flagging so the developer can confirm the intended behavior.

backend/src/main/java/com/skapp/community/crmplanner/payload/response/CrmDealNameExistsResponseDto.java
```diff
@@ -0,0 +1,12 @@
+package com.skapp.community.crmplanner.payload.response;
+
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+public class CrmDealNameExistsResponseDto {
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

Code duplication (guideline checklist #9). `CrmDealNameExistsResponseDto` is byte-for-byte identical to `CrmCompanyNameExistsResponseDto` (single `Boolean isExists` field), and `WorkLocationNameAvailabilityResponseDto` is the same shape too — there are now three identical name-exists DTOs. Consider consolidating into one shared `NameExistsResponseDto` (or a small generic availability DTO) in a common payload package and reusing it across CRM deal/company and work-location flows. This is consistent with the existing per-entity convention, so it is a suggestion rather than a blocker.

backend/src/test/java/com/skapp/community/crmplanner/controller/v1/CrmDealControllerIntegrationTest.java
```diff
@@ -117,6 +130,93 @@ private CrmDealCreateRequestDto validPayload(Long stageId, Long contactId) {
...
+			.andExpect(jsonPath(RESULTS_0_PATH + "['message']")
+				.value(messageUtil.getMessage(CrmMessageConstant.CRM_ERROR_DEAL_NAME_REQUIRED)));
+	}
+
+	@Test
+	@DisplayName("Check deal name exists with name exceeding max length - Returns Bad Request")
+	void checkDealNameExists_NameTooLong_ReturnsBadRequest() throws Exception {
+		String tooLongName = "A".repeat(256);
+		performGetExistsRequest(tooLongName).andDo(print())
+			.andExpect(status().isBadRequest())
```
ThinuwanW, line 214: 🤖 **Claude** · 💬 **Note**

Test precision nit: `checkDealNameExists_NameTooLong_ReturnsBadRequest` only asserts `STATUS_UNSUCCESSFUL` and not the specific message. Because `validateDealName` checks length before the regex, a 256-char 'A' string also matches the regex, so the test would still pass even if the wrong validation branch fired. Mirror the blank-name test and assert `messageUtil.getMessage(CrmMessageConstant.CRM_ERROR_DEAL_NAME_TOO_LONG)` to lock down the intended failure path.

backend/src/main/java/com/skapp/community/crmplanner/repository/CrmDealDao.java
```diff
@@ -12,6 +12,8 @@ public interface CrmDealDao extends JpaRepository<CrmDeal, Long>, CrmDealReposit
 
 	Optional<CrmDeal> findByIdAndIsDeletedFalse(Long id);
 
+	boolean existsByNameIgnoreCaseAndIsDeletedFalse(String name);
```
ThinuwanW, line 15: 🤖 **Claude** · 💬 **Note**

Performance nit (uncertain — flagging for the developer to decide). `existsByNameIgnoreCaseAndIsDeletedFalse` generates a case-insensitive `LOWER(name) = LOWER(?)` predicate on `crm_deal.name`, which has no index (entity declares only `@Column(name = "name", nullable = false)`). If the frontend calls this as a typeahead/on-blur uniqueness check it could run frequently and force a full scan as the table grows. The existing company endpoint shares this pattern, so it is not newly introduced; if it becomes hot, consider a functional/lower-case index via a Liquibase changeset.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
anusham-rc: Added required validation while create a deal

backend/src/main/java/com/skapp/community/crmplanner/payload/response/CrmDealNameExistsResponseDto.java
anusham-rc: added common dto CrmNameExistsResponseDto to remove duplication

backend/src/test/java/com/skapp/community/crmplanner/controller/v1/CrmDealControllerIntegrationTest.java
anusham-rc: added test validation

backend/src/main/java/com/skapp/community/crmplanner/repository/CrmDealDao.java
anusham-rc: we agreed to use TEXT for these columns, can not apply above solution, will add front end denounce on api integration

akilarootcode (dismissed):

ThinuwanW (dismissed):

PR #1470, opened 2026-06-15
Reverting contact no encryption and laying the new encryption system

ThinuwanW:
…t-on-read

### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/backend/pom.xml
+++ b/backend/pom.xml
@@ -283,6 +283,17 @@
             <artifactId>cloudfront</artifactId>
             <version>2.42.0</version>
         </dependency>
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>kms</artifactId>
+            <version>2.42.0</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>commons-logging</groupId>
+                    <artifactId>commons-logging</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
 
         <!-- Apache PDFBox -->
         <dependency>
--- a/backend/src/main/java/com/skapp/community/common/exception/GlobalExceptionHandler.java
+++ b/backend/src/main/java/com/skapp/community/common/exception/GlobalExceptionHandler.java
@@ -166,15 +166,6 @@ public class GlobalExceptionHandler {
 
 	@ExceptionHandler(DataAccessException.class)
 	public ResponseEntity<ResponseEntityDto> handleDataAccessException(DataAccessException e) {
-		// A field-encryption (or other server-side) fault thrown inside a JPA flush/load
-		// is
-		// rewrapped by Hibernate as a DataAccessException — unwrap it so the original,
-		// coded error surfaces instead of a generic database error.
-		InternalServerException internalServerException = findInternalServerException(e);
-		if (internalServerException != null) {
-			return handleInternalServerException(internalServerException);
-		}
-
 		HttpStatus status = HttpStatus.INTERNAL_SERVER_ERROR;
 		handleException(e, CommonMessageConstant.COMMON_ERROR_DATABASE_ERROR.name(), status);
 
@@ -186,14 +177,6 @@ public class GlobalExceptionHandler {
 
 	@ExceptionHandler(Exception.class)
 	public ResponseEntity<ResponseEntityDto> handleExceptions(Exception e) {
-		// Same unwrap for any other wrapper (e.g. a raw PersistenceException) that buries
-		// an
-		// InternalServerException in its cause chain.
-		InternalServerException internalServerException = findInternalServerException(e);
-		if (internalServerException != null) {
-			return handleInternalServerException(internalServerException);
-		}
-
 		HttpStatus status = HttpStatus.INTERNAL_SERVER_ERROR;
 		handleException(e, CommonMessageConstant.COMMON_ERROR_MODULE_EXCEPTION.name(), status);
 
@@ -201,20 +184,6 @@ public class GlobalExceptionHandler {
 				new ResponseEntityDto(true,
 						new ErrorResponse(status, e.getMessage(), CommonMessageConstant.COMMON_ERROR_MODULE_EXCEPTION)),
 				status);
-	}
-
-	/**
-	 * Walks the cause chain looking for an {@link InternalServerException}. Hibernate and
-	 * Spring wrap exceptions thrown during flush/load, so the original is rarely the
-	 * top-level type seen by these handlers.
-	 */
-	private InternalServerException findInternalServerException(Throwable throwable) {
-		for (Throwable current = throwable; current != null; current = current.getCause()) {
-			if (current instanceof InternalServerException internalServerException) {
-				return internalServerException;
-			}
-		}
-		return null;
 	}
 
 	@ExceptionHandler(ServletException.class)
--- a/backend/src/main/java/com/skapp/community/common/exception/GlobalExceptionHandler.java
+++ b/backend/src/main/java/com/skapp/community/common/exception/GlobalExceptionHandler.java
@@ -107,6 +107,15 @@ public class GlobalExceptionHandler {
 				new ResponseEntityDto(true, new ErrorResponse(status, e.getMessage(), e.getMessageKey())), status);
 	}
 
+	@ExceptionHandler(InternalServerException.class)
+	public ResponseEntity<ResponseEntityDto> handleInternalServerException(InternalServerException e) {
+		HttpStatus status = HttpStatus.INTERNAL_SERVER_ERROR;
+		handleException(e, e.getMessageKey().name(), status);
+
+		return new ResponseEntity<>(
+				new ResponseEntityDto(true, new ErrorResponse(status, e.getMessage(), e.getMessageKey())), status);
+	}
+
 	@ExceptionHandler(EntityNotFoundException.class)
 	public ResponseEntity<ResponseEntityDto> handleEntityNotFoundExceptions(EntityNotFoundException e) {
 		HttpStatus status = HttpStatus.NOT_FOUND;
@@ -157,6 +166,15 @@ public class GlobalExceptionHandler {
 
 	@ExceptionHandler(DataAccessException.class)
 	public ResponseEntity<ResponseEntityDto> handleDataAccessException(DataAccessException e) {
+		// A field-encryption (or other server-side) fault thrown inside a JPA flush/load
+		// is
+		// rewrapped by Hibernate as a DataAccessException — unwrap it so the original,
+		// coded error surfaces instead of a generic database error.
+		InternalServerException internalServerException = findInternalServerException(e);
+		if (internalServerException != null) {
+			return handleInternalServerException(internalServerException);
+		}
+
 		HttpStatus status = HttpStatus.INTERNAL_SERVER_ERROR;
 		handleException(e, CommonMessageConstant.COMMON_ERROR_DATABASE_ERROR.name(), status);
 
@@ -168,6 +186,14 @@ public class GlobalExceptionHandler {
 
 	@ExceptionHandler(Exception.class)
 	public ResponseEntity<ResponseEntityDto> handleExceptions(Exception e) {
+		// Same unwrap for any other wrapper (e.g. a raw PersistenceException) that buries
+		// an
+		// InternalServerException in its cause chain.
+		InternalServerException internalServerException = findInternalServerException(e);
+		if (internalServerException != null) {
+			return handleInternalServerException(internalServerException);
+		}
+
 		HttpStatus status = HttpStatus.INTERNAL_SERVER_ERROR;
 		handleException(e, CommonMessageConstant.COMMON_ERROR_MODULE_EXCEPTION.name(), status);
 
@@ -175,6 +201,20 @@ public class GlobalExceptionHandler {
 				new ResponseEntityDto(true,
 						new ErrorResponse(status, e.getMessage(), CommonMessageConstant.COMMON_ERROR_MODULE_EXCEPTION)),
 				status);
+	}
+
+	/**
+	 * Walks the cause chain looking for an {@link InternalServerException}. Hibernate and
+	 * Spring wrap exceptions thrown during flush/load, so the original is rarely the
+	 * top-level type seen by these handlers.
+	 */
+	private InternalServerException findInternalServerException(Throwable throwable) {
+		for (Throwable current = throwable; current != null; current = current.getCause()) {
+			if (current instanceof InternalServerException internalServerException) {
+				return internalServerException;
+			}
+		}
+		return null;
 	}
 
 	@ExceptionHandler(ServletException.class)
--- a/backend/src/main/java/com/skapp/community/common/exception/InternalServerException.java
+++ b/backend/src/main/java/com/skapp/community/common/exception/InternalServerException.java
@@ -1,0 +1,56 @@
+package com.skapp.community.common.exception;
+
+import com.skapp.community.common.constant.MessageConstant;
+import com.skapp.community.common.util.MessageUtil;
+import lombok.Getter;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.ApplicationContextAware;
+import org.springframework.stereotype.Component;
+
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ * Signals a server-side / operational fault that is <b>not</b> the caller's fault — a
+ * required resource is unavailable or an internal invariant is broken (e.g. an encryption
+ * key failed to load, a stored value is corrupt). Surfaces as HTTP 500 through
+ * {@code GlobalExceptionHandler}, which also unwraps it when it is rethrown from inside a
+ * JPA flush/load and rewrapped by Hibernate.
+ *
+ * <p>
+ * Contrast with {@link ModuleException}, which maps to HTTP 400 (a client error). Use
+ * this type when the request itself is valid but the server cannot fulfil it. Carries a
+ * {@link MessageConstant} key so the message flows through the standard
+ * message-properties mechanism, keeping every error string in one place.
+ * </p>
+ */
+@Getter
+public class InternalServerException extends RuntimeException {
+
+	private static final AtomicReference<MessageUtil> messageUtil = new AtomicReference<>();
+
+	private final transient MessageConstant messageKey;
+
+	public InternalServerException(MessageConstant messageKey, Object... args) {
+		super(getMessageUtil().getMessage(messageKey.getMessageKey(), args));
+		this.messageKey = messageKey;
+	}
+
+	private static MessageUtil getMessageUtil() {
+		MessageUtil util = messageUtil.get();
+		if (util == null) {
+			throw new IllegalStateException("MessageUtil not initialized");
+		}
+		return util;
+	}
+
+	@Component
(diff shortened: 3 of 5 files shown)


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds a transitional JPA AttributeConverter (LegacyEncryptedContactNoConverter) that stops encrypting the organization contact_no on write while still decrypting legacy "ENC:" values on read, paired with the EpOrganization entity switching to it. The converter logic is correct and faithful: null-safe, the decrypt-on-read failure path is already wrapped in ModuleException inside EncryptionDecryptionServiceImpl, and the static setter-injection pattern matches the existing EncryptionDecryptionConverter. The main gaps are the absence of unit tests for the new behavior and some duplication/placement considerations given its transitional nature.

Found **5** new issue(s): 🟣 2 nit(s)

backend/src/main/java/com/skapp/community/common/util/converter/LegacyEncryptedContactNoConverter.java
```diff
@@ -0,0 +1,80 @@
+package com.skapp.community.common.util.converter;
```
ThinuwanW, line 1: 🤖 **Claude** · 💬 **Note**

📁 **File-level observation**

Missing unit tests for the new converter. The sibling EncryptionDecryptionConverter has a dedicated test class (EncryptionDecryptionConverterTest), but this converter introduces distinct behavior that is currently unverified: convertToDatabaseColumn must now return the value unchanged with NO "ENC:" prefix (encryption stopped), while convertToEntityAttribute must still (a) decrypt legacy "ENC:" values, (b) return plaintext as-is, and (c) handle null. These are exactly the behaviors the safe rollout depends on. Add a LegacyEncryptedContactNoConverterTest mirroring the existing test that asserts plaintext-on-write and decrypt-on-read for legacy values.

backend/src/main/java/com/skapp/community/common/util/converter/LegacyEncryptedContactNoConverter.java
```diff
@@ -0,0 +1,80 @@
...
+	 * migration has converted every stored value back to plaintext.
+	 */
+	@Override
+	public String convertToEntityAttribute(String dbData) {
+		if (dbData == null) {
+			return null;
+		}
+		if (dbData.startsWith(ENCRYPTED_PREFIX)) {
+			return encryptionDecryptionService.decrypt(dbData.substring(ENCRYPTED_PREFIX.length()));
+		}
```
ThinuwanW, line 49: 🤖 **Claude** · 💬 **Note**

convertToEntityAttribute, the ENCRYPTED_PREFIX constant, and the static EncryptionDecryptionService setter-injection are near-identical copies of EncryptionDecryptionConverter. Given the class is explicitly transitional/disposable this duplication is defensible, but consider extracting the shared decrypt-on-read helper (or at least the "ENC:" prefix constant) so the two converters cannot drift apart while both exist. Minor — acceptable to leave if the migration/removal lands soon.

backend/src/main/java/com/skapp/community/common/util/converter/LegacyEncryptedContactNoConverter.java
```diff
@@ -0,0 +1,80 @@
...
+ */
+@Converter
+@Component
+public class LegacyEncryptedContactNoConverter implements AttributeConverter<String, String> {
+
+	// Prefix that marks values written by the previous (encrypting) converter.
+	private static final String ENCRYPTED_PREFIX = "ENC:";
+
+	private static EncryptionDecryptionService encryptionDecryptionService;
+
```
ThinuwanW, line 23: 🤖 **Claude** · 💬 **Note**

Package/module placement: this converter exists solely to support an enterprise-only field (EpOrganization.contactNo) yet lives in the community module. It works because enterprise may depend on community and it only references the community EncryptionDecryptionService, and it is consistent with the existing converter's location, so this is a nit. If the intent is to keep enterprise-specific transitional code out of community, it could instead live under com.skapp.enterprise; otherwise leaving it here is fine.

backend/src/main/java/com/skapp/community/common/util/converter/LegacyEncryptedContactNoConverter.java
```diff
@@ -0,0 +1,80 @@
...
+ * revert migration in skapp-migrations.
+ * </p>
+ *
+ * @see EncryptionDecryptionConverter the generic encrypt-at-rest converter, retained for
+ * fields that genuinely require field-level encryption
+ */
+@Converter
+@Component
+public class LegacyEncryptedContactNoConverter implements AttributeConverter<String, String> {
+
```
ThinuwanW, line 18: 🤖 **Claude** · 🟣 **Nit**

Malformed Javadoc: `@see EncryptionDecryptionConverter the generic encrypt-at-rest converter...` — the class-reference form of `@see` does not take trailing label prose, so the trailing text is treated as part of the reference and will render incorrectly / may produce a javadoc warning. Use a separate sentence with `{@link EncryptionDecryptionConverter}` instead, e.g. a normal paragraph referencing `{@link EncryptionDecryptionConverter}` for fields that genuinely require field-level encryption. Severity: nit.

backend/src/main/java/com/skapp/community/common/util/converter/LegacyEncryptedContactNoConverter.java
ThinuwanW, line 1: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

Paired-change dependency not verifiable in-repo: correctness of this converter (and the eventual safe removal of the decrypt-on-read branch) hinges on the referenced "GDPR contact-number revert migration in skapp-migrations", which lives in a separate repository. There is no Liquibase changeset or in-repo guard ensuring that migration ships before/with this change, nor preventing premature removal of this converter while legacy ENC: rows still exist. Confirm the external migration is coordinated with this deploy, and consider adding a brief note/ticket reference tying the converter's removal to migration completion so a future cleanup doesn't render legacy rows unreadable. Severity: nit.

ThinuwanW: no need of ticket and anything since both this change and the migration will rollout together and the whole migration implementation is there in the migration repo pr

backend/src/main/java/com/skapp/community/common/util/converter/LegacyEncryptedContactNoConverter.java
ThinuwanW: true but keeping this way under following reasons
1. existing code is also in the community so this is something more same as it
2. this and the existing code both will remove after completing the migration

backend/src/main/java/com/skapp/community/common/util/converter/LegacyEncryptedContactNoConverter.java
ThinuwanW: this is valid and AI generated comment was malformed.
fixed.

backend/src/main/java/com/skapp/community/common/util/converter/LegacyEncryptedContactNoConverter.java
ThinuwanW: valid
but this file added while keeping the existing as it is and this is a temp code which will remove soon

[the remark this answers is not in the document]
ThinuwanW: valid point but this is used in contact number where this is not possible. in the new implementation I'll add a guard and not in this since this is a temp code

backend/src/main/java/com/skapp/community/common/util/converter/LegacyEncryptedContactNoConverter.java
```diff
@@ -0,0 +1,80 @@
...
+	 */
+	@Override
+	public String convertToEntityAttribute(String dbData) {
+		if (dbData == null) {
+			return null;
+		}
+		if (dbData.startsWith(ENCRYPTED_PREFIX)) {
+			return encryptionDecryptionService.decrypt(dbData.substring(ENCRYPTED_PREFIX.length()));
+		}
+		return dbData;
```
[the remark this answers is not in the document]
ThinuwanW: added

backend/src/main/java/com/skapp/community/common/util/converter/LegacyEncryptedContactNoConverter.java
ThinuwanW: added

backend/src/main/java/com/skapp/community/common/util/converter/LegacyEncryptedContactNoConverter.java
```diff
@@ -0,0 +1,80 @@
...
+ * (and ultimately this whole converter) can be removed — see the GDPR contact-number
+ * revert migration in skapp-migrations.
+ * </p>
+ *
+ * @see EncryptionDecryptionConverter the generic encrypt-at-rest converter, retained for
+ * fields that genuinely require field-level encryption
+ */
+@Converter
+@Component
+public class LegacyEncryptedContactNoConverter implements AttributeConverter<String, String> {
```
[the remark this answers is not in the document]
ThinuwanW: this is used in submodule so we can take this as an invalid comment

backend/src/main/java/com/skapp/community/common/util/converter/LegacyEncryptedContactNoConverter.java
```diff
@@ -0,0 +1,80 @@
+package com.skapp.community.common.util.converter;
+
+import com.skapp.community.common.service.EncryptionDecryptionService;
+import jakarta.persistence.AttributeConverter;
+import jakarta.persistence.Converter;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+/**
+ * Transitional JPA AttributeConverter for the organization {@code contact_no} field.
```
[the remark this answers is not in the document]
ThinuwanW: not valid 
submodule probem

ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds an InternalServerException type (HTTP 500) with cause-chain unwrapping in GlobalExceptionHandler to surface coded server-side faults that Hibernate rewraps during flush/load, plus a transitional LegacyEncryptedContactNoConverter that stops encrypting contact_no on write while still decrypting legacy ENC: values on read. The changes are clean, consistent with existing patterns, and well-tested. Findings are minor: an asymmetry where the legacy converter's decrypt failures aren't covered by the new unwrap mechanism, code duplication between the two coded-exception types, and an unguarded cause-chain walk.

Found **6** new issue(s): 

backend/src/main/java/com/skapp/community/common/util/converter/LegacyEncryptedContactNoConverter.java
```diff
@@ -0,0 +1,82 @@
...
+	 * "ENC:" prefix, so the application keeps reading correct values until the data
+	 * migration has converted every stored value back to plaintext.
+	 */
+	@Override
+	public String convertToEntityAttribute(String dbData) {
+		if (dbData == null) {
+			return null;
+		}
+		if (dbData.startsWith(ENCRYPTED_PREFIX)) {
+			return encryptionDecryptionService.decrypt(dbData.substring(ENCRYPTED_PREFIX.length()));
```
ThinuwanW, line 48: 🤖 **Claude** · 💬 **Note**

This converter calls the community EncryptionDecryptionService.decrypt(), which throws ModuleException (mapped to HTTP 400) on failure — not InternalServerException. Since this runs inside Hibernate result-set mapping, a decrypt failure on a legacy ENC: row gets rewrapped as a DataAccessException, and the new findInternalServerException() unwrap only looks for InternalServerException. Net effect: the coded COMMON_ERROR_DECRYPTION_FAILED message is never surfaced — the client gets a generic COMMON_ERROR_DATABASE_ERROR 500, and semantically a server-side decrypt fault is being modeled as a client error (400). Consider wrapping decrypt failures here in InternalServerException so this path benefits from the very unwrap mechanism this PR adds, keeping it symmetric with the enterprise FieldEncryptionService path. Low impact since this is transitional code, but the infrastructure added doesn't actually cover this converter.

backend/src/main/java/com/skapp/community/common/exception/InternalServerException.java
```diff
@@ -0,0 +1,56 @@
+package com.skapp.community.common.exception;
```
ThinuwanW, line 1: 🤖 **Claude** · 💬 **Note**

📁 **File-level observation**

InternalServerException is a near-verbatim copy of ModuleException: same static AtomicReference<MessageUtil>, identical getMessageUtil() with the 'MessageUtil not initialized' guard, and a duplicate @Component MessageUtilInjector ApplicationContextAware inner class (now two beans doing the same MessageUtil wiring into separate static refs). Consider extracting the shared machinery into a common abstract base (e.g. AbstractCodedException holding messageKey + the MessageUtil lookup) so the two coded-exception types differ only in their HTTP-status semantics. Reduces the maintenance burden of keeping two injectors in sync.

backend/src/main/java/com/skapp/community/common/exception/GlobalExceptionHandler.java
```diff
@@ -177,6 +203,20 @@ public ResponseEntity<ResponseEntityDto> handleExceptions(Exception e) {
 				status);
 	}
 
+	/**
+	 * Walks the cause chain looking for an {@link InternalServerException}. Hibernate and
+	 * Spring wrap exceptions thrown during flush/load, so the original is rarely the
+	 * top-level type seen by these handlers.
+	 */
+	private InternalServerException findInternalServerException(Throwable throwable) {
+		for (Throwable current = throwable; current != null; current = current.getCause()) {
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

findInternalServerException() walks getCause() with no cycle/visited guard. A self-referential or cyclic cause chain would loop forever (vs. Spring's NestedExceptionUtils-style walkers that are cycle-safe). Cyclic chains are rare in practice, but consider a bounded loop or a visited set to be defensive, since this runs on every DataAccessException and unhandled Exception.

backend/src/main/java/com/skapp/community/common/exception/GlobalExceptionHandler.java
```diff
@@ -157,6 +166,15 @@ public ResponseEntity<ResponseEntityDto> handleSQLException(SQLException e) {
 
 	@ExceptionHandler(DataAccessException.class)
 	public ResponseEntity<ResponseEntityDto> handleDataAccessException(DataAccessException e) {
+		// A field-encryption (or other server-side) fault thrown inside a JPA flush/load
+		// is
+		// rewrapped by Hibernate as a DataAccessException — unwrap it so the original,
+		// coded error surfaces instead of a generic database error.
+		InternalServerException internalServerException = findInternalServerException(e);
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

The new unwrap helper recovers only InternalServerException from a Hibernate/Spring-wrapped cause chain — ModuleException is not unwrapped. This is asymmetric and directly bites the new LegacyEncryptedContactNoConverter: its convertToEntityAttribute() calls the community EncryptionDecryptionService.decrypt(), which throws ModuleException on a corrupt legacy 'ENC:' value during a JPA load. When Hibernate rewraps that into a DataAccessException, findInternalServerException() returns null and the response degrades to a generic 500 COMMON_ERROR_DATABASE_ERROR, losing the coded decryption message. Either also walk the chain for ModuleException (returning its mapped 400/coded response) or align the converter to throw InternalServerException so the intended coded 500 surfaces. As written the unwrap only helps the enterprise FieldEncryption path, not the converter this PR adds.

backend/src/test/java/com/skapp/community/common/util/converter/LegacyEncryptedContactNoConverterTest.java
```diff
@@ -0,0 +1,104 @@
+package com.skapp.community.common.util.converter;
```
ThinuwanW, line 1: 🤖 **Claude** · 💬 **Note**

📁 **File-level observation**

The converter's entire reason to exist is safely handling legacy encrypted values on read, yet the test only exercises happy paths (plaintext passthrough and successful legacy decrypt). There is no test for a value that carries the 'ENC:' prefix but whose payload is corrupt/non-Base64 — i.e. the failure path that motivated this PR's exception-handling changes. Add a test asserting the expected exception/behavior for a malformed 'ENC:'-prefixed value, so the decrypt-failure contract (and any future change to which exception type is thrown) is locked down.

backend/pom.xml
```diff
@@ -283,6 +283,17 @@
             <artifactId>cloudfront</artifactId>
             <version>2.42.0</version>
         </dependency>
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>kms</artifactId>
+            <version>2.42.0</version>
```
ThinuwanW, line 289: 🤖 **Claude** · 💬 **Note**

The new KMS dependency hardcodes version 2.42.0, the third software.amazon.awssdk artifact to repeat that literal (alongside the existing s3/cloudfront entries) with no shared version. This invites version drift across AWS SDK modules. Consider importing software.amazon.awssdk:bom in <dependencyManagement> (as is already done for azure-sdk-bom) and dropping the per-artifact <version>, so all AWS SDK artifacts stay aligned. Nit — consistent with current style, but the BOM avoids future skew.

ThinuwanW: already noted and is a separate improvement

backend/src/test/java/com/skapp/community/common/util/converter/LegacyEncryptedContactNoConverterTest.java
ThinuwanW: this operation is actively monitored and temp code so no need for edge cases

backend/src/main/java/com/skapp/community/common/exception/GlobalExceptionHandler.java
ThinuwanW: LegacyEncryptedContactNoConverter is temp code and have only basic exception handling

backend/src/main/java/com/skapp/community/common/exception/GlobalExceptionHandler.java
ThinuwanW: This new implementation is unnecessary for the existing code so reverted and kept the original way

backend/src/main/java/com/skapp/community/common/exception/InternalServerException.java
ThinuwanW: True but lets keep the pattern as it is and take as a separate improvement

backend/src/main/java/com/skapp/community/common/util/converter/LegacyEncryptedContactNoConverter.java
ThinuwanW: skipping since the temp code

akilarootcode (approved):

PR #1477, opened 2026-06-15
Fix/2499 filter panel UI issues

shakila-kamalasena:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/frontend/src/community/leave/components/molecules/RequestFilterMenuItems/RequestFilterMenuItems.tsx
+++ b/frontend/src/community/leave/components/molecules/RequestFilterMenuItems/RequestFilterMenuItems.tsx
@@ -213,6 +213,7 @@ const RequestFilterMenuItems = ({
             handleFilters(button, menuItemsData[1].type);
           }
         }}
+        className="max-h-full"
       />
     </BasicFilterStructure>
   );


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds a single Tailwind class `max-h-full` to the leave-type `SelectableItemList` to constrain its height (per the commit message, fixing the filter menu height). The change is low-risk, isolated, and consistent with how `SelectableItemList` is used elsewhere. No correctness, type-safety, or i18n concerns introduced.

frontend/src/community/leave/components/molecules/RequestFilterMenuItems/RequestFilterMenuItems.tsx
```diff
@@ -213,6 +213,7 @@ const RequestFilterMenuItems = ({
             handleFilters(button, menuItemsData[1].type);
           }
         }}
+        className="max-h-full"
```
ThinuwanW, line 216: 🤖 **Claude** · 🟣 **Nit**

`max-h-full` resolves to `max-height: 100%`, which only has an effect if the immediate parent rendered by `SelectableItemList`/`BasicFilterStructure` has an explicitly bounded height; otherwise it is a no-op and the intended scroll/clipping fix won't apply. Worth confirming in the browser that the type list actually clamps and scrolls. If the goal is a fixed cap, a concrete value (e.g. `max-h-[300px]`) plus `overflow-y-auto` would be more robust. Nit since this may already behave correctly given the surrounding layout.

ThinuwanW, line 216: 🤖 **Claude** · 🟡 **Suggestion**

`max-h-full` constrains height but provides no overflow behavior, so even when the max-height resolves correctly a long leave-type list will be clipped rather than scrollable. Comparable filter sections in this codebase pair height/structure with an explicit overflow utility (e.g. JobFamilies/UserRoles/Employment sections wrap their lists in `overflow-y-auto`). If the intent is a bounded scrollable area, this needs an accompanying `overflow-y-auto` (and a concrete max-height like `max-h-[NNpx]` rather than `max-h-full`, which depends on a constrained parent). Confirm `SelectableItemList` forwards `className` to its scroll container; if it doesn't, this class has no effect at all.

shakila-kamalasena: This is the needed behaviour

shakila-kamalasena: The parent has full overflow

thusala (approved):

ThinuwanW (approved):

PR #1484, opened 2026-06-16
Feat/2472 add skills to profile

shakila-kamalasena:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds an employee Skills feature (default skills from a bundled YAML + user-defined custom skills) across backend (new entities, DAO, service, controller, DDL) and frontend (ChipAutocomplete, SkillsDetailsSection, API hook, change detection). The architecture and conventions are mostly sound, but there is a critical defect: skills are persisted via a direct DAO save during employee creation while the Employee is still transient, which will fail for any new employee created with skills. Several secondary issues exist around a dead response field, an unused message constant, N+1 reads, and frontend/backend data duplication.

Found **17** new issue(s):

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
```diff
@@ -447,6 +450,11 @@ private Employee createEmployeeEntity(Employee employee, CreateEmployeeRequestDt
 			processEmployeeTeams(requestDto.getEmployment(), employee);
 		}
 
+		// Skills
+		if (requestDto != null && requestDto.getPersonal() != null) {
+			processEmployeeSkills(requestDto.getPersonal(), employee);
```
ThinuwanW, line 453: 🤖 **Claude** · 💬 **Note**

Critical: creating a new employee with skills will fail. processEmployeeSkills -> skillService.saveEmployeeSkills() runs inside createEmployeeEntity, which executes BEFORE the employee is persisted (userDao.save(user) is only called later at line 241). At this point employee.getEmployeeId() is null and the Employee is a transient instance. saveEmployeeSkills immediately calls employeeSkillDao.deleteByEmployeeEmployeeId(null) (no-op) and then employeeSkillDao.saveAll(...) where each EmployeeSkill references the transient Employee (no cascade on the @ManyToOne, employee_id is NOT NULL). Hibernate will throw a TransientObjectException / constraint violation and roll back the whole create. Every other process* helper (processEmployeeManagers, processEmployeeTeams, processEmployeeFamilies) instead mutates the employee's in-memory collection and relies on cascade at save — skills is the anomaly doing a direct DAO write. Fix: either move the skills save to after userDao.save(user) in createEmployee/updateEmployee, or model EmployeeSkill as a cascaded collection on Employee. Update works because the employee already has an ID; only the create path is broken.

backend/src/main/java/com/skapp/community/peopleplanner/payload/response/EmployeeDetailedResponseDto.java
```diff
@@ -92,4 +92,6 @@ public class EmployeeDetailedResponseDto {
 
 	private EmployeeRoleResponseDto employeeRole;
 
+	private List<SkillResponseDto> skills;
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

Dead field: the new `List<SkillResponseDto> skills` field is never populated. PeopleMapper marks it `@Mapping(target = "skills", ignore = true)` and no code calls setSkills on EmployeeDetailedResponseDto, so it is always null in every response that uses this DTO (directory listing, /me, search). The actual skills are returned only via EmployeePersonalDetailsDto in PeopleReadServiceImpl. Either remove this field (and the ignore mapping) or populate it where the DTO is built; leaving it ships a permanently-null field in the API contract.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/SkillServiceImpl.java
```diff
@@ -0,0 +1,112 @@
...
+		List<EmployeeSkill> employeeSkills = employeeSkillDao.findByEmployeeEmployeeId(employeeId);
+
+		return employeeSkills.stream().map(es -> {
+			if (es.getSkillType() == SkillType.DEFAULT) {
+				Optional<DefaultSkillLoader.DefaultSkill> defaultSkill = defaultSkillLoader.findById(es.getSkillId());
+				String name = defaultSkill.map(DefaultSkillLoader.DefaultSkill::name).orElse("Unknown");
+				return new SkillResponseDto(es.getSkillId(), name, SkillType.DEFAULT);
+			}
+			else {
+				Optional<CustomSkill> customSkill = customSkillDao.findById(es.getSkillId());
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

N+1 reads in getEmployeeSkillResponses: for every CUSTOM EmployeeSkill it calls customSkillDao.findById(...) individually inside the stream. An employee with many custom skills triggers one query per skill. Collect the custom skill IDs first and use customSkillDao.findAllById(ids) once, then build the response from the map. Same applies conceptually to the default lookup, though that one is in-memory.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/EmployeeSkillServiceImpl.java
```diff
@@ -0,0 +1,112 @@
...
+
+	private final DefaultSkillLoader defaultSkillLoader;
+
+	private final CustomSkillDao customSkillDao;
+
+	private final EmployeeSkillDao employeeSkillDao;
+
+	@Override
+	@Transactional
+	public List<EmployeeSkill> saveEmployeeSkills(Employee employee, List<EmployeeSkillDto> skills) {
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

Missing service entry logging. Per backend conventions new service methods should log entry (e.g. log.info("saveEmployeeSkills: execution started")). saveEmployeeSkills, getEmployeeSkillResponses, and getAllSkills have none. The class is already annotated @Slf4j but only logs a warn in one branch.

backend/src/main/java/com/skapp/community/peopleplanner/constant/PeopleMessageConstant.java
```diff
@@ -227,7 +227,8 @@ public enum PeopleMessageConstant implements MessageConstant {
 			"api.error.people.transfer-supervisors.team-record-not-found"),
 	PEOPLE_ERROR_EMPLOYEE_TERMINATION_OR_DELETION_ACTION_REQUIRED(
 			"api.error.people.employee-termination-or-deletion-action-required"),
-	PEOPLE_ERROR_VALIDATION_WORK_LOCATION_NOT_FOUND("api.error.people.validation-work-location-not-found");
+	PEOPLE_ERROR_VALIDATION_WORK_LOCATION_NOT_FOUND("api.error.people.validation-work-location-not-found"),
+	PEOPLE_ERROR_SKILL_NOT_FOUND("api.error.people.skill.not.found");
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

Unused constant: PEOPLE_ERROR_SKILL_NOT_FOUND (and its api.error.people.skill.not.found message) is added but never referenced anywhere. SkillServiceImpl handles missing skills with log.warn + a fallback name of "Unknown" rather than throwing a ModuleException with this key. Either wire it into the not-found handling (preferable for default-skill mismatches) or remove the constant and message to avoid dead code.

backend/src/main/java/com/skapp/community/peopleplanner/controller/v1/SkillController.java
```diff
@@ -0,0 +1,32 @@
...
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+@RequiredArgsConstructor
+@RequestMapping("/v1/people/skills")
+@Tag(name = "Skill Controller", description = "Endpoints for skill management")
+public class SkillController {
+
+	private final SkillService skillService;
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

Misleading API doc: the @Operation description says "Retrieves all default and custom skills" but getAllSkills() in SkillServiceImpl returns only custom skills (defaults live in skills.yml and are shipped to the frontend via the local skillConstants.ts file). Update the description to reflect that only custom skills are returned, or include defaults in the response.

frontend/src/community/people/constants/skillConstants.ts
```diff
@@ -0,0 +1,1003 @@
+export const SKILL_OPTIONS = [
```
ThinuwanW, line 1: 🤖 **Claude** · 💬 **Note**

📁 **File-level observation**

Data duplication / sync risk: the ~1000-entry default skill master list is maintained in two places — backend skills.yml and this frontend SKILL_OPTIONS constant — and the two must keep identical id↔name mappings for default-skill display to be correct (getSkillDisplayNames/findExistingSkill match SKILL_OPTIONS by id, while the backend resolves names from skills.yml by the same id). Any future edit to one file without the other silently corrupts displayed skill names. Consider a single source of truth (e.g., the frontend fetching the default list from the backend instead of bundling its own copy).

frontend/src/community/people/utils/peopleEditFlowUtils/personalDetailsChangesUtils.ts
```diff
@@ -362,6 +361,22 @@ export const getHealthAndOtherDetailsChanges = (
   return changes;
 };
 
+export const getSkillsDetailsChanges = (
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

getSkillsDetailsChanges compares skills positionally (newSkills.some((skill, index) => ... previousSkills[index])). The previous list comes from the backend sorted by name, while the new list reflects the user's insertion order, so the same set of skills in a different order is reported as a change (false-positive 'unsaved changes' / unnecessary update payload). Compare as order-independent sets keyed by skillType + (skillId|name) instead of by index.

frontend/src/community/people/api/PeopleApi.ts
```diff
@@ -77,6 +79,36 @@ export const useGetBannerData = (): UseQueryResult<number> => {
...
+    "skillsDetails"
+  );
+
+  return useQuery({
+    queryKey: peopleQueryKeys.GET_SKILLS,
+    queryFn: async () => authFetch.get(peoplesEndpoints.GET_SKILLS),
+    select: (response) =>
+      (response?.data?.results ?? []).map(
+        (skill: { id: number; name: string; skillType: string }) => ({
+          skillId: skill.id,
```
ThinuwanW, line 90: 🤖 **Claude** · 💬 **Note**

Minor type-safety / caching: the useGetSkills select callback types the raw item inline and uses `skill.skillType as SkillTypes` to coerce a string into the enum without validation (strict-mode `as` assertion the guidelines flag). Prefer a shared response interface and a safe mapping. Also, the skills master list is effectively static, but the query sets no staleTime, so it refetches on every mount/focus — consider staleTime: Infinity (or a long value).

backend/src/main/java/com/skapp/community/peopleplanner/model/EmployeeSkill.java
```diff
@@ -0,0 +1,40 @@
...
+
+	@Id
+	@ManyToOne(fetch = FetchType.LAZY)
+	@JoinColumn(name = "employee_id", nullable = false)
+	private Employee employee;
+
+	@Id
+	@Column(name = "skill_id", nullable = false)
+	private Long skillId;
+
```
ThinuwanW, line 35: 🤖 **Claude** · 💬 **Note**

Schema inconsistency (nit): the skill_type field uses columnDefinition = "text", but the Liquibase DDL defines skill_type as VARCHAR(255) (and it is part of the composite PK). Because Liquibase owns the schema, the entity columnDefinition is ignored at runtime, so this is harmless today but misleading. Align the entity columnDefinition with the actual column type (VARCHAR(255)).

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/EmployeeSkillServiceImpl.java
```diff
@@ -0,0 +1,112 @@
+package com.skapp.community.peopleplanner.service.impl;
```
ThinuwanW, line 1: 🤖 **Claude** · 💬 **Note**

📁 **File-level observation**

Testing: the new SkillServiceImpl (skill resolution, custom-skill create-or-reuse, delete-and-replace) and the ChipAutocomplete component have no accompanying tests. Per the repo testing conventions, service changes should ship with test coverage, especially given the create-time persistence ordering issue noted above which a service/integration test would have caught.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/EmployeeSkillServiceImpl.java
```diff
@@ -0,0 +1,112 @@
...
+				employeeSkill.setSkillId(skillDto.getSkillId());
+				employeeSkill.setSkillType(SkillType.DEFAULT);
+			}
+			else if (skillDto.getSkillType() == SkillType.CUSTOM) {
+				CustomSkill customSkill = findOrCreateCustomSkill(skillDto.getName());
+				employeeSkill.setSkillId(customSkill.getId());
+				employeeSkill.setSkillType(SkillType.CUSTOM);
+			}
+
+			employeeSkills.add(employeeSkill);
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

Unhandled bad-input path → 500. If skillDto.getSkillType() is null or any value other than DEFAULT/CUSTOM, neither branch sets skillId/skillType, yet the EmployeeSkill is still added to the list and passed to saveAll(). Persisting an EmployeeSkill with null skill_id/skill_type violates the NOT NULL columns (and the composite PK), throwing DataIntegrityViolationException that is not wrapped in a ModuleException → raw 500 to the client. Same risk for SkillType.CUSTOM with a null/blank name: findOrCreateCustomSkill(null) creates a CustomSkill with a null name and hits the NOT NULL constraint on ppl_custom_skill.name. EmployeeSkillDto has no validation (see related finding), so this is reachable from a malformed request. Fix: add an `else { log.warn("Unknown/empty skill type: {}", skillDto.getSkillType()); continue; }`, and guard against a blank name before findOrCreateCustomSkill (skip/throw ModuleException with PEOPLE_ERROR_SKILL_NOT_FOUND instead of inserting an invalid row).

backend/src/main/java/com/skapp/community/peopleplanner/payload/request/EmployeeSkillDto.java
```diff
@@ -0,0 +1,17 @@
...
+
+import com.skapp.community.peopleplanner.type.SkillType;
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+public class EmployeeSkillDto {
+
+	private Long skillId;
```
ThinuwanW, line 11: 🤖 **Claude** · 💬 **Note**

Missing bean validation. skillId/skillType/name are all unvalidated, and CreateEmployeeRequestDto's nested skills list is not marked @Valid, so no constraints are enforced. At minimum add @NotNull on skillType (the discriminator the service branches on) so a missing/invalid type is rejected by bean validation with a proper 400 rather than reaching the persistence layer. This is the validation source for the null-skillType crash noted in SkillServiceImpl.

backend/src/main/resources/community/db/changelog/ddl-script/people-ddl-script-v1-create-table-ppl-custom-skill.sql
```diff
@@ -0,0 +1,11 @@
+-- liquibase formatted sql
+
+-- changeset shakila:00116_create_table_ppl_custom_skill
+CREATE TABLE IF NOT EXISTS `ppl_custom_skill`
+(
+    `id`   bigint       NOT NULL AUTO_INCREMENT,
+    `name` text         NOT NULL,
```
ThinuwanW, line 7: 🤖 **Claude** · 💬 **Note**

No uniqueness guarantee on custom skill name. ppl_custom_skill.name has no unique index, but findOrCreateCustomSkill relies on findByNameIgnoreCase(...) returning at most one row (Optional<CustomSkill>). Two employee saves referencing the same new custom skill name (concurrent requests, or differing case before this method runs) will create duplicate rows; once duplicates exist, the Spring Data derived query throws IncorrectResultSizeDataAccessException on every subsequent lookup. Add a UNIQUE constraint on name (note: a TEXT column needs a key length in MySQL, e.g. switch to VARCHAR and add `CONSTRAINT UQ_ppl_custom_skill_name UNIQUE (name)`), or make findOrCreateCustomSkill resilient to multiple matches.

backend/src/main/java/com/skapp/community/peopleplanner/model/Employee.java
```diff
@@ -168,6 +168,9 @@ public class Employee extends Auditable<String> {
 	@JoinColumn(name = "work_location_id")
 	private WorkLocation workLocation;
 
+	@OneToMany(mappedBy = "employee", fetch = FetchType.LAZY)
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

Dead JPA relationship. The new `employeeSkills` @OneToMany collection is never read anywhere — the read path goes through EmployeeSkillDao.findByEmployeeEmployeeId() and writes go through EmployeeSkillDao.saveAll(); getEmployeeSkills() has no callers. It adds an unused lazy association to the aggregate. Either remove it, or use it consistently (and add it to the entity graph) if it is meant to be the access path.

backend/src/main/java/com/skapp/community/peopleplanner/component/DefaultSkillLoader.java
```diff
@@ -0,0 +1,57 @@
...
+		}
+	}
+
+	public Optional<DefaultSkill> findById(Long id) {
+		return defaultSkills.stream().filter(skill -> skill.id().equals(id)).findFirst();
+	}
+
+	public record DefaultSkill(Long id, String name) {
+	}
+
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

O(n) lookup on a ~1000-entry list. findById streams/filters the whole defaultSkills list on every call, and getEmployeeSkillResponses calls it once per DEFAULT skill an employee has. Back the loader with a Map<Long, DefaultSkill> built once in loadSkills() for O(1) lookups instead of a linear scan.

frontend/src/community/people/components/organisms/PersonDetailsSection/SubSections/SkillsDetailsSection.tsx
```diff
@@ -0,0 +1,121 @@
...
+import { forwardRef, useImperativeHandle } from "react";
+
+import PeopleLayout from "~community/common/components/templates/PeopleLayout/PeopleLayout";
+import { useTranslator } from "~community/common/hooks/useTranslator";
+import { SKILL_OPTIONS } from "~community/people/constants/skillConstants";
+import { SkillTypes } from "~community/people/enums/PeopleEnums";
+import { usePeopleStore } from "~community/people/store/store";
+import { FormMethods } from "~community/people/types/PeopleEditTypes";
+import { SkillType } from "~community/people/types/PeopleTypes";
+
```
ThinuwanW, line 12: 🤖 **Claude** · 💬 **Note**

Import convention deviation. ChipAutocomplete is imported via a deep relative path (`../../../molecules/ChipAutocomplete/ChipAutocomplete`) and the new component folder has no barrel `index.tsx`. Per the module conventions, components should expose a barrel `index.tsx` and be imported through the `~community/*` path alias rather than multi-level relative paths. Add an `index.tsx` re-export for ChipAutocomplete and import it as `~community/people/components/molecules/ChipAutocomplete`.

backend/src/main/java/com/skapp/community/peopleplanner/component/DefaultSkillLoader.java
shakila-kamalasena: Fixed

backend/src/main/java/com/skapp/community/peopleplanner/constant/PeopleMessageConstant.java
shakila-kamalasena: Fixed

backend/src/main/java/com/skapp/community/peopleplanner/controller/v1/SkillController.java
shakila-kamalasena: Fixed

backend/src/main/java/com/skapp/community/peopleplanner/controller/v1/SkillController.java
```diff
@@ -0,0 +1,32 @@
...
+
+	@Operation(summary = "Get all skills", description = "Retrieves all default and custom skills.")
+	@PreAuthorize("hasAnyRole('ROLE_SUPER_ADMIN','ROLE_PEOPLE_EMPLOYEE')")
+	@GetMapping(produces = MediaType.APPLICATION_JSON_VALUE)
```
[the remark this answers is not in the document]
shakila-kamalasena: Fixed

backend/src/main/java/com/skapp/community/peopleplanner/mapper/PeopleMapper.java
```diff
@@ -128,6 +128,7 @@ List<HolidayWorkLocationResponseDto> workLocationsToHolidayWorkLocationResponseD
 	@Mapping(target = "email", source = "user.email")
 	@Mapping(target = "isActive", source = "user.isActive")
 	@Mapping(target = "managers", source = "employeeManagers")
+	@Mapping(target = "skills", ignore = true)
 	EmployeeDetailedResponseDto employeeToEmployeeDetailedResponseDto(Employee employee);
```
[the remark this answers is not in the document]
shakila-kamalasena: Fixed

backend/src/main/java/com/skapp/community/peopleplanner/model/Employee.java
shakila-kamalasena: Removed

backend/src/main/java/com/skapp/community/peopleplanner/model/EmployeeSkill.java
shakila-kamalasena: Fixed

backend/src/main/java/com/skapp/community/peopleplanner/model/EmployeeSkill.java
```diff
@@ -0,0 +1,40 @@
...
+	@Id
+	@Enumerated(EnumType.STRING)
+	@Column(name = "skill_type", nullable = false, columnDefinition = "text")
+	private SkillType skillType;
```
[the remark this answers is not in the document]
shakila-kamalasena: kept default varchar(255)

backend/src/main/java/com/skapp/community/peopleplanner/payload/request/EmployeeSkillDto.java
shakila-kamalasena: We do not add them as a practice

backend/src/main/java/com/skapp/community/peopleplanner/payload/response/EmployeeDetailedResponseDto.java
shakila-kamalasena: This field is already populated.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
shakila-kamalasena: Fixed

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/EmployeeSkillServiceImpl.java
shakila-kamalasena: No need tests

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/EmployeeSkillServiceImpl.java
shakila-kamalasena: Added logs

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/SkillServiceImpl.java
```diff
@@ -0,0 +1,112 @@
...
+				employeeSkill.setSkillId(skillDto.getSkillId());
+				employeeSkill.setSkillType(SkillType.DEFAULT);
+			}
+			else if (skillDto.getSkillType() == SkillType.CUSTOM) {
+				CustomSkill customSkill = findOrCreateCustomSkill(skillDto.getName());
+				employeeSkill.setSkillId(customSkill.getId());
+				employeeSkill.setSkillType(SkillType.CUSTOM);
+			}
+
+			employeeSkills.add(employeeSkill);
```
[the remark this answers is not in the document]
shakila-kamalasena: Added an else { continue; } guard so entries with null or unrecognized skillType are skipped instead of being saved with null

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/EmployeeSkillServiceImpl.java
shakila-kamalasena: Addressed as above

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/SkillServiceImpl.java
shakila-kamalasena: Implemented a map

backend/src/main/resources/community/db/changelog/ddl-script/people-ddl-script-v1-create-table-ppl-custom-skill.sql
shakila-kamalasena: No need this

frontend/src/community/people/api/PeopleApi.ts
shakila-kamalasena: No need inifinite stale times

frontend/src/community/common/components/molecules/ChipAutocomplete/ChipAutocomplete.tsx
```diff
@@ -0,0 +1,269 @@
...
+              <input
+                ref={inputRef}
+                value={inputValue}
+                onChange={handleInputChange}
+                onFocus={() => setIsOpen(true)}
+                onKeyDown={handleKeyDown}
+                disabled={isDisabled}
+                placeholder={value.length === 0 ? placeholder : ""}
+                className="outline-none body1"
+              />
```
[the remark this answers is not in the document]
shakila-kamalasena: Added these attributes

frontend/src/community/people/components/organisms/PersonDetailsSection/SubSections/SkillsDetailsSection.tsx
shakila-kamalasena: Fixed this

frontend/src/community/people/utils/peopleEditFlowUtils/personalDetailsChangesUtils.ts
```diff
@@ -362,6 +361,22 @@ export const getHealthAndOtherDetailsChanges = (
...
+
+  return newSkills.some((skill, index) => {
+    const prev = previousSkills[index];
+    return (
+      skill.skillId !== prev.skillId ||
+      skill.skillType !== prev.skillType ||
+      skill.name !== prev.name
+    );
+  });
+};
```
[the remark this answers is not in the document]
shakila-kamalasena: No need tests

frontend/src/community/people/utils/peopleEditFlowUtils/personalDetailsChangesUtils.ts
shakila-kamalasena: Done as above

frontend/src/community/people/constants/skillConstants.ts
```diff
@@ -0,0 +1,1003 @@
...
+  { id: 1, label: "Python Programming" },
+  { id: 2, label: "JavaScript Development" },
+  { id: 3, label: "Java Programming" },
+  { id: 4, label: "C++ Programming" },
```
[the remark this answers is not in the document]
shakila-kamalasena: No need to do this

frontend/src/community/people/constants/skillConstants.ts
shakila-kamalasena: No need to do this

backend/src/main/java/com/skapp/community/peopleplanner/component/DefaultSkillLoader.java
```diff
@@ -0,0 +1,55 @@
...
+	private Map<Long, DefaultSkill> defaultSkillsById = new HashMap<>();
+
+	@PostConstruct
+	public void init() {
+		loadSkills();
+	}
+
+	private void loadSkills() {
+		try {
+			ClassPathResource resource = new ClassPathResource("community/common/skills.yml");
```
akilarootcode: use yml reader we already have method. check it.

backend/src/main/java/com/skapp/community/peopleplanner/component/DefaultSkillLoader.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+@Component
+public class DefaultSkillLoader {
+
+	private Map<Long, DefaultSkill> defaultSkillsById = new HashMap<>();
+
+	@PostConstruct
```
akilarootcode: no need this.

backend/src/main/java/com/skapp/community/peopleplanner/component/DefaultSkillLoader.java
```diff
@@ -0,0 +1,55 @@
...
+		catch (Exception e) {
+			defaultSkillsById = Collections.emptyMap();
+		}
+	}
+
+	public Optional<DefaultSkill> findById(Long id) {
+		return Optional.ofNullable(defaultSkillsById.get(id));
+	}
+
+	public record DefaultSkill(Long id, String name) {
```
akilarootcode: remove record method. in our code no record methods

backend/src/main/java/com/skapp/community/peopleplanner/controller/v1/SkillController.java
```diff
@@ -0,0 +1,32 @@
...
+@RequiredArgsConstructor
+@RequestMapping("/v1/people/skills")
+@Tag(name = "Skill Controller", description = "Endpoints for skill management")
+public class SkillController {
+
+	private final SkillService skillService;
+
+	@Operation(summary = "Get all custom skills", description = "Retrieves all custom skills.")
+	@PreAuthorize("hasAnyRole('ROLE_SUPER_ADMIN','ROLE_PEOPLE_EMPLOYEE')")
+	@GetMapping(produces = MediaType.APPLICATION_JSON_VALUE)
```
akilarootcode: remove media type

backend/src/main/java/com/skapp/community/peopleplanner/controller/v1/SkillController.java
```diff
@@ -0,0 +1,32 @@
...
+@RestController
+@RequiredArgsConstructor
+@RequestMapping("/v1/people/skills")
+@Tag(name = "Skill Controller", description = "Endpoints for skill management")
+public class SkillController {
+
+	private final SkillService skillService;
+
+	@Operation(summary = "Get all custom skills", description = "Retrieves all custom skills.")
+	@PreAuthorize("hasAnyRole('ROLE_SUPER_ADMIN','ROLE_PEOPLE_EMPLOYEE')")
```
akilarootcode: remove super admin

backend/src/main/java/com/skapp/community/peopleplanner/mapper/PeopleMapper.java
```diff
@@ -128,6 +128,7 @@ List<HolidayWorkLocationResponseDto> workLocationsToHolidayWorkLocationResponseD
 	@Mapping(target = "email", source = "user.email")
 	@Mapping(target = "isActive", source = "user.isActive")
 	@Mapping(target = "managers", source = "employeeManagers")
+	@Mapping(target = "skills", ignore = true)
```
akilarootcode: employee entity need to add mapping to skills

backend/src/main/java/com/skapp/community/peopleplanner/model/CustomSkill.java
```diff
@@ -0,0 +1,28 @@
...
+@NoArgsConstructor
+@Table(name = "ppl_custom_skill")
+public class CustomSkill {
+
+	@Id
+	@GeneratedValue(strategy = GenerationType.IDENTITY)
+	@Column(name = "id", nullable = false, updatable = false)
+	private Long id;
+
+	@Column(name = "name", nullable = false, columnDefinition = "text")
```
akilarootcode: remove column defenition

backend/src/main/java/com/skapp/community/peopleplanner/model/CustomSkill.java
```diff
@@ -0,0 +1,28 @@
...
+@Entity
+@Getter
+@Setter
+@NoArgsConstructor
+@Table(name = "ppl_custom_skill")
+public class CustomSkill {
+
+	@Id
+	@GeneratedValue(strategy = GenerationType.IDENTITY)
+	@Column(name = "id", nullable = false, updatable = false)
```
akilarootcode: remove updatable false

backend/src/main/java/com/skapp/community/peopleplanner/model/EmployeeSkillId.java
```diff
@@ -0,0 +1,25 @@
...
+import lombok.Setter;
+
+import java.io.Serializable;
+
+@Getter
+@Setter
+@NoArgsConstructor
+@AllArgsConstructor
+@EqualsAndHashCode
+public class EmployeeSkillId implements Serializable {
```
akilarootcode: why need serializable ? remove this serializable. 

Image

only this place has serializable. remove it.

backend/src/main/java/com/skapp/community/peopleplanner/payload/request/EmployeeSkillDto.java
```diff
@@ -0,0 +1,17 @@
+package com.skapp.community.peopleplanner.payload.request;
+
+import com.skapp.community.peopleplanner.type.SkillType;
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+public class EmployeeSkillDto {
```
akilarootcode, line 9: if you use Skill or EmployeeSkill use one of them dont mix the pattern

some places has EmployeeSkill but some places has Skll directly. use one pattern

backend/src/main/java/com/skapp/community/peopleplanner/repository/CustomSkillDao.java
```diff
@@ -0,0 +1,14 @@
+package com.skapp.community.peopleplanner.repository;
+
+import com.skapp.community.peopleplanner.model.CustomSkill;
+import org.springframework.data.jpa.repository.JpaRepository;
+import org.springframework.stereotype.Repository;
+
+import java.util.Optional;
+
+@Repository
+public interface CustomSkillDao extends JpaRepository<CustomSkill, Long> {
```
akilarootcode: custom employeeSkill

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleReadServiceImpl.java
```diff
@@ -115,6 +119,13 @@ private EmployeePersonalDetailsDto mapPersonalDetails(Employee employee,
 			EmployeeProfileViewAccessLevel accessLevel) {
 		EmployeePersonalDetailsDto dto = new EmployeePersonalDetailsDto();
 		dto.setGeneral(mapPersonalGeneralDetails(employee, accessLevel));
+		dto.setSkills(skillService.getEmployeeSkillResponses(employee.getEmployeeId()).stream().map(s -> {
+			EmployeeSkillDto skillDto = new EmployeeSkillDto();
+			skillDto.setSkillId(s.getId());
+			skillDto.setSkillType(s.getSkillType());
+			skillDto.setName(s.getName());
+			return skillDto;
+		}).toList());
```
akilarootcode, line 123: use existing pattern. setSkills(mapEmployeeSkills) use private method to map. otherwise code will complex

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
```diff
@@ -237,6 +240,8 @@ public ResponseEntityDto createEmployee(CreateEmployeeRequestDto requestDto) {
 
 		userDao.save(user);
 
+		processEmployeeSkills(requestDto.getPersonal(), employee);
```
akilarootcode: move this to createEmployeeEntity. all the processing must move inside to createEmployeeEntity. so entity creation part need to move

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
```diff
@@ -205,6 +206,8 @@ public class PeopleServiceImpl implements PeopleService {
 
 	private final EmployeeExportMapperService employeeExportMapperService;
 
+	private final SkillService skillService;
```
akilarootcode: employeeSkill service rename

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/EmployeeSkillServiceImpl.java
```diff
@@ -0,0 +1,133 @@
...
+
+	private final CustomSkillDao customSkillDao;
+
+	private final EmployeeSkillDao employeeSkillDao;
+
+	@Override
+	@Transactional
+	public List<EmployeeSkill> saveEmployeeSkills(Employee employee, List<EmployeeSkillDto> skills) {
+		log.info("saveEmployeeSkills: execution started");
+		employeeSkillDao.deleteByEmployeeEmployeeId(employee.getEmployeeId());
```
akilarootcode: why delete and flush ? use proper transactional mechanism

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/EmployeeSkillServiceImpl.java
```diff
@@ -0,0 +1,133 @@
...
+		employeeSkillDao.flush();
+
+		if (skills == null || skills.isEmpty()) {
```
akilarootcode: validations must top before delete. "

look EmployeeSkillDao but service has SkillService, use consistent naming

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/SkillServiceImpl.java
```diff
@@ -0,0 +1,133 @@
...
+		List<EmployeeSkill> employeeSkills = employeeSkillDao.findByEmployeeEmployeeId(employeeId);
+
+		List<Long> customSkillIds = employeeSkills.stream()
+			.filter(es -> es.getSkillType() == SkillType.CUSTOM)
+			.map(EmployeeSkill::getSkillId)
+			.toList();
+
+		Map<Long, CustomSkill> customSkillMap = customSkillDao.findAllById(customSkillIds)
+			.stream()
+			.collect(Collectors.toMap(CustomSkill::getId, Function.identity()));
```
akilarootcode: function.identity() do we need it ?

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/SkillServiceImpl.java
```diff
@@ -0,0 +1,133 @@
...
+
+		return employeeSkills.stream().map(es -> {
+			if (es.getSkillType() == SkillType.DEFAULT) {
+				DefaultSkillLoader.DefaultSkill defaultSkill = defaultSkillLoader.findById(es.getSkillId())
```
akilarootcode: nested if statements have. try to simplify to 1 if else layer

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/EmployeeSkillServiceImpl.java
```diff
@@ -0,0 +1,133 @@
...
+				}
+				return new SkillResponseDto(es.getSkillId(), customSkill.getName(), SkillType.CUSTOM);
+			}
+		}).sorted(Comparator.comparing(SkillResponseDto::getName, String.CASE_INSENSITIVE_ORDER)).toList();
+	}
+
+	private CustomSkill findOrCreateCustomSkill(String name) {
+		Optional<CustomSkill> existing = customSkillDao.findByNameIgnoreCase(name);
+		if (existing.isPresent()) {
+			return existing.get();
```
akilarootcode: simplify and use proper spacing between each logic

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/SkillServiceImpl.java
```diff
@@ -0,0 +1,133 @@
...
+		}
+		CustomSkill customSkill = new CustomSkill();
+		customSkill.setName(name);
+		return customSkillDao.save(customSkill);
+	}
+
+	@Override
+	public ResponseEntityDto getAllSkills() {
+		log.info("getAllSkills: execution started");
+		List<SkillResponseDto> customSkills = customSkillDao.findAll()
```
akilarootcode: after log add spacing

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/SkillServiceImpl.java
```diff
@@ -0,0 +1,133 @@
...
+			.stream()
+			.map(s -> new SkillResponseDto(s.getId(), s.getName(), SkillType.CUSTOM))
+			.sorted(Comparator.comparing(SkillResponseDto::getName, String.CASE_INSENSITIVE_ORDER))
```
akilarootcode: use db level sorting when find don't use service level sorting.

backend/src/main/resources/community/db/changelog/ddl-script/people-ddl-script-v1-create-table-ppl-employee-skill.sql
```diff
@@ -0,0 +1,13 @@
+-- liquibase formatted sql
+
+-- changeset shakila:00115_create_table_ppl_employee_skill
+CREATE TABLE IF NOT EXISTS `ppl_employee_skill`
+(
+    `employee_id` bigint       NOT NULL,
+    `skill_id`    bigint       NOT NULL,
+    `skill_type`  varchar(255) NOT NULL,
```
akilarootcode: use text

backend/src/main/java/com/skapp/community/peopleplanner/component/DefaultSkillLoader.java
```diff
@@ -0,0 +1,55 @@
...
+import org.yaml.snakeyaml.Yaml;
+
+import java.io.InputStream;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+@Component
```
akilarootcode: no need component. so use service level parsing instead. check email templates approach

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
```diff
@@ -317,6 +322,8 @@ public ResponseEntityDto updateEmployee(Long employeeId, CreateEmployeeRequestDt
 
 		userDao.save(user);
 
+		processEmployeeSkills(requestDto.getPersonal(), employee);
```
akilarootcode: remove process from here as well. so create methods are fully simple processing parts do inside private methods

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/EmployeeSkillServiceImpl.java
```diff
@@ -0,0 +1,133 @@
...
+	@Override
+	public ResponseEntityDto getAllSkills() {
+		log.info("getAllSkills: execution started");
+		List<SkillResponseDto> customSkills = customSkillDao.findAll()
+			.stream()
+			.map(s -> new SkillResponseDto(s.getId(), s.getName(), SkillType.CUSTOM))
+			.sorted(Comparator.comparing(SkillResponseDto::getName, String.CASE_INSENSITIVE_ORDER))
+			.toList();
+
+		return new ResponseEntityDto(false, (Object) customSkills);
```
akilarootcode: why object cast ? remove it. use proper response interface or dto

backend/src/main/java/com/skapp/community/peopleplanner/controller/v1/SkillController.java
```diff
@@ -0,0 +1,32 @@
...
+import org.springframework.security.access.prepost.PreAuthorize;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+@RequiredArgsConstructor
+@RequestMapping("/v1/people/skills")
+@Tag(name = "Skill Controller", description = "Endpoints for skill management")
+public class SkillController {
```
akilarootcode: for controller also use employeeSkill controller or use similar pattern

akilarootcode (changes requested):

backend/src/main/java/com/skapp/community/peopleplanner/component/DefaultSkillLoader.java
shakila-kamalasena: Done

backend/src/main/java/com/skapp/community/peopleplanner/component/DefaultSkillLoader.java
shakila-kamalasena: Removed

backend/src/main/java/com/skapp/community/peopleplanner/component/DefaultSkillLoader.java
shakila-kamalasena: Removed this method

backend/src/main/java/com/skapp/community/peopleplanner/controller/v1/SkillController.java
shakila-kamalasena: Removed

backend/src/main/java/com/skapp/community/peopleplanner/controller/v1/SkillController.java
shakila-kamalasena: Done, removed ROLE_SUPER_ADMIN

backend/src/main/java/com/skapp/community/peopleplanner/model/CustomSkill.java
shakila-kamalasena: Removed

backend/src/main/java/com/skapp/community/peopleplanner/model/CustomSkill.java
shakila-kamalasena: Removed

backend/src/main/java/com/skapp/community/peopleplanner/payload/request/EmployeeSkillDto.java
shakila-kamalasena: Used employee skill everywhere

backend/src/main/java/com/skapp/community/peopleplanner/repository/CustomSkillDao.java
shakila-kamalasena: Done

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/EmployeeSkillServiceImpl.java
shakila-kamalasena: Done, removed flush

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/EmployeeSkillServiceImpl.java
shakila-kamalasena: Delete is in the correct place as if a user sends an empty skills list (intending to clear all their skills), the method returns early without deleting the existing skills.

Made the naming consistent

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/SkillServiceImpl.java
shakila-kamalasena: Function.identity() is needed here, it's the value mapper for Collectors.toMap

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/SkillServiceImpl.java
shakila-kamalasena: Simplified by extracting to a method

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/EmployeeSkillServiceImpl.java
shakila-kamalasena: Done

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/SkillServiceImpl.java
shakila-kamalasena: Done

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/SkillServiceImpl.java
shakila-kamalasena: Removed sorting, no need of them

backend/src/main/resources/community/db/changelog/ddl-script/people-ddl-script-v1-create-table-ppl-employee-skill.sql
shakila-kamalasena: Cannot use text

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/EmployeeSkillServiceImpl.java
shakila-kamalasena: Removed

backend/src/main/java/com/skapp/community/peopleplanner/controller/v1/SkillController.java
shakila-kamalasena: Renamed

backend/src/main/java/com/skapp/community/peopleplanner/component/DefaultSkillLoader.java
shakila-kamalasena: Done

backend/src/main/java/com/skapp/community/peopleplanner/mapper/PeopleMapper.java
shakila-kamalasena: Added

backend/src/main/java/com/skapp/community/peopleplanner/model/EmployeeSkillId.java
shakila-kamalasena: Removed this

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleReadServiceImpl.java
shakila-kamalasena: Done

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
shakila-kamalasena: Removed and moved inside the createEmployeeEntity

hasalarootcode (dismissed):

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/EmployeeSkillServiceImpl.java
```diff
@@ -0,0 +1,159 @@
...
+
+		return employeeSkills.stream().map(es -> mapToEmployeeSkillResponseDto(es, customSkillMap)).toList();
+	}
+
+	@Override
+	public ResponseEntityDto getAllSkills() {
+		log.info("getAllSkills: execution started");
+
+		List<EmployeeSkillResponseDto> customSkills = customEmployeeSkillDao.findAll()
+			.stream()
```
akilarootcode, line 74: filter from db query

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/EmployeeSkillServiceImpl.java
```diff
@@ -0,0 +1,159 @@
...
+	}
+
+	@Override
+	public List<EmployeeSkillResponseDto> getEmployeeSkillResponses(Long employeeId) {
+		log.info("getEmployeeSkillResponses: execution started");
+
+		List<EmployeeSkill> employeeSkills = employeeSkillDao.findByEmployeeEmployeeId(employeeId);
+
+		List<Long> customSkillIds = employeeSkills.stream()
+			.filter(es -> es.getSkillType() == EmployeeSkillType.CUSTOM)
```
akilarootcode: filter from db

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/EmployeeSkillServiceImpl.java
```diff
@@ -0,0 +1,159 @@
...
+
+		Map<Long, CustomEmployeeSkill> customSkillMap = customEmployeeSkillDao.findAllById(customSkillIds)
+			.stream()
+			.collect(Collectors.toMap(CustomEmployeeSkill::getId, Function.identity()));
+
+		return employeeSkills.stream().map(es -> mapToEmployeeSkillResponseDto(es, customSkillMap)).toList();
+	}
+
+	@Override
+	public ResponseEntityDto getAllSkills() {
```
akilarootcode: get custom skills

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/EmployeeSkillServiceImpl.java
```diff
@@ -0,0 +1,159 @@
...
+		List<EmployeeSkill> employeeSkills = employeeSkillDao.findByEmployeeEmployeeId(employeeId);
+
+		List<Long> customSkillIds = employeeSkills.stream()
+			.filter(es -> es.getSkillType() == EmployeeSkillType.CUSTOM)
+			.map(EmployeeSkill::getSkillId)
+			.toList();
+
+		Map<Long, CustomEmployeeSkill> customSkillMap = customEmployeeSkillDao.findAllById(customSkillIds)
+			.stream()
+			.collect(Collectors.toMap(CustomEmployeeSkill::getId, Function.identity()));
```
akilarootcode: simplify this method

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/EmployeeSkillServiceImpl.java
```diff
@@ -0,0 +1,159 @@
...
+			}
+
+			employeeSkills.add(employeeSkill);
+		}
+
+		return employeeSkillDao.saveAll(employeeSkills);
+	}
+
+	@Override
+	public List<EmployeeSkillResponseDto> getEmployeeSkillResponses(Long employeeId) {
```
akilarootcode: use proper name

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/EmployeeSkillServiceImpl.java
```diff
@@ -0,0 +1,159 @@
...
+			EmployeeSkill employeeSkill = new EmployeeSkill();
+			employeeSkill.setEmployee(employee);
+
+			if (skillDto.getSkillType() == EmployeeSkillType.DEFAULT) {
+				getDefaultEmployeeSkillName(skillDto.getSkillId());
+				employeeSkill.setSkillId(skillDto.getSkillId());
+				employeeSkill.setSkillType(EmployeeSkillType.DEFAULT);
+			}
+			else if (skillDto.getSkillType() == EmployeeSkillType.CUSTOM) {
+				if (skillDto.getName() == null || skillDto.getName().isBlank()) {
```
akilarootcode: simplify if else ladder

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/EmployeeSkillServiceImpl.java
```diff
@@ -0,0 +1,159 @@
...
+		log.info("getEmployeeSkillResponses: execution started");
+
+		List<EmployeeSkill> employeeSkills = employeeSkillDao.findByEmployeeEmployeeId(employeeId);
+
+		List<Long> customSkillIds = employeeSkills.stream()
+			.filter(es -> es.getSkillType() == EmployeeSkillType.CUSTOM)
+			.map(EmployeeSkill::getSkillId)
+			.toList();
+
+		Map<Long, CustomEmployeeSkill> customSkillMap = customEmployeeSkillDao.findAllById(customSkillIds)
```
akilarootcode: no need hashmap all the places i think

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/EmployeeSkillServiceImpl.java
```diff
@@ -0,0 +1,159 @@
...
+			throw new ModuleException(PeopleMessageConstant.PEOPLE_ERROR_SKILL_NOT_FOUND);
+		}
+
+		return name;
+	}
+
+	private CustomEmployeeSkill findOrCreateCustomSkill(String name) {
+		Optional<CustomEmployeeSkill> existing = customEmployeeSkillDao.findByNameIgnoreCase(name);
+
+		if (existing.isPresent()) {
```
akilarootcode: remove this method and throw error if already exist. if exist find and return not a good behavior

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/EmployeeSkillServiceImpl.java
```diff
@@ -0,0 +1,159 @@
...
+@RequiredArgsConstructor
+public class EmployeeSkillServiceImpl implements EmployeeSkillService {
+
+	private static final String SKILLS_YAML_PATH = "community/common/skills.yml";
+
+	private final CustomEmployeeSkillDao customEmployeeSkillDao;
+
+	private final EmployeeSkillDao employeeSkillDao;
+
+	private Map<Long, String> defaultEmployeeSkillNames;
```
akilarootcode: setting all 1000 skills to variable can be a doubt. because it can be overflow the memory. set on redis or fetch always from resources will the good option i think.

akilarootcode (changes requested):

backend/src/main/java/com/skapp/community/peopleplanner/payload/employeeskill/DefaultEmployeeSkillsYaml.java
```diff
@@ -0,0 +1,14 @@
...
+import lombok.Getter;
+import lombok.Setter;
+
+import java.util.List;
+
+@Getter
+@Setter
+public class DefaultEmployeeSkillsYaml {
+
+	private List<DefaultEmployeeSkill> skills;
```
akilarootcode: change variable name as well defaultSkills likewise

akilarootcode (changes requested):

backend/src/main/java/com/skapp/community/peopleplanner/constant/PeopleMessageConstant.java
```diff
@@ -228,7 +228,8 @@ public enum PeopleMessageConstant implements MessageConstant {
 			"api.error.people.transfer-supervisors.team-record-not-found"),
 	PEOPLE_ERROR_EMPLOYEE_TERMINATION_OR_DELETION_ACTION_REQUIRED(
 			"api.error.people.employee-termination-or-deletion-action-required"),
-	PEOPLE_ERROR_VALIDATION_WORK_LOCATION_NOT_FOUND("api.error.people.validation-work-location-not-found");
+	PEOPLE_ERROR_VALIDATION_WORK_LOCATION_NOT_FOUND("api.error.people.validation-work-location-not-found"),
```
akilarootcode, line 231: not,found is wrong. not-found

akilarootcode (changes requested):

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/EmployeeSkillServiceImpl.java
shakila-kamalasena: Got from the criteria query

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/EmployeeSkillServiceImpl.java
shakila-kamalasena: Done

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/EmployeeSkillServiceImpl.java
shakila-kamalasena: Changed

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/EmployeeSkillServiceImpl.java
shakila-kamalasena: Removed

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/EmployeeSkillServiceImpl.java
shakila-kamalasena: Renamed to getEmployeeSkills

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/EmployeeSkillServiceImpl.java
shakila-kamalasena: Already removed this

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/EmployeeSkillServiceImpl.java
shakila-kamalasena: Done

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/EmployeeSkillServiceImpl.java
shakila-kamalasena: Removed and implemented to fetch always from resources

backend/src/main/java/com/skapp/community/peopleplanner/constant/PeopleMessageConstant.java
shakila-kamalasena: Changed this

backend/src/main/java/com/skapp/community/peopleplanner/payload/employeeskill/DefaultEmployeeSkillsYaml.java
shakila-kamalasena: Done

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/EmployeeSkillServiceImpl.java
shakila-kamalasena: Simplified by extracting two methods for DEFAULT and CUSTOM

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/EmployeeSkillServiceImpl.java
```diff
@@ -0,0 +1,139 @@
...
+
+		for (EmployeeSkillDto skillDto : skills) {
+			EmployeeSkill employeeSkill = new EmployeeSkill();
+			employeeSkill.setEmployee(employee);
+			employeeSkill.setSkillType(skillDto.getSkillType());
+
+			employeeSkill.setSkillId(switch (skillDto.getSkillType()) {
+				case DEFAULT -> resolveDefaultSkillId(skillDto);
+				case CUSTOM -> resolveCustomSkillId(skillDto);
+				case null -> throw new ModuleException(PeopleMessageConstant.PEOPLE_ERROR_SKILL_NOT_FOUND);
```
akilarootcode: remove null

akilarootcode (dismissed):

hasalarootcode (dismissed):

akilarootcode (dismissed):

backend/src/main/java/com/skapp/community/common/util/YamlReader.java
```diff
@@ -0,0 +1,26 @@
...
+import tools.jackson.dataformat.yaml.YAMLMapper;
+
+import java.io.InputStream;
+
+@UtilityClass
+public class YamlReader {
+
+	private static final YAMLMapper yamlMapper = YAMLMapper.builder().build();
+
+	public static <T> T read(String path, Class<T> clazz) {
```
ThinuwanW: clazz ???

backend/src/main/java/com/skapp/community/peopleplanner/mapper/PeopleMapper.java
```diff
@@ -128,8 +130,12 @@ List<HolidayWorkLocationResponseDto> workLocationsToHolidayWorkLocationResponseD
 	@Mapping(target = "email", source = "user.email")
 	@Mapping(target = "isActive", source = "user.isActive")
 	@Mapping(target = "managers", source = "employeeManagers")
+	@Mapping(target = "skills", source = "employeeSkills")
 	EmployeeDetailedResponseDto employeeToEmployeeDetailedResponseDto(Employee employee);
 
+	@Mapping(source = "skillId", target = "id")
```
ThinuwanW, line 136: we can remove mapper

backend/src/main/java/com/skapp/community/peopleplanner/payload/employeeskill/DefaultEmployeeSkillsYaml.java
```diff
@@ -0,0 +1,14 @@
+package com.skapp.community.peopleplanner.payload.employeeskill;
+
+import lombok.Getter;
+import lombok.Setter;
+
+import java.util.List;
+
+@Getter
+@Setter
+public class DefaultEmployeeSkillsYaml {
```
ThinuwanW, line 10: Dto at end

backend/src/main/java/com/skapp/community/peopleplanner/payload/employeeskill/DefaultEmployeeSkillsYaml.java
```diff
@@ -0,0 +1,14 @@
+package com.skapp.community.peopleplanner.payload.employeeskill;
+
+import lombok.Getter;
+import lombok.Setter;
+
+import java.util.List;
+
+@Getter
+@Setter
```
ThinuwanW, line 9: can check can remove setter and no warinings

backend/src/main/java/com/skapp/community/peopleplanner/repository/EmployeeSkillDao.java
```diff
@@ -0,0 +1,17 @@
...
+import org.springframework.stereotype.Repository;
+
+import java.util.List;
+
+@Repository
+public interface EmployeeSkillDao extends JpaRepository<EmployeeSkill, EmployeeSkillId>, EmployeeSkillRepository {
+
+	List<EmployeeSkill> findByEmployeeEmployeeId(Long employeeId);
+
+	void deleteByEmployeeEmployeeId(Long employeeId);
```
ThinuwanW: move to employee edit flow

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/EmployeeSkillServiceImpl.java
```diff
@@ -0,0 +1,138 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+@Slf4j
+@Service
+@RequiredArgsConstructor
+public class EmployeeSkillServiceImpl implements EmployeeSkillService {
```
ThinuwanW, line 28: reosurces/community/people/skills/skills.yml

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/EmployeeSkillServiceImpl.java
```diff
@@ -0,0 +1,138 @@
...
+
+	private static final String SKILLS_YAML_PATH = "community/common/skills.yml";
+
+	private final CustomEmployeeSkillDao customEmployeeSkillDao;
+
+	private final EmployeeSkillDao employeeSkillDao;
+
+	@Override
```
ThinuwanW, line 36: handle through the diff

akilarootcode (dismissed):

hasalarootcode (dismissed):

akilarootcode (dismissed):

ThinuwanW (approved):

PR #1491, opened 2026-06-16
fix: refactor attendance summary methods to fix bugs

yuhanga-rtc:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/backend/src/main/java/com/skapp/community/timeplanner/repository/impl/TimeRecordRepositoryImpl.java
+++ b/backend/src/main/java/com/skapp/community/timeplanner/repository/impl/TimeRecordRepositoryImpl.java
@@ -65,8 +65,22 @@ public class TimeRecordRepositoryImpl implements TimeRecordRepository {
 	@Override
 	public AttendanceSummaryDto getEmployeeAttendanceSummary(List<Long> employeeIds, LocalDate startDate,
 			LocalDate endDate) {
-		List<Predicate> additionalPredicates = new ArrayList<>();
-		return computeMergedAttendanceSummary(employeeIds, startDate, endDate, additionalPredicates);
+		CriteriaBuilder criteriaBuilder = entityManager.getCriteriaBuilder();
+		CriteriaQuery<AttendanceSummaryDto> criteriaQuery = criteriaBuilder.createQuery(AttendanceSummaryDto.class);
+		Root<TimeRecord> root = criteriaQuery.from(TimeRecord.class);
+
+		List<Predicate> predicates = new ArrayList<>();
+		predicates.add((root.get(TimeRecord_.employee).get(Employee_.employeeId).in(employeeIds)));
+		predicates.add(criteriaBuilder.between(root.get(TimeRecord_.date), startDate, endDate));
+
+		Predicate[] predArray = new Predicate[predicates.size()];
+		predicates.toArray(predArray);
+		criteriaQuery.where(predArray);
+		criteriaQuery.select(criteriaBuilder.construct(AttendanceSummaryDto.class,
+				criteriaBuilder.coalesce(criteriaBuilder.sum(root.get(TimeRecord_.workedHours)), 0.0),
+				criteriaBuilder.coalesce(criteriaBuilder.sum(root.get(TimeRecord_.breakHours)), 0.0)));
+		TypedQuery<AttendanceSummaryDto> typedQuery = entityManager.createQuery(criteriaQuery);
+		return typedQuery.getSingleResult();
 	}
 
 	@Override
@@ -94,39 +108,62 @@ public class TimeRecordRepositoryImpl implements TimeRecordRepository {
 	@Override
 	public AttendanceSummaryDto findManagerAssignUsersAttendanceSummary(Long managerId, List<Long> teamIds,
 			LocalDate startDate, LocalDate endDate, List<Long> employeeIds) {
-		if (employeeIds == null || employeeIds.isEmpty()) {
-			return new AttendanceSummaryDto(0.0f, 0.0f);
+		CriteriaBuilder criteriaBuilder = entityManager.getCriteriaBuilder();
+
+		CriteriaQuery<AttendanceSummaryDto> criteriaQuery = criteriaBuilder.createQuery(AttendanceSummaryDto.class);
+		Root<TimeRecord> root = criteriaQuery.from(TimeRecord.class);
+
+		// Predicates for the main query
+		List<Predicate> predicates = new ArrayList<>();
+		predicates.add(criteriaBuilder.between(root.get(TimeRecord_.date), startDate, endDate));
+		if (!employeeIds.isEmpty()) {
+
+			CriteriaBuilder.In<Long> inClause = criteriaBuilder
+				.in(root.get(TimeRecord_.employee).get(Employee_.employeeId));
+			for (Long employeeID : employeeIds) {
+				inClause.value(employeeID);
+			}
+			predicates.add(inClause);
 		}
-		return computeMergedAttendanceSummary(employeeIds, startDate, endDate, new ArrayList<>());
+
+		criteriaQuery.select(criteriaBuilder.construct(AttendanceSummaryDto.class,
+				criteriaBuilder.coalesce(criteriaBuilder.sum(root.get(TimeRecord_.workedHours)), 0.0),
+				criteriaBuilder.coalesce(criteriaBuilder.sum(root.get(TimeRecord_.breakHours)), 0.0)));
+		criteriaQuery.where(predicates.toArray(new Predicate[0]));
+
+		TypedQuery<AttendanceSummaryDto> typedQuery = entityManager.createQuery(criteriaQuery);
+		try {
+			return typedQuery.getSingleResult();
+		}
+		catch (NoResultException e) {
+			return new AttendanceSummaryDto(0.0F, 0.0F);
+		}
 	}
 
 	@Override
 	public TimeSheetSummaryData findTimeSheetSummaryData(LocalDate startDate, LocalDate endDate,
 			List<Long> employeeIds) {
-		if (employeeIds == null || employeeIds.isEmpty()) {
-			return new TimeSheetSummaryData(0.0, 0.0, 0.0);
-		}
-		AttendanceSummaryDto mergedSummary = computeMergedAttendanceSummary(employeeIds, startDate, endDate,
-				new ArrayList<>());
-
 		CriteriaBuilder criteriaBuilder = entityManager.getCriteriaBuilder();
-		CriteriaQuery<Tuple> avgQuery = criteriaBuilder.createTupleQuery();
-		Root<TimeRecord> root = avgQuery.from(TimeRecord.class);
+		CriteriaQuery<TimeSheetSummaryData> criteriaQuery = criteriaBuilder.createQuery(TimeSheetSummaryData.class);
+		Root<TimeRecord> root = criteriaQuery.from(TimeRecord.class);
 
 		List<Predicate> predicates = new ArrayList<>();
 		predicates.add(root.get(TimeRecord_.employee).get(Employee_.employeeId).in(employeeIds));
 		predicates.add(criteriaBuilder.between(root.get(TimeRecord_.date), startDate, endDate));
-		predicates.add(criteriaBuilder.isFalse(root.get(TimeRecord_.isManual)));
 
-		avgQuery.select(criteriaBuilder.tuple(
-				criteriaBuilder.coalesce(criteriaBuilder.avg(root.get(TimeRecord_.clockInTime)), 0.0),
-				criteriaBuilder.coalesce(criteriaBuilder.avg(root.get(TimeRecord_.clockOutTime)), 0.0)));
-		avgQuery.where(predicates.toArray(new Predicate[0]));
+		Predicate[] predArray = new Predicate[predicates.size()];
+		predicates.toArray(predArray);
+		criteriaQuery.where(predArray);
 
-		Tuple avgResult = entityManager.createQuery(avgQuery).getSingleResult();
+		criteriaQuery
+			.select(criteriaBuilder.construct(TimeSheetSummaryData.class,
+					criteriaBuilder.coalesce(criteriaBuilder.sum(root.get(TimeRecord_.workedHours)).as(Double.class),
+							0.0),
+					criteriaBuilder.coalesce(criteriaBuilder.avg(root.get(TimeRecord_.clockInTime)), 0.0),
+					criteriaBuilder.coalesce(criteriaBuilder.avg(root.get(TimeRecord_.clockOutTime)), 0.0)));
 
-		return new TimeSheetSummaryData((double) mergedSummary.getTotalWorkHours(), avgResult.get(0, Double.class),
-				avgResult.get(1, Double.class));
+		TypedQuery<TimeSheetSummaryData> typedQuery = entityManager.createQuery(criteriaQuery);
+		return typedQuery.getSingleResult();
 	}
 
 	@Override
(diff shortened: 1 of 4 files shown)


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR refactors TimeRecord aggregation to merge auto clock-in records with manual entries (deducting time overlap) across attendance-summary, manager-summary, and timesheet queries. The merge logic is sound and correctly handles null clock times, but it replaces DB-side SUM aggregation with full row fetch + in-memory aggregation (potential performance regression on wide ranges), introduces a subtle empty-employeeIds semantic change in getEmployeeAttendanceSummary, and silently collapses multiple non-manual records per employee/day.

Found **8** new issue(s): 🟡 4 suggestion(s), 🟣 1 nit(s)

backend/src/main/java/com/skapp/community/timeplanner/repository/impl/TimeRecordRepositoryImpl.java
```diff
@@ -471,6 +475,90 @@ protected EmployeeTimeRecord buildEmptyTimeRecord(Long employeeId, LocalDate dat
...
+		predicates.add(cb.between(root.get(TimeRecord_.date), startDate, endDate));
+
+		query.select(
+				cb.tuple(root.get(TimeRecord_.timeRecordId), root.get(TimeRecord_.employee).get(Employee_.employeeId),
+						root.get(TimeRecord_.date), cb.coalesce(root.get(TimeRecord_.workedHours), 0.0f),
+						cb.coalesce(root.get(TimeRecord_.breakHours), 0.0f), root.get(TimeRecord_.isManual),
+						root.get(TimeRecord_.clockInTime), root.get(TimeRecord_.clockOutTime)));
+		query.where(predicates.toArray(new Predicate[0]));
+
+		List<Tuple> records = entityManager.createQuery(query).getResultList();
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

Performance regression: the previous implementations of getEmployeeAttendanceSummary / findManagerAssignUsersAttendanceSummary / findTimeSheetSummaryData computed SUM(workedHours)/SUM(breakHours) on the database side and returned a single aggregate row. computeMergedAttendanceSummary now fetches every matching TimeRecord row (8 columns each) into memory via getResultList() and sums in Java. For manager summaries spanning many employees, and for getWorkedHoursForDateRange (called over 30- and 60-day windows in calculateWorkTimeUtilization), this can pull a large, unbounded result set. The row-level fetch is necessary for the overlap-merge feature, but consider bounding it (e.g. confirm date ranges/employee counts are constrained) or pushing the merge into a more selective query, and add a note that this is intentionally non-aggregated. Severity: should review for the manager/export paths.

backend/src/main/java/com/skapp/community/timeplanner/repository/impl/TimeRecordRepositoryImpl.java
```diff
@@ -471,6 +475,90 @@ protected EmployeeTimeRecord buildEmptyTimeRecord(Long employeeId, LocalDate dat
...
 	}
 
+	private AttendanceSummaryDto computeMergedAttendanceSummary(List<Long> employeeIds, LocalDate startDate,
+			LocalDate endDate, List<Predicate> additionalPredicates) {
+		CriteriaBuilder cb = entityManager.getCriteriaBuilder();
+		CriteriaQuery<Tuple> query = cb.createTupleQuery();
+		Root<TimeRecord> root = query.from(TimeRecord.class);
+
+		List<Predicate> predicates = new ArrayList<>(additionalPredicates);
+		if (employeeIds != null && !employeeIds.isEmpty()) {
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

Behavioral change for empty employeeIds in getEmployeeAttendanceSummary. The old getEmployeeAttendanceSummary unconditionally added root...employeeId.in(employeeIds); with an empty list JPA emits a false predicate (1=0), so the result was (0.0, 0.0). The shared helper instead skips the employee filter entirely when the list is empty, so an empty list now aggregates ALL employees in the date range. The same divergence affects findTimeSheetSummaryData, where computeMergedAttendanceSummary would sum all employees while the separate avg query still uses .in(emptyList) (=0), producing an inconsistent response. Current callers pass non-empty lists (List.of(userId), or the manager 'all' path which relied on this for findManagerAssignUsersAttendanceSummary), so this is latent, but the asymmetry is a trap. Suggest guarding getEmployeeAttendanceSummary/findTimeSheetSummaryData to return (0,0) on empty input, or document that empty means 'all' only for the manager method.

backend/src/main/java/com/skapp/community/timeplanner/repository/impl/TimeRecordRepositoryImpl.java
```diff
@@ -379,11 +340,54 @@ public List<EmployeeTimeRecord> findEmployeesTimeRecords(List<Long> employeeIds,
...
+			for (Tuple tuple : tuples) {
+				Boolean isManual = tuple.get(6, Boolean.class);
+				if (Boolean.TRUE.equals(isManual)) {
+					if (latestManualRecord == null
+							|| tuple.get(0, Long.class) > latestManualRecord.get(0, Long.class)) {
+						latestManualRecord = tuple;
+					}
+				}
+				else {
+					clockInRecord = tuple;
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

Multiple non-manual records per employee/day are silently collapsed. In both findEmployeesTimeRecords (here) and computeMergedAttendanceSummary, the loop does clockInRecord = tuple for every non-manual record, so only the last one encountered survives; the group ordering is DB-dependent (the query's groupBy is for aggregation, there is no orderBy), making the choice non-deterministic and dropping the hours of any other non-manual records. Manual records are reduced to the single highest-id entry, also discarding earlier manual entries. The prior findEmployeesTimeRecords used Collectors.toMap with no merge function, which would have thrown on a duplicate key, so multiple records per key is a newly-supported case worth getting right. Please confirm the invariant 'at most one non-manual record and only the latest manual record matters per employee/day' is correct; if multiple auto/manual records can legitimately coexist, their hours should be summed rather than dropped.

backend/src/main/java/com/skapp/community/timeplanner/repository/impl/TimeRecordRepositoryImpl.java
```diff
@@ -108,62 +94,35 @@ public Optional<TimeRecord> findIncompleteClockOutTimeRecords(LocalDate lastCloc
...
 		CriteriaBuilder criteriaBuilder = entityManager.getCriteriaBuilder();
-		CriteriaQuery<TimeSheetSummaryData> criteriaQuery = criteriaBuilder.createQuery(TimeSheetSummaryData.class);
-		Root<TimeRecord> root = criteriaQuery.from(TimeRecord.class);
+		CriteriaQuery<Tuple> avgQuery = criteriaBuilder.createTupleQuery();
+		Root<TimeRecord> root = avgQuery.from(TimeRecord.class);
 
 		List<Predicate> predicates = new ArrayList<>();
 		predicates.add(root.get(TimeRecord_.employee).get(Employee_.employeeId).in(employeeIds));
 		predicates.add(criteriaBuilder.between(root.get(TimeRecord_.date), startDate, endDate));
+		predicates.add(criteriaBuilder.isFalse(root.get(TimeRecord_.isManual)));
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Behavioral change in findTimeSheetSummaryData: the new average clock-in/clock-out query adds `criteriaBuilder.isFalse(root.get(TimeRecord_.isManual))`, so manual records are now excluded from the average clock-in/out times. The previous implementation computed avg(clockInTime)/avg(clockOutTime) over ALL matching records (manual + non-manual). If a manual record carries real clock-in/out times, it used to contribute to the averages and no longer does. Confirm this exclusion is intended; if manual entries should still influence the displayed average times, this is a regression. (Note: SQL AVG already ignores NULLs, so the change only matters for manual records that have non-null clock times.)

backend/src/main/java/com/skapp/community/timeplanner/repository/impl/TimeRecordRepositoryImpl.java
```diff
@@ -471,6 +475,90 @@ protected EmployeeTimeRecord buildEmptyTimeRecord(Long employeeId, LocalDate dat
 		return new EmployeeTimeRecordImpl(null, employeeId, date, 0.0f, 0.0f, null);
 	}
 
+	private AttendanceSummaryDto computeMergedAttendanceSummary(List<Long> employeeIds, LocalDate startDate,
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Code duplication: the per-group merge loop in computeMergedAttendanceSummary (lines 505-537) is nearly identical to the merge block in findEmployeesTimeRecords (lines 344-390) — same grouping, same clockIn/latestManual selection, same computeMergedHoursWithOverlap call, differing only in tuple indices (no JSON_ARRAYAGG column here) and whether it returns a record or accumulates totals. Extract a shared helper (e.g. a method that resolves a group of tuples into a merged worked/break-hours pair) to avoid the two copies drifting. They already diverge: findEmployeesTimeRecords rounds worked/break hours via cb.round(..., 2) while this method selects raw values, so the summary card total computed here will not equal the sum of the rounded per-day rows shown by findEmployeesTimeRecords.

backend/src/main/java/com/skapp/community/timeplanner/repository/impl/TimeRecordRepositoryImpl.java
```diff
@@ -471,6 +475,90 @@ protected EmployeeTimeRecord buildEmptyTimeRecord(Long employeeId, LocalDate dat
...
+			LocalDate endDate, List<Predicate> additionalPredicates) {
+		CriteriaBuilder cb = entityManager.getCriteriaBuilder();
+		CriteriaQuery<Tuple> query = cb.createTupleQuery();
+		Root<TimeRecord> root = query.from(TimeRecord.class);
+
+		List<Predicate> predicates = new ArrayList<>(additionalPredicates);
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Dead parameter: `additionalPredicates` is never populated — all three callers (getEmployeeAttendanceSummary, findManagerAssignUsersAttendanceSummary, findTimeSheetSummaryData) pass a freshly created empty ArrayList, and because computeMergedAttendanceSummary is private it cannot be extended by an enterprise subclass either. Remove the parameter (and the `new ArrayList<>(additionalPredicates)` copy) to reduce noise, or, if it was added for a future use, document the intent.

backend/src/main/java/com/skapp/community/timeplanner/repository/impl/TimeRecordRepositoryImpl.java
```diff
@@ -471,6 +475,90 @@ protected EmployeeTimeRecord buildEmptyTimeRecord(Long employeeId, LocalDate dat
...
+
+		return new AttendanceSummaryDto(totalWork, totalBreak);
+	}
+
+	private float[] computeMergedHoursWithOverlap(float ciWork, float ciBreak, Long ciIn, Long ciOut, float manWork,
+			float manBreak, Long manIn, Long manOut) {
+		float overlapWorkDeduction = 0f;
+		float overlapBreakDeduction = 0f;
+
+		if (ciIn != null && ciOut != null && manIn != null && manOut != null && ciOut > manIn && ciIn < manOut) {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Overlap-dedup gap: computeMergedHoursWithOverlap only deducts overlapping hours when the manual record has BOTH non-null clockInTime and clockOutTime (manIn/manOut). Manual time entries frequently store worked/break hours without clock-in/out timestamps; in that case the guard short-circuits and the result is `ciWork + manWork` with no deduction, double-counting the period covered by both the clock-in record and the manual correction for the same employee/day. If manual entries can represent corrections to an existing clock-in day, this over-counts total hours. Consider falling back to a non-time-based reconciliation (e.g. preferring the latest manual record's hours, or capping the sum) when manual clock times are absent.

backend/src/main/java/com/skapp/community/timeplanner/repository/impl/TimeRecordRepositoryImpl.java
```diff
@@ -471,6 +475,90 @@ protected EmployeeTimeRecord buildEmptyTimeRecord(Long employeeId, LocalDate dat
...
+				cb.tuple(root.get(TimeRecord_.timeRecordId), root.get(TimeRecord_.employee).get(Employee_.employeeId),
+						root.get(TimeRecord_.date), cb.coalesce(root.get(TimeRecord_.workedHours), 0.0f),
+						cb.coalesce(root.get(TimeRecord_.breakHours), 0.0f), root.get(TimeRecord_.isManual),
+						root.get(TimeRecord_.clockInTime), root.get(TimeRecord_.clockOutTime)));
+		query.where(predicates.toArray(new Predicate[0]));
+
+		List<Tuple> records = entityManager.createQuery(query).getResultList();
+
+		Map<String, List<Tuple>> grouped = records.stream()
+			.collect(Collectors.groupingBy(t -> t.get(1, Long.class) + "_" + t.get(2, LocalDate.class)));
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Grouping-key inconsistency: this method keys groups as `employeeId + "_" + date` (line 500), while findEmployeesTimeRecords keys them as `date + "_" + employeeId` (line 345). Both are internally consistent so there is no bug, but since the two blocks are otherwise parallel logic, aligning the key format (ideally via the shared helper suggested above) avoids confusion for future readers.

backend/src/main/java/com/skapp/community/timeplanner/repository/impl/TimeRecordRepositoryImpl.java
```diff
@@ -108,62 +94,35 @@ public Optional<TimeRecord> findIncompleteClockOutTimeRecords(LocalDate lastCloc
...
-		TypedQuery<AttendanceSummaryDto> typedQuery = entityManager.createQuery(criteriaQuery);
-		try {
-			return typedQuery.getSingleResult();
-		}
-		catch (NoResultException e) {
-			return new AttendanceSummaryDto(0.0F, 0.0F);
-		}
+		List<Long> filterEmployeeIds = employeeIds.isEmpty() ? null : employeeIds;
+		List<Predicate> additionalPredicates = new ArrayList<>();
+		return computeMergedAttendanceSummary(filterEmployeeIds, startDate, endDate, additionalPredicates);
```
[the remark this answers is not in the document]
yuhanga-rtc: Done

backend/src/main/java/com/skapp/community/timeplanner/repository/impl/TimeRecordRepositoryImpl.java
```diff
@@ -108,62 +94,35 @@ public Optional<TimeRecord> findIncompleteClockOutTimeRecords(LocalDate lastCloc
...
 	}
 
 	@Override
 	public TimeSheetSummaryData findTimeSheetSummaryData(LocalDate startDate, LocalDate endDate,
 			List<Long> employeeIds) {
+		AttendanceSummaryDto mergedSummary = computeMergedAttendanceSummary(employeeIds, startDate, endDate,
+				new ArrayList<>());
```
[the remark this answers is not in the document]
yuhanga-rtc: Done

ThinuwanW (approved):

PR #1498, opened 2026-06-18
fix: address book not updated when terminated user is reactivated

akilarootcode:
## Summary
- Reactivating a terminated employee left their AddressBook entry with isActive = false permanently
- Root cause: the reactivation path in updateUserStatus() returned early without publishing any event, while termination correctly published UserDeactivatedEvent and triggered the address book deactivation listener
- Fix: introduce UserReactivatedEvent, publish it in the reactivation path, and add a listener that sets addressBook.isActive = true — symmetric with the deactivation flow

## Changes
- New: UserReactivatedEvent extending UserEvent (mirrors UserDeactivatedEvent)
- PeopleServiceImpl: publish UserReactivatedEvent after saving the reactivated user in updateUserStatus()
- Enterprise submodule (skapp-ep-be-src PR #686): add handleUserReactivation listener in UserDeactivatedEventHandler

## Test plan
- [ ] Terminate an employee and verify their address book entry becomes inactive
- [ ] Reactivate the same employee and verify their address book entry becomes active again
- [ ] Confirm no regression on fresh user creation: address book entry is created and active

Generated with Claude Code

### DIFF
--- a/backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
@@ -28,6 +28,7 @@ import com.skapp.community.common.util.MessageUtil;
 import com.skapp.community.common.util.Validation;
 import com.skapp.community.common.util.event.UserCreatedEvent;
 import com.skapp.community.common.util.event.UserDeactivatedEvent;
+import com.skapp.community.common.util.event.UserReactivatedEvent;
 import com.skapp.community.common.util.transformer.PageTransformer;
 import com.skapp.community.leaveplanner.type.ManagerType;
 import com.skapp.community.peopleplanner.constant.PeopleConstants;
@@ -2554,11 +2555,12 @@ public class PeopleServiceImpl implements PeopleService {
 			employee.setTerminationDate(null);
 			user.setIsActive(true);
 
+			userDao.save(user);
+			applicationEventPublisher.publishEvent(new UserReactivatedEvent(this, user));
 			updateSubscriptionQuantity(1L, true, false);
 			userVersionService.upgradeUserVersion(user.getUserId(), VersionType.MAJOR);
 			invalidateUserCache();
 			invalidateUserAuthPicCache();
-			userDao.save(user);
 			log.info("updateUserStatus: execution ended - user activated");
 			return;
 		}
--- a/backend/src/main/java/com/skapp/community/common/util/event/UserReactivatedEvent.java
+++ b/backend/src/main/java/com/skapp/community/common/util/event/UserReactivatedEvent.java
@@ -1,0 +1,11 @@
+package com.skapp.community.common.util.event;
+
+import com.skapp.community.common.model.User;
+
+public class UserReactivatedEvent extends UserEvent {
+
+	public UserReactivatedEvent(Object source, User user) {
+		super(source, user);
+	}
+
+}


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds a new `UserReactivatedEvent` (mirroring the existing `UserDeactivatedEvent`/`UserCreatedEvent` classes exactly) and publishes it in `updateUserStatus` on the reactivation path, consistent with the existing deactivation publish at line 2616. The event class and publish call are idiomatic and correct. The key risk is that no `@EventListener` for `UserReactivatedEvent` exists in the repository, so the event has no consumer and the intended address-book reactivation will not happen unless a handler is supplied (e.g., in the enterprise submodule).

Found **2** new issue(s): 🟣 1 nit(s)

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
```diff
@@ -2559,6 +2560,7 @@ public void updateUserStatus(Long userId, AccountStatus status, boolean isDelete
 			invalidateUserCache();
 			invalidateUserAuthPicCache();
 			userDao.save(user);
+			applicationEventPublisher.publishEvent(new UserReactivatedEvent(this, user));
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

The commit publishes `UserReactivatedEvent` to 'sync address book on user reactivation', but there is no `@EventListener`/`@TransactionalEventListener` handling this event anywhere in the codebase. The only event handlers present are `UserCreateEventHandler`, `UserDeactivatedEventHandler`, and `QuartzJobHandler` (under enterprise/common/component/event/handler). `UserDeactivatedEventHandler.handleUserDeactivation` sets `addressBook.setIsActive(false)` on deactivation, but nothing sets it back to `true` on reactivation. As-is, this event is a no-op and the address book stays inactive after a user is reactivated. Add a corresponding handler (e.g., `UserReactivatedEventHandler` calling `addressBookDao.findByInternalUser(user).ifPresent(ab -> { ab.setIsActive(true); addressBookDao.save(ab); })`). If that handler lives in the enterprise submodule and is being added in a separate PR, please confirm/link it so this change isn't merged without its consumer.

ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Publish-ordering asymmetry with the deactivation path. In the deactivation branch the event is published immediately after `userDao.save(user)` and before the subscription/version/cache-invalidation steps (lines 2615-2621), whereas here the reactivation event is published only after updateSubscriptionQuantity, upgradeUserVersion, invalidateUserCache and invalidateUserAuthPicCache have already run. The two status-change paths are conceptually symmetric, so for consistency and to reduce surprise consider matching the ordering (publish right after the save, as deactivation does). Not a functional bug given the listener only reads the user, but worth aligning.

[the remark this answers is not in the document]
akilarootcode: Both points here are incorrect:

**1. Test coverage** — The reactivation endpoint is already covered by existing integration tests in `PeopleControllerIntegrationTest`:
- `reactivateTerminatedUser_WithValidTerminatedUser_ReturnsSuccess` (terminates the user first, then calls `/v1/people/user/reactivate/{id}` and asserts HTTP 200) — this exercises the full service path including the event publication.
- `reactivateTerminatedUser_WithActiveUser_ReturnsBadRequest`
- `reactivateTerminatedUser_WithNonExistentId_ReturnsNotFound`

**2. PR #686 link** — The companion PR #686 referenced in the description is in the **`rootcodelabs/skapp-ep-be-src`** repository (the enterprise backend submodule), not in `SkappHQ/skapp`. Copilot appears to have looked up PR #686 in the current repo instead. The link is correct as written.

akilarootcode: This is incorrect. The `UserReactivatedEvent` listener is added in the companion enterprise submodule PR **rootcodelabs/skapp-ep-be-src#686** (`UserDeactivatedEventHandler.handleUserReactivation`). These two PRs are designed to be merged together — the event is **published here** (community module) and **consumed there** (enterprise module). The finding is based on searching within this repo only, but the enterprise module lives in a separate git submodule repository.

akilarootcode: Valid observation. The event publish will be moved to immediately after `userDao.save(user)` to match the ordering in the deactivation path.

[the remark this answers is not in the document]
akilarootcode: its there in submodule

yuhanga-rtc (approved):

shakila-kamalasena (approved):

ThinuwanW (approved):

PR #1503, opened 2026-06-18
feat: added deal and contact search criteria

charith-rc:
### Summary

- Added search by deal name and contact name criteria

### How to test

1.

### Project Checklist

  - [x] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmTaskRepositoryImpl.java
+++ b/backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmTaskRepositoryImpl.java
@@ -194,7 +194,13 @@ public class CrmTaskRepositoryImpl implements CrmTaskRepository {
 
 		if (params.getSearchKeyword() != null && !params.getSearchKeyword().isBlank()) {
 			String escaped = StringUtils.escapeLikePattern(params.getSearchKeyword().trim().toLowerCase());
-			predicates.add(cb.like(cb.lower(root.get(CrmTask_.name)), "%" + escaped + "%"));
+
+			Join<CrmTask, CrmContact> contactJoin = root.join(CrmTask_.contact, JoinType.LEFT);
+			Join<CrmTask, CrmDeal> dealJoin = root.join(CrmTask_.deal, JoinType.LEFT);
+
+			predicates.add(cb.or(cb.like(cb.lower(root.get(CrmTask_.name)), "%" + escaped + "%"),
+					cb.like(cb.lower(contactJoin.get(CrmContact_.name)), "%" + escaped + "%"),
+					cb.like(cb.lower(dealJoin.get(CrmDeal_.name)), "%" + escaped + "%")));
 		}
 
 		if (params.getContactId() != null) {
--- a/backend/src/test/java/com/skapp/community/crmplanner/controller/v1/CrmTaskControllerIntegrationTest.java
+++ b/backend/src/test/java/com/skapp/community/crmplanner/controller/v1/CrmTaskControllerIntegrationTest.java
@@ -240,16 +240,37 @@ class CrmTaskControllerIntegrationTest {
 	}
 
 	@Test
-	@DisplayName("Get tasks with search keyword - Returns only matching tasks")
-	void getTasks_WithSearchKeyword_ReturnsMatchingTasks() throws Exception {
-		savedTask("Follow up call", false, false);
-		savedTask("Send proposal", false, false);
+	@DisplayName("Get tasks with search keyword matching contact name - Returns matching tasks")
+	void getTasks_WithSearchKeywordMatchingContactName_ReturnsMatchingTasks() throws Exception {
+		CrmContact contact = new CrmContact();
+		contact.setName("John Doe");
+		contact.setEmail("john@example.com");
+		contact.setOwner(employeeDao.getReferenceById(1L));
+		contact = crmContactDao.save(contact);
 
-		performGetRequest(authToken, "follow", null, null).andDo(print())
+		savedTask("Task 1", false, false, contact.getId());
+		savedTask("Task 2", false, false);
+
+		performGetRequest(authToken, "john", null, null).andDo(print())
 			.andExpect(status().isOk())
 			.andExpect(jsonPath(STATUS_PATH).value(STATUS_SUCCESSFUL))
 			.andExpect(jsonPath(RESULTS_0_PATH + "['tasks'].length()").value(1))
-			.andExpect(jsonPath(RESULTS_0_PATH + "['tasks'][0]['name']").value("Follow up call"));
+			.andExpect(jsonPath(RESULTS_0_PATH + "['tasks'][0]['name']").value("Task 1"));
+	}
+
+	@Test
+	@DisplayName("Get tasks with search keyword matching deal name - Returns matching tasks")
+	void getTasks_WithSearchKeywordMatchingDealName_ReturnsMatchingTasks() throws Exception {
+		CrmDeal deal = savedDeal("Special Deal", crmContactDao.getReferenceById(contactId), null);
+
+		savedTask("Task 1", false, false, contactId, deal);
+		savedTask("Task 2", false, false);
+
+		performGetRequest(authToken, "special", null, null).andDo(print())
+			.andExpect(status().isOk())
+			.andExpect(jsonPath(STATUS_PATH).value(STATUS_SUCCESSFUL))
+			.andExpect(jsonPath(RESULTS_0_PATH + "['tasks'].length()").value(1))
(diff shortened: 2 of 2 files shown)


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR extends CRM task keyword search to also match linked contact and deal names by adding LEFT joins in buildTaskPredicates. The change is functionally correct — contact/deal are @ManyToOne so the joins don't multiply rows or inflate the count query, and escaping reuses the existing escapeLikePattern consistently. Two minor points: the new search joins duplicate the joins already created by applyFetchGraph in the list/page query paths, and the test renames dropped the only coverage for searching completed tasks by task name.

Found **4** new issue(s):

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmTaskRepositoryImpl.java
```diff
@@ -194,7 +194,13 @@ private List<Predicate> buildTaskPredicates(CriteriaBuilder cb, Root<CrmTask> ro
 
 		if (params.getSearchKeyword() != null && !params.getSearchKeyword().isBlank()) {
 			String escaped = StringUtils.escapeLikePattern(params.getSearchKeyword().trim().toLowerCase());
-			predicates.add(cb.like(cb.lower(root.get(CrmTask_.name)), "%" + escaped + "%"));
+
+			Join<CrmTask, CrmContact> contactJoin = root.join(CrmTask_.contact, JoinType.LEFT);
```
ThinuwanW, line 198: 🤖 **Claude** · 💬 **Note**

In the findTasks and findCompletedTasks (content) paths, applyFetchGraph() already creates LEFT JOIN FETCH for CrmTask_.contact and CrmTask_.deal (lines 224-225). The new root.join(CrmTask_.contact, LEFT) / root.join(CrmTask_.deal, LEFT) here are tracked separately by Hibernate and produce a second pair of LEFT JOINs to crm_contact/crm_deal in the generated SQL whenever a search keyword is present — redundant work for those queries. The count query genuinely needs these joins (it doesn't call applyFetchGraph), so they can't simply be removed. Consider reusing the fetched joins for the predicate (e.g. capture the Fetch as a Join and reuse it, or only add explicit joins when fetch joins aren't present) to avoid the duplicate joins. Functionally harmless since these are to-one associations, hence suggestion rather than important.

backend/src/test/java/com/skapp/community/crmplanner/controller/v1/CrmTaskControllerIntegrationTest.java
```diff
@@ -383,16 +404,38 @@ void getCompletedTasks_WithNoCompletedTasks_ReturnsEmptyPage() throws Exception
...
+			.andExpect(status().isOk())
+			.andExpect(jsonPath(STATUS_PATH).value(STATUS_SUCCESSFUL))
+			.andExpect(jsonPath(RESULTS_0_PATH + "['items'].length()").value(1))
+			.andExpect(jsonPath(RESULTS_0_PATH + "['items'][0]['name']").value("Completed Task 1"))
+			.andExpect(jsonPath(RESULTS_0_PATH + "['totalItems']").value(1));
+	}
+
+	@Test
+	@DisplayName("Get completed tasks with search keyword matching deal name - Returns matching tasks")
+	void getCompletedTasks_WithSearchKeywordMatchingDealName_ReturnsMatchingTasks() throws Exception {
```
ThinuwanW, line 428: 🤖 **Claude** · 💬 **Note**

The renamed tests replaced the previous 'search by task name' cases with contact-name and deal-name cases. For the open-tasks path, name search is still covered by getTasks_SearchAndContactIdFilter_ReturnsIntersection (searches 'follow' against task name). For the completed-tasks path, however, there is now no test that verifies search matching the task's own name still works after this change. Consider keeping/adding a completed-tasks test that searches by task name to guard against regressions in the OR predicate's name branch.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmTaskRepositoryImpl.java
```diff
@@ -194,7 +194,13 @@ private List<Predicate> buildTaskPredicates(CriteriaBuilder cb, Root<CrmTask> ro
...
+			Join<CrmTask, CrmDeal> dealJoin = root.join(CrmTask_.deal, JoinType.LEFT);
+
+			predicates.add(cb.or(cb.like(cb.lower(root.get(CrmTask_.name)), "%" + escaped + "%"),
```
ThinuwanW, line 201: 🤖 **Claude** · 💬 **Note**

The three new/modified cb.like(...) calls pass the pattern but no ESCAPE character: cb.like(cb.lower(...), "%" + escaped + "%"). escapeLikePattern() prefixes a backslash to wildcard chars (% _ \), but cb.like with no escape arg does not declare a backslash escape, so the escaping is only honored if the DB happens to treat backslash as the default LIKE escape (MySQL does by default, but breaks under sql_mode=NO_BACKSLASH_ESCAPES). Every other CRM repository pairs escapeLikePattern with the escape char, e.g. CrmContactRepositoryImpl.java:88-90, CrmCompanyRepositoryImpl.java:185, CrmContactOwnerRepositoryImpl.java:91-93 all use cb.like(..., likePattern, '\\'). For consistency and correct wildcard neutralization, add the '\\' escape argument to all three like() calls here (the contact-name and deal-name lines being new make this a fresh occurrence).

backend/src/test/java/com/skapp/community/crmplanner/controller/v1/CrmTaskControllerIntegrationTest.java
```diff
@@ -383,16 +404,38 @@ void getCompletedTasks_WithNoCompletedTasks_ReturnsEmptyPage() throws Exception
...
+
+		performGetCompletedRequest(authToken, "0", "10", "jane", null, null).andDo(print())
+			.andExpect(status().isOk())
+			.andExpect(jsonPath(STATUS_PATH).value(STATUS_SUCCESSFUL))
+			.andExpect(jsonPath(RESULTS_0_PATH + "['items'].length()").value(1))
+			.andExpect(jsonPath(RESULTS_0_PATH + "['items'][0]['name']").value("Completed Task 1"))
+			.andExpect(jsonPath(RESULTS_0_PATH + "['totalItems']").value(1));
+	}
+
+	@Test
```
ThinuwanW, line 426: 🤖 **Claude** · 💬 **Note**

The original getCompletedTasks_WithSearchKeyword test matched the keyword 'call' against the task's own NAME ('Completed call'). Both replacement tests (matching contact name and deal name) only exercise the contactJoin/dealJoin branches; nothing in the completed-tasks path now searches against root.get(CrmTask_.name). The open-tasks path retains incidental name coverage via getTasks_SearchAndContactIdFilter, but the completed path has none. Add a completed-tasks test that searches by the task's own name to keep that OR branch covered.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmTaskRepositoryImpl.java
charith-rc: we dont need to explicitly set '\' as escape character in our case since MySQL defaults to it

backend/src/test/java/com/skapp/community/crmplanner/controller/v1/CrmTaskControllerIntegrationTest.java
charith-rc: intersection implies own path works anyway, omitting to prevent bulking tests

backend/src/test/java/com/skapp/community/crmplanner/controller/v1/CrmTaskControllerIntegrationTest.java
charith-rc: added one happy path test 6bb4d548c8bab4c47135a6dfafb4c5b0aa76893a

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmTaskRepositoryImpl.java
charith-rc: no need since only one record can match against a task from each entity. would also require breaking up the applyFetchGraph function.

anusham-rc (approved):

sanjula-rtc (approved):

akilarootcode (approved):

ThinuwanW (approved):

ThinuwanW (approved):

PR #1490, opened 2026-06-16
Feat/crm 197 open tasks view

charith-rc:
### Summary

- Scaffolding for my tasks and team tasks tab views with role restrictions

### How to test

1.

### Project Checklist

- [x] Changes build without any errors
- [x] Have written adequate test cases
- [x] Done developer testing in
  - [x] Chrome
  - [ ] Firefox
  - [ ] Safari
- [x] Code is formatted with `npm run format`
- [x] Code is linted with `npm run check-lint`
- [x] No unnecessary comments left in code
- [ ] Made corresponding changes to the documentation

#### Other

- [x] New atomic components added
- [x] New molecules added
- [ ] New pages(routes) added
- [ ] New dependencies installed

### PR Checklist

- [x] Pull request is raised from the correct source branch
- [x] Pull request is raised to the correct destination branch
- [x] Pull request is raised with correct title
- [x] Pull request is self reviewed
- [x] Pull request is self assigned
- [ ] Suitable pull request status labels are added (`ready-for-code-review`)

### Additional Information
- requires the TaskRow component from https://github.com/SkappHQ/skapp/pull/1373

PR #1506, opened 2026-06-19
Feat/crm 197 view tasks completed

charith-rc:
### Summary

- Extends task view to include completed tasks. adds paginated list of completed tasks with search enabled

### How to test

1.

### Project Checklist

  - [x] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
- Chain PR from : https://github.com/SkappHQ/skapp/pull/1490
- Includes base of tasks page and structure, required for this PR


frontend/src/community/crm/hooks/useGetTasksTabs.tsx
```diff
@@ -1,24 +1,34 @@
...
-      component: <></>
+      id: CrmTaskTabEnum.COMPLETED_TASKS,
+      label: translateText(["completedTasks"])
     }
   ];
+
+  if (!(isCrmSalesManager ?? false)) {
+    return allTabs.filter(
+      (tab) => !SALES_REP_RESTRICTED_TASK_TABS.includes(tab.id)
+    );
```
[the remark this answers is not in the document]
charith-rc: checks from the hierarchy of the roles, crm sales manager and upward would not be restricted.

logic is:
if not CrmSalesManager (which means only CrmSalesRep), filter out all tabs in the SALES_REP_RESTRICTED_TASK_TABS array

frontend/src/community/crm/components/atoms/TaskGroup/TaskGroup.tsx
```diff
@@ -0,0 +1,34 @@
...
+    <div className="flex flex-col">
+      {label && (
+        <div className="subtitle2 mb-2 sticky top-0 bg-white z-10">{label}</div>
+      )}
+      <div className="border border-secondary-accent rounded-lg overflow-hidden divide-y divide-secondary-accent">
+        {tasks.map((task) => (
+          <div key={task.id}>
+            {/* <TaskRow task={task} isCheckTaskVisible={isCheckTaskVisible} /> */}
+          </div>
+        ))}
```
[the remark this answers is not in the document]
charith-rc: requires integration from separate dependent PR, will be integrated as soon as it is merged

frontend/src/community/crm/api/TaskApi.ts
```diff
@@ -49,3 +70,37 @@ export const useUpdateTaskCompletion = (
...
+        page: pageParam,
+        size,
+        searchKeyword
+      }),
+    getNextPageParam: (lastPage) => {
+      const nextPage = lastPage.currentPage + 1;
+      return nextPage < lastPage.totalPages ? nextPage : undefined;
+    }
+  });
+};
```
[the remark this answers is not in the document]
charith-rc: uses a constant default page size. this will not change. no need to add size to query params.

ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds a 'Completed Tasks' tab to the CRM tasks table with an infinite-scroll, debounced-search list backed by a new `useInfiniteQuery` hook and a reusable `useInfiniteScroll` IntersectionObserver hook. The implementation is solid and follows existing patterns (query key factories, endpoint constants, `useDebounce`, empty/error states), but there are a few cleanups around dead code, null-safety in pagination, an empty sentinel element, and notable markup duplication with `OpenTasksTabContent`.

Found **10** new issue(s): 🟡 6 suggestion(s), 🟣 4 nit(s)

frontend/src/community/crm/api/TaskApi.ts
```diff
@@ -64,3 +70,37 @@ export const useUpdateTaskCompletion = (
...
```
ThinuwanW, line 119: 🤖 **Claude** · 🟡 **Suggestion**

`getNextPageParam` dereferences `lastPage.currentPage`/`lastPage.totalPages` without null-safety, yet `fetchCompletedTasks` returns `response?.data?.results?.[0]` which can be `undefined` at runtime (and the consumer's `useMemo` already defensively handles `page?.items ?? []`). A malformed success response would throw inside `getNextPageParam`. Guard it, e.g. `const nextPage = (lastPage?.currentPage ?? 0) + 1; return lastPage && nextPage < lastPage.totalPages ? nextPage : undefined;`

frontend/src/community/crm/api/utils/QueryKeys.ts
```diff
@@ -53,5 +53,10 @@ export const companyQueryKeys = {
 
 export const taskQueryKeys = {
   GET_TASK_DATA: ["get-task-data"],
-  GET_OPEN_TASKS: ["get-open-tasks"]
+  GET_OPEN_TASKS: ["get-open-tasks"],
+  GET_COMPLETED_TASKS: ["get-completed-tasks"],
```
ThinuwanW, line 59: 🤖 **Claude** · 🟡 **Suggestion**

`GET_COMPLETED_TASKS: ["get-completed-tasks"]` is dead code — only `GET_COMPLETED_TASKS_BY_SEARCH(...)` is referenced anywhere (verified via grep). Remove the unused standalone key to avoid confusion, or use it as the shared root for invalidation.

frontend/src/community/crm/components/molecules/TaskTabContent/TaskTabContent.tsx
```diff
@@ -0,0 +1,119 @@
...
+          }
+          icon={<SearchIcon />}
+        />
+      );
+    }
+
+    return (
+      <div className="flex flex-col flex-1 min-h-0 px-2 pb-4 gap-4 overflow-y-auto">
+        <TaskGroup tasks={tasks} isCheckTaskVisible={false} />
+        <div ref={loadingRef} />
```
ThinuwanW, line 160: 🤖 **Claude** · 🟡 **Suggestion**

The infinite-scroll sentinel `<div ref={loadingRef} />` has zero height. Combined with the `useInfiniteScroll` default `threshold = 1.0` (requires 100% of the target visible), IntersectionObserver triggering on a zero-area element is browser-dependent and may not fire reliably. Give the sentinel a small height/min-height (or pass `threshold={0}`/a `rootMargin`) so `onLoadMore` triggers consistently. Also consider rendering a loading indicator here while `isFetchingNextPage`.

frontend/src/community/crm/components/atoms/TaskGroup/TaskGroup.tsx
```diff
@@ -5,17 +5,26 @@ import { CrmTaskDetailType } from "~community/crm/types/CommonTypes";
...
+  label?: string;
   tasks: CrmTaskDetailType[];
+  isCheckTaskVisible?: boolean;
 }
 
-const TaskGroup: FC<TaskGroupProps> = ({ label, tasks }) => {
+const TaskGroup: FC<TaskGroupProps> = ({
+  label,
+  tasks,
+  isCheckTaskVisible = true
```
ThinuwanW, line 16: 🤖 **Claude** · 🟣 **Nit**

The new `isCheckTaskVisible` prop is effectively dead — it is only referenced inside the commented-out `TaskRow` line (line 26) and has no runtime effect since `TaskRow` rendering is commented out. The tab content passes `isCheckTaskVisible={false}` but it does nothing yet. This is acceptable as WIP scaffolding, but flag so it isn't forgotten when `TaskRow` is wired up.

frontend/src/community/crm/components/molecules/TaskTabContent/TaskTabContent.tsx
```diff
@@ -0,0 +1,119 @@
+import {
```
ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Significant markup duplication with `OpenTasksTabContent.tsx`: the search `InputField`, the `isLoading` skeleton, the `isError` `EmptyDataView`, and the empty-state (`emptyDataState`/`emptySearchState`) branch are nearly identical between the two tab components. Per the guidelines (extract shared component when 2+ components have near-identical markup), consider extracting a shared wrapper (e.g. `TasksTabShell` handling search header + loading/error/empty states) and passing the list body as children.

frontend/src/community/crm/api/TaskApi.ts
```diff
@@ -64,3 +70,37 @@ export const useUpdateTaskCompletion = (
...
+  const response = await authFetch.get(taskEndpoints.GET_COMPLETED_TASKS, {
+    params: { page, size, searchKeyword }
+  });
+  return response?.data?.results?.[0];
+};
+
+export const useGetCompletedTasks = (searchKeyword: string, size: number) => {
+  return useInfiniteQuery({
+    initialPageParam: 0,
+    queryKey: taskQueryKeys.GET_COMPLETED_TASKS_BY_SEARCH(searchKeyword),
```
ThinuwanW, line 106: 🤖 **Claude** · 🟣 **Nit**

`size` is used in `queryFn` but is not part of the query key (`GET_COMPLETED_TASKS_BY_SEARCH(searchKeyword)` keys only on the search term). It works today because `size` is always the constant `DEFAULT_PAGE_SIZE`, but if a caller ever passes a different page size the cache won't differentiate. Consider including `size` in the key factory for correctness.

frontend/src/community/common/hooks/useInfiniteScroll.ts
```diff
@@ -0,0 +1,46 @@
...
+    [hasNextPage, isLoading, onLoadMore],
+  );
+
+  useEffect(() => {
+    const element = loadingRef.current;
+    if (!element) return;
+
+    const observer = new IntersectionObserver(handleObserver, {
+      threshold,
+      rootMargin,
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Fragile sentinel attachment for a reusable hook. The effect that creates the IntersectionObserver depends only on [handleObserver, threshold, rootMargin]; it reads loadingRef.current but does not react to the element mounting later. In CompletedTasksTabContent the sentinel is conditionally rendered (only in the success branch, not during the skeleton/loading branch), so the observer only attaches because handleObserver happens to change when `hasNextPage` flips false→true after the first page resolves. A future consumer with a stable onLoadMore and a sentinel that mounts without any of these deps changing would never get an observer attached. Make this robust by using a callback ref (set state when the node mounts and include it in the deps) instead of relying on dependency churn to re-run the effect after the element appears.

frontend/src/community/common/hooks/useInfiniteScroll.ts
```diff
@@ -0,0 +1,46 @@
+'use client';
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Inconsistent with the established infinite-scroll mechanism. Other CRM list views (ContactTable, CompanyTable, DealsSection) already implement infinite scrolling through the shared `Table` component from @rootcodelabs/skapp-ui via its `hasMore`/`onLoadMore`/`infiniteScrollLoadingMessage` props (see ContactTable.tsx loadMore + Table usage). This PR introduces a parallel, bespoke IntersectionObserver hook to do the same job. Since TaskGroup isn't a Table the existing props can't be used verbatim, but consider whether the shared component/abstraction can be reused (or the new hook adopted by the existing tables) so the codebase has one infinite-scroll pattern rather than two.

ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`'use client'` is an App Router directive and is a no-op in this Next.js 14 Pages Router project — every component here is already client-side. It can be removed; the only other occurrences are a handful of enterprise files and are similarly unnecessary.

ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

New shared/reusable hook added under common/hooks with no corresponding test. Unlike the feature components, a generic hook intended for reuse benefits from a unit test (e.g. asserting onLoadMore fires only when isIntersecting && hasNextPage && !isLoading, and that the observer disconnects on unmount). Per the testing guideline, consider adding useInfiniteScroll.test.ts.

charith-rc: same job but different design. this is required to handle the design's intended behaviour. using a Table component here is incorrect

charith-rc: this hook is a one to one copy from the PM module

frontend/src/community/common/hooks/useInfiniteScroll.ts
charith-rc: this hook is a one to one copy from the PM module

frontend/src/community/common/hooks/useInfiniteScroll.ts
charith-rc: this hook is a one to one copy from the PM module

frontend/src/community/crm/components/molecules/TaskTabContent/TaskTabContent.tsx
charith-rc: refactored components to sit in one TasksTabContent component cleanly 2f57672c816bde227637c08dfb8de09f46b20788

frontend/src/community/crm/components/molecules/TaskTabContent/TaskTabContent.tsx
charith-rc: existing usage pattern from PM module

frontend/src/community/crm/api/utils/QueryKeys.ts
charith-rc: still used in some in progress implementations. will remove this once integrated

frontend/src/community/crm/api/TaskApi.ts
charith-rc: backend will always return a result as defined in the existing structure, therefore this is unnecesssary

frontend/src/community/crm/api/TaskApi.ts
charith-rc: size is a constant. no need to bulk query key

frontend/src/community/crm/components/atoms/TaskGroup/TaskGroup.tsx
charith-rc: will be used with task row integration

anusham-rc (dismissed):

thusala (dismissed):

thusala (approved):

ThinuwanW (approved):

PR #1526, opened 2026-06-22
feat: add metrics section to contact details panel

dulsara-manakal:
### Summary

- Wired up with a metrics card in the contact side panel. 
- add a mapper to  map metrics card  API data

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR cleanly wires a new metrics card section into the contact side panel via a pure mapping util and adds the corresponding i18n keys. The wiring, types, query gating, and translation namespace are all correct. Two minor improvements are worth considering around i18n interpolation idiom and the hardcoded currency symbol.

Found **6** new issue(s): 🟡 1 suggestion(s), 🟣 1 nit(s)

frontend/src/community/crm/utils/contactUtil.ts
```diff
@@ -31,3 +36,45 @@ export const mergeAndPrioritizeCompanyDropdownItems = (
...
 };
+
+export const mapContactToMetricItems = (
+  contact: CrmContactDetailResponseType,
+  translateText: (keys: string[]) => string
+): MetricItem[] => {
+  const overdueChip =
+    contact.overdueTasksCount > 0
+      ? {
+          label: translateText(["metrics", "overdueChipLabel"]).replace(
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The overdue label is built with a manual `String.replace("{{count}}", ...)` instead of i18next's built-in interpolation, which is the established pattern everywhere else in this codebase (e.g. `translateText(["..."], { name })` in BulkUploadSummary, TransferMembersModal, and the sibling `{{name}}` keys in this same crmModule.json). The manual replace is brittle (breaks if the placeholder format/whitespace changes or is reordered in another locale) and the narrowed param type `translateText: (keys: string[]) => string` is what forces it — the real `useTranslator` return type accepts a second `interpolationValues` arg. Fix: widen the param type to `(keys: string[], interpolationValues?: Record<string, unknown>) => string` and call `translateText(["metrics", "overdueChipLabel"], { count: contact.overdueTasksCount })`. (Note: i18next treats `count` as the pluralization key; since no `_one`/`_other` variants are defined it falls back to plain interpolation, so this is safe, but be aware if plural forms are added later.)

frontend/src/community/crm/utils/contactUtil.ts
```diff
@@ -31,3 +36,45 @@ export const mergeAndPrioritizeCompanyDropdownItems = (
...
+    },
+    {
+      id: "activeDealsCount",
+      title: translateText(["metrics", "activeDeals"]),
+      amount: String(contact.activeDealsCount)
+    },
+    {
+      id: "totalRevenue",
+      title: translateText(["metrics", "totalRevenue"]),
+      amount: contact.totalRevenue,
```
ThinuwanW, line 72: 🤖 **Claude** · 🟣 **Nit**

`totalRevenue` and `pipelineRevenue` are flagged `isCurrency: true`, which routes them through `formatValue` in crmUtil.ts — that helper hardcodes a `$` prefix. `CrmContactDetailResponseType` carries no `currencyCode`, so revenue is silently rendered as USD regardless of the contact's actual currency. Pre-existing limitation of `formatValue` (also used for company metrics), but this PR newly surfaces it for contact revenue. Consider threading a currency code through or confirming USD-only is acceptable for community mode.

frontend/src/community/crm/components/organisms/ContactSidePanel/ContactSidePanel.tsx
```diff
@@ -98,24 +100,28 @@ const ContactSidePanel: FC<SidePanelProps> = ({ isOpen, onClose }) => {
         />
       }
     >
-      <div className="flex flex-col pb-4 gap-4">
-        {contact && (
+      {contact && (
```
ThinuwanW, line 108: 🤖 **Claude** · 💬 **Note**

Behavioral drift from the refactor: previously the `Tabs`, the `<hr>`, and `renderTabContent()` rendered regardless of whether `contact` was loaded (only `SidePanelContactInfo` was gated). Now the entire body is wrapped in `{contact && (...)}`, so while `useGetContactById` is fetching the panel body is completely blank. The hook only exposes `isError` here — `isLoading`/`isFetching` is never consumed, so there is no skeleton/spinner during the fetch (Checklist #4 loading states, #5 behavioral preservation). Consider rendering a loading/skeleton state while the query is pending instead of an empty body.

frontend/src/community/crm/components/organisms/ContactSidePanel/ContactSidePanel.tsx
```diff
@@ -98,24 +100,28 @@ const ContactSidePanel: FC<SidePanelProps> = ({ isOpen, onClose }) => {
...
           />
-        )}
 
-        <div className="flex flex-col pt-2 w-full">
-          <Tabs
-            tabs={tabs}
-            activeTabId={activeTab}
-            onTabChange={(tabId) => setActiveTab(tabId as SidePanelTabEnum)}
+          <SidePanelMetricCards
+            metrics={mapContactToMetricItems(contact, translateText)}
```
ThinuwanW, line 116: 🤖 **Claude** · 💬 **Note**

`mapContactToMetricItems(contact, translateText)` is called inline in JSX, so a fresh `MetricItem[]` (and 5 `translateText` calls) is recomputed on every render of the panel, even when `contact` is unchanged. Wrap it in `useMemo(() => mapContactToMetricItems(contact, translateText), [contact])` to avoid the per-render allocation passed down to the memoizable card list.

frontend/src/community/crm/utils/contactUtil.ts
```diff
@@ -31,3 +36,45 @@ export const mergeAndPrioritizeCompanyDropdownItems = (
 
   return [...prioritized, ...deprioritized];
 };
+
+export const mapContactToMetricItems = (
+  contact: CrmContactDetailResponseType,
+  translateText: (keys: string[]) => string
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

The `translateText` parameter is locally re-declared as `(keys: string[]) => string`, which duplicates the actual return type of `useTranslator` and silently drops its optional second `interpolationValues?: Record<string, any>` argument. This both diverges from the source-of-truth signature (will need manual updating if the translator changes) and structurally prevents this helper from ever using i18next interpolation. Prefer exporting/reusing the translator's return type (e.g. `ReturnType<typeof useTranslator>`) rather than re-typing it inline.

frontend/src/community/crm/utils/contactUtil.ts
```diff
@@ -31,3 +36,45 @@ export const mergeAndPrioritizeCompanyDropdownItems = (
...
+          ),
+          variant: CrmMetricLabelThemeEnum.RED
+        }
+      : undefined;
+
+  return [
+    {
+      id: "openTasksCount",
+      title: translateText(["metrics", "openTasks"]),
+      amount: String(contact.openTasksCount),
```
ThinuwanW, line 61: 🤖 **Claude** · 💬 **Note**

Null-safety nit (Checklist #2): `String(contact.openTasksCount)` / `String(contact.activeDealsCount)` will render the literal string "undefined" in the UI if the backend response ever omits these numeric fields. The type marks them as required `number`, but if there is any chance of drift, default them (e.g. `String(contact.openTasksCount ?? 0)`) so the metric card shows a sane value instead of "undefined".

ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds a metrics section (open tasks, active deals, revenue) to the contact details side panel via a new `mapContactToMetricItems` util, a translation block, and a component refactor. The changes are type-safe and the translation keys map correctly. Findings are minor: a deviation from the established i18next interpolation pattern, a loading-state gap introduced by the refactor, and a small render-performance nit.

_4 finding(s) already posted in a previous review._

frontend/src/community/crm/utils/contactUtil.ts
```diff
@@ -31,3 +36,45 @@ export const mergeAndPrioritizeCompanyDropdownItems = (
 
   return [...prioritized, ...deprioritized];
 };
+
+export const mapContactToMetricItems = (
```
ThinuwanW, line 40: 🤖 **Claude** · 🟡 **Suggestion**

`mapContactToMetricItems` is a new exported util containing branching logic (the conditional `overdueChip` is only present when `overdueTasksCount > 0`), but no `.test.ts` was added for it. Per the testing guidelines, add a unit test covering both the overdue (chip present, RED variant, count interpolated) and non-overdue (chip undefined) paths, plus the currency vs. non-currency `amount` mapping. A pure mapping function like this is cheap to test and protects against future drift in the metric ordering/ids.

frontend/src/community/crm/components/organisms/ContactSidePanel/ContactSidePanel.tsx
```diff
@@ -98,24 +100,28 @@ const ContactSidePanel: FC<SidePanelProps> = ({ isOpen, onClose }) => {
...
           />
+
+          <div className="flex flex-col pt-2 w-full">
+            <Tabs
+              tabs={tabs}
+              activeTabId={activeTab}
+              onTabChange={(tabId) => setActiveTab(tabId as SidePanelTabEnum)}
+            />
+            <hr className="border-secondary-accent" />
```
ThinuwanW, line 125: 🤖 **Claude** · 🟣 **Nit**

During the refactor the `<hr className="border-secondary-accent" />` was moved from being a direct child of the outer `gap-4` flex column (where it had gap spacing above and below) to inside the `flex flex-col pt-2 w-full` Tabs container directly after `<Tabs>` (no gap between them). This changes the divider's spacing/position relative to the previous layout. If the visual change is intentional, ignore; otherwise verify the divider still sits where intended relative to the tabs and tab content.

frontend/src/community/crm/utils/contactUtil.ts
dulsara-manakal: Fixed this.

frontend/src/community/crm/utils/contactUtil.ts
dulsara-manakal: This part will be implemented as a feature in the upcoming version.

frontend/src/community/crm/components/organisms/ContactSidePanel/ContactSidePanel.tsx
dulsara-manakal: Skelotons will add in another [PR](https://pr.skapp.dev/view/f68cf611-78df-4973-b855-38be78cacbdb).

frontend/src/community/crm/utils/contactUtil.ts
dulsara-manakal: Fixed by using `interpolationValues`.

frontend/src/community/crm/utils/contactUtil.ts
dulsara-manakal: BE will return these values correctly, and with other checks `contact &&` and `isLoading && ` , this will be an unnecessary fallback.

frontend/src/community/crm/components/organisms/ContactSidePanel/ContactSidePanel.tsx
dulsara-manakal: This is intentional.

frontend/src/community/crm/components/organisms/ContactSidePanel/ContactSidePanel.tsx
dulsara-manakal: This is only an array mapping with translation keys object lookup, this won't take a lot of time and will not affect the performance a lot.

frontend/src/community/crm/utils/contactUtil.ts
dulsara-manakal: Add a unit test for this util.

sanjula-rtc (dismissed):

hasalarootcode (dismissed):

ThinuwanW (dismissed):

PR #1528, opened 2026-06-22
fix(people): fix subscription increment for deactivated guest user conversion

akilarootcode:
Captured wasDeactivatedGuest flag before mutating accountStatus to PENDING, so deactivated guest -> employee conversions correctly trigger subscription increment. Simplified redundant null/role checks already guaranteed by findGuestUserByEmail filter.

### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
@@ -214,10 +214,12 @@ public class PeopleServiceImpl implements PeopleService {
 		boolean isGuestConversion = false;
 
 		Optional<User> existingGuestUser = findGuestUserByEmail(requestDto);
+		boolean wasDeactivatedGuest = false;
 		if (existingGuestUser.isPresent()) {
 			user = existingGuestUser.get();
 			user.setIsActive(true);
 			employee = user.getEmployee();
+			wasDeactivatedGuest = employee.getAccountStatus() == AccountStatus.DEACTIVATED;
 			// Reset to PENDING so createUserEntity triggers credential setup and
 			// invitation email
 			employee.setAccountStatus(AccountStatus.PENDING);
@@ -243,6 +245,10 @@ public class PeopleServiceImpl implements PeopleService {
 			peopleEmailService.sendUserInvitationEmail(user, tempPassword);
 			updateSubscriptionQuantity(1L, true, false);
 		}
+		else if (wasDeactivatedGuest) {
+			updateSubscriptionQuantity(1L, true, false);
+		}
+
 		invalidateUserCache();
 		invalidateUserAuthPicCache();
 
@@ -262,10 +268,12 @@ public class PeopleServiceImpl implements PeopleService {
 		CreateEmployeeRequestDto createEmployeeRequestDto = createEmployeeRequest(employeeQuickAddDto);
 
 		Optional<User> existingGuestUser = findGuestUserByEmail(createEmployeeRequestDto);
+		boolean wasDeactivatedGuest = false;
 		if (existingGuestUser.isPresent()) {
 			user = existingGuestUser.get();
 			user.setIsActive(true);
 			employee = user.getEmployee();
+			wasDeactivatedGuest = employee.getAccountStatus() == AccountStatus.DEACTIVATED;
 			// Reset to PENDING so createUserEntity triggers credential setup and
 			// invitation email
 			employee.setAccountStatus(AccountStatus.PENDING);
@@ -286,6 +294,9 @@ public class PeopleServiceImpl implements PeopleService {
 			// Skip for guest conversion: email is sent via createUserEntity ->
 			// resendInvitationEmail
 			peopleEmailService.sendUserInvitationEmail(user, tempPassword);
+			updateSubscriptionQuantity(1L, true, false);
+		}
+		else if (wasDeactivatedGuest) {
 			updateSubscriptionQuantity(1L, true, false);
 		}
 		invalidateUserCache();


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR correctly fixes subscription under-counting when a deactivated guest user is converted to a full employee, by capturing the pre-reset account status and re-incrementing the subscription only for the deactivated case. This mirrors the decrement that happens during guest deactivation (updateUserStatus → updateSubscriptionQuantity(1L,false,false)), so the logic is sound. Remaining comments are non-blocking: duplicated logic across the two entry points, a transactional-boundary asymmetry, and missing test coverage for the new branch.

Found **5** new issue(s): 🟡 1 suggestion(s)

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
```diff
@@ -243,6 +245,10 @@ public ResponseEntityDto createEmployee(CreateEmployeeRequestDto requestDto) {
 			peopleEmailService.sendUserInvitationEmail(user, tempPassword);
 			updateSubscriptionQuantity(1L, true, false);
 		}
+		else if (wasDeactivatedGuest) {
```
ThinuwanW, line 248: 🤖 **Claude** · 💬 **Note**

The guest-conversion detection (`wasDeactivatedGuest`) plus the new `else if (wasDeactivatedGuest) { updateSubscriptionQuantity(1L, true, false); }` branch is now duplicated verbatim in both createEmployee (lines 217/222/248-250) and quickAddEmployee (lines 271/276/299-301). Per checklist item #9, consider extracting the shared guest-conversion handling (locate existing guest user, reactivate, capture deactivated state, reset to PENDING, and post-save subscription increment) into a private helper so the two paths cannot drift apart in future fixes.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
```diff
@@ -214,10 +214,12 @@ public ResponseEntityDto createEmployee(CreateEmployeeRequestDto requestDto) {
 		boolean isGuestConversion = false;
 
 		Optional<User> existingGuestUser = findGuestUserByEmail(requestDto);
+		boolean wasDeactivatedGuest = false;
```
ThinuwanW, line 217: 🤖 **Claude** · 💬 **Note**

📁 **File-level observation**

Transactional-boundary asymmetry: createEmployee is annotated @Transactional (line 206) but quickAddEmployee (line 258) is not, and the class has no class-level @Transactional. Both methods perform writes (userDao.save, event publishing) and now both invoke updateSubscriptionQuantity. In the community build updateSubscriptionQuantity is a no-op, but the enterprise override performs real subscription work; without a transaction on quickAddEmployee, a failure after userDao.save could leave the user persisted while the subscription update is skipped/partially applied. Consider adding @Transactional to quickAddEmployee for consistency with createEmployee.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
```diff
@@ -214,10 +214,12 @@ public ResponseEntityDto createEmployee(CreateEmployeeRequestDto requestDto) {
...
 		if (existingGuestUser.isPresent()) {
 			user = existingGuestUser.get();
 			user.setIsActive(true);
 			employee = user.getEmployee();
+			wasDeactivatedGuest = employee.getAccountStatus() == AccountStatus.DEACTIVATED;
```
ThinuwanW, line 222: 🤖 **Claude** · 💬 **Note**

The fix assumes any non-deactivated guest already counts toward the subscription, so it only re-increments for DEACTIVATED guests. This holds for guests created via the approval path (EpGuestUserServiceImpl line 421 increments) and reactivation, but guests invited via the privileged bulk path (inviteSingleGuestUser → createAndSaveEmployee) are saved as PENDING without ever incrementing the subscription. Converting such a PENDING guest would therefore not increment here either, leaving it uncounted. Worth confirming whether privileged-path PENDING guests can reach this conversion flow; if so, the increment condition may need to cover them too.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
ThinuwanW, line 248: 🤖 **Claude** · 💬 **Note**

No test coverage exists for createEmployee/quickAddEmployee guest-conversion (no tests reference these methods). Since this PR changes subscription side-effects, consider adding tests asserting updateSubscriptionQuantity(1L, true, false) is invoked exactly once when converting a DEACTIVATED guest and is NOT invoked when converting an ACTIVE/PENDING guest, to lock in the intended behavior.

ThinuwanW, line 248: 🤖 **Claude** · 🟡 **Suggestion**

Code duplication (guidelines checklist #9): this PR adds the identical `boolean wasDeactivatedGuest` detection (lines 217/222 and 271/276) and the identical `else if (wasDeactivatedGuest) { updateSubscriptionQuantity(1L, true, false); }` block (lines 248-250 and 299-301) to both createEmployee and quickAddEmployee, which were already near-identical. Two copies of subscription-billing logic will drift over time (e.g., a future fix applied to only one path would silently over/under-count billing). Recommend extracting the guest-conversion handling into a shared private helper — e.g. a method that takes the existing guest User/Employee and returns whether a re-increment is needed, and/or a helper like `incrementSubscriptionIfNeeded(isGuestConversion, wasDeactivatedGuest)` that both methods call. As a smaller alternative, collapse the two branches into one condition: `if (!isGuestConversion || wasDeactivatedGuest) { updateSubscriptionQuantity(1L, true, false); }` with the invitation email kept under the separate `if (!isGuestConversion)` guard, so the billing rule lives in exactly one place per method.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
akilarootcode: its existing method

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
akilarootcode: raise another pr foe that fix also

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
akilarootcode: no need tests

akilarootcode: no need to address duplicate issue

akilarootcode: will address duplication on another pr.

yuhanga-rtc (approved):

shakila-kamalasena (approved):

ThinuwanW (approved):

PR #1530, opened 2026-06-23
Fix/2327: Main content loader content gets split when there a success/error/warning toast at some instances

hasalarootcode:
…ror/warning toast at some instances

### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/frontend/src/community/common/components/molecules/FullScreenLoader/FullScreenLoader.tsx
+++ b/frontend/src/community/common/components/molecules/FullScreenLoader/FullScreenLoader.tsx
@@ -5,9 +5,10 @@ import { useTranslator } from "~community/common/hooks/useTranslator";
 
 interface Props {
   fullPage?: boolean;
+  zIndex?: number;
 }
 
-const FullScreenLoader = ({ fullPage = true }: Props) => {
+const FullScreenLoader = ({ fullPage = true, zIndex = ZIndexEnums.MAX }: Props) => {
   const theme = useTheme();
   const translateAria = useTranslator(
     "commonAria",
@@ -24,7 +25,7 @@ const FullScreenLoader = ({ fullPage = true }: Props) => {
         alignItems: "center",
         justifyContent: "center",
         backgroundColor: theme.palette.background.default,
-        zIndex: ZIndexEnums.MAX
+        zIndex: zIndex
       }}
       role="status"
       aria-live="polite"
--- a/frontend/src/community/common/components/organisms/ContentWithDrawer/ContentWithDrawer.tsx
+++ b/frontend/src/community/common/components/organisms/ContentWithDrawer/ContentWithDrawer.tsx
@@ -6,6 +6,7 @@ import FullScreenLoader from "~community/common/components/molecules/FullScreenL
 import ToastMessage from "~community/common/components/molecules/ToastMessage/ToastMessage";
 import AppBar from "~community/common/components/organisms/AppBar/AppBar";
 import Drawer from "~community/common/components/organisms/Drawer/Drawer";
+import { ZIndexEnums } from "~community/common/enums/CommonEnums";
 import useRouteLoading from "~community/common/hooks/useRouteLoading";
 import { useTranslator } from "~community/common/hooks/useTranslator";
 import {
@@ -46,7 +47,7 @@ const ContentWithDrawer = ({ children }: Props) => {
                 minHeight: "100dvh"
               }}
             >
-              <FullScreenLoader fullPage={false} />
+              <FullScreenLoader fullPage={false} zIndex={ZIndexEnums.MODAL} />
             </div>
           ) : (
             <main


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds an optional `zIndex` prop to `FullScreenLoader` (defaulting to `ZIndexEnums.MAX` to preserve existing behavior) and uses `ZIndexEnums.MODAL` in `ContentWithDrawer` so the in-content loader sits below the toast layer (TOAST=1450), fixing the toast/content overlap. The default-preserving change is safe — all other callers omit the prop and keep MAX. The main gap is that the sibling layout `ContentWithoutDrawer` has the identical loader+toast pattern but was not updated, so the same bug persists there.

Found **4** new issue(s): 🟡 2 suggestion(s), 🟣 2 nit(s)

frontend/src/community/common/components/organisms/ContentWithDrawer/ContentWithDrawer.tsx
```diff
@@ -46,7 +47,7 @@ const ContentWithDrawer = ({ children }: Props) => {
                 minHeight: "100dvh"
               }}
             >
-              <FullScreenLoader fullPage={false} />
+              <FullScreenLoader fullPage={false} zIndex={ZIndexEnums.MODAL} />
```
ThinuwanW, line 50: 🤖 **Claude** · 🟡 **Suggestion**

The sibling layout `ContentWithoutDrawer.tsx` (line 27: `{loading && <FullScreenLoader />}`) has the exact same loader-over-toast pattern this PR is fixing, but still uses the default `zIndex = MAX (2000)`, which is above `TOAST (1450)`. The same toast/content-overlap bug this PR fixes will still occur on unprotected pages (e.g. login error toasts). Consider applying the same fix there: `<FullScreenLoader zIndex={ZIndexEnums.MODAL} />` (note it defaults to `fullPage={true}` there, so verify the overlap still resolves with `position: fixed`).

frontend/src/community/common/components/molecules/FullScreenLoader/FullScreenLoader.tsx
```diff
@@ -5,9 +5,10 @@ import { useTranslator } from "~community/common/hooks/useTranslator";
 
 interface Props {
   fullPage?: boolean;
+  zIndex?: number;
```
ThinuwanW, line 8: 🤖 **Claude** · 🟣 **Nit**

The new `zIndex?: number` prop is typed as a plain `number`, but every call site passes a `ZIndexEnums` member and the default is `ZIndexEnums.MAX`. Typing it as `zIndex?: ZIndexEnums` would keep call sites constrained to the established z-index token scale and avoid accidental hardcoded magic numbers. Minor — `number` works since the enum is numeric.

frontend/src/community/common/components/organisms/ContentWithDrawer/ContentWithDrawer.tsx
ThinuwanW, line 50: 🤖 **Claude** · 🟡 **Suggestion**

Semantic misuse of `ZIndexEnums.MODAL` (1300) as a loader z-index. MODAL is meant for modal dialogs, and the value was clearly chosen only because it happens to fall below TOAST (1450). This creates two latent fragilities: (1) the root-level controllers still rendered during loading (QuickSetupModalController/MyRequestModalController/TimeWidgetPopupController) or any open MUI Modal also sit at 1300, so they would tie with the loader rather than reliably stack above/below it; (2) if the enum is ever reordered (the enum file's own comment says these values are slated to be replaced), the loader-vs-toast relationship silently breaks. Recommend adding a dedicated semantic constant (e.g. `CONTENT_LOADER`) in ZIndexEnums that is explicitly defined to be below TOAST, and using that here instead of borrowing MODAL.

frontend/src/community/common/components/molecules/FullScreenLoader/FullScreenLoader.tsx
ThinuwanW, line 8: 🤖 **Claude** · 🟣 **Nit**

The new public `zIndex` prop adds a configurable behavior to a shared molecule but no `FullScreenLoader.test.tsx` exists (confirmed none in the repo). Per the testing guideline, consider adding a test (wrapped in MockTheme) asserting the loader renders with the default `ZIndexEnums.MAX` and honors an overridden value, so future enum/refactor changes don't silently regress the toast-stacking fix this PR introduces.

frontend/src/community/common/components/organisms/ContentWithDrawer/ContentWithDrawer.tsx
hasalarootcode: That's the behaviour needed

frontend/src/community/common/components/molecules/FullScreenLoader/FullScreenLoader.tsx
hasalarootcode: Components should keep the general behaviour

frontend/src/community/common/components/organisms/ContentWithDrawer/ContentWithDrawer.tsx
hasalarootcode: This fires when navigating to page, because of that this comment is not valid

frontend/src/community/common/components/molecules/FullScreenLoader/FullScreenLoader.tsx
hasalarootcode: Don't need a test

shakila-kamalasena (approved):

ThinuwanW (approved):

PR #1542, opened 2026-06-25
feat: enable Bucket4j + Redis rate limiting for enterprise endpoints

akilarootcode:
## Summary
Enables IP-based rate limiting across all enterprise endpoints using Bucket4j backed by the existing Lettuce/Redis connection. This is the parent PR that ties together the submodule changes.

## Changes (parent repo)
- Add Bucket4j `bucket4j_jdk17-core` + `bucket4j_jdk17-lettuce` 8.19.0 dependencies.
- Fix `handleTooManyRequestException` parameter type (`ServletException` → `TooManyRequestsException`) so the 429 handler is actually matched.
- Update `too-many-requests-exception` message to a user-facing string.

## Submodule PRs (merge first; pointers updated separately)
- src: rootcodelabs/skapp-ep-be-src#696
- config: rootcodelabs/skapp-ep-be-resources-config#121
- tests: rootcodelabs/skapp-ep-be-tests#24

## Behaviour notes
- Per-IP token bucket; sets `X-RateLimit-*` headers, returns `429` when exceeded.
- Fail-closed on Redis unavailability.
- Defaults: capacity 100, refill 100 / 60s.


ThinuwanW (commented):
## 🤖 Claude Code Review

This is a small, clean enablement PR for Bucket4j-based rate limiting. The core change correctly fixes a real bug in GlobalExceptionHandler where the @ExceptionHandler(TooManyRequestsException.class) method declared a ServletException parameter — since TooManyRequestsException extends RuntimeException (not ServletException), the handler would have failed to bind and the exception would have fallen through to the generic handler, returning 500 instead of 429. The dependency additions and message text update are correct and consistent with existing conventions. No blocking issues found.

Found **1** new issue(s):

backend/pom.xml
```diff
@@ -229,6 +229,18 @@
             <artifactId>caffeine</artifactId>
         </dependency>
 
+        <!-- Bucket4j Rate Limiting -->
```
ThinuwanW, line 232: 🤖 **Claude** · 💬 **Note**

The two Bucket4j artifacts (bucket4j_jdk17-core and bucket4j_jdk17-lettuce) both hardcode version 8.19.0 in separate <version> tags. These must always move together — bumping one and forgetting the other yields a classpath with mismatched Bucket4j modules. Extract a shared <bucket4j.version>8.19.0</bucket4j.version> property in <properties> and reference it as ${bucket4j.version} in both dependencies so the versions cannot drift. Nit only, since the project already hardcodes versions for other non-BOM deps.

akilarootcode: Considered, but we reverted to hardcoded versions to stay consistent with the other non-BOM dependencies in this pom, per maintainer preference.

yuhanga-rtc (approved):

shakila-kamalasena (approved):

ThinuwanW (approved):

ThinuwanW (approved):

PR #1543, opened 2026-06-25
fix:  fix the logic of the createTask and do some changes in response…

dulsara-manakal:
… Dto

### Summary

- Change the task creation logic 
- change the delay response

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR reworks CrmTaskServiceImpl.createTask to derive company/contact from the contact and deal entities, and changes CrmDealResponseDto to expose a nested stage object plus description. The createTask rewrite introduces behavioral regressions: the request's companyId is now silently ignored, cross-entity consistency validations are dropped or made tautological, and several existing integration tests (which still expect Bad Request) are not updated and will fail. The CrmDealResponseDto shape change is also a breaking API contract change for frontend consumers.

Found **7** new issue(s):

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmTaskServiceImpl.java
```diff
@@ -156,24 +156,22 @@ public ResponseEntityDto createTask(CrmTaskCreateRequestDto requestDto) {
...
 				.orElseThrow(() -> new ModuleException(CrmMessageConstant.CRM_ERROR_CONTACT_NOT_FOUND));
 			task.setContact(contact);
-		}
-
-		if (requestDto.getCompanyId() != null) {
-			company = crmCompanyDao.findByIdAndIsDeletedFalse(requestDto.getCompanyId())
-				.orElseThrow(() -> new ModuleException(CrmMessageConstant.CRM_ERROR_COMPANY_NOT_FOUND));
-			task.setCompany(company);
+			task.setCompany(contact.getCompany());
+			company = contact.getCompany();
```
ThinuwanW, line 185: 🤖 **Claude** · 💬 **Note**

Regression: the request's companyId is no longer read anywhere. validateTaskTargets(contactId, companyId, dealId) still treats companyId as a valid standalone target, so a request with only companyId passes validation but the task is persisted with NO company (task.setCompany is only reached via the contactId or dealId branches). Two problems: (1) a company-only task silently loses its association, and (2) a non-existent/deleted companyId is no longer validated (the old crmCompanyDao.findByIdAndIsDeletedFalse(...).orElseThrow(CRM_ERROR_COMPANY_NOT_FOUND) is gone). Also breaks existing tests createTask_ContactCompanyMismatch_ReturnsBadRequest, the contact-no-company test, and createTask_DealCompanyMismatch_ReturnsBadRequest in CrmTaskControllerIntegrationTest, which assert Bad Request. Either keep handling companyId (load + validate + setCompany) or, if company is intentionally always derived, remove companyId from CrmTaskCreateRequestDto and from validateTaskTargets and update the tests.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmTaskServiceImpl.java
```diff
@@ -156,24 +156,22 @@ public ResponseEntityDto createTask(CrmTaskCreateRequestDto requestDto) {
...
 		}
 
 		if (requestDto.getDealId() != null) {
 			deal = crmDealDao.findByIdAndIsDeletedFalse(requestDto.getDealId())
 				.orElseThrow(() -> new ModuleException(CrmMessageConstant.CRM_ERROR_DEAL_NOT_FOUND));
 			task.setDeal(deal);
+			task.setContact(deal.getContact());
```

Behavioral regression: when both contactId and dealId are supplied, task.setContact(deal.getContact()) and contact = deal.getContact() silently overwrite the user-supplied contact instead of validating consistency. Previously validateDealBelongsToContact(deal, <user-contact>) would throw CRM_ERROR_TASK_DEAL_CONTACT_MISMATCH when the deal's contact differed from the requested contact. Now a mismatched contactId is ignored rather than rejected. This breaks createTask_DealContactMismatch_ReturnsBadRequest (not updated in this PR). Validate the deal against the originally-requested contact/company before overwriting, or intentionally drop the contactId input and update the test.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmTaskServiceImpl.java
```diff
@@ -156,24 +156,22 @@ public ResponseEntityDto createTask(CrmTaskCreateRequestDto requestDto) {
...
+			task.setCompany(deal.getCompany());
+			contact = deal.getContact();
+			company = deal.getCompany();
+			CrmValidations.validateDealBelongsToContact(deal, contact);
```
ThinuwanW, line 193: 🤖 **Claude** · 💬 **Note**

Dead/tautological validation: validateDealBelongsToContact(deal, contact) and validateDealBelongsToContact... validateDealBelongsToCompany(deal, company) are called immediately after contact = deal.getContact() and company = deal.getCompany(). Since both arguments are derived from the same deal, the equality checks (deal.getContact().getId().equals(contact.getId()), etc.) can never fail. These calls are no-ops and should be removed, or the validations should run against the user-supplied contactId/companyId (see the prior two findings) so they actually catch mismatches.

backend/src/main/java/com/skapp/community/crmplanner/payload/response/CrmDealResponseDto.java
```diff
@@ -11,9 +11,9 @@ public class CrmDealResponseDto {
 
 	private String name;
 
-	private String stageName;
+	private String description;
 
-	private String stageColor;
+	private CrmDealStageResponseDto stage;
```
ThinuwanW, line 17: 🤖 **Claude** · 💬 **Note**

Breaking API contract change: stageName and stageColor flat fields were removed and replaced with a nested CrmDealStageResponseDto stage object (and a new description field). This DTO is returned by the deals list/create/update endpoints (CrmDealServiceImpl). Frontend consumers still read stageName/stageColor (frontend CrmDealListItem in CommonTypes.ts and DealsTable.tsx), so the deals table will lose stage name/color rendering until the frontend is updated. Confirm a coordinated frontend change ships with this, or keep the flat fields for backward compatibility.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmTaskServiceImpl.java
```diff
@@ -156,24 +156,22 @@ public ResponseEntityDto createTask(CrmTaskCreateRequestDto requestDto) {
...
+			CrmValidations.validateDealBelongsToCompany(deal, company);
```
ThinuwanW, line 194: 🤖 **Claude** · 💬 **Note**

Missing paired test update (CI will fail). The refactored createTask no longer reads companyId, auto-derives company from contact/deal, and dropped validateContactBelongsToCompany, so several existing integration tests in CrmTaskControllerIntegrationTest now assert error responses the new code never produces: createTask_ContactCompanyMismatch_ReturnsBadRequest (expects CRM_ERROR_TASK_CONTACT_COMPANY_MISMATCH), createTask_ContactNoCompanyButCompanySpecified_ReturnsBadRequest, createTask_DealContactMismatch_ReturnsBadRequest (expects CRM_ERROR_TASK_DEAL_CONTACT_MISMATCH), createTask_DealCompanyMismatch_ReturnsBadRequest, and createTask_DealNoCompanyButCompanySpecified_ReturnsBadRequest. With the new logic these inputs all return 201 Created instead of 400, so the assertions fail. Either restore the validation behavior or update/remove these tests (and add coverage for the new auto-derive behavior) within this PR so the suite stays green.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmTaskServiceImpl.java
```diff
@@ -156,24 +156,22 @@ public ResponseEntityDto createTask(CrmTaskCreateRequestDto requestDto) {
 			contact = crmContactDao.findByIdAndIsDeletedFalse(requestDto.getContactId())
 				.orElseThrow(() -> new ModuleException(CrmMessageConstant.CRM_ERROR_CONTACT_NOT_FOUND));
 			task.setContact(contact);
-		}
-
-		if (requestDto.getCompanyId() != null) {
-			company = crmCompanyDao.findByIdAndIsDeletedFalse(requestDto.getCompanyId())
-				.orElseThrow(() -> new ModuleException(CrmMessageConstant.CRM_ERROR_COMPANY_NOT_FOUND));
-			task.setCompany(company);
+			task.setCompany(contact.getCompany());
```

Behavioral asymmetry between createTask and editTask for the same operation. createTask now ignores the request's companyId, auto-derives company from contact/deal, and skips validateContactBelongsToCompany, while editTask (lines 232-252) still reads contactId/companyId/dealId independently and runs validateContactBelongsToCompany / validateDealBelongsToContact / validateDealBelongsToCompany against the user-supplied values. The result: a contact-from-company-A + company-B payload is rejected on edit but silently accepted on create. Align the two paths so create and update apply the same target resolution and validation rules; otherwise document why they intentionally differ.

backend/src/main/java/com/skapp/community/crmplanner/payload/response/CrmDealResponseDto.java
ThinuwanW, line 17: 🤖 **Claude** · 💬 **Note**

Missing paired frontend change for the response-shape change. This PR is backend-only, but frontend consumers still read the removed flat fields: DealsTable.tsx:179-181 uses STAGE_COLOR_MAP[deal.stageColor] and renders deal.stageName, and CommonTypes.ts:267-268 still declares stageName/stageColor. After this change the API returns a nested stage object instead, so stageColor/stageName become undefined and the Deals table renders no color and a blank stage label. Update the frontend type and DealsTable to read deal.stage.color / deal.stage.name in the same change set (or ship them together).

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmTaskServiceImpl.java
dulsara-manakal: Remove the company ID because there can be tasks without a company ID.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmTaskServiceImpl.java
dulsara-manakal: Now only set companyId from contact only.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmTaskServiceImpl.java
dulsara-manakal: Fix this and remove the validation since the front end doesn't support this.

backend/src/main/java/com/skapp/community/crmplanner/payload/response/CrmDealResponseDto.java
dulsara-manakal: This actually fixes with this FE [PR.](https://pr.skapp.dev/view/196da4a0-141f-4ce2-9bd3-7d3809d450a8)

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmTaskServiceImpl.java
dulsara-manakal: Fixed them.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmTaskServiceImpl.java
dulsara-manakal: Fixed the editTask as well.

backend/src/main/java/com/skapp/community/crmplanner/payload/response/CrmDealResponseDto.java
dulsara-manakal: This actually fixes in this FE [PR.](https://pr.skapp.dev/view/196da4a0-141f-4ce2-9bd3-7d3809d450a8)

sanjula-rtc (dismissed):

ThinuwanW (commented):
## 🤖 Claude Code Review

This PR removes the independent `companyId` task target (company is now derived from the contact) and adds new filters: deals by `contactId` and contacts by `dealId`, plus reshapes `CrmDealResponseDto` to expose a nested `stage` object and `description`. The mapper, repository, and DTO changes are correct and consistent. The main concern is an asymmetry introduced in `CrmTaskServiceImpl.editTask`, which no longer validates that an attached deal belongs to the task's contact, while `createTask` still does — allowing inconsistent task data on edit.

Found **6** new issue(s): 🔴 1 important, 🟡 2 suggestion(s), 🟣 1 nit(s)

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmTaskServiceImpl.java
```diff
@@ -250,8 +238,6 @@ public ResponseEntityDto editTask(Long id, CrmTaskEditRequestDto requestDto) {
 		}
 
 		CrmValidations.validateContactBelongsToCompany(task.getContact(), task.getCompany());
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

editTask dropped both validateDealBelongsToContact and validateDealBelongsToCompany, but createTask (lines 166-167) still calls validateDealBelongsToContact. This is an asymmetry that allows inconsistent data: e.g. PATCH with only dealId attaches a deal belonging to a different contact than the task's existing contact, or PATCH with only contactId switches the contact while leaving a deal that belongs to the old contact — neither is now caught. The remaining validateContactBelongsToCompany call here is also effectively dead, since company is always derived from contact.getCompany() (line 231), so it can never mismatch. Restore the deal-contact consistency check in editTask (using task.getDeal()/task.getContact()) to match createTask, and add an editTask integration test for the deal/contact mismatch case (there is currently none).

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmTaskServiceImpl.java
```diff
@@ -128,8 +128,7 @@ public ResponseEntityDto createTask(CrmTaskCreateRequestDto requestDto) {
 
 		CrmValidations.validateTaskName(requestDto.getName());
 		CrmValidations.validateTaskTypeId(requestDto.getTypeId());
-		CrmValidations.validateTaskTargets(requestDto.getContactId(), requestDto.getCompanyId(),
-				requestDto.getDealId());
+		CrmValidations.validateTaskTargets(requestDto.getContactId(), requestDto.getDealId());
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

After removing the companyId lookup from createTask/editTask, the injected `crmCompanyDao` field (line 54) is no longer used anywhere in this class — remove the unused dependency. Relatedly, CrmTaskEditRequestDto still declares `companyId` (line 29) which editTask no longer reads, so clients passing it get it silently ignored. Remove that field too, matching the removal already done in CrmTaskCreateRequestDto, to keep the two request DTOs consistent.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
```diff
@@ -151,14 +151,23 @@ private List<Predicate> buildLookupPredicates(CriteriaBuilder cb, Root<CrmContac
 			predicates.add(cb.like(cb.lower(contact.get(CrmContact_.name)), "%" + escaped + "%", '\\'));
 		}
 
+		if (filterDto.getDealId() != null) {
```
ThinuwanW, line 157: 🤖 **Claude** · 🟡 **Suggestion**

The new contact-lookup filter by dealId (and the corresponding CrmContactFilterDto.dealId field) has no integration test, unlike the symmetric deal-by-contactId filter which got CrmDealControllerIntegrationTest.getDeals_FilterByContactId. Add a findContactsForLookup test covering the dealId filter to verify the subquery returns only the contact linked to that deal (and that the count query stays consistent, since buildLookupPredicates is reused for both).

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmTaskServiceImpl.java
```diff
@@ -235,12 +228,7 @@ public ResponseEntityDto editTask(Long id, CrmTaskEditRequestDto requestDto) {
 			CrmContact contact = crmContactDao.findByIdAndIsDeletedFalse(requestDto.getContactId())
 				.orElseThrow(() -> new ModuleException(CrmMessageConstant.CRM_ERROR_CONTACT_NOT_FOUND));
 			task.setContact(contact);
-		}
-
-		if (requestDto.getCompanyId() != null) {
-			CrmCompany company = crmCompanyDao.findByIdAndIsDeletedFalse(requestDto.getCompanyId())
-				.orElseThrow(() -> new ModuleException(CrmMessageConstant.CRM_ERROR_COMPANY_NOT_FOUND));
-			task.setCompany(company);
+			task.setCompany(contact.getCompany());
```
ThinuwanW, line 258: 🤖 **Claude** · 💬 **Note**

Asymmetric / orphaned field: this PR removed `companyId` from CrmTaskCreateRequestDto and removed the company-lookup block from editTask, but CrmTaskEditRequestDto still declares `private Long companyId` (CrmTaskEditRequestDto.java:29) and `requestDto.getCompanyId()` is no longer read anywhere. A client can still POST companyId to the edit endpoint and it will be silently ignored, which is a confusing contract. Remove `companyId` from CrmTaskEditRequestDto for symmetry with the create DTO.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmTaskServiceImpl.java

Dead validation in editTask. `validateContactBelongsToCompany(task.getContact(), task.getCompany())` can no longer fail: whenever a contact is set, company is assigned directly from `contact.getCompany()` (line 231), and when contactId is absent both fields are unchanged and already consistent. The check is now redundant and can be removed to avoid implying a constraint that no longer exists (and to match createTask, which dropped it).

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmTaskServiceImpl.java
```diff
@@ -156,24 +155,18 @@ public ResponseEntityDto createTask(CrmTaskCreateRequestDto requestDto) {
...
-				.orElseThrow(() -> new ModuleException(CrmMessageConstant.CRM_ERROR_COMPANY_NOT_FOUND));
+			company = contact.getCompany();
 			task.setCompany(company);
 		}
 
 		if (requestDto.getDealId() != null) {
 			deal = crmDealDao.findByIdAndIsDeletedFalse(requestDto.getDealId())
 				.orElseThrow(() -> new ModuleException(CrmMessageConstant.CRM_ERROR_DEAL_NOT_FOUND));
 			task.setDeal(deal);
+			CrmValidations.validateDealBelongsToCompany(deal, company);
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Validation order changed in createTask: it now runs validateDealBelongsToCompany before validateDealBelongsToContact, whereas the original order ran the contact check first. For a deal that mismatches both the contact and the company, the error returned to the client changes from CRM_ERROR_TASK_DEAL_CONTACT_MISMATCH to CRM_ERROR_TASK_DEAL_COMPANY_MISMATCH. Minor, but if the previous error precedence was intentional, swap the two calls back.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmTaskServiceImpl.java
dulsara-manakal: Acknowledge and fix this.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmTaskServiceImpl.java
dulsara-manakal: removed `CrmCompanyDao` import and field injection. removed companyId field   from `CrmTaskEditRequestDto `

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
dulsara-manakal: Add test

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmTaskServiceImpl.java
dulsara-manakal: Removed companyId from CrmTaskEditRequestDto

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmTaskServiceImpl.java
dulsara-manakal: Already fixed this.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmTaskServiceImpl.java
dulsara-manakal: Fix the order.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmTaskServiceImpl.java
```diff
@@ -128,8 +125,7 @@ public ResponseEntityDto createTask(CrmTaskCreateRequestDto requestDto) {
 
 		CrmValidations.validateTaskName(requestDto.getName());
 		CrmValidations.validateTaskTypeId(requestDto.getTypeId());
-		CrmValidations.validateTaskTargets(requestDto.getContactId(), requestDto.getCompanyId(),
-				requestDto.getDealId());
+		CrmValidations.validateTaskTargets(requestDto.getContactId(), requestDto.getDealId());
```
sanjula-rtc: remove

backend/src/main/java/com/skapp/community/crmplanner/util/CrmValidations.java
```diff
@@ -202,8 +202,8 @@ public static void validateTaskTypeId(Long typeId) {
 		}
 	}
 
-	public static void validateTaskTargets(Long contactId, Long companyId, Long dealId) {
-		if (contactId == null && companyId == null && dealId == null) {
+	public static void validateTaskTargets(Long contactId, Long dealId) {
```
sanjula-rtc: remove

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
```diff
@@ -140,7 +140,7 @@ public Page<CrmContact> findContactsForLookup(CrmContactFilterDto filterDto, Pag
 		return new PageImpl<>(typedQuery.getResultList(), pageable, getLookupTotalCount(cb, filterDto));
 	}
 
-	private List<Predicate> buildLookupPredicates(CriteriaBuilder cb, Root<CrmContact> contact,
+	private List<Predicate> buildLookupPredicates(CriteriaBuilder cb, CriteriaQuery<?> query, Root<CrmContact> contact,
```
sanjula-rtc: CriteriaQuery<T> query,

sanjula-rtc (changes requested):

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmTaskServiceImpl.java
dulsara-manakal: Removed

backend/src/main/java/com/skapp/community/crmplanner/util/CrmValidations.java
dulsara-manakal: Removed

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
dulsara-manakal: Add 'T' instead of a wildcard.

sanjula-rtc (dismissed):

akilarootcode (dismissed):

akilarootcode (dismissed):

ThinuwanW (approved):

PR #1545, opened 2026-06-25
PII communty fallbacks and encrypting fields phase 1

ThinuwanW:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR introduces a community no-op FieldEncryptionService + JPA converter and annotates several PII columns for field-level encryption (enterprise-backed), following the existing EncryptionDecryptionConverter pattern. The abstraction, no-op default, static-setter injection, and tests are sound and consistent with conventions. The main concern is that one newly-encrypted field (middleName) is used in a LIKE search predicate, which will silently break that search path under enterprise encryption; secondary concerns are column-width capacity for encrypted values and a throwaway generator test committed to main.

Found **5** new issue(s):

backend/src/main/java/com/skapp/community/peopleplanner/model/Employee.java
```diff
@@ -64,6 +66,7 @@ public class Employee extends Auditable<String> {
 	private Gender gender;
 
 	@Column(name = "middle_name")
+	@Convert(converter = FieldEncryptionConverter.class)
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

middleName is now field-encrypted, but it is used in a case-insensitive LIKE search predicate in EmployeeTeamRepositoryImpl.getEmployeesByTeamIds (criteriaBuilder.like on Employee_.middleName, ~line 132). The enterprise KMS implementation uses random-IV AES-GCM, so the stored column is non-deterministic ciphertext (ENC:<alias>:<base64>) and the LIKE will never match plaintext keywords — middle-name search silently returns no rows in enterprise builds (first/last name still match, masking the regression). This directly violates the converter's own Javadoc constraint ('only after verifying that field is not used in WHERE/ORDER BY/unique constraints/joins'). In community builds the no-op keeps it working, so this won't surface in community tests. Fix: either drop middleName from the search predicate, or do not encrypt middleName (it is low-sensitivity compared to ssn/nin/passport), or switch that field to a searchable/deterministic scheme. At minimum, confirm the search behavior is acceptable for enterprise.

backend/src/main/java/com/skapp/community/peopleplanner/model/Employee.java
```diff
@@ -79,6 +82,7 @@ public class Employee extends Auditable<String> {
 	private String addressLine1;
 
 	@Column(name = "personal_email")
+	@Convert(converter = FieldEncryptionConverter.class)
```
ThinuwanW, line 84: 🤖 **Claude** · 💬 **Note**

Column-capacity risk for encrypted values. The enterprise wire format ENC:<alias>:<base64(12-byte IV + ciphertext + 16-byte GCM tag)> is substantially longer than the plaintext (roughly 1.33x + prefix/alias overhead). middle_name, personal_email (varchar(255)) and the EmployeeFamily first_name/last_name/parent_name (varchar(255)) could overflow varchar(255) for long-but-valid inputs, causing silent truncation or insert failures once enterprise encryption is enabled — and truncated ciphertext is unrecoverable. This PR adds the @Convert annotations but ships no Liquibase changeset widening these columns. Please confirm the enterprise migration widens these columns (or document the dependency) before encryption is turned on. ssn/nin/passport_no are short enough to be safe.

backend/src/test/java/com/skapp/community/common/util/converter/LegacyContactNoEncryptionGeneratorTest.java
```diff
@@ -0,0 +1,60 @@
+package com.skapp.community.common.util.converter;
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

This is a self-described throwaway manual generator ('Delete this file once seeding/testing is done') committed to main. It hardcodes contact numbers, prints via System.out.println, and depends on EncryptionDecryptionServiceImpl + an external secret to mint ENC: values for a contact_no revert migration that is otherwise unrelated to this PR's field-encryption work. Committing one-off operational scripts as JUnit tests in the permanent codebase is undesirable — it tends to outlive its purpose and adds noise. Consider keeping it out of version control (a documented one-off script / gist) or, if it must be tracked, place it under a clearly-scoped tooling location rather than src/test and reference the migration it supports. (It is Assumptions-guarded so it won't fail CI — that part is fine.)

backend/src/main/java/com/skapp/community/peopleplanner/model/Employee.java
ThinuwanW: 🤖 **Claude** · 💬 **Note**

ADDITIONAL location the first pass missed for the middle_name searchability problem. Besides EmployeeTeamRepositoryImpl (already reported), LeaveRequestRepositoryImpl.java:1007-1011 (the manager pending-leave-requests search) also builds a predicate `like(lower(employee.get(Employee_.middleName)), pattern)` OR-ed with first/last name. Once middle_name is encrypted with random-IV GCM in enterprise builds, this predicate can never match, so name search there silently loses the middle-name branch (and matches encrypted blobs against the lowercased pattern). Fix in the same way as EmployeeTeamRepositoryImpl: drop middleName from the search OR, or search against a non-encrypted/blind-indexed column. Grep `Employee_.middleName` across repositories to ensure every search/sort/where usage is updated together.

backend/src/main/java/com/skapp/community/common/util/converter/FieldEncryptionConverter.java
```diff
@@ -0,0 +1,54 @@
+package com.skapp.community.common.util.converter;
```
ThinuwanW, line 1: 🤖 **Claude** · 💬 **Note**

📁 **File-level observation**

Code duplication / parallel mechanism (guideline 'Code duplication'). This class is structurally identical to the existing EncryptionDecryptionConverter (both are AttributeConverter<String,String> delegating to an encryption service via a static @Autowired setter), but the two now have divergent semantics: EncryptionDecryptionConverter encrypts even in community builds (via EncryptionDecryptionServiceImpl, as the LegacyContactNoEncryptionGeneratorTest shows) and owns the 'ENC:' prefix in the converter, whereas FieldEncryptionConverter is a no-op in community and pushes prefix/wire-format ownership into the service. Having two field-encryption converters with different community behavior and different prefix ownership is a maintenance trap (a dev applying the 'wrong' one silently gets plaintext-at-rest or a different wire format). Consolidate to one converter, or document explicitly in both classes why both must coexist and which to use for new PII fields.

ThinuwanW: true. but this kept for the purpose until the release. once everything stable, all legacy codes would get deleted

backend/src/main/java/com/skapp/community/peopleplanner/model/Employee.java
ThinuwanW: this was a critical mistake claude has done and missed from my manual review. it hasnt checked the predecated for any search query so this missed. i removed the converter from this and did a through review for all other encrypting fields and all others safe. so this comment can resolve now

backend/src/test/java/com/skapp/community/common/util/converter/LegacyContactNoEncryptionGeneratorTest.java
ThinuwanW: deleted since no tests needed for this temp legacy encryption

backend/src/main/java/com/skapp/community/peopleplanner/model/Employee.java
ThinuwanW: this is true
converted all encrypting column to TEXT from varchar. so no risk now

backend/src/main/java/com/skapp/community/peopleplanner/model/Employee.java
ThinuwanW: true. 
removed the converter for middle name so no issue now
and verified other encrypting fields - no one using these predicated so all good now

akilarootcode (approved):

akilarootcode (approved):

PR #1546, opened 2026-06-26
Add CRM tier limit guard

sanjula-rtc:
## Summary
- Guard CRM add actions for contacts, companies, deals, and tasks
- Add community fallback CRM limit guard/modal implementations
- Update enterprise submodule pointer for CRM limit modal support

## Verification
- git diff --check
- Filtered TypeScript check for touched CRM paths


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds CRM resource-limit guarding to community CRM pages via an enterprise hook (with community fallback stubs), and invalidates a shared limitation query after create mutations. The wiring is consistent and the fallback/enterprise query keys match (`get-crm-limitation`), so invalidation works in both modes. However, a newly added community query-key export is dead code with a mismatched value that could become a silent bug, and the limitation cache is only refreshed on create — not on delete — leaving stale limit state.

Found **7** new issue(s):

frontend/src/community/crm/api/utils/QueryKeys.ts
```diff
@@ -62,3 +62,7 @@ export const taskQueryKeys = {
     searchKeyword
   ]
 };
+
+export const crmLimitationQueryKeys = {
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

This newly added `crmLimitationQueryKeys` export is dead code AND has a value that conflicts with the real key. All four create mutations (CompanyApi, ContactApi, TaskApi, crmDealApi) import `crmLimitationQueryKeys` from `~enterprise/crm/api/utils/QueryKeys`, whose value (and the fallback's) is `["get-crm-limitation"]`. This community copy uses `["crm-limitation"]` and is never imported anywhere. It's a trap: if a future developer imports this one by mistake, `invalidateQueries` will silently no-op against the `["get-crm-limitation"]` query used by `useGetCrmLimitation`. Remove this export.

frontend/src/community/crm/api/CompanyApi.ts
```diff
@@ -15,6 +15,7 @@ import {
 } from "../types/CommonTypes";
 import { companyEndpoints } from "./utils/ApiEndpoints";
 import { companyQueryKeys } from "./utils/QueryKeys";
+import { crmLimitationQueryKeys } from "~enterprise/crm/api/utils/QueryKeys";
```
ThinuwanW, line 18: 🤖 **Claude** · 💬 **Note**

📁 **File-level observation**

Limit-cache invalidation was added to create mutations but not to delete mutations. `useDeleteCompany` (CompanyApi.ts) and `useDeleteTask` (TaskApi.ts) — and any contact/deal delete — reduce a resource's used count and can clear `limitReached`, but they do not invalidate `crmLimitationQueryKeys.GET_CRM_LIMITATION`. As a result, a user who is at the limit and deletes a record will still be blocked by `guardCrmCreate` (which uses the cached `data` when present) until the query goes stale or the page reloads. Add `queryClient.invalidateQueries({ queryKey: crmLimitationQueryKeys.GET_CRM_LIMITATION })` to the delete mutations' `onSuccess` as well.

frontend/pages/community/crm/companies.tsx
```diff
@@ -9,9 +9,13 @@ import CompanySidePanel from "~community/crm/components/organisms/CompanySidePan
 import { CompanyTable } from "~community/crm/components/organisms/CompanyTable/CompanyTable";
 import { useCrmStore } from "~community/crm/store/store";
 import { CrmModalTypes } from "~community/crm/types/ModalTypes";
+import CrmLimitModalController from "~enterprise/crm/components/organisms/CrmLimitModalController/CrmLimitModalController";
+import useCrmLimitGuard from "~enterprise/crm/hooks/useCrmLimitGuard";
+import { CrmLimitResource } from "~enterprise/crm/types/CrmLimitTypes";
 
 const Companies: NextPage = () => {
   const translateText = useTranslator("crmModule", "companies");
+  const { guardCrmCreate } = useCrmLimitGuard();
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

`guardCrmCreate` is async in the enterprise implementation and will `await refetch()` on the first click when no limitation data is cached, but the create button gets no loading/disabled feedback. The hook even returns `isCheckingCrmLimit` for exactly this, yet none of the pages (companies/contacts/deals/tasks) destructure or use it. During the in-flight check the primary button appears inert and is repeatedly clickable, which can fire multiple refetches. Consider wiring `isCheckingCrmLimit` to the primary button's loading/disabled state across these pages.

frontend/pages/community/crm/companies.tsx
```diff
@@ -9,9 +9,13 @@ import CompanySidePanel from "~community/crm/components/organisms/CompanySidePan
 import { CompanyTable } from "~community/crm/components/organisms/CompanyTable/CompanyTable";
 import { useCrmStore } from "~community/crm/store/store";
 import { CrmModalTypes } from "~community/crm/types/ModalTypes";
+import CrmLimitModalController from "~enterprise/crm/components/organisms/CrmLimitModalController/CrmLimitModalController";
+import useCrmLimitGuard from "~enterprise/crm/hooks/useCrmLimitGuard";
+import { CrmLimitResource } from "~enterprise/crm/types/CrmLimitTypes";
```
ThinuwanW, line 15: 🤖 **Claude** · 💬 **Note**

Both useCrmLimitGuard implementations return `isCheckingCrmLimit` (enterprise: `isFetching`; fallback: `false`), but no consumer destructures or uses it. In enterprise mode the first click triggers an awaited `refetch()` (see enterprise useCrmLimitGuard), during which the create button gives no feedback and stays enabled — rapid double-clicks fire multiple refetches and can run `onAllowed` twice (opening the modal/side panel twice). This applies to all four pages and SidePanelTasksSection. Fix: destructure `isCheckingCrmLimit` and pass it to the primary/add button as `disabled`/loading state (and the button in SidePanelTasksSection).

frontend/src/fallback/crm/types/CrmLimitTypes.ts
```diff
@@ -0,0 +1,7 @@
+export enum CrmLimitResource {
+  CONTACTS = "contacts",
+  COMPANIES = "companies",
+  DEALS = "deals",
+  DEAL_STAGES = "dealStages",
```
ThinuwanW, line 5: 🤖 **Claude** · 💬 **Note**

DEAL_STAGES is defined as a CrmLimitResource and is fully supported by the upgrade modal (dedicated `dealStageLimitImage` and translation keys in the enterprise modal), yet no code calls `guardCrmCreate(CrmLimitResource.DEAL_STAGES)` and the deal-stage creation flow (configurations module) never invalidates `crmLimitationQueryKeys.GET_CRM_LIMITATION`. This is a missing paired change: creating a deal stage will not refresh the limit cache, and stage creation is never gated. Fix: apply the guard to the deal-stage create action and add the limitation invalidation to the deal-stage create mutation, or confirm DEAL_STAGES is intentionally deferred.

frontend/src/fallback/crm/api/CrmLimitationApi.ts
```diff
@@ -0,0 +1,17 @@
+import type { CrmLimitationResponse } from "~enterprise/crm/types/CrmLimitTypes";
```
ThinuwanW, line 1: 🤖 **Claude** · 💬 **Note**

This fallback stub (and the fallback CrmLimitUpgradeModal.tsx) appears to be dead code in community mode. `~enterprise/crm/api/CrmLimitationApi` and `~enterprise/crm/components/molecules/CrmLimitUpgradeModal` are imported only by enterprise files (useCrmLimitGuard / CrmLimitModalController), which are not compiled in community builds — the community build uses the self-contained fallback hook (imports nothing from CrmLimitationApi) and the fallback CrmLimitModalController that returns null without importing the modal. No community file imports these two paths. If they are only kept to mirror the enterprise tree, that's defensible, but they are never resolved; consider removing them or documenting why the mirror is needed.

frontend/src/community/crm/components/molecules/SidePanelTasksSection/SidePanelTasksSection.tsx
```diff
@@ -42,9 +45,11 @@ const SidePanelTasksSection: FC<Props> = ({
   );
 
   const handleAddTask = () => {
-    setPreselectedContact(preselectedContact);
-    setTaskModalType(CrmModalTypes.ADD_TASK_MODAL);
-    setIsTaskModalOpen(true);
+    guardCrmCreate(CrmLimitResource.TASKS, () => {
```
ThinuwanW, line 48: 🤖 **Claude** · 💬 **Note**

When the TASKS limit is reached, `guardCrmCreate` only sets store state (`setLimitedResource`); the upgrade modal renders only where a `CrmLimitModalController` is mounted. This component does not render one and relies on the parent page doing so. It works today because both consuming pages (contacts/companies via Contact/CompanySidePanel) mount the controller, but the coupling is implicit and fragile — a future page rendering this section without a controller would silently swallow the limit block (no modal, no task created). Consider documenting this requirement or mounting the controller alongside the section.

frontend/src/community/crm/api/utils/QueryKeys.ts
sanjula-rtc: removed

frontend/src/community/crm/api/CompanyApi.ts
sanjula-rtc: added

frontend/src/fallback/crm/types/CrmLimitTypes.ts
sanjula-rtc: added

frontend/src/fallback/crm/api/CrmLimitationApi.ts
sanjula-rtc: removed

frontend/src/community/crm/components/molecules/SidePanelTasksSection/SidePanelTasksSection.tsx
sanjula-rtc: this is the strucuture of our pages. so this is acceptable

frontend/pages/community/crm/companies.tsx
sanjula-rtc: added

frontend/pages/community/crm/companies.tsx
sanjula-rtc: added

frontend/src/community/crm/components/molecules/SidePanelTasksSection/SidePanelTasksSection.tsx
hasalarootcode: lets try to move that modal controller to a common wrapper

hasalarootcode (changes requested):

hasalarootcode (dismissed):

frontend/pages/community/crm/companies.tsx
```diff
@@ -35,17 +38,20 @@ const Companies: NextPage = () => {
...
-    setCompanyModalType(CrmModalTypes.ADD_COMPANY_MODAL);
+    guardCrmCreate(CrmLimitResource.COMPANIES, () => {
+      setIsCompanyModalOpen(true);
+      setCompanyModalType(CrmModalTypes.ADD_COMPANY_MODAL);
+    });
   };
 
   return (
-    <ContentLayout
+    <CrmLayout
```
ThinuwanW: keep both

ThinuwanW (approved):

PR #1554, opened 2026-06-29
fix(auth): keep employees signed in for the full refresh-token lifetime

akilarootcode:
## Problem

Employee users were redirected to `/signin` after ~2-3 days instead of staying signed in for the full 30-day refresh-token lifetime.

**Root cause:** the Next.js `middleware.ts` authorizes protected routes using **only** the client-JS-set `accessToken` cookie and never consults the httpOnly refresh token. Browsers evict script-set (`document.cookie`) cookies well before their nominal `expires` — notably Safari/iOS ITP, which caps them to a few days — while the server-set httpOnly refresh cookie legitimately lives ~30 days. Once the access cookie is gone, the middleware redirects to `/signin` before any client-side refresh can run, even though the refresh token is still valid. This is also why the timing was browser-dependent (~2-3 days) rather than exactly 30.

## Fix (frontend only — backend already issues a correct 30-day token)

- **`middleware.ts`** — if there is no `accessToken` cookie but the durable refresh-token cookie is present, let the request through so the client can silently mint a new access token. Enterprise checks the **exact** tenant-scoped cookie `{tenantId}_refreshToken` (tenant derived from the subdomain via the shared `getSubdomain` util); when the subdomain carries no tenant it falls through to `/signin` (no fallback). Community mode checks `refreshToken`.
- **`authUtils.ts` (`getAccessToken`)** — when the `accessToken` cookie is missing, attempt a refresh from the refresh token instead of returning `null` immediately. Previously it only refreshed when the cookie existed but the JWT was expired, so an evicted cookie meant no refresh attempt at all.

## Behavior after the fix

- Access cookie evicted, refresh token still valid → silent refresh, user stays signed in.
- Refresh token genuinely expired/invalid → existing `authAxios` interceptor signs out (30-day cap preserved; no redirect loop).
- Enterprise host with no tenant in the subdomain → redirect to `/signin`.

## Testing

- **Manual:** sign in, then in DevTools delete only the `accessToken` cookie (keep `{tenant}_refreshToken`) and navigate to `/dashboard` → stays signed in and `POST /v1/auth/session/refresh-token` returns `200`. Delete both cookies → redirected to `/signin`.
- TypeScript check clean on both changed files.

ChatGPT Image Jun 26, 2026, 03_38_44 PM


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR fixes premature employee sign-out by letting the middleware pass requests through when an access token is absent but a refresh-token cookie is present, and by having getAccessToken() attempt a silent refresh instead of returning null. The core logic is sound and the cookie-naming matches the backend (refreshToken for community, {tenant}_refreshToken for enterprise). The main concern is that role-based authorization in the middleware is bypassed for the request that passes through on refresh-token-only, plus a couple of minor edge cases.

Found **7** new issue(s): 🔴 2 important, 🟡 1 suggestion(s)

frontend/middleware.ts
```diff
@@ -335,9 +341,25 @@ export function middleware(request: NextRequest) {
...
   } else {
-    return NextResponse.redirect(new URL(ROUTES.AUTH.SIGNIN, request.url));
+    refreshTokenCookieName = "refreshToken";
   }
+
+  const hasRefreshToken =
+    refreshTokenCookieName !== null &&
+    Boolean(request.cookies.get(refreshTokenCookieName)?.value);
+
+  if (!token && hasRefreshToken) {
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

Authorization gap to verify: when the access token is absent but any non-empty refresh-token cookie exists, the request is allowed through with NextResponse.next() while `roles` is empty (claims came from an empty token). This means none of the role-based gates in this middleware run for that request — a low-privilege user whose access token has just expired can deep-link to a super-admin/admin route (e.g. /setup-organization, /configurations) and the page will render, because the only server-side gate (this middleware) is skipped for that navigation. The refresh-token value is opaque/httpOnly so the middleware cannot validate it either. This is acceptable only if every protected page enforces its own role guard after the client-side refresh resolves. Please confirm page-level guards exist; otherwise consider redirecting through a dedicated refresh route that re-runs the role check once the new access token is set.

frontend/middleware.ts
```diff
@@ -335,9 +341,25 @@ export function middleware(request: NextRequest) {
     return NextResponse.redirect(
       new URL(ROUTES.AUTH.UNAUTHORIZED, request.url)
     );
+  }
+
+  const tenantId = getSubdomain(request.nextUrl.hostname) as string;
+  let refreshTokenCookieName: string | null;
+  if (isEnterpriseMode()) {
+    refreshTokenCookieName = tenantId ? `${tenantId}_refreshToken` : null;
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

Enterprise fallback mismatch with the backend. Here, when `tenantId` is falsy in enterprise mode you set refreshTokenCookieName = null, so hasRefreshToken becomes false and the user is bounced to sign-in. But the backend (EpCookieUtil.getRefreshTokenCookieName) falls back to the bare "refreshToken" name when tenantId is null/empty. To stay consistent with the backend contract, the falsy branch should fall back to "refreshToken" rather than null. Edge case (getSubdomain rarely returns empty), but worth aligning.

frontend/middleware.ts
```diff
@@ -335,9 +341,25 @@ export function middleware(request: NextRequest) {
     return NextResponse.redirect(
       new URL(ROUTES.AUTH.UNAUTHORIZED, request.url)
     );
+  }
+
+  const tenantId = getSubdomain(request.nextUrl.hostname) as string;
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

`getSubdomain(...) as string` uses a type assertion to silence the `string | string[]` union return type. Per the guidelines, prefer fixing the type over asserting — e.g. call getSubdomain with an explicit single-value contract or narrow with a runtime check, so an accidental array return isn't masked. Minor.

frontend/src/community/auth/utils/authUtils.ts
```diff
@@ -170,7 +170,7 @@ export const getAccessToken = async (): Promise<string | null> => {
   const currentAccessToken = getCookieValue("accessToken");
 
   if (!currentAccessToken) {
-    return null;
+    return await getNewAccessToken();
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

getAccessToken() now calls getNewAccessToken() whenever the accessToken cookie is missing — including for genuinely logged-out users on any page that calls this. Since the httpOnly refresh cookie can't be read client-side, this always fires a POST to REFRESH_TOKEN; on a missing/invalid refresh cookie the backend returns COMMON_ERROR_MISSING_COOKIE_IN_TOKEN / COMMON_ERROR_INVALID_REFRESH_TOKEN, which the authInterceptor response handler turns into signOut() (redirect to /signin). Verify this doesn't cause an unnecessary refresh request (and potential redirect loop) when getAccessToken is invoked on the sign-in/public pages for never-authenticated users.

frontend/middleware.ts
```diff
@@ -16,9 +16,15 @@ import {
   SenderTypes,
   SuperAdminType
 } from "~community/common/types/AuthTypes";
-import { checkRestrictedRoutesAndRedirect } from "~community/common/utils/commonUtil";
+import {
```

📁 **File-level observation**

Sign-out robustness regression from the combination of both changes (distinct from the generic allow-through concern). clearCookies() (authUtils.ts) only clears the non-HttpOnly accessToken cookie client-side and relies on the backend SIGNOUT API to revoke/clear the HttpOnly refresh-token cookie — but that API call is wrapped in try/catch and merely logs on failure. If SIGNOUT fails (network error, 5xx), the refresh-token cookie survives. With the new middleware fall-through (lines 354-360), a request with no accessToken but a surviving refresh cookie now returns NextResponse.next() instead of redirecting to sign-in, and getAccessToken() then calls getNewAccessToken() and silently re-authenticates the user — so 'Sign out' can leave the user logged in. Before this PR the user was effectively logged out at the UI even when SIGNOUT failed. Fix: have clearCookies() detect SIGNOUT failure and surface it / block navigation, or guard the middleware fall-through against the just-signed-out state (e.g. a short-lived 'signedOut' marker cookie), so a failed server revocation cannot result in silent re-auth.

frontend/src/community/auth/utils/authUtils.ts
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Beyond the extra network call already noted, this change can create a sign-out/redirect loop for logged-out users — a materially worse consequence than a wasted fetch. requestInterceptorConfig in axiosInterceptor.ts awaits getAccessToken() for EVERY authFetch/authFetchV2 request before the '/app-setup-status' / '/refresh-token' skip check, so even calls to public/no-auth endpoints (e.g. app-setup-status invoked from the sign-in or landing page) now trigger getNewAccessToken(). When the refresh fails for a genuinely logged-out user and the backend returns COMMON_ERROR_INVALID_REFRESH_TOKEN / COMMON_ERROR_MISSING_COOKIE_IN_TOKEN, authInterceptor.ts's response interceptor calls signOut(), which does window.location.href = SIGNIN?callback=... — reloading the sign-in page, which fires the public call again, repeating the refresh→signOut→redirect cycle. Fix: only attempt getNewAccessToken() when a refresh token is actually present, or move the auto-refresh out of the unconditional getAccessToken() path so unauthenticated/public flows don't trigger it.

frontend/middleware.ts

Separate from the `as string` type-assertion concern: tenantId is computed unconditionally on every middleware invocation via the enterprise-only getSubdomain(), but it is only used inside the isEnterpriseMode() branch (line 349). In community mode this is dead computation and unnecessarily couples the shared middleware path to an enterprise module on every request. Move `const tenantId = getSubdomain(request.nextUrl.hostname)` inside the `if (isEnterpriseMode())` block so it isn't evaluated (and the enterprise util isn't exercised) for community-mode requests.

frontend/middleware.ts
akilarootcode: Fixed in 01e43d6a2 (and 7f73185). The tenant is now derived primarily from the readable `tenant` cookie the backend sets alongside the refresh cookie; when it's absent and the host subdomain is falsy we fall back to the bare `"refreshToken"` name, plus an extra `|| request.cookies.get("refreshToken")` check — so it now matches EpCookieUtil.getRefreshTokenCookieName's fallback contract.

frontend/middleware.ts
akilarootcode: Fixed in 01e43d6a2. Replaced the `as string` assertion with a runtime narrowing: `Array.isArray(hostSubdomain) ? hostSubdomain[0] : hostSubdomain`, so an accidental array return is handled rather than masked.

akilarootcode: Restructured in 01e43d6a2. The `isEnterpriseMode()` branch is gone; the tenant is now taken from the `tenant` cookie (used in both community and enterprise), so it's no longer dead computation in community mode. `getSubdomain` remains only as a fallback when the cookie is missing.

frontend/src/community/auth/utils/authUtils.ts
akilarootcode: Addressed in 01e43d6a2. getAccessToken() no longer fires /refresh-token unconditionally — it only calls getNewAccessToken() when the middleware-exposed `hasRefreshToken` flag is "true". With no refresh token it returns null (the middleware redirects protected routes to sign-in), so never-authenticated users on public/sign-in pages don't trigger a doomed refresh.

akilarootcode: Addressed in 01e43d6a2. Because getAccessToken() now returns null instead of calling signOut() when there's no refresh token, the unconditional getAccessToken() in requestInterceptorConfig no longer drives a refresh→signOut→redirect cycle on public endpoints (e.g. /app-setup-status from the sign-in/landing page). As defense-in-depth, signOut() also early-returns when already on /signin or /enterprise/signin.

frontend/middleware.ts
akilarootcode: Addressed in 01e43d6a2. clearCookies() now sets a non-HttpOnly `signedOut` marker; the middleware treats the session as signed out while it's present (hasRefreshToken forced false), so a refresh cookie that survived a failed backend SIGNOUT can no longer silently re-authenticate via the allow-through. The marker is cleared in setAccessToken() on the next sign-in. Residual: it's session-scoped, so the narrow case of failed-SIGNOUT + browser restart + still-valid refresh cookie isn't covered (the HttpOnly cookie can only be revoked server-side); open to making it longer-lived if preferred.

frontend/middleware.ts
akilarootcode: Acknowledged — this allow-through (no access token + valid refresh token → next()) is pre-existing (introduced in 0d48562, not the flag change), retained to let the client silently refresh. Mitigation in place: the data on these routes is fetched via authFetch with a Bearer token and the backend enforces per-endpoint role authorization, so the allow-through only renders the client shell for the brief window until the refresh resolves; the very next request/navigation carries a populated token and is gated normally by this middleware. Leaving the deeper 'refresh route that re-runs role checks' as a follow-up; flagging for confirmation that page-level guards exist on the sensitive routes.

akilarootcode:
Superseded by #1560 (clean branch off develop with the final middleware + client refresh approach).

PR #1555, opened 2026-06-29
feat: new endpoint to get related tasks

YasinduRC:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmTaskController.java
+++ b/backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmTaskController.java
@@ -87,7 +87,7 @@ public class CrmTaskController {
 
 	@Operation(summary = "Get related tasks",
 			description = "Returns paginated tasks filtered by contactId and/or dealId.")
-	@GetMapping("/related-tasks")
+	@GetMapping("/related")
 	@PreAuthorize("hasRole('ROLE_CRM_SALES_REPRESENTATIVE')")
 	public ResponseEntity<ResponseEntityDto> getRelatedTasks(CrmTaskRelatedFilterDto filterDto) {
 		ResponseEntityDto response = taskService.getRelatedTasks(filterDto);
--- a/backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmTaskController.java
+++ b/backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmTaskController.java
@@ -5,6 +5,7 @@ import com.skapp.community.crmplanner.payload.request.CrmTaskCompletedFilterDto;
 import com.skapp.community.crmplanner.payload.request.CrmTaskCreateRequestDto;
 import com.skapp.community.crmplanner.payload.request.CrmTaskEditRequestDto;
 import com.skapp.community.crmplanner.payload.request.CrmTaskFilterDto;
+import com.skapp.community.crmplanner.payload.request.CrmTaskRelatedFilterDto;
 import com.skapp.community.crmplanner.service.CrmTaskService;
 import io.swagger.v3.oas.annotations.Operation;
 import io.swagger.v3.oas.annotations.tags.Tag;
@@ -81,6 +82,15 @@ public class CrmTaskController {
 	@PreAuthorize("hasRole('ROLE_CRM_SALES_REPRESENTATIVE')")
 	public ResponseEntity<ResponseEntityDto> deleteTask(@PathVariable Long id) {
 		ResponseEntityDto response = taskService.deleteTask(id);
+		return new ResponseEntity<>(response, HttpStatus.OK);
+	}
+
+	@Operation(summary = "Get related tasks",
+			description = "Returns paginated tasks filtered by contactId and/or dealId.")
+	@GetMapping("/related-tasks")
+	@PreAuthorize("hasRole('ROLE_CRM_SALES_REPRESENTATIVE')")
+	public ResponseEntity<ResponseEntityDto> getRelatedTasks(CrmTaskRelatedFilterDto filterDto) {
+		ResponseEntityDto response = taskService.getRelatedTasks(filterDto);
 		return new ResponseEntity<>(response, HttpStatus.OK);
 	}
 
--- a/backend/src/main/java/com/skapp/community/crmplanner/payload/request/CrmTaskRelatedFilterDto.java
+++ b/backend/src/main/java/com/skapp/community/crmplanner/payload/request/CrmTaskRelatedFilterDto.java
@@ -1,0 +1,20 @@
+package com.skapp.community.crmplanner.payload.request;
+
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+public class CrmTaskRelatedFilterDto {
+
+	private Long contactId;
+
+	private Long dealId;
+
+	private Long excludeTaskId;
+
+	private int page = 0;
+
+	private int size = 10;
+
+}
--- a/backend/src/main/java/com/skapp/community/crmplanner/payload/request/CrmTaskRelatedFilterDto.java
+++ b/backend/src/main/java/com/skapp/community/crmplanner/payload/request/CrmTaskRelatedFilterDto.java
@@ -11,8 +11,6 @@ public class CrmTaskRelatedFilterDto {
 
 	private Long dealId;
 
-	private Long excludeTaskId;
-
 	private int page = 0;
 
 	private int size = 10;
--- a/backend/src/main/java/com/skapp/community/crmplanner/repository/CrmTaskRepository.java
+++ b/backend/src/main/java/com/skapp/community/crmplanner/repository/CrmTaskRepository.java
@@ -3,6 +3,7 @@ package com.skapp.community.crmplanner.repository;
 import com.skapp.community.crmplanner.model.CrmTask;
 import com.skapp.community.crmplanner.payload.request.CrmTaskCompletedFilterDto;
 import com.skapp.community.crmplanner.payload.request.CrmTaskFilterDto;
+import com.skapp.community.crmplanner.payload.request.CrmTaskRelatedFilterDto;
 import com.skapp.community.crmplanner.type.CrmContactTaskMetrics;
 import com.skapp.community.crmplanner.type.CrmTaskSummary;
 import org.springframework.data.domain.Page;
@@ -27,5 +28,7 @@ public interface CrmTaskRepository {
 	Page<CrmTask> findCompletedTasks(Long ownerId, CrmTaskCompletedFilterDto filterDto, Pageable pageable);
 
 	Map<Long, Long> countTasksByDealIds(List<Long> dealIds);
+
+	Page<CrmTask> findRelatedTasksPaginated(CrmTaskRelatedFilterDto filterDto, Pageable pageable);
 
 }
--- a/backend/src/main/java/com/skapp/community/crmplanner/repository/CrmTaskRepository.java
+++ b/backend/src/main/java/com/skapp/community/crmplanner/repository/CrmTaskRepository.java
@@ -29,6 +29,6 @@ public interface CrmTaskRepository {
 
 	Map<Long, Long> countTasksByDealIds(List<Long> dealIds);
 
-	Page<CrmTask> findRelatedTasksPaginated(CrmTaskRelatedFilterDto filterDto, Pageable pageable);
+	Page<CrmTask> findRelatedTasks(CrmTaskRelatedFilterDto filterDto, Pageable pageable);
 
 }
--- a/backend/src/main/java/com/skapp/community/crmplanner/repository/CrmTaskRepository.java
+++ b/backend/src/main/java/com/skapp/community/crmplanner/repository/CrmTaskRepository.java
@@ -29,6 +29,6 @@ public interface CrmTaskRepository {
 
 	Map<Long, Long> countTasksByDealIds(List<Long> dealIds);
 
-	Page<CrmTask> findRelatedTasks(CrmTaskRelatedFilterDto filterDto, Pageable pageable);
+	Page<CrmTask> findRelatedTasks(CrmTaskRelatedFilterDto filterDto, Long ownerId, Pageable pageable);
 
 }
--- a/backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmTaskRepositoryImpl.java
+++ b/backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmTaskRepositoryImpl.java
@@ -248,14 +248,15 @@ public class CrmTaskRepositoryImpl implements CrmTaskRepository {
 	}
 
 	@Override
-	public Page<CrmTask> findRelatedTasks(CrmTaskRelatedFilterDto filterDto, Pageable pageable) {
+	public Page<CrmTask> findRelatedTasks(CrmTaskRelatedFilterDto filterDto, Long ownerId, Pageable pageable) {
 		CriteriaBuilder cb = entityManager.getCriteriaBuilder();
 		CriteriaQuery<CrmTask> query = cb.createQuery(CrmTask.class);
 		Root<CrmTask> task = query.from(CrmTask.class);
 
 		applyFetchGraph(task);
 
-		CrmTaskRelatedParams params = new CrmTaskRelatedParams(filterDto.getContactId(), filterDto.getDealId());
+		CrmTaskRelatedParams params = new CrmTaskRelatedParams(filterDto.getContactId(), filterDto.getDealId(),
+				ownerId);
 		List<Predicate> predicates = buildRelatedTaskPredicates(cb, task, params);
 
 		query.select(task)
@@ -284,6 +285,10 @@ public class CrmTaskRepositoryImpl implements CrmTaskRepository {
 			CrmTaskRelatedParams params) {
 		List<Predicate> predicates = new ArrayList<>();
 		predicates.add(cb.isFalse(root.get(CrmTask_.isDeleted)));
+
+		if (params.getOwnerId() != null) {
+			predicates.add(cb.equal(root.get(CrmTask_.owner).get(Employee_.employeeId), params.getOwnerId()));
+		}
 
 		List<Predicate> contextPredicates = new ArrayList<>();
 		if (params.getContactId() != null) {
--- a/backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmTaskRepositoryImpl.java
+++ b/backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmTaskRepositoryImpl.java
@@ -248,15 +248,14 @@ public class CrmTaskRepositoryImpl implements CrmTaskRepository {
 	}
 
 	@Override
-	public Page<CrmTask> findRelatedTasksPaginated(CrmTaskRelatedFilterDto filterDto, Pageable pageable) {
+	public Page<CrmTask> findRelatedTasks(CrmTaskRelatedFilterDto filterDto, Pageable pageable) {
 		CriteriaBuilder cb = entityManager.getCriteriaBuilder();
 		CriteriaQuery<CrmTask> query = cb.createQuery(CrmTask.class);
 		Root<CrmTask> task = query.from(CrmTask.class);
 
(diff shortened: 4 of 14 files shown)


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds a paginated 'related tasks' endpoint (controller, service, repository, DTO/params, tests) following the existing CRM task patterns cleanly. The main concern is a missing ORDER BY in the new paginated criteria query, which makes pagination non-deterministic; secondary concerns are an owner-scoping asymmetry versus the other task list endpoints, duplicated overdue logic, and unvalidated pagination/filter inputs.

Found **8** new issue(s): 🟡 1 suggestion(s), 🟣 2 nit(s)

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmTaskRepositoryImpl.java
```diff
@@ -245,6 +247,58 @@ private void applyFetchGraph(Root<CrmTask> root) {
...
+		CriteriaQuery<CrmTask> query = cb.createQuery(CrmTask.class);
+		Root<CrmTask> task = query.from(CrmTask.class);
+
+		applyFetchGraph(task);
+
+		CrmTaskRelatedParams params = new CrmTaskRelatedParams(filterDto.getContactId(), filterDto.getDealId(),
+				filterDto.getExcludeTaskId());
+		List<Predicate> predicates = buildRelatedTaskPredicates(cb, task, params);
+
+		query.select(task).where(predicates.toArray(new Predicate[0]));
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

findRelatedTasksPaginated builds a paginated query (setFirstResult/setMaxResults) but has no ORDER BY clause. The sibling paginated query buildFindCompletedTasksQuery orders by lastModifiedDate desc then id desc, and findTasks orders by dueAt/id. Without a deterministic sort, the database is free to return rows in arbitrary/inconsistent order between the content query and across pages, causing duplicated or skipped items when the client paginates. Add an explicit orderBy (e.g. cb.asc/desc on dueAt and a tiebreaker on CrmTask_.id) to query.select(task) before creating the TypedQuery.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmTaskServiceImpl.java
```diff
@@ -121,6 +124,32 @@ public ResponseEntityDto getCompletedTasks(CrmTaskCompletedFilterDto filterDto)
 		return new ResponseEntityDto(false, response);
 	}
 
+	@Override
+	@Transactional(readOnly = true)
+	public ResponseEntityDto getRelatedTasks(CrmTaskRelatedFilterDto filterDto) {
```
ThinuwanW, line 128: 🤖 **Claude** · 💬 **Note**

getRelatedTasks applies no owner restriction, while getTasks and getCompletedTasks scope results to the current sales rep's ownerId (CrmUtil.isCrmSalesRepresentative) and getTaskById denies cross-owner access via CrmValidations.isOwnerRestrictedForRepresentative. As written, any CRM_SALES_REPRESENTATIVE can enumerate tasks for any contactId/dealId, including tasks owned by other reps. Note getContactById is similarly unscoped, so this may be intentional for contact/deal context views — please confirm the intended visibility and, if reps should only see their own related tasks, pass the resolved ownerId into the filter and predicate as the other list endpoints do.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmTaskServiceImpl.java
```diff
@@ -121,6 +124,32 @@ public ResponseEntityDto getCompletedTasks(CrmTaskCompletedFilterDto filterDto)
...
+		log.info("getRelatedTasks: execution started");
+
+		Pageable pageable = PageRequest.of(filterDto.getPage(), filterDto.getSize());
+		Page<CrmTask> taskPage = crmTaskDao.findRelatedTasksPaginated(filterDto, pageable);
+
+		List<CrmTaskDetailResponseDto> tasks = taskPage.getContent().stream().map(task -> {
+			CrmTaskDetailResponseDto dto = crmMapper.crmTaskToCrmTaskDetailResponseDto(task);
+			dto.setIsOverdue(!Boolean.TRUE.equals(task.getIsCompleted()) && task.getDueAt() != null
```

The isOverdue computation (!Boolean.TRUE.equals(task.getIsCompleted()) && task.getDueAt() != null && dueAt.isBefore(LocalDate.now().atStartOfDay())) is duplicated verbatim from CrmContactServiceImpl#getContactById (around line 328). Extract a shared helper (e.g. a static CrmUtil.isTaskOverdue(CrmTask) or have the mapper populate it) so the overdue rule lives in one place.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmTaskRepositoryImpl.java
```diff
@@ -245,6 +247,58 @@ private void applyFetchGraph(Root<CrmTask> root) {
...
+		predicates.add(cb.isFalse(root.get(CrmTask_.isDeleted)));
+
+		List<Predicate> contextPredicates = new ArrayList<>();
+		if (params.getContactId() != null) {
+			contextPredicates.add(cb.equal(root.get(CrmTask_.contact).get(CrmContact_.id), params.getContactId()));
+		}
+		if (params.getDealId() != null) {
+			contextPredicates.add(cb.equal(root.get(CrmTask_.deal).get(CrmDeal_.id), params.getDealId()));
+		}
+		predicates.add(cb.or(contextPredicates.toArray(new Predicate[0])));
```
ThinuwanW, line 300: 🤖 **Claude** · 💬 **Note**

When both contactId and dealId are null, contextPredicates is empty and cb.or(new Predicate[0]) evaluates to false, so the query silently returns zero rows (and total=0) regardless of excludeTaskId. There is no validation that at least one of contactId/dealId is supplied (unlike createTask which calls validateTaskTargets). Consider validating that at least one context filter is present and returning a clear ModuleException, rather than silently returning an empty page for a malformed request.

backend/src/main/java/com/skapp/community/crmplanner/payload/request/CrmTaskRelatedFilterDto.java
```diff
@@ -0,0 +1,20 @@
...
+@Setter
+public class CrmTaskRelatedFilterDto {
+
+	private Long contactId;
+
+	private Long dealId;
+
+	private Long excludeTaskId;
+
+	private int page = 0;
```
ThinuwanW, line 14: 🤖 **Claude** · 💬 **Note**

page and size have no bean-validation bounds. A negative page (e.g. ?page=-1) makes PageRequest.of throw IllegalArgumentException, which is not wrapped in ModuleException and surfaces as a 500; size is unbounded, allowing a very large page request. Add @Min(0) on page and @Min(1)/@Max(...) on size (and @Validated/@Valid on the binding) consistent with how pagination DTOs should be guarded. Note CrmTaskCompletedFilterDto shares this gap, so this is a pre-existing pattern worth addressing here.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmTaskServiceImpl.java
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Code duplication (guidelines checklist #9). This map block — `crmTaskToCrmTaskDetailResponseDto(task)` followed by `dto.setIsOverdue(!Boolean.TRUE.equals(task.getIsCompleted()) && task.getDueAt() != null && task.getDueAt().isBefore(LocalDate.now().atStartOfDay()))` — is a byte-for-byte copy of CrmContactServiceImpl.getContactById (lines 326-331). The same overdue logic also exists as Criteria expressions in CrmTaskRepositoryImpl.findOpenTaskSummaryByContactIds and findTaskMetricsByContactId, so the rule now lives in 3+ places. Any correctness fix (e.g. the overdue computation concern raised in the first pass) must be applied in every copy. Extract a single helper — e.g. a `CrmTaskDetailResponseDto toDetailDto(CrmTask)` in a shared util/mapper @AfterMapping, or a `boolean isTaskOverdue(CrmTask)` in CrmUtil — and call it from both services.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmTaskRepositoryImpl.java
```diff
@@ -245,6 +247,58 @@ private void applyFetchGraph(Root<CrmTask> root) {
 		root.fetch(CrmTask_.deal, JoinType.LEFT);
 	}
 
+	@Override
```
ThinuwanW, line 250: 🤖 **Claude** · 🟣 **Nit**

Code duplication (guidelines checklist #9). findRelatedTasksPaginated repeats the entire pagination/count scaffolding of buildFindCompletedTasksQuery almost verbatim (createQuery, applyFetchGraph, setFirstResult/setMaxResults, separate count CriteriaQuery with cb.count, PageImpl assembly) — only the predicate-builder call differs. Consider extracting a private generic helper such as `Page<CrmTask> paginate(Pageable pageable, BiFunction<CriteriaBuilder, Root<CrmTask>, List<Predicate>> predicateBuilder)` that both methods reuse, reducing the risk of the two diverging (the missing ORDER BY noted in the first pass is exactly this kind of drift).

backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmTaskController.java
```diff
@@ -84,4 +85,13 @@ public ResponseEntity<ResponseEntityDto> deleteTask(@PathVariable Long id) {
 		return new ResponseEntity<>(response, HttpStatus.OK);
 	}
 
+	@Operation(summary = "Get related tasks",
+			description = "Returns paginated tasks filtered by contactId and/or dealId.")
+	@GetMapping("/related-tasks")
+	@PreAuthorize("hasRole('ROLE_CRM_SALES_REPRESENTATIVE')")
+	public ResponseEntity<ResponseEntityDto> getRelatedTasks(CrmTaskRelatedFilterDto filterDto) {
```
ThinuwanW, line 92: 🤖 **Claude** · 🟣 **Nit**

The getRelatedTasks parameter binds CrmTaskRelatedFilterDto from query params but is not annotated `@Valid` (nor is the class `@Validated`). This compounds the missing page/size bounds noted in the first pass: even if `@Min`/`@Max` bounds are added to the DTO fields, they will not be enforced for query-parameter binding unless `@Valid` is placed on this controller parameter. Add `@Valid` here when the DTO gains validation annotations.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmTaskServiceImpl.java
YasinduRC: Since the overdue logic is handled by the frontend (Inside taskRow),  this logic and feild in the responseDTO is not needed.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmTaskRepositoryImpl.java
YasinduRC: The findTasks method uses dueAt nulls-last + id tiebreaker. Applied the same order to findRelatedTasksPaginated.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmTaskServiceImpl.java
YasinduRC: Since the overdue logic is handled by the frontend (Inside taskRow), this logic and feild in the responseDTO is not needed.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmTaskRepositoryImpl.java
YasinduRC: This is the existing codebase pattern, we write this pagination/count scaffolding inline rather than extracting separate functions for it.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmTaskRepositoryImpl.java
YasinduRC: In real usage, both contactId and dealId can  be null, the frontend will guard the API call and show the empty view directly rather than making the request, so throwing an error here would be incorrect for a valid UI state.

backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmTaskController.java
```diff
@@ -84,4 +85,13 @@ public ResponseEntity<ResponseEntityDto> deleteTask(@PathVariable Long id) {
 		return new ResponseEntity<>(response, HttpStatus.OK);
 	}
 
+	@Operation(summary = "Get related tasks",
+			description = "Returns paginated tasks filtered by contactId and/or dealId.")
+	@GetMapping("/related-tasks")
```
sanjula-rtc: crm/task/related

backend/src/main/java/com/skapp/community/crmplanner/util/CrmValidations.java
```diff
@@ -208,6 +208,7 @@ public static void validateTaskTargets(Long contactId, Long companyId, Long deal
 		}
 	}
 
+
```
sanjula-rtc: remove new line

backend/src/main/java/com/skapp/community/crmplanner/payload/request/CrmTaskRelatedFilterDto.java
```diff
@@ -0,0 +1,20 @@
...
+
+@Getter
+@Setter
+public class CrmTaskRelatedFilterDto {
+
+	private Long contactId;
+
+	private Long dealId;
+
+	private Long excludeTaskId;
```
sanjula-rtc: let's remove this

backend/src/main/java/com/skapp/community/crmplanner/repository/CrmTaskRepository.java
```diff
@@ -28,4 +29,6 @@ public interface CrmTaskRepository {
 
 	Map<Long, Long> countTasksByDealIds(List<Long> dealIds);
 
+	Page<CrmTask> findRelatedTasksPaginated(CrmTaskRelatedFilterDto filterDto, Pageable pageable);
```
sanjula-rtc: findRelatedTasks

backend/src/main/java/com/skapp/community/crmplanner/type/CrmTaskRelatedParams.java
```diff
@@ -0,0 +1,16 @@
...
+@Getter
+@AllArgsConstructor
+public class CrmTaskRelatedParams {
+
+	private final Long contactId;
+
+	private final Long dealId;
+
+	private final Long excludeTaskId;
+
```
sanjula-rtc, line 15: remove excludeTaskId

sanjula-rtc (changes requested):

backend/src/main/java/com/skapp/community/crmplanner/util/CrmValidations.java
YasinduRC: Removed the new line that was accidentally added

backend/src/main/java/com/skapp/community/crmplanner/payload/request/CrmTaskRelatedFilterDto.java
YasinduRC: Removed the excludeTaskId, it will be filtered out from the frontend

backend/src/main/java/com/skapp/community/crmplanner/repository/CrmTaskRepository.java
YasinduRC: Removed the Paginated from the repository function name

backend/src/main/java/com/skapp/community/crmplanner/type/CrmTaskRelatedParams.java
YasinduRC: Removed the excludeTaskId, it will be filtered out from the frontend

backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmTaskController.java
YasinduRC: updated the endpoint name

backend/src/main/java/com/skapp/community/crmplanner/util/CrmValidations.java
```diff
@@ -208,6 +208,7 @@ public static void validateTaskTargets(Long contactId, Long companyId, Long deal
 		}
 	}
```
sanjula-rtc, line 215: check the build and test errors as well

backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmTaskController.java
YasinduRC: Since there are no` @Min/@Max `constraints on this DTO yet, adding `@Valid` now would be a no-op and misleading.

backend/src/main/java/com/skapp/community/crmplanner/payload/request/CrmTaskRelatedFilterDto.java
YasinduRC: This pattern (no Min/Max on page/size) is consistent across all CRM filter DTOs Eg: CrmCompanyFilterDto, CrmContactFilterDto, CrmDealFilterDto, and CrmTaskCompletedFilterDto all share it.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmTaskServiceImpl.java
YasinduRC: Added the restriction so sales reps can only view their own tasks

backend/src/main/java/com/skapp/community/crmplanner/util/CrmValidations.java
YasinduRC: checked test and build pass correctly

sanjula-rtc (dismissed):

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmTaskServiceImpl.java
```diff
@@ -121,6 +123,33 @@ public ResponseEntityDto getCompletedTasks(CrmTaskCompletedFilterDto filterDto)
 		return new ResponseEntityDto(false, response);
 	}
 
+	@Override
+	@Transactional(readOnly = true)
+	public ResponseEntityDto getRelatedTasks(CrmTaskRelatedFilterDto filterDto) {
+		log.info("getRelatedTasks: execution started");
+
```
akilarootcode, line 130: add service level validations and throws errors

akilarootcode (changes requested):

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmTaskRepositoryImpl.java
YasinduRC: added the service level validations with a exception when dealID and ContactId is NULL

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmTaskServiceImpl.java
YasinduRC: added the service level validations with a exception when dealID and ContactId is NULL

akilarootcode (dismissed):

akilarootcode:
address test faliure

akilarootcode (approved):

ThinuwanW (approved):

PR #1560, opened 2026-06-30
fix(auth): refresh access token via middleware & client to stop premature sign-out

akilarootcode:
## Problem
Employees were being signed out prematurely when the access token expired, instead of being silently refreshed.

## Changes (frontend only)

**`middleware.ts`** — when there is no access token:
- Derives the tenant from the request subdomain (`getSubdomain(request.nextUrl.hostname)`).
- Enterprise mode requires `{tenantId}_refreshToken`; community mode uses `refreshToken`.
- If a valid (non-expired) refresh token exists, calls the refresh-token API to mint an access token so the **role/route checks run with real claims**.
- If the refresh token is missing or expired → redirect to `/signin` (enterprise: also when the subdomain/tenant can't be resolved).

**`authUtils.ts`**
- `getAccessToken()` now refreshes on a **missing or expired** access token — covering client-side API calls that happen between navigations (the middleware only runs on navigations).
- Signs out when the refresh yields no token.
- Access-token cookie lifetime now tracks the token's own `exp` claim (instead of a fixed number of days).
- `signOut()` no longer redirect-loops when already on `/signin` or `/enterprise/signin`.

## Notes
- The middleware mints only to evaluate access for the navigation; the client issues its own refresh to persist the cookie (one extra refresh on a cold navigation, by design — keeps the middleware simple and loop-free).
- No backend changes included.


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds middleware-level access-token refresh using the refresh-token cookie, switches the access-token cookie expiry to track the JWT's own exp claim, and makes getAccessToken sign the user out when refresh fails. The core logic is sound, but the middleware refresh path does not persist the newly minted access token or forward the refresh endpoint's Set-Cookie back to the browser, which causes redundant refreshes and can break sessions if refresh tokens are rotated. A few smaller concerns around an unreachable branch, a hardcoded endpoint, and a refresh-token JWT assumption are also noted.

Found **9** new issue(s):

frontend/middleware.ts
```diff
@@ -208,6 +245,44 @@ export function middleware(request: NextRequest) {
...
+      refreshToken &&
+      !isTokenExpired(refreshToken)
+    ) {
+      const minted = await requestAccessTokenFromRefresh(
+        refreshTokenCookieName,
+        refreshToken,
+        tenantId
+      );
+
+      if (minted) {
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

The minted access token is stored only in the local `token` variable and used for routing, but it is never written back to the browser. Both `NextResponse.next()` (line 405) and the redirects create fresh responses with no Set-Cookie, so: (1) the client app still has no accessToken cookie and will immediately call getNewAccessToken again on mount — a redundant refresh on every cold navigation; (2) more seriously, the refresh endpoint's own Set-Cookie response (a rotated/single-use refresh token, and the new access token) is discarded by this server-side fetch. If the backend rotates refresh tokens, the browser keeps the now-invalidated refresh token, and the next client-side refresh will fail and force a sign-out. Fix: capture the refresh Response's `set-cookie` header and the minted token, build a `NextResponse` (e.g. `const res = NextResponse.next()`), and forward them (`res.cookies.set('accessToken', minted, {...})` plus the rotated refresh cookie) on the returned response.

frontend/middleware.ts
```diff
@@ -208,6 +245,44 @@ export function middleware(request: NextRequest) {
...
+    }
+
+    const refreshToken = refreshTokenCookieName
+      ? request.cookies.get(refreshTokenCookieName)?.value
+      : undefined;
+
+    if (
+      refreshTokenCookieName &&
+      refreshToken &&
+      !isTokenExpired(refreshToken)
```
ThinuwanW, line 269: 🤖 **Claude** · 💬 **Note**

`isTokenExpired(refreshToken)` decodes the refresh token via decodeJWTToken (base64 of `token.split('.')[1]`). This only works if the refresh token is a decodable JWT containing an `exp` claim. If the backend issues opaque refresh tokens, decodeJWTToken throws, isTokenExpired returns true, and the refresh is skipped entirely — the user is redirected to /signin despite holding a valid refresh token, breaking this whole feature. Please confirm the refresh token is a JWT; if it is not guaranteed to be, gate the refresh on cookie presence rather than on isTokenExpired.

frontend/src/community/auth/utils/authUtils.ts
```diff
@@ -169,16 +170,18 @@ export const getAccessToken = async (): Promise<string | null> => {
...
+  if (currentAccessToken && !isTokenExpired(currentAccessToken)) {
+    return currentAccessToken;
   }
 
-  if (isTokenExpired(currentAccessToken)) {
-    const newToken = await getNewAccessToken();
-    return newToken;
+  const newAccessToken = await getNewAccessToken();
+
+  if (!newAccessToken) {
```
ThinuwanW, line 180: 🤖 **Claude** · 💬 **Note**

getAccessToken now calls signOut() (default redirect=true) on ANY refresh failure, including transient/network errors from getNewAccessToken. Previously a failed refresh just returned null without logging the user out. Since getAccessToken is called from many places (including React Query refetches), a single transient backend hiccup will now redirect the user to /signin and clear cookies. Consider distinguishing auth failures (401/invalid refresh) from transient errors, or at least confirm this aggressive sign-out is intended.

frontend/middleware.ts
```diff
@@ -192,11 +201,39 @@ const allowedRoutes: Record<
...
+  refreshTokenCookieName: string,
+  refreshToken: string,
+  tenantId: string | undefined
+): Promise<string | null> => {
+  if (!refreshTokenCookieName || !tenantId || !refreshToken) return null;
 
-  const claims = extractClaimsFromToken(token || "");
+  try {
+    const response = await fetch(
+      `${process.env.NEXT_PUBLIC_API_URL}/v1/auth/session/refresh-token`,
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

The refresh URL `/v1/auth/session/refresh-token` is hardcoded here, but an endpoint constant already exists: `authenticationEndpoints.REFRESH_TOKEN` (`${ApiVersions.V1}/auth/session/refresh-token` in ~enterprise/common/api/utils/ApiEndpoints). Per the guidelines, use the endpoint constant instead of a hardcoded path so the two stay in sync.

frontend/middleware.ts
```diff
@@ -208,6 +245,44 @@ export function middleware(request: NextRequest) {
     return NextResponse.next();
   }
 
+  if (!token) {
+    const subdomain = getSubdomain(request.nextUrl.hostname);
+    const tenantId = Array.isArray(subdomain) ? subdomain[0] : subdomain;
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

`getSubdomain(request.nextUrl.hostname)` is called with a single argument, so `multipleValues` defaults to false and it always returns a `string` (parts[0]), never an array. The `Array.isArray(subdomain) ? subdomain[0] : subdomain` branch is therefore unreachable at runtime and only exists to satisfy the `string | string[]` return type. Simplify to `const tenantId = getSubdomain(request.nextUrl.hostname);` (or call with the explicit single-value intent) to avoid the misleading dead branch.

frontend/src/community/auth/utils/authUtils.ts
```diff
@@ -127,11 +127,12 @@ export const getNewAccessToken = async (): Promise<string | null> => {
 
 export const setAccessToken = (token: string) => {
   if (typeof window !== "undefined") {
-    const expiryDate = new Date(
-      Date.now() + COOKIE_EXPIRY_DAYS * unitConversion.MILLISECONDS_PER_DAY
-    );
+    const expirySeconds = extractClaimsFromToken(token)?.exp as number | undefined;
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

Minor: extractClaimsFromToken returns a non-nullable `Record<string, any>` (it returns `{}` on failure), so the `?.` on `extractClaimsFromToken(token)?.exp` is dead. The `as number | undefined` assertion is also a type cast to silence the `any`. Cleaner: `const claims = extractClaimsFromToken(token); const expirySeconds = typeof claims.exp === 'number' ? claims.exp : undefined;` which validates the runtime type instead of asserting it.

frontend/middleware.ts
```diff
@@ -208,6 +245,44 @@ export function middleware(request: NextRequest) {
     return NextResponse.next();
   }
 
+  if (!token) {
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

The new session-restoration path only runs `if (!token)` — it refreshes when the accessToken cookie is ABSENT, but never when it is present-but-expired. The PR added `isTokenExpired` and applies it to the refresh token (line 266) yet never validates the access token's own expiry. Because legacy cookies set by the previous deployment use the long COOKIE_EXPIRY_DAYS lifetime (and even new cookies have a brief overlap window around `exp` due to second-granular cookie purging / clock skew), an expired-but-still-present accessToken bypasses refresh entirely. The middleware then calls `extractClaimsFromToken(token)` on the expired JWT and authorizes routing using stale `roles`/`tenantStatus`. Fix: guard with `if (!token || isTokenExpired(token))` so an expired access token also triggers the refresh flow, matching the intent of this feature.

frontend/middleware.ts
```diff
@@ -192,11 +201,39 @@ const allowedRoutes: Record<
...
-  const claims = extractClaimsFromToken(token || "");
+  try {
+    const response = await fetch(
+      `${process.env.NEXT_PUBLIC_API_URL}/v1/auth/session/refresh-token`,
+      {
+        method: "POST",
+        headers: {
+          "Content-Type": "application/json",
+          cookie: `${refreshTokenCookieName}=${refreshToken}`,
+          "X-Tenant-ID": tenantId
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

In community mode the refresh request always sets `X-Tenant-ID` to `tenantId`, which is just the first hostname label (e.g. "localhost", "app", "www") since `getSubdomain` is called unconditionally. Tenancy is not a concept in community mode (the backend's community `getRefreshTokenFromCookies(request, tenantId)` ignores the tenant per CookieUtil.java), so the header is meaningless there; more importantly `requestAccessTokenFromRefresh`'s guard `if (!tenantId ...) return null` makes community refresh depend on a value it should not need. Recommend deriving/sending tenantId and X-Tenant-ID only when `isEnterpriseMode()` is true, and omitting both in community mode.

frontend/middleware.ts
```diff
@@ -192,11 +201,39 @@ const allowedRoutes: Record<
   ...commonRoutes
 };
 
-export function middleware(request: NextRequest) {
-  // Get accessToken from cookies
-  const token = request.cookies.get("accessToken")?.value;
+const requestAccessTokenFromRefresh = async (
```
ThinuwanW, line 205: 🤖 **Claude** · 💬 **Note**

Code duplication: `requestAccessTokenFromRefresh` re-implements the same refresh call and response-shape parsing (`data?.results?.[0]?.accessToken`) that `getNewAccessToken` in authUtils.ts already performs. This means the brittle `results[0].accessToken` access path and the refresh endpoint reference now live in two places and must be kept in sync (the first pass already noted the hardcoded URL on line 213). Consider extracting a shared helper for parsing the refresh response, or at minimum a shared constant for the `results[0].accessToken` shape, so a backend response-shape change only needs one edit.

frontend/middleware.ts
akilarootcode: Fixed in d443d5833. The guard is now `if (!token || isTokenExpired(token))`, so an expired-but-present access token also triggers the refresh flow; if the refresh yields no token it redirects to sign-in (no more routing on stale claims).

frontend/middleware.ts
akilarootcode: Fixed in d443d5833. Now uses `${process.env.NEXT_PUBLIC_API_URL}${authenticationEndpoints.REFRESH_TOKEN}` instead of the hardcoded path.

frontend/src/community/auth/utils/authUtils.ts
akilarootcode: Fixed in d443d5833. Replaced with `const claims = extractClaimsFromToken(token); const expirySeconds = typeof claims.exp === "number" ? claims.exp : undefined;` — runtime validation instead of the `as` cast / dead optional chain.

frontend/middleware.ts
akilarootcode: Confirmed: the refresh token is a JWT — it is produced by `jwtService.generateRefreshToken(...)` (HS-signed, with an `exp` claim), so `decodeJWTToken`/`isTokenExpired` work. The call is also already gated on cookie presence (`refreshToken && ...`) before the expiry check, so an unreadable value would just be treated as expired rather than NPE.

frontend/middleware.ts
akilarootcode: Intentional trade-off (discussed and accepted): the middleware mints purely to evaluate role/route access with real claims; the client (`getAccessToken` → `getNewAccessToken`) issues its own refresh to persist the cookie. The accepted cost is one extra refresh on a cold navigation, in exchange for keeping the middleware simple and free of redirect-to-self loops. Re: rotation — the refresh-token endpoint does not rotate/invalidate the refresh token (the client reuses the same refresh cookie across a session today), so discarding the server-side fetch's Set-Cookie is safe here. If we ever move to rotating refresh tokens, we'll forward the Set-Cookie.

frontend/middleware.ts
akilarootcode: Kept intentionally. `getSubdomain` is typed `string | string[]`, so without the `Array.isArray` narrowing TS won't let the result flow into `${tenantId}_refreshToken` / the `string | undefined` param. A prior review rejected the `as string` cast, so this narrowing is the type-safe alternative. The proper fix is an overloaded return type on `getSubdomain`, which is a shared util out of scope here.

frontend/middleware.ts
akilarootcode: Acknowledged. Full sharing isn't clean because the two run in different runtimes: `requestAccessTokenFromRefresh` is Edge-middleware using `fetch` (manual cookie/X-Tenant-ID, no cookie persistence), while `getNewAccessToken` is client-side using the `authAxios` instance (interceptors, withCredentials, setAccessToken). The only real overlap is the `results[0].accessToken` shape; happy to extract a tiny shared parser/constant for that if preferred.

frontend/middleware.ts
akilarootcode: Fixed in c72685be9. tenantId/X-Tenant-ID are now derived and sent only when isEnterpriseMode(); community mode uses the plain 'refreshToken' cookie and requestAccessTokenFromRefresh no longer requires a tenant id.

frontend/src/community/auth/utils/authUtils.ts
akilarootcode: Intentional. The guaranteed sign-out is by design so the redirect to /signin does not depend on the backend returning a specific messageKey. The response interceptor still handles auth-specific failures, and this is the final safety net for a no-session state. Accepting that a transient backend error can also sign the user out is a deliberate trade-off for this PR.

hasalarootcode (dismissed):

hasalarootcode (dismissed):

PR #1570, opened 2026-07-02
Fix/job families cannot be deleted issue

gaindurootcode:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

The PR replaces the job-family fetch used by getAllJobFamilies with a new repository method that batch-loads active job titles, correctly fixing a pre-existing bug where inactive titles leaked into the response. The core logic is sound under the readOnly transaction, but the new method mutates managed entities backed by a CascadeType.ALL association and does not address the existing N+1 on employees. No blocking bugs found.

Found **6** new issue(s):

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/JobFamilyRepositoryImpl.java
```diff
@@ -76,6 +79,50 @@ public List<JobFamily> getJobFamiliesByEmployeeCount() {
...
+		criteriaQuery.orderBy(criteriaBuilder.desc(criteriaBuilder.count(employeeJoin.get(Employee_.employeeId))));
+
+		List<JobFamily> jobFamilies = entityManager.createQuery(criteriaQuery).getResultList();
+		if (jobFamilies.isEmpty()) {
+			return jobFamilies;
+		}
+
+		List<Long> jobFamilyIds = jobFamilies.stream().map(JobFamily::getJobFamilyId).toList();
+		Map<Long, Set<JobTitle>> activeJobTitlesByFamilyId = getActiveJobTitlesByJobFamilyIds(jobFamilyIds);
+		jobFamilies.forEach(jobFamily -> jobFamily
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

setJobTitles(...) replaces the persistent collection on a *managed* JobFamily entity. JobFamily.jobTitles is @ManyToMany with cascade = CascadeType.ALL, so if a flush ever occurred, removing the inactive titles from the collection would delete join rows and cascade a REMOVE onto the JobTitle entities themselves. This is only safe today because the caller (getAllJobFamilies) is @Transactional(readOnly = true), which sets FlushMode.MANUAL. That is fragile: if readOnly is ever dropped, or this method is reused from a write transaction, it risks deleting inactive JobTitle rows. Consider returning the active titles via a separate projection/DTO (like getJobFamilyOverview / getJobTitlesByJobFamily do) or a fetch-join into a fresh query instead of mutating the managed association, so behavior does not depend on the transaction's flush mode.

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/JobFamilyRepositoryImpl.java
```diff
@@ -76,6 +79,50 @@ public List<JobFamily> getJobFamiliesByEmployeeCount() {
 		return orderedIds.stream().map(jobFamilyMap::get).filter(Objects::nonNull).toList();
 	}
 
+	@Override
+	public List<JobFamily> getJobFamiliesWithJobTitles() {
+		CriteriaBuilder criteriaBuilder = entityManager.getCriteriaBuilder();
+
+		CriteriaQuery<JobFamily> criteriaQuery = criteriaBuilder.createQuery(JobFamily.class);
+		Root<JobFamily> root = criteriaQuery.from(JobFamily.class);
+		Join<JobFamily, Employee> employeeJoin = root.join(JobFamily_.employees, JoinType.LEFT);
```
ThinuwanW, line 88: 🤖 **Claude** · 💬 **Note**

The main query LEFT-joins employees only to compute the count for ordering; it does not fetch them. Since JobFamilyResponseDetailDto exposes an `employees` list, the MapStruct mapper will call getEmployees() on each returned JobFamily, triggering one lazy SELECT per family (N+1). This N+1 also existed in the old getJobFamiliesByEmployeeCount path so it is not newly introduced, but since this method is being rewritten it would be a good opportunity to address it (e.g. an @EntityGraph / fetch join on employees, or a dedicated projection) — or confirm the frontend does not consume the employees field and consider dropping it from the DTO.

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/JobFamilyRepositoryImpl.java
```diff
@@ -76,6 +79,50 @@ public List<JobFamily> getJobFamiliesByEmployeeCount() {
...
+			.setJobTitles(activeJobTitlesByFamilyId.getOrDefault(jobFamily.getJobFamilyId(), new HashSet<>())));
+
+		return jobFamilies;
+	}
+
+	private Map<Long, Set<JobTitle>> getActiveJobTitlesByJobFamilyIds(List<Long> jobFamilyIds) {
+		CriteriaBuilder criteriaBuilder = entityManager.getCriteriaBuilder();
+		CriteriaQuery<JobFamilyTitle> criteriaQuery = criteriaBuilder.createQuery(JobFamilyTitle.class);
+		Root<JobFamilyTitle> root = criteriaQuery.from(JobFamilyTitle.class);
+		root.fetch(JobFamilyTitle_.jobTitle, JoinType.INNER);
```
ThinuwanW, line 104: 🤖 **Claude** · 💬 **Note**

getActiveJobTitlesByJobFamilyIds both fetch-joins jobTitle (root.fetch(JobFamilyTitle_.jobTitle, INNER)) and separately navigates root.get(JobFamilyTitle_.jobTitle).get(JobTitle_.isActive) in the WHERE. Depending on the Hibernate version this can emit two joins to job_title rather than reusing the fetch join. Minor; consider capturing the fetch as a Join (Join titleJoin = (Join) root.fetch(...)) and reusing it in the predicate to guarantee a single join.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/JobServiceImpl.java
```diff
@@ -59,7 +59,7 @@ public class JobServiceImpl implements JobService {
 	public ResponseEntityDto getAllJobFamilies() {
 		log.info("getAllJobFamilies: execution started");
 
-		List<JobFamily> jobFamilies = jobFamilyDao.getJobFamiliesByEmployeeCount();
+		List<JobFamily> jobFamilies = jobFamilyDao.getJobFamiliesWithJobTitles();
```
ThinuwanW, line 64: 🤖 **Claude** · 💬 **Note**

This is a genuine behavioral change (getAllJobFamilies now returns only active job titles instead of all associated titles) but there is no test covering getAllJobFamilies or the new getJobFamiliesWithJobTitles repository method. Consider adding a test that asserts inactive job titles are excluded and that ordering by employee count is preserved, to lock in the fix.

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/JobFamilyRepositoryImpl.java
```diff
@@ -76,6 +79,50 @@ public List<JobFamily> getJobFamiliesByEmployeeCount() {
 		return orderedIds.stream().map(jobFamilyMap::get).filter(Objects::nonNull).toList();
 	}
 
+	@Override
+	public List<JobFamily> getJobFamiliesWithJobTitles() {
```
ThinuwanW, line 83: 🤖 **Claude** · 💬 **Note**

Code duplication: getJobFamiliesWithJobTitles re-implements the exact 'active job families ordered by employee-count desc' selection that getJobFamiliesByEmployeeCount already performs (both LEFT-join employees, filter isActive=true, group by job family, order by count(employeeId) desc). The only real difference is how job titles are attached afterwards. Consider extracting a shared private helper that returns the ordered active JobFamily list, and have both public methods build on it — otherwise future ordering/filtering changes must be made in two places and will drift. Note one minor inconsistency introduced by the copy: the new version drops the coalesce(count, 0) wrapper used in getJobFamiliesByEmployeeCount (harmless since COUNT of a LEFT-joined column yields 0, not null, but worth aligning for readability).

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/JobServiceImpl.java
ThinuwanW, line 64: 🤖 **Claude** · 💬 **Note**

Divergence / missing paired change across the two 'list jobs' paths. The community getAllJobFamilies now uses getJobFamiliesWithJobTitles, which pushes active-job-title filtering into the query. The enterprise EpJobServiceImpl.getJobs() (backend/src/main/java/com/skapp/enterprise/common/service/impl/EpJobServiceImpl.java:35-38) still calls the old getJobFamiliesByEmployeeCount() and then lazily loads every job title and filters with filterActiveJobTitles() in Java — the exact lazy-load-all-then-filter pattern the new query was meant to eliminate. Two nearly identical operations now use different loading strategies, so a change to job-title filtering rules will silently apply to only one of them. Confirm whether the enterprise path should also adopt getJobFamiliesWithJobTitles (or a shared helper) to keep behavior and performance consistent; if the divergence is intentional, a short comment explaining why would help.

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/JobFamilyRepositoryImpl.java
gaindurootcode: Not needed

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/JobServiceImpl.java
gaindurootcode: No testes needed

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/JobFamilyRepositoryImpl.java
gaindurootcode: This is intentional

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/JobFamilyRepositoryImpl.java
gaindurootcode: Not needed

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/JobServiceImpl.java
gaindurootcode: Out of scope for this PR

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/JobFamilyRepositoryImpl.java
gaindurootcode: Removed all the managed entities from modifying. Used DTOs instead

PR #1578, opened 2026-07-02
feat: enforce HTTPS for website validation and update tests

dulsara-manakal:
### Summary

- Add validation to company website
- only https website urls are valid 
- http urls cannot be add

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR tightens website URL validation to require the HTTPS scheme, both in the backend (URI scheme check via a new HTTPS_SCHEME constant) and frontend (a rewritten regex), with corresponding test updates. The changes are consistent across layers, all existing integration tests already use https values so nothing breaks, and the new regex correctly satisfies its test cases. Only minor suggestions remain around the intentional http rejection and a pre-existing frontend/backend path-character asymmetry.

frontend/src/community/common/regex/regexPatterns.ts
```diff
@@ -86,7 +86,7 @@ export function isValidDateInYYYYMMDDFormat(): RegExp {
 }
 
 export function isValidWebsiteUrl(): RegExp {
-  return /^(https?:\/\/)?(www\.)?([a-zA-Z0-9-]+\.)+[a-zA-Z]{2,}(:\d+)?(\/[^\s]*)?(\?[^\s]*)?(#[^\s]*)?$/i;
+  return /^https:\/\/(www\.)?[a-z0-9-]{1,63}(\.[a-z0-9-]{1,63}){0,9}\.[a-z]{2,63}(:\d{1,5})?([/?#]\S*)?$/i;
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Behavioral tightening: both the frontend regex and backend now reject any non-https URL (http://, www.-only, bare domains). Any existing company records stored with an http:// website will fail validation when the company is next edited/saved, and users whose sites are http-only cannot save a website. This appears intentional given the PR goal, but please confirm it's a deliberate product decision and that pre-existing http website data won't cause save failures on edit. If http should remain acceptable, allow an optional scheme; otherwise no change needed.

ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

The path/query/fragment segment `([/?#]\S*)?` uses `\S*`, which permits characters that are illegal in a URI (e.g. `|`, `{`, `}`, `^`, backtick, backslash). The backend validates via `new URI(website)`, which rejects these, so a value like `https://acme.com/a|b` passes frontend validation but triggers a server-side CRM_ERROR_WEBSITE_INVALID. This asymmetry pre-dates the PR, but since the regex was fully rewritten here it's a good opportunity to restrict the path to a stricter character set (e.g. RFC 3986 pchar) so the two layers agree.

frontend/src/community/crm/utils/companyValidations.ts
```diff
@@ -32,7 +35,10 @@ export const addCompanyValidations = (translator: TranslatorFunctionType) =>
       .nullable()
       .optional()
       .transform((v) => (v === "" ? null : v))
-      .url(translator(["validations", "website"]))
+      .matches(isValidWebsiteUrl(), {
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`excludeEmptyString: true` is redundant here: the preceding `.transform((v) => (v === "" ? null : v))` already converts empty strings to null, and Yup's `.matches()` skips validation on null. Harmless, but can be dropped for clarity.

backend/src/main/java/com/skapp/community/crmplanner/util/CrmValidations.java
```diff
@@ -104,7 +104,7 @@ public static void validateWebsite(String website) {
 
 		try {
 			URI uri = new URI(website);
-			if (!uri.isAbsolute()) {
+			if (!uri.isAbsolute() || !CrmConstants.HTTPS_SCHEME.equalsIgnoreCase(uri.getScheme())) {
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

The `!uri.isAbsolute()` check is now redundant. `URI.isAbsolute()` is defined as `scheme != null`, so whenever a URI is not absolute its scheme is null, and `CrmConstants.HTTPS_SCHEME.equalsIgnoreCase(null)` already returns false — the second clause `!HTTPS_SCHEME.equalsIgnoreCase(uri.getScheme())` alone covers every non-absolute case. The `!uri.isAbsolute()` term can never independently trigger. Consider dropping it (or keeping only the scheme check) for clarity: `if (!CrmConstants.HTTPS_SCHEME.equalsIgnoreCase(uri.getScheme())) { ... }`.

frontend/src/community/common/regex/regexPatterns.ts
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

The `(www\.)?` group is redundant. A `www` label already matches the general leading-label token `[a-z0-9-]{1,63}`, so every string that matches with the optional `www\.` also matches without it (e.g. `https://www.example.com` → `www` as the leading label, `.example` as a subdomain group, `.com` as TLD). The group adds no additional accepted inputs and can be removed to simplify the pattern.

ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Residual frontend/backend validation asymmetry. The frontend regex requires a dotted host with a letters-only TLD, whereas the backend `validateWebsite` (CrmValidations.java) only requires `uri.isAbsolute()` + https scheme + `toURL()`. As a result the backend accepts values the UI rejects — e.g. `https://localhost`, `https://192.168.1.1`, single-label hosts like `https://intranet`, and opaque forms like `https:example.com`. A company created/edited directly via the API (or seeded/imported) can therefore hold a website the UI considers invalid, and the two layers disagree on the same field. Consider aligning: give the backend an equivalent host/TLD check, or base the frontend check on a URL parser so both accept the same set. If the divergence is intentional, a short comment noting it would help.

ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

The subdomain group `(\.[a-z0-9-]{1,63}){0,9}` caps the number of intermediate labels at 9 (max ~11 labels total including leading label and TLD). Legitimate deeply-nested hostnames (more than ~11 dotted labels) will be silently rejected. This is an edge case, but if arbitrary valid hostnames must be supported consider relaxing the `{0,9}` bound (e.g. `{0,}` with a total-length guard, which `.max()` already provides).

dulsara-manakal: This is intentional and done according to the requirements.

dulsara-manakal: Tighten the regex to match RFC 3986's pchar set for the path/query/fragment portion

frontend/src/community/crm/utils/companyValidations.ts
dulsara-manakal: drop that line

backend/src/main/java/com/skapp/community/crmplanner/util/CrmValidations.java
dulsara-manakal: remove that part.

frontend/src/community/common/regex/regexPatterns.ts
dulsara-manakal: fixed this

dulsara-manakal: The total length is already capped via .max(CHARACTER_LENGTH) in the Yup schema, so this doesn't open up unbounded input.

dulsara-manakal: BE end already handle it with `https` and since FE already validates this, malformed website link wont be able to  enter to DB

dulsara-manakal: This regex change won't affect anything existing since this isn't used anywhere else.

hasalarootcode (dismissed):

frontend/src/community/common/regex/regexPatterns.ts
```diff
@@ -86,7 +86,7 @@ export function isValidDateInYYYYMMDDFormat(): RegExp {
 }
 
 export function isValidWebsiteUrl(): RegExp {
-  return /^(https?:\/\/)?(www\.)?([a-zA-Z0-9-]+\.)+[a-zA-Z]{2,}(:\d+)?(\/[^\s]*)?(\?[^\s]*)?(#[^\s]*)?$/i;
+  return /^https:\/\/([a-z0-9-]+\.)+[a-z]{2,63}(:\d{1,5})?(?:[/?#][a-z0-9\-._~!$&'()*+,;=:@/?%]*)?$/i;
```
akilarootcode: add as new pattern

akilarootcode (changes requested):

frontend/src/community/common/regex/regexPatterns.ts
dulsara-manakal: added new one

frontend/src/community/crm/utils/companyValidations.ts
```diff
@@ -2,8 +2,12 @@ import * as Yup from "yup";
 
 import { characterLengths } from "~community/common/constants/stringConstants";
 import { isValidPhoneNumber } from "~community/common/regex/regexPatterns";
-import { CrmIndustryEnum } from "~community/crm/enums/common";
 import { TranslatorFunctionType } from "~community/common/types/CommonTypes";
+import { CrmIndustryEnum } from "~community/crm/enums/common";
+
+export function isValidCompanyWebsiteUrl(): RegExp {
```
akilarootcode: move this to regex file

akilarootcode (changes requested):

frontend/src/community/crm/utils/companyValidations.ts
dulsara-manakal: moved

akilarootcode (approved):

ThinuwanW (approved):

PR #1582, opened 2026-07-06
fix: notify supervisor on employee leave request cancellation

akilarootcode:
## Summary

Supervisors were not receiving any notification or email when an employee cancelled a leave request.

**Root cause:** The UI cancel action calls `DELETE /leaves/{id}` → `LeaveServiceImpl.deleteLeaveRequestById`, which only set the status to `CANCELLED` and deducted entitlement hours — it never sent any email or in-app notification. The cancellation notification logic existed **only** in the separate PATCH path (`updateLeaveRequestByEmployee`), which the cancel action does not use.

## Fix

- Extracted the four cancel email/notification calls into a shared private helper `sendCancelLeaveRequestEmailsAndNotifications(currentUser, leaveRequest)`.
- Invoke it from `deleteLeaveRequestById` (the actual UI cancel path) — **this is the fix**.
- Refactored the existing PATCH path (`updateLeaveRequestByEmployee`) to call the same helper (no behavior change there), so both cancel paths stay in sync.

The helper sends, on cancellation:
- Employee confirmation email + in-app notification
- Supervisor(s) email + in-app notification

Recipients resolve via `employeeManagerDao.findByEmployee(...)` (primary + secondary supervisors), filtered by `PeopleUtil.filterManagersByLeaveRoles(...)` (Leave Manager / Leave Admin) — consistent with the apply/approve flows.

## Bug report

- **Env:** Enterprise - Core - DEV
- **Severity:** Major
- **Steps:** Submit a leave request as employee → cancel it → check supervisor's notifications and email inbox.
- **Before:** Supervisor received nothing.
- **After:** Supervisor receives both an in-app notification and an email.

## Testing

- Backend compiles cleanly (`mvn compile`), spring-javaformat validation passes.
- Manual end-to-end verification on DEV recommended: cancel a leave request as an employee and confirm the assigned supervisor (with a Leave Manager/Admin role) receives both the notification and the email.

### DIFF
--- a/backend/src/main/java/com/skapp/community/leaveplanner/service/impl/LeaveServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/leaveplanner/service/impl/LeaveServiceImpl.java
@@ -340,22 +340,8 @@ public class LeaveServiceImpl implements LeaveService {
 
 		leaveRequest = leaveRequestDao.save(leaveRequest);
 
-		boolean isSingleDay = leaveRequest.getStartDate().equals(leaveRequest.getEndDate());
-		List<EmployeeManager> employeeManagers = employeeManagerDao.findByEmployee(currentUser.getEmployee());
-
 		if (leavePatchRequestDto.getLeaveRequestStatus().equals(LeaveRequestStatus.CANCELLED)) {
-			leaveEmailService.sendCancelLeaveRequestEmployeeEmail(currentUser.getEmail(),
-					currentUser.getEmployee().getFirstName(), currentUser.getEmployee().getLastName(),
-					leaveRequest.getLeaveState().toString(), leaveRequest.getLeaveType().getName(),
-					leaveRequest.getStartDate(), leaveRequest.getEndDate(), isSingleDay);
-			leaveEmailService.sendCancelLeaveRequestManagerEmail(employeeManagers,
-					currentUser.getEmployee().getFirstName(), currentUser.getEmployee().getLastName(),
-					leaveRequest.getLeaveState().toString(), leaveRequest.getLeaveType().getName(),
-					leaveRequest.getStartDate(), leaveRequest.getEndDate(), isSingleDay);
-			leaveNotificationService.sendCancelLeaveRequestEmployeeNotification(currentUser.getEmployee(),
-					employeeManagers, leaveRequest, isSingleDay);
-			leaveNotificationService.sendCancelLeaveRequestManagerNotification(currentUser.getEmployee(),
-					employeeManagers, leaveRequest, isSingleDay);
+			sendCancelLeaveRequestEmailsAndNotifications(currentUser, leaveRequest);
 		}
 
 		LeaveRequestByIdResponseDto leaveRequestResponseDto = leaveMapper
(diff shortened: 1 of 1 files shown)


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR extracts the leave-cancellation email/notification logic into a private helper (`sendCancelLeaveRequestEmailsAndNotifications`) and invokes it from `deleteLeaveRequestById`, so supervisors are now notified when an employee deletes/cancels a request. The extract-method refactor is behavior-preserving — arguments match the original code exactly, and `leaveState`/`leaveType` are non-nullable per the entity so no new NPE is introduced. The only notable concern is that the delete path lacks the idempotency guard the update path has, so already-cancelled/revoked requests will now trigger duplicate notifications.

Found **3** new issue(s): 🟡 3 suggestion(s)

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/LeaveServiceImpl.java
```diff
@@ -442,13 +428,33 @@ public ResponseEntityDto deleteLeaveRequestById(@NonNull Long id) {
 
 		leaveRequest = leaveRequestDao.save(leaveRequest);
 
+		sendCancelLeaveRequestEmailsAndNotifications(currentUser, leaveRequest);
```
ThinuwanW, line 431: 🤖 **Claude** · 🟡 **Suggestion**

deleteLeaveRequestById only rejects APPROVED/DENIED (lines 421-424), so CANCELLED and REVOKED requests still fall through, get re-set to CANCELLED, re-deducted via deductUsedHoursFromLeaveEntitlement, and — new with this PR — now fire cancellation emails/notifications again. The employee-facing update path (updateLeaveRequestByEmployee, lines 331-333) guards against same-status transitions and throws LEAVE_ERROR_LEAVE_REQUEST_UPDATE_ON_SAME_STATUS; the delete path has no equivalent guard, creating an asymmetry that is now user-visible via duplicate notifications. Consider rejecting when the request is already CANCELLED (and deciding whether REVOKED should be allowed), e.g. throw LEAVE_ERROR_INVALID_LEAVE_REQUEST_STATUS_EMPLOYEE for those statuses too, so notifications are only sent on a genuine PENDING→CANCELLED transition.

ThinuwanW, line 431: 🤖 **Claude** · 🟡 **Suggestion**

The behavioral change here (notifying employee + supervisor on deletion) has no test coverage — there is no test referencing deleteLeaveRequestById in backend/src/test. Add/extend a service test that verifies leaveEmailService and leaveNotificationService are invoked (verify(...) on the two email + two notification methods) when a PENDING request is deleted, to lock in the fix and guard against regression.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/LeaveServiceImpl.java
```diff
@@ -340,22 +340,8 @@ public ResponseEntityDto updateLeaveRequestByEmployee(@NonNull LeavePatchRequest
 
 		leaveRequest = leaveRequestDao.save(leaveRequest);
 
-		boolean isSingleDay = leaveRequest.getStartDate().equals(leaveRequest.getEndDate());
-		List<EmployeeManager> employeeManagers = employeeManagerDao.findByEmployee(currentUser.getEmployee());
-
 		if (leavePatchRequestDto.getLeaveRequestStatus().equals(LeaveRequestStatus.CANCELLED)) {
```
ThinuwanW, line 343: 🤖 **Claude** · 🟡 **Suggestion**

Potential NullPointerException gating the new helper call. `leavePatchRequestDto.getLeaveRequestStatus().equals(LeaveRequestStatus.CANCELLED)` dereferences leaveRequestStatus without a null check. The earlier block at line 330 explicitly guards `getLeaveRequestStatus() != null`, and LeavePatchRequestDto.leaveRequestStatus has no @NotNull, so a valid PATCH that sets only isViewed=true (a real 'mark as viewed' flow handled at line 337) leaves the field null and this line throws NPE -> 500. This is pre-existing but lives in the hunk this PR modified and now guards the extracted sendCancelLeaveRequestEmailsAndNotifications call. Fix: invert to a null-safe comparison, e.g. `LeaveRequestStatus.CANCELLED.equals(leavePatchRequestDto.getLeaveRequestStatus())`, or reuse the `!= null` guard.

akilarootcode: Good catch, but this is pre-existing behaviour that this PR only relocated (via the extract-method refactor), not introduced. As this PR is a focused bug fix for the missing cancellation notification, I'll keep the null-safety hardening out of scope and track it separately.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/LeaveServiceImpl.java
akilarootcode: Agreed on the asymmetry. The re-set/re-deduction on the delete path is pre-existing though; this PR only restores the missing cancellation email/notification. Keeping the idempotency guard out of scope for this bug fix — it can be handled in a follow-up.

akilarootcode: Fair point. Scoping this PR to the bug fix itself; will add test coverage for the notify-on-delete behaviour in a separate follow-up.

shakila-kamalasena (approved):

yuhanga-rtc (approved):

ThinuwanW (approved):

PR #1583, opened 2026-07-06
fix: show country for restricted profile viewers

akilarootcode:
## Summary

Fixes the bug where the **country value is not displayed** in a user's profile even though a country has been set (reported for the Super Admin user, but affects any user).

### Root cause
The country field is rendered inside the **always-visible general details section** of the profile (read-only view), but it reads its value from `personal.contact.country`. On the backend, `country` was only populated inside the **sensitive contact block**, which is gated behind `accessLevel.canSeeSensitiveData()`. So when a regular user views another user's profile (`RESTRICTED` access), the whole contact block is dropped and the country field renders with an empty value.

### Fix
In `PeopleReadServiceImpl.mapPersonalDetails`, always set `country` on the contact DTO for every viewer. Viewers with sensitive-data access still get the full contact block (email, phone, address); restricted viewers now get only the country populated, keeping the other contact details hidden.

## Testing
- Regular user viewing another user's profile now sees the country value.
- Sensitive contact fields (email, phone, address) remain hidden for restricted viewers.
- Full-access / own-profile views are unaffected.

### DIFF
--- a/backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleReadServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleReadServiceImpl.java
@@ -121,6 +121,10 @@ public class PeopleReadServiceImpl implements PeopleReadService {
 		dto.setGeneral(mapPersonalGeneralDetails(employee, accessLevel));
 		dto.setSkills(mapEmployeeSkills(employee));
 
+		EmployeePersonalContactDetailsDto contactDto = new EmployeePersonalContactDetailsDto();
+		contactDto.setCountry(employee.getCountry());
+		dto.setContact(contactDto);
+
 		if (accessLevel.canSeeSensitiveData()) {
 			dto.setContact(mapPersonalContactDetails(employee));
 


ThinuwanW (commented):
## 🤖 Claude Code Review

The PR correctly adds the employee's country to the personal contact details for RESTRICTED profile viewers, matching the stated intent. The DTO field, entity field, and access-level branching all check out, and behavior for sensitive-data viewers is preserved since mapPersonalContactDetails overwrites the country-only DTO. Only a minor redundancy is worth noting.

Found **3** new issue(s): 🟡 1 suggestion(s), 🟣 2 nit(s)

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleReadServiceImpl.java
```diff
@@ -121,6 +121,10 @@ private EmployeePersonalDetailsDto mapPersonalDetails(Employee employee,
 		dto.setGeneral(mapPersonalGeneralDetails(employee, accessLevel));
 		dto.setSkills(mapEmployeeSkills(employee));
 
+		EmployeePersonalContactDetailsDto contactDto = new EmployeePersonalContactDetailsDto();
```
ThinuwanW, line 124: 🤖 **Claude** · 🟣 **Nit**

Minor: for OWN_PROFILE/FULL_ACCESS viewers (canSeeSensitiveData() == true), this newly-created contactDto is allocated and set on the DTO only to be immediately discarded and overwritten by mapPersonalContactDetails(employee) at line 128-129. Consider only building the country-only contact for the restricted case, e.g. move these three lines into an `else` branch of the `if (accessLevel.canSeeSensitiveData())` block. This avoids the wasted allocation and makes the intent (country-only fallback for restricted viewers) clearer. Functionally the current code is correct, so this is a readability/cleanliness suggestion.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleReadServiceImpl.java
```diff
@@ -121,6 +121,10 @@ private EmployeePersonalDetailsDto mapPersonalDetails(Employee employee,
...
+		contactDto.setCountry(employee.getCountry());
```
ThinuwanW, line 125: 🤖 **Claude** · 🟡 **Suggestion**

Single-source-of-truth / drift risk: `country` is now mapped from `employee.getCountry()` in two independent places — this new block (line 125) and `mapPersonalContactDetails` (line 184). If the source of country ever changes (e.g. it should come from `personalInfo` or a lookup), one path will silently diverge and RESTRICTED viewers will see a different value than privileged viewers. Prefer restructuring so the minimal contact DTO is built only for the RESTRICTED case (e.g. an `else` branch to the existing `canSeeSensitiveData()` check) so the privileged path remains the sole full mapper. This is a maintainability angle distinct from the redundant-allocation note.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleReadServiceImpl.java
ThinuwanW, line 124: 🤖 **Claude** · 🟣 **Nit**

Missing test coverage for the new behavior. This PR changes the response contract: RESTRICTED viewers previously received a null `contact` object and now receive a non-null `EmployeePersonalContactDetailsDto` populated only with `country` (all other fields null). There is no unit test for PeopleReadServiceImpl and PeopleControllerIntegrationTest does not appear to assert this. Add a test asserting that for a RESTRICTED viewer `personal.contact` is non-null, `country` is populated, and sensitive fields (personalEmail, contactNo, addressLine1/2, city, state, postalCode) remain null — guarding against regression and against accidental sensitive-field exposure.

akilarootcode: Intentional trade-off. Setting the country-only `contactDto` unconditionally keeps "country is always returned" as the obvious base case at the top of the method, with the full contact mapper layered on top only for privileged viewers. The cost is a single lightweight DTO allocation per request for privileged viewers, which I considered negligible next to the extra branching. Keeping as-is for readability.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleReadServiceImpl.java
akilarootcode: Both paths resolve the value from the same `employee.getCountry()` getter, so there's still a single source of truth for the value itself — the only duplication is the `setCountry` call, not the source. If country's origin ever moves (e.g. to `personalInfo` or a lookup), it changes in the getter and both paths follow automatically. I deliberately kept the unconditional country-set so "country is always exposed" is explicit rather than tucked into an `else`. Noted on the maintainability angle though.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleReadServiceImpl.java
akilarootcode: Fair point — the response contract did change for RESTRICTED viewers (`contact` goes from null to a country-only DTO). This is worth a regression test asserting that for a RESTRICTED viewer `personal.contact` is non-null with `country` populated, and all sensitive fields (`personalEmail`, `contactNo`, `addressLine1/2`, `city`, `state`, `postalCode`) remain null — guarding both against regression and accidental sensitive-field exposure. Will add coverage for this.

yuhanga-rtc (approved):

ThinuwanW (approved):

PR #1586, opened 2026-07-06
feat: contact name search added to the contact lookup

anusham-rc:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
+++ b/backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
@@ -26,7 +26,6 @@ import jakarta.persistence.criteria.Predicate;
 import jakarta.persistence.criteria.Root;
 import jakarta.persistence.criteria.Subquery;
 
-import java.util.Optional;
 import lombok.RequiredArgsConstructor;
 import org.springframework.data.domain.Page;
 import org.springframework.data.domain.PageImpl;
@@ -127,9 +126,10 @@ public class CrmContactRepositoryImpl implements CrmContactRepository {
 		CriteriaBuilder cb = entityManager.getCriteriaBuilder();
 		CriteriaQuery<CrmContact> query = cb.createQuery(CrmContact.class);
 		Root<CrmContact> contact = query.from(CrmContact.class);
-		contact.fetch(CrmContact_.company, JoinType.LEFT);
+		Join<CrmContact, CrmCompany> company = (Join<CrmContact, CrmCompany>) contact.fetch(CrmContact_.company,
+				JoinType.LEFT);
 
-		List<Predicate> predicates = buildLookupPredicates(cb, contact, filterDto);
+		List<Predicate> predicates = buildLookupPredicates(cb, contact, company, filterDto);
 		query.where(predicates.toArray(new Predicate[0]));
 		query.orderBy(cb.asc(cb.lower(contact.get(CrmContact_.name))), cb.asc(contact.get(CrmContact_.id)));
 
@@ -141,15 +141,14 @@ public class CrmContactRepositoryImpl implements CrmContactRepository {
 	}
 
 	private List<Predicate> buildLookupPredicates(CriteriaBuilder cb, Root<CrmContact> contact,
-			CrmContactFilterDto filterDto) {
+			Join<CrmContact, CrmCompany> company, CrmContactFilterDto filterDto) {
 		List<Predicate> predicates = new ArrayList<>();
 		predicates.add(cb.isFalse(contact.get(CrmContact_.isDeleted)));
 
 		String searchKeyword = filterDto.getSearchKeyword();
 		if (searchKeyword != null && !searchKeyword.isBlank()) {
-			String escaped = StringUtils.escapeLikePattern(searchKeyword.trim().toLowerCase());
+			String escaped = StringUtils.escapeLikePattern(searchKeyword.trim().toLowerCase(Locale.ROOT));
 			String likePattern = "%" + escaped + "%";
-			Join<CrmContact, CrmCompany> company = contact.join(CrmContact_.company, JoinType.LEFT);
 			predicates.add(cb.or(cb.like(cb.lower(contact.get(CrmContact_.name)), likePattern, '\\'),
 					cb.like(cb.lower(company.get(CrmCompany_.name)), likePattern, '\\')));
 		}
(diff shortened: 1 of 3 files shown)


ThinuwanW (commented):
## 🤖 Claude Code Review

This is a small, well-scoped change that extends the contact lookup search to also match company name via a LEFT join, and updates the Swagger description accordingly. The logic is correct (LEFT join preserves companyless contacts, no row duplication since company is @ManyToOne, LIKE pattern properly escaped), but the new code introduces a redundant second join to crm_company and an inconsistent locale handling compared to its sibling metric method.

Found **4** new issue(s):

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
```diff
@@ -148,7 +148,10 @@ private List<Predicate> buildLookupPredicates(CriteriaBuilder cb, Root<CrmContac
 		String searchKeyword = filterDto.getSearchKeyword();
 		if (searchKeyword != null && !searchKeyword.isBlank()) {
 			String escaped = StringUtils.escapeLikePattern(searchKeyword.trim().toLowerCase());
-			predicates.add(cb.like(cb.lower(contact.get(CrmContact_.name)), "%" + escaped + "%", '\\'));
+			String likePattern = "%" + escaped + "%";
+			Join<CrmContact, CrmCompany> company = contact.join(CrmContact_.company, JoinType.LEFT);
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

Redundant/duplicate join to crm_company in the lookup query. findContactsForLookup already fetch-joins the company at line 130 (contact.fetch(CrmContact_.company, JoinType.LEFT)), but buildLookupPredicates creates a second, separate join here when a searchKeyword is present. Hibernate treats a fetch join and an explicit join() as distinct nodes, so the generated SQL will contain two LEFT JOINs on crm_company. It's functionally correct (company is @ManyToOne, so no row multiplication), but wasteful. The sibling findContacts/buildPredicates already handles this cleanly by casting the fetch to a Join and passing it down: `Join<CrmContact, CrmCompany> company = (Join<CrmContact, CrmCompany>) contact.fetch(...)`. Apply the same pattern here — cast the fetch join in findContactsForLookup and pass it into buildLookupPredicates rather than re-joining. (Note: the count query's join at line 105/164 is separate and fine.)

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
```diff
@@ -148,7 +148,10 @@ private List<Predicate> buildLookupPredicates(CriteriaBuilder cb, Root<CrmContac
 		String searchKeyword = filterDto.getSearchKeyword();
 		if (searchKeyword != null && !searchKeyword.isBlank()) {
 			String escaped = StringUtils.escapeLikePattern(searchKeyword.trim().toLowerCase());
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

Locale inconsistency with the sibling method. This line lowercases the keyword via searchKeyword.trim().toLowerCase() with no locale, whereas buildPredicates (line 86) uses toLowerCase(Locale.ROOT). Since this method now also drives company-name matching, use the same locale-independent form — searchKeyword.trim().toLowerCase(Locale.ROOT) — to avoid locale-dependent case folding (e.g. Turkish 'I'/'i') and keep the two search paths consistent. Minor/nit; this line is pre-existing but the new company search relies on it.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
```diff
@@ -148,7 +148,10 @@ private List<Predicate> buildLookupPredicates(CriteriaBuilder cb, Root<CrmContac
...
-			predicates.add(cb.like(cb.lower(contact.get(CrmContact_.name)), "%" + escaped + "%", '\\'));
+			String likePattern = "%" + escaped + "%";
```
ThinuwanW, line 151: 🤖 **Claude** · 💬 **Note**

Code duplication: the keyword-search construction (escapeLikePattern -> "%"+escaped+"%" -> cb.or(cb.like(cb.lower(name), ...))) is now near-identical to the block in buildPredicates (lines 86-90). Both build the same escaped like-pattern and OR a set of lower(...) LIKE clauses, differing only in which extra field(s) are searched (owner names vs company name). Consider extracting a shared helper, e.g. `Predicate keywordSearch(CriteriaBuilder cb, String keyword, Expression<String>... fields)`, that centralizes the trim/lowercase (with a single Locale), escaping, pattern-wrapping, and cb.or assembly. This also removes the risk of the two copies drifting further apart (as they already have with the Locale.ROOT difference).

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
```diff
@@ -148,7 +148,10 @@ private List<Predicate> buildLookupPredicates(CriteriaBuilder cb, Root<CrmContac
...
+			Join<CrmContact, CrmCompany> company = contact.join(CrmContact_.company, JoinType.LEFT);
+			predicates.add(cb.or(cb.like(cb.lower(contact.get(CrmContact_.name)), likePattern, '\\'),
```
ThinuwanW, line 152: 🤖 **Claude** · 💬 **Note**

Missing paired test change: this PR introduces new behavior (lookup now matches on company name via a LEFT join, and companyless contacts must still be returned/counted). There is no test covering findContactsForLookup / getContactsLookup / getLookupTotalCount (grep found none). Add a repository or integration test asserting: (a) a keyword matching only the company name returns the contact, (b) a keyword matching only the contact name still returns it, and (c) contacts with a null company are not dropped and are counted consistently between the data query and getLookupTotalCount. This guards the LEFT-join + OR semantics against future regressions.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
anusham-rc: Fixed, reused the existing fetch join by casting it and passing it into the predicate builder (same pattern as the contacts list query), so the generated SQL now has a single LEFT JOIN on crm_company; the count query keeps its own join as required.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
anusham-rc: Fixed, switched to toLowerCase(Locale.ROOT) to match the contacts list search path and avoid locale-dependent case folding.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
anusham-rc: keeping the two search blocks separate intentionally, since they target different field sets (owner names vs company name) and may evolve independently; both now use the same Locale.ROOT lowercasing and escaping, so the earlier drift is resolved.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
anusham-rc: Integration tests updated

sanjula-rtc (approved):

akilarootcode (approved):

ThinuwanW (approved):

PR #1590, opened 2026-07-06
fix: vadation added and update we when creating an contact

anusham-rc:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/frontend/src/community/crm/constants/contactConstants.ts
+++ b/frontend/src/community/crm/constants/contactConstants.ts
@@ -2,3 +2,6 @@ export const CONTACT_SEARCH_DEBOUNCE_DELAY = 500;
 export const DEFAULT_PAGE_SIZE = 10;
 export const DEFAULT_COMPANY_PAGE_SIZE = 100;
 export const ALL_COMPANIES = undefined;
+export const CONTACT_NAME_MAX_LENGTH = 255;
+export const CONTACT_EMAIL_MAX_LENGTH = 100;
+export const CONTACT_NUMBER_MAX_LENGTH = 15;
--- a/frontend/src/community/crm/regex/crmRegexPatterns.ts
+++ b/frontend/src/community/crm/regex/crmRegexPatterns.ts
@@ -3,5 +3,5 @@ export function isDealNameValid(): RegExp {
 }
 
 export function isContactNameValid(): RegExp {
-  return /^[\p{L} \-.,]+$/u;
+  return /^[\p{L}\p{N} \-.,']+$/u;
 }
--- a/frontend/src/community/crm/regex/crmRegexPatterns.ts
+++ b/frontend/src/community/crm/regex/crmRegexPatterns.ts
@@ -1,3 +1,7 @@
 export function isDealNameValid(): RegExp {
   return /^[\p{L}\p{N} \-.,&'()/:@#|]+$/u;
 }
+
+export function isContactNameValid(): RegExp {
+  return /^[\p{L} \-.,]+$/u;
+}
--- a/frontend/src/community/crm/utils/contactValidations.ts
+++ b/frontend/src/community/crm/utils/contactValidations.ts
@@ -26,8 +26,7 @@ export const addContactValidations = (translator: TranslatorFunctionType) =>
       .trim()
       .max(CONTACT_EMAIL_MAX_LENGTH, translator(["validations", "emailLength"]))
       .matches(isValidEmail(), {
-        message: translator(["validations", "invalidEmail"]),
-        excludeEmptyString: true
+        message: translator(["validations", "invalidEmail"])
       })
       .required(translator(["validations", "email"])),
     contactNumber: Yup.string()
--- a/frontend/src/community/crm/utils/contactValidations.ts
+++ b/frontend/src/community/crm/utils/contactValidations.ts
@@ -1,10 +1,14 @@
 import * as Yup from "yup";
 
-import { characterLengths } from "~community/common/constants/stringConstants";
 import {
   isValidEmail,
   isValidPhoneNumber
 } from "~community/common/regex/regexPatterns";
+import {
+  CONTACT_EMAIL_MAX_LENGTH,
+  CONTACT_NAME_MAX_LENGTH
+} from "~community/crm/constants/contactConstants";
+import { isContactNameValid } from "~community/crm/regex/crmRegexPatterns";
 
 type TranslatorFunctionType = (suffixes: string[]) => string;
 
@@ -13,31 +17,37 @@ export const addContactValidations = (translator: TranslatorFunctionType) =>
     name: Yup.string()
       .trim()
       .required(translator(["validations", "name"]))
-      .max(
-        characterLengths.NAME_LENGTH,
-        translator(["validations", "nameLength"])
-      ),
+      .max(CONTACT_NAME_MAX_LENGTH, translator(["validations", "nameLength"]))
+      .matches(isContactNameValid(), {
+        message: translator(["validations", "nameInvalidCharacters"]),
+        excludeEmptyString: true
+      }),
     email: Yup.string()
-      .trim()
-      .required(translator(["validations", "email"]))
-      .matches(isValidEmail(), translator(["validations", "invalidEmail"]))
-      .max(
-        characterLengths.CHARACTER_LENGTH,
-        translator(["validations", "characterLength"])
+      .test(
+        "email-required",
+        translator(["validations", "email"]),
+        (inputEmail) => Boolean(inputEmail?.trim())
+      )
+      .test(
+        "email-format",
+        translator(["validations", "invalidEmail"]),
+        (inputEmail) =>
+          !inputEmail?.trim() || isValidEmail().test(inputEmail.trim())
+      )
+      .test(
+        "email-max-length",
+        translator(["validations", "invalidEmail"]),
+        (inputEmail) =>
+          !inputEmail || inputEmail.length <= CONTACT_EMAIL_MAX_LENGTH
       ),
     contactNumber: Yup.string()
       .nullable()
       .optional()
-      .trim()
-      .max(
-        characterLengths.PHONE_NUMBER_LENGTH_MAX,
-        translator(["validations", "contactNumberLength"])
-      )
       .test(
         "valid-contact-number",
         translator(["validations", "contactNumber"]),
         function (inputContactNumber) {
-          if (!inputContactNumber) {
+          if (!inputContactNumber || inputContactNumber === "") {
             return true;
           }
 
--- a/frontend/src/community/crm/utils/contactValidations.ts
+++ b/frontend/src/community/crm/utils/contactValidations.ts
@@ -36,9 +36,10 @@ export const addContactValidations = (translator: TranslatorFunctionType) =>
       )
       .test(
         "email-max-length",
-        translator(["validations", "invalidEmail"]),
+        translator(["validations", "emailLength"]),
         (inputEmail) =>
-          !inputEmail || inputEmail.length <= CONTACT_EMAIL_MAX_LENGTH
+          !inputEmail?.trim() ||
+          inputEmail.trim().length <= CONTACT_EMAIL_MAX_LENGTH
       ),
     contactNumber: Yup.string()
       .nullable()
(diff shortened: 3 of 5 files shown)


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR refactors contact form validation to use CRM-specific length constants and a contact-name regex, replacing shared `characterLengths`. The frontend constraints (name regex, 255-char name, 15-char phone) correctly mirror the backend `CrmValidations`/`CrmConstants` source of truth, and removed translation keys are not referenced elsewhere. Two minor issues remain: a mislabeled/effectively-unreachable email length check and a redundant empty-string condition.

Found **7** new issue(s): 🟡 1 suggestion(s), 🟣 1 nit(s)

frontend/src/community/crm/utils/contactValidations.ts
```diff
@@ -13,31 +17,37 @@ export const addContactValidations = (translator: TranslatorFunctionType) =>
...
+      )
+      .test(
+        "email-format",
+        translator(["validations", "invalidEmail"]),
+        (inputEmail) =>
+          !inputEmail?.trim() || isValidEmail().test(inputEmail.trim())
+      )
+      .test(
+        "email-max-length",
+        translator(["validations", "invalidEmail"]),
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The `email-max-length` test surfaces the `invalidEmail` message ("Invalid email address") when the email exceeds CONTACT_EMAIL_MAX_LENGTH. A length violation showing a format error is misleading to users. Add a dedicated message key (e.g. `emailLength`: "Email is too long") in crmModule.json and reference it here. Note also this test is largely unreachable in practice because the email InputField already sets maxLength={CONTACT_EMAIL_MAX_LENGTH}, so it can double as defensive validation but should still carry the correct message. Additionally it checks the untrimmed `inputEmail.length` while the format test trims — consider trimming here too for consistency.

frontend/src/community/crm/utils/contactValidations.ts
```diff
@@ -13,31 +17,37 @@ export const addContactValidations = (translator: TranslatorFunctionType) =>
...
-      .max(
-        characterLengths.PHONE_NUMBER_LENGTH_MAX,
-        translator(["validations", "contactNumberLength"])
-      )
       .test(
         "valid-contact-number",
         translator(["validations", "contactNumber"]),
         function (inputContactNumber) {
-          if (!inputContactNumber) {
+          if (!inputContactNumber || inputContactNumber === "") {
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

The condition `if (!inputContactNumber || inputContactNumber === "")` is redundant — `!inputContactNumber` already evaluates to true for `""`, `null`, and `undefined`. Simplify to `if (!inputContactNumber?.trim())` (also handling whitespace-only input, since the old `.trim()` transform was removed from this field) or just `if (!inputContactNumber)`.

frontend/src/community/crm/regex/crmRegexPatterns.ts
```diff
@@ -1,3 +1,7 @@
 export function isDealNameValid(): RegExp {
   return /^[\p{L}\p{N} \-.,&'()/:@#|]+$/u;
 }
+
+export function isContactNameValid(): RegExp {
+  return /^[\p{L} \-.,]+$/u;
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

NEW behavioral restriction: the contact `name` field previously had no character validation (only required + maxLength). This PR adds `.matches(isContactNameValid())` with regex `^[\p{L} \-.,]+$` which rejects apostrophes and digits, so common legitimate names like "O'Brien", "D'Angelo", or "3M" now fail validation and cannot be saved. Note the sibling `isDealNameValid` regex deliberately includes `'` (and `\p{N}`). This mirrors the backend `CONTACT_NAME_REGEX`, so if the omission is unintended it must be fixed in both places — recommend adding `'` (and reconsider digits) to allow valid names.

frontend/src/community/crm/constants/contactConstants.ts
```diff
@@ -1,4 +1,7 @@
 export const CONTACT_SEARCH_DEBOUNCE_DELAY = 500;
 export const DEFAULT_PAGE_SIZE = 10;
 export const DEFAULT_COMPANY_PAGE_SIZE = 100;
-export const ALL_COMPANIES = undefined;
\ No newline at end of file
+export const ALL_COMPANIES = undefined;
+export const CONTACT_NAME_MAX_LENGTH = 255;
+export const CONTACT_EMAIL_MAX_LENGTH = 100;
```
ThinuwanW, line 6: 🤖 **Claude** · 💬 **Note**

`CONTACT_EMAIL_MAX_LENGTH = 100` is enforced on the frontend (input maxLength + `email-max-length` test) but there is NO corresponding backend/DB constraint — `CrmValidations.validateContactEmail` only checks blank + format, and the `email` column has no length limit. RFC 5321 permits addresses up to 254 chars, so this cap can reject valid emails that the backend would happily accept. Consider raising the limit to match a real backend/DB constraint (or add the matching backend validation) so the constraint is consistent and doesn't block legitimate input.

frontend/src/community/crm/constants/contactConstants.ts
```diff
@@ -1,4 +1,7 @@
...
+export const CONTACT_NUMBER_MAX_LENGTH = 15;
```
ThinuwanW, line 7: 🤖 **Claude** · 💬 **Note**

`CONTACT_NUMBER_MAX_LENGTH = 15` is used only for the input `maxLength`, while the actual contact-number Yup validation uses `isValidPhoneNumber()` which derives its max from a different constant (`characterLengths.PHONE_NUMBER_LENGTH_MAX`). Two independent sources of truth for the same limit (both 15 today) will silently drift if one is changed. Consider deriving the input maxLength from the same phone-length constant used by the regex, or documenting the coupling.

frontend/src/community/crm/utils/contactValidations.ts
```diff
@@ -13,31 +17,37 @@ export const addContactValidations = (translator: TranslatorFunctionType) =>
...
+        "email-format",
+        translator(["validations", "invalidEmail"]),
+        (inputEmail) =>
+          !inputEmail?.trim() || isValidEmail().test(inputEmail.trim())
+      )
+      .test(
+        "email-max-length",
+        translator(["validations", "invalidEmail"]),
+        (inputEmail) =>
+          !inputEmail || inputEmail.length <= CONTACT_EMAIL_MAX_LENGTH
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

Inconsistent trimming across the email tests: `email-required` and `email-format` operate on `inputEmail.trim()`, but `email-max-length` checks the raw `inputEmail.length`. Trailing/leading whitespace therefore counts toward the length limit here but is stripped elsewhere. Trim consistently (e.g. `inputEmail.trim().length`) so all three tests judge the same normalized value.

frontend/src/community/crm/utils/contactValidations.ts
```diff
@@ -1,10 +1,14 @@
 import * as Yup from "yup";
 
-import { characterLengths } from "~community/common/constants/stringConstants";
 import {
   isValidEmail,
   isValidPhoneNumber
 } from "~community/common/regex/regexPatterns";
+import {
```
ThinuwanW, line 7: 🤖 **Claude** · 💬 **Note**

📁 **File-level observation**

This file's validation logic was substantially rewritten (new name regex, restructured multi-test email validation, changed contact-number handling), yet there is no `contactValidations.test.ts`. The codebase has an established convention of unit-testing validation utils (e.g. `peopleValidations.test.ts`, `common/utils/__test__/validation.test.ts`). Add a test covering the new name-character rule, the email required/format/max-length precedence, and the optional contact-number cases to lock in behavior.

frontend/src/community/crm/utils/contactValidations.ts
anusham-rc: proper error message added

frontend/src/community/crm/utils/contactValidations.ts
anusham-rc: the regex now tests the trimmed value, matching what the submit handlers send.

frontend/src/community/crm/regex/crmRegexPatterns.ts
anusham-rc: regex updated accept apostrophes (ex: O'Brien), requirement confirmed

frontend/src/community/crm/constants/contactConstants.ts
anusham-rc: this is the requirement

frontend/src/community/crm/constants/contactConstants.ts
anusham-rc: this is the requirement

frontend/src/community/crm/utils/contactValidations.ts
anusham-rc: Fixed the max-length test now checks.

frontend/src/community/crm/utils/contactValidations.ts
anusham-rc: currently tests added only to utils

frontend/src/community/crm/utils/contactValidations.ts
```diff
@@ -13,36 +17,39 @@ export const addContactValidations = (translator: TranslatorFunctionType) =>
...
       .trim()
       .required(translator(["validations", "name"]))
-      .max(
-        characterLengths.NAME_LENGTH,
-        translator(["validations", "nameLength"])
-      ),
+      .max(CONTACT_NAME_MAX_LENGTH, translator(["validations", "nameLength"]))
+      .matches(isContactNameValid(), {
+        message: translator(["validations", "nameInvalidCharacters"]),
+        excludeEmptyString: true
```
akilarootcode, line 22: check this excludeEmptyString needed or not

frontend/src/community/crm/utils/contactValidations.ts
```diff
@@ -13,36 +17,39 @@ export const addContactValidations = (translator: TranslatorFunctionType) =>
...
-      .trim()
-      .required(translator(["validations", "email"]))
-      .matches(isValidEmail(), translator(["validations", "invalidEmail"]))
-      .max(
-        characterLengths.CHARACTER_LENGTH,
-        translator(["validations", "characterLength"])
+      .test(
+        "email-required",
+        translator(["validations", "email"]),
+        (inputEmail) => Boolean(inputEmail?.trim())
```
akilarootcode: check validation order

frontend/src/community/crm/utils/contactValidations.ts
```diff
@@ -13,36 +17,39 @@ export const addContactValidations = (translator: TranslatorFunctionType) =>
...
+      )
+      .test(
+        "email-format",
+        translator(["validations", "invalidEmail"]),
+        (inputEmail) =>
+          !inputEmail?.trim() || isValidEmail().test(inputEmail.trim())
```
akilarootcode: check trim() behavior. remove unnecessary trims

akilarootcode (changes requested):

frontend/src/community/crm/utils/contactValidations.ts
anusham-rc: order updated

frontend/src/community/crm/utils/contactValidations.ts
anusham-rc: removed unnecessary trims

frontend/src/community/crm/utils/contactValidations.ts
anusham-rc: without the flag, user who leaves the optional phone field empty gets "invalid contact number" and can't submit the form.

akilarootcode: check this again

anusham-rc: removed unnecessary excludeEmptyString check for email

akilarootcode (approved):

ThinuwanW (approved):

PR #1591, opened 2026-07-07
Add company name truncation with tooltip in table

YasinduRC:
### Summary

- 

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds a custom `render` to the company name column so the full name shows in a `title` tooltip on hover — a small, low-risk fix. The intent is correct and type-safe (`row.name` is a non-null string). The only concern is that the added `<span className="truncate">` uses the `truncate` utility on an inline element where it has no effect, and it diverges from the module's established truncation pattern.

Found **3** new issue(s): 🟡 1 suggestion(s)

frontend/src/community/crm/components/organisms/CompanyTable/CompanyTable.tsx
```diff
@@ -65,6 +65,13 @@ export const CompanyTable: FC = () => {
       header: translateText(["table", "columns", "nameHeader"]),
       key: "name",
       className: "truncate",
+      render(_value, row) {
+        return (
+          <span className="truncate" title={row.name}>
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The `truncate` Tailwind utility (overflow:hidden; text-overflow:ellipsis; white-space:nowrap) has no visual effect on a default inline `<span>` — those properties require a block/inline-block element with a constrained width. The established pattern in the same module (DealsTable.tsx lines 169, 190, 198) is `className="body2 block w-full truncate"`. The cell still truncates via the column-level `className: "truncate"` on line 67, so the tooltip goal is met, but the span's `truncate` class is dead/misleading. Suggest changing to `className="body2 block w-full truncate"` (or at least `block w-full truncate`) to match the sibling tables and make truncation on the span itself actually functional.

frontend/src/community/crm/components/organisms/CompanyTable/CompanyTable.tsx
```diff
@@ -65,6 +65,13 @@ export const CompanyTable: FC = () => {
       header: translateText(["table", "columns", "nameHeader"]),
       key: "name",
       className: "truncate",
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

Duplicated/competing truncation declarations. The column already carries `className: "truncate"` (line 67), which is the established cell-level truncation pattern used elsewhere (e.g. CustomerContactTable applies `className: "truncate"` on the column and no render). This PR adds a second `truncate` on an inner `<span>` (line 70). You now have two truncation mechanisms on the same cell, and it's ambiguous which one is meant to win — one of them is effectively dead. Pick a single mechanism: either keep only the column `className: "truncate"` (but then you can't attach the `title` tooltip), or drop the column `className` and make the span actually truncate with `className="block truncate"` plus a bounded width. Leaving both is misleading and reads as an incomplete change.

frontend/src/community/crm/components/organisms/CompanyTable/CompanyTable.tsx
```diff
@@ -65,6 +65,13 @@ export const CompanyTable: FC = () => {
...
+      render(_value, row) {
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

The render signature marks the first param unused (`_value`) but then reads `row.name` — which, because `key` is `"name"`, is the exact same value the cell already receives. This diverges from the sibling ContactTable name column which uses `render(value, row)` and renders `value` directly. Prefer `render(value)` and use `value` for consistency and to avoid the awkward unused-param + property-lookup pattern.

frontend/src/community/crm/components/organisms/CompanyTable/CompanyTable.tsx
YasinduRC: Removed the the cell level turncate since the pattern of the codebase is to use truncate inside the span

frontend/src/community/crm/components/organisms/CompanyTable/CompanyTable.tsx
YasinduRC: fixed. Changed render(_value, row) to render(value) and now render value directly instead of reaching back into row.name.

frontend/src/community/crm/components/organisms/CompanyTable/CompanyTable.tsx
YasinduRC: Fixed, changed className="truncate" to className="body2 block w-full truncate" so it's actually a block-level element with truncate taking effect, also matching with the pattern used in the codebase

thusala (approved):

akilarootcode (approved):

ThinuwanW (approved):

PR #1593, opened 2026-07-07
feat: Deal Sidepanel Delete and Update Bugs,

YasinduRC:
…t messages

### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR implements delete-deal functionality by adding a `useDeleteDeal` React Query mutation hook, wiring it into `DeleteDealModal` with success/error toasts and loading/disabled states, and passing `dealId` down from `DealSidePanel`. The implementation is clean and mirrors the existing `useDeleteDealStage`/`useCreateDeal` patterns closely, including query-key invalidation and i18n. Only minor, forward-looking observations were found.

Found **5** new issue(s): 🟡 3 suggestion(s), 🟣 2 nit(s)

frontend/src/community/crm/api/crmDealApi.ts
```diff
@@ -231,6 +231,25 @@ export const useDeleteDealStage = (
...
+const deleteDeal = async (id: number): Promise<void> => {
+  await authFetch.delete(crmDealEndpoints.DELETE_DEAL(id));
+};
+
+export const useDeleteDeal = (onSuccess: () => void, onError: () => void) => {
+  const queryClient = useQueryClient();
+  return useMutation({
+    mutationFn: deleteDeal,
+    onSuccess: () => {
+      queryClient.invalidateQueries({ queryKey: crmDealQueryKeys.ALL });
```
ThinuwanW, line 281: 🤖 **Claude** · 🟡 **Suggestion**

On delete success only `crmDealQueryKeys.ALL` (['crm-deals']) and the CRM limitation key are invalidated. The Kanban board list uses a separate key (`crmBoardQueryKeys.GET_DEALS_GROUPED_BY_STAGES` / ['crm-board-deals-grouped-by-stages']) which is NOT invalidated. Today this is not a live bug because the delete modal is only reachable from the list/table view (the Kanban card `onDealClick` in DealsKanbanBoard.tsx is a no-op `() => {}`, and the board query is disabled while in list view so it refetches on view switch). However, once Kanban cards are wired to open the DealSidePanel, deleting a deal will leave a stale card on the board. Consider also invalidating the board query key here (mirrors the same latent gap in `useCreateDeal`).

frontend/src/community/crm/components/molecules/DeleteDealModal/DeleteDealModal.tsx
```diff
@@ -1,20 +1,60 @@
...
+    setSelectedDealId: store.setSelectedDealId,
+    closeCrmSidePanel: store.closeCrmSidePanel
+  }));
+
+  const handleSuccess = (): void => {
+    setToastMessage({
+      open: true,
+      toastType: ToastType.SUCCESS,
+      title: translateText(["toastMessages", "successTitle"]),
+      description: translateText(["toastMessages", "successDescription"], {
```
ThinuwanW, line 33: 🤖 **Claude** · 🟣 **Nit**

The success toast interpolates `dealName`, which comes from `deal?.name ?? ""` in DealSidePanel. The kebab menu (and thus the delete action) is rendered even while `useGetDealById` is still loading (headerActions sits outside the `isLoading` branch), so if a user deletes before the deal detail loads, `dealName` is "" and the toast reads " has been deleted successfully." Minor edge case; consider falling back to a generic phrasing when the name is empty.

frontend/src/community/crm/components/organisms/DealSidePanel/DealSidePanel.tsx
```diff
@@ -142,6 +142,7 @@ const DealSidePanel: FC = () => {
       <DeleteDealModal
         isOpen={isDeleteModalOpen}
         onClose={() => setIsDeleteModalOpen(false)}
+        dealId={selectedDealId!}
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`dealId={selectedDealId!}` uses a non-null assertion to force a nullable store value into `number`, and `handleDeleteDeal` in DeleteDealModal then calls `deleteDeal(dealId)` unconditionally. If `selectedDealId` is ever null when the modal fires, this issues DELETE `/deal/null`. The sibling modals avoid this: DeleteTaskModalContent reads `selectedTaskId` from the store and guards with `if (selectedTaskId === null) return;` before mutating, and DeleteCompanyModalContent reads `selectedCompany.id` from the store. Fix: either read `selectedDealId` from the store inside DeleteDealModal (it already subscribes to the store) and guard for null before calling `deleteDeal`, or add a `if (dealId == null) return;` guard in `handleDeleteDeal` instead of relying on the `!` assertion. Guidelines flag `!`/`as` assertions used to silence nullability rather than handle it.

frontend/src/community/crm/components/molecules/DeleteDealModal/DeleteDealModal.tsx
```diff
@@ -1,20 +1,60 @@
 import { SmallModal } from "@rootcodelabs/skapp-ui";
 
 import Icon from "~community/common/components/atoms/Icon/Icon";
+import { ToastType } from "~community/common/enums/ComponentEnums";
```
ThinuwanW, line 4: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

This modal re-implements the delete/cancel button markup (variant, DeleteButtonIcon, iconPosition, isLoading/disabled=isPending) directly via SmallModal's `buttons` config, duplicating what the shared `CrmDeleteModalContent` presenter already provides and what DeleteCompanyModalContent/DeleteTaskModalContent reuse. It also diverges architecturally by taking `dealId`/`dealName` as props while its siblings read the selected entity from the store (source of truth) as self-contained `FC` components. Consider aligning with the established pattern — read the deal id/name from the store and/or render through CrmDeleteModalContent — to keep the delete flows consistent and reduce duplicated button markup.

frontend/src/community/crm/api/crmDealApi.ts
ThinuwanW, line 281: 🤖 **Claude** · 🟣 **Nit**

On successful delete only `crmDealQueryKeys.ALL` (`["crm-deals"]`) and the CRM limitation key are invalidated. The individual deal-detail cache uses `DEAL_BY_ID(id)` = `["crm-deal", id]` (singular "crm-deal"), which is NOT a prefix match of `["crm-deals"]`, so the deleted deal's detail query remains cached (stale). Since the side panel closes on success this is low-impact today, but it leaves orphaned cache. Consider `queryClient.removeQueries({ queryKey: crmDealQueryKeys.DEAL_BY_ID(id) })` in the mutation (the id is available via the `deleteDeal` variable / onSuccess args) to fully clean up.

YasinduRC: Added invalidateQueries for the board key in both useDeleteDeal and useCreateDeal

YasinduRC: Since removeQueries is not a practice we use, so it moved it to  ["crm-deals"] name space, and now it will invalidated by the iinvalidateQueries({ queryKey: ALL })

frontend/src/community/crm/components/molecules/DeleteDealModal/DeleteDealModal.tsx
YasinduRC: Used CrmDeleteModalContent   and to read selectedDealId from useCrmStore internally and the deal name is passed as the prop, for the modal

frontend/src/community/crm/components/molecules/DeleteDealModal/DeleteDealModal.tsx
YasinduRC: added the skeletons for the for header and headerActions  because other sidepanels also used the same pattern

frontend/src/community/crm/components/organisms/DealSidePanel/DealSidePanel.tsx
YasinduRC: added thye guard to selectedID

PR #1595, opened 2026-07-07
Fix/ Job family job title not filtering issue

gaindurootcode:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/backend/src/main/java/com/skapp/community/peopleplanner/service/impl/JobServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/peopleplanner/service/impl/JobServiceImpl.java
@@ -60,6 +60,7 @@ public class JobServiceImpl implements JobService {
 		log.info("getAllJobFamilies: execution started");
 
 		List<JobFamily> jobFamilies = jobFamilyDao.getJobFamiliesByEmployeeCount();
+		jobFamilies.forEach(jobFamily -> jobFamily.setJobTitles(filterActiveJobTitles(jobFamily.getJobTitles())));
 
 		List<JobFamilyResponseDetailDto> jobFamilyResponseDetailDtos = peopleMapper
 			.jobFamilyListToJobFamilyResponseDetailDtoList(jobFamilies);


ThinuwanW (commented):
## 🤖 Claude Code Review

The PR adds a single line to filter out inactive job titles from the job-titles collection when listing all job families. This is a legitimate fix — the underlying `getJobFamiliesByEmployeeCount` repository query uses a non-fetch join with an `isActive` on-clause that does NOT filter the lazily-loaded collection, so filtering had to be done in the service layer (mirroring the existing enterprise `EpJobServiceImpl.getJobs()` pattern). The change is correct and safe under the read-only transaction, but it introduces a behavioral inconsistency with the sibling `getJobFamilyById` endpoint, which still returns inactive titles.

Found **3** new issue(s): 🔴 1 important, 🟡 2 suggestion(s)

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/JobServiceImpl.java
```diff
@@ -60,6 +60,7 @@ public ResponseEntityDto getAllJobFamilies() {
 		log.info("getAllJobFamilies: execution started");
 
 		List<JobFamily> jobFamilies = jobFamilyDao.getJobFamiliesByEmployeeCount();
+		jobFamilies.forEach(jobFamily -> jobFamily.setJobTitles(filterActiveJobTitles(jobFamily.getJobTitles())));
```
ThinuwanW, line 63: 🤖 **Claude** · 🟡 **Suggestion**

Mutating a managed entity's collection purely to shape the response is fragile. `jobTitles` is a `@ManyToMany(cascade = CascadeType.ALL)` (JobFamily line 41); replacing it with a filtered subset marks the collection dirty, so removed (inactive) titles would be deleted from the `job_family_title` join table on flush. This is safe today only because the method is `@Transactional(readOnly = true)` (flush disabled), but the safety is implicit and would break silently if `readOnly` were ever removed or an explicit flush occurred. Consider filtering at the mapper/DTO layer instead of mutating the persistent entity.

ThinuwanW, line 63: 🤖 **Claude** · 🔴 **Important**

Behavioral asymmetry / missing paired change: `getAllJobFamilies` now filters out inactive job titles (line 63), but `getJobFamilyById` (lines 74-87) returns the same `JobFamilyResponseDetailDto` WITHOUT filtering. Its query `getJobFamilyByIdWithJobTitles` fetch-joins ALL job titles (JobFamilyRepositoryImpl.java:151 fetches jobTitles with no is_active predicate) and maps them directly. Result: the list endpoint hides soft-deleted (isActive=false) titles while the detail-by-id endpoint leaks them — inconsistent API behavior for the same DTO. Since `addJobFamily` (line 124) and `updateJobFamily` (line 196) also filter active titles, hiding inactive titles is clearly the intended contract, so `getJobFamilyById` is the outlier. Fix: apply `filterActiveJobTitles(jobFamily.getJobTitles())` in `getJobFamilyById` before mapping (or add an is_active predicate to the fetch join).

ThinuwanW, line 63: 🤖 **Claude** · 🟡 **Suggestion**

Missing test coverage for the new behavior. `JobServiceImplUnitTest` has no test referencing `getAllJobFamilies` or the active-title filtering, so this behavioral change (excluding isActive=false titles from the response) is unverified. Add a unit test that returns a JobFamily containing both active and inactive JobTitle entities and asserts only active titles appear in the mapped `JobFamilyResponseDetailDto`.

gaindurootcode: Known issue. Out of scope of this PR

gaindurootcode: Since the requirement do not specify this fltering on the getJobFamilyById  we will not fix it.

gaindurootcode: We do not write test cases for ow. Not needed

akilarootcode (approved):

ThinuwanW (approved):

ThinuwanW (approved):

ThinuwanW (approved):

PR #1597, opened 2026-07-07
Feat/API integration for deal detailed view

YasinduRC:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR wires up inline editing for the deal side panel (title, description, amount, stage, priority, owner, contact) via a new useEditDeal mutation, with shared field validators and toast feedback. The mutation, query-key invalidation, types, and i18n are all correctly structured. However, the amount field's onChange wiring breaks discard/escape semantics (unsaved values persist visually), and PropertyField commits saves even when the value is unchanged, causing redundant PATCH requests and misleading success toasts.

Found **8** new issue(s): 🔴 1 important, 🟡 2 suggestion(s)

frontend/src/community/crm/components/organisms/DealSidePanel/DealPropertiesSidebar.tsx
```diff
@@ -98,18 +112,32 @@ const DealPropertiesSidebar: FC<DealPropertiesSidebarProps> = ({
           label={translateText(["value"])}
           value={amount}
           placeholder={translateText(["placeholders", "none"])}
+          validate={(value) => validateDealAmount(value, translateText)}
           onChange={setAmount}
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

`onChange={setAmount}` on the amount PropertyField updates the parent `amount` state on every keystroke. Because PropertyField feeds `amount` back in as its `value` prop and syncs `inputValue` from it via useEffect, pressing Escape (or clicking away after a failed commit) resets `inputValue` to `value` — which has already been mutated to the typed-but-unsaved text. Result: discarding an amount edit leaves the edited (unsaved) value displayed and in local state until a refetch, diverging from the backend. Title/Description avoid this by keeping edit state internal and only calling `onSave`. Remove `onChange={setAmount}` and rely solely on `onSave` (which already calls `setAmount(value)`).

frontend/src/community/crm/components/molecules/PropertyField/PropertyField.tsx
```diff
@@ -42,26 +44,47 @@ const PropertyField: FC<PropertyFieldProps> = ({
...
+  const commit = (nextValue: string): boolean => {
+    if (validate) {
+      const error = validate(nextValue);
+      if (error) {
+        setValidationError(error);
+        return false;
+      }
+    }
+    setValidationError(undefined);
+    if (onSave) {
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

`commit` calls `onSave(nextValue)` unconditionally, with no dirty check. Because click-outside also invokes `commit`, simply focusing the amount field and clicking away (without changing anything) fires a PATCH request and shows the "Deal updated" success toast for a no-op edit. DealTitleSection/DealDescriptionSection guard with `if (trimmed !== original)`. Add an equivalent dirty check here (e.g. skip `onSave` when `nextValue === value`) so unchanged values don't trigger network writes. The same lack of a dirty check affects the stage/priority/owner/contact handlers in DealPropertiesSidebar — reselecting the current value will also fire a redundant update.

frontend/src/community/crm/api/crmDealApi.ts
```diff
@@ -130,6 +131,41 @@ export const useGetDealById = (
...
+};
+
+export const useEditDeal = (
+  onSuccess: () => void,
+  onError: (error: AxiosError) => void
+) => {
+  const queryClient = useQueryClient();
+  return useMutation({
+    mutationFn: editDeal,
+    onSuccess: ({ id }) => {
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

`onSuccess: ({ id })` destructures the mutation result, but `editDeal` returns `response?.data?.results?.[0]` which is typed as CrmDealDetailResponseType yet can be `undefined` at runtime (empty/absent results array). Destructuring `id` from `undefined` would throw inside onSuccess. Prefer the mutation `variables` for the id, which is always available: `onSuccess: (_data, variables) => { ... crmDealQueryKeys.DEAL_BY_ID(variables.id) ... }`. This is both safer and guarantees the correct deal detail cache is invalidated even if the response body is empty.

frontend/src/community/crm/components/organisms/DealSidePanel/DealPropertiesSidebar.tsx
```diff
@@ -98,18 +112,32 @@ const DealPropertiesSidebar: FC<DealPropertiesSidebarProps> = ({
...
+          onSave={(value) => {
+            setAmount(value);
+            onAmountChange(value);
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

When the amount is cleared, `onSave("")` → `onAmountChange("")` → `updateDeal({ amount: "" })` sends an empty string. `CrmDealEditFields.amount` is typed `string | null`, so clearing likely should send `null` rather than `""` to unset the value. Confirm the backend accepts an empty string for clearing amount; if not, convert empty input to `null` before calling onAmountChange.

frontend/src/community/crm/components/organisms/DealSidePanel/DealSidePanel.tsx
```diff
@@ -133,7 +167,19 @@ const DealSidePanel: FC = () => {
                   />
                 </div>
               </div>
-              <DealPropertiesSidebar deal={deal!} isOpen={isOpen} />
+              <DealPropertiesSidebar
+                deal={deal!}
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The panel only branches on `isLoading`; when the GET (`useGetDealById`) errors or resolves without a deal (e.g. deal not found / deleted), `isLoading` is false and `deal` is `undefined`, yet `DealPropertiesSidebar` is rendered with `deal={deal!}`. That component immediately dereferences `deal.amount`, `deal.priority`, `deal.stageId`, `deal.owner`, `deal.contact` (DealPropertiesSidebar.tsx:48-55), so the whole panel throws at runtime. Note the title/description sections already guard with `deal?.name ?? ""`, making the unguarded sidebar an inconsistency. There are even unused `errors.dealNotFoundTitle`/`dealNotFoundDescription` translation keys that suggest a not-found state was intended but never wired up. Fix: render an error/not-found state when `isError || !deal` (using those translation keys) before rendering the sidebar, and remove the `deal!` assertion.

frontend/src/community/crm/components/organisms/DealSidePanel/DealSidePanel.tsx
```diff
@@ -46,8 +49,33 @@ const DealSidePanel: FC = () => {
     closeCrmSidePanel();
   };
 
+  const { setToastMessage } = useToast();
+
   const { data: deal, isLoading } = useGetDealById(selectedDealId!);
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`useGetDealById(selectedDealId!)` is called with a non-null assertion, but `selectedDealId` is `null` whenever no deal is selected (panel closed). `useGetDealById` has no `enabled` flag (crmDealApi.ts:124-132), so the query fires with `id === null`, producing a query key `["crm-deal", null]` and a request to `/deal/null`. Add an `enabled: id != null` (or pass `selectedDealId` and gate with `enabled: !!selectedDealId`) to the query and drop the `!` assertion, per the guideline that queries depending on runtime state must set `enabled`.

frontend/src/community/crm/components/organisms/DealSidePanel/DealDescriptionSection.tsx
```diff
@@ -2,33 +2,55 @@ import { ButtonV2, TextArea } from "@rootcodelabs/skapp-ui";
 import { FC, KeyboardEventHandler, useState } from "react";
 
 import { useTranslator } from "~community/common/hooks/useTranslator";
+import { validateDealDescription } from "~community/crm/utils/dealValidations";
```
ThinuwanW, line 6: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

`DealTitleSection` and `DealDescriptionSection` implement a near-identical inline-edit state machine (`isEditing`/`edited*`/`error` state, `handleClick`/`handleChange`/`handleSave`/`handleDiscard`, key handlers, dirty-vs-original comparison, validate-on-change). `PropertyField` is yet a third variant of the same pattern. Per the duplication guideline, extract the shared inline-edit logic into a reusable hook (e.g. `useInlineEdit`) or component to avoid divergence between the title/description/property edit flows.

frontend/src/community/crm/components/organisms/DealSidePanel/DealSidePanel.tsx
YasinduRC: to properly handling isError/deal-not-found needs its own dedicated view/empty state (it is edge case the we faced in all 4 sidepanels) so we've deferred it rather than half-wiring it here. Also i removed the misleading translation keys

frontend/src/community/crm/components/organisms/DealSidePanel/DealSidePanel.tsx
YasinduRC: /deal/null can't happen in here, DealSidePanel is conditionally mounted with selectedDealId !== nullso it only exists while a deal is selected. On close we null out selectedDealId, which unmounts the panel. Kept assertion because SelectedDealId cannot be null

frontend/src/community/crm/components/organisms/DealSidePanel/DealPropertiesSidebar.tsx
YasinduRC: Removed the onChange and rely only on onSave

frontend/src/community/crm/api/crmDealApi.ts
YasinduRC: Since the backend cannot return undefined in any case, so destructing from the response, also this is the same pattern followed by other panels

frontend/src/community/crm/components/organisms/DealSidePanel/DealPropertiesSidebar.tsx
YasinduRC: Fixed added a backend change if we pass a empty string backend will check and set it to null

frontend/src/community/crm/components/molecules/PropertyField/PropertyField.tsx
YasinduRC: added guards with if conditions to check the if the values are changed from the existing state

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
```diff
@@ -413,7 +413,7 @@ public ResponseEntityDto editDeal(Long id, CrmDealEditRequestDto requestDto) {
 
 		if (requestDto.getAmount() != null) {
 			CrmValidations.validateDealAmount(requestDto.getAmount());
-			deal.setAmount(requestDto.getAmount());
+			deal.setAmount(requestDto.getAmount().isBlank() ? null : requestDto.getAmount());
```
akilarootcode: remove empty check

frontend/src/community/common/assets/languages/english/crmModule.json
```diff
@@ -443,9 +443,18 @@
...
-        "dealNotFoundDescription": "This deal is no longer available."
+      "toastMessages": {
+        "editSuccessTitle": "Deal updated",
+        "editSuccessDescription": "Your changes have been saved.",
+        "editErrorTitle": "Failed to update deal",
+        "editErrorDescription": "Something went wrong. Please try again"
+      },
+      "validations": {
+        "dealNameRequired": "Deal name is required",
+        "dealNameInvalidChars": "Deal name can only contain letters, numbers, spaces, and - . , & ' ( ) / : @ # |",
```
akilarootcode, line 456: check with the requirement

frontend/src/community/crm/api/crmDealApi.ts
```diff
@@ -121,12 +122,45 @@ const fetchDealById = async (
...
+    enabled
+  });
+};
+
+const editDeal = async ({
+  id,
+  fields
+}: {
+  id: number;
+  fields: CrmDealEditFields;
```
akilarootcode: use interface

frontend/src/community/crm/api/crmDealApi.ts
```diff
@@ -121,12 +122,45 @@ const fetchDealById = async (
...
+  );
+  return response?.data?.results?.[0];
+};
+
+export const useEditDeal = (onError: (error: AxiosError) => void) => {
+  const queryClient = useQueryClient();
+  return useMutation({
+    mutationFn: editDeal,
+    onSuccess: ({ id }) => {
+      queryClient.invalidateQueries({ queryKey: crmDealQueryKeys.ALL });
```
akilarootcode: invalidation of ALL is wrong practice. check with team

frontend/src/community/crm/api/crmDealApi.ts
```diff
@@ -121,12 +122,45 @@ const fetchDealById = async (
...
+    fields
+  );
+  return response?.data?.results?.[0];
+};
+
+export const useEditDeal = (onError: (error: AxiosError) => void) => {
+  const queryClient = useQueryClient();
+  return useMutation({
+    mutationFn: editDeal,
+    onSuccess: ({ id }) => {
```
akilarootcode: Add type

frontend/src/community/crm/components/molecules/PropertyField/PropertyField.tsx
```diff
@@ -21,7 +21,7 @@ interface PropertyFieldProps {
   inputType?: PropertyFieldInputType;
   min?: number;
   max?: number;
-  errorMessage?: string;
+  validate?: (value: string) => string | undefined;
```
akilarootcode: remove undefined

frontend/src/community/crm/components/molecules/PropertyField/PropertyField.tsx
```diff
@@ -42,26 +42,45 @@ const PropertyField: FC<PropertyFieldProps> = ({
     placeholder ?? translateText(["placeholders", "none"]);
   const [isEditing, setIsEditing] = useState<boolean>(false);
   const [inputValue, setInputValue] = useState<string>(value);
+  const [validationError, setValidationError] = useState<string | undefined>(
```
akilarootcode: remove undefined with empty string

frontend/src/community/crm/components/molecules/PropertyField/PropertyField.tsx
```diff
@@ -42,26 +42,45 @@ const PropertyField: FC<PropertyFieldProps> = ({
...
+    undefined
+  );
   const inputRef = useRef<HTMLDivElement>(null);
 
   useEffect(() => {
     setInputValue(value);
   }, [value]);
 
+  const commit = (nextValue: string): boolean => {
```
akilarootcode: rename the variable

frontend/src/community/crm/components/molecules/PropertyField/PropertyField.tsx
```diff
@@ -42,26 +42,45 @@ const PropertyField: FC<PropertyFieldProps> = ({
...
+    if (validate) {
+      const error = validate(nextValue);
```
akilarootcode: rename as errorMessage

frontend/src/community/crm/components/molecules/PropertyField/PropertyField.tsx
```diff
@@ -42,26 +42,45 @@ const PropertyField: FC<PropertyFieldProps> = ({
...
+      if (error) {
+        setValidationError(error);
+        return false;
+      }
+    }
+    setValidationError(undefined);
+    if (nextValue !== value) {
```
akilarootcode: check the diff first

frontend/src/community/crm/components/molecules/PropertyField/PropertyField.tsx
```diff
@@ -42,26 +42,45 @@ const PropertyField: FC<PropertyFieldProps> = ({
...
 
+  const commit = (nextValue: string): boolean => {
+    if (validate) {
+      const error = validate(nextValue);
+      if (error) {
+        setValidationError(error);
+        return false;
+      }
+    }
+    setValidationError(undefined);
```
akilarootcode: use formik

akilarootcode (changes requested):

frontend/src/community/crm/api/crmDealApi.ts
YasinduRC: moved to a interface

frontend/src/community/crm/api/crmDealApi.ts
YasinduRC: Removed now it uses the store

frontend/src/community/crm/api/crmDealApi.ts
YasinduRC: Removed since the query keys are remvoed

frontend/src/community/crm/components/molecules/PropertyField/PropertyField.tsx
YasinduRC: Removed in the refactor

frontend/src/community/crm/components/molecules/PropertyField/PropertyField.tsx
YasinduRC: Removed in the refactor

frontend/src/community/crm/components/molecules/PropertyField/PropertyField.tsx
YasinduRC: Removed in the refactor

frontend/src/community/crm/components/molecules/PropertyField/PropertyField.tsx
YasinduRC: Removed in the refactor

frontend/src/community/crm/components/molecules/PropertyField/PropertyField.tsx
YasinduRC: Removed in the refactor

frontend/src/community/crm/components/molecules/PropertyField/PropertyField.tsx
YasinduRC: Used formic and refactor

ThinuwanW (commented):
## 🤖 Claude Code Review

This PR wires up inline deal editing (backend edit mapping + frontend inline-edit hook, store, and validation). The core logic is sound and types are largely consistent, but the `useGetDealsInfinite` return type was changed from `UseInfiniteQueryResult` to `UseQueryResult` which breaks the type contract with its consumer, and there is no rollback of local UI state when an edit mutation fails.

Found **11** new issue(s): 🔴 1 important, 🟡 2 suggestion(s), 🟣 3 nit(s)

frontend/src/community/crm/api/crmDealApi.ts
```diff
@@ -29,7 +30,7 @@ import { crmDealQueryKeys } from "./utils/QueryKeys";
 export const useGetDealsInfinite = (
   params: CrmDealFilterParams,
   enabled: boolean = true
-): UseInfiniteQueryResult<CrmDealPaginatedResponse> => {
+): UseQueryResult<CrmDealPaginatedResponse> => {
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`useGetDealsInfinite` still returns `useInfiniteQuery(...)` but its declared return type was changed to `UseQueryResult<CrmDealPaginatedResponse>`. `UseQueryResult` does not expose `fetchNextPage`, `hasNextPage`, or `isFetchingNextPage`, and its `data` is `CrmDealPaginatedResponse` (no `.pages`). The sole consumer `DealsSection.tsx` destructures `fetchNextPage`, `hasNextPage`, `isFetchingNextPage` and reads `data?.pages.flatMap(...)`, all of which will fail type-checking (or is at minimum an incorrect/misleading annotation). Revert to `UseInfiniteQueryResult<CrmDealPaginatedResponse>` and re-add the removed import.

frontend/src/community/crm/components/organisms/DealSidePanel/DealSidePanel.tsx
```diff
@@ -42,7 +53,30 @@ const DealSidePanel: FC = () => {
...
+
+  useEffect(() => {
+    if (dealDetail) {
+      updateDealInStore(dealDetail);
+    }
+  }, [dealDetail]);
+
+  const selectedDeal = getDealById(selectedDealId!);
+
+  const { mutate: editDeal } = useEditDeal(() => {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

On edit failure only an error toast is shown; there is no rollback. `DealPropertiesSidebar` mutates its local state (amount/priority/stage/owner/contact) before/independently of server confirmation, so when `useEditDeal` fails the sidebar continues to display the unsaved value while the store/backend still hold the old value, misrepresenting what was persisted. Consider reverting the optimistic local state (or driving the fields off the store deal) inside the `onError` path, or making the sidebar fields controlled by the store `selectedDeal`.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmDealRepositoryImpl.java
```diff
@@ -47,6 +48,12 @@ public class CrmDealRepositoryImpl implements CrmDealRepository {
 
 	private final EntityManager entityManager;
 
+	private Expression<BigDecimal> safeAmountAsDecimal(CriteriaBuilder cb, Path<String> amount) {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`safeAmountAsDecimal` correctly guards the empty-string/null `amount` -> BigDecimal cast, but the same unguarded `deal.get(CrmDeal_.amount).cast(BigDecimal.class)` remains in `CrmCompanyRepositoryImpl` (lines ~69, 76) and `CrmContactRepositoryImpl` (line ~70). Since this PR now lets users edit a deal's amount (including to empty), those aggregation queries can still throw on an empty amount. Apply the same safe-cast helper there for consistency, or extract it to a shared util.

frontend/src/community/crm/components/organisms/DealSidePanel/DealSidePanel.tsx
```diff
@@ -42,7 +53,30 @@ const DealSidePanel: FC = () => {
...
-  const { data: deal, isLoading } = useGetDealById(selectedDealId!);
+  const { setToastMessage } = useToast();
+
+  const { data: dealDetail } = useGetDealById(selectedDealId!);
+
+  useEffect(() => {
+    if (dealDetail) {
+      updateDealInStore(dealDetail);
+    }
+  }, [dealDetail]);
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`useEffect(..., [dealDetail])` omits `updateDealInStore` from its dependency array (and `DealsSection.tsx:54` omits `setDeals` from `[allDeals]`). These are stable Zustand setters so behavior is unaffected, but `react-hooks/exhaustive-deps` will flag them — add the setters to the deps arrays.

frontend/src/community/crm/api/crmDealApi.ts
```diff
@@ -118,18 +119,46 @@ export const useGetDealLookup = (
...
+): Promise<CrmDealResponseType> => {
   const response = await authFetch.get(crmDealEndpoints.GET_DEAL_BY_ID(id));
   return response?.data?.results?.[0];
 };
 
 export const useGetDealById = (
-  id: number
-): UseQueryResult<CrmDealDetailResponseType> => {
+  id: number,
+  enabled: boolean = true
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

The new `enabled` parameter added to `useGetDealById` is never supplied by the only caller (`DealSidePanel` passes just the id). It's harmless because the panel only mounts when `selectedDealId !== null`, but the parameter is currently dead — either wire it up (e.g. `enabled: selectedDealId != null`) as an explicit guard, or drop it.

frontend/src/community/common/assets/languages/english/crmModule.json
```diff
@@ -447,9 +447,18 @@
         "save": "Save",
         "discard": "Discard"
       },
-      "errors": {
-        "dealNotFoundTitle": "Deal not found",
-        "dealNotFoundDescription": "This deal is no longer available."
+      "toastMessages": {
+        "editSuccessTitle": "Deal updated",
```

`toastMessages.editSuccessTitle`/`editSuccessDescription` are added but never used — only the error toast is shown after an edit. Either surface a success toast in `useEditDeal`/`DealSidePanel` (consistent with the create flow which toasts on success) or remove the unused keys.

frontend/src/community/crm/components/molecules/PropertyField/PropertyField.tsx
```diff
@@ -33,20 +25,24 @@ const PropertyField: FC<PropertyFieldProps> = ({
...
-  }, [value]);
+  const {
+    isEditing,
+    value: editedValue,
+    error,
+    startEditing,
+    changeValue,
+    save,
+    discard
+  } = useInlineEditForm({ value, validate, onSave, onChange });
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

`useInlineEditForm({ value, validate, onSave, onChange })` passes `onChange`, but `UseInlineEditFormParams` only declares `value`/`validate`/`onSave`. Under strict TS this is an excess-property error (TS2353), and separately PropertyField's `onSave?` (optional) is passed into the hook's required `onSave` (TS2322) — the build breaks. Beyond the type error, the `onChange` prop is now dead: the old component invoked `onChange` for onChange-only callers, but the hook never uses it, so any onChange-only usage silently no-ops. Fix: drop `onChange` from the hook call (and from `PropertyFieldProps` if unused), and make the hook's `onSave` optional (or guarantee a value before calling it) to match the prop.

frontend/src/community/crm/api/crmDealApi.ts
```diff
@@ -118,18 +119,46 @@ export const useGetDealLookup = (
...
+  );
+  return response?.data?.results?.[0];
+};
+
+export const useEditDeal = (onError: (error: AxiosError) => void) => {
+  const { updateDeal, updateDealInStage } = useCrmStore((store) => ({
+    updateDeal: store.updateDeal,
+    updateDealInStage: store.updateDealInStage
+  }));
+  return useMutation({
```
ThinuwanW, line 157: 🤖 **Claude** · 💬 **Note**

`useEditDeal.onSuccess` updates the Zustand store directly but never calls `queryClient.invalidateQueries`, unlike `useCreateDeal` (which invalidates `crmDealQueryKeys.ALL` + limitation). After an edit the `GET_DEALS` infinite cache, this deal's `DEAL_BY_ID` cache, deal lookups, and backend contact/company deal metrics (amount/stage/contact changes) all become stale for other consumers that read React Query instead of the store. Consider invalidating `crmDealQueryKeys.ALL` (and relevant contact/company metric keys) on success.

frontend/src/community/crm/components/organisms/DealSidePanel/DealPropertiesSidebar.tsx
```diff
@@ -14,34 +14,49 @@ import {
...
+  isOpen,
+  onStageChange,
+  onAmountChange,
+  onPriorityChange,
+  onOwnerChange,
+  onContactChange
 }) => {
   const translateText = useTranslator("crmModule", "deals", "sidePanel");
 
   const [amount, setAmount] = useState<string>(deal.amount ?? "");
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

All local state (`amount`, `priority`, `selectedStageId`, `selectedOwner`, `selectedContact`) is seeded from the `deal` prop via `useState` initializers, which only run on mount. `DealSidePanel` renders `<DealPropertiesSidebar deal={selectedDeal}>` with no `key` (DealSidePanel.tsx:172), so if `selectedDealId` changes while the panel stays mounted (opening a different deal without closing), the sidebar keeps the previous deal's values. Add `key={deal.id}` to `<DealPropertiesSidebar>` (or reset local state in an effect keyed on `deal.id`). Note DealTitleSection avoids this because useInlineEditForm uses `enableReinitialize`, but the sidebar's plain useState does not.

frontend/src/community/crm/utils/kanbanUtil.ts
```diff
@@ -130,6 +143,49 @@ export const addDealToStageMap = (
       : stage
   );
 
+export const updateDealInStageMap = (
```
ThinuwanW, line 144: 🤖 **Claude** · 💬 **Note**

The new `updateDealInStageMap` implements nontrivial cross-stage move logic (remove from source, append to target, adjust totalCount) but has no unit tests, even though kanbanUtil.test.ts was modified in this PR. Additionally, if `deal.stageId` is not present in `stageMap` (e.g. a stage not loaded / filtered off the board), the deal is filtered out of its source stage and `totalCount` is decremented, but no target stage adds it back — the deal silently vanishes. Add a guard for the missing-target-stage case and add tests covering same-stage update, cross-stage move, and deal-not-found.

frontend/src/community/crm/utils/kanbanUtil.ts
```diff
@@ -62,6 +62,19 @@ export const mapCreatedDealToSlice = (
   stageId: deal.stage.id
 });
 
+export const mapEditedDealToSlice = (
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

`mapEditedDealToSlice` is almost identical to `mapCreatedDealToSlice` — the only difference is `taskCount` (0 vs omitted). Consider consolidating into one mapper that optionally takes `taskCount` to avoid the two functions drifting apart when the slice shape changes.

frontend/src/community/crm/api/crmDealApi.ts
YasinduRC: Fixed  UseInfiniteQueryResult used

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmDealRepositoryImpl.java
YasinduRC: Outdated, removed this mysql handles that operation

frontend/src/community/crm/components/organisms/DealSidePanel/DealSidePanel.tsx
YasinduRC: added the updateDealInStore

frontend/src/community/crm/api/crmDealApi.ts
YasinduRC: added selectedDealId != null to the API call

frontend/src/community/common/assets/languages/english/crmModule.json
YasinduRC: removed the unused keys

frontend/src/community/crm/components/organisms/DealSidePanel/DealSidePanel.tsx
YasinduRC: Fixed, instead of passing the deals as a prop  now it get deal from the store

frontend/src/community/crm/components/organisms/DealSidePanel/DealPropertiesSidebar.tsx
YasinduRC: Fixed, instead of passing the deals as a prop  now it get deal from the store

frontend/src/community/crm/api/crmDealApi.ts
YasinduRC: Since we are getting from the store no need of invalidate crmDealQueryKeys.ALL

frontend/src/community/crm/utils/kanbanUtil.ts
YasinduRC: Used the existing utill since edit path ignores the task count

frontend/src/community/crm/utils/kanbanUtil.ts
YasinduRC: added unitTests

frontend/src/community/crm/components/molecules/PropertyField/PropertyField.tsx
YasinduRC: Removed the onChange and onSave was already required in every caller

frontend/src/community/common/assets/languages/english/crmModule.json
YasinduRC: Checked add Deal has the same validation message

frontend/src/community/crm/components/organisms/DealSidePanel/DealDescriptionSection.tsx
YasinduRC: Moved to a useInlineEdit hook and used fomic

hasalarootcode (dismissed):

frontend/src/community/crm/api/crmDealApi.ts
```diff
@@ -116,20 +118,47 @@ export const useGetDealLookup = (
...
+}: CrmDealEditPayload): Promise<CrmDealResponseType> => {
+  const response = await authFetch.patch(
+    crmDealEndpoints.EDIT_DEAL(id),
+    fields
+  );
+  return response?.data?.results?.[0];
+};
+
+export const useEditDeal = (onSuccess?: () => void, onError?: () => void) => {
+  const { updateDeal, updateDealInStage } = useCrmStore((store) => ({
```
ThinuwanW: do all the extra operations in the component  level and keep this simple

frontend/src/community/crm/utils/kanbanUtil.ts
```diff
@@ -130,6 +130,49 @@ export const addDealToStageMap = (
       : stage
   );
 
+export const updateDealInStageMap = (
```
ThinuwanW, line 144: rework this to support the order with the order indexing logics

ThinuwanW (changes requested):

frontend/src/community/crm/api/crmDealApi.ts
YasinduRC: Moved kept the api level clean and simple, and handled it in the component level

frontend/src/community/crm/utils/kanbanUtil.ts
YasinduRC: -  instead of locally appending the moved card, on a stage-change success we refetch just the source and target lanes from page 0 and replace them in the board store. Server order becomes client order, and both lanes' pagination (currentPage, hasNextPage, totalCount) resets automatically from the fresh response.

- all other lanes keep their loaded pages and scroll state,only the two affected lanes reset.

- non-stage edits still use the in-place updateDealInStage patch

ThinuwanW (approved):

PR #1599, opened 2026-07-08
Feat/create contact validations updated

anusham-rc:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR makes companyId optional when creating a CRM contact and widens the contact-name regex to permit digits and apostrophes. The changes are consistent with the nullable company_id column and existing null-company handling elsewhere (lookup/detail mappers). No blocking issues found; a couple of suggestions around company-existence validation and test coverage for the new optional-company path.

Found **6** new issue(s):

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmContactServiceImpl.java
```diff
@@ -91,7 +91,8 @@ public ResponseEntityDto createContact(CrmContactCreateRequestDto requestDto) {
 			throw new ModuleException(CrmMessageConstant.CRM_ERROR_CONTACT_EMAIL_ALREADY_EXISTS);
 		}
 
-		CrmCompany company = crmCompanyDao.getReferenceById(requestDto.getCompanyId());
+		CrmCompany company = requestDto.getCompanyId() != null
```
ThinuwanW, line 94: 🤖 **Claude** · 💬 **Note**

Now that company is optional and companyId is no longer validated on create, a non-null but non-existent companyId still flows straight into crmCompanyDao.getReferenceById(...), which returns a lazy proxy without hitting the DB. The invalid FK only surfaces as a DataIntegrityViolationException at flush time, producing a 500 rather than a clean ModuleException. (Note the editContact path has the same pattern.) Consider verifying existence with existsById/findByIdAndIsDeletedFalse and throwing ModuleException(CRM_ERROR_COMPANY_NOT_FOUND) when a companyId is supplied but doesn't resolve.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmContactServiceImpl.java
```diff
@@ -81,7 +81,7 @@ public ResponseEntityDto createContact(CrmContactCreateRequestDto requestDto) {
 		log.info("createContact: execution started");
 
 		validateContactPayload(requestDto.getName(), requestDto.getEmail(), requestDto.getContactNumber(),
-				requestDto.getOwnerId(), requestDto.getCompanyId());
+				requestDto.getOwnerId());
```
ThinuwanW, line 84: 🤖 **Claude** · 💬 **Note**

The core new behavior — creating a contact with no companyId (company == null) — is not covered by any test in CrmContactControllerIntegrationTest; all createContact tests still pass a valid companyId. Add a test asserting a contact can be created with companyId omitted, so this newly-permitted path is protected against regressions.

backend/src/main/java/com/skapp/community/crmplanner/constant/CrmConstants.java
```diff
@@ -38,7 +38,7 @@ public class CrmConstants {
 	public static final Set<Role> ASSIGNABLE_CRM_ROLES = Set.of(Role.CRM_ADMIN, Role.CRM_SALES_MANAGER,
 			Role.CRM_SALES_REPRESENTATIVE);
 
-	public static final String CONTACT_NAME_REGEX = "^[\\p{L} \\-.,]+$";
+	public static final String CONTACT_NAME_REGEX = "^[\\p{L}\\p{N} \\-.,']+$";
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

The widened regex uses a straight ASCII apostrophe (') but not the typographic apostrophe (U+2019), so names like "O’Brien" pasted from many editors/mobile keyboards will still be rejected as invalid. Also \p{N} allows any Unicode numeric character (e.g. superscripts, other scripts' digits), which is broader than ASCII 0-9. Confirm both are intended; if only ASCII digits/apostrophes are desired, tighten to [0-9] and consider adding U+2019.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmContactServiceImpl.java
ThinuwanW, line 94: 🤖 **Claude** · 💬 **Note**

Asymmetry (Checklist #10): create now supports a company-less contact (companyId == null → company left null), but editContact (lines 149-153) treats a null companyId as "leave unchanged", so there is no way to DETACH a company from an existing contact once set. If the product intends contacts to be optionally company-linked, edit should be able to clear it too (e.g. an explicit sentinel or a separate flag), otherwise a contact created with a company can never be reverted to company-less. Confirm this is intended; if not, add an unset path in editContact and a covering test.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmContactServiceImpl.java
ThinuwanW, line 84: 🤖 **Claude** · 💬 **Note**

📁 **File-level observation**

Dead code (Checklist #5) in editContact line 150: `CrmValidations.validateCompanyId(requestDto.getCompanyId())` is called inside `if (requestDto.getCompanyId() != null)`, but validateCompanyId only throws when the argument is null — so guarded by the non-null check it can never throw and is a no-op. This PR removed validateCompanyId from the create path, leaving edit as its only CrmContact caller where it is now effectively unreachable. Remove the redundant call in editContact (the id-existence problem it appears to guard against is actually handled — or not — by getReferenceById on the next line, same as create).

backend/src/main/java/com/skapp/community/crmplanner/constant/CrmConstants.java
ThinuwanW: 🤖 **Claude** · 💬 **Note**

The widened CONTACT_NAME_REGEX now adds `\p{N}`, so an all-numeric string (e.g. "12345") passes contact-name validation. For a person/contact name this may be unintended. Note DEAL_STAGE_NAME_REGEX (line 28) uses a `(?=.*\p{L})` lookahead to require at least one letter — consider the same here if names must contain a letter. Also note there is no positive test asserting the newly-allowed characters (digit, comma, apostrophe) are accepted, only a negative test for '@'/'!'; add one to lock in the intended behavior.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmContactServiceImpl.java
anusham-rc: handled now in both create and edit

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmContactServiceImpl.java
anusham-rc: handled in both scenarios

backend/src/main/java/com/skapp/community/crmplanner/constant/CrmConstants.java
anusham-rc: this regex is intentional

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmContactServiceImpl.java
anusham-rc: when creating this is intentional,
when edit company detaching will be handle separately

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmContactServiceImpl.java
anusham-rc: fixed, create/edit now use findByIdAndIsDeletedFalse(...).orElseThrow(COMPANY_NOT_FOUND)

backend/src/main/java/com/skapp/community/crmplanner/constant/CrmConstants.java
anusham-rc: Requirement is that names must not contain numbers at all, so \p{N} has been removed from CONTACT_NAME_REGEX (rather than adding the lookahead), keeping the apostrophe/comma widening — positive and negative tests added for both.

akilarootcode (approved):

ThinuwanW (approved):

PR #1600, opened 2026-07-08
feat: kanban board deatail view

anusham-rc:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

Small, well-targeted PR wiring up deal-detail navigation from the Kanban board and fixing two dropdown value-matching issues (stage id stringification, and rendering the selected contact when it's absent from the search results). All changes are type-consistent and follow existing patterns (DealsSection uses the identical click handler; store/enum targets exist). No correctness bugs found; one observation about local-only state persistence.

frontend/src/community/crm/components/organisms/DealSidePanel/DealPropertiesSidebar.tsx
```diff
@@ -61,8 +61,8 @@ const DealPropertiesSidebar: FC<DealPropertiesSidebarProps> = ({
   const stageOptions = useMemo(
     () =>
       dealStages.map((stage) => ({
-        id: stage.id,
-        value: stage.id,
+        id: String(stage.id),
```
ThinuwanW, line 64: 🤖 **Claude** · 🟡 **Suggestion**

This cast correctly fixes the stage dropdown selection matching (selectedStageId is already a string), but note that changing the stage — like amount, priority, owner, and contact in this sidebar — only updates local useState and is never persisted via a mutation/invalidateQueries. Now that the dropdown is actually selectable, users can change the stage with no backend effect. If wiring the save is a follow-up, that's fine; if not, this selection should trigger a deal-update mutation. Flagging so it's a conscious decision.

frontend/src/community/crm/components/organisms/DealsKanbanBoard/DealsKanbanBoard.tsx
```diff
@@ -50,6 +50,12 @@ const DealsKanbanBoard: FC<DealsKanbanBoardProps> = ({
     openCrmSidePanel(CrmSidePanelTypes.ADD_DEAL_SIDE_PANEL);
   };
 
+  const handleDealClick = (dealId: number) => {
```
ThinuwanW, line 53: 🤖 **Claude** · 🟡 **Suggestion**

Code duplication: `handleDealClick` here is identical to `handleDealOnClick` in DealsSection.tsx (lines 52-55) — both call `setSelectedDealId(id)` then `openCrmSidePanel(CrmSidePanelTypes.DEAL_DETAIL_SIDE_PANEL)`. Per guideline checklist #9 (2+ components with the same logic → extract a custom hook), extract this into a shared hook (e.g. `useOpenDealDetail()` returning an `openDeal(dealId)` callback) so the store keys and panel type stay in sync across the list and board views. Note also that DealsSection reads these via a `useCrmStore` selector while this file uses `useCrmStore.getState()`; consolidating removes that inconsistency too.

frontend/src/community/crm/components/molecules/ContactPopupSearch/ContactPopupSearch.tsx
```diff
@@ -65,7 +65,9 @@ const ContactPopupSearch: FC<Props> = ({
     option: DropdownOption | null,
     triggerProps: TriggerProps
   ) => {
-    const contact = findById(contacts, Number(option?.id), getContactId);
+    const contact =
```
ThinuwanW, line 68: 🤖 **Claude** · 🟡 **Suggestion**

Behavioral inconsistency with the owner-search pattern. This change patches only `handleRenderTrigger` so the trigger displays a selected contact that isn't in the current `contacts` search results, but `buildContactOptions` (dealUtil.ts) — unlike `buildOwnerOptions`, which prepends the selected item when missing — still omits it, so `handleRenderOption` (line 94) will not render the currently-selected contact in the dropdown list (it returns `null`). The selected contact therefore shows in the trigger but cannot be seen/highlighted in the open list. Prefer mirroring `buildOwnerOptions`: have `buildContactOptions` accept `selectedContact` and prepend it when absent, then this trigger fallback becomes unnecessary and both surfaces stay consistent.

frontend/src/community/crm/components/organisms/DealsKanbanBoard/DealsKanbanBoard.tsx
```diff
@@ -75,7 +81,7 @@ const DealsKanbanBoard: FC<DealsKanbanBoardProps> = ({
                 totalCount={stageDeals?.totalCount ?? 0}
                 isOver={overStageId === stage.id}
                 searchKeyword={searchKeyword}
-                onDealClick={() => {}}
+                onDealClick={handleDealClick}
```
ThinuwanW, line 84: 🤖 **Claude** · 🟣 **Nit**

This changes `onDealClick` from a no-op `() => {}` to real navigation that opens the detail side panel. Because the click target (DealCard's title `div` with `role="button"`) lives inside DraggableDealCard, whose parent carries the dnd-kit drag listeners, verify that completing a drag-and-drop of a card does not also fire the title's `onClick` and unexpectedly open the detail panel after a drop. If it does, guard the handler using the sortable `isDragging` state or stop propagation. Flagging as a nit since browsers usually suppress the synthetic click after a real drag, but the behavior was inert before this PR so it is now newly reachable.

frontend/src/community/crm/components/organisms/DealSidePanel/DealPropertiesSidebar.tsx
anusham-rc: Yes, intentional for now — the deal-update endpoint isn't integrated yet, so these fields are local-only in this PR. Persistence will be wired in another pr

frontend/src/community/crm/components/organisms/DealsKanbanBoard/DealsKanbanBoard.tsx
anusham-rc: Extracted the shared logic into a useDealDetailPanel hook (open/close) and updated both views plus the side panel to use it, also resolves the selector vs getState() inconsistency.

frontend/src/community/crm/components/molecules/ContactPopupSearch/ContactPopupSearch.tsx
anusham-rc: edit deal is not implemented, this handles in separate pr

frontend/src/community/crm/components/organisms/DealsKanbanBoard/DealsKanbanBoard.tsx
anusham-rc: dnd-kit handles this itself: once the 8px activation constraint is met, PointerSensor adds a capture-phase document click listener calling stopPropagation and keeps it ~50ms past the drop (AbstractPointerSensor.detach), so a completed drag can't fire the title's onClick. No guard needed.

frontend/src/community/crm/hooks/useDealDetailPanel.ts
```diff
@@ -0,0 +1,20 @@
+import { useCrmStore } from "~community/crm/store/store";
+import { CrmSidePanelTypes } from "~community/crm/types/SidePanelTypes";
+
+const useDealDetailPanel = () => {
```
akilarootcode: roleback this to old one

akilarootcode (changes requested):

frontend/src/community/crm/hooks/useDealDetailPanel.ts
anusham-rc: reverted and remove unwanted hook

akilarootcode (approved):

ThinuwanW (approved):

PR #1610, opened 2026-07-10
refactor: enhance deal stage sorting and display logic

anusham-rc:
### Problem
Stage lists were sorted purely by `orderIndex`. Since `findNextOrderIndex()` assigns
max(orderIndex) + 1 across **all** stages, newly created open stages received a higher
index than the seeded Won/Lost stages — so they appeared **after** Won and Lost on the
board instead of between Initial and the terminal stages.

### Solution
Stages are now returned in a fixed display order: **Initial → Open (by order index) → Won → Lost.**

- `CrmDealStageType` now carries an explicit `displayOrder` field (`INITIAL(1), OPEN(2), WON(3), LOST(4)`),
  so the ordering is self-documenting and doesn't rely on enum declaration order.
- Added `CrmUtil.sortStagesForDisplay(...)`, which sorts by the type's display order first,
  then by `orderIndex` within the same type.
- Applied the sort at all stage-listing points:
  - `CrmDealServiceImpl.getBoardInitData()` — board init stages
  - `CrmDealStageServiceImpl.getDealStages()` — stage list endpoint
  - `CrmDealStageServiceImpl.reorderDealStages()` — response previously returned stages
    in their pre-reorder fetch order; now reflects the newly assigned order

### Notes
- Sorting is applied **after** `filterVisibleDealStages(...)`, so enterprise tier-visibility
  overrides are unaffected.
- No schema or data changes; ordering is handled in the service layer.

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
@@ -236,8 +236,8 @@ public class CrmDealServiceImpl implements CrmDealService {
 	public ResponseEntityDto getBoardInitData() {
 		log.info("getBoardInitData: execution started");
 
-		List<CrmDealStage> visibleStages = filterVisibleDealStages(
-				crmDealStageDao.findAllByIsDeletedFalseOrderByOrderIndexAsc());
+		List<CrmDealStage> visibleStages = CrmUtil.sortStagesForDisplay(
+				filterVisibleDealStages(crmDealStageDao.findAllByIsDeletedFalseOrderByOrderIndexAsc()));
 		List<CrmBoardStageResponseDto> stages = crmMapper.crmDealStagesToCrmBoardStageResponseDtos(visibleStages);
 
 		List<CrmBoardContactResponseDto> contacts = crmContactDao.findAllContactsForBoardInit()
--- a/backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealStageServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealStageServiceImpl.java
@@ -14,6 +14,7 @@ import com.skapp.community.crmplanner.repository.CrmDealDao;
 import com.skapp.community.crmplanner.repository.CrmDealStageDao;
 import com.skapp.community.crmplanner.service.CrmDealStageService;
 import com.skapp.community.crmplanner.type.CrmDealStageType;
+import com.skapp.community.crmplanner.util.CrmUtil;
 import com.skapp.community.crmplanner.util.CrmValidations;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
@@ -45,8 +46,8 @@ public class CrmDealStageServiceImpl implements CrmDealStageService {
 	public ResponseEntityDto getDealStages() {
 		log.info("getDealStages: execution started");
 
-		List<CrmDealStage> stages = filterVisibleDealStages(
-				crmDealStageDao.findAllByIsDeletedFalseOrderByOrderIndexAsc());
+		List<CrmDealStage> stages = CrmUtil.sortStagesForDisplay(
+				filterVisibleDealStages(crmDealStageDao.findAllByIsDeletedFalseOrderByOrderIndexAsc()));
 
 		log.info("getDealStages: execution ended with {} result(s)", stages.size());
 
@@ -184,7 +185,8 @@ public class CrmDealStageServiceImpl implements CrmDealStageService {
 
 		log.info("reorderDealStages: execution ended");
 
-		return new ResponseEntityDto(false, crmMapper.crmDealStagesToCrmDealStageResponseDtos(existingStages));
+		return new ResponseEntityDto(false,
+				crmMapper.crmDealStagesToCrmDealStageResponseDtos(CrmUtil.sortStagesForDisplay(existingStages)));
 	}
 
 	private void updateStageTypesAfterReorder(List<CrmDealStage> reorderedStages) {
--- a/backend/src/main/java/com/skapp/community/crmplanner/type/CrmDealStageType.java
+++ b/backend/src/main/java/com/skapp/community/crmplanner/type/CrmDealStageType.java
@@ -1,7 +1,16 @@
 package com.skapp.community.crmplanner.type;
 
+import lombok.Getter;
+
+@Getter
 public enum CrmDealStageType {
 
-	INITIAL, OPEN, WON, LOST
+	INITIAL(1), OPEN(2), WON(3), LOST(4);
+
+	private final int displayOrder;
+
+	CrmDealStageType(int displayOrder) {
+		this.displayOrder = displayOrder;
+	}
 
 }
--- a/backend/src/main/java/com/skapp/community/crmplanner/util/CrmUtil.java
+++ b/backend/src/main/java/com/skapp/community/crmplanner/util/CrmUtil.java
@@ -6,6 +6,7 @@ import com.skapp.community.crmplanner.mapper.CrmMapper;
 import com.skapp.community.crmplanner.model.CrmCompany;
 import com.skapp.community.crmplanner.model.CrmContact;
 import com.skapp.community.crmplanner.model.CrmDeal;
+import com.skapp.community.crmplanner.model.CrmDealStage;
 import com.skapp.community.crmplanner.payload.response.CrmContactDetailResponseDto;
 import com.skapp.community.crmplanner.payload.response.CrmContactListItemDto;
 import com.skapp.community.crmplanner.payload.response.CrmContactLookupResponseDto;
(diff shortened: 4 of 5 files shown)


ThinuwanW (commented):
## 🤖 Claude Code Review

Clean, focused change that adds a display-order to CrmDealStageType and a shared CrmUtil.sortStagesForDisplay helper, applied consistently across getDealStages, getBoardInitData, and reorderDealStages. The sort correctly groups stages by type then orderIndex so newly created OPEN stages no longer render after terminal WON/LOST stages. No correctness or security issues found; only minor test-coverage and defensive-safety observations.

Found **5** new issue(s): 🟡 2 suggestion(s), 🟣 3 nit(s)

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealStageServiceImpl.java
```diff
@@ -45,8 +46,8 @@ public class CrmDealStageServiceImpl implements CrmDealStageService {
 	public ResponseEntityDto getDealStages() {
 		log.info("getDealStages: execution started");
 
-		List<CrmDealStage> stages = filterVisibleDealStages(
-				crmDealStageDao.findAllByIsDeletedFalseOrderByOrderIndexAsc());
+		List<CrmDealStage> stages = CrmUtil.sortStagesForDisplay(
```
ThinuwanW, line 49: 🤖 **Claude** · 🟡 **Suggestion**

The new sort behavior is not exercised by any test. The existing integration tests only use DefaultCrmDealStageValues, whose orderIndex order already coincides with the stage-type display order (INITIAL=1, OPEN=2..5, WON=6, LOST=7), so sortStagesForDisplay produces an identical result and the change is effectively untested. The actual bug this fixes only manifests when an OPEN stage's orderIndex exceeds the terminal stages' orderIndex (e.g., after creating a new stage via createDealStage, which assigns findNextOrderIndex()). Add a test that creates a new OPEN stage (giving it the highest orderIndex) and asserts GET /v1/crm/deal/stage still returns it before WON/LOST, to lock in the intended grouping behavior.

backend/src/main/java/com/skapp/community/crmplanner/util/CrmUtil.java
```diff
@@ -15,9 +16,20 @@
...
 import lombok.experimental.UtilityClass;
 
+import java.util.Comparator;
+import java.util.List;
+
 @UtilityClass
 public class CrmUtil {
 
+	private static final Comparator<CrmDealStage> STAGE_DISPLAY_ORDER = Comparator
+		.comparingInt((CrmDealStage stage) -> stage.getStageType().getDisplayOrder())
```
ThinuwanW, line 26: 🤖 **Claude** · 🟣 **Nit**

The comparator dereferences stage.getStageType().getDisplayOrder() and CrmDealStage::getOrderIndex without null guards. This is safe today because both columns are nullable=false and every caller passes DB-loaded entities, so it is not a reachable NPE in current code — flagging only as a low-confidence defensive note. If a non-persisted CrmDealStage (missing stageType/orderIndex) is ever passed to sortStagesForDisplay, sorting will throw an NPE. Consider using Comparator.nullsLast on the extractors if that ever becomes a possibility.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
```diff
@@ -236,8 +236,8 @@ public ResponseEntityDto getDealsByStages(CrmDealsByStagesRequestDto requestDto)
 	public ResponseEntityDto getBoardInitData() {
 		log.info("getBoardInitData: execution started");
 
-		List<CrmDealStage> visibleStages = filterVisibleDealStages(
-				crmDealStageDao.findAllByIsDeletedFalseOrderByOrderIndexAsc());
+		List<CrmDealStage> visibleStages = CrmUtil.sortStagesForDisplay(
```
ThinuwanW, line 239: 🤖 **Claude** · 🟣 **Nit**

Code duplication: the composed pipeline `CrmUtil.sortStagesForDisplay(filterVisibleDealStages(crmDealStageDao.findAllByIsDeletedFalseOrderByOrderIndexAsc()))` is now repeated verbatim here and in CrmDealStageServiceImpl.getDealStages (line 49-50). Since both must stay in sync to render stages consistently on the board and the stage-list screen, consider extracting a single helper (e.g. a `getVisibleStagesForDisplay()` shared method or a CrmUtil helper that takes the dao result) so a future change to the ordering/filter pipeline only has to be made once.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealStageServiceImpl.java
```diff
@@ -45,8 +46,8 @@ public class CrmDealStageServiceImpl implements CrmDealStageService {
...
+				filterVisibleDealStages(crmDealStageDao.findAllByIsDeletedFalseOrderByOrderIndexAsc()));
```
ThinuwanW, line 50: 🤖 **Claude** · 🟣 **Nit**

Redundant ordering: the query `findAllByIsDeletedFalseOrderByOrderIndexAsc()` sorts by order_index in SQL, but the result is now fully re-sorted in memory by STAGE_DISPLAY_ORDER (stageType.displayOrder, then orderIndex). The SQL ORDER BY no longer influences the final result at the getDealStages/getBoardInitData call sites, so it is wasted DB work. Not incorrect, but worth noting the SQL ordering is now superseded; if kept, a brief comment explaining it is retained only because the query method is shared would avoid future confusion.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealStageServiceImpl.java
```diff
@@ -184,7 +185,8 @@ public ResponseEntityDto reorderDealStages(List<CrmDealStageReorderRequestDto> c
 
 		log.info("reorderDealStages: execution ended");
 
-		return new ResponseEntityDto(false, crmMapper.crmDealStagesToCrmDealStageResponseDtos(existingStages));
+		return new ResponseEntityDto(false,
+				crmMapper.crmDealStagesToCrmDealStageResponseDtos(CrmUtil.sortStagesForDisplay(existingStages)));
```
ThinuwanW, line 189: 🤖 **Claude** · 🟡 **Suggestion**

Response asymmetry the PR leaves unaddressed: reorderDealStages now returns `sortStagesForDisplay(existingStages)`, but `existingStages` was reduced to only non-terminal stages (WON/LOST filtered out at lines 153-155), whereas getDealStages returns sortStagesForDisplay over ALL visible stages including terminal ones. Both responses map to CrmDealStageResponseDto and now both claim to be 'sorted for display', so a client that replaces its stage list with the reorder response will silently drop WON/LOST. This terminal-filtering predates the PR, but the PR makes the two paths look symmetric while their contents differ — either return the full visible+sorted set from reorder (re-fetch and sort all visible stages) or confirm the frontend intentionally ignores terminal stages in the reorder response.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealStageServiceImpl.java
anusham-rc: Added an integration test that creates a new OPEN stage and verifies it appears before WON/LOST in the GET response.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
anusham-rc: filterVisibleDealStages is a protected hook overridden by the enterprise services, so this can't move to a shared helper without breaking that — keeping it in line with the existing template-method pattern.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealStageServiceImpl.java
anusham-rc: The query is shared with the reorder flow and a few test helpers that still rely on the SQL ordering, so leaving it as is.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealStageServiceImpl.java
anusham-rc: Reorder only accepts non terminal stages (WON/LOST in the payload is rejected), so the response intentionally mirrors the request. The FE doesn't replace its full stage list from this response, so keeping as is.

backend/src/main/java/com/skapp/community/crmplanner/util/CrmUtil.java
anusham-rc: Both fields are non-nullable in the DB and only persisted entities reach this sort, so skipping the null guards to keep it lean.

backend/src/main/java/com/skapp/community/crmplanner/type/CrmDealStageType.java
```diff
@@ -1,7 +1,16 @@
...
+
+@Getter
 public enum CrmDealStageType {
 
-	INITIAL, OPEN, WON, LOST
+	INITIAL(1), OPEN(2), WON(3), LOST(4);
+
+	private final int displayOrder;
+
+	CrmDealStageType(int displayOrder) {
```
akilarootcode, line 12: remove this and use required args constructor

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
akilarootcode, line 239: reorder using create side

akilarootcode (changes requested):

ThinuwanW (approved):

PR #1612, opened 2026-07-10
fix: remove the task gate for deleted companies

charith-rc:
### Summary

- Removed the deleted company filter from the tasks query to prevent tasks from disappearing when associated company is deleted.

### How to test

1.

### Project Checklist

  - [x] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmTaskRepositoryImpl.java
+++ b/backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmTaskRepositoryImpl.java
@@ -129,11 +129,8 @@ public class CrmTaskRepositoryImpl implements CrmTaskRepository {
 							cb.lessThan(task.get(CrmTask_.dueAt), cb.literal(LocalDate.now().atStartOfDay()))), 1L)
 					.otherwise(0L))));
 
-		Join<CrmTask, CrmCompany> companyJoin = task.join(CrmTask_.company, JoinType.LEFT);
-
 		query.where(task.get(CrmTask_.contact).get(CrmContact_.id).in(contactIds),
-				cb.isFalse(task.get(CrmTask_.isCompleted)), cb.isFalse(task.get(CrmTask_.isDeleted)),
-				cb.or(cb.isNull(companyJoin), cb.isFalse(companyJoin.get(CrmCompany_.isDeleted))));
+				cb.isFalse(task.get(CrmTask_.isCompleted)), cb.isFalse(task.get(CrmTask_.isDeleted)));
 
 		query.groupBy(task.get(CrmTask_.contact).get(CrmContact_.id));
 
@@ -151,14 +148,12 @@ public class CrmTaskRepositoryImpl implements CrmTaskRepository {
 		Join<CrmTask, CrmContact> directContact = task.join(CrmTask_.contact, JoinType.LEFT);
 		Join<CrmTask, CrmDeal> deal = task.join(CrmTask_.deal, JoinType.LEFT);
 		Join<CrmDeal, CrmContact> dealContact = deal.join(CrmDeal_.contact, JoinType.LEFT);
-		Join<CrmTask, CrmCompany> companyJoin = task.join(CrmTask_.company, JoinType.LEFT);
 
 		query.distinct(true);
 		query.where(cb.and(
 				cb.or(cb.equal(directContact.get(CrmContact_.id), contactId),
 						cb.equal(dealContact.get(CrmContact_.id), contactId)),
-				cb.isFalse(task.get(CrmTask_.isDeleted)),
-				cb.or(cb.isNull(companyJoin), cb.isFalse(companyJoin.get(CrmCompany_.isDeleted)))));
+				cb.isFalse(task.get(CrmTask_.isDeleted))));
 
 		return entityManager.createQuery(query).getResultList();
 	}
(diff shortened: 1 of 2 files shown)


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR removes the deleted-company filter from five CrmTask queries so tasks of soft-deleted companies remain visible, making task behavior consistent with contacts and deals (which already stay visible after a company is deleted). The Criteria API refactor is clean — the companyJoin is now created lazily only when a companyId filter is present, with no dangling joins or unused imports. The main gap is that the deleted integration test verified real behavior and was removed without a replacement, leaving the new behavior and the still-valid contact/deal visibility assertions untested.

Found **3** new issue(s): 🟡 1 suggestion(s), 🟣 1 nit(s)

backend/src/test/java/com/skapp/community/crmplanner/controller/v1/CrmCompanyControllerIntegrationTest.java
```diff
@@ -1,28 +1,6 @@
 package com.skapp.community.crmplanner.controller.v1;
```
ThinuwanW, line 1: 🤖 **Claude** · 💬 **Note**

📁 **File-level observation**

The integration test 'deleteCompany_WithAssociatedRecords_...' was deleted entirely rather than updated. It covered three still-valid behaviors: (1) contacts remain visible after their company is deleted, (2) deals remain visible, and (3) — the new behavior introduced by this PR — tasks should now ALSO remain visible. By removing the whole test, the new task-visibility behavior ships with zero coverage, and the previously-verified contact/deal visibility guarantees lose their regression protection. Per the review guidelines ('service changes without corresponding test updates'), consider re-adding the test with the tasks assertion inverted to expect the task to now be returned by crmTaskDao.findTasks(...), keeping the contact/deal visibility assertions intact.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmTaskRepositoryImpl.java
```diff
@@ -325,12 +316,9 @@ public Map<Long, Long> countTasksByDealIds(List<Long> dealIds) {
 		CriteriaQuery<Tuple> query = cb.createTupleQuery();
 		Root<CrmTask> task = query.from(CrmTask.class);
 
-		Join<CrmTask, CrmCompany> companyJoin = task.join(CrmTask_.company, JoinType.LEFT);
-
 		query.select(cb.tuple(task.get(CrmTask_.deal).get(CrmDeal_.id), cb.count(task.get(CrmTask_.id))));
 		query.where(task.get(CrmTask_.deal).get(CrmDeal_.id).in(dealIds), cb.isFalse(task.get(CrmTask_.isDeleted)),
-				cb.isFalse(task.get(CrmTask_.isCompleted)),
-				cb.or(cb.isNull(companyJoin), cb.isFalse(companyJoin.get(CrmCompany_.isDeleted))));
+				cb.isFalse(task.get(CrmTask_.isCompleted)));
```
ThinuwanW, line 321: 🤖 **Claude** · 🟡 **Suggestion**

Beyond the task-list visibility the deleted test covered, this same filter removal changes two other consumer surfaces that the test did NOT exercise: countTasksByDealIds (line 321 — powers deal task-count badges) and findTaskMetricsByContactId/findOpenTaskSummaryByContactIds (lines 202/133 — power a contact's open/overdue task counts). After this change, tasks whose company was soft-deleted are counted again in these metrics. Since CrmCompanyServiceImpl.deleteCompany (line 160) only soft-deletes the company itself and does NOT cascade to its tasks, those tasks remain fully active and counted. Please confirm this is the intended behavior for the deal-count and contact-metric consumers specifically (not just task lists); if the product intent is still to suppress tasks of deleted companies, the correct fix is to cascade the soft-delete to tasks in deleteCompany rather than relying on a query filter. If it is intended, add/restore an integration test asserting deal counts and contact metrics INCLUDE such tasks so the behavior is pinned.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmTaskRepositoryImpl.java
```diff
@@ -205,8 +199,7 @@ public CrmContactTaskMetrics findTaskMetricsByContactId(Long contactId) {
 		query.where(cb.and(
 				cb.or(cb.equal(directContact.get(CrmContact_.id), contactId),
 						cb.equal(dealContact.get(CrmContact_.id), contactId)),
-				cb.isFalse(task.get(CrmTask_.isDeleted)),
-				cb.or(cb.isNull(companyJoin), cb.isFalse(companyJoin.get(CrmCompany_.isDeleted)))));
+				cb.isFalse(task.get(CrmTask_.isDeleted))));
```
ThinuwanW, line 202: 🤖 **Claude** · 🟣 **Nit**

After removing the company join, findByContactIdWithAssociations (lines 148-156) and findTaskMetricsByContactId (lines 185-202) now share an identical join scaffold (directContact / deal / dealContact LEFT joins) and identical WHERE predicate (contactId matches directContact OR dealContact, AND isDeleted false). They differ only in the SELECT clause. Consider extracting the shared join setup + predicate into a private helper (e.g. buildContactTaskCriteria) to avoid two copies drifting apart in future edits.

charith-rc: since these are criteria queries, extracting predicates might cause unnecessary coupling in these queries that would break other sites if one query needed to change.

backend/src/test/java/com/skapp/community/crmplanner/controller/v1/CrmCompanyControllerIntegrationTest.java
charith-rc: reverted the test delete and updated test

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmTaskRepositoryImpl.java
charith-rc: the tasks should not be cascade deleted as per requirement. we should still show the tasks even if companies are soft deleted

akilarootcode (approved):

ThinuwanW (approved):

PR #1614, opened 2026-07-10
refactor: simplify deal stage sorting and update order index while creating stages

anusham-rc:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealStageServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealStageServiceImpl.java
@@ -14,7 +14,6 @@ import com.skapp.community.crmplanner.repository.CrmDealDao;
 import com.skapp.community.crmplanner.repository.CrmDealStageDao;
 import com.skapp.community.crmplanner.service.CrmDealStageService;
 import com.skapp.community.crmplanner.type.CrmDealStageType;
-import com.skapp.community.crmplanner.util.CrmUtil;
 import com.skapp.community.crmplanner.util.CrmValidations;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
@@ -25,6 +24,7 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
@@ -46,8 +46,8 @@ public class CrmDealStageServiceImpl implements CrmDealStageService {
 	public ResponseEntityDto getDealStages() {
 		log.info("getDealStages: execution started");
 
-		List<CrmDealStage> stages = CrmUtil.sortStagesForDisplay(
-				filterVisibleDealStages(crmDealStageDao.findAllByIsDeletedFalseOrderByOrderIndexAsc()));
+		List<CrmDealStage> stages = filterVisibleDealStages(
+				crmDealStageDao.findAllByIsDeletedFalseOrderByOrderIndexAsc());
 
 		log.info("getDealStages: execution ended with {} result(s)", stages.size());
 
@@ -74,13 +74,35 @@ public class CrmDealStageServiceImpl implements CrmDealStageService {
 		stage.setDescription(requestDto.getDescription());
 		stage.setColor(requestDto.getColor().name());
 		stage.setStageType(CrmConstants.DEFAULT_DEAL_STAGE_TYPE);
-		stage.setOrderIndex(crmDealStageDao.findNextOrderIndex());
+		stage.setOrderIndex(resolveOrderIndexForNewStage());
 
 		CrmDealStage saved = crmDealStageDao.save(stage);
 
 		log.info("createDealStage: execution ended, created stage id={}", saved.getId());
 
 		return new ResponseEntityDto(false, crmMapper.crmDealStageToCrmDealStageResponseDto(saved));
+	}
+
+	private Integer resolveOrderIndexForNewStage() {
+		List<CrmDealStage> existingStages = crmDealStageDao.findAllByIsDeletedFalseOrderByOrderIndexAsc();
+
+		Optional<Integer> firstTerminalOrderIndex = existingStages.stream()
+			.filter(stage -> CrmConstants.TERMINAL_STAGES.contains(stage.getStageType()))
+			.map(CrmDealStage::getOrderIndex)
+			.findFirst();
+
+		if (firstTerminalOrderIndex.isEmpty()) {
+			return crmDealStageDao.findNextOrderIndex();
+		}
+
+		Integer newOrderIndex = firstTerminalOrderIndex.get();
+		List<CrmDealStage> stagesToShift = existingStages.stream()
+			.filter(stage -> stage.getOrderIndex() >= newOrderIndex)
+			.toList();
+		stagesToShift.forEach(stage -> stage.setOrderIndex(stage.getOrderIndex() + 1));
+		crmDealStageDao.saveAll(stagesToShift);
+
+		return newOrderIndex;
 	}
 
 	@Override
@@ -185,8 +207,7 @@ public class CrmDealStageServiceImpl implements CrmDealStageService {
 
 		log.info("reorderDealStages: execution ended");
 
-		return new ResponseEntityDto(false,
-				crmMapper.crmDealStagesToCrmDealStageResponseDtos(CrmUtil.sortStagesForDisplay(existingStages)));
+		return new ResponseEntityDto(false, crmMapper.crmDealStagesToCrmDealStageResponseDtos(existingStages));
 	}
 
 	private void updateStageTypesAfterReorder(List<CrmDealStage> reorderedStages) {
--- a/backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealStageServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealStageServiceImpl.java
@@ -20,6 +20,7 @@ import lombok.extern.slf4j.Slf4j;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
@@ -102,6 +103,9 @@ public class CrmDealStageServiceImpl implements CrmDealStageService {
 		stagesToShift.forEach(stage -> stage.setOrderIndex(stage.getOrderIndex() + 1));
 		crmDealStageDao.saveAll(stagesToShift);
 
+		log.info("resolveOrderIndexForNewStage: shifted {} stage(s) to insert new stage at order index {}",
+				stagesToShift.size(), newOrderIndex);
+
 		return newOrderIndex;
 	}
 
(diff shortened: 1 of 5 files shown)


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR removes the stageType-based display sort (CrmUtil.sortStagesForDisplay and CrmDealStageType.displayOrder) and instead maintains display order directly in order_index by inserting new OPEN stages before terminal (WON/LOST) stages and shifting terminals. The core logic in resolveOrderIndexForNewStage is correct and well-tested. The main risks are backward-compatibility for existing tenants whose custom stages were created under the old logic (order_index above terminals) and a subtle change in the reorder response array ordering.

Found **5** new issue(s):

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealStageServiceImpl.java
```diff
@@ -46,8 +46,8 @@ public class CrmDealStageServiceImpl implements CrmDealStageService {
 	public ResponseEntityDto getDealStages() {
 		log.info("getDealStages: execution started");
 
-		List<CrmDealStage> stages = CrmUtil.sortStagesForDisplay(
-				filterVisibleDealStages(crmDealStageDao.findAllByIsDeletedFalseOrderByOrderIndexAsc()));
+		List<CrmDealStage> stages = filterVisibleDealStages(
```
ThinuwanW, line 50: 🤖 **Claude** · 💬 **Note**

Behavioral/data-migration risk from removing sortStagesForDisplay. Display order now depends entirely on order_index. Under the OLD code, createDealStage set order_index = findNextOrderIndex() (max+1), so any custom OPEN stage created before this change received an order_index HIGHER than the seeded terminal stages (WON=6, LOST=7 → new stage=8). Correct display was previously guaranteed by sortStagesForDisplay (which sorted by stageType.displayOrder first). With that sort now gone, getDealStages/getBoardInitData order purely by order_index, so those pre-existing custom OPEN stages will render AFTER WON/LOST for any tenant that already created stages. There is no Liquibase data-migration changeset in this PR to normalize existing order_index values. Please confirm whether the deal-stage creation feature is already live in production; if so, add a one-time migration that re-sequences order_index so terminal stages sit last (INITIAL, all OPEN, WON, LOST).

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealStageServiceImpl.java
```diff
@@ -185,8 +207,7 @@ public ResponseEntityDto reorderDealStages(List<CrmDealStageReorderRequestDto> c
 
 		log.info("reorderDealStages: execution ended");
 
-		return new ResponseEntityDto(false,
-				crmMapper.crmDealStagesToCrmDealStageResponseDtos(CrmUtil.sortStagesForDisplay(existingStages)));
+		return new ResponseEntityDto(false, crmMapper.crmDealStagesToCrmDealStageResponseDtos(existingStages));
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

Reorder response array order changed. `existingStages` is built from findAllByIsDeletedFalseOrderByOrderIndexAsc() (sorted by the OLD order_index) and then each entity's orderIndex is mutated to the requested value, but the List itself is never re-sorted. Previously CrmUtil.sortStagesForDisplay re-sorted by the new order before mapping, so the returned DTO array reflected the requested order; now it reflects the pre-reorder order (only the orderIndex field values are updated). If any consumer relies on array position rather than the orderIndex field this is a regression. Consider sorting existingStages by the new orderIndex before mapping, e.g. `existingStages.stream().sorted(Comparator.comparing(CrmDealStage::getOrderIndex)).toList()`, so the response order matches the persisted order. The reorder tests only assert DB state via findById, so this ordering change is currently untested — add an assertion on the response array order if it matters to the client.

ThinuwanW: 🤖 **Claude** · 💬 **Note**

Invariant gap exposed by removing type-based display sorting. reorderDealStages only assigns the requested orderIndex values to non-terminal stages, and validateDealStageReorderRequest (CrmValidations) only checks non-null + uniqueness — it does NOT verify the indices form a contiguous 1..N range below the terminal stages' indices. Previously sortStagesForDisplay forced INITIAL/OPEN before WON/LOST regardless of order_index, masking this. Now that getDealStages/getBoardInitData order purely by order_index, an authenticated admin sending e.g. [1,2,3,4,99] (still 5 unique entries, so it passes the size and uniqueness checks) persists an OPEN stage with orderIndex 99, which will then render AFTER WON(7)/LOST(8) on the board. Fix: validate that the requested indices are the contiguous set expected (or below the minimum terminal orderIndex), or re-normalize non-terminal indices to 1..N before persisting so the terminal-last invariant is guaranteed independent of client input.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealStageServiceImpl.java
```diff
@@ -83,6 +83,28 @@ public ResponseEntityDto createDealStage(CrmDealStageCreateRequestDto requestDto
...
+	private Integer resolveOrderIndexForNewStage() {
+		List<CrmDealStage> existingStages = crmDealStageDao.findAllByIsDeletedFalseOrderByOrderIndexAsc();
+
+		Optional<Integer> firstTerminalOrderIndex = existingStages.stream()
+			.filter(stage -> CrmConstants.TERMINAL_STAGES.contains(stage.getStageType()))
+			.map(CrmDealStage::getOrderIndex)
+			.findFirst();
+
+		if (firstTerminalOrderIndex.isEmpty()) {
+			return crmDealStageDao.findNextOrderIndex();
```
ThinuwanW, line 96: 🤖 **Claude** · 💬 **Note**

Redundant DB round-trip. In the firstTerminalOrderIndex.isEmpty() branch, existingStages has already been loaded, so the next order index can be derived in memory (max(orderIndex)+1 over existingStages) instead of issuing a second query via crmDealStageDao.findNextOrderIndex(). Minor efficiency nit — consider computing it from the already-fetched list.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealStageServiceImpl.java
```diff
@@ -83,6 +83,28 @@ public ResponseEntityDto createDealStage(CrmDealStageCreateRequestDto requestDto
 		return new ResponseEntityDto(false, crmMapper.crmDealStageToCrmDealStageResponseDto(saved));
 	}
 
+	private Integer resolveOrderIndexForNewStage() {
```
ThinuwanW, line 87: 🤖 **Claude** · 💬 **Note**

resolveOrderIndexForNewStage performs a silent data mutation — it shifts and persists (saveAll) the order_index of the terminal stages — but emits no log line. Given the logging convention in this codebase (entry logging + auditable writes), consider a log.info noting how many stages were shifted and the chosen insertion index, to aid debugging of ordering issues. Nit.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealStageServiceImpl.java
anusham-rc: no migration needed

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealStageServiceImpl.java
anusham-rc: Fixed. The response is now built in the new orderIndex order, so the returned array matches the persisted order. Also added response-order assertions to the reorder happy-path test.

anusham-rc: Fixed by re-normalizing instead of validating: the requested indices are now treated as relative order only, and the existing persisted index set is reassigned in that order. A payload like [1,2,3,4,99] persists as [1,2,3,4,5], so reordered stages can never land after WON/LOST. Added an integration test covering this case.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealStageServiceImpl.java
anusham-rc: Keeping this as is. The extra query only runs in the branch where no terminal stages exist, which shouldn't happen with seeded data, so the impact is negligible, and findNextOrderIndex keeps that fallback self contained.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealStageServiceImpl.java
anusham-rc: Added resolveOrderIndexForNewStage now logs how many stages were shifted and the chosen insertion index.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealStageServiceImpl.java
```diff
@@ -83,6 +84,31 @@ public ResponseEntityDto createDealStage(CrmDealStageCreateRequestDto requestDto
...
+		}
+
+		Integer newOrderIndex = firstTerminalOrderIndex.get();
+		List<CrmDealStage> stagesToShift = existingStages.stream()
+			.filter(stage -> stage.getOrderIndex() >= newOrderIndex)
+			.toList();
+		stagesToShift.forEach(stage -> stage.setOrderIndex(stage.getOrderIndex() + 1));
+		crmDealStageDao.saveAll(stagesToShift);
+
+		log.info("resolveOrderIndexForNewStage: shifted {} stage(s) to insert new stage at order index {}",
```
yuhanga-rtc: remove

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealStageServiceImpl.java
```diff
@@ -161,32 +187,34 @@ public ResponseEntityDto reorderDealStages(List<CrmDealStageReorderRequestDto> c
...
 
-		CrmDealStage firstExistingStage = existingStagesMap.get(firstRequestedStage.getId());
+		CrmDealStage firstExistingStage = existingStagesMap.get(orderedRequest.getFirst().getId());
 		if (firstExistingStage == null || firstExistingStage.getStageType() != CrmDealStageType.INITIAL) {
 			throw new ModuleException(CrmMessageConstant.CRM_ERROR_DEAL_STAGE_REORDER_INVALID_REQUEST);
 		}
 
-		changedStages.forEach(newStage -> {
-			CrmDealStage stage = existingStagesMap.get(newStage.getId());
+		List<CrmDealStage> reorderedStages = IntStream.range(0, orderedRequest.size()).mapToObj(i -> {
```
yuhanga-rtc: use meaningful name for ''i''

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealStageServiceImpl.java
anusham-rc: removed

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealStageServiceImpl.java
anusham-rc: added position as i

akilarootcode (approved):

ThinuwanW (approved):

PR #1619, opened 2026-07-10
fix: handle company unlinking in contact edit and update frontend logic

anusham-rc:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmContactServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmContactServiceImpl.java
@@ -154,6 +154,9 @@ public class CrmContactServiceImpl implements CrmContactService {
 				.orElseThrow(() -> new ModuleException(CrmMessageConstant.CRM_ERROR_COMPANY_NOT_FOUND));
 			contact.setCompany(company);
 		}
+		else {
+			contact.setCompany(null);
+		}
 
 		if (requestDto.getOwnerId() != null) {
 			CrmValidations.validateOwnerId(requestDto.getOwnerId());
--- a/backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmContactServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmContactServiceImpl.java
@@ -149,13 +149,17 @@ public class CrmContactServiceImpl implements CrmContactService {
 			contact.setContactNumber(normalizeNullableText(requestDto.getContactNumber()));
 		}
 
-		if (requestDto.getCompanyId() != null) {
-			CrmCompany company = crmCompanyDao.findByIdAndIsDeletedFalse(requestDto.getCompanyId())
-				.orElseThrow(() -> new ModuleException(CrmMessageConstant.CRM_ERROR_COMPANY_NOT_FOUND));
-			contact.setCompany(company);
-		}
-		else {
-			contact.setCompany(null);
+		if (requestDto.getCompanyId().isPresent()) {
+			Long companyId = requestDto.getCompanyId().get();
+
+			if (companyId != null) {
+				CrmCompany company = crmCompanyDao.findByIdAndIsDeletedFalse(companyId)
+					.orElseThrow(() -> new ModuleException(CrmMessageConstant.CRM_ERROR_COMPANY_NOT_FOUND));
+				contact.setCompany(company);
+			}
+			else {
+				contact.setCompany(null);
+			}
 		}
 
 		if (requestDto.getOwnerId() != null) {
--- a/backend/src/test/java/com/skapp/community/crmplanner/controller/v1/CrmContactControllerIntegrationTest.java
+++ b/backend/src/test/java/com/skapp/community/crmplanner/controller/v1/CrmContactControllerIntegrationTest.java
@@ -455,7 +455,7 @@ class CrmContactControllerIntegrationTest {
 	}
 
 	@Test
-	@DisplayName("Edit contact with partial payload - Only provided fields updated")
+	@DisplayName("Edit contact with partial payload - Omitted fields preserved except company, which requires companyId")
 	void editContact_PartialPayload_OnlyProvidedFieldsUpdated() throws Exception {
 		Long companyId = savedCompany().getId();
 		Long contactId = savedContact(companyId, "original@example.com").getId();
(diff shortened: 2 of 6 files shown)


ThinuwanW (commented):
## 🤖 Claude Code Review

The PR enables unlinking a contact's company by treating a null companyId as 'remove the company', and updates the frontend to always send companyId so partial edits don't accidentally clear it. The backend change is functionally correct and well-tested, but it silently breaks PATCH partial-update semantics for the companyId field only, creating a data-loss risk for any API consumer that omits companyId while relying on the endpoint's otherwise-partial behavior.

Found **5** new issue(s):

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmContactServiceImpl.java
```diff
@@ -154,6 +154,9 @@ public ResponseEntityDto editContact(Long id, CrmContactEditRequestDto requestDt
 				.orElseThrow(() -> new ModuleException(CrmMessageConstant.CRM_ERROR_COMPANY_NOT_FOUND));
 			contact.setCompany(company);
 		}
+		else {
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

This `else { contact.setCompany(null); }` makes companyId behave differently from every other field on this PATCH endpoint. name/email/contactNumber/ownerId are only touched when non-null (true partial-update semantics), but now any request that omits companyId will unlink the company. Because `CrmContactEditRequestDto.companyId` is a plain `Long`, Jackson maps both an absent field and an explicit `null` to the same value — the service cannot distinguish 'don't touch the company' from 'remove the company'. The frontend was updated to always send companyId, but this is a shared service/REST contract: any other consumer (integrations, future callers, or a partial name-only update) will silently wipe the company association. Consider distinguishing presence from null explicitly (e.g. `JsonNullable<Long>` via jackson-databind-nullable, or a boolean `clearCompany` flag) so an omitted companyId leaves the association untouched while an explicit null clears it. At minimum, document the required-always contract and add a code comment explaining the asymmetry with the other fields.

frontend/src/community/crm/components/molecules/EditContactModalContent/EditContactModalContent.tsx
```diff
@@ -91,7 +91,8 @@ const EditContactModalContent = () => {
 
     editContact({
       id: selectedContact?.id,
-      ...changedFields
+      ...changedFields,
+      companyId: normalizedValues.companyId
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

`companyId: normalizedValues.companyId` overrides whatever `...changedFields` already spread in, so companyId is now always sent — this is the compensating change for the new backend semantics and is functionally correct. Two notes: (1) the explicit key is partially redundant since getChangedContactFields already includes companyId when it changed, so consider a brief comment explaining that companyId must always be sent because the backend treats an omitted companyId as an unlink; (2) this tightly couples the frontend to a backend quirk — if the backend is changed to distinguish absent vs. null (see the backend finding), this line should be revisited.

ThinuwanW: 🤖 **Claude** · 💬 **Note**

Regression from always re-sending companyId: because the compensator sends the contact's existing companyId on every edit (even name/email/owner-only edits), the backend re-runs `crmCompanyDao.findByIdAndIsDeletedFalse(companyId)` (CrmContactServiceImpl.java:153) every time. `findByIdWithAssociations` fetches the company via LEFT JOIN without an isDeleted filter (CrmContactRepositoryImpl.java:183), so a contact linked to a company that was later soft-deleted still exposes that companyId to the client. When the user then edits any unrelated field, the compensator re-sends that stale id, the backend lookup returns empty, and the whole edit throws CRM_ERROR_COMPANY_NOT_FOUND — the contact becomes uneditable. Previously (companyId omitted) these edits succeeded. Fix: only include companyId in the payload when it actually changed, and represent intentional unlink distinctly (e.g. backend accepts field-absent = keep, explicit null = clear via a JsonNullable/Optional wrapper) rather than forcing companyId on every request.

backend/src/test/java/com/skapp/community/crmplanner/controller/v1/CrmContactControllerIntegrationTest.java
```diff
@@ -454,16 +460,28 @@ void editContact_PartialPayload_OnlyProvidedFieldsUpdated() throws Exception {
...
 		Long contactId = savedContact(companyId, "original@example.com").getId();
 
-		CrmContactEditRequestDto dto = new CrmContactEditRequestDto();
-		dto.setName("Updated Name Only");
-		// email, contactNumber, companyId, ownerId are null
-
-		performPatchRequest(contactId, dto).andDo(print())
+		// email, contactNumber, ownerId are omitted and stay untouched; companyId is
+		// always carried by the edit payload since null means "remove the company"
+		performPatchRawRequest(contactId, "{\"name\": \"Updated Name Only\", \"companyId\": " + companyId + "}")
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

Test-coverage gap for the new destructive contract. The renamed test now always sends companyId, and `editContact_NullCompanyId_CompanyUnlinked` only covers an explicit `"companyId": null`. The riskiest new behavior for any non-modal API consumer is that a PATCH which OMITS companyId entirely (field absent) also silently clears the company. There is no test asserting that a field-absent payload (e.g. `{"name": "X"}`) unlinks the company. Add one so the contract is documented and regressions are caught. Also, the DisplayName "Only provided fields updated" is now misleading since companyId is implicitly touched even when not conceptually 'provided'.

frontend/src/community/crm/components/molecules/EditContactModalContent/EditContactModalContent.tsx
```diff
@@ -91,7 +91,8 @@ const EditContactModalContent = () => {
 
     editContact({
       id: selectedContact?.id,
-      ...changedFields
+      ...changedFields,
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

Dead comparison in the shared util: `getChangedContactFields` (crmUtil.ts:45-47) still computes and returns `changedFields.companyId`, but this is its only consumer and the very next line overrides companyId with `normalizedValues.companyId` unconditionally. The companyId branch in getChangedContactFields is therefore effectively unreachable/redundant for change detection. Either drop the companyId comparison from the util (documenting that companyId is handled separately) or drop the override — keeping both makes the intent ambiguous for future maintainers.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmContactServiceImpl.java
anusham-rc: this is intentional

frontend/src/community/crm/components/molecules/EditContactModalContent/EditContactModalContent.tsx
anusham-rc: this is intentional

anusham-rc: this is intentional

backend/src/test/java/com/skapp/community/crmplanner/controller/v1/CrmContactControllerIntegrationTest.java
anusham-rc: test description updated

frontend/src/community/crm/components/molecules/EditContactModalContent/EditContactModalContent.tsx
anusham-rc: The comparison isn't dead — it makes company-only edits pass the empty-changes guard; the override is required because the backend unlinks the company when companyId is absent. Removing either introduces a bug.

thusala (dismissed):

thusala:
FE Approved

yuhanga-rtc (dismissed):

akilarootcode (approved):

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmContactServiceImpl.java
akilarootcode: use 		CommonModuleUtils.setIfExists(() -> requestDto.getEmployment().getEmploymentDetails().getEmployeeNumber(),
				employee::setIdentificationNo);

akilarootcode (changes requested):

akilarootcode (approved):

ThinuwanW (approved):

PR #1623, opened 2026-07-13
feat: enhance CRM task and company management with nullable fields an…

dulsara-manakal:
…d cascading updates

### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmContactServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmContactServiceImpl.java
@@ -102,7 +102,7 @@ public class CrmContactServiceImpl implements CrmContactService {
 		CrmContact contact = new CrmContact();
 		contact.setName(requestDto.getName());
 		contact.setEmail(lowercaseEmail);
-		contact.setContactNumber(CrmUtil.normalizeNullableText(requestDto.getContactNumber()));
+		contact.setContactNumber(requestDto.getContactNumber());
 		contact.setCompany(company);
 		contact.setOwner(owner);
 
@@ -147,7 +147,7 @@ public class CrmContactServiceImpl implements CrmContactService {
 
 		if (requestDto.getContactNumber() != null) {
 			CrmValidations.validateContactNumber(requestDto.getContactNumber());
-			contact.setContactNumber(CrmUtil.normalizeNullableText(requestDto.getContactNumber()));
+			contact.setContactNumber(requestDto.getContactNumber());
 		}
 
 		if (requestDto.getCompanyId().isPresent()) {
--- a/backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmContactServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmContactServiceImpl.java
@@ -150,13 +150,17 @@ public class CrmContactServiceImpl implements CrmContactService {
 			contact.setContactNumber(CrmUtil.normalizeNullableText(requestDto.getContactNumber()));
 		}
 
-		if (requestDto.getCompanyId() != null) {
-			CrmCompany company = crmCompanyDao.findByIdAndIsDeletedFalse(requestDto.getCompanyId())
-				.orElseThrow(() -> new ModuleException(CrmMessageConstant.CRM_ERROR_COMPANY_NOT_FOUND));
-			updateContactCompany(contact, company);
-		}
-		else {
-			updateContactCompany(contact, null);
+		if (requestDto.getCompanyId().isPresent()) {
+			Long companyId = requestDto.getCompanyId().get();
+
+			if (companyId != null) {
+				CrmCompany company = crmCompanyDao.findByIdAndIsDeletedFalse(companyId)
+					.orElseThrow(() -> new ModuleException(CrmMessageConstant.CRM_ERROR_COMPANY_NOT_FOUND));
+				contact.setCompany(company);
+			}
+			else {
+				contact.setCompany(null);
+			}
 		}
 
 		if (requestDto.getOwnerId() != null) {
--- a/backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmContactServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmContactServiceImpl.java
@@ -49,6 +49,7 @@ import java.math.BigDecimal;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.function.Function;
 import java.util.stream.Collectors;
@@ -101,7 +102,7 @@ public class CrmContactServiceImpl implements CrmContactService {
 		CrmContact contact = new CrmContact();
 		contact.setName(requestDto.getName());
 		contact.setEmail(lowercaseEmail);
-		contact.setContactNumber(normalizeNullableText(requestDto.getContactNumber()));
+		contact.setContactNumber(CrmUtil.normalizeNullableText(requestDto.getContactNumber()));
 		contact.setCompany(company);
 		contact.setOwner(owner);
 
@@ -146,13 +147,16 @@ public class CrmContactServiceImpl implements CrmContactService {
 
 		if (requestDto.getContactNumber() != null) {
 			CrmValidations.validateContactNumber(requestDto.getContactNumber());
-			contact.setContactNumber(normalizeNullableText(requestDto.getContactNumber()));
+			contact.setContactNumber(CrmUtil.normalizeNullableText(requestDto.getContactNumber()));
 		}
 
 		if (requestDto.getCompanyId() != null) {
 			CrmCompany company = crmCompanyDao.findByIdAndIsDeletedFalse(requestDto.getCompanyId())
 				.orElseThrow(() -> new ModuleException(CrmMessageConstant.CRM_ERROR_COMPANY_NOT_FOUND));
-			contact.setCompany(company);
+			updateContactCompany(contact, company);
+		}
+		else {
+			updateContactCompany(contact, null);
 		}
 
 		if (requestDto.getOwnerId() != null) {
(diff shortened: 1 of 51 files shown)


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR lets CRM entities clear nullable text fields (empty string → null) and lets edit endpoints unlink associations by omitting/nulling them, cascading a contact's company change to its tasks and deals. The changes are internally consistent and well-tested, but the biggest concerns are the newly asymmetric partial-update contract (omitted association fields now delete data instead of being ignored) and the frontend `useEditContact` mutation not invalidating task/deal caches even though the backend now cascades company changes to them.

Found **6** new issue(s): 🔴 1 important, 🟡 1 suggestion(s), 🟣 1 nit(s)

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmContactServiceImpl.java
```diff
@@ -146,13 +147,16 @@ public ResponseEntityDto editContact(Long id, CrmContactEditRequestDto requestDt
...
 		}
 
 		if (requestDto.getCompanyId() != null) {
 			CrmCompany company = crmCompanyDao.findByIdAndIsDeletedFalse(requestDto.getCompanyId())
 				.orElseThrow(() -> new ModuleException(CrmMessageConstant.CRM_ERROR_COMPANY_NOT_FOUND));
-			contact.setCompany(company);
+			updateContactCompany(contact, company);
+		}
+		else {
+			updateContactCompany(contact, null);
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

Behavior/contract change: `editContact` (and `editTask` in CrmTaskServiceImpl, lines 255-278) now treat an omitted/null `companyId`/`contactId`/`dealId` as an explicit request to REMOVE the association (the `else { updateContactCompany(contact, null); }` branch), while every other field in the same DTO (name, email, contactNumber, ownerId, priority, etc.) still uses the classic PATCH semantics of 'null = leave unchanged'. This asymmetry within one endpoint is a silent data-loss footgun: any caller that sends a partial update without re-including these fields will unlink the company/contact/deal (the existing `editContact_PartialPayload` test had to be changed to always resend companyId to keep passing). All current frontend callers were updated, but this makes the API fragile for future callers and any other consumer (mobile/enterprise/direct API). Consider making the removal intent explicit (e.g. a dedicated flag or a sentinel), or documenting the special-case fields clearly, so an omitted field cannot silently destroy data.

frontend/src/community/crm/components/molecules/EditContactModalContent/EditContactModalContent.tsx
```diff
@@ -91,7 +91,8 @@ const EditContactModalContent = () => {
 
     editContact({
       id: selectedContact?.id,
-      ...changedFields
+      ...changedFields,
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

Missing React Query invalidation for cascaded data. This PR makes the backend `editContact` cascade a company change to all of the contact's tasks and deals (cascadeCompanyToContactAssociations). But `useEditContact` (ContactApi.ts) performs no `invalidateQueries`, and this onSuccess only calls `updateContact(data)` on the Zustand store (contact only). As a result the task lists (`taskQueryKeys.GET_OPEN_TASKS`/`GET_COMPLETED_TASKS`) and deal board (`crmDealQueryKeys.ALL`) will display stale company associations until a manual refresh. Note `useDeleteContact` already invalidates exactly these keys because it cascades to tasks/deals — `useEditContact` should do the same now that it also cascades. Fix: add `queryClient.invalidateQueries` for the open/completed task keys and `crmDealQueryKeys.ALL` in `useEditContact` (or in this onSuccess).

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmTaskServiceImpl.java
```diff
@@ -251,17 +252,30 @@ public ResponseEntityDto editTask(Long id, CrmTaskEditRequestDto requestDto) {
...
+				task.setContact(null);
+				task.setCompany(null);
+			}
 		}
 
-		if (requestDto.getDealId() != null) {
-			CrmDeal deal = crmDealDao.findByIdAndIsDeletedFalse(requestDto.getDealId())
-				.orElseThrow(() -> new ModuleException(CrmMessageConstant.CRM_ERROR_DEAL_NOT_FOUND));
-			task.setDeal(deal);
+		Long currentDealId = task.getDeal() != null ? task.getDeal().getId() : null;
```

Possible inconsistent (orphan) state reachable via `editTask`: if a request sends `contactId=null` (clearing contact → also clears company at lines 263-265) while `dealId` is left equal to the current deal, the deal block is skipped and the task keeps a non-null deal with a null contact and null company. The final consistency check `if (task.getDeal() != null && task.getContact() != null)` (line 281) is then skipped because contact is null, so nothing catches it, even though the retained deal still references its own contact/company. The community UI avoids this (handleClearContact in TaskModalForm now also clears the deal), but a direct API call can produce a task whose deal contradicts its (empty) contact. Consider clearing the deal when the contact is cleared, or validating that a deal cannot remain when the contact is removed.

frontend/src/community/crm/components/organisms/TaskSidePanel/TaskSidePanel.tsx
```diff
@@ -139,7 +140,12 @@ const TaskSidePanel: FC = () => {
 
   const handleMarkAsDone = () => {
     updateTaskCompletion(
-      { id: selectedTaskId!, isCompleted: true },
+      {
+        id: selectedTaskId!,
+        isCompleted: true,
+        contactId: selectedTask?.contactId ?? null,
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`handleMarkAsDone` (and the related-tasks `contactId` at line 95) reads `selectedTask?.contactId`, but the task-detail response backing this panel (`CrmTaskViewResponseDto` via `useGetTaskById`) has NO `contactId` field — it only exposes a nested `contact` object (the flat `contactId` exists only on the list DTO `CrmTaskResponseDto`). `CrmTaskDetailType.contactId` is therefore not honored by the detail contract. When the store task originates from the detail response (e.g. a task not previously loaded via the task tab, or a related task), `selectedTask.contactId` is `undefined`, so `contactId ?? null` sends `null`. With the new `editTask` semantics (null contactId => clear contact + company), marking the task done then silently unlinks its contact and company (data loss), while the deal is preserved (uses `deal?.id`), producing an inconsistent state. TaskRow (`task.contact?.id`) and EditTaskModalContent (`selectedTask?.contact?.id`) already use the nested object. Fix: use `selectedTask?.contact?.id ?? null` in both `handleMarkAsDone` and the `contactId` derivation at line 95.

frontend/src/community/crm/components/molecules/TaskRow/TaskRow.tsx
```diff
@@ -49,7 +49,12 @@ const TaskRow: FC<Props> = ({
   const handleToggleChange = (checked: boolean) => {
     setTaskCompleted(checked);
     updateCompletion(
-      { id: task.id, isCompleted: checked },
+      {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The completion toggle (and TaskSidePanel's mark-as-done) now go through `useUpdateTask`, which can change a task's contact/deal/company associations, yet `useUpdateTask` (TaskApi.ts) only invalidates task query keys — it never invalidates `contactQueryKeys.CONTACT_BY_ID` or deal-detail caches. Note `useCreateTask` DOES invalidate `CONTACT_BY_ID(contactId)`, so this is an inconsistency. After a toggle/edit that alters associations, contact-detail and deal-detail views showing the task's linkage can be left stale. Consider invalidating the affected contact/deal detail query keys in `useUpdateTask`'s `onSuccess`, mirroring `useCreateTask`.

backend/src/main/java/com/skapp/community/crmplanner/util/CrmUtil.java
```diff
@@ -34,6 +34,10 @@ public boolean isCrmSalesRepresentative(User user) {
 		return user.getEmployee().getEmployeeRole().getCrmRole() == Role.CRM_SALES_REPRESENTATIVE;
 	}
 
+	public String normalizeNullableText(String value) {
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`normalizeNullableText` collapses to null only via `value.isEmpty()`, but the paired validators (`validateWebsite`, `validateAddress`, `validateContactNumber`) treat blank the same as null via `isBlank()`. A whitespace-only value (e.g. "   ") from a non-trimming API client passes validation but is persisted verbatim instead of being normalized to null, leaving inconsistent blank data. Use `value == null || value.isBlank() ? null : value` (optionally trimming) to align with the validators.

frontend/src/community/crm/components/molecules/EditContactModalContent/EditContactModalContent.tsx
dulsara-manakal: this actualy handle with store values

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmContactServiceImpl.java
dulsara-manakal: Agreed. For MVP we're going with the "always-send the association fields" contract (all current FE callers do this).

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmTaskServiceImpl.java
dulsara-manakal: According to the requirements, a task can have a deal and contact independently. If we remove contact means, we don't want to remove the deal.

frontend/src/community/crm/components/molecules/TaskRow/TaskRow.tsx
dulsara-manakal: we send this because we dont want to set contactId or dealId as null. but in FE dont allow to change that value when toggle on and off

frontend/src/community/crm/components/organisms/TaskSidePanel/TaskSidePanel.tsx
dulsara-manakal: Fixed this.

backend/src/main/java/com/skapp/community/crmplanner/util/CrmUtil.java
dulsara-manakal: BE has a global trimmer, no need to worry about spaced empty strings.

akilarootcode (approved):

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmContactServiceImpl.java
```diff
@@ -146,7 +147,7 @@ public ResponseEntityDto editContact(Long id, CrmContactEditRequestDto requestDt
 
 		if (requestDto.getContactNumber() != null) {
 			CrmValidations.validateContactNumber(requestDto.getContactNumber());
-			contact.setContactNumber(normalizeNullableText(requestDto.getContactNumber()));
+			contact.setContactNumber(CrmUtil.normalizeNullableText(requestDto.getContactNumber()));
```
akilarootcode: remove util

akilarootcode (changes requested):

akilarootcode (approved):

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmContactServiceImpl.java
dulsara-manakal: removed it.

ThinuwanW (approved):

PR #1624, opened 2026-07-13
fix: prevent body scroll when side panel is open

YasinduRC:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/frontend/pages/community/crm/deals.tsx
+++ b/frontend/pages/community/crm/deals.tsx
@@ -44,7 +44,7 @@ const Deals: NextPage = () => {
     >
       <>
         <SidePanelWrapper
-          isOpen={selectedDealId !== null || isCrmSidePanelOpen}
+          isOpen={isCrmSidePanelOpen}
         >
           {selectedDealId !== null && <DealSidePanel />}
           <AddDealSidePanel />
--- a/frontend/pages/community/crm/deals.tsx
+++ b/frontend/pages/community/crm/deals.tsx
@@ -18,10 +18,13 @@ const Deals: NextPage = () => {
   const translateText = useTranslator("crmModule", "deals");
   const { guardCrmCreate, isCheckingCrmLimit } = useCrmLimitGuard();
 
-  const { openCrmSidePanel, selectedDealId } = useCrmStore((store) => ({
-    openCrmSidePanel: store.openCrmSidePanel,
-    selectedDealId: store.selectedDealId
-  }));
+  const { openCrmSidePanel, selectedDealId, isCrmSidePanelOpen } = useCrmStore(
+    (store) => ({
+      openCrmSidePanel: store.openCrmSidePanel,
+      selectedDealId: store.selectedDealId,
+      isCrmSidePanelOpen: store.isCrmSidePanelOpen
+    })
+  );
 
   const handleAddDeal = () => {
     guardCrmCreate(CrmLimitResource.DEALS, () =>
@@ -40,7 +43,9 @@ const Deals: NextPage = () => {
       onPrimaryButtonClick={handleAddDeal}
     >
       <>
-        <SidePanelWrapper>
+        <SidePanelWrapper
+          isOpen={selectedDealId !== null || isCrmSidePanelOpen}
+        >
           {selectedDealId !== null && <DealSidePanel />}
           <AddDealSidePanel />
         </SidePanelWrapper>
--- a/frontend/src/community/crm/components/atoms/SidePanelWrapper/SidePanelWrapper.tsx
+++ b/frontend/src/community/crm/components/atoms/SidePanelWrapper/SidePanelWrapper.tsx
@@ -1,4 +1,4 @@
-import { FC, ReactNode } from "react";
+import { FC, ReactNode,useEffect } from "react";
 
 import { ZIndexEnums } from "~community/common/enums/CommonEnums";
 
@@ -7,6 +7,13 @@ interface Props {
 }
 
 const SidePanelWrapper: FC<Props> = ({ children }) => {
+  useEffect(() => {
+    document.body.style.overflow = "hidden";
+    return () => {
+      document.body.style.overflow = "unset";
+    };
+  }, []);
+
   return <div style={{ zIndex: ZIndexEnums.CRM_SIDE_PANEL }}>{children}</div>;
 };
 
--- a/frontend/src/community/crm/components/atoms/SidePanelWrapper/SidePanelWrapper.tsx
+++ b/frontend/src/community/crm/components/atoms/SidePanelWrapper/SidePanelWrapper.tsx
@@ -1,18 +1,22 @@
-import { FC, ReactNode,useEffect } from "react";
+import { FC, ReactNode, useEffect } from "react";
 
 import { ZIndexEnums } from "~community/common/enums/CommonEnums";
 
 interface Props {
   children: ReactNode;
+  isOpen?: boolean;
 }
 
-const SidePanelWrapper: FC<Props> = ({ children }) => {
+const SidePanelWrapper: FC<Props> = ({ children, isOpen = true }) => {
   useEffect(() => {
+    if (!isOpen) return;
+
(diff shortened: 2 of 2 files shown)


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds a body-scroll lock to SidePanelWrapper via a mount/unmount useEffect. On tasks, contacts, and companies pages the wrapper is conditionally rendered so the lock works as intended, but on the deals page the wrapper is always mounted (panels are conditional inside it), which means body scroll will be locked permanently while on Deals even when no side panel is open. The cleanup also hardcodes the restore value rather than preserving the previous overflow, diverging from the existing pattern in the codebase.

Found **4** new issue(s): 🟡 1 suggestion(s), 🟣 1 nit(s)

frontend/src/community/crm/components/atoms/SidePanelWrapper/SidePanelWrapper.tsx
```diff
@@ -7,6 +7,13 @@ interface Props {
 }
 
 const SidePanelWrapper: FC<Props> = ({ children }) => {
+  useEffect(() => {
```
ThinuwanW, line 11: 🤖 **Claude** · 💬 **Note**

This effect locks body scroll on mount and unlocks on unmount, which assumes SidePanelWrapper is only mounted while a side panel is open. That holds for tasks.tsx, contacts.tsx, and companies.tsx (which render `{selectedId && <SidePanelWrapper>}`), but NOT for deals.tsx: there `<SidePanelWrapper>` is rendered unconditionally and the panels are toggled inside it (`{selectedDealId !== null && <DealSidePanel />}` / `<AddDealSidePanel />`). As a result, on the Deals page `document.body.style.overflow` is set to 'hidden' as soon as the page loads and stays locked even when no panel is open, breaking page scrolling. Fix by making the scroll lock depend on whether a panel is actually open (e.g. accept an `isOpen` prop and guard the effect / include it in the dependency array), or make deals.tsx conditionally mount SidePanelWrapper like the other three pages.

frontend/src/community/crm/components/atoms/SidePanelWrapper/SidePanelWrapper.tsx
```diff
@@ -7,6 +7,13 @@ interface Props {
...
+    document.body.style.overflow = "hidden";
+    return () => {
+      document.body.style.overflow = "unset";
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

The cleanup hardcodes `document.body.style.overflow = "unset"` instead of restoring the previous value. The established pattern in this codebase (see enterprise/sign/.../PdfViewerWithSignature.tsx) captures `const originalOverflow = document.body.style.overflow` and restores it on cleanup. If another component (e.g. a modal) also locks body scroll while this panel is open, unmounting the panel will reset overflow and unlock scrolling prematurely. Consider capturing and restoring the original value for consistency and correctness.

frontend/src/community/crm/components/atoms/SidePanelWrapper/SidePanelWrapper.tsx
ThinuwanW, line 11: 🤖 **Claude** · 🟣 **Nit**

Divergence from the established scroll-lock pattern's accessibility handling. The existing pattern in ConsentModalWithOverlay.tsx (lines 25-53) not only toggles body overflow but also sets `aria-hidden` on the background main content so screen readers/keyboard focus don't reach content behind the open panel. This new effect only locks visual scroll and does nothing for a11y, so background content remains reachable via Tab/AT while the panel is presented. Consider mirroring the existing pattern (set/remove `aria-hidden` on the background container) if the side panel is meant to be a focus-capturing overlay, or confirm the underlying `SidePanel` UI component already handles focus trapping.

frontend/src/community/crm/components/atoms/SidePanelWrapper/SidePanelWrapper.tsx
```diff
@@ -7,6 +7,13 @@ interface Props {
...
+    };
+  }, []);
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

No test coverage for the newly introduced side effect. SidePanelWrapper has no `.test.tsx` file, and mount/unmount body-overflow behavior is exactly the kind of side effect that regresses silently. Add a test (with MockTheme wrapper) asserting `document.body.style.overflow` becomes 'hidden' on mount and is reset on unmount, so the scroll-lock/restore contract is locked down before further changes.

frontend/src/community/crm/components/atoms/SidePanelWrapper/SidePanelWrapper.tsx
akilarootcode: address this comment

frontend/src/community/crm/components/atoms/SidePanelWrapper/SidePanelWrapper.tsx
YasinduRC: We do not consider components when writing frontend tests

frontend/src/community/crm/components/atoms/SidePanelWrapper/SidePanelWrapper.tsx
YasinduRC: Added the established pattern in the existing code,cleannup now restores the captured originalOverflow instead of hardcoding "unset"

frontend/src/community/crm/components/atoms/SidePanelWrapper/SidePanelWrapper.tsx
YasinduRC: Added a isOpen prop to deals and added it to the useEffect logic

YasinduRC: Fixed as the established pattern in the existing code

YasinduRC: The shared SidePanel already uses the correct ARIA mechanism (aria-modal), and manually hiding the parent container would accidentally hide the panel itself

frontend/pages/community/crm/deals.tsx
```diff
@@ -40,7 +43,9 @@ const Deals: NextPage = () => {
       onPrimaryButtonClick={handleAddDeal}
     >
       <>
-        <SidePanelWrapper>
+        <SidePanelWrapper
+          isOpen={selectedDealId !== null || isCrmSidePanelOpen}
```
thusala: remove selectedDealId !== null |

thusala (changes requested):

frontend/pages/community/crm/deals.tsx
YasinduRC: Removed

thusala (approved):

akilarootcode (approved):

frontend/src/community/crm/components/atoms/SidePanelWrapper/SidePanelWrapper.tsx
```diff
@@ -1,12 +1,23 @@
...
 
-const SidePanelWrapper: FC<Props> = ({ children }) => {
+const SidePanelWrapper: FC<Props> = ({ children, isOpen = true }) => {
+  useEffect(() => {
+    if (!isOpen) return;
+
+    const originalOverflow = document.body.style.overflow;
+    document.body.style.overflow = "hidden";
+    return () => {
+      document.body.style.overflow = originalOverflow;
```
ThinuwanW, line 17: I see two problems
1. directly manipulating styles
2. updating body styles 
but since u r pushing to the release I'll merge this

ThinuwanW (approved):

PR #1631, opened 2026-07-14
[Esign] Handle BankId Redirect concern

erandi-rootcode:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/frontend/src/community/common/utils/commonUtil.ts
+++ b/frontend/src/community/common/utils/commonUtil.ts
@@ -573,6 +573,19 @@ export const isMobileDevice = (): boolean => {
   );
 };
 
+export const isIOSDevice = (): boolean => {
+  const userAgent = globalThis.navigator?.userAgent ?? "";
+  if (/iPhone|iPad|iPod/i.test(userAgent)) {
+    return true;
+  }
+  // iPadOS 13+ reports a desktop ("Macintosh") user agent, so fall back to
+  // detecting a Mac platform that also reports touch support.
+  return (
+    globalThis.navigator?.platform === "MacIntel" &&
+    (globalThis.navigator?.maxTouchPoints ?? 0) > 1
+  );
+};
+
 export const replaceTabQueryParam = (path: string, tabId: string): void => {
   const [basePath, query] = path.split("?");
   const params = new URLSearchParams(query);
--- a/frontend/src/community/common/utils/commonUtil.ts
+++ b/frontend/src/community/common/utils/commonUtil.ts
@@ -1,6 +1,7 @@
 import { SxProps, Theme } from "@mui/material";
 import { NextRequest, NextResponse } from "next/server";
 
+import { characterLengths } from "~community/common/constants/stringConstants";
 import { HOURS_PER_DAY } from "~community/common/constants/timeConstants";
 import {
   alphaNumericNamePatternWithSpecialCharacters,
@@ -8,6 +9,7 @@ import {
   matchInvalidEmailCharactersSearchPattern,
   removeNonAlphaNumericCharactersPattern
 } from "~community/common/regex/regexPatterns";
+import { AdminTypes } from "~community/common/types/AuthTypes";
 import {
   DropdownListType,
   FileUploadType
@@ -19,11 +21,8 @@ import {
 import { JobFamilies } from "~community/people/types/JobRolesTypes";
 import { getShortDayName } from "~community/people/utils/holidayUtils/commonUtils";
 
-import { AdminTypes } from "~community/common/types/AuthTypes";
-
 import { appModes } from "../constants/configs";
 import ROUTES from "../constants/routes";
-import { characterLengths } from "~community/common/constants/stringConstants";
 
 export const getBlinkClass = (shouldBlink: boolean): string =>
   shouldBlink ? "animate-pulse" : "";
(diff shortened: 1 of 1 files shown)


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds a single, well-documented `isIOSDevice()` utility to `commonUtil.ts` for use in the enterprise BankID redirect flow. The function is SSR-safe (optional chaining), correctly typed, and non-duplicative. The main concern is that its iPadOS-13+ fallback branch is effectively unreachable in the one place it is consumed, plus a minor reliance on the deprecated `navigator.platform`.

frontend/src/community/common/utils/commonUtil.ts
```diff
@@ -573,6 +573,19 @@ export const isMobileDevice = (): boolean => {
...
 
+export const isIOSDevice = (): boolean => {
+  const userAgent = globalThis.navigator?.userAgent ?? "";
+  if (/iPhone|iPad|iPod/i.test(userAgent)) {
+    return true;
+  }
+  // iPadOS 13+ reports a desktop ("Macintosh") user agent, so fall back to
+  // detecting a Mac platform that also reports touch support.
+  return (
+    globalThis.navigator?.platform === "MacIntel" &&
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The MacIntel + `maxTouchPoints` fallback (added, per the comment, to catch iPadOS 13+ which reports a "Macintosh" UA) is unreachable in the only caller. In `useBankIdVerificationFlow.ts` (line 323-324), `isIOSDevice()` is only evaluated inside `if (isMobileDevice())`, and `isMobileDevice()` tests `/Android|iPhone|iPad|iPod/i` against the UA — which is `false` for an iPadOS 13+ device reporting "Macintosh". So such iPads take the non-mobile `else` branch and never reach this fallback, meaning the intended iOS-specific `redirect=null` handling for iPadOS 13+ works only by coincidence (the else branch also hardcodes `redirect=null`). Consider aligning `isMobileDevice()` to use the same MacIntel/touch detection (e.g. call `isIOSDevice()` inside it), or document that this fallback is intended for future/other callers, so the two detection paths stay consistent.

ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`navigator.platform` is deprecated and, per MDN, may be frozen/removed by some browsers. For iPad detection the more future-proof signal is `navigator.maxTouchPoints > 1` combined with a Mac-like UA (e.g. `/Macintosh/.test(userAgent)`) rather than `platform === "MacIntel"`. Not a functional bug today, but worth considering to avoid the branch silently becoming a no-op if `platform` is dropped.

frontend/src/community/common/utils/commonUtil.ts
```diff
@@ -573,6 +573,19 @@ export const isMobileDevice = (): boolean => {
   );
 };
 
+export const isIOSDevice = (): boolean => {
+  const userAgent = globalThis.navigator?.userAgent ?? "";
+  if (/iPhone|iPad|iPod/i.test(userAgent)) {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Code duplication: the iOS user-agent check `/iPhone|iPad|iPod/i` here is a subset of the regex literal already declared in the adjacent `isMobileDevice()` (`/Android|iPhone|iPad|iPod/i`, line 571). Two util functions in the same file now maintain overlapping device-detection regexes, so a future change (e.g. adding another Apple device token) has to be made in two places and can silently drift. Suggest extracting a shared constant, e.g. `const IOS_UA_REGEX = /iPhone|iPad|iPod/i;` (and an `ANDROID`/mobile variant), and referencing it from both functions so the source of truth is single.

frontend/src/community/common/utils/commonUtil.ts
```diff
@@ -573,6 +573,19 @@ export const isMobileDevice = (): boolean => {
   );
 };
 
+export const isIOSDevice = (): boolean => {
```
ThinuwanW, line 578: 🤖 **Claude** · 🟡 **Suggestion**

Missing test coverage for the new util. A test file already exists at `frontend/src/community/common/utils/__test__/commonUtil.test.ts`, but neither `isIOSDevice` nor `isMobileDevice` is exercised there. Since this function drives a production payment/e-sign redirect decision (BankID `redirect=null` vs encoded href), add unit tests covering: iPhone/iPad/iPod UA → true, Android/desktop UA → false, the MacIntel + `maxTouchPoints > 1` branch → true, and the SSR case where `globalThis.navigator` is undefined → false (verifying it does not throw). Mock `globalThis.navigator` per case.

frontend/src/community/common/utils/commonUtil.ts
erandi-rootcode: Valid Real gap (iPad → desktop/custom-scheme). Fix widens shared isMobileDevice — needs sign-off.

fixed.

erandi-rootcode: Valid and Fixed.

frontend/src/community/common/utils/commonUtil.ts
erandi-rootcode: extracted IOS_UA_REGEX and MOBILE_UA_REGEX as shared module constants, referenced from both functions.

PR #1632, opened 2026-07-14
fix: update deal movement logic to place deals at the top of

anusham-rc:
the target stage and add method to find minimum order index by stage

### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
@@ -387,11 +387,8 @@ public class CrmDealServiceImpl implements CrmDealService {
 		}
 
 		if (previousDealId == null && nextDealId == null) {
-			String existingOrderIndex = crmDealDao.findMaxOrderIndexByStageId(stageId);
-			if (existingOrderIndex != null) {
-				throw new ModuleException(CrmMessageConstant.CRM_ERROR_DEAL_ORDER_NEIGHBOURS_REQUIRED);
-			}
-			return FractionalIndexUtil.generateKeyBetween(null, null);
+			String existingOrderIndex = crmDealDao.findMinOrderIndexByStageId(stageId);
+			return FractionalIndexUtil.generateKeyBetween(null, existingOrderIndex);
 		}
 
 		return FractionalIndexUtil.generateKeyBetween(previousOrderIndex, nextOrderIndex);
--- a/backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
@@ -387,8 +387,8 @@ public class CrmDealServiceImpl implements CrmDealService {
 		}
 
 		if (previousDealId == null && nextDealId == null) {
-			String existingOrderIndex = crmDealDao.findMinOrderIndexByStageId(stageId);
-			return FractionalIndexUtil.generateKeyBetween(null, existingOrderIndex);
+			String minOrderIndex = crmDealDao.findMinOrderIndexByStageId(stageId);
+			return FractionalIndexUtil.generateKeyBetween(null, minOrderIndex);
 		}
 
 		return FractionalIndexUtil.generateKeyBetween(previousOrderIndex, nextOrderIndex);
--- a/backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmBoardController.java
+++ b/backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmBoardController.java
@@ -55,7 +55,7 @@ public class CrmBoardController {
 	}
 
 	@Operation(summary = "Move deal to a different stage",
-			description = "Moves a deal to a different stage (swimlane) on the Kanban board. Provide previousDealId and/or nextDealId to position the deal between neighbours, otherwise it is appended to the end of the target stage.")
+			description = "Moves a deal to a different stage (swimlane) on the Kanban board. Provide previousDealId and/or nextDealId to position the deal between neighbours, otherwise it is placed at the top of the target stage.")
 	@PatchMapping("/deal-move-between-stages")
 	@PreAuthorize("hasAnyRole('ROLE_CRM_SALES_REPRESENTATIVE')")
 	public ResponseEntity<ResponseEntityDto> moveDeal(@RequestBody CrmDealUpdateStageRequestDto requestDto) {
--- a/backend/src/main/java/com/skapp/community/crmplanner/repository/CrmDealRepository.java
+++ b/backend/src/main/java/com/skapp/community/crmplanner/repository/CrmDealRepository.java
@@ -24,6 +24,8 @@ public interface CrmDealRepository {
 
 	String findMaxOrderIndexByStageId(Long stageId);
 
+	String findMinOrderIndexByStageId(Long stageId);
+
 	List<CrmDeal> findByContactIdWithAssociations(Long contactId);
 
 	CrmDeal findByIdWithAssociations(Long id);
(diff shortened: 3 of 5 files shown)


ThinuwanW (commented):
## 🤖 Claude Code Review

The PR changes moveDeal so that dropping a deal into a non-empty stage without neighbours now places it at the top (before the current minimum order index) instead of rejecting the request. The new repository method mirrors findMaxOrderIndexByStageId correctly, aggregate null-handling on empty stages is safe, and the fetch ordering (ascending orderIndex) makes 'top' correct. However, an existing integration test still asserts the old error behavior and will now fail, and the PR ships no test coverage for the new path.

Found **5** new issue(s):

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
```diff
@@ -387,11 +387,8 @@ private String generateOrderIndex(Long dealId, Long stageId, Long previousDealId
 		}
 
 		if (previousDealId == null && nextDealId == null) {
-			String existingOrderIndex = crmDealDao.findMaxOrderIndexByStageId(stageId);
-			if (existingOrderIndex != null) {
-				throw new ModuleException(CrmMessageConstant.CRM_ERROR_DEAL_ORDER_NEIGHBOURS_REQUIRED);
-			}
-			return FractionalIndexUtil.generateKeyBetween(null, null);
+			String existingOrderIndex = crmDealDao.findMinOrderIndexByStageId(stageId);
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

Behavior change breaks an existing test that was not updated. Previously, moving a deal (no neighbours) into a non-empty stage threw CRM_ERROR_DEAL_ORDER_NEIGHBOURS_REQUIRED; now it places the deal at the top. The integration test CrmBoardControllerIntegrationTest.moveDeal_NoNeighbors_StageHasDeals_ReturnsError (backend/src/test/.../v1/CrmBoardControllerIntegrationTest.java:189-204) still asserts a 400 + neighbours-required message and will now FAIL against the new behavior. Update this test to assert the deal is placed at the top (orderIndex < existing minimum, HTTP 200), and add positive coverage for the 'move to top of non-empty stage' path. Note CRM_ERROR_DEAL_ORDER_NEIGHBOURS_REQUIRED is still used by reorderDeal (line 286), so the constant/message is not dead.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
```diff
@@ -387,11 +387,8 @@ private String generateOrderIndex(Long dealId, Long stageId, Long previousDealId
...
+			return FractionalIndexUtil.generateKeyBetween(null, existingOrderIndex);
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

Nit / edge case: FractionalIndexUtil.generateKeyBetween(null, existingOrderIndex) can throw IllegalArgumentException ("cannot decrement any more") once order keys reach the lower bound. Since moving to the top now happens by default (no neighbours) and can be repeated, keys monotonically decrease and could eventually hit the floor. IllegalArgumentException is not a ModuleException, so it would surface as an unhandled 500 rather than a clean error response. Consider wrapping generateKeyBetween failures in a ModuleException with a message constant so the client gets the standard ErrorResponse. This is largely theoretical (also latent in the reorder path), so flagging as low-priority.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
ThinuwanW: 🤖 **Claude** · 💬 **Note**

Behavioral asymmetry between the two flows that share generateOrderIndex. reorderDeal (line 285-287) still rejects a both-null (no previous/next) request with CRM_ERROR_DEAL_ORDER_NEIGHBOURS_REQUIRED, while moveDeal/updateDealStage now silently accepts both-null and places the deal at the top via this new branch. As a result this both-null branch is unreachable from reorderDeal and only serves moveDeal. Please confirm this divergence is intended: either reorderDeal should also support 'move to top' (drop its guard and reuse this branch), or the shared method should not carry logic that one caller can never hit. Document the reasoning so the two 'position without neighbours' paths don't drift further.

ThinuwanW: 🤖 **Claude** · 💬 **Note**

Misleading variable name left over from the old logic. `existingOrderIndex` previously held the max index used only for an emptiness/existence check; it now holds the stage's minimum order index and is passed as the upper bound to generateKeyBetween. Rename to `minOrderIndex` (or `topOrderIndex`) so the intent — inserting before the current top — is clear.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
ThinuwanW: 🤖 **Claude** · 💬 **Note**

Concurrency: duplicate order index risk on the new top-insertion path. findMinOrderIndexByStageId + generateKeyBetween(null, min) is deterministic, so two concurrent moves-to-top of the same stage read the same min and both compute the identical new key, yielding two deals with the same orderIndex and an unstable board order (no unique constraint on stage+orderIndex enforces otherwise). This mirrors the pre-existing createDeal(max+generate) pattern but is newly reachable here. Consider serializing per-stage index generation (e.g. a pessimistic lock on the stage row) or handling a collision retry, at least for the reorder/move flows.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
anusham-rc: Updated the tests

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
anusham-rc: Checked this, it can't actually happen. When keys get close to the smallest possible value, FractionalIndexUtil switches to a different strategy (making the key longer instead of smaller), so it never hits the "cannot decrement" error. And even getting near that point would take an astronomically large number of moves-to-top.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
anusham-rc: This is intentional, reorderDeal moves a card within the same stage, so a request with no neighbours doesn't make sense there (there's nowhere to move it). moveDeal is different: dropping a card into another stage without neighbours simply means "put it at the top".

anusham-rc: Fixed, renamed existingOrderIndex to minOrderIndex.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
anusham-rc: this same race already exists in createDeal and the other reorder paths  it's very rare in practice (two users moving cards at the exact same moment), and fixing it properly means adding stage-level locking to all these paths. We'll handle that as a separate task instead of this bugfix.

yuhanga-rtc (approved):

akilarootcode (approved):

ThinuwanW (approved):

PR #1633, opened 2026-07-14
feat: update crm_deal table schema to include utf8mb4 collation

anusham-rc:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR correctly fixes a real ordering bug: crm_deal.order_index holds a base-62 fractional-index string that must be compared byte-wise (matching FractionalIndexUtil's use of Java String.compareTo), so the new utf8mb4_bin collation is the right choice and is correctly scoped to the only string-based order_index column. The concern is delivery: the fix edits an already-applied Liquibase changeset rather than adding an ALTER migration, so it will not reach existing databases and risks a checksum-validation failure on startup.

Found **3** new issue(s):

backend/src/main/resources/community/db/changelog/ddl-script/common-ddl-script-v1-create-tables.sql
```diff
@@ -651,22 +651,22 @@ CREATE TABLE IF NOT EXISTS `crm_deal_stage`
...
+    `id`                 bigint                                         NOT NULL AUTO_INCREMENT,
+    `created_by`         text                                                    DEFAULT NULL,
+    `created_date`       datetime(6)                                             DEFAULT NULL,
+    `last_modified_by`   text                                                    DEFAULT NULL,
+    `last_modified_date` datetime(6)                                             DEFAULT NULL,
+    `name`               text                                           NOT NULL,
+    `description`        text                                                    DEFAULT NULL,
+    `stage_id`           bigint                                         NOT NULL,
+    `priority`           text                                           NOT NULL,
+    `order_index`        text CHARACTER SET utf8mb4 COLLATE utf8mb4_bin NOT NULL,
```
ThinuwanW, line 663: 🤖 **Claude** · 💬 **Note**

The collation fix is correct, but it is applied by editing an already-applied Liquibase changeset (`-- changeset AkilaSachintha:common-ddl-script-v1-create-tables`) instead of adding a new migration. Two problems for any database where crm_deal already exists: (1) Liquibase stores a checksum of this changeset in DATABASECHANGELOG; changing the file content changes the checksum, and with no `runOnChange`/`validCheckSum` and standard `enabled: true` config (application-prd.yml), startup will fail with a checksum-validation (mismatch) error. (2) Even if validation is bypassed, `CREATE TABLE IF NOT EXISTS` is a no-op on an existing table, so the collation change never applies — existing deployments keep the wrong (case-insensitive) collation and the ordering bug persists. Only fresh installs get the fix. Recommend reverting this edit and adding a new changeset (e.g. `common-ddl-script-v1-alter-crm-deal-order-index-collation.sql`, wired into db.changelog.yml) with: `ALTER TABLE crm_deal MODIFY COLUMN order_index TEXT CHARACTER SET utf8mb4 COLLATE utf8mb4_bin NOT NULL;`. This is safe on existing rows (collation change only affects comparison, not stored values) and fixes both new and existing databases.

backend/src/main/resources/community/db/changelog/ddl-script/common-ddl-script-v1-create-tables.sql
```diff
@@ -683,7 +683,7 @@ CREATE TABLE IF NOT EXISTS `crm_task`
     `last_modified_date` datetime(6)      DEFAULT NULL,
     `name`               text    NOT NULL,
     `type_id`            bigint  NOT NULL,
-    `priority`           text        NOT NULL,
+    `priority`           text    NOT NULL,
```
ThinuwanW, line 686: 🤖 **Claude** · 💬 **Note**

Scope creep / unrelated edit: this change to `crm_task.priority` is a purely cosmetic whitespace/alignment edit (`text        NOT NULL` → `text    NOT NULL`) that has nothing to do with the stated goal ('utf8mb4 collation for order_index'). It touches the crm_task table definition for no functional reason and further mutates the already-applied changeset, expanding the checksum delta beyond what the fix requires. Revert this line so the changeset diff contains only the intended collation change.

backend/src/main/resources/community/db/changelog/ddl-script/common-ddl-script-v1-create-tables.sql
ThinuwanW, line 663: 🤖 **Claude** · 💬 **Note**

Charset/collation consistency: `order_index` is now the ONLY column in the entire DDL script with an explicit `CHARACTER SET utf8mb4 COLLATE utf8mb4_bin`; every other text column (including `name`, `description`, `amount`, `priority` in this same table) inherits the database/schema default charset. If the schema default charset is not utf8mb4, this column silently diverges from its siblings, and any future expression that compares/concatenates `order_index` with another text column of the default charset will raise MySQL 'Illegal mix of collations' at runtime. Current queries only sort/MAX on order_index alone so there is no error today, but consider documenting why only this column is overridden, or pinning the charset at the table/database level, to make the intent explicit and avoid a latent trap.

anusham-rc: this is the requirement

backend/src/main/resources/community/db/changelog/ddl-script/common-ddl-script-v1-create-tables.sql
anusham-rc: check with the team

backend/src/main/resources/community/db/changelog/ddl-script/common-ddl-script-v1-create-tables.sql
anusham-rc: this is the practice, since we have separate migration tool

yuhanga-rtc (approved):

akilarootcode (approved):

ThinuwanW (approved):

PR #1634, opened 2026-07-14
feat(attendance): clock-in/out-only + fingerprint attendance toggles, ADMS ingestion foundation

akilarootcode:
﻿## What

Two new per-tenant attendance settings (Configurations -> Attendance) and the community-side foundation for fingerprint (ADMS) attendance:

- **"Clock in / clock out only (no breaks)"** — a row in Clock in Settings (`CLOCK_IN_OUT_ONLY`): hides pause/resume in the clock widget and the backend rejects PAUSE/RESUME.
- **"Fingerprint attendance"** — its own section (`FINGERPRINT_ATTENDANCE_ENABLED`): device punches become attendance (first punch of the day = clock-in, last punch finalized as clock-out at 23:55 org time via an EventBridge-driven internal call), and the section lists registered devices with last-sync status. Ingestion/finalization logic lives in the enterprise repo.
- **Clock event provenance**: new `clock_in_source` / `clock_out_source` columns on `time_record` (`WEB`, `MOBILE`, `DEVICE`, `MANUAL`) stamped by each flow; community Liquibase migration included (tenant-DB migration is in skapp-migrations#183).
- `EmployeeRepository.updateLastClockInDate` — criteria-based single-column update; a full entity save would re-encrypt field-encrypted columns.
- Bumps the enterprise submodules to the companion feature branches.

## Companion PRs (merge together)
- rootcodelabs/skapp-ep-be-src#715 (ingestion, finalization, schedule lifecycle, device endpoint, internal API)
- rootcodelabs/skapp-ep-fe-src#754 (FingerprintDevices section + device hook)
- rootcodelabs/skapp-ep-be-tests#28 (new coverage + fixes 18 pre-existing ADMS test failures)
- rootcodelabs/skapp-migrations#183 (EventBridge schedule endpoints + tenant DB migration 00121)

## Verification
- Backend full suite green (514 tests); FE typecheck clean on touched files
- E2E on the local stack: toggles round-trip from the UI, PAUSE rejected while restricted, device punches create records with `DEVICE` sources only when the fingerprint toggle is on (raw logs always kept), finalization completes the day with correct worked hours, devices section shows availability + last sync
- Device punches are converted org-timezone -> UTC (web/mobile clients submit UTC; devices punch local wall time)

## Ops follow-up (AWS)
EventBridge bus rule for `attendance-finalization-trigger` -> API destination -> `POST /internal/v1/ep/time/attendance/finalize` (connection auth `x-api-key`), mirroring the existing ai-insights rule.


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds two attendance config toggles (clock-in/out-only, fingerprint attendance), a TimeRecordSource enum with clock-in/out source columns, and a repository method to update last clock-in date — all wired correctly into existing patterns and consumed by enterprise code. The clock-in/out-only feature is fully functional in community mode; the main concern is that the fingerprint section is rendered unconditionally in the shared community component even though the feature is enterprise-only, which is inconsistent with how GeoFencing is encapsulated. Remaining items are minor convention nits.

Found **7** new issue(s): 🔴 1 important, 🟡 2 suggestion(s)

frontend/src/community/attendance/components/organisms/AttendanceConfiguration/AttendanceConfiguration.tsx
```diff
@@ -159,6 +171,37 @@ const AttendanceConfiguration = (): JSX.Element => {
           onSwitchChange={handleSwitchChange}
         />
 
+        <Typography
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

The entire Fingerprint Attendance section (title, description, and the enable toggle) is rendered unconditionally in the shared community component. Only the `FingerprintDevices` device list is behind the enterprise/fallback pattern (its community fallback returns an empty fragment). This is asymmetric with GeoFencing, where the whole section lives inside the `~enterprise/.../GeoFencingSettings` component and its fallback renders nothing — so geofencing is hidden in community mode, but the fingerprint toggle is visible and toggleable in community mode. Since fingerprint sync (AdmsTimeRecordSyncServiceImpl) is enterprise-only, community users can enable FINGERPRINT_ATTENDANCE_ENABLED for a feature that does nothing. Recommend moving the whole fingerprint block (title/description/switch) into an enterprise component with an empty community fallback, mirroring GeoFencingSettings, or otherwise gating it by mode.

backend/src/main/resources/community/db/changelog/ddl-script/time-ddl-script-v1-alter-table-time-record-add-source-columns.sql
```diff
@@ -0,0 +1,8 @@
+-- liquibase formatted sql
+
+-- changeset akila:00121_alter_table_time_record_add_clock_in_out_source_columns
+ALTER TABLE `time_record`
+    ADD COLUMN `clock_in_source` text NULL,
```
ThinuwanW, line 5: 🤖 **Claude** · 💬 **Note**

The new `clock_in_source`/`clock_out_source` columns are typed `text`, but the codebase convention for `@Enumerated(EnumType.STRING)` columns is `varchar(255)` (e.g., `pm_role`, `esign_role`, `crm_role` in the common DDL scripts). `text` is inconsistent, harder to index, and larger than needed for a short enum name. Suggest changing both columns to `varchar(255) DEFAULT NULL` for consistency.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/AttendanceConfigServiceImpl.java
```diff
@@ -132,4 +142,19 @@ public boolean getAttendanceConfigByType(AttendanceConfigType attendanceConfigTy
 		return Boolean.parseBoolean(config.getAttendanceConfigValue());
 	}
 
+	@Override
+	public boolean isAttendanceConfigEnabled(AttendanceConfigType attendanceConfigType) {
```
ThinuwanW, line 146: 🤖 **Claude** · 💬 **Note**

The new service method `isAttendanceConfigEnabled` has no entry logging. The module convention (see the other methods in this class) is `log.info("methodName: execution started")`. Add an entry log line for consistency (nit — the pre-existing `getAttendanceConfigByType` also lacks it, so low priority).

backend/src/main/java/com/skapp/community/timeplanner/type/TimeRecordSource.java
```diff
@@ -0,0 +1,7 @@
+package com.skapp.community.timeplanner.type;
+
+public enum TimeRecordSource {
+
+	WEB, MOBILE, DEVICE, MANUAL
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

The `MANUAL` enum value is introduced but never assigned anywhere in community or enterprise code (only WEB, MOBILE, DEVICE are set). Manually-created time records therefore leave `clockInSource`/`clockOutSource` null rather than MANUAL. If manual entries are intended to be tagged, wire `MANUAL` into the manual-record creation path; otherwise this is a harmless forward-looking value. Flagging as a nit for the developer to confirm intent.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeServiceImpl.java
```diff
@@ -1980,6 +1985,7 @@ private void recordClockInAndClockOut(User currentUser, long timeInMillis, TimeR
 				throw new ModuleException(TimeMessageConstant.TIME_ERROR_CLOCK_IN_NOT_EXISTS_FOR_CURRENT_DATE);
 			}
 			timeRecord.get().setClockOutTime(timeInMillis);
+			timeRecord.get().setClockOutSource(TimeRecordSource.WEB);
```
ThinuwanW, line 1989: 🤖 **Claude** · 🔴 **Important**

Source assignment is incomplete: this PR sets clockOutSource=WEB only in recordClockInAndClockOut, but the sibling web clock-out path updateCurrentUserIncompleteTimeRecords (around line 702) calls timeRecord.setClockOutTime(...) / setCompleted(true) without ever setting clockOutSource. Records completed via that endpoint will persist clock_out_source = NULL even though they are web clock-outs, defeating the purpose of the new source column and making downstream reconciliation (which distinguishes WEB vs DEVICE punches) unreliable. The time-request approval/edit paths (setClockInTime/setClockOutTime around lines 1348–1484) similarly leave both source columns unset. Fix: set the appropriate TimeRecordSource (WEB for the incomplete-record clock-out; MANUAL for approved edits) on every path that mutates clock-in/clock-out times so the column is populated consistently.

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/EmployeeRepositoryImpl.java
```diff
@@ -1616,4 +1617,16 @@ public List<Employee> findAllActiveEmployeesExcludingRole(Role excludedRole, Set
 		return entityManager.createQuery(query).getResultList();
 	}
 
+	@Override
```
ThinuwanW, line 1620: 🤖 **Claude** · 🟡 **Suggestion**

updateLastClockInDate uses a JPA bulk CriteriaUpdate (executeUpdate). Bulk updates bypass Hibernate dirty-checking and the Auditable listener, so lastModifiedBy/lastModifiedDate on the Employee row are NOT updated (a convention every entity extends Auditable for), and any Employee instance already loaded in the persistence context keeps a stale lastClockInDate for the rest of the transaction. The returned update count is also discarded, so a non-existent employeeId silently no-ops. Consider either loading the employee and using a normal setter+save (keeps audit fields correct) or, if the bulk update is intentional for performance, documenting the audit-skip and asserting the update count is 1.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeServiceImpl.java
```diff
@@ -534,6 +535,9 @@ public ResponseEntityDto addTimeRecord(AddTimeRecordDto addTimeRecordDto) {
 		}
 		else if (addTimeRecordDto.getRecordActionType() == TimeRecordActionTypes.RESUME
 				|| addTimeRecordDto.getRecordActionType() == TimeRecordActionTypes.PAUSE) {
+			if (attendanceConfigService.isAttendanceConfigEnabled(AttendanceConfigType.CLOCK_IN_OUT_ONLY)) {
```
ThinuwanW, line 538: 🤖 **Claude** · 🟡 **Suggestion**

New service behavior lands without test coverage. The addTimeRecord PAUSE/RESUME gating that throws TIME_ERROR_PAUSE_RESUME_NOT_AVAILABLE when CLOCK_IN_OUT_ONLY is enabled, and the new AttendanceConfigService.isAttendanceConfigEnabled method, have no accompanying unit tests — the only test changes in this PR are mechanical constructor-argument fixes for the widened AttendanceConfigRequestDto. Add tests covering: (a) PAUSE/RESUME rejected when clock-in-out-only is enabled and allowed when disabled, and (b) isAttendanceConfigEnabled returning false when the config row is absent vs true/false based on the stored value.

akilarootcode:
Thanks for the review — went through each point:

**Addressed**
- **Fingerprint section rendered in community mode** — the whole Fingerprint Attendance section (title/description/toggle/devices) is now gated behind `useGetEnvironment() === ENTERPRISE`, so it's hidden in community mode (matches the reviewer's "or otherwise gating it by mode" alternative to the enterprise-component pattern).
- **Source not set on the incomplete clock-out path** — `updateCurrentUserIncompleteTimeRecords` now sets `clockOutSource = WEB`.
- **Test coverage** — added `isAttendanceConfigEnabled` unit tests (absent row → false, stored true/false). The PAUSE/RESUME gating is covered end-to-end.

**Intentional — not changing**
- **`text` vs `varchar(255)` for the source columns** — deliberate product decision to use `text` for these columns.
- **Entry logging on `isAttendanceConfigEnabled`** — logging is intentionally kept minimal for this feature (as the note itself observes, the sibling `getAttendanceConfigByType` also has none).
- **`updateLastClockInDate` bulk `CriteriaUpdate`** — intentional: a full entity save re-encrypts the field-encrypted `identificationNo` on flush, which we specifically want to avoid; and a device/system clock-in isn't a user modification, so skipping the Auditable fields is the desired behaviour. Nothing re-reads the employee in the same transaction.
- **`MANUAL` enum value unused** — intentional forward-looking value; the time-request approval/manual-entry paths can adopt it in a follow-up. Leaving the enum complete now.

frontend/src/community/attendance/components/organisms/AttendanceConfiguration/AttendanceConfiguration.tsx
akilarootcode: Fixed - the whole Fingerprint Attendance section (title/description/toggle/devices) is now gated behind useGetEnvironment() === ENTERPRISE, so it is hidden in community mode (the 'gate it by mode' alternative you suggested).

backend/src/main/resources/community/db/changelog/ddl-script/time-ddl-script-v1-alter-table-time-record-add-source-columns.sql
akilarootcode: Intentional - deliberate decision to use text for the source columns, so keeping as-is.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/AttendanceConfigServiceImpl.java
akilarootcode: Not changing - logging is intentionally kept minimal for this feature (as the note observes, the sibling getAttendanceConfigByType also has no entry log).

backend/src/main/java/com/skapp/community/timeplanner/type/TimeRecordSource.java
akilarootcode: Intentional forward-looking value; the manual-entry / time-request-approval paths can adopt MANUAL in a follow-up. Keeping the enum complete.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeServiceImpl.java
akilarootcode: Addressed for the clear web path - updateCurrentUserIncompleteTimeRecords now sets clockOutSource = WEB. The time-request approval/edit paths are MANUAL provenance and are deferred to a follow-up (they are pre-existing multi-method flows).

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/EmployeeRepositoryImpl.java
akilarootcode: Intentional - a full entity save re-encrypts the field-encrypted identificationNo on flush, which we specifically avoid; and a device/system clock-in is not a user modification, so skipping the Auditable fields is desired. Nothing re-reads the employee in the same transaction.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeServiceImpl.java
akilarootcode: Added isAttendanceConfigEnabled unit tests (absent row -> false, stored true/false). The PAUSE/RESUME gating is covered end-to-end.

frontend/src/community/attendance/components/organisms/AttendanceConfiguration/AttendanceConfiguration.tsx
```diff
@@ -126,6 +131,15 @@ const AttendanceConfiguration = (): JSX.Element => {
                   handleSwitchChange("isClockInOnLeaveDays", checked)
                 }
               />
+              <SwitchRow
+                labelId="clock-in-out-only"
+                label={attendanceConfigurations(["isClockInClockOutOnly"])}
+                checked={config.isClockInClockOutOnly ?? false}
```
thusala: do we really need this fallback to false? doesnt it handled from inital value that u set to store?

frontend/src/community/attendance/components/organisms/AttendanceConfiguration/AttendanceConfiguration.tsx
```diff
@@ -159,6 +173,40 @@ const AttendanceConfiguration = (): JSX.Element => {
...
+                  }
+                />
+              )}
+            </Box>
+
+            {initialConfig?.isFingerprintAttendanceEnabled && (
+              <FingerprintDevices />
+            )}
+          </>
+        )}
```
thusala: move this part to enterprise component and use tailwind

thusala (changes requested):

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
```diff
@@ -30,6 +31,10 @@ const Timer = ({ disabled }: TimerProps): JSX.Element => {
   const isBelow600 = useMediaQuery()(MediaQueries.BELOW_600);
   const translateText = useTranslator("attendanceModule", "timeWidget");
 
+  const { data: attendanceConfig } = useGetAttendanceConfiguration();
+  const isClockInClockOutOnly =
+    attendanceConfig?.isClockInClockOutOnly === true;
```
hasalarootcode: why do we need to check this? since backend should return a boolean

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
```diff
@@ -84,7 +89,7 @@ const Timer = ({ disabled }: TimerProps): JSX.Element => {
             : "00:00:00"}
         </Typography>
       )}
-      <PlayButton />
+      {!isClockInClockOutOnly && <PlayButton />}
```
hasalarootcode: use it direcly and might need a loader as wel

frontend/src/community/attendance/components/organisms/AttendanceConfiguration/AttendanceConfiguration.tsx
```diff
@@ -159,6 +173,40 @@ const AttendanceConfiguration = (): JSX.Element => {
           onSwitchChange={handleSwitchChange}
         />
 
+        {isEnterprise && (
+          <>
+            <Typography variant="h2" sx={classes.sectionTitle}>
```
hasalarootcode: dont use mui

hasalarootcode (changes requested):

frontend/src/community/attendance/components/organisms/AttendanceConfiguration/AttendanceConfiguration.tsx
akilarootcode: Done — made isClockInClockOutOnly required in the type (the config always carries it), so the ?? false fallback is removed.

frontend/src/community/attendance/components/organisms/AttendanceConfiguration/AttendanceConfiguration.tsx
akilarootcode: Done — moved the whole fingerprint section into a new ~enterprise FingerprintSettings component (with a fallback), tailwind-based, mirroring GeoFencingSettings; the community file no longer has the inline enterprise block.

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
akilarootcode: Done — dropped the === true; the config value is used directly now.

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
akilarootcode: Done — using attendanceConfig?.isClockInClockOutOnly directly (removed the intermediate const).

frontend/src/community/attendance/components/organisms/AttendanceConfiguration/AttendanceConfiguration.tsx
akilarootcode: Done — that fingerprint title/section moved to the tailwind-based FingerprintSettings component, so the MUI Typography here is gone.

hasalarootcode (dismissed):

thusala (dismissed):

shakila-kamalasena (dismissed):

shakila-kamalasena (approved):

shakila-kamalasena (approved):

thusala (approved):

ThinuwanW (approved):

PR #1635, opened 2026-07-15
feat/google workspace sync

umeshaliyanage:
### Summary

- added new non superadmin restriction routes, added new drawerHiddenProtectedRoutes and added a tool tip message 

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR wires up the Google Workspace people-sync feature: new notification types/templates, route config, an icon, modal routing, and Google Cloud dependencies. Most changes are consistent (notification enums match front/back-end contracts), but there is a corrupted SVG namespace, a fallback stub missing its prop type that will break the community typecheck, and an access-control tightening on the existing Add-People route that should be confirmed as intentional.

Found **7** new issue(s): 🔴 1 important, 🟡 1 suggestion(s), 🟣 1 nit(s)

frontend/src/community/common/assets/Icons/GoogleIcon.tsx
```diff
@@ -14,30 +14,72 @@ const GoogleIcon = ({
       id={id}
       width={width}
       height={height}
-      viewBox="0 0 21 20"
+      viewBox="0 0 48 48"
       fill="none"
-      xmlns="http://www.w3.org/2000/svg"
+      xmlns="w3.org/2000/svg"
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

The `xmlns` attribute was corrupted from `http://www.w3.org/2000/svg` to `w3.org/2000/svg`, which is not a valid namespace URI. While browsers are lenient with inline SVG, this breaks any serialization/standalone use of the SVG and is clearly unintentional. Restore `xmlns="http://www.w3.org/2000/svg"`.

frontend/src/fallback/people/components/molecules/ConnectGoogleWorkspaceModal/ConnectGoogleWorkspaceModal.tsx
```diff
@@ -0,0 +1,5 @@
+const ConnectGoogleWorkspaceModal = () => {
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

This fallback stub declares no props (`() => <></>`), but DirectoryPopupController renders it with `isOpen` and `onClose` props. In community mode `~enterprise/*` resolves to `src/fallback/*`, so the strict TypeScript build will fail with 'Property isOpen/onClose does not exist'. Follow the existing fallback pattern (see fallback IntegrationSettings.tsx) and declare a matching props interface, e.g. `interface Props { isOpen: boolean; onClose: () => void; }` accepted (and ignored) by the stub.

frontend/src/community/common/constants/routes.ts
```diff
@@ -181,4 +184,9 @@ export const employeeRestrictedRoutes = [
   ROUTES.PEOPLE.ADD
 ];
 
-export const managerRestrictedRoutes = [ROUTES.PEOPLE.ADD];
+export const nonSuperAdminRestrictedRoutes = [
+  ROUTES.PEOPLE.ADD,
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

`ROUTES.PEOPLE.ADD` is carried over into the new `nonSuperAdminRestrictedRoutes` list, and middleware now passes `AdminTypes.SUPER_ADMIN` as the required role (previously `managerRestrictedRoutes` required `AdminTypes.PEOPLE_ADMIN`). Net effect: a People Admin who is not a Super Admin can no longer reach `/people/directory/add`, an access regression for existing functionality bundled with the Google-import routes. Confirm this tightening is intended; if only the Google import/sync routes should be super-admin-gated, keep ADD in a separate PEOPLE_ADMIN-gated list.

backend/pom.xml
```diff
@@ -222,6 +229,26 @@
...
         </dependency>
+        <dependency>
+            <groupId>com.google.apis</groupId>
+            <artifactId>google-api-services-admin-directory</artifactId>
+            <version>directory_v1-rev20220919-2.0.0</version>
+        </dependency>
+        <dependency>
+            <groupId>com.google.auth</groupId>
+            <artifactId>google-auth-library-oauth2-http</artifactId>
+            <version>1.19.0</version>
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

`google-auth-library-oauth2-http` (1.19.0) and `google-api-client` (2.2.0) are pinned with explicit versions, but these artifacts are also managed by the newly imported `libraries-bom` 26.32.0 (which `google-cloud-secretmanager` relies on). Overriding them with older explicit versions risks NoSuchMethod/version-skew at runtime against the BOM-aligned secretmanager client. Prefer removing the explicit versions and letting the BOM manage them, or verify the pinned versions are compatible with 26.32.0.

frontend/src/community/common/constants/routes.ts
```diff
@@ -181,4 +184,9 @@ export const employeeRestrictedRoutes = [
...
+  ROUTES.PEOPLE.GOOGLE_IMPORT_SYNCING,
```
ThinuwanW, line 190: 🤖 **Claude** · 🔴 **Important**

The new routes GOOGLE_IMPORT_SYNCING, GOOGLE_IMPORT_REVIEW and SYNC_CHANGES are placed in `nonSuperAdminRestrictedRoutes`, so middleware.ts only allows ROLE_SUPER_ADMIN. However the entry points into this flow are exposed to People Admins, not just super admins: the enterprise directory page gates the 'Import' button on `isPeopleAdmin` (useSessionData → user has ROLE_PEOPLE_ADMIN), and UploadTypeSelectModal.handleGoogleWorkspaceSelect does `router.push(ROUTES.PEOPLE.GOOGLE_IMPORT_REVIEW)` when Google is connected, while GoogleWorkspaceSyncBanner pushes to SYNC_CHANGES. A user who has ROLE_PEOPLE_ADMIN but not ROLE_SUPER_ADMIN will see the Import button, start the flow, then be redirected to /unauthorized as soon as they land on the review/syncing/sync-changes route. Either broaden the required role to include PEOPLE_ADMIN (add a second role check) or gate the UI entry points on `isSuperAdmin` so the visible affordances match the route authorization.

frontend/src/community/common/utils/notificationUtils.ts
```diff
@@ -163,6 +167,11 @@ export const handleNotifyRow = ({
     if (resourceId && !isNaN(Number(resourceId))) {
       router.push(ROUTES.SIGN.INBOX_INFO.ID(Number(resourceId)));
     }
+  } else if (
+    notificationType === NotificationItemsTypes.GOOGLE_WORKSPACE_USER_REMOVED ||
+    notificationType === NotificationItemsTypes.EXTERNAL_SYNC_COMPLETED
+  ) {
+    router.push(ROUTES.PEOPLE.SYNC_CHANGES);
```
ThinuwanW, line 174: 🤖 **Claude** · 🟡 **Suggestion**

handleNotifyRow navigates every recipient of a GOOGLE_WORKSPACE_USER_REMOVED / EXTERNAL_SYNC_COMPLETED notification to ROUTES.PEOPLE.SYNC_CHANGES, but that route is restricted to ROLE_SUPER_ADMIN via nonSuperAdminRestrictedRoutes (same root cause as the routes.ts finding). If these notifications are ever delivered to People Admins/Managers, clicking the notification will bounce them to /unauthorized instead of opening the sync-changes page. Confirm the backend only ever targets super admins with these notification types, or make the destination role-aware.

frontend/src/community/common/assets/Icons/GoogleIcon.tsx
```diff
@@ -14,30 +14,72 @@ const GoogleIcon = ({
...
-      />
-      <path
-        d="M13.5068 14.2234C11.1701 15.7233 8.05997 15.0449 6.56013 12.7082C6.33471 12.357 6.15389 11.9791 6.02185 11.5833L3.23779 13.7423C4.65211 16.5579 7.53398 18.3345 10.6848 18.3333C12.7134 18.3883 14.6863 17.666 16.1999 16.3143L13.5068 14.2234Z"
-        fill="#309B4D"
-      />
-      <path
-        d="M18.5335 8.48486H10.6851V11.7083H15.1699C14.9824 12.7387 14.3816 13.6474 13.507 14.2235C13.5044 14.2252 13.5016 14.2265 13.499 14.2282C13.5016 14.2265 13.5044 14.2252 13.507 14.2235L16.2001 16.3145L16.2002 16.3144C17.8588 14.717 18.7618 12.4909 18.685 10.1894C18.6855 9.61783 18.6348 9.04741 18.5335 8.48486Z"
-        fill="#4285F4"
-      />
+      <g clipPath="url(#clip0_300_7501)">
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

The new markup hardcodes internal SVG element ids (clip0_300_7501, paint0_radial_300_7501, paint1_radial_300_7501, paint2_linear_300_7501) referenced via url(#...). If more than one GoogleIcon renders on the same page (e.g. the option-card icon plus the connect-modal icon), these ids collide, producing invalid duplicate DOM ids; browsers resolve url(#id) to the first match in document order, so all instances share the first icon's gradients/clip. Consider suffixing the def ids with a unique/uid per instance (e.g. derived from the `id` prop) so each icon references its own defs.

frontend/src/community/common/assets/Icons/GoogleIcon.tsx
SanithuGunasena07: fixed, restored xmlns="http://www.w3.org/2000/svg"

frontend/src/fallback/people/components/molecules/ConnectGoogleWorkspaceModal/ConnectGoogleWorkspaceModal.tsx
SanithuGunasena07: fixed, stub now declares { isOpen, onClose } matching the real component's props.

frontend/src/community/common/constants/routes.ts
SanithuGunasena07: Fixed: removed ROUTES.PEOPLE.ADD from nonSuperAdminRestrictedRoutes in routes.ts

backend/pom.xml
SanithuGunasena07: removed the explicit versions, both now managed by the libraries-bom like google-cloud-secretmanager

frontend/src/community/common/constants/routes.ts
SanithuGunasena07: fixed

frontend/src/community/common/utils/notificationUtils.ts
SanithuGunasena07: Confirmed — both notification types are only ever created via notifySuperAdmins(), which queries isSuperAdmin = true recipients exclusively (EpGoogleWorkspacePersonSyncServiceImpl.java). No other code path creates them, so this is safe as-is.

frontend/src/community/common/assets/Icons/GoogleIcon.tsx
SanithuGunasena07: not needed

ThinuwanW (commented):
## 🤖 Claude Code Review

This PR wires up an enterprise Google Workspace people-sync feature: new Google dependencies, a WebSocket broadcast helper, notification enums/templates, route rewrites, and UI gating. The most significant issue is a missing fallback stub for the `GoogleWorkspaceSyncApi` hook now imported by a community page, which will break community-mode builds that don't check out the enterprise submodule. There is also an incidental access-control loosening on the add-people route and an unrelated removal of the CRM route guard in `middleware.ts`.

Found **7** new issue(s): 🔴 1 important, 🟡 4 suggestion(s), 🟣 2 nit(s)

frontend/pages/community/configurations/index.tsx
```diff
@@ -12,13 +12,19 @@ import { AdminTypes } from "~community/common/types/AuthTypes";
 import { getConfigurationTabs } from "~community/configurations/utils/configurationTabsUtil";
 import { useGetEnvironment } from "~enterprise/common/hooks/useGetEnvironment";
 import { getEnterpriseConfigurationTabs } from "~enterprise/configurations/utils/configurationTabsUtil";
+import { useGetGoogleConnectionStatus } from "~enterprise/people/api/GoogleWorkspaceSyncApi";
```
ThinuwanW, line 16: 🤖 **Claude** · 🔴 **Important**

`useGetGoogleConnectionStatus` is imported from `~enterprise/people/api/GoogleWorkspaceSyncApi`, but no fallback exists at `src/fallback/people/api/GoogleWorkspaceSyncApi.ts`. Per tsconfig, `~enterprise/*` resolves to `src/enterprise/*` then `src/fallback/*`; in a community-mode build where the enterprise submodule is a stub, this import cannot resolve and the build fails. The PR correctly added fallbacks for `ConnectGoogleWorkspaceModal` and `UploadTypeSelectModal` but omitted the API. Add `src/fallback/people/api/GoogleWorkspaceSyncApi.ts` exporting a no-op `useGetGoogleConnectionStatus` (e.g. returning a disabled `useQuery` / `{ data: undefined }` shape) so community builds compile.

frontend/middleware.ts
```diff
@@ -301,23 +301,14 @@ export function middleware(request: NextRequest) {
...
-    ) {
-      return NextResponse.redirect(
-        new URL(ROUTES.AUTH.UNAUTHORIZED, request.url)
-      );
-    }
-
-    // Check manager restricted routes
-    const managerRedirect = checkRestrictedRoutesAndRedirect(
+    // Check super-admin restricted routes
+    const nonSuperAdminRedirect = checkRestrictedRoutesAndRedirect(
```
ThinuwanW, line 319: 🤖 **Claude** · 🟡 **Suggestion**

This PR also deletes the explicit CRM guard (`pathname.startsWith(ROUTES.CRM.BASE) && !roles.includes(CRM_SALES_REPRESENTATIVE) -> UNAUTHORIZED`). CRM access is still gated by the `isAllowed`/`allowedRoutes` check (SUPER_ADMIN, CRM_ADMIN, CRM_SALES_MANAGER, CRM_SALES_REPRESENTATIVE), so the removal is not an obvious hole, but it changes CRM authorization behavior and is unrelated to the Google Workspace feature. Confirm this cleanup is intentional and belongs in this PR.

backend/src/main/java/com/skapp/community/common/component/WebSocketHandler.java
```diff
@@ -66,6 +66,19 @@ public void sendNotificationToUser(String userId, String message) {
 		}
 	}
 
+	public void broadcast(String message) {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`broadcast()` sends the message to every open session in the global `sessions` map, which is keyed only by userId with no tenant scoping. If this is invoked for tenant-scoped sync notifications in a multi-tenant deployment, it would deliver notifications to users of all tenants. It currently has no callers in the reviewed diff/codebase. Ensure the (enterprise) caller either scopes recipients to the relevant tenant or that broadcasting to all connected users is genuinely intended; otherwise add a tenant/user-filtered variant.

frontend/pages/community/configurations/index.tsx
```diff
@@ -59,6 +68,17 @@ const Configurations: NextPage = () => {
...
   };
 
+  // If the currently active tab disappears from visibleTabs (e.g. the People
+  // tab hides itself once auto-sync is disabled and the Google connection is
+  // torn down), fall back to the Attendance tab rather than showing a stale,
+  // no-longer-listed tab.
+  useEffect(() => {
+    if (!router.isReady || visibleTabs.length === 0) return;
+    if (activeTab && !visibleTabs.some((tab) => tab.id === activeTab)) {
+      handleTabChange("attendance");
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The fallback when the active tab disappears hardcodes `handleTabChange("attendance")`. If the current user (only super admins see the People tab today, so this is safe in practice) ever lacks the attendance tab, `activeTab` would be set to a tab not present in `visibleTabs`, and the effect could re-fire `router.replace` repeatedly. Prefer falling back to `visibleTabs[0]?.id` so the fallback always targets an actually-visible tab.

backend/src/main/java/com/skapp/community/common/component/WebSocketHandler.java
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`broadcast(String message)` is added but never invoked anywhere in the codebase — the enterprise sync service (`EpGoogleWorkspacePersonSyncServiceImpl`) sends its people-sync notifications via `notifySuperAdmins(...)`/`sendNotificationToUser`, not `broadcast`. This is dead code as of this PR. Either wire it into the sync notification flow (and address the tenant-scoping concern already raised on it before doing so) or remove it to avoid shipping an unused, un-tenant-scoped broadcast path to production.

frontend/src/community/people/components/organisms/DirectoryPopupController/DirectoryPopupController.tsx
```diff
@@ -15,9 +15,15 @@ import { usePeopleStore } from "~community/people/store/store";
 import { DirectoryModalTypes } from "~community/people/types/ModalTypes";
 import { QuickSetupModalTypeEnums } from "~enterprise/common/enums/Common";
 import { useCommonEnterpriseStore } from "~enterprise/common/store/commonStore";
+import ConnectGoogleWorkspaceModal from "~enterprise/people/components/molecules/ConnectGoogleWorkspaceModal/ConnectGoogleWorkspaceModal";
+import UploadTypeSelectModal from "~enterprise/people/components/molecules/UploadTypeSelectModal/UploadTypeSelectModal";
 
 const DirectoryPopupController = () => {
   const translatedTexts = useTranslator("peopleModule", "peoples");
+  const translateGoogleWorkspace = useTranslator(
```
ThinuwanW, line 23: 🤖 **Claude** · 🟣 **Nit**

This is a community-module component, but it now calls `useTranslator("peopleEnterprise", "googleWorkspaceImport")` — the `peopleEnterprise` namespace is only registered via the enterprise i18n bundle (`src/enterprise/common/assets/languages/english/english.ts`). In a community build that namespace is absent, so `translateGoogleWorkspace(["chooserTitle"])` would render the raw key instead of a label if the UPLOAD_TYPE_SELECT modal is ever reached. It's low-risk today because the Import flow is enterprise-only, but consider guarding this behind the enterprise-mode check (or moving the title logic into the enterprise `UploadTypeSelectModal`) so a community component isn't coupled to an enterprise translation namespace.

frontend/middleware.ts
```diff
@@ -301,23 +301,14 @@ export function middleware(request: NextRequest) {
...
       request,
-      managerRestrictedRoutes,
-      AdminTypes.PEOPLE_ADMIN,
+      nonSuperAdminRestrictedRoutes,
```
ThinuwanW, line 321: 🤖 **Claude** · 🟣 **Nit**

The new pages (`GOOGLE_IMPORT_SYNCING`, `GOOGLE_IMPORT_REVIEW`, `SYNC_CHANGES`) are added to `nonSuperAdminRestrictedRoutes` and checked against `AdminTypes.SUPER_ADMIN`, but they are never added to `superAdminRoutes` (or any entry in `allowedRoutes`). The restricted-route check only runs inside `if (isAllowed)`, and `isAllowed` requires a role whose allow-list is a prefix of the path. Since `superAdminRoutes` contains no `/people` prefix, a SUPER_ADMIN can only reach these pages if they also carry a role that grants `/people` access (e.g. PEOPLE_ADMIN via `adminRoutes`); otherwise they hit the default-deny redirect to UNAUTHORIZED before the super-admin gate is even evaluated. Please verify super admins reliably reach these new pages, and if not, add the `/people/directory/import-google` + `/people/directory/sync-changes` prefixes to `superAdminRoutes`.

frontend/pages/community/configurations/index.tsx
SanithuGunasena07: added src/fallback/people/api/GoogleWorkspaceSyncApi.ts exporting a no-op useGetGoogleConnectionStatus (returns { data: undefined }), matching the existing fallback stub convention in the same directory.

frontend/middleware.ts
SanithuGunasena07: Confirmed via git history — this wasn't intentional, it was dropped as a merge-conflict casualty (added in 9cec67a7d, disappeared in 5195b40c2 "after merge conflict resolved"). Restored it

backend/src/main/java/com/skapp/community/common/component/WebSocketHandler.java
SanithuGunasena07: no callers exist anywhere in the backend, including the Google Workspace notification path (which correctly uses sendNotificationToUser per recipient). Since broadcast() was unused dead code with no tenant scoping, removed it.

frontend/pages/community/configurations/index.tsx
SanithuGunasena07: switched to visibleTabs[0]?.id, guaranteeing the fallback always targets a tab that's actually visible to the current user.

frontend/src/community/people/components/organisms/DirectoryPopupController/DirectoryPopupController.tsx
SanithuGunasena07: removed

frontend/middleware.ts
SanithuGunasena07: Added the three route constants to superAdminRoutes

frontend/src/community/common/constants/routes.ts
```diff
@@ -67,7 +67,10 @@ const ROUTES = {
     PENDING: "/people/directory/pending",
     USER_ACCOUNT: "/user-account",
     EDIT: (id: any) => `/people/directory/edit/${id}`,
-    ADD: "/people/directory/add"
+    ADD: "/people/directory/add",
+    GOOGLE_IMPORT_SYNCING: "/people/directory/import-google/syncing",
+    GOOGLE_IMPORT_REVIEW: "/people/directory/import-google/review",
```
hasalarootcode, line 74: move to enterppise routes

backend/src/main/java/com/skapp/community/common/type/EmailBodyTemplates.java
```diff
@@ -104,7 +104,11 @@ public enum EmailBodyTemplates implements EmailTemplates {
 	// E-Sign Templates
 	ESIGN_DOCUMENT_SIGN_REQUEST("esign-document-sign-request"), ESIGN_DOCUMENT_COMPLETED("esign-document-completed"),
 	ESIGN_DOCUMENT_DECLINED("esign-document-declined"), ESIGN_DOCUMENT_VOIDED("esign-document-voided"),
-	ESIGN_DOCUMENT_REMINDER("esign-document-reminder"), ESIGN_DOCUMENT_EXPIRED("esign-document-expired");
+	ESIGN_DOCUMENT_REMINDER("esign-document-reminder"), ESIGN_DOCUMENT_EXPIRED("esign-document-expired"),
+
+	// People Module — External Sync Templates
+	PEOPLE_MODULE_EXTERNAL_SYNC_COMPLETED("people-module-external-sync-completed"),
```
yuhanga-rtc: move to people section

backend/pom.xml
```diff
@@ -222,6 +229,26 @@
...
+        </dependency>
+        <!-- version managed by the BOM above -->
+        <dependency>
+            <groupId>com.google.api-client</groupId>
+            <artifactId>google-api-client</artifactId>
+        </dependency>
+        <!-- GCP Secret Manager — version managed by the BOM above -->
+        <dependency>
+            <groupId>com.google.cloud</groupId>
+            <artifactId>google-cloud-secretmanager</artifactId>
```
yuhanga-rtc: remove if not needed

backend/pom.xml
```diff
@@ -28,6 +28,13 @@
                 <type>pom</type>
                 <scope>import</scope>
             </dependency>
+            <dependency>
+                <groupId>com.google.cloud</groupId>
```
yuhanga-rtc: Check this with the use case

yuhanga-rtc (changes requested):

frontend/pages/community/configurations/index.tsx
```diff
@@ -12,13 +12,19 @@ import { AdminTypes } from "~community/common/types/AuthTypes";
...
 import { getEnterpriseConfigurationTabs } from "~enterprise/configurations/utils/configurationTabsUtil";
+import { useGetGoogleConnectionStatus } from "~enterprise/people/api/GoogleWorkspaceSyncApi";
 
 const Configurations: NextPage = () => {
   const { user } = useAuth();
   const router = useRouter();
   const translateText = useTranslator("configurations");
   const environment = useGetEnvironment();
   const isEnterprise = environment === appModes.ENTERPRISE;
+  const isSuperAdmin = !!user?.roles?.includes(AdminTypes.SUPER_ADMIN);
```
hasalarootcode: use the existing hook which is available in your other places

frontend/pages/community/configurations/index.tsx
```diff
@@ -31,11 +37,14 @@ const Configurations: NextPage = () => {
   const visibleTabs = useMemo(() => {
     const userRoles = user?.roles || [];
     return allTabs.filter((tab) => {
-      return userRoles.some((role) =>
+      const hasRequiredRole = userRoles.some((role) =>
         tab.requiredRoles.includes(role as AdminTypes)
       );
+      if (!hasRequiredRole) return false;
```
hasalarootcode: let's try to simplify this

frontend/pages/community/configurations/index.tsx
```diff
@@ -59,6 +68,16 @@ const Configurations: NextPage = () => {
     );
   };
 
+  useEffect(() => {
+    if (!router.isReady || visibleTabs.length === 0) return;
```
hasalarootcode: move to a util and call here

frontend/pages/_app.tsx
```diff
@@ -90,13 +90,32 @@ function MyApp({
...
-  const shouldUseWebSocketProvider =
-    process.env.NEXT_PUBLIC_MODE !== appModes.ENTERPRISE;
+  const isEnterpriseMode = process.env.NEXT_PUBLIC_MODE === appModes.ENTERPRISE;
 
   return (
     <div className={inter.className}>
       <AuthProvider>
-        {shouldUseWebSocketProvider ? (
+        {isEnterpriseMode ? (
+          <WebSocketProvider>
```
hasalarootcode: remove websocket provider from enterprise

frontend/src/community/common/assets/Icons/GoogleIcon.tsx
```diff
@@ -9,35 +9,86 @@ const GoogleIcon = ({
   svgProps,
   onClick
 }: IconProps): JSX.Element => {
+  // Suffix the def ids with a per-instance uid so multiple GoogleIcons on the
+  // same page don't collide on url(#...) references (browsers resolve to the
+  // first matching id in document order otherwise).
+  const uid = useId();
```
hasalarootcode: remove these

frontend/src/community/common/types/ToastTypes.ts
```diff
@@ -7,7 +7,7 @@ export interface ToastProps {
   title: string;
   description?: string;
   toastType?: ToastType;
-  autoHideDuration?: number;
+  autoHideDuration?: number | null;
```
hasalarootcode, line 10: check this, if you cannot handle without adding this null

hasalarootcode (changes requested):

backend/src/main/java/com/skapp/community/common/type/EmailBodyTemplates.java
SanithuGunasena07: Moved

backend/pom.xml
SanithuGunasena07: Removed as it was dead code

backend/pom.xml
SanithuGunasena07: removed as it was not required

frontend/src/community/common/constants/routes.ts
SanithuGunasena07: Moved to a new enterprise/people/constants/routes.ts

frontend/pages/community/configurations/index.tsx
SanithuGunasena07: Switched to useSessionData()'s isSuperAdmin (already used the same way elsewhere, e.g. UploadTypeSelectModal.tsx).

frontend/pages/community/configurations/index.tsx
SanithuGunasena07: Simplified to one line, iterating tab.requiredRoles instead of userRoles — same check, but it also removes the need for the role as AdminTypes cast, since AdminTypes is already assignable into userRoles' broader union type.

frontend/pages/community/configurations/index.tsx
SanithuGunasena07: Extracted into getFallbackTabId() in configurationTabsUtil.tsx, alongside getConfigurationTabs. Takes visibleTabs/activeTabId, returns undefined if the current tab is still valid or the first visible tab's id otherwise.

frontend/src/community/common/assets/Icons/GoogleIcon.tsx
SanithuGunasena07: Removed

frontend/src/community/common/types/ToastTypes.ts
SanithuGunasena07: it's necessary, not optional. autoHideDuration flows straight into MUI's <Snackbar>, whose own type is number | null, where null means "never auto-dismiss." That's used for the two persistent error toasts in GoogleWorkspaceImportReview.tsx (import/update failures) that should stay open until manually closed there's no numeric value that means "never." Our type just mirrors MUI's own contract.

backend/pom.xml
SanithuGunasena07: Correction: that removal was wrong. The libraries-bom wasn't only versioning google-cloud-secretmanager it also manages google-auth-library-oauth2-http and google-api-client, which are genuinely used (Google OAuth/Calendar integration). Restored the BOM import; kept google-cloud-secretmanager itself removed since that one really is unused.

SanithuGunasena07: the BOM was managing versions for three dependencies, not just the one flagged (google-cloud-secretmanager, unused, removed). The other two (google-auth-library-oauth2-http, google-api-client) are genuinely used for the Google OAuth/Calendar integration.

Re-importing the whole libraries-bom to fix that turned out to be worse it pulls in a huge transitive dependency tree (gRPC, protobuf, guava, etc.) and caused the entire build to fail to resolve dependencies (even unrelated ones like lombok/jakarta.persistence).

Final fix: dropped the BOM entirely and pinned explicit versions directly on just those two dependencies (google-auth-library-oauth2-http:1.24.0, google-api-client:2.7.0) instead. google-cloud-secretmanager stays removed.

frontend/pages/_app.tsx
umeshaliyanage: removed webhook

backend/src/main/java/com/skapp/community/common/component/WebSocketHandler.java
umeshaliyanage: removed broadcast when webhook was removed

frontend/src/community/common/constants/routes.ts
hasalarootcode: suggestion is wrong based on current approach, lets keep it community

frontend/pages/community/configurations/index.tsx
hasalarootcode: Let's use a util for this specific case without changing the existing implementation

frontend/src/community/common/constants/routes.ts
SanithuGunasena07: moved the three routes back into ROUTES.PEOPLE in community/common/constants/routes.ts and removed the enterprise/people/constants/routes.ts file. Updated all 9 call sites back to ROUTES.PEOPLE.

frontend/pages/community/configurations/index.tsx
SanithuGunasena07: Extracted into hasRequiredRole() in configurationTabsUtil.tsx

backend/src/main/java/com/skapp/community/common/util/StringUtils.java
```diff
@@ -21,4 +28,21 @@ public static String escapeLikePattern(String input) {
...
 	}
 
+	public static String convertToCommaSeperatedString(Set<String> values) {
+		if (values == null || values.isEmpty()) {
+			return null;
+		}
+		return values.stream().filter(Objects::nonNull).collect(Collectors.joining(EpCommonConstants.COMMA_DELIMITER));
+	}
+
+	public static Set<String> convertToList(String value) {
```
yuhanga-rtc, line 38: Can we check this to simplify

backend/pom.xml
```diff
@@ -222,7 +222,21 @@
...
             <version>v1-rev20240313-2.0.0</version>
         </dependency>
-
+        <dependency>
+            <groupId>com.google.apis</groupId>
+            <artifactId>google-api-services-admin-directory</artifactId>
+            <version>directory_v1-rev20220919-2.0.0</version>
+        </dependency>
+        <dependency>
+            <groupId>com.google.auth</groupId>
```
yuhanga-rtc, line 231: Remove after using existing library in implementation

backend/pom.xml
```diff
@@ -222,7 +222,21 @@
             <artifactId>google-api-services-people</artifactId>
             <version>v1-rev20240313-2.0.0</version>
         </dependency>
-
+        <dependency>
+            <groupId>com.google.apis</groupId>
+            <artifactId>google-api-services-admin-directory</artifactId>
+            <version>directory_v1-rev20220919-2.0.0</version>
```
yuhanga-rtc: Check latest version

yuhanga-rtc (changes requested):

backend/src/main/java/com/skapp/community/common/util/StringUtils.java
SanithuGunasena07: Simplified convertToList was duplicating the exact null-or-blank check that isNullOrBlank already does right above it in the same class. Now just calls isNullOrBlank(value).

backend/pom.xml
SanithuGunasena07: done

hasalarootcode (dismissed):

backend/pom.xml
SanithuGunasena07: needed for now

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/EmployeeRepositoryImpl.java
```diff
@@ -111,7 +111,10 @@ public Page<Employee> findEmployees(EmployeeFilterDto employeeFilterDto, Pageabl
 		List<Predicate> predicates = new ArrayList<>();
 
 		predicates.add(criteriaBuilder.notEqual(root.get(Employee_.ACCOUNT_STATUS), AccountStatus.DELETED));
-		predicates.add(criteriaBuilder.notEqual(roleJoin.get(EmployeeRole_.PM_ROLE), Role.PM_GUEST_EMPLOYEE));
+		// predicates.add(criteriaBuilder.notEqual(roleJoin.get(EmployeeRole_.PM_ROLE),
```
yuhanga-rtc: Remove comment

SanithuGunasena07: removed

yuhanga-rtc (dismissed):

backend/src/main/java/com/skapp/community/common/type/NotificationCategory.java
```diff
@@ -7,7 +7,7 @@
 @AllArgsConstructor
 public enum NotificationCategory {
 
-	PEOPLE("people"), LEAVE("leave"), ATTENDANCE("attendance"), OKR("okr"), ESIGN("esign");
+	PEOPLE("people"), LEAVE("leave"), ATTENDANCE("attendance"), OKR("okr"), ESIGN("esign"), PEOPLE_SYNC("people_sync");
```
akilarootcode: PEOPLE_SYNC remove

backend/src/main/java/com/skapp/community/common/util/StringUtils.java
```diff
@@ -21,4 +27,23 @@ public static String escapeLikePattern(String input) {
 		return ValidationConstant.LIKE_WILDCARD_PATTERN.matcher(input).replaceAll("\\\\$1");
 	}
 
+	private static final String COMMA_DELIMITER = ",";
```
akilarootcode: move to top of the method

backend/src/main/java/com/skapp/community/common/util/StringUtils.java
```diff
@@ -21,4 +27,23 @@ public static String escapeLikePattern(String input) {
...
+
+	public static String convertToCommaSeperatedString(Set<String> values) {
+		if (values == null || values.isEmpty()) {
+			return null;
+		}
+		return values.stream().filter(Objects::nonNull).collect(Collectors.joining(COMMA_DELIMITER));
```
akilarootcode: remove .filter(Objects::nonNull)

backend/src/main/java/com/skapp/community/common/util/StringUtils.java
```diff
@@ -21,4 +27,23 @@ public static String escapeLikePattern(String input) {
...
+	}
+
+	public static Set<String> convertToList(String value) {
+		if (isNullOrBlank(value)) {
+			return new LinkedHashSet<>();
+		}
+		return Arrays.stream(value.split(COMMA_DELIMITER))
+			.map(String::trim)
```
akilarootcode: .map(String::trim)
			.filter(s -> !s.isEmpty()) remove

backend/pom.xml
```diff
@@ -222,7 +222,21 @@
...
-
+        <dependency>
+            <groupId>com.google.apis</groupId>
+            <artifactId>google-api-services-admin-directory</artifactId>
+            <version>directory_v1-rev20260531-2.0.0</version>
+        </dependency>
+        <dependency>
+            <groupId>com.google.auth</groupId>
+            <artifactId>google-auth-library-oauth2-http</artifactId>
+            <version>1.24.0</version>
```
akilarootcode: update versions if possible

backend/pom.xml
```diff
@@ -222,7 +222,21 @@
...
+        </dependency>
+        <dependency>
+            <groupId>com.google.api-client</groupId>
+            <artifactId>google-api-client</artifactId>
```
akilarootcode, line 237: update to latest version

frontend/src/community/configurations/utils/configurationTabsUtil.tsx
```diff
@@ -7,6 +8,21 @@ import TimeConfigurations from "~community/configurations/components/organisms/T
 import WorkLocationsTable from "~community/configurations/components/organisms/WorkLocationsTable/WorkLocationsTable";
 import { ConfigurationTab } from "~community/configurations/types/ConfigurationTabTypes";
 
+export const hasRequiredRole = (
+  tab: ConfigurationTab,
+  userRoles: NonNullable<User["roles"]>
```
akilarootcode: remove this and use Enum directly

frontend/pages/community/configurations/index.tsx
```diff
@@ -59,6 +69,14 @@ const Configurations: NextPage = () => {
...
+    const fallbackTabId = getFallbackTabId(visibleTabs, activeTab);
```
akilarootcode: remove useEffect and onSuccess or OnError do tab redirection handling part.

frontend/pages/_app.tsx
```diff
@@ -154,4 +154,4 @@ MyApp.getInitialProps = async (appContext: AppContext) => {
   return { ...appProps };
 };
 
-export default MyApp;
+export default MyApp;
```
akilarootcode: remove enter

frontend/src/community/common/assets/languages/english/peopleModule.json
```diff
@@ -309,6 +309,7 @@
     },
     "addPeople": "Add People",
     "addBulkPeople": "Bulk Upload",
+    "importPeople": "Import",
```
akilarootcode: change key also to import

frontend/src/community/common/utils/notificationUtils.ts
```diff
@@ -15,6 +15,23 @@ import {
   NotificationSummaryType
 } from "~community/common/types/notificationTypes";
 
+interface NotificationTypeHandler {
+  icon: FC<IconProps>;
+  getRoute: () => string;
+}
```
akilarootcode: ### Need to check this again ###

frontend/src/community/common/utils/notificationUtils.ts
```diff
@@ -43,7 +60,9 @@ export const getNotificationIcon = (
     case NotificationItemsTypes.ESIGN_DOCUMENT_DECLINED_OWNER:
       return EsignatureModuleIcon;
     default:
```
akilarootcode, line 49: handle all notification scenarios using case:

frontend/src/community/common/utils/notificationUtils.ts
```diff
@@ -163,6 +182,11 @@ export const handleNotifyRow = ({
     if (resourceId && !isNaN(Number(resourceId))) {
       router.push(ROUTES.SIGN.INBOX_INFO.ID(Number(resourceId)));
     }
+  } else if (notificationType) {
+    const handler = notificationTypeHandlerRegistry.get(notificationType);
```
akilarootcode: remove this

frontend/src/community/configurations/utils/configurationTabsUtil.tsx
```diff
@@ -7,6 +8,21 @@ import TimeConfigurations from "~community/configurations/components/organisms/T
...
+): boolean => tab.requiredRoles.some((role) => userRoles.includes(role));
+
+export const getFallbackTabId = (
```
akilarootcode: remove this as well

frontend/src/community/people/components/organisms/DirectoryPopupController/DirectoryPopupController.tsx
```diff
@@ -50,7 +56,7 @@ const DirectoryPopupController = () => {
       case DirectoryModalTypes.ADD_NEW_RESOURCE:
         return "Add people";
       case DirectoryModalTypes.UPLOAD_TYPE_SELECT:
-        return translatedTexts(["uploadTypeSelectorModalTitle"]);
+        return translateGoogleWorkspace(["chooserTitle"]);
```
akilarootcode, line 61: move this to separate case: and revert

frontend/src/community/people/components/organisms/DirectoryPopupController/DirectoryPopupController.tsx
```diff
@@ -122,9 +128,34 @@ const DirectoryPopupController = () => {
         DirectoryModalTypes.GUEST_TO_INTERNAL_USER_CONFIRMATION && (
         <GuestToInternalUserConfirmationModal />
       )}
+      {directoryModalType === DirectoryModalTypes.UPLOAD_TYPE_SELECT && (
+        <UploadTypeSelectModal />
```
akilarootcode, line 134: render conditionally using isEnable inside the component

akilarootcode (changes requested):

backend/src/main/java/com/skapp/community/common/type/NotificationCategory.java
SanithuGunasena07: Removed PEOPLE_SYNC, now using NotificationCategory.PEOPLE for the sync notification

backend/src/main/java/com/skapp/community/common/util/StringUtils.java
SanithuGunasena07: Moved

backend/src/main/java/com/skapp/community/common/util/StringUtils.java
SanithuGunasena07: Removed .filter(Objects::nonNull)

backend/src/main/java/com/skapp/community/common/util/StringUtils.java
SanithuGunasena07: Keeping this - group names can contain commas (e.g. "APAC, Region 1"), and since we join/split on commas with no escaping, a name like that gets accidentally split apart, leaving a stray space or blank entry behind. .trim() and the empty-check clean up that leftover mess.

backend/pom.xml
SanithuGunasena07: Updated to 1.48.0

backend/pom.xml
SanithuGunasena07: Updated to 2.8.0

frontend/src/community/configurations/utils/configurationTabsUtil.tsx
SanithuGunasena07: done

frontend/pages/_app.tsx
SanithuGunasena07: removed

frontend/src/community/common/assets/languages/english/peopleModule.json
SanithuGunasena07: Renamed

frontend/src/community/common/utils/notificationUtils.ts
SanithuGunasena07: the file had zero importers, so its registration code never ran, leaving these notification types with no icon or click-navigation. Fixed by importing it in _app.tsx.

frontend/src/community/common/utils/notificationUtils.ts
SanithuGunasena07: removed the registry (NotificationTypeHandler/notificationTypeHandlerRegistry/registerNotificationTypeHandler) entirely and added explicit case branches for EXTERNAL_SYNC_COMPLETED/GOOGLE_WORKSPACE_USER_REMOVED in both the icon switch and the click-navigation logic, matching how Leave/Attendance/E-Sign already work. Deleted the now-unnecessary registerPersonSyncNotifications.ts and its _app.tsx import.

frontend/src/community/common/utils/notificationUtils.ts
SanithuGunasena07: fixed as part of fixing https://github.com/SkappHQ/skapp/pull/1635#discussion_r3663322638

frontend/src/community/configurations/utils/configurationTabsUtil.tsx
SanithuGunasena07: fixed

frontend/src/community/people/components/organisms/DirectoryPopupController/DirectoryPopupController.tsx
SanithuGunasena07: @umeshaliyanage

frontend/src/community/people/components/organisms/DirectoryPopupController/DirectoryPopupController.tsx
SanithuGunasena07: @umeshaliyanage

frontend/pages/community/configurations/index.tsx
SanithuGunasena07: @umeshaliyanage

frontend/src/community/people/components/organisms/DirectoryPopupController/DirectoryPopupController.tsx
umeshaliyanage: moved chooserTitle to separate case

frontend/src/community/people/components/organisms/DirectoryPopupController/DirectoryPopupController.tsx
umeshaliyanage: added isEnable to connectGoogleWorkspace modal and handled conditionally

frontend/pages/community/configurations/index.tsx
umeshaliyanage: removed the fallback because if the query got undefined, it will fallback to tab 1

backend/src/main/java/com/skapp/community/common/util/StringUtils.java
akilarootcode: check this again

backend/src/main/java/com/skapp/community/common/util/MessageUtil.java
```diff
@@ -21,11 +21,11 @@ public String getMessage(MessageConstant messageConstant) {
 		return messageSource.getMessage(messageConstant.getMessageKey(), null, Locale.getDefault());
 	}
 
-	public String getMessage(String code, Object[] args) {
+	public String getMessage(String code, Object... args) {
 		return messageSource.getMessage(code, args, Locale.getDefault());
 	}
 
-	public String getMessage(MessageConstant messageConstant, Object[] args) {
+	public String getMessage(MessageConstant messageConstant, Object... args) {
```
akilarootcode: roleback

backend/src/main/resources/community/templates/notification/notification-templates.json
```diff
@@ -317,6 +317,29 @@
         "title": "Document expired",
         "message": "{{documentName}} is expiring today. Please review and sign."
       }
+    ],
+
```
akilarootcode: remove whitespace

backend/src/main/resources/community/templates/notification/notification-templates.json
```diff
@@ -317,6 +317,29 @@
...
+    "people_sync": [
```
akilarootcode: move to people

frontend/pages/_app.tsx
akilarootcode: check again

akilarootcode (changes requested):

backend/src/main/java/com/skapp/community/common/util/MessageUtil.java
SanithuGunasena07: Reverted

backend/src/main/resources/community/templates/notification/notification-templates.json
SanithuGunasena07: removed

backend/src/main/resources/community/templates/notification/notification-templates.json
SanithuGunasena07: moved

backend/src/main/java/com/skapp/community/common/util/StringUtils.java
SanithuGunasena07: removed

backend/src/main/java/com/skapp/community/common/util/MessageUtil.java
```diff
@@ -21,7 +21,7 @@ public String getMessage(MessageConstant messageConstant) {
 		return messageSource.getMessage(messageConstant.getMessageKey(), null, Locale.getDefault());
 	}
 
-	public String getMessage(String code, Object[] args) {
+	public String getMessage(String code, Object... args) {
```
akilarootcode: roleback this again

backend/src/main/java/com/skapp/community/timeplanner/type/TimeRecordSource.java
```diff
@@ -4,4 +4,4 @@ public enum TimeRecordSource {
 
 	WEB, FINGERPRINT_DEVICE
 
-}
+}
```
akilarootcode: revert this also

frontend/pages/_app.tsx
```diff
@@ -153,5 +153,4 @@ MyApp.getInitialProps = async (appContext: AppContext) => {
 
   return { ...appProps };
 };
-
-export default MyApp;
+export default MyApp;
```
akilarootcode: revert this also

akilarootcode (changes requested):

akilarootcode:
@SanithuGunasena07 @umeshaliyanage check the build failure

backend/src/main/java/com/skapp/community/common/util/MessageUtil.java
SanithuGunasena07: rolebacked

backend/src/main/java/com/skapp/community/timeplanner/type/TimeRecordSource.java
SanithuGunasena07: done

frontend/pages/_app.tsx
SanithuGunasena07: done

akilarootcode (dismissed):

akilarootcode (approved):

PR #1650, opened 2026-07-20
fix: increment open deals count for company upon deal creation

YasinduRC:
### Summary
Fixing the issue of CRM - Companies Detailed View - When Deals are added from within the Detailed view, Open Deals count do not increment until the view is refreshed and the detailed panel is reopened

Updating the store by one when the a new deal  added via the Company Detail View, this is scoped only for the company detail panel 

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

The PR correctly adds an optimistic openDeals increment to the CRM store when a deal is created from the inline SidePanelAddDeal, following existing immutable-update patterns with proper null safety and type coverage. The core logic is sound and reconciles cleanly with server refetches via mergeWithExisting, so no double-counting occurs. Findings are minor: a missing unit test for the new utility and an inconsistency with the other deal-creation entry point.

Found **3** new issue(s): 🔴 1 important, 🟡 2 suggestion(s)

frontend/src/community/crm/utils/companyUtil.ts
```diff
@@ -10,6 +10,16 @@ export const mergeCompanyUpdate = (
     company.id === update.id ? { ...company, ...update } : company
   );
 
+export const withIncrementedOpenDeals = (
```
ThinuwanW, line 13: 🤖 **Claude** · 🟡 **Suggestion**

The new `withIncrementedOpenDeals` utility has no unit test, even though this file already has a colocated test (`companyUtil.test.ts`) covering `mapCompanyToMetricItems`. Add a test verifying: (1) the matching company's `openDeals` is incremented by 1, (2) a company with `openDeals: null` becomes 1 (exercising the `?? 0` branch), (3) non-matching companies are left unchanged, and (4) the result is a new array (immutability). This util is now the single source of truth for the metric's optimistic update, so it should be covered.

frontend/src/community/crm/components/molecules/SidePanelAddDeal/SidePanelAddDeal.tsx
```diff
@@ -66,6 +71,11 @@ const SidePanelAddDeal: FC<Props> = ({ onClose, defaultContact }) => {
   };
 
   const handleCreateDealSuccess = () => {
+    const companyId = selectedContact?.company?.id;
```
ThinuwanW, line 74: 🤖 **Claude** · 🔴 **Important**

The deal-create flow updates only the Zustand store (`incrementCompanyOpenDeals`) but never invalidates the company-metrics React Query cache. `useCreateDeal.onSuccess` (crmDealApi.ts) invalidates deal/contact/limitation queries, yet creating a deal also changes company data (`openDeals`), and `companyQueryKeys.GET_COMPANY_DATA` is not invalidated. Because `CompanyTable` repopulates the store from that cache via `setCompanies` → `mergeWithExisting` (which does `{...current, ...incoming}`, so the server value wins), when the table remounts the cached pre-deal metrics briefly overwrite the incremented `openDeals` before the background refetch corrects it — a visible flicker/revert. This is the exact 'forgot invalidateQueries after a mutation that affects cached data' case in the guidelines. Fix: in `useCreateDeal.onSuccess` add `queryClient.invalidateQueries({ queryKey: companyQueryKeys.GET_COMPANY_DATA })` (server-authoritative), or keep the optimistic path but also mirror it into the cache with `queryClient.setQueryData` so store and cache stay in sync.

frontend/src/community/crm/utils/companyUtil.ts
ThinuwanW, line 13: 🤖 **Claude** · 🟡 **Suggestion**

`withIncrementedOpenDeals` (and therefore the whole feature) is a silent no-op when the target company is not present in the store's `companies` array. `SidePanelAddDeal` is rendered from both `CompanySidePanel` and `ContactSidePanel` (via `SidePanelDealSection`); when a deal is added from a context where the company list/table was never loaded (e.g. the contact side panel), `companies` won't contain the company, the `.map` matches nothing, and the count is never updated. Combined with the missing company-metrics invalidation above, there is no fallback that would eventually correct it. Prefer refreshing the count via query invalidation (see the finding on SidePanelAddDeal.tsx) rather than relying on the company already being in the store, or document that this optimistic path only applies when the company is loaded.

YasinduRC: Added a unit test

frontend/src/community/crm/components/molecules/SidePanelAddDeal/SidePanelAddDeal.tsx
YasinduRC: openDeals is only shown in the side panel's metric card, not in the companies table,  there's nothing on the table to revert when it remounts, and the panel re-reads from the store after the list has already refetched. 

So there is no need of query invalidation

frontend/src/community/crm/utils/companyUtil.ts
YasinduRC: Since the increment function calls on the add deal button inside the company side panel, there is no way to open a side panel if the company is not in the store

anusham-rc, line 13: check and refactor this with getCompanyById

YasinduRC: It invokes a slice method from within the utility, but implementing it manually here results in about the same level of complexity.

PR #1651, opened 2026-07-20
fix: merge the templates instead of replacing

gaindurootcode:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/backend/src/main/java/com/skapp/community/common/service/impl/EmailServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/common/service/impl/EmailServiceImpl.java
@@ -16,6 +16,7 @@ import tools.jackson.dataformat.yaml.YAMLMapper;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
@@ -152,13 +153,24 @@ public class EmailServiceImpl implements EmailService {
 						String innerKey = innerEntry.getKey();
 						List<EmailTemplateMetadata> metadataList = innerEntry.getValue();
 
-						templateDetailsMap.get(outerKey).put(innerKey, metadataList);
+						mergeModuleTemplates(templateDetailsMap.get(outerKey), innerKey, metadataList);
 					}
 				}
 			}
 		}
 		catch (IOException e) {
 			log.warn("Failed to load templates from {}: {}", path, e.getMessage());
+		}
+	}
+
+	private void mergeModuleTemplates(Map<String, List<EmailTemplateMetadata>> moduleTemplates, String module,
+			List<EmailTemplateMetadata> incomingTemplates) {
+		List<EmailTemplateMetadata> existingTemplates = moduleTemplates.computeIfAbsent(module,
+				key -> new ArrayList<>());
+
+		for (EmailTemplateMetadata incomingTemplate : incomingTemplates) {
+			existingTemplates.removeIf(template -> template.getId().equals(incomingTemplate.getId()));
+			existingTemplates.add(incomingTemplate);
 		}
 	}
 
--- a/backend/src/main/java/com/skapp/community/common/service/impl/EmailServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/common/service/impl/EmailServiceImpl.java
@@ -22,6 +22,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 
 @Service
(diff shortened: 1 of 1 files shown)


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR correctly fixes a real bug: when the enterprise EmailServiceImpl subclass loads community then enterprise template YAMLs, the old `put(innerKey, metadataList)` overwrote a module's entire template list, discarding community-only templates in that module. The new `mergeModuleTemplates` helper dedupes by template id and appends, preserving overrides while keeping non-overlapping templates, and uses a fresh mutable ArrayList. The change is well-scoped and behavior-preserving for the currently shipped data (where enterprise is a superset), while protecting against template loss in the general case.

Found **4** new issue(s): 🟣 3 nit(s)

backend/src/main/java/com/skapp/community/common/service/impl/EmailServiceImpl.java
```diff
@@ -162,6 +163,17 @@ protected void addTemplatesFromPath(String path) {
 		}
 	}
 
+	private void mergeModuleTemplates(Map<String, List<EmailTemplateMetadata>> moduleTemplates, String module,
+			List<EmailTemplateMetadata> incomingTemplates) {
+		List<EmailTemplateMetadata> existingTemplates = moduleTemplates.computeIfAbsent(module,
+				key -> new ArrayList<>());
+
+		for (EmailTemplateMetadata incomingTemplate : incomingTemplates) {
+			existingTemplates.removeIf(template -> template.getId().equals(incomingTemplate.getId()));
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

Null-safety nit: `template.getId().equals(incomingTemplate.getId())` will throw NPE if any existing template has a null `id` (EmailTemplateMetadata.id has no @NotNull and YAML could omit it). All shipped YAML entries currently define an id, so this is low risk, but for robustness prefer `java.util.Objects.equals(template.getId(), incomingTemplate.getId())`, which is null-safe on both sides. The developer can decide whether the controlled YAML source makes this unnecessary.

ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Asymmetric null-id handling (distinct from the first-pass finding, which only covers the *existing* template's id). Here `incomingTemplate.getId()` is passed to `equals()`: if an incoming template has a null id, `existing.getId().equals(null)` returns false for every existing entry, so no dedup happens and the null-id template is silently appended — allowing duplicate/unremovable entries to accumulate across loads. Switching the comparison to `java.util.Objects.equals(template.getId(), incomingTemplate.getId())` handles both directions safely and also resolves the first-pass NPE, making it the single clean fix for the whole comparison.

backend/src/main/java/com/skapp/community/common/service/impl/EmailServiceImpl.java
```diff
@@ -162,6 +163,17 @@ protected void addTemplatesFromPath(String path) {
 		}
 	}
 
+	private void mergeModuleTemplates(Map<String, List<EmailTemplateMetadata>> moduleTemplates, String module,
+			List<EmailTemplateMetadata> incomingTemplates) {
+		List<EmailTemplateMetadata> existingTemplates = moduleTemplates.computeIfAbsent(module,
+				key -> new ArrayList<>());
+
+		for (EmailTemplateMetadata incomingTemplate : incomingTemplates) {
```
ThinuwanW, line 176: 🤖 **Claude** · 🟣 **Nit**

Behavioral drift / new unhandled NPE. The old code (`...put(innerKey, metadataList)`) tolerated a null value list at load time (it would only fail later at query time). The new enhanced-for loop `for (EmailTemplateMetadata incomingTemplate : incomingTemplates)` throws a NullPointerException immediately if a module key in the YAML has an empty/null value (e.g. `moduleName:` with no children). That NPE is not an IOException, so it escapes the try/catch in `addTemplatesFromPath` and propagates uncaught. Add a guard (e.g. `if (incomingTemplates == null) return;` or `Collections.emptyList()` fallback) to preserve the previous tolerance.

backend/src/main/java/com/skapp/community/common/service/impl/EmailServiceImpl.java
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

O(n*m) merge: `existingTemplates.removeIf(...)` performs a full ArrayList scan (plus element shifts) for every incoming template, inside the loop over all incoming templates. Template lists are small so this is not urgent, but if the intent is dedup-by-id, a `LinkedHashMap<String, EmailTemplateMetadata>` keyed by id (preserving insertion order, put overwrites) would give O(1) upsert and clearer semantics than remove-then-add.

gaindurootcode: switched to Objects.equals(template.getId(), incomingTemplate.getId()), which is null-safe on both sides.

gaindurootcode: Fixed with Objects.equals(...)

backend/src/main/java/com/skapp/community/common/service/impl/EmailServiceImpl.java
gaindurootcode: added an early if (incomingTemplates == null) return; guard to preserve the previous tolerance.

backend/src/main/java/com/skapp/community/common/service/impl/EmailServiceImpl.java
gaindurootcode: Not an issue.

shakila-kamalasena (approved):

akilarootcode (approved):

ThinuwanW (approved):

PR #1657, opened 2026-07-21
Fix/2505 1982 2372 notification, teams and add holiday modal ui issues

shakila-kamalasena:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/frontend/src/community/common/components/organisms/AppBar/AppBar.tsx
+++ b/frontend/src/community/common/components/organisms/AppBar/AppBar.tsx
@@ -76,9 +76,13 @@ const AppBar = () => {
   const { data: unreadCount } = useGetUnreadNotificationsCount();
 
   useEffect(() => {
+    if (unreadCount === undefined) return;
+
     setNotifyData({
-      unreadCount: unreadCount || 0
+      unreadCount
     });
+
+    localStorage.setItem("unReadMsgCount", unreadCount.toString());
   }, [setNotifyData, unreadCount]);
 
   useEffect(() => {
--- a/frontend/src/community/common/components/organisms/AppBar/AppBar.tsx
+++ b/frontend/src/community/common/components/organisms/AppBar/AppBar.tsx
@@ -76,7 +76,7 @@ const AppBar = () => {
   const { data: unreadCount } = useGetUnreadNotificationsCount();
 
   useEffect(() => {
-    if (unreadCount === undefined) return;
+    if (unreadCount == null) return;
 
     setNotifyData({
       unreadCount
--- a/frontend/src/community/common/assets/languages/english/peopleModule.json
+++ b/frontend/src/community/common/assets/languages/english/peopleModule.json
@@ -50,7 +50,7 @@
     "noOfSupervisorsExceedingToastMessageDescription": "Only 3 supervisors are allowed per team.",
     "emptySearchResult": {
       "title": "No results found",
-      "description": "No results found for the requested search query. Please retry."
+      "description": "We couldn't find any matches for your search. Please try using a different term."
     },
     "emptyScreen": {
       "title": "No teams available",
--- a/frontend/src/community/common/components/molecules/DurationSelector/DurationSelector.tsx
+++ b/frontend/src/community/common/components/molecules/DurationSelector/DurationSelector.tsx
@@ -200,7 +200,7 @@ const DurationSelector = <T,>({
     >
       <Stack sx={classes.container}>
         <Typography
-          variant="label"
+          variant="body1"
           sx={{
             color: error ? theme.palette.error.contrastText : "common.black"
           }}
--- a/frontend/src/community/common/components/molecules/DurationSelector/DurationSelector.tsx
+++ b/frontend/src/community/common/components/molecules/DurationSelector/DurationSelector.tsx
@@ -200,12 +200,12 @@ const DurationSelector = <T,>({
     >
       <Stack sx={classes.container}>
         <Typography
-          variant="body1"
+          variant="label"
           sx={{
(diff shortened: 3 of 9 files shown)


ThinuwanW (commented):
## 🤖 Claude Code Review

The TeamsTable empty-state refactor is a correct improvement — it fixes the prior bug where a search yielding no matches (while teams existed) showed the 'No teams available' screen with an Add button, by correctly using the Table's isSearching/noSearchResults API. The i18n copy change is fine (only the english locale exists). The AppBar change is functionally sound but introduces a mode-inconsistency: it writes the unReadMsgCount localStorage key unconditionally, whereas the corresponding decrement (useMarkNotificationAsRead) and reads are all enterprise-gated for the FCM push flow.

Found **4** new issue(s): 🔴 1 important, 🟣 1 nit(s)

frontend/src/community/common/components/organisms/AppBar/AppBar.tsx
```diff
@@ -76,9 +76,13 @@ const AppBar = () => {
...
 
   useEffect(() => {
+    if (unreadCount === undefined) return;
+
     setNotifyData({
-      unreadCount: unreadCount || 0
+      unreadCount
     });
+
+    localStorage.setItem("unReadMsgCount", unreadCount.toString());
```
ThinuwanW, line 85: 🤖 **Claude** · 💬 **Note**

The `unReadMsgCount` localStorage key exists to seed the enterprise FCM push-notification baseline: it's read in `useFCMToken.ts` (enterprise) and the decrement in `useMarkNotificationAsRead` (notificationsApi.ts) is explicitly gated behind `appModes.ENTERPRISE === process.env.APP_MODE`. This new write is unconditional, so in community mode it writes a value that is never read and never maintained by the mark-as-read path — dead/inconsistent state. More importantly, in enterprise mode this write runs on every refetch of `useGetUnreadNotificationsCount` (React Query default refetchOnWindowFocus/onMount), which can clobber a real-time FCM increment with a momentarily-stale server count. Recommend gating this write to enterprise mode (mirroring the mark-as-read logic) so localStorage stays consistent with the code that actually consumes it.

frontend/src/community/common/components/organisms/AppBar/AppBar.tsx
```diff
@@ -76,9 +76,13 @@ const AppBar = () => {
   const { data: unreadCount } = useGetUnreadNotificationsCount();
```
ThinuwanW, line 76: 🤖 **Claude** · 💬 **Note**

Type safety: `useGetUnreadNotificationsCount` returns `data.results[0]` with no type annotation, so `unreadCount` is inferred as `any`. The new code calls `unreadCount.toString()` and assigns it straight into `notifyData.unreadCount` (typed `number`), both of which silently assume the API yields a number — if it ever returns an object, `.toString()` produces "[object Object]" and the badge breaks with no compile-time signal. Consider giving the query hook an explicit return type (e.g. `useQuery<number>` / typing `data.results[0]`).

frontend/src/community/common/components/organisms/AppBar/AppBar.tsx
ThinuwanW, line 85: 🤖 **Claude** · 🔴 **Important**

Null-safety regression: the new guard only bails on `unreadCount === undefined`, but the previous code (`unreadCount || 0`) also handled `null`. `useGetUnreadNotificationsCount` returns `data.results[0]` (typed `any`), so if the API ever returns `null` for that element, `unreadCount === undefined` is false and `unreadCount.toString()` on line 85 throws an unhandled TypeError, breaking the AppBar render. It also writes a non-number into `notifyData.unreadCount`, which is typed `number` in notificationsSlice. Fix: guard with `if (unreadCount == null) return;` (covers both null and undefined) or use `String(unreadCount ?? 0)` and `setNotifyData({ unreadCount: unreadCount ?? 0 })`.

ThinuwanW, line 85: 🤖 **Claude** · 🟣 **Nit**

Mode-gating inconsistency: this `localStorage.setItem("unReadMsgCount", ...)` runs unconditionally in both community and enterprise modes, but the value is only ever read/decremented under an enterprise guard (`appModes.ENTERPRISE === process.env.APP_MODE` in notificationsApi.ts:98, and useFCMToken.ts which is enterprise-only). In community mode this is a dead write. Consider gating the write with the same enterprise mode check so the seeding stays consistent with where the key is actually consumed.

frontend/src/community/common/components/organisms/AppBar/AppBar.tsx
shakila-kamalasena: Existing code

frontend/src/community/common/components/organisms/AppBar/AppBar.tsx
shakila-kamalasena: No need enterprise gates

shakila-kamalasena: no need

shakila-kamalasena: Fixed

hasalarootcode (approved):

akilarootcode (approved):

ThinuwanW (approved):

PR #1660, opened 2026-07-21
feat(crm): integrate limit guard for adding deals in Kanban board

dulsara-manakal:
### Summary

- The upgrade limitation modal should be displayed when attempting to add a deal from the Swimlane View after reaching the free plan limit.

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/frontend/src/community/crm/components/organisms/DealsKanbanBoard/DealsKanbanBoard.tsx
+++ b/frontend/src/community/crm/components/organisms/DealsKanbanBoard/DealsKanbanBoard.tsx
@@ -49,7 +49,7 @@ const DealsKanbanBoard: FC<DealsKanbanBoardProps> = ({
   const { guardCrmCreate, isCheckingCrmLimit } = useCrmLimitGuard();
 
   const handleAddDeal = (stageId: number) => {
-    void guardCrmCreate(CrmLimitResource.DEALS, () => {
+    guardCrmCreate(CrmLimitResource.DEALS, () => {
       const { setPreselectedStageId, openCrmSidePanel } =
         useCrmStore.getState();
       setPreselectedStageId(stageId);
--- a/frontend/src/community/crm/components/organisms/DealsKanbanBoard/DealsKanbanBoard.tsx
+++ b/frontend/src/community/crm/components/organisms/DealsKanbanBoard/DealsKanbanBoard.tsx
@@ -16,6 +16,8 @@ import { useBoardData } from "~community/crm/hooks/useBoardData";
 import { useKanbanDrag } from "~community/crm/hooks/useKanbanDrag";
 import { useCrmStore } from "~community/crm/store/store";
 import { CrmSidePanelTypes } from "~community/crm/types/SidePanelTypes";
+import useCrmLimitGuard from "~enterprise/crm/hooks/useCrmLimitGuard";
+import { CrmLimitResource } from "~enterprise/crm/types/CrmLimitTypes";
 
 interface DealsKanbanBoardProps {
   searchKeyword?: string;
@@ -44,10 +46,15 @@ const DealsKanbanBoard: FC<DealsKanbanBoardProps> = ({
     handleDragEnd
   } = useKanbanDrag();
 
+  const { guardCrmCreate } = useCrmLimitGuard();
+
   const handleAddDeal = (stageId: number) => {
-    const { setPreselectedStageId, openCrmSidePanel } = useCrmStore.getState();
-    setPreselectedStageId(stageId);
-    openCrmSidePanel(CrmSidePanelTypes.ADD_DEAL_SIDE_PANEL);
+    guardCrmCreate(CrmLimitResource.DEALS, () => {
+      const { setPreselectedStageId, openCrmSidePanel } =
+        useCrmStore.getState();
+      setPreselectedStageId(stageId);
+      openCrmSidePanel(CrmSidePanelTypes.ADD_DEAL_SIDE_PANEL);
+    });
   };
 
   const handleDealClick = (dealId: number) => {
--- a/frontend/src/community/crm/components/organisms/DealsKanbanBoard/DealsKanbanBoard.tsx
+++ b/frontend/src/community/crm/components/organisms/DealsKanbanBoard/DealsKanbanBoard.tsx
@@ -46,7 +46,7 @@ const DealsKanbanBoard: FC<DealsKanbanBoardProps> = ({
     handleDragEnd
   } = useKanbanDrag();
 
-  const { guardCrmCreate } = useCrmLimitGuard();
+  const { guardCrmCreate, isCheckingCrmLimit } = useCrmLimitGuard();
 
   const handleAddDeal = (stageId: number) => {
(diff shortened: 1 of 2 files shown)


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR correctly wraps the Kanban board's per-stage 'Add Deal' action in the CRM limit guard, matching the existing pattern used by the deals page and side panels. The one gap is that it ignores the hook's isCheckingCrmLimit loading flag, which every other caller uses to disable the add button while the limitation query is pending — during that window the guard allows creation and the limit can be bypassed.

Found **3** new issue(s): 🔴 1 important, 🟡 1 suggestion(s), 🟣 1 nit(s)

frontend/src/community/crm/components/organisms/DealsKanbanBoard/DealsKanbanBoard.tsx
```diff
@@ -44,10 +46,15 @@ const DealsKanbanBoard: FC<DealsKanbanBoardProps> = ({
     handleDragEnd
   } = useKanbanDrag();
 
+  const { guardCrmCreate } = useCrmLimitGuard();
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The limit check is only half-integrated: `isCheckingCrmLimit` returned by `useCrmLimitGuard()` is not consumed. In enterprise mode, while `useGetCrmLimitation()` is still pending, `limitationData` is `undefined`, so `guardCrmCreate` treats the resource as not limited and calls `onAllowed()` — opening the Add Deal panel even when the deal limit has been reached. Every other caller (deals.tsx L19/L41, SidePanelDealSection L47/L89-90/L137-138, SidePanelAddDeal L110/L113, contacts/companies/tasks pages) destructures `isCheckingCrmLimit` and uses it to disable/show a loading state on the add button, closing this bypass window. Fix: destructure `isCheckingCrmLimit` here, add a `disabled`/`isLoading` prop to `DealStageLane`'s add-deal `ButtonV2`, and thread the flag through `onAddDeal` so the per-stage add button is gated while the limit is being checked.

frontend/src/community/crm/components/organisms/DealsKanbanBoard/DealsKanbanBoard.tsx
```diff
@@ -44,10 +46,15 @@ const DealsKanbanBoard: FC<DealsKanbanBoardProps> = ({
...
+
   const handleAddDeal = (stageId: number) => {
-    const { setPreselectedStageId, openCrmSidePanel } = useCrmStore.getState();
-    setPreselectedStageId(stageId);
-    openCrmSidePanel(CrmSidePanelTypes.ADD_DEAL_SIDE_PANEL);
+    guardCrmCreate(CrmLimitResource.DEALS, () => {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Fail-open beyond the initial-load window: the enterprise `guardCrmCreate` computes `isLimited = limitationData?.[resource]?.limitReached`, so whenever `limitationData` is undefined the guard runs `onAllowed()` and opens the Add Deal panel. This is not limited to the initial pending state (that transient case is the first-pass finding) — if `useGetCrmLimitation` errors permanently, `limitationData` stays undefined and every per-lane add click silently bypasses the enterprise limit with no toast/error shown to the user. The new Kanban usage inherits this. Consider having the guard treat an errored/undefined limitation response as 'block + show feedback' rather than 'allow', or surface the query error here so a user isn't silently allowed past a limit that couldn't be verified.

ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`guardCrmCreate` is declared `async` in the enterprise implementation and returns a Promise, but it is invoked here fire-and-forget with no `await` and no `.catch()`, creating a floating promise. If a future implementation of the guard rejects (e.g., it awaits a network call), this becomes an unhandled rejection with no user feedback and may trip `@typescript-eslint/no-floating-promises`. Prefer `void guardCrmCreate(...)` to make the intent explicit, or attach a `.catch()` that surfaces a toast. (Note: the existing call sites share this pattern, so a broader cleanup could be warranted.)

frontend/src/community/crm/components/organisms/DealsKanbanBoard/DealsKanbanBoard.tsx
dulsara-manakal: Fixed this, add that `isCheckingCrmLimit ` too.

frontend/src/community/crm/components/organisms/DealsKanbanBoard/DealsKanbanBoard.tsx
dulsara-manakal: BE always checks the limitation even though an existing hook fails with network failure. Otherwise, be guaranteed to send the correct response.

dulsara-manakal: Use the existing pattern everywhere, and that return promise always resolves immediately and can never reject.

hasalarootcode: Whenever undefined, the default should be false

hasalarootcode: this is also valid

hasalarootcode (changes requested):

frontend/src/community/crm/components/organisms/DealsKanbanBoard/DealsKanbanBoard.tsx
dulsara-manakal: Confirmed. In the `useCrmLimitGuard` hook, whenever it is undefined or BE call is failed, it is always false.

dulsara-manakal: Add the `void` operator as mentioned in the comment. For the record though, `@typescript-eslint/no-floating-promises` won't actually flag this in our setup: it's not part of any preset in our `.eslintrc.json`

dulsara-manakal: Had to remove `void` because of a sonar issue. 
image

dulsara-manakal: removed `void` that's unnecessary. fix this from hook in this [PR](https://github.com/rootcodelabs/skapp-ep-fe-src/pull/762)

hasalarootcode (approved):

akilarootcode (approved):

ThinuwanW (approved):

PR #1661, opened 2026-07-22
Fix/2334 geofence validation alert

shakila-kamalasena:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/frontend/src/community/attendance/hooks/useRecordAttendance.ts
+++ b/frontend/src/community/attendance/hooks/useRecordAttendance.ts
@@ -6,7 +6,10 @@ import { AttendanceSlotType } from "~community/attendance/types/attendanceTypes"
 import { getCurrentLocation } from "~community/attendance/utils/geolocationUtils";
 import { appModes } from "~community/common/constants/configs";
 import { convertDateToUTC } from "~community/common/utils/dateTimeUtils";
-import { useUpdateEmployeeStatusWithLocation } from "~enterprise/attendance/api/AttendanceApi";
+import {
+  useGetUserGeofenceStatus,
+  useUpdateEmployeeStatusWithLocation
+} from "~enterprise/attendance/api/AttendanceApi";
 import { useGetEnvironment } from "~enterprise/common/hooks/useGetEnvironment";
 
 export const useRecordAttendance = (
@@ -23,11 +26,16 @@ export const useRecordAttendance = (
   const isEnterprise = environment === appModes.ENTERPRISE;
 
   const { data: attendanceConfig } = useGetAttendanceConfiguration();
-  const isGeoFencingEnabled = attendanceConfig?.isGeoFencingEnabled;
+  const isGeoFencingEnabled: boolean = attendanceConfig?.isGeoFencingEnabled;
+
+  const { data: geofenceStatus } = useGetUserGeofenceStatus(
+    isGeoFencingEnabled && isEnterprise
+  );
+  const isGeofenceConfigured = geofenceStatus?.geofenceConfigured;
 
   const recordAttendance = useCallback(
     (slotType: AttendanceSlotType) => {
-      if (isGeoFencingEnabled && isEnterprise) {
+      if (isGeoFencingEnabled && isEnterprise && isGeofenceConfigured) {
         getCurrentLocation().then(({ latitude, longitude }) => {
           mutateWithLocation({
             recordActionType: slotType,
@@ -40,7 +48,13 @@ export const useRecordAttendance = (
         mutate(slotType);
       }
     },
-    [isGeoFencingEnabled, isEnterprise, mutateWithLocation, mutate]
+    [
+      isGeoFencingEnabled,
+      isEnterprise,
+      isGeofenceConfigured,
+      mutateWithLocation,
+      mutate
+    ]
   );
 
   return {
--- a/frontend/src/community/attendance/hooks/useRecordAttendance.ts
+++ b/frontend/src/community/attendance/hooks/useRecordAttendance.ts
@@ -28,37 +28,47 @@ export const useRecordAttendance = (
   const { data: attendanceConfig } = useGetAttendanceConfiguration();
   const isGeoFencingEnabled: boolean = attendanceConfig?.isGeoFencingEnabled;
 
-  const { data: geofenceStatus } = useGetUserGeofenceStatus(
-    isGeoFencingEnabled && isEnterprise
-  );
+  const {
+    data: geofenceStatus,
+    isLoading: isGeofenceStatusLoading,
+    isError: isGeofenceStatusError
+  } = useGetUserGeofenceStatus(isGeoFencingEnabled && isEnterprise);
   const isGeofenceConfigured = geofenceStatus?.geofenceConfigured;
 
   const recordAttendance = useCallback(
     (slotType: AttendanceSlotType) => {
-      if (isGeoFencingEnabled && isEnterprise && isGeofenceConfigured) {
-        getCurrentLocation().then(({ latitude, longitude }) => {
-          mutateWithLocation({
-            recordActionType: slotType,
-            time: convertDateToUTC(new Date().toISOString()) as string,
-            latitude,
-            longitude
+      if (isGeoFencingEnabled && isEnterprise) {
+        if (isGeofenceStatusError) {
+          onError?.();
+          return;
+        }
+        if (isGeofenceConfigured) {
+          getCurrentLocation().then(({ latitude, longitude }) => {
+            mutateWithLocation({
+              recordActionType: slotType,
+              time: convertDateToUTC(new Date().toISOString()) as string,
+              latitude,
+              longitude
+            });
           });
-        });
-      } else {
-        mutate(slotType);
+          return;
+        }
       }
+      mutate(slotType);
     },
     [
       isGeoFencingEnabled,
       isEnterprise,
+      isGeofenceStatusError,
       isGeofenceConfigured,
       mutateWithLocation,
-      mutate
+      mutate,
+      onError
     ]
   );
 
   return {
     recordAttendance,
-    isPending: isPending || isEpPending
+    isPending: isPending || isEpPending || isGeofenceStatusLoading
   };
 };
--- a/frontend/src/community/attendance/hooks/useRecordAttendance.ts
+++ b/frontend/src/community/attendance/hooks/useRecordAttendance.ts
@@ -33,7 +33,7 @@ export const useRecordAttendance = (
     isLoading: isGeofenceStatusLoading,
     isError: isGeofenceStatusError
   } = useGetUserGeofenceStatus(isGeoFencingEnabled && isEnterprise);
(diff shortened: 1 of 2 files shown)


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds a geofence-configured check before using location-based attendance recording, gating it behind a new `useGetUserGeofenceStatus` query. The logic and query wiring (enabled flag, optional chaining, useCallback deps) are correct, but the async geofence-status query's loading/error state is not surfaced to the UI, creating a window where attendance can be recorded without geofence validation. There is also a misleading `boolean` type annotation on a value that is `undefined` at runtime.

Found **6** new issue(s): 🔴 3 important, 🟡 2 suggestion(s), 🟣 1 nit(s)

frontend/src/community/attendance/hooks/useRecordAttendance.ts
```diff
@@ -23,11 +26,16 @@ export const useRecordAttendance = (
   const isEnterprise = environment === appModes.ENTERPRISE;
 
   const { data: attendanceConfig } = useGetAttendanceConfiguration();
-  const isGeoFencingEnabled = attendanceConfig?.isGeoFencingEnabled;
+  const isGeoFencingEnabled: boolean = attendanceConfig?.isGeoFencingEnabled;
+
+  const { data: geofenceStatus } = useGetUserGeofenceStatus(
+    isGeoFencingEnabled && isEnterprise
+  );
+  const isGeofenceConfigured = geofenceStatus?.geofenceConfigured;
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Potential silent geofence bypass. `isGeofenceConfigured` comes from the async `useGetUserGeofenceStatus` query, but its loading/error state is not reflected in the returned `isPending` (line 62 only ORs the two mutation pending flags). Consequently: (1) if the user clicks clock-in before the query resolves, `isGeofenceConfigured` is `undefined`, so the code falls into the `else` branch and calls `mutate(slotType)` — recording attendance WITHOUT location even though geofencing is enabled; the button in ClockWidget.tsx is not disabled during this window. (2) If the query errors, `geofenceStatus` stays undefined and geofencing is silently skipped. Since geofencing is a validation control, gate on the query being resolved: include `isLoading`/`isFetching` from `useGetUserGeofenceStatus` in the returned `isPending` (so the button is disabled until status is known), and/or handle the query error explicitly rather than defaulting to the no-location path.

frontend/src/community/attendance/hooks/useRecordAttendance.ts
```diff
@@ -23,11 +26,16 @@ export const useRecordAttendance = (
   const isEnterprise = environment === appModes.ENTERPRISE;
 
   const { data: attendanceConfig } = useGetAttendanceConfiguration();
-  const isGeoFencingEnabled = attendanceConfig?.isGeoFencingEnabled;
+  const isGeoFencingEnabled: boolean = attendanceConfig?.isGeoFencingEnabled;
```
ThinuwanW, line 29: 🤖 **Claude** · 🟡 **Suggestion**

Misleading type annotation. `attendanceConfig` is untyped (`useGetAttendanceConfiguration` returns `result?.data?.results[0]` as `any`) and `AttendanceConfigurationType.isGeoFencingEnabled` is declared `isGeoFencingEnabled?: boolean`. At runtime this value is `boolean | undefined` (undefined before the config loads), yet it is annotated `: boolean`. TypeScript only accepts this because the source is `any`, so the annotation asserts a type the value doesn't actually have. Prefer `const isGeoFencingEnabled = attendanceConfig?.isGeoFencingEnabled ?? false;` to make it genuinely boolean, rather than annotating around an `any`.

frontend/src/community/attendance/hooks/useRecordAttendance.ts
```diff
@@ -23,11 +26,16 @@ export const useRecordAttendance = (
...
+  const { data: geofenceStatus } = useGetUserGeofenceStatus(
+    isGeoFencingEnabled && isEnterprise
+  );
+  const isGeofenceConfigured = geofenceStatus?.geofenceConfigured;
 
   const recordAttendance = useCallback(
     (slotType: AttendanceSlotType) => {
-      if (isGeoFencingEnabled && isEnterprise) {
+      if (isGeoFencingEnabled && isEnterprise && isGeofenceConfigured) {
         getCurrentLocation().then(({ latitude, longitude }) => {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`getCurrentLocation().then(...)` has no `.catch()`. If the browser geolocation call rejects (permission denied, timeout, insecure context), the rejection is unhandled, `mutateWithLocation` never fires, and the user gets no feedback (the `onError` passed to the hook only covers mutation failures, not geolocation failures). This path is directly exercised by this PR's condition; add a `.catch()` that surfaces an error toast (or falls back) so a location failure isn't a silent no-op.

frontend/src/community/attendance/hooks/useRecordAttendance.ts
```diff
@@ -6,7 +6,10 @@ import { AttendanceSlotType } from "~community/attendance/types/attendanceTypes"
 import { getCurrentLocation } from "~community/attendance/utils/geolocationUtils";
 import { appModes } from "~community/common/constants/configs";
 import { convertDateToUTC } from "~community/common/utils/dateTimeUtils";
-import { useUpdateEmployeeStatusWithLocation } from "~enterprise/attendance/api/AttendanceApi";
+import {
+  useGetUserGeofenceStatus,
+  useUpdateEmployeeStatusWithLocation
```
ThinuwanW, line 11: 🤖 **Claude** · 🔴 **Important**

Missing paired change in the fallback stub. `~enterprise/*` resolves to `./src/enterprise/*` and then `./src/fallback/*` (tsconfig.json:22), so in a community-only checkout (no enterprise submodule) this import resolves to `src/fallback/attendance/api/AttendanceApi.ts`, which only exports `useUpdateEmployeeStatusWithLocation` — it does NOT export `useGetUserGeofenceStatus`. That makes the community build fail with 'Module has no exported member useGetUserGeofenceStatus'. Fix: add a matching `useGetUserGeofenceStatus` (and the `UserGeofenceStatus` type) stub to `src/fallback/attendance/api/AttendanceApi.ts`, mirroring the existing pattern (e.g. return `{ data: { geofenceConfigured: false } }`).

frontend/src/community/attendance/hooks/useRecordAttendance.ts
```diff
@@ -23,11 +26,16 @@ export const useRecordAttendance = (
   const isEnterprise = environment === appModes.ENTERPRISE;
 
   const { data: attendanceConfig } = useGetAttendanceConfiguration();
-  const isGeoFencingEnabled = attendanceConfig?.isGeoFencingEnabled;
+  const isGeoFencingEnabled: boolean = attendanceConfig?.isGeoFencingEnabled;
+
+  const { data: geofenceStatus } = useGetUserGeofenceStatus(
+    isGeoFencingEnabled && isEnterprise
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The `enabled` guard is defeated when config hasn't loaded. `isGeoFencingEnabled` comes from `attendanceConfig?.isGeoFencingEnabled`, which is `undefined` until the config query resolves. `undefined && isEnterprise` evaluates to `undefined`, and passing `undefined` to `useGetUserGeofenceStatus(enabled: boolean = true)` triggers the default parameter value `true` — so the geofence-status request fires on initial render (and in any state where the flag is undefined), including before config loads, regardless of mode. Fix: coerce to a real boolean, e.g. `useGetUserGeofenceStatus(Boolean(isGeoFencingEnabled) && isEnterprise)`, so a falsy/undefined config yields `false` rather than falling through to the `true` default.

frontend/src/community/attendance/hooks/useRecordAttendance.ts
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`isGeofenceConfigured` has no explicit type; it is inferred as `boolean | undefined` (from `geofenceStatus?.geofenceConfigured`). Strict-mode conventions here call for explicit types. Consider `const isGeofenceConfigured: boolean = geofenceStatus?.geofenceConfigured ?? false;` — this also makes the undefined-during-loading behavior explicit at the source.

frontend/src/community/attendance/hooks/useRecordAttendance.ts
shakila-kamalasena: Added

frontend/src/community/attendance/hooks/useRecordAttendance.ts
shakila-kamalasena: No need to add this type of fallbacks

frontend/src/community/attendance/hooks/useRecordAttendance.ts
shakila-kamalasena: no need boolean cast here

frontend/src/community/attendance/hooks/useRecordAttendance.ts
shakila-kamalasena: Done

shakila-kamalasena: No need as it already has a type

frontend/src/community/attendance/hooks/useRecordAttendance.ts
shakila-kamalasena: no need

hasalarootcode (dismissed):

frontend/src/community/attendance/hooks/useRecordAttendance.ts
```diff
@@ -23,28 +26,49 @@ export const useRecordAttendance = (
...
   const { data: attendanceConfig } = useGetAttendanceConfiguration();
-  const isGeoFencingEnabled = attendanceConfig?.isGeoFencingEnabled;
+  const isGeoFencingEnabled: boolean = attendanceConfig?.isGeoFencingEnabled;
+
+  const {
+    data: geofenceStatus,
+    isLoading: isGeofenceStatusLoading,
+    isError: isGeofenceStatusError
+  } = useGetUserGeofenceStatus(isGeoFencingEnabled && isEnterprise);
+  const isGeofenceConfigured = geofenceStatus?.geofenceConfigured;
```
akilarootcode: isGeofenceConfigured

akilarootcode (changes requested):

frontend/src/community/attendance/hooks/useRecordAttendance.ts
shakila-kamalasena: Done

akilarootcode (approved):

ThinuwanW (approved):

PR #1666, opened 2026-07-23
Fix/2632 sso profile picture persistence

shakila-kamalasena:
…URLs

### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/backend/src/main/java/com/skapp/community/common/util/CommonModuleUtils.java
+++ b/backend/src/main/java/com/skapp/community/common/util/CommonModuleUtils.java
@@ -283,10 +283,6 @@ public class CommonModuleUtils {
 		}
 	}
 
-	public static boolean isExternalAuthPicUrl(String authPic) {
-		return authPic != null && (authPic.startsWith("https://"));
-	}
-
 	/**
 	 * Safely extracts and sets a value from a nested structure only if not null. Combines
 	 * safeGet and setIfNotNull for cleaner syntax.
@@ -342,6 +338,14 @@ public class CommonModuleUtils {
 	public static boolean isGuestEmployee(Employee employee) {
 		return employee != null && employee.getEmployeeRole() != null
 				&& Role.PM_GUEST_EMPLOYEE == employee.getEmployeeRole().getPmRole();
+	}
+
+	public static boolean isExternalAuthPicUrl(String authPic) {
+		return authPic != null && authPic.startsWith("https://");
+	}
+
+	public static boolean shouldReplaceAuthPic(String existing, String incoming) {
+		return incoming != null && !incoming.equals(existing) && (existing == null || isExternalAuthPicUrl(existing));
 	}
 
 }
--- a/backend/src/main/java/com/skapp/community/common/util/CommonModuleUtils.java
+++ b/backend/src/main/java/com/skapp/community/common/util/CommonModuleUtils.java
@@ -283,6 +283,10 @@ public class CommonModuleUtils {
 		}
 	}
 
+	public static boolean isExternalAuthPicUrl(String authPic) {
+		return authPic != null && (authPic.startsWith("https://"));
+	}
+
 	/**
 	 * Safely extracts and sets a value from a nested structure only if not null. Combines
 	 * safeGet and setIfNotNull for cleaner syntax.


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds a small utility method isExternalAuthPicUrl to CommonModuleUtils that distinguishes external provider auth-picture URLs (returned as-is) from app-managed S3 paths (which get a signed URL generated). The logic is correct for the Google/SSO callers and includes a null guard. The only issues are stylistic/edge-case: it lacks the Javadoc that every other public method in this file has, and its scheme check is case-sensitive and https-only.

Found **4** new issue(s): 🟡 1 suggestion(s), 🟣 3 nit(s)

backend/src/main/java/com/skapp/community/common/util/CommonModuleUtils.java
```diff
@@ -283,6 +283,10 @@ public static <T> void setIfNotNull(T value, Consumer<T> setter) {
 		}
 	}
 
+	public static boolean isExternalAuthPicUrl(String authPic) {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Missing Javadoc. Every other public method in CommonModuleUtils (safeGet, setIfNotNull, setIfExists, isEmpty) is documented with a Javadoc block explaining intent, params, and return value. This new public method has none, and its name is ambiguous (what counts as "external"? why only https?). Add a Javadoc clarifying that it identifies auth pictures that are already fully-qualified external URLs (e.g. OAuth provider avatars) versus app-managed S3 object paths that must be signed, so callers understand the branch semantics.

backend/src/main/java/com/skapp/community/common/util/CommonModuleUtils.java
```diff
@@ -283,6 +283,10 @@ public static <T> void setIfNotNull(T value, Consumer<T> setter) {
...
+		return authPic != null && (authPic.startsWith("https://"));
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

The scheme check is case-sensitive and matches only "https://". Any stored authPic using a different scheme/case (e.g. "http://", "HTTPS://") is classified as NOT external, so the caller in EpUserServiceImpl.mapEmployeeToAuthPicDto (line 165) falls through to generateAuthPicUrl, which prepends S3_PROFILE_PIC_THUMBNAIL_PATH to the full URL and generates a bogus signed URL — a broken image. This is safe today because Google/OAuth avatars are always lowercase https, but the assumption is undocumented and brittle. Consider a case-insensitive check that also accepts http:// (e.g. authPic.regionMatches(true, 0, "http", 0, 4)) or at minimum document the https-only assumption. Low likelihood, so flagged as a nit.

ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Redundant parentheses around `(authPic.startsWith("https://"))`. The `startsWith` call is a single boolean expression that does not need grouping. Simplify to `return authPic != null && authPic.startsWith("https://");` to match the style of the other predicate methods in this class (e.g., isEmpty, isGuestEmployee).

backend/src/main/java/com/skapp/community/common/util/CommonModuleUtils.java
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Method placement breaks a logically-grouped cluster. `isExternalAuthPicUrl` is inserted between `setIfNotNull` (line 280) and `setIfExists` (line 297), which together with `safeGet` form a documented, related group of safe get/set helpers. This URL-validation predicate is unrelated to those helpers and interrupts the cluster. Consider moving it next to other standalone predicates (e.g., near `isGuestEmployee` or `isValidFloat`) for better cohesion.

shakila-kamalasena: No need javadocs

backend/src/main/java/com/skapp/community/common/util/CommonModuleUtils.java
shakila-kamalasena: no need

backend/src/main/java/com/skapp/community/common/util/CommonModuleUtils.java
shakila-kamalasena: Fixed

backend/src/main/java/com/skapp/community/common/util/CommonModuleUtils.java
shakila-kamalasena: Fixed

PR #1673, opened 2026-07-24
Feat/contact name split contact section Backend

YasinduRC:
### Summary

Replaces the single name field on CRM contacts with separate firstName (required) and lastName (optional) fields, end‑to‑end across the persistence, validation, API, and query layers. This aligns CRM contacts with how people/owners are already modeled and enables first/last‑aware search and sorting.

Scope: Backend only. Frontend changes will be addressed in a separate PR.

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR cleanly splits CrmContact.name into firstName/lastName across the entity, DTOs, mapper, validations, service, and repository search/order clauses, with thorough test coverage. The Java-side refactor is consistent and complete. However, it is missing the Liquibase schema migration for the column change, which will break every contact read/write on any Liquibase-managed (production) database.

Found **4** new issue(s): 🔴 1 important, 🟡 3 suggestion(s)

backend/src/main/java/com/skapp/community/crmplanner/model/CrmContact.java
```diff
@@ -29,8 +29,11 @@ public class CrmContact extends Auditable<String> {
 	@Column(name = "id", nullable = false, updatable = false)
 	private Long id;
 
-	@Column(name = "name", nullable = false)
-	private String name;
+	@Column(name = "first_name", nullable = false)
```
ThinuwanW, line 32: 🤖 **Claude** · 🔴 **Important**

This changes the JPA column mapping from `name` to `first_name` (NOT NULL) + `last_name`, but no Liquibase changeset was added under `backend/src/main/resources/community/db/changelog/`. The existing DDL (common-ddl-script-v1-create-tables.sql:624) still defines `crm_contact.name text NOT NULL` and has no `first_name`/`last_name` columns. On any Liquibase-managed DB (production), Hibernate will fail on every insert/select: it stops writing the still-NOT-NULL `name` column and references non-existent `first_name`/`last_name` columns. This is NOT caught by the test suite because tests use `ddl-auto: create` with `liquibase.enabled: false` (src/test/resources/application.yml), so the schema is generated from the entity. Add an incremental changeset (following the existing `common-ddl-script-v1-alter-table-*` pattern registered in db.changelog.yml) that: (1) adds `first_name` and `last_name` columns, (2) backfills them from existing `name` data, (3) sets `first_name` NOT NULL after backfill, and (4) drops the old `name` column (or makes it nullable if kept for rollback safety).

backend/src/main/java/com/skapp/community/crmplanner/model/CrmContact.java
```diff
@@ -52,4 +55,8 @@ public class CrmContact extends Auditable<String> {
 	@Column(name = "is_deleted", nullable = false)
 	private Boolean isDeleted = false;
 
+	public String getFullName() {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`getFullName()` uses `lastName != null` to decide whether to append it, but `lastName` is optional and can be persisted as an empty/blank string: `validateContactLastName` returns early (no error) for blank values, and `createContact`/`editContact` set the raw value, while global JSON string-trimming turns "  " into "". When `lastName` is an empty string, `getFullName()` returns `firstName + " "` with a trailing space, which then surfaces in `CrmDealResponseDto.contactName`. Prefer `StringUtils.hasText(lastName)` (or check `!lastName.isBlank()`) instead of `!= null` so empty last names collapse to just the first name.

backend/src/main/java/com/skapp/community/crmplanner/payload/response/CrmContactResponseDto.java
```diff
@@ -11,7 +11,9 @@ public class CrmContactResponseDto {
 
 	private Long id;
 
-	private String name;
+	private String firstName;
+
```
ThinuwanW, line 15: 🤖 **Claude** · 🟡 **Suggestion**

Breaking API response-contract change. All contact response DTOs (`CrmContactResponseDto`, `CrmContactDetailResponseDto`, `CrmContactListItemDto`, `CrmContactLookupResponseDto`, `CrmBoardContactResponseDto`) drop the `name` field in favor of `firstName`/`lastName` without a version bump. Any existing API consumer (frontend, integrations) reading `results[].name` will silently break. Confirm the frontend and any external consumers are updated in lockstep, or expose a computed `fullName`/`name` field alongside the split fields for backward compatibility if consumers still depend on it.

backend/src/main/java/com/skapp/community/crmplanner/payload/request/CrmContactCreateRequestDto.java
```diff
@@ -7,7 +7,9 @@
 @Setter
 public class CrmContactCreateRequestDto {
 
-	private String name;
+	private String firstName;
```
ThinuwanW, line 10: 🤖 **Claude** · 🟡 **Suggestion**

DTO validation annotations remain absent after the rename. `firstName` is a required, non-empty field (enforced only later in the service via `CrmValidations.validateContactFirstName`), yet the request DTO carries no `@NotBlank`/size annotations and the controller likely has no `@Valid` trigger. This matches the pre-existing pattern (the old `name` had none either), so it is not a regression, but the rename was a natural opportunity to add `@NotBlank`/`@Size(max = CONTACT_NAME_MAX_LENGTH)` on `firstName` and `@Size` on `lastName` so bean validation and the service checks stay consistent. Consider adding them, or leave as-is to preserve the current service-layer-only validation convention.

backend/src/main/java/com/skapp/community/crmplanner/model/CrmContact.java
YasinduRC: Migrations are created in a separate PR

backend/src/main/java/com/skapp/community/crmplanner/model/CrmContact.java
YasinduRC: used StringUtils.hasText(lastName and fixed

backend/src/main/java/com/skapp/community/crmplanner/payload/request/CrmContactCreateRequestDto.java
YasinduRC: As a practice we do not write validations on DTOs

backend/src/main/java/com/skapp/community/crmplanner/payload/response/CrmContactResponseDto.java
YasinduRC: Frontend is updated also they will deployed together

backend/src/main/java/com/skapp/community/crmplanner/model/CrmContact.java
```diff
@@ -52,4 +56,8 @@ public class CrmContact extends Auditable<String> {
...
```
anusham-rc: check this

anusham-rc (changes requested):

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmContactServiceImpl.java
```diff
@@ -116,7 +117,8 @@ public ResponseEntityDto createContact(CrmContactCreateRequestDto requestDto) {
 		Employee owner = crmOwnerResolver.resolveOwner(requestDto.getOwnerId(), currentUser);
 
 		CrmContact contact = new CrmContact();
-		contact.setName(requestDto.getName());
+		contact.setFirstName(requestDto.getFirstName());
```
erandi-rootcode, line 120: Check the behavior and apply fix if needed. 

trim the requestDto.getFirstName() and requestDto.getLastName() before saving. 
else if " Jane " like value is sent, even though it's validated it will save as the same with the leading and trailing spaces.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmContactServiceImpl.java
```diff
@@ -146,9 +148,15 @@ public ResponseEntityDto editContact(Long id, CrmContactEditRequestDto requestDt
 			throw new ModuleException(CrmMessageConstant.CRM_ERROR_CONTACT_EDIT_DENIED);
 		}
 
-		if (requestDto.getName() != null) {
-			CrmValidations.validateContactName(requestDto.getName());
-			contact.setName(requestDto.getName());
+		if (requestDto.getFirstName() != null) {
+			CrmValidations.validateContactFirstName(requestDto.getFirstName());
+			contact.setFirstName(requestDto.getFirstName());
```
erandi-rootcode, line 153: Check the behavior and apply fix if needed. 
trim here as well

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmContactServiceImpl.java
```diff
@@ -146,9 +148,15 @@ public ResponseEntityDto editContact(Long id, CrmContactEditRequestDto requestDt
...
+		}
+
+		if (requestDto.getLastName() != null) {
+			String lastName = requestDto.getLastName();
+			CrmValidations.validateContactLastName(lastName);
+			contact.setLastName(StringUtils.hasText(lastName) ? lastName : null);
```
erandi-rootcode: Check the behavior and apply fix if needed. 

trim here too 
 StringUtils.hasText(lastName) ? lastName.trim() : null

erandi-rootcode (changes requested):

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmContactServiceImpl.java
YasinduRC: Checked -  no fix needed. The app has a global Jackson deserializer that trims every incoming JSON string field before it ever reaches the DTO getters

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmContactServiceImpl.java
YasinduRC: Checked -  no fix needed. The app has a global Jackson deserializer that trims every incoming JSON string field before it ever reaches the DTO getters

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmContactServiceImpl.java
YasinduRC: Checked -  no fix needed. The app has a global Jackson deserializer that trims every incoming JSON string field before it ever reaches the DTO getters

erandi-rootcode (dismissed):

backend/src/main/java/com/skapp/community/crmplanner/model/CrmContact.java
YasinduRC: will discuss with the team

backend/src/main/java/com/skapp/community/crmplanner/util/CrmValidations.java
```diff
@@ -45,17 +45,31 @@ public static void validateCompanyId(Long companyId) {
...
+	}
+
+	public static void validateContactLastName(String lastName) {
+		if (lastName == null || lastName.isBlank()) {
+			return;
 		}
 
-		if (name.trim().length() > CrmConstants.CONTACT_NAME_MAX_LENGTH) {
-			throw new ModuleException(CrmMessageConstant.CRM_ERROR_CONTACT_NAME_TOO_LONG);
+		if (lastName.trim().length() > CrmConstants.CONTACT_NAME_MAX_LENGTH) {
```
akilarootcode: trim() remove trim()

akilarootcode (changes requested):

backend/src/main/java/com/skapp/community/crmplanner/util/CrmValidations.java
YasinduRC: removed the trim

backend/src/main/java/com/skapp/community/crmplanner/model/CrmContact.java
YasinduRC: Removed the firstFullNam() from the entity, used firstName and lastName separately

erandi-rootcode (dismissed):

akilarootcode (dismissed):

PR #1677, opened 2026-07-27
feat: read module role restrictions from module_roles_restriction

anusham-rc:
@
## What

Phase 1 of the module roles restriction alignment: `GET /v1/roles/restrictions/{module}` now reads `module_roles_restriction` (the new comma-separated `RoleLevel` table) instead of the legacy `module_role_restriction` boolean table.

## Changes

- **`ModuleRoleRestrictionResponseDto`** — adds `restrictions: List<RoleLevel>` and `restrictableRoles: List<RoleLevel>`. Additive; `isAdmin`/`isManager` stay and are now *derived* from the role set rather than read from the legacy row.
- **`RolesServiceImpl`** — new shared read accessors `getRestrictedRoleLevels()` and `getRestrictedRoleLevels(ModuleType)`. A module with no row simply has no restrictions; no legacy fallback.
- **`PeopleUtil`** — three shared helpers for the new table: `getSecondaryRestrictionRole(ModuleType)` (the module manager-level role: `ESIGN → SENDER`, `CRM → SALES_MANAGER`, else `MANAGER`), `parseRestrictions(String)`, `toRestrictionsString(Set<RoleLevel>)`.
- **`PeopleMapper`** — removes `restrictRoleToRestrictRoleResponseDto`, which lost its only caller when the response stopped being a field-to-field map of the legacy entity.
- **`RolesServiceImplUnitTest`** — new; 16 tests covering both read paths and the write path.

The legacy write is untouched, so this phase stays revertible.

## Notes for reviewers

Two things reach slightly beyond the single endpoint:

1. `GET /v1/roles` (`processModuleRoles`) also switched from the legacy table to the new read accessor. Role-assignment enforcement therefore reads the new table as of this PR. Behaviourally inert as long as the two tables agree, which the Phase 0 gate verifies.
2. `buildModuleRolesRestriction` now serializes via `EnumSet`, so stored values are declaration-ordered and deduped — `[MANAGER, ADMIN]` used to store `"MANAGER,ADMIN"` and now stores `"ADMIN,MANAGER"`. Reads parse into a set either way, so no functional change, but any check that compares the stored string literally will see the normalization.

`getSecondaryRestrictionRole` is now `static`, so `EpRolesServiceImpl` can no longer override the module-to-manager-role mapping. It does not today.

## Testing

`mvn test -Dtest=RolesServiceImplUnitTest` — 16 passed. `mvn test-compile` clean, including the enterprise submodule. Format check (`spring-javaformat`) passes.


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR refactors module role restrictions from a boolean pair (is_admin/is_manager on the legacy module_role_restriction table) to a comma-separated RoleLevel string on the new module_roles_restriction table, adding parse/serialize helpers in PeopleUtil and good unit coverage. The core logic is sound and well-tested, but the read path now sources exclusively from the new table while nothing migrates existing legacy data, the legacy table is still written but never read, and a batch-loader helper is dead code.

Found **7** new issue(s): 🟡 1 suggestion(s), 🟣 2 nit(s)

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
```diff
@@ -130,54 +130,70 @@ private ModuleRolesRestriction buildModuleRolesRestriction(
...
+
+		moduleRolesRestrictionDao.findAll()
+			.forEach(moduleRolesRestriction -> restrictedRoleLevels.put(moduleRolesRestriction.getModule(),
+					PeopleUtil.parseRestrictions(moduleRolesRestriction.getRestrictions())));
+
+		return restrictedRoleLevels;
+	}
+
+	protected Set<RoleLevel> getRestrictedRoleLevels(ModuleType module) {
+		return moduleRolesRestrictionDao.findById(module)
```
ThinuwanW, line 155: 🤖 **Claude** · 💬 **Note**

The read path for restrictions now sources exclusively from the new `module_roles_restriction` table (getRestrictedRoleLevels), but no Liquibase data-migration changeset backfills it from the legacy `module_role_restriction` table (which is still written at line 120). If any deployment already has restrictions stored in the legacy table, after this deploy the new table is empty, so getAllowedRoles / validateRestrictedRoleAssignment will treat every module as unrestricted — silently re-enabling assignment of ADMIN/MANAGER roles that admins had restricted. This is an authorization gate weakening. Add a data-migration changeset that reads is_admin/is_manager per module and writes the equivalent comma-separated restrictions string, or confirm the legacy table was never populated in a released build.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
```diff
@@ -130,54 +130,70 @@ private ModuleRolesRestriction buildModuleRolesRestriction(
...
 		return moduleRolesRestriction;
 	}
 
-	private RoleLevel getSecondaryRestrictionRole(ModuleType module) {
-		return switch (module) {
-			case ESIGN -> RoleLevel.SENDER;
-			case CRM -> RoleLevel.SALES_MANAGER;
-			default -> RoleLevel.MANAGER;
-		};
+	protected Map<ModuleType, Set<RoleLevel>> getRestrictedRoleLevels() {
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

The no-arg `getRestrictedRoleLevels()` (findAll-based batch loader) is never called anywhere in community or enterprise code — it is dead code. It appears intended to let getAllowedRoles load all module restrictions in a single query instead of the current per-module findById inside processModuleRoles (line 254, invoked once per module in the stream). Either wire it into getAllowedRoles to eliminate the per-module query, or delete it.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
```diff
@@ -111,8 +111,8 @@ public ResponseEntityDto updateRoleRestrictions(ModuleRoleRestrictionRequestDto
 				&& !moduleRoleRestrictionRequestDto.getRestrictions().isEmpty()) {
 			List<RoleLevel> restrictions = moduleRoleRestrictionRequestDto.getRestrictions();
 			moduleRoleRestrictionRequestDto.setIsAdmin(restrictions.contains(RoleLevel.ADMIN));
-			moduleRoleRestrictionRequestDto.setIsManager(
-					restrictions.contains(getSecondaryRestrictionRole(moduleRoleRestrictionRequestDto.getModule())));
+			moduleRoleRestrictionRequestDto.setIsManager(restrictions
+				.contains(PeopleUtil.getSecondaryRestrictionRole(moduleRoleRestrictionRequestDto.getModule())));
 		}
 
 		ModuleRoleRestriction moduleRoleRestriction = peopleMapper
```
ThinuwanW, line 118: 🤖 **Claude** · 💬 **Note**

updateRoleRestrictions still persists to the legacy `moduleRoleRestrictionDao` (line 120), but the legacy table is no longer read by any code path (the new unit test even asserts verifyNoInteractions on it for reads). This makes it a write-only, effectively dead persistence that adds an extra DB write per update and can drift from the source-of-truth new table. If it is kept intentionally for rollback safety, add a comment stating so; otherwise remove the legacy write (and eventually the DAO/entity).

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
```diff
@@ -111,8 +111,8 @@ public ResponseEntityDto updateRoleRestrictions(ModuleRoleRestrictionRequestDto
 				&& !moduleRoleRestrictionRequestDto.getRestrictions().isEmpty()) {
 			List<RoleLevel> restrictions = moduleRoleRestrictionRequestDto.getRestrictions();
 			moduleRoleRestrictionRequestDto.setIsAdmin(restrictions.contains(RoleLevel.ADMIN));
-			moduleRoleRestrictionRequestDto.setIsManager(
-					restrictions.contains(getSecondaryRestrictionRole(moduleRoleRestrictionRequestDto.getModule())));
+			moduleRoleRestrictionRequestDto.setIsManager(restrictions
```
ThinuwanW, line 114: 🤖 **Claude** · 💬 **Note**

updateRoleRestrictions accepts and persists any RoleLevel in the `restrictions` list without validating it against getRestrictableRoles(module). A caller can store a meaningless/inconsistent restriction (e.g. EMPLOYEE for PEOPLE, or SALES_MANAGER for a non-CRM module), which is written verbatim yet never surfaced by isAdmin/isManager or restrictableRoles. Consider validating that the requested restrictions are a subset of the module's restrictable roles and rejecting invalid entries with a ModuleException/ValidationException using a message constant.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
```diff
@@ -130,54 +130,70 @@ private ModuleRolesRestriction buildModuleRolesRestriction(
...
-		}
+		Set<RoleLevel> restrictedRoleLevels = getRestrictedRoleLevels(module);
 
-		ModuleRoleRestriction moduleRoleRestriction = restrictedRole.get();
-		ModuleRoleRestrictionResponseDto moduleRoleRestrictionResponseDto = peopleMapper
-			.restrictRoleToRestrictRoleResponseDto(moduleRoleRestriction);
+		ModuleRoleRestrictionResponseDto moduleRoleRestrictionResponseDto = new ModuleRoleRestrictionResponseDto();
+		moduleRoleRestrictionResponseDto.setModule(module);
+		moduleRoleRestrictionResponseDto.setRestrictions(List.copyOf(restrictedRoleLevels));
+		moduleRoleRestrictionResponseDto.setRestrictableRoles(getRestrictableRoles(module));
```
ThinuwanW, line 181: 🤖 **Claude** · 🟡 **Suggestion**

getRestrictedRoleByModule now unconditionally calls getRestrictableRoles(module), which rebuilds a full EnumMap via initializeRolesForModule() on every invocation. This method is the workhorse of validateRestrictedRoleAssignment (line 548), which only reads getIsAdmin()/getIsManager() and never uses restrictableRoles. During a single validateRoles() call it is invoked once per module (5x in community, 6x in enterprise), so the restrictableRoles/EnumMap work is fully wasted on that path. Consider splitting the response building: keep the lightweight isAdmin/isManager derivation used by validation, and only compute restrictableRoles for the controller-facing read (or cache initializeRolesForModule()'s result).

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
```diff
@@ -130,54 +130,70 @@ private ModuleRolesRestriction buildModuleRolesRestriction(
...
-			return newRestrictRole;
-		}
+		Set<RoleLevel> restrictedRoleLevels = getRestrictedRoleLevels(module);
 
-		ModuleRoleRestriction moduleRoleRestriction = restrictedRole.get();
-		ModuleRoleRestrictionResponseDto moduleRoleRestrictionResponseDto = peopleMapper
-			.restrictRoleToRestrictRoleResponseDto(moduleRoleRestriction);
+		ModuleRoleRestrictionResponseDto moduleRoleRestrictionResponseDto = new ModuleRoleRestrictionResponseDto();
+		moduleRoleRestrictionResponseDto.setModule(module);
+		moduleRoleRestrictionResponseDto.setRestrictions(List.copyOf(restrictedRoleLevels));
```
ThinuwanW, line 180: 🤖 **Claude** · 🟣 **Nit**

The response 'restrictions' list is populated from whatever is stored (getRestrictedRoleLevels returns the raw parsed set), while 'restrictableRoles' is constrained to ADMIN + the module's secondary role. Because updateRoleRestrictions persists any RoleLevel without validating it against restrictable roles (first-pass finding #4), the two fields can disagree — e.g. a stored EMPLOYEE surfaces in 'restrictions' but never in 'restrictableRoles'. A UI that renders checkboxes from restrictableRoles and marks selection from restrictions would silently drop such a value. Consider intersecting restrictions with restrictableRoles before returning, so the response is internally consistent.

backend/src/main/java/com/skapp/community/peopleplanner/util/PeopleUtil.java
```diff
@@ -91,6 +100,47 @@ public static List<EmployeeManager> filterManagersByAttendanceRoles(List<Employe
...
+			}
+			catch (IllegalArgumentException e) {
+				log.warn("parseRestrictions: skipping unknown role level {}", roleLevel);
+			}
+		}
+
+		return roleLevels;
+	}
+
+	public static String toRestrictionsString(Set<RoleLevel> roleLevels) {
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

toRestrictionsString does EnumSet.copyOf(roleLevels) purely to force ordinal ordering before joining; every caller already passes an EnumSet (buildModuleRolesRestriction), so the copy is redundant defensive work. If the intent is to guarantee deterministic ordering regardless of caller, a brief comment stating that would prevent a future reader from removing it; otherwise stream roleLevels directly. Minor — flagging only for clarity, not correctness.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
anusham-rc: we already handled this migrations in a separate PR

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
anusham-rc: removed getRestrictedRoleLevels

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
anusham-rc: Intentional, the legacy write keeps this phase
it will be removed with the update api pr.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
anusham-rc: Pre-existing and unreachable this phase (the FE sends booleans only, never `restrictions`
next PR adds this validation when the payload becomes `add[]`/`remove[]`.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
anusham-rc: Deferred to GET /role api PR, which will cover the N+1 fix for this path.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
anusham-rc: Only reachable if invalid data is stored, which write-side validation prevents in a later PR,
filtering on read would hide bad rows rather than stop them.

backend/src/main/java/com/skapp/community/peopleplanner/util/PeopleUtil.java
anusham-rc: Only reachable if invalid data is stored, which update PR's validation prevents in a later phase  
filtering on read would hide bad rows rather than stop them.

backend/src/main/java/com/skapp/community/peopleplanner/util/PeopleUtil.java
```diff
@@ -91,6 +100,47 @@ public static List<EmployeeManager> filterManagersByAttendanceRoles(List<Employe
...
+		for (String restriction : restrictions.split(RESTRICTION_SEPARATOR)) {
+			String roleLevel = restriction.trim();
+			if (roleLevel.isEmpty()) {
+				continue;
+			}
+			try {
+				roleLevels.add(RoleLevel.valueOf(roleLevel));
+			}
+			catch (IllegalArgumentException e) {
+				log.warn("parseRestrictions: skipping unknown role level {}", roleLevel);
```
shakila-kamalasena: remove log warn

shakila-kamalasena (dismissed):

backend/src/main/java/com/skapp/community/peopleplanner/util/PeopleUtil.java
anusham-rc: removed and changed the implementation

backend/src/test/java/com/skapp/community/peopleplanner/service/RolesServiceImplUnitTest.java
```diff
@@ -0,0 +1,270 @@
...
+import org.mockito.junit.jupiter.MockitoExtension;
+
+import java.util.List;
+import java.util.Optional;
+
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+@ExtendWith(MockitoExtension.class)
+class RolesServiceImplUnitTest {
```
akilarootcode: remvoe this test

akilarootcode (changes requested):

backend/src/test/java/com/skapp/community/peopleplanner/service/RolesServiceImplUnitTest.java
anusham-rc: removed, and added integration tests

shakila-kamalasena (dismissed):

akilarootcode (approved):

shakila-kamalasena (approved):

PR #1678, opened 2026-07-27
feat: render module role restrictions from the restrictions list

anusham-rc:
### Summary

- Drive the restricted user roles modal from the backend response instead of a hardcoded per-module switch. `restrictableRoles` decides which checkboxes render and `restrictions` seeds which ones are checked, so a new module or role level needs no frontend change.
- Add a `RoleLevel` enum (`ADMIN`, `MANAGER`, `SENDER`, `EMPLOYEE`, `GUEST`, `NONE`, `SALES_MANAGER`, `SALES_REPRESENTATIVE`) and a label-key map, plus the missing `salesRepresentativeRoleLabel`, `employeeRoleLabel`, `guestRoleLabel` and `noneRoleLabel` translations.
- Replace the `isAdmin` / `isManager` / `isSender` formik fields with a single `selected: RoleLevel[]`. This removes the ESIGN-specific mapping that reused `isManager` for the Sender role.
- Split the update payload into `UserRoleRestrictionsUpdateType`. The PATCH endpoint still only accepts the boolean pair, so `isManager` is derived from whichever manager level role (`MANAGER`, `SENDER`, `SALES_MANAGER`) is selected. Both the type and the mapping are marked deprecated and go away when the backend moves to an add/remove delta payload.
- Drop the unused `useGetUserRoleRestrictions` hook and give `getUserRoleRestrictions` an explicit return type.
- Guard the module page against `Modules.NONE` and move `setIsPending(false)` into a `finally` so a failed fetch no longer leaves the button stuck in the pending state.

### How to test

1. Go to **Configurations → User Roles** and open any module (People, Leave, Attendance, E-Sign, Invoice, Project Management, CRM).
2. Click the primary button to open the **Restricted User Roles** modal. Confirm the checkboxes match the `restrictableRoles` returned by the endpoint for that module, with correct labels (CRM shows Sales Manager, E-Sign shows Sender, etc.).
3. Confirm the already restricted roles come back pre-checked from `restrictions`.
4. Toggle roles and save. Verify the request succeeds, the success toast appears, and reopening the modal shows the persisted selection.
5. Open the modal without changing anything and confirm the save button stays disabled until the form is dirty.
6. Simulate a failed restrictions fetch (throttle or block the GET) and confirm the button leaves the pending state instead of hanging.

### Project Checklist

  - [x] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information

- Depends on the backend returning `restrictions` and `restrictableRoles` on `GET /user-role-restrictions/{module}`. Without those fields the modal renders no checkboxes.
- The `isAdmin` / `isManager` booleans are kept on `UserRoleRestrictionsType` and on the PATCH payload only for backwards compatibility, and are marked `@deprecated`. Both are removed once the backend switches to the delta payload.


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR refactors the RestrictedUserRolesModal from hardcoded per-module role logic to a data-driven approach that renders checkboxes from backend-provided `restrictableRoles` and preselects from `restrictions`. The new `RoleLevel` enum correctly matches the backend, and the isAdmin/isManager collapse logic mirrors the backend's transitional mapping. However, the frontend now depends entirely on two response fields (`restrictableRoles`, `restrictions`) that the current backend GET endpoint does not return, which would leave the modal with zero checkboxes.

Found **10** new issue(s): 🔴 1 important, 🟡 3 suggestion(s), 🟣 1 nit(s)

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
```diff
@@ -63,53 +84,40 @@ const RestrictedUserRolesModal = ({ initialData }: Props) => {
...
-      case Modules.INVOICE:
-      case Modules.PM:
-        return ["isAdmin"];
-      case Modules.CRM:
-        return ["isAdmin", "isManager"];
-      default:
-        return [];
-    }
-  };
+  const restrictableRoles = initialData?.restrictableRoles ?? [];
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

Breaking dependency on backend fields that don't exist yet. The modal now renders checkboxes exclusively from `initialData?.restrictableRoles` (line 112) and preselects from `initialData?.restrictions` (line 106), but the backend GET `/v1/roles/restrictions/{module}` response DTO (`ModuleRoleRestrictionResponseDto`, used by both community `RolesServiceImpl` and enterprise `EpRolesServiceImpl`) only returns `module`, `isAdmin`, and `isManager`. With the current backend both new fields are `undefined`, so `restrictableRoles ?? []` yields no checkboxes and `restrictions ?? []` preselects nothing — the modal becomes completely non-functional (regression from the previous local `getRestrictableRoles()` logic). Confirm a coordinated backend change adds `restrictions` and `restrictableRoles` to the response DTO before merging; otherwise this must not ship.

frontend/src/community/configurations/types/UserRolesTypes.ts
```diff
@@ -27,6 +27,14 @@ export interface UserRoleRestrictionsType {
   module: Modules;
   isAdmin: boolean;
   isManager: boolean;
+  restrictions: RoleLevel[];
```
ThinuwanW, line 30: 🤖 **Claude** · 💬 **Note**

`restrictions` and `restrictableRoles` are declared as required (`RoleLevel[]`) on `UserRoleRestrictionsType`, but the backend response can omit them (they are `undefined` at runtime with the current API). This makes the type lie about the runtime shape and defeats the `?? []` guards (TS considers them unreachable when `initialData` is defined). Mark them optional (`restrictions?: RoleLevel[]; restrictableRoles?: RoleLevel[];`) to accurately model the response and keep the null-safety guards meaningful.

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
```diff
@@ -63,53 +84,40 @@ const RestrictedUserRolesModal = ({ initialData }: Props) => {
...
-    // TODO: For ESIGN, the backend uses isManager to represent Sender role.
-    // This mapping will be removed after the backend's migration is completed.
-    const payload: UserRoleRestrictionsType = {
+    const payload: UserRoleRestrictionsUpdateType = {
       module: moduleType,
-      isAdmin: values.isAdmin,
-      isManager:
-        moduleType === Modules.ESIGN ? values.isSender : values.isManager
+      isAdmin: values.selected.includes(RoleLevel.ADMIN),
+      isManager: values.selected.some((roleLevel) =>
```
ThinuwanW, line 98: 🤖 **Claude** · 💬 **Note**

The frontend re-implements backend business logic. `SECONDARY_ROLE_LEVELS` + the isAdmin/isManager derivation here duplicates `getSecondaryRestrictionRole()` and the identical `restrictions.contains(...)` mapping already present in the backend `updateRoleRestrictions` (RolesServiceImpl lines 110-116), which already accepts and prefers a `restrictions` array. Since the backend supports the array payload, consider sending `restrictions: values.selected` directly and letting the backend own the collapse, avoiding two copies of the same mapping drifting out of sync (e.g. if a module ever gets a secondary role not in this hardcoded set). If the phased approach is intentional, this is fine to defer.

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
```diff
@@ -63,53 +84,40 @@ const RestrictedUserRolesModal = ({ initialData }: Props) => {
...
-  };
+  const restrictableRoles = initialData?.restrictableRoles ?? [];
+
+  const toggleRoleLevel = (roleLevel: RoleLevel) => {
+    const selected = values.selected.includes(roleLevel)
+      ? values.selected.filter((selectedRole) => selectedRole !== roleLevel)
+      : [...values.selected, roleLevel];
 
-  const restrictableRoles = getRestrictableRoles(moduleType);
+    setFieldValue("selected", selected);
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

`toggleRoleLevel` appends newly-checked roles to the end of the array, so unchecking then re-checking a role changes element order relative to `initialValues`. Formik's `dirty` uses a deep equality check, so `[ADMIN, MANAGER]` vs `[MANAGER, ADMIN]` reports as dirty even though the selected set is unchanged, spuriously enabling the Save button. Compare as sets (or keep a stable order) when computing selection to avoid a no-op save being enabled.

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
```diff
@@ -5,16 +5,37 @@ import { useFormik } from "formik";
...
+  [RoleLevel.MANAGER]: "managerRoleLabel",
+  [RoleLevel.SENDER]: "senderRoleLabel",
+  [RoleLevel.SALES_MANAGER]: "salesManagerRoleLabel",
+  [RoleLevel.EMPLOYEE]: "employeeRoleLabel",
+  [RoleLevel.GUEST]: "guestRoleLabel",
+  [RoleLevel.NONE]: "noneRoleLabel",
+  [RoleLevel.SALES_REPRESENTATIVE]: "salesRepresentativeRoleLabel"
+};
+
+/** Role levels the backend reports through the deprecated isManager flag. and will be removed in the future */
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

Minor grammar/typo in the doc comment: "...through the deprecated isManager flag. and will be removed in the future" — the sentence breaks mid-way (period followed by lowercase "and"). Reword to e.g. "...through the deprecated isManager flag, and will be removed in the future."

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
```diff
@@ -63,53 +84,40 @@ const RestrictedUserRolesModal = ({ initialData }: Props) => {
...
+    selected: RoleLevel[];
+  }>({
     initialValues: {
-      isAdmin: initialData !== undefined ? initialData?.isAdmin : false,
-      isManager: initialData !== undefined ? initialData?.isManager : false,
-      isSender:
-        initialData !== undefined && moduleType === Modules.ESIGN
-          ? initialData?.isManager
-          : false
+      selected: initialData?.restrictions ?? []
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Behavioral regression in initial checked state: the form's `selected` is initialized from `initialData?.restrictions`, but the backend response DTO (ModuleRoleRestrictionResponseDto: getRestrictedRoleByModule) only returns `module`, `isAdmin`, and `isManager` — it never returns `restrictions`. So `selected` is always `[]` at runtime and previously-saved restrictions will NEVER show as checked when the modal opens, even though the backend does report them via isAdmin/isManager. The old code correctly seeded checkboxes from `initialData.isAdmin`/`isManager`/`isSender`. This is distinct from the missing-checkbox-list issue (restrictableRoles): even if the list rendered, nothing would be pre-checked. Fix: derive `selected` from the isAdmin/isManager flags the backend actually returns (mapping the module's secondary role), or extend the backend response DTO to include `restrictions` before shipping.

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
```diff
@@ -63,53 +84,40 @@ const RestrictedUserRolesModal = ({ initialData }: Props) => {
...
   };
 
+  // This endpoint still accepts only the isAdmin/isManager pair, where
+  // isManager stands for whichever manager level role the module has. Replaced
+  // by an add/remove delta payload in the next phase.
   const handleSubmit = () => {
-    // TODO: For ESIGN, the backend uses isManager to represent Sender role.
-    // This mapping will be removed after the backend's migration is completed.
-    const payload: UserRoleRestrictionsType = {
+    const payload: UserRoleRestrictionsUpdateType = {
```
ThinuwanW, line 95: 🤖 **Claude** · 🟡 **Suggestion**

Lossy/silent-drop mapping on submit: the payload only sends `isAdmin` and `isManager` (derived via SECONDARY_ROLE_LEVELS). Any restrictable role that is neither ADMIN nor a secondary manager-level role (SALES_REPRESENTATIVE, EMPLOYEE, GUEST, NONE — all present in RoleLevel and the label map) can be checked in the UI but is silently discarded, because it maps to neither boolean. The backend request DTO (ModuleRoleRestrictionRequestDto) already accepts and prefers a `restrictions: RoleLevel[]` field (see RolesServiceImpl.updateRoleRestrictions lines 110-116). Fix: send `restrictions: values.selected` in the update payload instead of collapsing to two booleans — this also removes the client-side re-implementation of getSecondaryRestrictionRole. If deferring to a later phase, restrict the rendered checkboxes to representable roles so users can't make selections that are silently lost.

frontend/src/community/configurations/types/UserRolesTypes.ts
```diff
@@ -27,6 +27,14 @@ export interface UserRoleRestrictionsType {
...
+  restrictableRoles: RoleLevel[];
+}
+
+export interface UserRoleRestrictionsUpdateType {
```
ThinuwanW, line 34: 🤖 **Claude** · 🟡 **Suggestion**

Code duplication: `UserRoleRestrictionsUpdateType` re-declares the exact three fields (module/isAdmin/isManager) already present on `UserRoleRestrictionsType`. If the base type changes, these drift silently. Define it as `Pick<UserRoleRestrictionsType, "module" | "isAdmin" | "isManager">` (or make `UserRoleRestrictionsType` extend it) so the shared shape stays in sync.

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
```diff
@@ -129,34 +137,15 @@ const RestrictedUserRolesModal = ({ initialData }: Props) => {
...
-            )}
-            {restrictableRoles.includes("isSender") && (
+            {restrictableRoles.map((roleLevel) => (
               <Checkbox
-                label={translateText(["senderRoleLabel"])}
-                name="isSender"
-                checked={values.isSender}
-                onChange={() => setFieldValue("isSender", !values.isSender)}
+                key={roleLevel}
+                label={translateText([ROLE_LEVEL_LABEL_KEYS[roleLevel]])}
```
ThinuwanW, line 148: 🤖 **Claude** · 🟣 **Nit**

Runtime null-safety: `ROLE_LEVEL_LABEL_KEYS[roleLevel]` and `translateText([...])` assume every entry of `restrictableRoles` is a valid `RoleLevel`. That array is sourced from an untyped JSON API response (getUserRoleRestrictions returns `data.data.results[0]`), so a role string the frontend enum doesn't know about yields `undefined` as the key, producing a broken/empty label rather than failing safely. Consider filtering `restrictableRoles` to known `RoleLevel` values (or guarding the lookup) before rendering.

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
```diff
@@ -5,16 +5,37 @@ import { useFormik } from "formik";
 import Checkbox from "~community/common/components/atoms/Checkbox/Checkbox";
 import Icon from "~community/common/components/atoms/Icon/Icon";
 import Tooltip from "~community/common/components/atoms/Tooltip/Tooltip";
-import { Modules } from "~community/common/enums/CommonEnums";
+import { Modules, RoleLevel } from "~community/common/enums/CommonEnums";
```
ThinuwanW, line 8: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Missing test coverage for a security-adjacent refactor: the modal's form model changed fundamentally (three booleans + per-module getRestrictableRoles switch → a RoleLevel[] with toggle/mapping logic), yet there is no `RestrictedUserRolesModal.test.tsx`. Per the testing guideline, add tests (with MockTheme) covering toggleRoleLevel, the isAdmin/isManager derivation in handleSubmit, and empty/undefined initialData, so the mapping to the backend payload is pinned before further phases change it.

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
anusham-rc: added both restrictions and restrictableRoles to the response DTO.

anusham-rc: Backend change is #1677 — adds both `restrictions` and `restrictableRoles` to the response DTO. Merging that first.

frontend/src/community/configurations/types/UserRolesTypes.ts
anusham-rc: #1677 always returns both as arrays, never null, required is the accurate shape. The `?? []` still covers `initialData === undefined`.

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
anusham-rc: Intentional this phase, the update API still takes booleans; removed when it becomes `add[]`/`remove[]`.

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
anusham-rc: Fixed, comparing the selection as a set instead of Formik's deep-equality `dirty`, which reports a reorder as a change.

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
anusham-rc: Fixed.

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
anusham-rc: #1677 returns `restrictions` now, with `isAdmin`/`isManager` derived from it.

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
anusham-rc: Can't happen, `restrictableRoles` is only ADMIN + the module's secondary role, which is exactly what the two booleans carry.

frontend/src/community/configurations/types/UserRolesTypes.ts
anusham-rc: The update type becomes `add[]`/`remove[]` next phase, so a `Pick<>` link would be unwound right away.

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
anusham-rc: Backend already skips unknown levels on read, and `restrictableRoles` comes from the enum itself.

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
anusham-rc: Extracted the toggle and the change detection into `configurations/utils/userRoles/roleRestrictionUtils.ts` with 12 tests, including the reorder case. The `isAdmin`/`isManager` derivation stays inline and untested 
it's the transitional shim and goes with the update-API PR.

shakila-kamalasena (dismissed):

akilarootcode (dismissed):

shakila-kamalasena (dismissed):

akilarootcode (approved):

shakila-kamalasena (approved):

PR #1683, opened 2026-07-28
feat: add search functionality for deal ID in CRM deal queries

dulsara-manakal:
### Summary

- Search for a Deal Using its Deal ID

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds deal-ID matching to the existing CRM deal search predicates in both `buildPredicates` (paginated deal list) and `addSearchKeywordPredicates` (board-by-stage/count queries), applied symmetrically and consistently. The implementation is mechanically correct, respects the `isDeleted` filter, and is covered by integration tests including a soft-delete case. The main consideration is that IDs are matched with a substring `LIKE %keyword%` rather than an exact match, which can surface unintended results.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmDealRepositoryImpl.java
```diff
@@ -112,7 +112,9 @@ private List<Predicate> buildPredicates(CriteriaBuilder cb, Root<CrmDeal> deal,
 			String keyword = "%" + filterDto.getSearchKeyword().toLowerCase() + "%";
 			Join<CrmDeal, CrmContact> contactJoin = deal.join(CrmDeal_.contact, JoinType.LEFT);
 			Join<CrmDeal, Employee> ownerJoin = deal.join(CrmDeal_.owner, JoinType.LEFT);
-			predicates.add(cb.or(cb.like(cb.lower(deal.get(CrmDeal_.name)), keyword),
+
+			predicates.add(cb.or(cb.like(deal.get(CrmDeal_.id).as(String.class), keyword),
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Deal ID is matched with `cb.like(id.as(String.class), "%keyword%")`, i.e. substring matching. A user searching the keyword "5" will match every deal whose ID *contains* 5 (5, 15, 25, 50, 500, ...), and any numeric keyword also matches the OR'd name/owner/contact LIKEs. For IDs, exact match is usually the intended UX. Consider matching the ID exactly and only when the keyword is fully numeric — e.g. guard `if (searchKeyword.matches("\\d+")) predicates.add(cb.equal(deal.get(CrmDeal_.id), Long.valueOf(searchKeyword)))` combined into the OR, or at minimum `cb.equal(deal.get(CrmDeal_.id).as(String.class), searchKeyword)` for exact-string ID matching. This same substring behavior applies to the identical predicate at line 168; fix both together to keep the two search paths symmetric.

backend/src/test/java/com/skapp/community/crmplanner/controller/v1/CrmDealControllerIntegrationTest.java
```diff
@@ -316,6 +316,42 @@ void getDeals_SalesRep_ReturnsOnlyOwnDeals() throws Exception {
 			.andExpect(jsonPath(RESULTS_0_PATH + "['items'][0]['name']").value("Rep Deal"));
 	}
 
+	@Test
```
ThinuwanW, line 319: 🤖 **Claude** · 🟣 **Nit**

Because ID matching is substring-based (`LIKE %id%`), these tests are subtly coupled to the auto-increment values assigned at run time. If the target deal and the "Unrelated Deal" ever receive IDs where one is a substring of the other (e.g. 5 vs 15, or when IDENTITY values overlap across seeded data), the `.length()==1` assertion becomes flaky. It's safe with adjacent multi-digit IDs today, but if the production predicate is changed to exact match (see the repository finding) these tests remain valid; if it stays as LIKE, consider asserting the target ID is present within the result set rather than asserting an exact count of 1, to harden against ID-substring collisions.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmDealRepositoryImpl.java
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Code duplication: the 6-clause search OR (id, name, contact name, owner first/last/full name) is now copied verbatim in two places — inlined here in buildPredicates (lines 116-122) and again in addSearchKeywordPredicates (lines 168-174). This PR had to add the identical `cb.like(deal.get(CrmDeal_.id).as(String.class), keyword)` line to both copies, which is exactly the drift risk duplication creates: a future change to the searchable fields must remember both sites. Fix: have buildPredicates delegate to addSearchKeywordPredicates (both already build the same joins and keyword string) so the search clause lives in one method. Severity suggestion since the behavior is currently correct.

backend/src/test/java/com/skapp/community/crmplanner/controller/v1/CrmBoardControllerIntegrationTest.java
```diff
@@ -439,6 +439,23 @@ void getDealsByStages_SalesRep_ReturnsOnlyOwnDeals() throws Exception {
 			.andExpect(jsonPath(RESULTS_0_PATH + "['deals'][0]['name']").value("Rep Deal"));
 	}
 
+	@Test
```
ThinuwanW, line 442: 🤖 **Claude** · 🟡 **Suggestion**

Test-coverage gap for the board endpoint's new ID search. The board path exercises TWO query methods that both got the new id predicate: findDealsByStageId (the fetched deals) and countDealsByStageIds (the per-stage total). This test only asserts `['deals'].length()` == 1, so the count-query path with an ID keyword is never verified — a mismatch between the fetch predicate and the count predicate for ID search would go undetected. Also, unlike the deal-controller suite (which added getDeals_SearchKeywordMatchesSoftDeletedDealId_ReturnsEmptyList), there is no soft-deleted-ID case here. Suggest adding `.andExpect(jsonPath(RESULTS_0_PATH + "['totalCount']").value(1))` and a soft-deleted-ID test for parity.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmDealRepositoryImpl.java
dulsara-manakal: That is the requirement and also aligns with skapp-pm.

backend/src/test/java/com/skapp/community/crmplanner/controller/v1/CrmDealControllerIntegrationTest.java
dulsara-manakal: `deal` and `"Unrelated Deal"` are created back-to-back in the same test, so their auto-increment IDs are sequential neighbours (N/N+1).  like 14 and 15. One is never hiding inside the other. So this bug cannot happen in our test, because we always make the two deals one after another.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmDealRepositoryImpl.java
dulsara-manakal: fixed this.

backend/src/test/java/com/skapp/community/crmplanner/controller/v1/CrmBoardControllerIntegrationTest.java
dulsara-manakal: Added test

yuhanga-rtc (approved):

akilarootcode (approved):

PR #1685, opened 2026-07-29
Feature/birthday notification be implementation

gaindurootcode:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds an org-wide birthday-notification feature: a new `SpecialNotificationStatus` entity + Liquibase changeset, a generic `SpecialNotificationService` for org-config-backed notification settings and per-employee last-viewed tracking, a `PeopleConfigController` for admin config, two employee-facing endpoints on `PeopleController`, and a Criteria-API birthday lookup with SELF/TEAM/ORGANIZATION scoping plus a Feb-29 fallback. Overall the layering, naming, `@Transactional`/`@PreAuthorize` usage, mapper wiring, DDL/entity alignment (verified field-by-field, including `Auditable` columns and the composite PK) and integration-test coverage are solid, and the enterprise `EpPeopleServiceImpl` super-constructor has already been updated for the two new `PeopleServiceImpl` dependencies. The main correctness problem is the `DataIntegrityViolationException` swallow in `markNotificationViewed`, which cannot actually recover the transaction; the rest are validation, error-wrapping, efficiency and design/consistency issues.

Found **28** new issue(s): 🔴 1 important, 🟡 17 suggestion(s), 🟣 10 nit(s)

backend/src/main/java/com/skapp/community/common/service/impl/SpecialNotificationServiceImpl.java
```diff
@@ -0,0 +1,96 @@
...
+				newStatus.setEmployee(employeeDao.getReferenceById(employeeId));
+				newStatus.setSpecialNotificationType(type);
+				return newStatus;
+			});
+		specialNotificationStatus.setLastViewedDate(viewedDate);
+
+		try {
+			specialNotificationStatusDao.saveAndFlush(specialNotificationStatus);
+		}
+		catch (DataIntegrityViolationException e) {
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Catching `DataIntegrityViolationException` after `saveAndFlush` does not make the concurrent-insert case recoverable. When the flush fails, Hibernate has already marked the persistence context / transaction as rollback-only (JPA-mandated behaviour), so swallowing the exception and logging "treating as success" does not help: `markNotificationViewed` joins the caller's transaction (`PeopleServiceImpl.markTodayBirthdayNotificationsAsViewed` is `@Transactional`), and when Spring commits it will throw `UnexpectedRollbackException` → HTTP 500 for the user, with a misleading WARN in the logs saying it succeeded. Also note the entity is saved via `merge` (composite `@IdClass` with all id attributes populated ⇒ Spring Data's `isNew` is false), i.e. SELECT-then-INSERT, so the duplicate-key window is real. Fix options: (a) push the write into its own transaction so the failure is contained — e.g. move the save into a separate bean method annotated `@Transactional(propagation = Propagation.REQUIRES_NEW)` and catch around that call; (b) do a real upsert (`INSERT ... ON DUPLICATE KEY UPDATE last_viewed_date = ?` via a native/`@Modifying` query, or an `UPDATE`-then-`INSERT`-if-0-rows in the repository impl); or (c) drop the try/catch and the `saveAndFlush` (use `save`) and let the rare conflict surface as a normal error rather than a silently-broken transaction.

backend/src/main/java/com/skapp/community/peopleplanner/controller/v1/PeopleConfigController.java
```diff
@@ -0,0 +1,45 @@
...
+		ResponseEntityDto response = peopleConfigService.getBirthdayNotificationConfigs();
+		return new ResponseEntity<>(response, HttpStatus.OK);
+	}
+
+	@Operation(summary = "Update birthday notification configuration",
+			description = "Partially updates the organization-wide configuration controlling birthday notifications.")
+	@PatchMapping(value = "/birthday-notifications")
+	@PreAuthorize("hasAnyRole('ROLE_PEOPLE_ADMIN')")
+	public ResponseEntity<ResponseEntityDto> updateBirthdayNotificationConfigs(
+			@RequestBody BirthdayNotificationConfigRequestDto requestDto) {
```
ThinuwanW, line 40: 🤖 **Claude** · 🟡 **Suggestion**

`@RequestBody BirthdayNotificationConfigRequestDto requestDto` is missing `@Valid`, and the DTO itself carries no bean-validation annotations, so nothing is validated at the boundary. Every other `@RequestBody` in `PeopleController` uses `@Valid` (e.g. `addNewEmployee`, `updateEmployee`). Add `@Valid` here, and add the constraints you actually want on the DTO (see the separate comment about the scope flags). Also consider `@RequestBody(required = false)` or documenting that an empty body is `{}` — a PATCH with no body currently returns 400 rather than a no-op.

backend/src/main/java/com/skapp/community/peopleplanner/payload/request/BirthdayNotificationConfigRequestDto.java
```diff
@@ -0,0 +1,16 @@
...
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+public class BirthdayNotificationConfigRequestDto {
+
+	private Boolean isTurnedOn;
+
+	private Boolean isOrganizationWide;
```
ThinuwanW, line 12: 🤖 **Claude** · 🟡 **Suggestion**

`isOrganizationWide` and `isTeamWide` encode a tri-state (SELF / TEAM / ORGANIZATION) as two independent booleans, so the persisted config can hold contradictory state (`both true`), and `PeopleServiceImpl.resolveBirthdayNotificationScope` silently resolves that by preferring ORGANIZATION — a rule only discoverable by reading the impl (the integration test `...WithBothOrganizationAndTeamFlags_UsesOrganizationScope` pins the behaviour but doesn't make it valid input). Worse, `isTurnedOn = true` with both flags false (the state produced by a PATCH that only sets `isTurnedOn`) silently degrades to SELF-only notifications, which is almost certainly not what an admin who "turned birthdays on" expects. Recommend replacing the two booleans with a single `BirthdayNotificationScope` enum field (`@NotNull` on PATCH-supplied values) alongside `isTurnedOn`; if the two-boolean shape must be kept for the frontend contract, at least reject `isOrganizationWide && isTeamWide` with a `ModuleException` + message constant, and document/choose an explicit default when notifications are on but no scope is set.

backend/src/main/java/com/skapp/community/common/service/impl/SpecialNotificationServiceImpl.java
```diff
@@ -0,0 +1,96 @@
...
+
+	@Override
+	@Transactional(readOnly = true)
+	public <T extends SpecialNotificationConfig> T getConfig(SpecialNotificationType type, Class<T> configClass) {
+		String configValue = organizationConfigDao
+			.findOrganizationConfigByOrganizationConfigType(type.getOrganizationConfigType().name())
+			.map(OrganizationConfig::getOrganizationConfigValue)
+			.orElse("{}");
+
+		return objectMapper.readValue(configValue, configClass);
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`objectMapper.readValue(configValue, configClass)` is unguarded. `configValue` comes from the `organization_config` table, so any malformed/legacy JSON (hand-edited row, schema drift after a config-class change, or a value written by an older/newer version) throws an unchecked `JacksonException` that escapes to `GlobalExceptionHandler` as a raw 500 rather than a structured `ErrorResponse`. This is exactly the checklist item for JSON parsing. The codebase already has the pattern and the message key — wrap it like `EsMigrationDocumentRepairServiceImpl` does: catch and `throw new ModuleException(CommonMessageConstant.COMMON_ERROR_JSON_STRING_TO_OBJECT_CONVERSION_FAILED)`. Since this is a config read on a hot employee-facing path (`getTodayBirthdayNotifications`), falling back to the all-false default and logging `log.error("...", e)` may be even better than failing the request.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
```diff
@@ -1472,6 +1486,75 @@ public ResponseEntityDto reassignSupervisorsAndTerminateOrDeleteEmployee(Long us
...
+			return BirthdayNotificationScope.ORGANIZATION;
+		}
+		if (Boolean.TRUE.equals(birthdayNotificationConfig.getIsTeamWide())) {
+			return BirthdayNotificationScope.TEAM;
+		}
+		return BirthdayNotificationScope.SELF;
+	}
+
+	private LocalDate resolveBirthdayNotificationDate() {
+		return LocalDate.now(ZoneId.of(organizationService.getOrganizationTimeZone()));
```
ThinuwanW, line 1627: 🤖 **Claude** · 🟡 **Suggestion**

`ZoneId.of(organizationService.getOrganizationTimeZone())` throws `NullPointerException` when the organization row exists but `organization_time_zone` is null. `OrganizationServiceImpl.getOrganizationTimeZone()` only defaults to `"UTC"` when *no* organization row is found (`orElse("UTC")` on the `Optional`); the column is nullable (`Organization.organizationTimeZone`, `@Column(length = 100)`) and `OrganizationDto.organizationTimeZone` has no `@NotNull`, so a tenant created without a timezone will 500 on every call to both new endpoints. An invalid/legacy zone string would similarly throw `ZoneRulesException`. The enterprise module already has `EpDateTimeUtils.resolveZoneId(String)` for exactly this; add an equivalent null-/invalid-safe resolve in `DateTimeUtils` (defaulting to UTC) and use it here.

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/EmployeeRepositoryImpl.java
```diff
@@ -1631,4 +1634,80 @@ public void updateLastClockInDate(Long employeeId, LocalDate lastClockInDate) {
...
+		predicates.add(cb.isNotNull(personalInfoJoin.get(EmployeePersonalInfo_.birthDate)));
+		predicates.add(buildBirthdayOnDatePredicate(cb, personalInfoJoin, date));
+		predicates.add(buildBirthdayScopePredicate(cb, criteriaQuery, root, currentEmployeeId, scope));
+
+		criteriaQuery.select(root)
+			.where(predicates.toArray(new Predicate[0]))
+			.distinct(true)
+			.orderBy(cb.asc(cb.lower(root.get(Employee_.firstName))), cb.asc(cb.lower(root.get(Employee_.lastName))));
+
+		return entityManager.createQuery(criteriaQuery).getResultList();
```
ThinuwanW, line 1667: 🤖 **Claude** · 🟡 **Suggestion**

`findEmployeesWithBirthdayOn` returns full `Employee` entities and the mapper then reads `user.email` for each one. `Employee.user` is a default-EAGER `@OneToOne`, and the `user` table is only *implicitly joined* here for the `isActive` predicate — an implicit join does not fetch — so Hibernate issues one extra SELECT per returned employee (classic N+1). Add an explicit fetch (`root.fetch(Employee_.user)`, or reuse a single `Join` for both the predicate and a fetch) or, better, select a tuple/DTO projection of just `employeeId, firstName, lastName, authPic, email` since only five scalar fields are ever used. Separately, `MONTH(birth_date)`/`DAY(birth_date)` are non-sargable, so this is a full scan of `employee_personal_info` joined to `employee` on every app open per user; if the employee table grows, consider a persisted/generated `birth_month`/`birth_day` pair with an index, or caching the day's result set.

backend/src/main/java/com/skapp/community/common/service/impl/SpecialNotificationServiceImpl.java
```diff
@@ -0,0 +1,96 @@
...
+
+	private final SpecialNotificationStatusDao specialNotificationStatusDao;
+
+	private final EmployeeDao employeeDao;
+
+	private final JsonMapper objectMapper;
+
+	@Override
+	@Transactional(readOnly = true)
+	public <T extends SpecialNotificationConfig> T getConfig(SpecialNotificationType type, Class<T> configClass) {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`getConfig` (line 36) and `getLastViewedDate` (line 64) have no entry logging, while `saveConfig` and `markNotificationViewed` in the same class follow the convention. Add `log.info("getConfig: execution started for type {}", type)` / `log.info("getLastViewedDate: execution started for employee {} and type {}", employeeId, type)` (and matching "execution ended" lines) so the new service is consistent with the rest of the codebase and traceable.

backend/src/main/java/com/skapp/community/common/payload/SpecialNotificationConfig.java
```diff
@@ -0,0 +1,5 @@
+package com.skapp.community.common.payload;
+
+public interface SpecialNotificationConfig {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`SpecialNotificationConfig` is an empty marker interface, and `<T extends SpecialNotificationConfig> T getConfig(SpecialNotificationType type, Class<T> configClass)` takes the type and the config class as two unrelated parameters. Nothing prevents `getConfig(SpecialNotificationType.BIRTHDAY, SomeOtherConfig.class)` — the generic bound buys no real safety, and every future notification type must remember the correct pairing at each call site. Consider binding the config class to the enum (e.g. `SpecialNotificationType.BIRTHDAY(OrganizationConfigType.BIRTHDAY_NOTIFICATIONS, BirthdayNotificationConfig.class)` with `getConfigClass()`), so `getConfig(type)` can return the right type and `saveConfig` can assert the config matches the type. If the marker interface stays, it should at least document what implementors must guarantee (JSON round-trippable, all fields defaulted).

backend/src/main/java/com/skapp/community/common/payload/response/BirthdayNotificationConfigResponseDto.java
```diff
@@ -0,0 +1,17 @@
+package com.skapp.community.common.payload.response;
+
+import com.skapp.community.common.payload.SpecialNotificationConfig;
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+public class BirthdayNotificationConfigResponseDto implements SpecialNotificationConfig {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

This class serves three roles at once: the API response DTO, the *persisted* JSON schema in `organization_config.config_value` (via `saveConfig`), and the mutable patch target in `PeopleConfigServiceImpl.updateBirthdayNotificationConfigs`. That couples the on-disk config format to a `*ResponseDto` — renaming or restructuring the response later silently breaks reading already-stored rows, and the field defaults (`= false`) exist only to make the `"{}"` fallback work. Recommend splitting: a `BirthdayNotificationConfig` model in `common.payload` (or better, in the people module) that is persisted and mutated, and a thin `BirthdayNotificationConfigResponseDto` mapped from it for the API. Related placement point: the birthday-specific request/response DTOs live under `common.payload.request`/`common.payload.response` but are referenced only by `peopleplanner` (`PeopleConfigService`, `PeopleConfigController`, `PeopleServiceImpl`) — only the generic `SpecialNotification*` pieces belong in `common`; the birthday-specific ones belong in `peopleplanner.payload`.

backend/src/test/java/com/skapp/community/peopleplanner/controller/v1/PeopleBirthdayNotificationControllerIntegrationTest.java
```diff
@@ -0,0 +1,491 @@
+package com.skapp.community.peopleplanner.controller.v1;
```
ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

The Feb-29 fallback in `buildBirthdayOnDatePredicate` is the only non-trivial branch in the new query and it has zero coverage — the controller tests always derive the date from the current clock, so the fallback can only ever fire on one real calendar day per year. Because `findEmployeesWithBirthdayOn(date, currentEmployeeId, scope)` takes the date as a parameter, this is directly testable at the repository level: add a test that seeds an employee with `birthDate = 1996-02-29`, calls `employeeDao.findEmployeesWithBirthdayOn(LocalDate.of(2027, 2, 28), ...)` (non-leap year → expect the employee to be returned) and `LocalDate.of(2028, 2, 28)` (leap year → expect them *not* returned, since 2028-02-29 exists). Worth also asserting the Feb-28-born employee is still returned in both cases.

backend/src/main/java/com/skapp/community/peopleplanner/payload/response/BirthdayNotificationViewedResponseDto.java
```diff
@@ -0,0 +1,16 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import lombok.Setter;
+
+import java.time.LocalDate;
+
+@Getter
+@Setter
+@AllArgsConstructor
+public class BirthdayNotificationViewedResponseDto {
```
ThinuwanW, line 12: 🤖 **Claude** · 🟣 **Nit**

`BirthdayNotificationViewedResponseDto` holds only `lastViewedDate`, which is already the first field of `BirthdayNotificationResponseDto`. Two new DTOs where one is a strict prefix of the other is the duplication the guidelines call out. Either return `new BirthdayNotificationResponseDto(today, List.of())` from `markTodayBirthdayNotificationsAsViewed` (keeping the two endpoints' response shapes symmetric, which is nicer for the frontend) or drop the wrapper and return the date via the `ResponseEntityDto` acknowledgement.

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/EmployeeRepositoryImpl.java
```diff
@@ -1631,4 +1634,80 @@ public void updateLastClockInDate(Long employeeId, LocalDate lastClockInDate) {
...
+		Join<Employee, EmployeePersonalInfo> personalInfoJoin = root.join(Employee_.personalInfo);
+
+		List<Predicate> predicates = new ArrayList<>();
+		predicates.add(cb.isTrue(root.get(Employee_.user).get(User_.isActive)));
+		predicates.add(root.get(Employee_.accountStatus).in(Set.of(AccountStatus.ACTIVE, AccountStatus.PENDING)));
+
+		Join<Employee, EmployeeRole> roleJoin = root.join(Employee_.employeeRole, JoinType.LEFT);
+		predicates.add(PeopleUtil.notGuestEmployeePredicate(cb, roleJoin));
+
+		predicates.add(cb.isNotNull(personalInfoJoin.get(EmployeePersonalInfo_.birthDate)));
```
ThinuwanW, line 1658: 🤖 **Claude** · 🟣 **Nit**

`cb.isNotNull(personalInfoJoin.get(EmployeePersonalInfo_.birthDate))` is dead: the very next predicate compares `MONTH(birth_date)`/`DAY(birth_date)` for equality, and in SQL `MONTH(NULL) = 2` evaluates to NULL, so null birth dates are already excluded (the inner join on `personalInfo` likewise already drops employees with no personal-info row). Harmless but it reads as a required guard. Note the integration test's `givePersonalInfo(FIFTH_EMPLOYEE_ID, null)` at line 390 doesn't actually exercise it either, since employee 5 has no matching month/day regardless — so that line of the test is also a no-op.

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/EmployeeRepositoryImpl.java
```diff
@@ -1631,4 +1634,80 @@ public void updateLastClockInDate(Long employeeId, LocalDate lastClockInDate) {
...
+		predicates.add(buildBirthdayOnDatePredicate(cb, personalInfoJoin, date));
+		predicates.add(buildBirthdayScopePredicate(cb, criteriaQuery, root, currentEmployeeId, scope));
+
+		criteriaQuery.select(root)
+			.where(predicates.toArray(new Predicate[0]))
+			.distinct(true)
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`.distinct(true)` looks unnecessary here: every join is to-one (`personalInfo` 1:1, `employeeRole` 1:1, implicit `user` 1:1) and the team check is an `EXISTS` subquery rather than a join, so no row multiplication is possible. On MySQL this adds a `SELECT DISTINCT` over all `employee` columns (including long text columns) plus a sort/dedup step for no benefit. Also, `selfPredicate` on line 1688 is computed before the `scope == ORGANIZATION` early return and is unused on that path — move it below the ORGANIZATION check.

backend/src/test/java/com/skapp/community/peopleplanner/controller/v1/PeopleBirthdayNotificationControllerIntegrationTest.java
```diff
@@ -0,0 +1,491 @@
...
+
+	private static final Long OUTSIDER_EMPLOYEE_ID = 3L;
+
+	private static final Long SECOND_OUTSIDER_EMPLOYEE_ID = 4L;
+
+	private static final Long FIFTH_EMPLOYEE_ID = 5L;
+
+	private static final Long SEEDED_TEAM_ID = 1L;
+
+	private static final Long TOKEN_USER_ID = 1L;
```
ThinuwanW, line 92: 🤖 **Claude** · 🟣 **Nit**

`TOKEN_USER_ID = 1L` is passed to `generateAccessToken` for *both* `CURRENT_USER_EMAIL` (user1, id 1) and `RESTRICTED_USER_EMAIL` (user2, id 2), so the restricted token carries a userId claim that doesn't match its subject. It happens to pass today because `JwtAuthFilter`/`UserServiceImpl.getCurrentUser()` resolve the principal from the email claim and only `checkVersionMismatch` uses the id — but it's misleading, and the assertion at line 433 (`createdBy == TOKEN_USER_ID`) is only meaningful for user1. Other tests in the repo pass the matching id (e.g. `PolicyLeaveTypeControllerIntegrationTest` uses `2L` for `user2@gmail.com`). Suggest a `tokenFor(String email, Long userId)` overload, or deriving the id from the loaded user. Same issue in `PeopleConfigControllerIntegrationTest`.

backend/src/main/java/com/skapp/community/peopleplanner/controller/v1/PeopleConfigController.java
```diff
@@ -0,0 +1,45 @@
...
+@RequestMapping("/v1/people-config")
+@Tag(name = "People Config Controller", description = "Operations related to people module config functionalities")
+public class PeopleConfigController {
+
+	private final PeopleConfigService peopleConfigService;
+
+	@Operation(summary = "Get birthday notification configuration",
+			description = "Returns the organization-wide configuration controlling birthday notifications.")
+	@GetMapping(value = "/birthday-notifications")
+	@PreAuthorize("hasAnyRole('ROLE_PEOPLE_ADMIN')")
```
ThinuwanW, line 29: 🤖 **Claude** · 🟣 **Nit**

`hasAnyRole('ROLE_PEOPLE_ADMIN')` omits `ROLE_SUPER_ADMIN`, whereas most admin-level people endpoints use `hasAnyRole('ROLE_SUPER_ADMIN','ROLE_PEOPLE_ADMIN')` (both variants exist in `PeopleController`, so this is a consistency point rather than a hole). It is harmless today because `RolesServiceImpl` always grants `PEOPLE_ADMIN` alongside `isSuperAdmin`, but if that ever changes a super admin would get 403 on their own org's birthday config. Worth aligning both methods here — and the two new `ROLE_PEOPLE_EMPLOYEE`-only endpoints in `PeopleController` (lines 235/244) — with the dominant `'ROLE_SUPER_ADMIN', ...` pattern.

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/EmployeeRepositoryImpl.java
```diff
@@ -1631,4 +1634,80 @@ public void updateLastClockInDate(Long employeeId, LocalDate lastClockInDate) {
...
+			.orderBy(cb.asc(cb.lower(root.get(Employee_.firstName))), cb.asc(cb.lower(root.get(Employee_.lastName))));
+
+		return entityManager.createQuery(criteriaQuery).getResultList();
+	}
+
+	private Predicate buildBirthdayOnDatePredicate(CriteriaBuilder cb,
+			Join<Employee, EmployeePersonalInfo> personalInfoJoin, LocalDate date) {
+		Expression<Integer> month = cb.function(CriteriaBuilderSqlFunction.MONTH.getFunctionName(), Integer.class,
```
ThinuwanW, line 1672: 🤖 **Claude** · 🟡 **Suggestion**

Wrapping the column in `MONTH(birth_date)` / `DAY(birth_date)` makes the predicate non-sargable: no index on `employee_personal_info.birth_date` can ever be used, so every call scans the whole `employee` x `employee_personal_info` join. This is not a background job — `GET /v1/people/birthday-notifications/today` is called by every authenticated employee on (probably) every app load, so a 10k-employee tenant pays a full scan per user per load, and with ORGANIZATION scope the result set is also unbounded. Options, cheapest first: (a) cache the day's result per (date, scope) since it can only change once per day; (b) short-circuit before hitting the DB when scope is SELF by reading the current employee's own `birthDate` (the common case for a disabled/self-only config); (c) if this needs to scale, persist a generated/indexed `birth_month_day` (e.g. SMALLINT `MMDD`) column on `employee_personal_info` and compare against a computed literal, which keeps the Feb-29 OR-branch working and makes the lookup an index range scan.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
```diff
@@ -1472,6 +1486,75 @@ public ResponseEntityDto reassignSupervisorsAndTerminateOrDeleteEmployee(Long us
...
+			.employeeListToEmployeeBirthdayResponseDtoList(employeesWithBirthdays);
+		response.forEach(dto -> dto.setIsCurrentUser(dto.getEmployeeId().equals(currentEmployeeId)));
+
+		log.info("getTodayBirthdayNotifications: execution ended, {} birthday(s) found", response.size());
+		return new ResponseEntityDto(false, new BirthdayNotificationResponseDto(lastViewedDate, response));
+	}
+
+	@Override
+	@Transactional
+	public ResponseEntityDto markTodayBirthdayNotificationsAsViewed() {
```
ThinuwanW, line 1597: 🤖 **Claude** · 🟡 **Suggestion**

`markTodayBirthdayNotificationsAsViewed()` writes a `SpecialNotificationStatus` row unconditionally — it never reads the birthday config, and it never checks that anything was actually shown. Two consequences: (1) rows accumulate for a feature that is disabled, and (2) it is order-dependent in a user-visible way — if a client calls mark-viewed while the feature is off (or before the admin enables it earlier the same day), `getTodayBirthdayNotifications()` will then hit the `lastViewedDate.isEqual(today)` short-circuit and suppress the popup for the rest of that day even after the admin turns the feature on. Fix: read the config first (reuse the same `specialNotificationService.getConfig(...)` call `getTodayBirthdayNotifications` makes) and return early with the existing `lastViewedDate` when `isTurnedOn` is not TRUE, instead of persisting. Note `getTodayBirthdayNotifications` already has this exact guard, so the two endpoints currently disagree about whether the feature flag matters.

backend/src/main/java/com/skapp/community/common/type/OrganizationConfigType.java
```diff
@@ -2,6 +2,6 @@
 
 public enum OrganizationConfigType {
 
-	LEAVE_CYCLE, HOURS_PER_DAY, EMAIL_CONFIGS
+	LEAVE_CYCLE, HOURS_PER_DAY, EMAIL_CONFIGS, BIRTHDAY_NOTIFICATIONS
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`BIRTHDAY_NOTIFICATIONS` is the only `OrganizationConfigType` with no default row written at organization setup. `LEAVE_CYCLE` is seeded by `leaveCycleService.setLeaveCycleDefaultConfigs()` and `EMAIL_CONFIGS` by `saveEmailServerConfigs(...)`, both invoked from `OrganizationServiceImpl.setDefaultOrganizationConfigs()`; `HOURS_PER_DAY` is seeded too. Because nothing seeds this one, correctness rests entirely on `getConfig`'s `orElse("{}")` fallback plus the `= false` field initialisers in `BirthdayNotificationConfigResponseDto` — i.e. the same unguarded deserialisation path already flagged elsewhere. It also means the API cannot distinguish "never configured" from "explicitly disabled", so the admin UI has no way to show an unset state. Add a `setBirthdayNotificationDefaultConfig()` call to `setDefaultOrganizationConfigs()` that persists an explicit all-false config, following the `setLeaveCycleDefaultConfigs()` pattern.

backend/src/main/java/com/skapp/community/common/service/impl/SpecialNotificationServiceImpl.java
```diff
@@ -0,0 +1,96 @@
...
+			.findOrganizationConfigByOrganizationConfigType(type.getOrganizationConfigType().name())
+			.map(OrganizationConfig::getOrganizationConfigValue)
+			.orElse("{}");
+
+		return objectMapper.readValue(configValue, configClass);
+	}
+
+	@Override
+	@Transactional
+	public void saveConfig(SpecialNotificationType type, SpecialNotificationConfig config) {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`saveConfig` is a near-verbatim copy of the org-config upsert block already in `OrganizationServiceImpl.saveEmailServerConfigs` (lines ~140-146): same `findOrganizationConfigByOrganizationConfigType(...).orElseGet(() -> new OrganizationConfig(key, value))` then `setOrganizationConfigValue(value)` then `save(...)`. `getConfig` likewise mirrors `getEmailServiceConfigs`. Per the duplication guideline, extract a single typed org-config accessor (e.g. `OrganizationConfigService#read(String key, Class<T>)` / `#write(String key, Object value)`) in the `common` module and have both `OrganizationServiceImpl` and this class delegate to it. Separately, this class currently mixes two unrelated responsibilities — organization-wide config read/write and per-employee view state — which is why the DTO/marker-interface layering came out awkward; splitting the config half out would resolve both.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleConfigServiceImpl.java
```diff
@@ -0,0 +1,63 @@
...
+			config.setIsOrganizationWide(requestDto.getIsOrganizationWide());
+		}
+		if (requestDto.getIsTeamWide() != null) {
+			config.setIsTeamWide(requestDto.getIsTeamWide());
+		}
+
+		specialNotificationService.saveConfig(SpecialNotificationType.BIRTHDAY, config);
+
+		log.info("updateBirthdayNotificationConfigs: execution ended");
+		return new ResponseEntityDto(
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The PATCH returns only an acknowledgement message while the GET on the same resource returns the config object, so a client can never learn the resulting state from the write it just performed. `PeopleConfigControllerIntegrationTest` demonstrates the cost: every update test has to issue a follow-up `performGetRequest(...)` to see what was stored. Since the service already holds the merged `config` object at this point, return it: `return new ResponseEntityDto(false, config)` (or include both message and config). That also makes the partial-update semantics self-documenting to the caller, which matters given the flags are a tri-state encoded as independent booleans.

backend/src/main/java/com/skapp/community/common/payload/request/BirthdayNotificationConfigRequestDto.java
```diff
@@ -0,0 +1,16 @@
+package com.skapp.community.common.payload.request;
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Package placement is inconsistent for this feature. `BirthdayNotificationConfigRequestDto` and `BirthdayNotificationConfigResponseDto` sit in `com.skapp.community.common.payload.request`/`.response`, but every other piece of the feature that consumes them lives in `peopleplanner`: `PeopleConfigController`, `PeopleConfigService(Impl)`, `PeopleMessageConstant.PEOPLE_SUCCESS_BIRTHDAY_NOTIFICATION_CONFIG_UPDATED`, and `BirthdayNotificationScope` (which is in `peopleplanner.type` while `SpecialNotificationType` is in `common.type`). Nothing in `common` references these two DTOs. Move them to `com.skapp.community.peopleplanner.payload.request`/`.response` alongside the rest of the feature, and keep only the genuinely generic pieces (`SpecialNotificationConfig`, `SpecialNotificationType`, `SpecialNotificationStatus`) in `common`.

backend/src/main/java/com/skapp/community/peopleplanner/payload/response/EmployeeBirthdayResponseDto.java
```diff
@@ -0,0 +1,22 @@
+package com.skapp.community.peopleplanner.payload.response;
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Four of this DTO's six fields (`employeeId`, `firstName`, `lastName`, `authPic`) are identical to the existing `EmployeeBasicDetailsResponseDto`, which `PeopleMapper` already maps from `Employee` via `employeeToEmployeeBasicDetailsResponseDto`. Per the duplication guideline, either extend `EmployeeBasicDetailsResponseDto` and add only `email` + `isCurrentUser`, or compose it as a nested field. Also consider whether `email` is needed at all here — the birthday popup renders name plus avatar, and `authPic` + name already cover that; dropping `email` would shrink the payload every employee receives about every other employee.

backend/src/test/java/com/skapp/community/peopleplanner/controller/v1/PeopleBirthdayNotificationControllerIntegrationTest.java
```diff
@@ -0,0 +1,491 @@
...
+			assertThat(specialNotificationStatusDao.count()).isZero();
+		}
+
+	}
+
+	@Nested
+	@DisplayName("Organization Time Zone Tests")
+	class OrganizationTimeZoneTests {
+
+		private static final String FAR_AHEAD_TIME_ZONE = "Pacific/Kiritimati";
```
ThinuwanW, line 426: 🤖 **Claude** · 🟡 **Suggestion**

`OrganizationTimeZoneTests` computes its expectation with the same expression production uses (`LocalDate.now(ZoneId.of(FAR_AHEAD_TIME_ZONE))`), so it only detects a regression to UTC during the ~10 hours per day when the UTC date and the Kiritimati (UTC+14) date actually differ — for the other 14 hours it passes even if `resolveBirthdayNotificationDate()` ignored the org timezone entirely. Make it deterministic by asserting against a fixed injected clock, or at minimum assert the stored date is `utcToday` or `utcToday.plusDays(1)` *and* separately assert it equals the Kiritimati date, so the intent is explicit. Related: the whole class takes `today = DateTimeUtils.getCurrentUtcDate()` in `@BeforeEach` while production resolves the date from the organization timezone at request time — that only agrees because `data.sql` seeds no `organization` row and `getOrganizationTimeZone()` falls back to "UTC". Adding a seeded organization with a non-UTC timezone would silently break every test in this class; a comment or an explicit UTC-organization fixture would make that dependency visible.

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/EmployeeRepositoryImpl.java
```diff
@@ -1631,4 +1634,80 @@ public void updateLastClockInDate(Long employeeId, LocalDate lastClockInDate) {
...
+		predicates.add(PeopleUtil.notGuestEmployeePredicate(cb, roleJoin));
+
+		predicates.add(cb.isNotNull(personalInfoJoin.get(EmployeePersonalInfo_.birthDate)));
+		predicates.add(buildBirthdayOnDatePredicate(cb, personalInfoJoin, date));
+		predicates.add(buildBirthdayScopePredicate(cb, criteriaQuery, root, currentEmployeeId, scope));
+
+		criteriaQuery.select(root)
+			.where(predicates.toArray(new Predicate[0]))
+			.distinct(true)
+			.orderBy(cb.asc(cb.lower(root.get(Employee_.firstName))), cb.asc(cb.lower(root.get(Employee_.lastName))));
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

The `orderBy` sorts on `lower(firstName)` then `lower(lastName)` with no stable tie-breaker, so two employees sharing both names come back in whatever order the DB chooses, and that order can differ between calls. Append `cb.asc(root.get(Employee_.employeeId))` as a final sort key so the response is deterministic (the test relies on ordering, so a tie-breaker also makes the assertions robust if seed names ever change).

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/EmployeeRepositoryImpl.java
```diff
@@ -1631,4 +1634,80 @@ public void updateLastClockInDate(Long employeeId, LocalDate lastClockInDate) {
...
+					cb.equal(day, PeopleConstants.LEAP_DAY_OF_FEBRUARY));
+			return cb.or(onDate, onLeapDay);
+		}
+
+		return onDate;
+	}
+
+	private Predicate buildBirthdayScopePredicate(CriteriaBuilder cb, CriteriaQuery<Employee> criteriaQuery,
+			Root<Employee> root, Long currentEmployeeId, BirthdayNotificationScope scope) {
+		Predicate selfPredicate = cb.equal(root.get(Employee_.employeeId), currentEmployeeId);
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`selfPredicate` is built at the top of `buildBirthdayScopePredicate` but is unused on the ORGANIZATION path, which returns `cb.conjunction()` immediately after. Move the `selfPredicate` construction below the `if (scope == BirthdayNotificationScope.ORGANIZATION)` early return so it is only built where it is actually used.

backend/src/main/java/com/skapp/community/peopleplanner/constant/PeopleConstants.java
```diff
@@ -43,4 +43,10 @@ public class PeopleConstants {
 
 	public static final int MAX_EMPLOYEE_NUMBER_LENGTH = 10;
 
+	public static final int FEBRUARY_MONTH_VALUE = 2;
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

These three constants re-derive things `java.time` already provides, and their names are inconsistent with each other (`SHORT_FEBRUARY_LAST_DAY` vs `LEAP_DAY_OF_FEBRUARY`). `buildBirthdayOnDatePredicate`'s guard can be written without any of them: `date.getMonth() == Month.FEBRUARY && date.getDayOfMonth() == date.lengthOfMonth()` is exactly "last day of February", which is 28 only in non-leap years, so it subsumes the separate `!date.isLeapYear()` check too. If you keep the constants, at least rename for consistency (e.g. `FEBRUARY_SHORT_LAST_DAY` / `FEBRUARY_LEAP_DAY`) and note that `FEBRUARY_MONTH_VALUE` duplicates `Month.FEBRUARY.getValue()`.

backend/src/main/java/com/skapp/community/common/service/SpecialNotificationService.java
```diff
@@ -0,0 +1,19 @@
...
+
+public interface SpecialNotificationService {
+
+	<T extends SpecialNotificationConfig> T getConfig(SpecialNotificationType type, Class<T> configClass);
+
+	void saveConfig(SpecialNotificationType type, SpecialNotificationConfig config);
+
+	Optional<LocalDate> getLastViewedDate(Long employeeId, SpecialNotificationType type);
+
+	void markNotificationViewed(Long employeeId, SpecialNotificationType type, LocalDate viewedDate);
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`markNotificationViewed` breaks the established `markXAsY` naming used by the sibling notification API — `NotificationService.markNotificationAsRead`, `markAllNotificationsAsRead`, `markNotificationTypeAsRead`, and `AITokenService.markChatbotUsageAsNudged`. Rename to `markNotificationAsViewed` for consistency (the controller-facing `markTodayBirthdayNotificationsAsViewed` already follows the convention).

backend/src/main/java/com/skapp/community/peopleplanner/controller/v1/PeopleController.java
```diff
@@ -229,6 +229,24 @@ public ResponseEntity<ResponseEntityDto> getNotificationSettings() {
 		return new ResponseEntity<>(response, HttpStatus.OK);
 	}
 
+	@Operation(summary = "Get today's birthday notifications",
+			description = "Returns the birthdays the current employee should be notified about today, if not already viewed.")
+	@GetMapping(value = "/birthday-notifications/today")
+	@PreAuthorize("hasAnyRole('ROLE_PEOPLE_EMPLOYEE')")
```
ThinuwanW, line 255: 🤖 **Claude** · 🟣 **Nit**

Both new endpoints use `hasAnyRole('ROLE_PEOPLE_EMPLOYEE')`, while the sibling employee-facing endpoints in this same controller (lines 69 and 126) use `hasAnyRole('ROLE_SUPER_ADMIN','ROLE_PEOPLE_EMPLOYEE')`. In practice this is currently harmless — `RolesServiceImpl.saveSuperAdminRoles` always sets `peopleRole = PEOPLE_ADMIN`, and `AuthorityServiceImpl.addRoleHierarchy` grants `ROLE_PEOPLE_EMPLOYEE` to admins — so no real user is locked out today. Flagging only so that whoever normalises the role expressions on `PeopleConfigController` applies the same treatment here rather than leaving the two controllers inconsistent. Also, `hasAnyRole` with a single argument reads better as `hasRole('PEOPLE_EMPLOYEE')`.

backend/src/main/java/com/skapp/community/common/service/impl/SpecialNotificationServiceImpl.java
gaindurootcode: Fixed - the try/catch and saveAndFlush are gone, it's a plain save now, so a genuine conflict surfaces as a normal error instead of a silently rollback-only transaction.

backend/src/main/java/com/skapp/community/peopleplanner/controller/v1/PeopleConfigController.java
gaindurootcode: Added @Valid

backend/src/main/java/com/skapp/community/peopleplanner/payload/request/BirthdayNotificationConfigRequestDto.java
gaindurootcode: This is intentional. This reflects how the configs are presented, and is required to save the state properly

backend/src/main/java/com/skapp/community/common/service/impl/SpecialNotificationServiceImpl.java
gaindurootcode: Project standard is to not use try catch and let the global exception handler handle it

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
gaindurootcode: This is an invalid issue. There cannot be a null organisation time zone.

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/EmployeeRepositoryImpl.java
gaindurootcode: Used root.fetch instead of root.join to fix this issue. Fixed

backend/src/main/java/com/skapp/community/common/service/impl/SpecialNotificationServiceImpl.java
gaindurootcode: Since these are not directly controller called services only the db writing methods were given the logs

backend/src/main/java/com/skapp/community/common/payload/response/BirthdayNotificationConfigResponseDto.java
gaindurootcode: Renamed to BirthdayNotificationConfigDto and moved into peopleplanner.payload.response so the name matches all three roles; not splitting it, since a twin class with three identical fields plus a mapper is worse duplication than the naming issue

backend/src/test/java/com/skapp/community/peopleplanner/controller/v1/PeopleBirthdayNotificationControllerIntegrationTest.java
gaindurootcode: Can't be reached from a integration test

backend/src/main/java/com/skapp/community/peopleplanner/payload/response/BirthdayNotificationViewedResponseDto.java
gaindurootcode: Not fixed. the frontend syncs its cache from this lastViewedDate so it can't move to the acknowledgement, and reusing the other DTO would ship a permanently empty employeeBirthdays array.

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/EmployeeRepositoryImpl.java
gaindurootcode: Keeping it since it is safer and will not cause any issues.

backend/src/test/java/com/skapp/community/peopleplanner/controller/v1/PeopleBirthdayNotificationControllerIntegrationTest.java
gaindurootcode: Fixed. tokenFor(email, userId) now takes the id explicitly and user2@gmail.com gets 2L in both new test classes.

backend/src/main/java/com/skapp/community/peopleplanner/controller/v1/PeopleConfigController.java
gaindurootcode: Not fixed. Roles are hierarchical, so we deliberately declare only the least-privileged role that should have access and higher roles inherit it.

backend/src/main/java/com/skapp/community/common/type/OrganizationConfigType.java
gaindurootcode: Not needed. It will write the default value when they first set the config.

backend/src/main/java/com/skapp/community/common/service/impl/SpecialNotificationServiceImpl.java
gaindurootcode: This is expected. But fixing this would mean changing the email-config code that has nothing to do with this PR all while to stop 4 lines of code being duplicated. So not fixed

backend/src/main/java/com/skapp/community/common/payload/request/BirthdayNotificationConfigRequestDto.java
gaindurootcode: Moved both into peopleplanner.payload.

backend/src/main/java/com/skapp/community/peopleplanner/payload/response/EmployeeBirthdayResponseDto.java
gaindurootcode: reused existing employee basic info dto

backend/src/test/java/com/skapp/community/peopleplanner/controller/v1/PeopleBirthdayNotificationControllerIntegrationTest.java
gaindurootcode: Added a test

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/EmployeeRepositoryImpl.java
gaindurootcode: added

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/EmployeeRepositoryImpl.java
gaindurootcode: Fixed together with the distinct comment - selfPredicate is now built only on the paths that use it.

backend/src/main/java/com/skapp/community/peopleplanner/constant/PeopleConstants.java
gaindurootcode: All three constants deleted - the condition is now date.getMonth() == Month.FEBRUARY && date.getDayOfMonth() == date.lengthOfMonth() && !date.isLeapYear(), which reads as its own documentation (DateTimeUtils.FEBRUARY already existed, so it was a triple).

backend/src/main/java/com/skapp/community/common/service/SpecialNotificationService.java
gaindurootcode: Renamed to markNotificationAsViewed

backend/src/main/java/com/skapp/community/peopleplanner/controller/v1/PeopleController.java
gaindurootcode: Same as the other role comment - the hierarchy means declaring the least-privileged role is intentional.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleConfigServiceImpl.java
gaindurootcode: refactored return the entity

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/EmployeeRepositoryImpl.java
gaindurootcode: Fixed. distinct removed and selfPredicate moved below the ORGANIZATION early return.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
gaindurootcode: added isEnabled flag so this problem wont occure

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/EmployeeRepositoryImpl.java
gaindurootcode: Not fixed. Will be done in the future

backend/src/main/java/com/skapp/community/peopleplanner/controller/v1/PeopleConfigController.java
```diff
@@ -0,0 +1,46 @@
...
+		ResponseEntityDto response = peopleConfigService.getBirthdayNotificationConfigs();
+		return new ResponseEntity<>(response, HttpStatus.OK);
+	}
+
+	@Operation(summary = "Update birthday notification configuration",
+			description = "Partially updates the organization-wide configuration controlling birthday notifications.")
+	@PatchMapping(value = "/birthday-notifications")
+	@PreAuthorize("hasAnyRole('ROLE_PEOPLE_ADMIN')")
+	public ResponseEntity<ResponseEntityDto> updateBirthdayNotificationConfigs(
+			@Valid @RequestBody BirthdayNotificationConfigRequestDto requestDto) {
```
shakila-kamalasena: Removed @Valid

backend/src/main/resources/community/db/changelog/ddl-script/common-ddl-script-v1-create-table-com-special-notification-status.sql
```diff
@@ -0,0 +1,17 @@
+-- liquibase formatted sql
+
+-- changeset GainduNuhansith:00122_create_table_com_special_notification_status
+CREATE TABLE IF NOT EXISTS `com_special_notification_status`
+(
+    `employee_id`               bigint       NOT NULL,
+    `special_notification_type` varchar(255) NOT NULL,
+    `last_viewed_date`          date         NOT NULL,
+    `created_by`                text                  DEFAULT NULL,
```
shakila-kamalasena: Remove default null

shakila-kamalasena (changes requested):

backend/src/main/resources/community/db/changelog/ddl-script/common-ddl-script-v1-create-table-com-special-notification-status.sql
gaindurootcode: removed default null

backend/src/main/java/com/skapp/community/common/model/SpecialNotificationStatusId.java
```diff
@@ -0,0 +1,37 @@
...
+
+	private Long employee;
+
+	private SpecialNotificationType specialNotificationType;
+
+	@Override
+	public boolean equals(Object object) {
+		if (this == object)
+			return true;
+		if (!(object instanceof SpecialNotificationStatusId that))
```
akilarootcode: no need object and instance of

backend/src/main/java/com/skapp/community/common/model/SpecialNotificationStatusId.java
```diff
@@ -0,0 +1,37 @@
...
+			return false;
+		return Objects.equals(getEmployee(), that.getEmployee())
+				&& getSpecialNotificationType() == that.getSpecialNotificationType();
+	}
+
+	@Override
```
akilarootcode: no need hashcode

backend/src/main/java/com/skapp/community/common/model/SpecialNotificationStatusId.java
```diff
@@ -0,0 +1,37 @@
...
+import lombok.Setter;
+
+import java.io.Serializable;
+import java.util.Objects;
+
+@Getter
+@Setter
+@NoArgsConstructor
+@AllArgsConstructor
+public class SpecialNotificationStatusId implements Serializable {
```
akilarootcode, line 17: no need serializable

backend/src/main/java/com/skapp/community/common/service/impl/SpecialNotificationServiceImpl.java
```diff
@@ -0,0 +1,89 @@
...
+
+	private final JsonMapper objectMapper;
+
+	@Override
+	@Transactional(readOnly = true)
+	public SpecialNotificationConfig getSpecialNotificationConfig(SpecialNotificationType type) {
+		String configValue = organizationConfigDao
+			.findOrganizationConfigByOrganizationConfigType(type.getOrganizationConfigType().name())
+			.map(OrganizationConfig::getOrganizationConfigValue)
+			.orElse("{}");
```
akilarootcode: remove fallback. strimg value need to map to dto

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/EmployeeRepositoryImpl.java
```diff
@@ -1632,6 +1635,86 @@ public void updateLastClockInDate(Long employeeId, LocalDate lastClockInDate) {
...
+					cb.equal(day, Month.FEBRUARY.maxLength()));
+			return cb.or(onDate, onLeapDay);
+		}
+
+		return onDate;
+	}
+
+	private Predicate buildBirthdayScopePredicate(CriteriaBuilder cb, CriteriaQuery<Employee> criteriaQuery,
+												  Root<Employee> root, Long viewerEmployeeId, BirthdayNotificationScope scope) {
+		if (scope == BirthdayNotificationScope.ORGANIZATION) {
```
akilarootcode, line 1693: why need conjunction

backend/src/main/resources/community/db/changelog/ddl-script/common-ddl-script-v1-create-table-com-special-notification-status.sql
```diff
@@ -0,0 +1,23 @@
+-- liquibase formatted sql
+
+-- changeset GainduNuhansith:00122_create_table_com_special_notification_status
+CREATE TABLE IF NOT EXISTS `com_special_notification_status`
+(
+    `employee_id`               bigint       NOT NULL,
+    `special_notification_type` varchar(255) NOT NULL,
+    `last_viewed_date`          date         NOT NULL,
```
akilarootcode, line 8: remove default null

akilarootcode (changes requested):

backend/src/main/java/com/skapp/community/common/model/SpecialNotificationStatusId.java
gaindurootcode: we need serializable to implement the composite primary key

backend/src/main/java/com/skapp/community/common/model/SpecialNotificationStatusId.java
gaindurootcode: use lombok

backend/src/main/java/com/skapp/community/common/model/SpecialNotificationStatusId.java
gaindurootcode: used lombok

backend/src/main/resources/community/db/changelog/ddl-script/common-ddl-script-v1-create-table-com-special-notification-status.sql
gaindurootcode: removed

backend/src/main/java/com/skapp/community/common/service/impl/SpecialNotificationServiceImpl.java
gaindurootcode: removed

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/EmployeeRepositoryImpl.java
gaindurootcode: in here we are adding a filtering condition based on the scope. For ORGANIZATION scope there is no restriction at all - everybody's birthday is visible - so the helper needs to hand back something that's a valid Predicate but filters nothing. Returning null isn't an option: predicates.toArray(...) would pass a null into where(...) and Hibernate blows up. conjunction()  is the standard neutral element for that.

backend/src/main/java/com/skapp/community/common/payload/SpecialNotificationConfig.java
gaindurootcode: removed the generics entirely

backend/src/main/resources/community/db/changelog/ddl-script/common-ddl-script-v1-create-table-com-special-notification-status.sql
```diff
@@ -0,0 +1,34 @@
...
+    `created_by` text,
+    `created_date` datetime
```
akilarootcode, line 10: format this as well

akilarootcode (changes requested):

akilarootcode (approved):

PR #1686, opened 2026-07-30
Feat/2837 payroll id and tin

shakila-kamalasena:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds `payrollId` and `tin` to the employee record across backend (entity, Liquibase, validation, uniqueness endpoints, bulk upload) and frontend (form fields, Yup rules, bulk CSV, edit-diff). The core plumbing is coherent and consistent with existing patterns, but the uniqueness guarantee is not actually enforceable — the columns are unindexed `text` with no unique constraint, so the check-then-insert races in the 6-thread bulk executor and across concurrent requests. More seriously, the fields are presented as admin-only in the UI and gated behind an admin-only lookup endpoint, yet the write path (`PATCH /v1/people/employee/{id}`, open to `ROLE_PEOPLE_EMPLOYEE`) and the read path (managers and self) apply no such restriction. There are also two functional bugs: a cleared payroll ID/TIN can never be saved, and non-admin users get a guaranteed 403 on every employment-section save.

Found **30** new issue(s): 🔴 13 important, 🟡 12 suggestion(s), 🟣 5 nit(s)

backend/src/main/resources/community/db/changelog/ddl-script/people-ddl-script-v1-alter-table-employee-add-payroll-id-column.sql
```diff
@@ -0,0 +1,7 @@
+-- liquibase formatted sql
+
+-- changeset shakila:people-ddl-script-v1-alter-table-employee-add-payroll-id-column
+ALTER TABLE `employee`
+    ADD COLUMN `payroll_id` text;
```
ThinuwanW, line 5: 🤖 **Claude** · 🔴 **Important**

Both new columns are declared as `text` with no index and no unique constraint, but the whole feature is built around uniqueness. Two consequences: (1) `existsByPayrollId` / `existsByTin` run a full table scan on every create, update and bulk row — `addBulkEmployees` runs 6 threads and performs 2 of these scans per row, so a 1000-row upload is 2000 full scans of `employee`; (2) uniqueness is only enforced by a check-then-insert with no DB backstop, so two concurrent bulk tasks, two concurrent admins, or a bulk upload racing a single create will happily insert duplicates (the in-memory `seenPayrollIds`/`seenTins` sets only cover one request). MySQL also cannot put a plain UNIQUE index on a `TEXT` column without a prefix length, so this column type forecloses ever adding the constraint. Change both DDL scripts to `varchar(50)` (matching `PeopleConstants.MAX_PAYROLL_ID_LENGTH`/`MAX_TIN_LENGTH`) and add a unique index in the same changeset; correspondingly add `length = 50` to the `@Column` annotations on `Employee.payrollId`/`Employee.tin` (Employee.java:93-97), which currently have no length while sibling columns like `identification_no` do.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
```diff
@@ -429,6 +432,13 @@ private Employee createEmployeeEntity(Employee employee, CreateEmployeeRequestDt
 		CommonModuleUtils.setIfExists(
 				() -> requestDto.getEmployment().getIdentificationAndDiversityDetails().getEeoJobCategory(),
 				employee::setEeo);
+		CommonModuleUtils.setIfExists(
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Payroll ID and TIN are treated as admin-only everywhere they are surfaced (the UI tooltip says "Visible to Admins and Super Admins only", the field is `readOnly` unless `isPeopleAdmin`, and the uniqueness endpoints are `@PreAuthorize("hasAnyRole('ROLE_PEOPLE_ADMIN')")`), but the write path enforces nothing. `PATCH /v1/people/employee/{id}` is annotated `hasAnyRole('ROLE_SUPER_ADMIN','ROLE_PEOPLE_EMPLOYEE')`, `updateEmployee` performs no ownership or field-level check, and `createEmployeeEntity` applies `identificationAndDiversityDetails.payrollId`/`.tin` unconditionally. Any authenticated employee can therefore set or overwrite their own — or any other employee's — payroll ID and TIN with a hand-crafted request, and `EmployeeValidationServiceImpl` will happily validate it for them. Gate these two fields on the caller's role in `createEmployeeEntity` (or reject the sub-object in `validateCreateEmployeeRequestEmploymentDetails` when the current user lacks `PEOPLE_ADMIN`/`SUPER_ADMIN`), the same way system permissions are handled.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleReadServiceImpl.java
```diff
@@ -273,6 +273,8 @@ private EmployeeEmploymentIdentificationAndDiversityDetailsDto mapIdentification
 		});
 
 		dto.setEeoJobCategory(employee.getEeo());
+		dto.setPayrollId(employee.getPayrollId());
```
ThinuwanW, line 276: 🤖 **Claude** · 🔴 **Important**

`mapIdentificationAndDiversityDetails` is invoked whenever `accessLevel.canSeeSensitiveData()` is true, which per `EmployeeProfileViewAccessLevel` covers `FULL_ACCESS` (Super Admin, People Admin **and People Manager**) and `OWN_PROFILE` (any employee viewing themselves). So adding `payrollId`/`tin` here exposes them to every people manager for their reports and to every employee for their own record — directly contradicting the tooltip text shipped in this same PR ("Visible to Admins and Super Admins only"). Either restrict these two fields to `FULL_ACCESS`-minus-manager (e.g. a new `canSeeAdminOnlyIdentifiers()` on the enum checking for the admin roles specifically) or correct the tooltip copy so the UI does not promise a confidentiality guarantee the API does not provide.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
```diff
@@ -495,6 +505,10 @@ private Employee createEmployeeEntity(Employee employee, CreateEmployeeRequestDt
 		return employee;
 	}
 
+	private static String normalizeIdentifier(String value) {
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Clearing a payroll ID or TIN is impossible. `normalizeIdentifier` maps blank to `null`, and `CommonModuleUtils.setIfExists` (CommonModuleUtils.java:293) only invokes the setter when the supplied value is non-null — so the previous value is silently retained. This is reachable from the normal edit flow: `isFieldDifferentAndValid("", "OLD")` returns true, so the frontend puts `payrollId: ""` in the PATCH body, the backend drops it, and the UI then re-renders the old value after refetch with no error shown. Handle blank explicitly, e.g. `if (requestDto...getIdentificationAndDiversityDetails() != null && payrollId != null) employee.setPayrollId(payrollId.isBlank() ? null : payrollId.trim());` instead of routing it through `setIfExists`.

backend/src/main/java/com/skapp/community/peopleplanner/controller/v1/PeopleController.java
```diff
@@ -172,6 +174,21 @@ public ResponseEntity<ResponseEntityDto> getEmployeeByIdOrEmail(
 		return new ResponseEntity<>(response, HttpStatus.OK);
 	}
 
+	@GetMapping(value = "/check-payroll-id-uniqueness")
+	@PreAuthorize("hasAnyRole('ROLE_PEOPLE_ADMIN')")
```
ThinuwanW, line 180: 🤖 **Claude** · 🔴 **Important**

`hasAnyRole('ROLE_PEOPLE_ADMIN')` excludes super admins. `AuthorityServiceImpl.addSuperAdminAuthority` grants `ROLE_SUPER_ADMIN` independently of the people-module role hierarchy — it does not imply `ROLE_PEOPLE_ADMIN`. Every other admin-scoped endpoint on this controller (create, update, export, bulk, terminate, delete) uses `hasAnyRole('ROLE_SUPER_ADMIN','ROLE_PEOPLE_ADMIN')`. As written, a super admin who is not also a people admin gets a 403 from both new endpoints while the create/update endpoints they can still reach do run the same duplicate check server-side, so they get an opaque save failure instead of inline field feedback. Add `'ROLE_SUPER_ADMIN'`. The same gap exists on the frontend: `IdentificationDetailsSection` uses only `isPeopleAdmin` for `readOnly`, so a super admin cannot edit fields the tooltip explicitly tells them they own — use `isPeopleAdmin || isSuperAdmin` there.

frontend/src/community/people/components/organisms/EmploymentFormSection/EmploymentDetailsForm.tsx
```diff
@@ -135,6 +153,44 @@ const EmploymentDetailsForm = ({
       Object.keys(identificationFormErrors).length === 0;
 
     if (employmentFormIsValid && identificationFormIsValid) {
+      let hasIdentificationUniquenessError = false;
+
+      if (payrollId) {
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The uniqueness refetches fire for every user who saves the employment section, not just admins. `PeopleFormSections` computes `isReadOnly = !isPeopleAdmin && userId !== employeeId`, so a regular employee or people manager editing their own profile renders this form with `isReadOnly === false`, and `payrollId`/`tin` are populated from the store with their existing values — so both `if (payrollId)` and `if (tin)` are truthy and two requests go out to endpoints that require `ROLE_PEOPLE_ADMIN`. Result: two guaranteed 403s on every self-save. The failures are also silently swallowed — `refetch()` resolves with `data: undefined`, which reads as "no duplicate", so a genuine network/permission failure is indistinguishable from a clean check. Guard the calls with `isPeopleAdmin` (from `useSessionData`) **and** with whether the value actually changed versus `initialEmployee`, and surface a toast if the refetch returns an `error` rather than treating it as a pass. The two awaits can also be a single `Promise.all` instead of running serially.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
```diff
@@ -495,6 +505,10 @@ private Employee createEmployeeEntity(Employee employee, CreateEmployeeRequestDt
...
+		return (value == null || value.isBlank()) ? null : value;
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Nothing trims these identifiers, so whitespace defeats the uniqueness check. Yup's `.trim()` in `peopleValidations.ts` is a transform applied during validation only — it does not mutate `formik.values` or the Zustand store, so the submitted payload keeps leading/trailing spaces. `normalizeIdentifier` only tests `isBlank()` and returns the value untouched, and the bulk path (`peopleMapper.employeeBulkDtoToEmployee`) does not normalize at all. Consequence: `"P-100"` and `" P-100"` are stored as two distinct rows and `existsByPayrollId` will not flag them. There is also a length mismatch — `Validations.validatePayrollId` checks the raw length while Yup checks the trimmed length, so a 50-character value with a trailing space passes the client and is rejected by the server. Trim in `normalizeIdentifier` (`value.trim()`) and apply the same normalization in the bulk validator/mapper path.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
```diff
@@ -1351,6 +1368,45 @@ public ResponseEntityDto getEmployeeByIdOrEmail(EmployeeDataValidationDto employ
 		return new ResponseEntityDto(false, employeeDataValidationResponseDto);
 	}
 
+	@Override
+	public ResponseEntityDto checkPayrollIdUniqueness(PayrollIdUniquenessCheckDto payrollIdUniquenessCheckDto) {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`checkPayrollIdUniqueness` and `checkTinUniqueness` are the only public methods in this service without `@Transactional` and without entry/exit logging. Every neighbour — `getEmployeeByIdOrEmail` immediately above, `searchEmployeesByEmail`, `getLoginPendingEmployeeCount` — has `@Transactional` plus `log.info("methodName: execution started")`. Add `@Transactional(readOnly = true)` (these are pure reads, so `readOnly` is the correct variant per the guidelines) and the standard entry/exit log lines.

backend/src/main/java/com/skapp/community/peopleplanner/controller/v1/PeopleController.java
```diff
@@ -172,6 +174,21 @@ public ResponseEntity<ResponseEntityDto> getEmployeeByIdOrEmail(
...
+	public ResponseEntity<ResponseEntityDto> checkPayrollIdUniqueness(
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The two new endpoints are missing three things every neighbouring endpoint has: `@Operation(summary=..., description=...)` for the OpenAPI docs, `produces = MediaType.APPLICATION_JSON_VALUE`, and `@Valid` on the bound DTO. `@Valid` is currently moot because `PayrollIdUniquenessCheckDto`/`TinUniquenessCheckDto` carry no constraints at all — add `@NotBlank` and `@Size(max = PeopleConstants.MAX_PAYROLL_ID_LENGTH)` on the string fields so an oversized or empty query param is rejected at the binding layer instead of silently returning a response DTO with a null flag. Compare `getEmployeeByIdOrEmail` directly above, which does use `@Valid EmployeeDataValidationDto`.

backend/src/main/java/com/skapp/community/peopleplanner/payload/request/PayrollIdExistsCheckDto.java
```diff
@@ -0,0 +1,14 @@
+package com.skapp.community.peopleplanner.payload.request;
```
ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

This feature ships four new DTOs, two service methods, two DAO method pairs, two controller endpoints and two bulk validators that are byte-for-byte identical apart from the field name — exactly the duplication the guidelines ask to consolidate. The codebase already has the right pattern for this: `check-email-identification-no` takes one `EmployeeDataValidationDto` and returns one `EmployeeDataValidationResponseDto` carrying several independent existence booleans (`isWorkEmailExists`, `isIdentificationNoExists`, `isGuestUser`, ...). Extending that request/response pair with `payrollId`/`isPayrollIdExists` and `tin`/`isTinExists` would drop four classes, one endpoint, one round-trip from the save path, and would let the frontend reuse the existing `useCheckEmailAndIdentificationNo` hook and `EmployeeDataExists` type (which this PR already widens with both flags).

backend/src/main/java/com/skapp/community/peopleplanner/payload/response/PayrollIdExistsResponseDto.java
```diff
@@ -0,0 +1,14 @@
...
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+
+@Getter
+@Setter
+@NoArgsConstructor
+public class PayrollIdUniquenessResponseDto {
+
+	private Boolean isPayrollIdExists;
```
ThinuwanW, line 12: 🤖 **Claude** · 🟡 **Suggestion**

`isPayrollIdExists` is a boxed `Boolean` that is only assigned inside the `payrollId != null && !payrollId.isBlank()` branch, so a request with a missing or blank param returns `{"isPayrollIdExists": null}`. Callers then have to distinguish null from false. Initialise the field to `false` (`private Boolean isPayrollIdExists = false;`) or use a primitive `boolean`, and do the same in `TinUniquenessResponseDto`.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
```diff
@@ -1892,6 +1948,44 @@ public void validateSocialSecurityNumber(String socialSecurityNumber, List<Strin
 					new Object[] { PeopleConstants.MAX_SSN_LENGTH, "First Name" }));
 	}
 
+	public void validatePayrollIdInBulk(String payrollId, List<String> errors, Set<String> seenPayrollIds) {
+		if (payrollId == null || payrollId.isBlank()) {
+			return;
+		}
+
```
ThinuwanW, line 1965: 🤖 **Claude** · 🟡 **Suggestion**

Two issues in `validatePayrollIdInBulk` (and identically in `validateTinInBulk`). First, after appending the max-length error it falls through to the DB lookup and the seen-set check, so a single oversized value can produce three stacked error strings for one cell — return immediately after the length error, as the value is invalid regardless. Second, `seenPayrollIds.add(payrollId)` executes during validation, before the row is known to be persistable: if row 3 fails for an unrelated reason (bad email, missing passport) its payroll ID is still recorded, so a later row reusing that ID is rejected with "Payroll ID already used in this file" even though no employee was created with it. Record the ID only after `createNewEmployeeFromBulk` succeeds, or accept the tradeoff and document it.

frontend/src/community/people/components/organisms/EmploymentFormSection/SubSections/IdentificationDetailsSection.tsx
```diff
@@ -56,6 +59,22 @@ const IdentificationDetailsSection = forwardRef<FormMethods, Props>(
     const { values, errors, handleChange, setFieldError, setFieldValue } =
       formik;
 
+    useImperativeHandle(ref, () => ({
```
ThinuwanW, line 96: 🤖 **Claude** · 🟡 **Suggestion**

This `useImperativeHandle` is new — the component was previously `forwardRef<FormMethods, Props>` with the ref never wired up, so `identificationDetailsRef.current` was always `null` in `EmploymentDetailsForm`. That means `identificationFormErrors` was always `{}` (`undefined || {}`), `identificationFormIsValid` was always `true`, and `submitForm()`/`resetForm()` on this section were no-ops. Wiring the handle is almost certainly the right fix, but it is a behavioural change well beyond adding two fields: the pre-existing `ssn` max-20 Yup rule now actually gates saving, so any employee whose stored SSN exceeds 20 characters will be unable to save the employment section until the value is corrected, and Cancel now resets this sub-form where it previously did not. Please confirm this is intended and worth calling out in the release notes.

frontend/src/community/people/components/organisms/EmploymentFormSection/SubSections/IdentificationDetailsSection.tsx
```diff
@@ -165,6 +184,58 @@ const IdentificationDetailsSection = forwardRef<FormMethods, Props>(
                 ariaLabel={translateAria(["selectEEOJobCategory"])}
               />
             </Grid>
+
```
ThinuwanW, line 221: 🤖 **Claude** · 🟡 **Suggestion**

The parallel legacy components were not updated: `src/community/people/components/organisms/AddNewResourceFlow/EmploymentDetailsSection/IdentificationDetailsSection.tsx` has no payroll ID / TIN inputs, and it is still reachable from three live routes — `/people/directory/add-new-resource` (pages/community/people/directory/add-new-resource.tsx), `/user-account` (pages/community/user-account.tsx) and `/people/directory/edit-all-information/[id]`. Users going through those screens cannot see or set the new fields, so the feature is only half-present depending on which entry point is used. Either mirror the fields into that component (note it reads a different store slice, `employeeIdentificationAndDiversityDetails`) or, if those screens are dead code superseded by `PeopleFormSections`, remove the routes as part of this change.

frontend/src/community/people/utils/constants/constants.ts
```diff
@@ -46,7 +46,9 @@ export const USER_BULK_HEADERS = [
   "ssn",
   "ethnicity",
   "eeo",
-  "passportNo"
+  "passportNo",
```
ThinuwanW, line 49: 🤖 **Claude** · 🟡 **Suggestion**

`useUserBulkValidation.isCsvValid` requires `headerList.length === USER_BULK_HEADERS.length` and that every entry in `USER_BULK_HEADERS` is present. Appending `payrollId` and `tin` therefore invalidates every CSV produced from a previously downloaded template — those uploads will now fail with the generic invalid-CSV error and no indication that the template changed. The bundled `.xlsx` is updated in this PR, but users keep local copies. Consider relaxing `isCsvValid` to treat the two new columns as optional (they are optional everywhere else: `payrollId?: string | null` in `BulkUploadUser`, no-op in `validatePayrollIdInBulk` when blank), or surfacing a "please download the latest template" message on header-count mismatch.

frontend/src/community/people/components/organisms/EmploymentFormSection/SubSections/IdentificationDetailsSection.tsx
```diff
@@ -165,6 +184,58 @@ const IdentificationDetailsSection = forwardRef<FormMethods, Props>(
...
+              <Grid size={{ xs: 12, md: 6, xl: 4 }}>
+                <InputField
+                  label={translateText(["payrollId"])}
+                  inputType="text"
+                  value={values?.payrollId ?? ""}
+                  placeHolder={translateText(["enterPayrollId"])}
+                  onChange={handleInput}
+                  inputName="payrollId"
+                  error={errors.payrollId ?? ""}
+                  maxLength={50}
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`maxLength={50}` is hardcoded on both new inputs while the Yup schema uses `characterLengths.CHARACTER_LENGTH` and the backend uses `PeopleConstants.MAX_PAYROLL_ID_LENGTH`/`MAX_TIN_LENGTH` — three places to keep in sync, and the translated error strings (`payrollIdMaxLengthError`, `tinMaxLengthError`) hardcode "50" a fourth time. `CHARACTER_LENGTH` is also a generic catch-all rather than a field-specific constant. Add `PAYROLL_ID_LENGTH = 50` / `TIN_LENGTH = 50` to `characterLengths` and reference them from both the `maxLength` prop and the Yup rule.

frontend/src/community/people/types/UserBulkUploadTypes.ts
```diff
@@ -60,6 +60,8 @@ export interface BulkUploadUser {
   contractState: string | null;
   employmentStatus: string | null;
   passportNo: string | null;
+  payrollId?: string | null;
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`payrollId?: string | null` and `tin?: string | null` are declared optional, while every sibling CSV column in `BulkUploadUser` (including the equally-optional `passportNo`, `middleName`, `linkedIn`) is a required `string | null`. Since `isCsvValid` enforces that both headers are present in the file, the parsed object will always carry the keys — drop the `?` for consistency with the rest of the interface. Same applies to the two additions in `BulkEmployeeDetails` (EmployeeTypes.tsx:178).

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
```diff
@@ -58,8 +58,10 @@
 import com.skapp.community.peopleplanner.payload.request.EmployeeProgressionsDto;
 import com.skapp.community.peopleplanner.payload.request.EmployeeQuickAddDto;
 import com.skapp.community.peopleplanner.payload.request.NotificationSettingsPatchRequestDto;
+import com.skapp.community.peopleplanner.payload.request.PayrollIdUniquenessCheckDto;
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

No tests accompany this change on either side. The backend adds two service methods, two controller endpoints, four DAO derived queries and two bulk validators with non-trivial branching (self-exclusion via `*AndEmployeeIdNot`, blank short-circuit, in-file duplicate detection across a synchronized set shared by 6 threads) with no unit or `@SpringBootTest`/`@AutoConfigureMockMvc` coverage; the frontend adds two React Query hooks, a new imperative handle and new Yup rules with no `.test.tsx`. At minimum, cover the self-exclusion path in the uniqueness check (an admin re-saving an unchanged payroll ID must not be flagged as a duplicate) and the in-file duplicate path in bulk upload — both are easy to regress.

backend/src/main/java/com/skapp/community/peopleplanner/model/Employee.java
```diff
@@ -90,6 +90,12 @@ public class Employee extends Auditable<String> {
 	@Column(name = "identification_no", length = 15)
 	private String identificationNo;
 
+	@Column(name = "payroll_id")
+	private String payrollId;
+
+	@Column(name = "tin")
```
ThinuwanW, line 96: 🤖 **Claude** · 🔴 **Important**

`tin` (a government tax identifier) is persisted in plaintext, while every comparable identifier in the codebase is encrypted at rest via `FieldEncryptionConverter`: `Employee.personalEmail` (Employee.java:84) and `EmployeePersonalInfo.ssn`/`nin`/`passportNo` (EmployeePersonalInfo.java:56/67/71). TIN sits in the *same* UI section as SSN and is classified the same way in the tooltip copy, so the inconsistency looks unintentional rather than deliberate.

This is not a one-line fix, because the feature's uniqueness design blocks the obvious remedy. `FieldEncryptionConverter`'s own javadoc states it must only be applied "after verifying that field is not used in WHERE / ORDER BY / unique constraints / joins, since random-IV GCM ciphertext is not searchable" — and `existsByTin` / `existsByTinAndEmployeeIdNot` are exactly equality WHERE predicates. Adding `@Convert` today would silently break every uniqueness check in enterprise builds (community's service is a plaintext passthrough, so it would pass CI and fail only in enterprise).

Please make this an explicit decision before merge, one of:
1. Store the encrypted value in `tin` and add a separate deterministic blind-index column (e.g. `tin_hash` = HMAC-SHA256(tenant key, normalized tin)), with a unique index on the hash; point `existsByTin*` at the hash column instead of the raw value.
2. Keep TIN plaintext and record the rationale in a code comment next to the field plus the security notes, so the divergence from ssn/nin/passportNo is a documented choice rather than an oversight.

Whichever way, decide it now — retrofitting encryption after production rows exist requires a backfill migration, and the accompanying unique index (already raised in the first pass) has to be built on whichever column ends up being the comparison key.

backend/src/main/java/com/skapp/community/peopleplanner/repository/EmployeeDao.java
```diff
@@ -43,4 +43,12 @@ public interface EmployeeDao
 	List<Employee> findByWorkLocationWorkLocationIdAndAccountStatusIn(Long workLocationId,
 			Set<AccountStatus> accountStatuses);
 
+	boolean existsByPayrollId(String payrollId);
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

All four new derived queries (`existsByPayrollId`, `existsByPayrollIdAndEmployeeIdNot`, `existsByTin`, `existsByTinAndEmployeeIdNot`) match against every row in `employee` regardless of `accountStatus`. Employee removal in this codebase is a soft delete: `deleteUser` calls `updateUserStatus(userId, AccountStatus.DELETED, true)` (PeopleServiceImpl.java:1446), which keeps the row and only frees the *email* by rewriting it to `DELETED_<timestamp>_<email>` (PeopleConstants.DELETED_PREFIX). Termination likewise leaves the row with `TERMINATED`.

Failure scenario: an admin creates employee A with payroll ID `EMP-0042`, then deletes A. Six months later a new hire is assigned the same payroll ID by the payroll system. Every path now rejects it — the pre-check endpoint returns `isPayrollIdExists: true`, `EmployeeValidationServiceImpl.validatePayrollId` throws `PEOPLE_ERROR_PAYROLL_ID_ALREADY_EXIST` on save, and the bulk upload row fails with "This Payroll ID is already assigned to another employee". The ID is unusable forever with no UI anywhere that reveals which (deleted) record holds it, and no way for an admin to release it. Same for TIN, which is worse because a TIN legitimately follows a person across rehires.

Fix: decide the intended lifetime of these identifiers and encode it. Either (a) scope the checks to live accounts, e.g. `existsByPayrollIdAndAccountStatusIn(String payrollId, Set<AccountStatus> statuses)` called with the active/pending set (mirroring the existing `existsByEmployeeIdAndAccountStatusIn` / `findByAccountStatusIn` pattern already in this DAO), or (b) clear `payrollId`/`tin` in `updateUserStatus` when `isDelete` is true, right beside the existing email-prefixing branch, so deletion releases them the same way it releases the email. Option (a) also needs the eventual unique index to be a partial/filtered index or dropped in favour of application-level checks, so pick it deliberately.

backend/src/main/java/com/skapp/community/peopleplanner/util/Validations.java
```diff
@@ -384,6 +384,18 @@ public static void validateEmployeeNumber(String employeeNumber) {
 					List.of(String.valueOf(PeopleConstants.MAX_EMPLOYEE_NUMBER_LENGTH)));
 	}
 
+	public static void validatePayrollId(String payrollId) {
```
ThinuwanW, line 387: 🤖 **Claude** · 🔴 **Important**

`validatePayrollId` and `validateTin` only enforce a maximum length — no character-set validation at all. Every sibling identifier in this codebase validates its shape as well: `validateEmployeeNumber` (Validations.java:378) requires `ALPHANUMERIC_REGEX`, `validateNIN` requires `VALID_NIN_NUMBER_REGEXP`, and `validatePassportNumber` / `validateSocialSecurityNumber` in PeopleServiceImpl both require `ALPHANUMERIC_REGEX`. The frontend does not compensate: `employeeIdentificationDetailsValidation` (peopleValidations.ts:321) is `.trim().max(...)` only, and the bulk validator `useUserBulkValidation.validateUser` (useUserBulkValidation.ts:71) is a bare length check, unlike its neighbour for `identificationNo` which calls `hasOnlyAlphaNumericWithSpecialCharacters()`.

So a payroll ID or TIN of ``, a newline-containing value, or a right-to-left-override / zero-width-joiner string is accepted and persisted. Two concrete consequences: (1) CSV/report export of these values becomes a formula- and delimiter-injection vector; (2) unicode look-alikes (Cyrillic `А` vs Latin `A`) defeat the whole uniqueness feature while rendering identically to the admin reviewing the directory.

Fix: add a pattern check to both methods in the same style as `validateEmployeeNumber`, e.g. `if (!payrollId.trim().matches(ALPHANUMERIC_REGEX)) throw new ValidationException(PeopleMessageConstant.PEOPLE_ERROR_VALIDATION_PAYROLL_ID_FORMAT, ...)`, with matching message constants and `people-messages.properties` entries. `ALPHANUMERIC_REGEX` is already statically imported in this file. Mirror the same rule in the Yup schema and in `validatePayrollIdInBulk`/`validateTinInBulk` so all three entry points agree — if hyphens/slashes are legitimately needed for real payroll IDs, define a dedicated regex rather than leaving the field unconstrained.

backend/src/main/java/com/skapp/community/peopleplanner/payload/request/PayrollIdUniquenessCheckDto.java
```diff
@@ -0,0 +1,14 @@
+package com.skapp.community.peopleplanner.payload.request;
+
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+public class PayrollIdUniquenessCheckDto {
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Neither `PayrollIdUniquenessCheckDto` nor `TinUniquenessCheckDto` carries a single bean-validation annotation, and `PeopleController.checkPayrollIdUniqueness` / `checkTinUniqueness` (PeopleController.java:180, 187) bind them without `@Valid`. Compare the immediately preceding endpoints: `searchEmployeesByEmail` and `getEmployeeByIdOrEmail` both use `@Valid`, and `EmployeeDataValidationDto` annotates both of its String fields with `@NotNull`.

Two consequences beyond the missing-annotation style issue:
1. No length bound. The pre-check runs `existsByPayrollId` on a value of any size, so the UI happily reports "unique" for a 500-character payroll ID that `Validations.validatePayrollId` will then reject with a length error on save. The pre-check and the save path disagree about what is acceptable, which is exactly the asymmetry the pre-check exists to prevent.
2. A completely absent `payrollId` parameter is not rejected; it falls into the silent no-op branch (`if (payrollId != null && !payrollId.isBlank())`) and returns a response whose only field is null — the boxed-Boolean problem already raised in the first pass, but reachable here specifically because nothing enforces presence.

Fix: annotate the field on each DTO with `@NotBlank` and `@Size(max = PeopleConstants.MAX_PAYROLL_ID_LENGTH)` (respectively `MAX_TIN_LENGTH`), and add `@Valid` to both controller parameters so `GlobalExceptionHandler` turns a bad request into a 400 with the standard `ErrorResponse` instead of a 200 carrying a null flag.

frontend/src/community/people/components/organisms/EmploymentFormSection/EmploymentDetailsForm.tsx
```diff
@@ -135,6 +153,44 @@ const EmploymentDetailsForm = ({
...
+        const { data: payrollIdCheckResult } =
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The uniqueness pre-check fails open. `const { data: payrollIdCheckResult } = await refetchPayrollIdUniqueness();` destructures only `data`; the `error` / `isError` fields of the `refetch()` result are discarded. When the request fails for any reason — 403 (the endpoint is gated to `ROLE_PEOPLE_ADMIN`, which excludes super admins), a network drop, a 500 — `data` is `undefined`, `payrollIdCheckResult?.isPayrollIdExists` evaluates falsy, `hasIdentificationUniquenessError` stays `false`, and the form submits as though the value had been confirmed unique. The identical pattern applies to the TIN block at line 175.

Nothing surfaces the failure either: `useCheckPayrollIdUniqueness` / `useCheckTinUniqueness` (PeopleApi.ts:302-334) declare no `onError`, and `axiosInterceptor.ts` has no global error toast, so the user sees no indication the check was skipped. The save is ultimately still rejected server-side by `EmployeeValidationServiceImpl.validatePayrollId`, so this is not a data-integrity hole — but the user loses the inline field-level error this whole code path was built to produce and instead gets a generic mutation-failure toast pointing at no field.

Fix: treat a failed check as blocking rather than passing. Capture the full result and branch on it, e.g. `const result = await refetchPayrollIdUniqueness(); if (result.isError) { setToastMessage({ type: ToastType.ERROR, ... }); return; }` before reading `result.data`, and do the same for TIN.

While here: the two checks are awaited sequentially, adding two serial round-trips to every employment-section save. Run them concurrently with `Promise.all`, and skip each one entirely when the value is unchanged from `initialEmployee?.employment?.identificationAndDiversityDetails` — an unchanged payroll ID cannot have become non-unique, and this also narrows the over-firing already noted in the first pass.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
```diff
@@ -1892,6 +1948,44 @@ public void validateSocialSecurityNumber(String socialSecurityNumber, List<Strin
...
+		if (payrollId.length() > PeopleConstants.MAX_PAYROLL_ID_LENGTH) {
+			errors.add(messageUtil.getMessage(PeopleMessageConstant.PEOPLE_ERROR_EXCEEDING_MAX_CHARACTER_LIMIT,
+					new Object[] { PeopleConstants.MAX_PAYROLL_ID_LENGTH, "Payroll ID" }));
+		}
+
+		if (employeeDao.existsByPayrollId(payrollId)) {
+			errors.add(messageUtil.getMessage(PeopleMessageConstant.PEOPLE_ERROR_PAYROLL_ID_ALREADY_EXIST));
+		}
+
+		if (!seenPayrollIds.add(payrollId)) {
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The in-file duplicate detection and the database duplicate detection disagree about case, so bulk upload can both miss real duplicates and report the wrong error for them.

`seenPayrollIds` is a plain `HashSet<String>` (line 1275), so `seenPayrollIds.add(payrollId)` is case- and accent-*sensitive*. The DB check on the next-but-one line, `employeeDao.existsByPayrollId(payrollId)`, resolves under the column's collation — the new columns are created as bare `text` (people-ddl-script-v1-alter-table-employee-add-payroll-id-column.sql:5) and therefore inherit the schema default, which on MySQL 8 is `utf8mb4_0900_ai_ci`: case- and accent-*insensitive*. `validateTinInBulk` has the same split.

Failure scenario: a CSV contains row 1 with payroll ID `P100` and row 12 with `p100`. `seenPayrollIds` treats them as two distinct strings, so neither gets `PEOPLE_ERROR_PAYROLL_ID_ALREADY_IN_FILE`. The six worker threads process rows concurrently (`Executors.newFixedThreadPool(6)`), so whether the DB check catches it is a race: if row 1 has already committed, row 12 fails with "already assigned to another employee" — misleading, since the collision is inside the file the admin just uploaded — and if the two transactions overlap, both rows commit and the tenant ends up with two employees the DB considers to have the same payroll ID. That second state is unrecoverable through the UI: editing either employee without touching the payroll ID triggers `existsByPayrollIdAndEmployeeIdNot`, which matches the other row case-insensitively and blocks the save.

Fix: normalize once, then compare consistently everywhere. Introduce a single helper (e.g. `payrollId.trim().toUpperCase(Locale.ROOT)`) and use its output for the `seenPayrollIds`/`seenTins` membership test, for the `existsBy*` lookup, and for what is persisted — `createNewEmployeeFromBulk` already does exactly this for `identificationNo` at line 2225-2226. Note that `employeeBulkDtoToEmployee` maps `payrollId`/`tin` implicitly by name, so the bulk path currently bypasses the `normalizeIdentifier` helper the create/update path uses; normalizing on the DTO before mapping fixes both problems at once. Whichever normal form you choose must also match the one used by any unique index added later.

frontend/src/community/people/components/organisms/EmploymentFormSection/SubSections/IdentificationDetailsSection.tsx
```diff
@@ -165,6 +184,58 @@ const IdentificationDetailsSection = forwardRef<FormMethods, Props>(
...
+                  componentStyle={{
+                    flex: 1,
+                    mt: "0rem"
+                  }}
+                  readOnly={!isPeopleAdmin || isInputsDisabled}
```
ThinuwanW, line 237: 🤖 **Claude** · 🟡 **Suggestion**

`readOnly={!isPeopleAdmin || isInputsDisabled}` (and line 229 for TIN) locks these fields for super admins. `useSessionData.isPeopleAdmin` is `user?.roles?.includes(AdminTypes.PEOPLE_ADMIN)` (useSessionData.ts:74) — it does not consider `AdminTypes.SUPER_ADMIN`, which the hook exposes separately as `isSuperAdmin`.

A super admin without an explicit PEOPLE_ADMIN role therefore sees both fields rendered but non-editable, with the tooltip "Only Admins can edit this field" — directly contradicting the other tooltip this PR adds two lines up, "Visible to Admins and Super Admins only" (peopleModule.json:849-850). This is the client-side counterpart of the role gap on the endpoint; fixing only the `@PreAuthorize` would still leave the field greyed out in the UI.

Fix: pull `isSuperAdmin` from the same `useSessionData()` call and gate on `const canEditIdentifiers = isPeopleAdmin || isSuperAdmin;`, using it for both `readOnly` and the tooltip ternary on both inputs.

backend/src/main/java/com/skapp/community/peopleplanner/controller/v1/PeopleController.java
```diff
@@ -172,6 +174,21 @@ public ResponseEntity<ResponseEntityDto> getEmployeeByIdOrEmail(
...
+			PayrollIdUniquenessCheckDto payrollIdUniquenessCheckDto) {
+		ResponseEntityDto response = peopleService.checkPayrollIdUniqueness(payrollIdUniquenessCheckDto);
+		return new ResponseEntity<>(response, HttpStatus.OK);
+	}
+
+	@GetMapping(value = "/check-tin-uniqueness")
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Both new endpoints are `GET`, so the payroll ID and the TIN travel in the URL query string (`/v1/people/check-tin-uniqueness?employeeId=12&tin=...`). Query strings are captured in places request bodies are not: web-server and load-balancer access logs, APM/tracing spans, proxy logs, and browser history. A TIN is a government tax identifier — the same class of data this codebase encrypts at rest for SSN, NIN and passport number — so writing it to plaintext infrastructure logs on every keystroke-triggered save undercuts that protection regardless of what the database does.

I recognise the neighbouring `check-email-identification-no` endpoint uses the same GET-with-query-params shape, which is presumably why this followed suit; an email address is a materially lower-sensitivity value than a tax ID, so consistency with that endpoint is not a strong reason to keep it here.

Fix: make these `@PostMapping` endpoints taking the existing DTOs as `@RequestBody` (with `@Valid`, per the separate finding). They are already non-idempotent in spirit — a point-in-time uniqueness probe rather than a cacheable resource fetch — so POST is also the better HTTP semantic, and it stops React Query and any intermediate cache from retaining the value. The frontend change is confined to swapping `authFetch.get(url, { params })` for `authFetch.post(url, body)` in `useCheckPayrollIdUniqueness` / `useCheckTinUniqueness`.

frontend/src/community/people/api/PeopleApi.ts
```diff
@@ -299,6 +299,40 @@ export const useCheckEmailAndIdentificationNo = (
   });
 };
 
+export const useCheckPayrollIdUniqueness = (
+  employeeId: string | undefined,
+  payrollId: string
+): UseQueryResult<Partial<EmployeeDataExists>> => {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Both new hooks are typed `UseQueryResult<Partial<EmployeeDataExists>>`, and to make that compile `isPayrollIdExists?: boolean` and `isTinExists?: boolean` were appended to `EmployeeDataExists` (EmployeeTypes.tsx:451-452). That interface models the response of a different endpoint (`check-email-identification-no` → `EmployeeDataValidationResponseDto`, whose fields are `isIdentificationNoExists`, `isWorkEmailExists`, `isGoogleDomain`, `isGuestUser`). The backend responses here are distinct single-field DTOs, `PayrollIdUniquenessResponseDto` and `TinUniquenessResponseDto`.

Two costs: every existing consumer of `useCheckEmailAndIdentificationNo` now sees two optional properties its endpoint never returns, and the `Partial<>` wrapper makes the four fields that endpoint *does* always return optional too — so the type no longer tells a reader which fields are guaranteed. It is the type-level equivalent of the `as` assertions the guidelines call out: widening a type to silence an error rather than modelling the actual contract.

Fix: declare `export interface PayrollIdUniquenessResponse { isPayrollIdExists: boolean; }` and `export interface TinUniquenessResponse { isTinExists: boolean; }` mirroring the backend DTOs, type each hook as `UseQueryResult<PayrollIdUniquenessResponse>` / `UseQueryResult<TinUniquenessResponse>`, and revert lines 451-452 of EmployeeTypes.tsx. Note that until the boxed-`Boolean` issue on the response DTO is fixed the field really can arrive as null, so either fix that server-side (preferred) or type it `boolean | null` here — but state it explicitly rather than hiding it behind `Partial`.

frontend/src/community/people/components/organisms/EmploymentFormSection/SubSections/IdentificationDetailsSection.tsx
```diff
@@ -165,6 +184,58 @@ const IdentificationDetailsSection = forwardRef<FormMethods, Props>(
                 ariaLabel={translateAria(["selectEEOJobCategory"])}
               />
             </Grid>
+
+            {!isReadOnly && (
```
ThinuwanW, line 222: 🤖 **Claude** · 🟡 **Suggestion**

The two new fields are wrapped in `{!isReadOnly && (...)}`, so they disappear entirely in read-only mode, while all three pre-existing fields in the same section (SSN, ethnicity, EEO job category) render unconditionally and pass `readOnly={isReadOnly}` to the input. Two inconsistencies follow from the divergence.

First, a people manager viewing a subordinate's profile gets `isReadOnly === true` (`!isPeopleAdmin && userId !== employeeId`, PeopleFormSections.tsx:36) and sees no Payroll ID or TIN row at all — not even an empty read-only field — so the section silently renders a different shape depending on who is looking, and there is no indication the fields exist. Second, hiding is done purely in the client: `mapIdentificationAndDiversityDetails` still populates `payrollId`/`tin` in the response the manager's browser receives, so this is presentation-only and confers no actual protection.

Fix: pick one model and apply it consistently. If these values may be read but not edited by non-admins, render them like their siblings with `readOnly={isReadOnly || !canEditIdentifiers}` and drop the conditional wrapper. If they must not be readable at all outside the admin roles, keep them out of the markup *and* stop serialising them server-side for those viewers, so the UI and the payload agree.

backend/src/main/java/com/skapp/community/peopleplanner/model/Employee.java
```diff
@@ -90,6 +90,12 @@ public class Employee extends Auditable<String> {
 	@Column(name = "identification_no", length = 15)
 	private String identificationNo;
 
+	@Column(name = "payroll_id")
```
ThinuwanW, line 93: 🤖 **Claude** · 🟣 **Nit**

`@Column(name = "payroll_id")` and `@Column(name = "tin")` declare no `length`, so the JPA metadata implies the default 255 while the Liquibase changesets create the columns as unbounded `text` and the business rule enforced in `Validations` and `PeopleConstants.MAX_PAYROLL_ID_LENGTH`/`MAX_TIN_LENGTH` is 50. Three different widths for the same field. Neighbouring columns in this entity are explicit — `identificationNo` is `length = 15`, `phone` is `length = 15`, `authPic` is `length = 500`.

Beyond tidiness this matters if Hibernate schema validation is ever enabled, and `text` is the wrong physical type for a 50-character identifier you intend to index (the first pass already flagged the missing index; MySQL cannot index a `TEXT` column without a prefix length, so the type choice will have to change anyway).

Fix: set `length = 50` on both `@Column` annotations and change the two DDL scripts to `ADD COLUMN \`payroll_id\` varchar(50)` / `varchar(50)`. Since neither changeset has shipped yet, edit them in place rather than adding a follow-up migration.

frontend/src/community/common/assets/languages/english/peopleModule.json
```diff
@@ -838,11 +838,22 @@
...
       "enterSSN": "Enter social security number",
+      "enterPayrollId": "Enter payroll ID",
+      "enterTin": "Enter TIN",
       "selectEthnicity": "Select ethnicity",
       "selectEEOJobCategory": "Select EEO job category",
       "eeoTooltip": "The Equal Opportunity Office (EOO) job categories ensure fair representation within the organization, crucial for reporting diversity and inclusion efforts",
+      "payrollIdTooltip": "A unique internal identifier used for payroll processing. Visible to Admins and Super Admins only.",
+      "tinTooltip": "Tax Identification Number, used for tax and statutory reporting. Visible to Admins and Super Admins only.",
       "maxSSNCharacterLimitError": "SSN should not exceed 20 characters",
+      "payrollIdMaxLengthError": "Payroll ID cannot exceed 50 characters",
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`"payrollIdMaxLengthError": "Payroll ID cannot exceed 50 characters"` and the TIN equivalent on the next line hardcode the limit in the copy, while the schema that raises them reads it from a constant: `.max(characterLengths.CHARACTER_LENGTH, translator(["payrollIdMaxLengthError"]))` (peopleValidations.ts:321-331). The backend does this correctly — `api.error.people.validation.payroll-id-length=Payroll ID cannot exceed {0} characters` is parameterised with `PeopleConstants.MAX_PAYROLL_ID_LENGTH`. So raising the limit later silently leaves the user-facing message stating the old number.

Fix: parameterise the strings (`"Payroll ID cannot exceed {{max}} characters"`) and pass the value through, e.g. `translator(["payrollIdMaxLengthError"], { max: characterLengths.CHARACTER_LENGTH })`, matching how the backend message is already written.

backend/src/main/java/com/skapp/community/peopleplanner/controller/v1/PeopleController.java
shakila-kamalasena: We use the minimum role, so PEOPLE_ADMIN covers SUPER_ADMIN as well

backend/src/main/java/com/skapp/community/peopleplanner/controller/v1/PeopleController.java
shakila-kamalasena: Added @Operation, others we do not use as a practice

backend/src/main/java/com/skapp/community/peopleplanner/controller/v1/PeopleController.java
shakila-kamalasena: Not needed, followed the same pattern as the email

backend/src/main/java/com/skapp/community/peopleplanner/model/Employee.java
shakila-kamalasena: We do not use entity level valdations

backend/src/main/java/com/skapp/community/peopleplanner/model/Employee.java
shakila-kamalasena: Will check later about PII encryption as we cannot use the existing FieldEncryptionConverter

backend/src/main/java/com/skapp/community/peopleplanner/payload/request/PayrollIdExistsCheckDto.java
shakila-kamalasena: Better to have them separatedly

backend/src/main/java/com/skapp/community/peopleplanner/payload/request/PayrollIdUniquenessCheckDto.java
shakila-kamalasena: We do not use dto level validations

backend/src/main/java/com/skapp/community/peopleplanner/payload/response/PayrollIdExistsResponseDto.java
shakila-kamalasena: No need

backend/src/main/java/com/skapp/community/peopleplanner/repository/EmployeeDao.java
shakila-kamalasena: Not needed as they can be reactivated

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
shakila-kamalasena: No need

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
shakila-kamalasena: Valid comment, fixed that

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
shakila-kamalasena: We do not use trim

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
shakila-kamalasena: no need

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
shakila-kamalasena: Removed the seenPayrollIds approach

backend/src/main/java/com/skapp/community/peopleplanner/util/Validations.java
shakila-kamalasena: No need character validations for now

backend/src/main/resources/community/db/changelog/ddl-script/people-ddl-script-v1-alter-table-employee-add-payroll-id-column.sql
shakila-kamalasena: we declare them as text, and no need index or unique

frontend/src/community/common/assets/languages/english/peopleModule.json
shakila-kamalasena: Fixed

frontend/src/community/people/api/PeopleApi.ts
shakila-kamalasena: Fixed by creating two separate response types

frontend/src/community/people/components/organisms/EmploymentFormSection/SubSections/IdentificationDetailsSection.tsx
shakila-kamalasena: This is intended

frontend/src/community/people/components/organisms/EmploymentFormSection/SubSections/IdentificationDetailsSection.tsx
shakila-kamalasena: No need to update the legacy componentes

frontend/src/community/people/components/organisms/EmploymentFormSection/SubSections/IdentificationDetailsSection.tsx
shakila-kamalasena: Frontend gating is correct, backend is done

frontend/src/community/people/components/organisms/EmploymentFormSection/SubSections/IdentificationDetailsSection.tsx
shakila-kamalasena: Fixed

frontend/src/community/people/components/organisms/EmploymentFormSection/SubSections/IdentificationDetailsSection.tsx
shakila-kamalasena: This works as needed

frontend/src/community/people/components/organisms/EmploymentFormSection/EmploymentDetailsForm.tsx
shakila-kamalasena: Fixed

frontend/src/community/people/components/organisms/EmploymentFormSection/EmploymentDetailsForm.tsx
shakila-kamalasena: If there is error, the user cannot submit

frontend/src/community/people/types/UserBulkUploadTypes.ts
shakila-kamalasena: Fixed

frontend/src/community/people/utils/constants/constants.ts
shakila-kamalasena: No need

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
shakila-kamalasena: Fixed

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleReadServiceImpl.java
shakila-kamalasena: Fixed

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
shakila-kamalasena: Removed this approach

frontend/src/community/common/constants/stringConstants.ts
```diff
@@ -3,6 +3,8 @@ export enum characterLengths {
   JOB_TITLE_LENGTH = 50,
   NAME_LENGTH = 50,
   CHARACTER_LENGTH = 50,
+  PAYROLL_ID_LENGTH = 50,
```
hasalarootcode: change this to a constant

frontend/src/community/people/api/utils/QueryKeys.ts
```diff
@@ -30,6 +30,25 @@ export const peopleQueryKeys = {
       identificationNo
     ].filter((val) => val !== undefined);
   },
+  PAYROLL_ID_UNIQUENESS_KEYS: function (
```
hasalarootcode: dont make the query key complex. make it simple as possible

frontend/src/community/people/api/PeopleApi.ts
```diff
@@ -299,6 +301,40 @@ export const useCheckEmailAndIdentificationNo = (
   });
 };
 
+export const useCheckPayrollIdUniqueness = (
+  employeeId: string | undefined,
```
hasalarootcode: make it optional

frontend/src/community/people/api/PeopleApi.ts
```diff
@@ -299,6 +301,40 @@ export const useCheckEmailAndIdentificationNo = (
...
+  payrollId: string
+): UseQueryResult<PayrollIdUniquenessResponse> => {
+  return useQuery({
+    queryKey: peopleQueryKeys.PAYROLL_ID_UNIQUENESS_KEYS(employeeId, payrollId),
+    queryFn: async () => {
```
hasalarootcode: handler

frontend/src/community/people/components/organisms/EmploymentFormSection/EmploymentDetailsForm.tsx
```diff
@@ -87,6 +91,21 @@ const EmploymentDetailsForm = ({
     employeeNumber
   );
 
+  const employeeIdForUniquenessCheck = employee?.common?.employeeId;
+  const payrollId =
+    employee?.employment?.identificationAndDiversityDetails?.payrollId || "";
```
hasalarootcode: remove fallback since default should be null

frontend/src/community/people/components/organisms/EmploymentFormSection/EmploymentDetailsForm.tsx
```diff
@@ -135,6 +154,44 @@ const EmploymentDetailsForm = ({
       Object.keys(identificationFormErrors).length === 0;
 
     if (employmentFormIsValid && identificationFormIsValid) {
+      let hasIdentificationUniquenessError = false;
+
+      if (isPeopleAdmin && payrollId) {
```
hasalarootcode: handle this like employment section's workemail

frontend/src/community/people/hooks/useUserBulkValidation.ts
```diff
@@ -67,7 +67,10 @@ const useUserBulkValidation = () => {
       (!user?.careerProgressionStartDate ||
         dateValidation(user?.careerProgressionStartDate)) &&
       validateCareerProgressionFields(user) &&
-      user?.passportNo
+      user?.passportNo &&
+      (!user?.payrollId ||
```
hasalarootcode: check this

hasalarootcode (changes requested):

backend/src/main/java/com/skapp/community/peopleplanner/controller/v1/PeopleController.java
```diff
@@ -172,6 +174,25 @@ public ResponseEntity<ResponseEntityDto> getEmployeeByIdOrEmail(
 		return new ResponseEntity<>(response, HttpStatus.OK);
 	}
 
+	@Operation(summary = "Check if a Payroll ID is unique",
+			description = "This endpoint checks whether the provided Payroll ID is already assigned to another employee.")
+	@GetMapping(value = "/check-payroll-id-uniqueness")
```
akilarootcode: exists/payroll-id

backend/src/main/java/com/skapp/community/peopleplanner/payload/request/PayrollIdUniquenessCheckDto.java
akilarootcode: exist check

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/EmployeeValidationServiceImpl.java
```diff
@@ -351,6 +357,49 @@ public void validateCreateEmployeeRequestPersonalDetails(EmployeePersonalDetails
...
+		String payrollId = identificationDetails.getPayrollId();
+		if (payrollId == null || payrollId.isBlank()) {
+			return;
+		}
+
+		Validations.validatePayrollId(payrollId);
+
+		boolean isDuplicate;
+		if (user.getEmployee() != null) {
+			isDuplicate = employeeDao.existsByPayrollIdAndEmployeeIdNot(payrollId, user.getEmployee().getEmployeeId());
```
akilarootcode: handle using query

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/EmployeeValidationServiceImpl.java
```diff
@@ -351,6 +357,49 @@ public void validateCreateEmployeeRequestPersonalDetails(EmployeePersonalDetails
...
+		String tin = identificationDetails.getTin();
+		if (tin == null || tin.isBlank()) {
+			return;
+		}
+
+		Validations.validateTin(tin);
+
+		boolean isDuplicate;
+		if (user.getEmployee() != null) {
+			isDuplicate = employeeDao.existsByTinAndEmployeeIdNot(tin, user.getEmployee().getEmployeeId());
```
akilarootcode: same

akilarootcode (changes requested):

backend/src/main/java/com/skapp/community/peopleplanner/controller/v1/PeopleController.java
shakila-kamalasena: Renamed

backend/src/main/java/com/skapp/community/peopleplanner/payload/request/PayrollIdUniquenessCheckDto.java
shakila-kamalasena: Renamed to TinExistCheckDto and PayrollIdExistCheckDto

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/EmployeeValidationServiceImpl.java
shakila-kamalasena: Added to a query

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/EmployeeValidationServiceImpl.java
shakila-kamalasena: Added to a query

frontend/src/community/common/constants/stringConstants.ts
shakila-kamalasena: Changed to const

frontend/src/community/people/api/utils/QueryKeys.ts
shakila-kamalasena: Fixed

frontend/src/community/people/api/PeopleApi.ts
shakila-kamalasena: Fixed

frontend/src/community/people/api/PeopleApi.ts
shakila-kamalasena: Moved

frontend/src/community/people/components/organisms/EmploymentFormSection/EmploymentDetailsForm.tsx
shakila-kamalasena: Removed

frontend/src/community/people/components/organisms/EmploymentFormSection/EmploymentDetailsForm.tsx
shakila-kamalasena: Done

frontend/src/community/people/hooks/useUserBulkValidation.ts
shakila-kamalasena: This is needed with the existing pattern

hasalarootcode (dismissed):

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
```diff
@@ -1351,6 +1387,36 @@ public ResponseEntityDto getEmployeeByIdOrEmail(EmployeeDataValidationDto employ
 		return new ResponseEntityDto(false, employeeDataValidationResponseDto);
 	}
 
+	@Override
+	public ResponseEntityDto checkPayrollIdUniqueness(PayrollIdExistCheckDto payrollIdExistCheckDto) {
```
akilarootcode: rename this to proper name

frontend/src/community/people/api/utils/QueryKeys.ts
```diff
@@ -30,6 +30,16 @@ export const peopleQueryKeys = {
       identificationNo
     ].filter((val) => val !== undefined);
   },
+  PAYROLL_ID_UNIQUENESS_KEYS: (employeeId?: string, payrollId?: string) => [
```
akilarootcode: rename this as well

frontend/src/community/people/utils/peopleValidations.ts
```diff
@@ -311,12 +317,47 @@ export const employeeCareerDetailsValidation = (
...
+    payrollId: Yup.string()
+      .trim()
+      .max(
+        PAYROLL_ID_LENGTH,
+        translator(["payrollIdMaxLengthError"], {
+          max: String(PAYROLL_ID_LENGTH)
+        })
+      )
+      .test(
+        "is-unique-payroll-id",
```
akilarootcode: rename this also

akilarootcode (changes requested):

frontend/src/community/people/utils/peopleValidations.ts
shakila-kamalasena: Done

frontend/src/community/people/api/utils/QueryKeys.ts
shakila-kamalasena: DOne

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java

akilarootcode (approved):

PR #1688, opened 2026-07-30
feat: add contact number validation and country code handling in CRM form

dulsara-manakal:
### Summary
Adds phone number + country code input to CRM Contact and Company forms (Create and Edit) — same flag/dial-code/number widget the People module already uses. Backend now actually checks the number format instead of just checking length.

## Backend changes

**File:** `CrmValidations.java` → `validateContactNumber()`

- **Before:** only checked length (≤15 chars). Any characters allowed.
- **After:** checks the whole string against one regex:
  ```
  ^[0-9]{1,4} [0-9]{7,15}$
  ```
  - First part = country code, 1-4 digits
  - Exactly one space
  - Second part = number, 7-15 digits, nothing else allowed

**To verify with a third-party regex checker** (e.g. regex101.com): paste `^[0-9]{1,4} [0-9]{7,15}$` and test:
- `94 771234567` → valid
- `94771234567` → invalid (no space)
- `+94 771234567` → invalid (plus sign not allowed)
- `94 77-123-4567` → invalid (dash not allowed)

Field stays **optional** — empty is still allowed.

Regex constant lives in `CrmConstants.CONTACT_NUMBER_REGEX`. Tests updated in `CrmValidationsTest.java` — 8 focused cases, one per real failure reason (was 10, trimmed 2 that tested the same thing twice).

## Frontend changes

- Contact and Company forms (Create + Edit) now use the shared `InputPhoneNumber` component instead of a plain text box.
- Country code and number are kept as **two separate fields** in the form (not combined) — matches how People and Org Setup already do it. They only get joined into one string (`"94 771234567"`) right before sending to the backend, not while typing.
- Frontend only checks the number part (`^[0-9]{7,15}$`). Country code isn't checked on the frontend — it always comes from the country dropdown, the user can never type it by hand, so there's nothing to validate there.
- **Bug fixed:** editing a Contact and changing *only* the phone number used to silently do nothing — no save, no error, nothing. Fixed by making sure a phone-only change is actually detected before deciding whether to call the API.

Got it — here's a more detailed, explicit version you can drop straight into that section.

## Why this approach

Before deciding, we checked exactly what People and Org Setup do today, on both frontend and backend.

**People (Employee contact number)**
- **Frontend:** Formik keeps country code and number as two separate fields. Combines them into one string (`"94 771234567"`) on every keystroke, and saves that combined value into a shared store. Validation only checks the number's *length* (7-15 characters) — no check on what characters are actually in it.
- **Backend (normal Add/Edit Employee flow):** takes the whole combined string and checks it against one loose pattern: `^(\+)?[0-9 ]{7,15}$`. This allows a `+` sign and extra spaces anywhere inside — it does not separately check the country code part vs. the number part.
- **Backend (CSV bulk-import only — a completely different, separate code path):** this one actually *does* split the string on the space and checks country code and number separately (`^[0-9]{1,4}$` for the code, then the number). But this stricter check only runs for bulk-import — normal Add/Edit Employee never uses it.

**Org Setup (Organization contact number)**
- **Frontend:** same as People — two separate fields, combined into one string only at the final submit.
- **Backend:** one pattern over the whole combined string: `^[+]?[0-9\- ]{7,25}$`. This allows `+`, dashes, and spaces anywhere, and allows up to 25 characters. It never splits the string apart at all — it doesn't know or care where the country code ends and the number begins.

**CRM (what we built)**
- **Frontend:** two separate fields, same as both — combined into one string only right before sending to the backend. Only the number part is checked (`^[0-9]{7,15}$`, digits only). Country code isn't checked on the frontend at all, because it can only ever come from the dropdown — the user can't type it by hand, so there's nothing to validate there.
- **Backend:** one pattern over the whole combined string: `^[0-9]{1,4} [0-9]{7,15}$`. This *does* enforce structure — it requires digits, then exactly one space, then digits, nothing else. So it's stricter than both People's normal flow and Org Setup's check, and it's closer in spirit to People's bulk-import checker (which is the only other place that actually cares about the two parts separately) — just written as one clean regex instead of manually splitting the string first.

**We did NOT add a separate `countryCode` field to the API or database.** Backend still receives one combined string, same as People and Org Setup do. Reason: the database column is a single string either way, so splitting the API wouldn't remove any work — it would just move the "combine" step from the frontend to the backend. Also, People's own DTO already has a spare `countryCode` field the backend never actually uses — dead weight we didn't want to repeat.


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR changes the CRM contact-number format to a mandatory `<countryCode> <number>` shape (backend regex `^[0-9]{1,4} [0-9]{7,15}$`) and swaps the plain input for the shared `InputPhoneNumber` component on both the company and contact modals. The wiring is largely consistent with the existing `InputPhoneNumber` usage in the people module, i18n keys are reused, and the backend/frontend length bounds (7–15) line up. However, the change is a breaking format migration with no backfill for existing rows, and `splitContactNumber` silently mis-parses every legacy number, which causes duplicated country codes to be written back on company edit.

Found **20** new issue(s): 🔴 5 important, 🟡 10 suggestion(s), 🟣 5 nit(s)

frontend/src/community/crm/utils/crmUtil.ts
```diff
@@ -125,6 +132,33 @@ export const groupItemsByPriority = <T extends Id>(
...
+export const splitContactNumber = (
+  contactNumber: string,
+  defaultCountryCode: string
+): { countryCode: string; number: string } => {
+  if (!contactNumber) {
+    return { countryCode: defaultCountryCode, number: "" };
+  }
+
+  const spaceIndex = contactNumber.indexOf(" ");
+  if (spaceIndex <= 0) {
```
ThinuwanW, line 144: 🤖 **Claude** · 🔴 **Important**

Legacy contact numbers get a duplicated country code written back. Every number stored before this PR is digits-only with no space (the old Yup rule was `^[0-9]{7,15}$` and the old backend rule was just `length <= 15`), so `splitContactNumber` always takes the `spaceIndex <= 0` branch and returns `{countryCode: default, number: "94771234567"}` — i.e. the whole stored value, country code included, lands in the number field. `EditCompanyModalContent.submitEditCompany` always sends `contactNumber` in the PATCH payload (it is not diffed), so saving *any* field on a company with a legacy number rewrites it to `"94 94771234567"`. That passes the new backend regex (11 digits ≤ 15) so it is silently persisted, and repeating the edit prefixes it again. Contacts are only protected by accident because `getChangedContactFields` diffs the field, but they show the same wrong value in the UI. Fix: make `splitContactNumber` recognise legacy input — e.g. only treat the prefix as a country code when the string matches `^\d{1,4} \d{7,15}$`, and otherwise attempt to strip a known dial code from the front of the digits (using `countryCodeList`) before falling back to the default. Also consider not sending `contactNumber` from the company edit modal when it is unchanged.

backend/src/main/java/com/skapp/community/crmplanner/constant/CrmConstants.java
```diff
@@ -50,6 +50,11 @@ public class CrmConstants {
 
 	public static final int PHONE_MAX_LENGTH = 15;
 
+	public static final int PHONE_MIN_LENGTH = 7;
+
+	public static final String CONTACT_NUMBER_REGEX = "^[0-9]{1,4} [0-9]{" + PHONE_MIN_LENGTH + "," + PHONE_MAX_LENGTH
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`CONTACT_NUMBER_REGEX` makes the space-separated country code mandatory, which is a breaking change to the `/v1/crm/companies` and `/v1/crm/contacts` create/edit contracts: any previously valid value (e.g. `"94771234567"`) is now rejected with `CRM_ERROR_CONTACT_NUMBER_INVALID`. There is no Liquibase changeset in this PR to backfill existing `crm_contact.contact_number` / `crm_company.contact_number` rows into the new format, so stored data and the validator now disagree — any consumer that round-trips a fetched number back to the API gets a 400. Add a data-migration changeset under `community/db/changelog/` that rewrites existing values to `<dialCode> <number>` (or leaves them and has the validator accept the legacy bare-digit form during a transition). Also, `api.error.crm.validation.contact-number=Invalid contact number` in `crm-messages.properties` no longer tells the caller what the expected shape is — worth updating it to mention the `<country code> <number>` format now that the constraint is much stricter.

frontend/src/community/crm/components/molecules/CompanyModalForm/CompanyModalForm.tsx
```diff
@@ -80,17 +91,16 @@ const CompanyModalForm: FC<CompanyModalFormProps> = ({
         fullWidth
       />
 
-      <InputField
-        name="contactNumber"
+      <InputPhoneNumber
```
ThinuwanW, line 88: 🤖 **Claude** · 🟡 **Suggestion**

The input's max length and the validation range disagree. `InputPhoneNumber` caps the number field at `getPhoneNumberMaxLength(countryCodeValue)` = `PHONE_NUMBER_LENGTH_MAX - countryCode.length` (13 for a 2-digit code, 11 for a 4-digit code), but both the Yup rule (`isValidPhoneNumber()` = `^[0-9]{7,15}$`) and the new backend regex (`[0-9]{7,15}` for the number part alone) accept up to 15 digits. `getPhoneNumberMaxLength` was written for the combined-length model; under the new split model the subtraction is wrong, so users can never enter a valid 14–15 digit national number. Either pass an explicit `maxLength` of `PHONE_NUMBER_LENGTH_MAX` for these fields, or tighten the Yup/backend upper bound to match what the input actually allows. Same issue applies to `ContactModalForm.tsx:251`.

frontend/src/community/crm/components/molecules/CompanyModalForm/CompanyModalForm.tsx
```diff
@@ -80,17 +91,16 @@ const CompanyModalForm: FC<CompanyModalFormProps> = ({
...
-        placeholder={translateText(["placeholders", "contactNumber"])}
-        onChange={handleChange}
-        onBlur={handleBlur}
-        errorMessage={errors.contactNumber}
-        state={errors.contactNumber ? "error" : "default"}
-        aria-label={translateText(["ariaLabels", "contactNumber"])}
-        fullWidth
+        countryCodeValue={values.countryCode}
+        onChangeCountry={handleChangeCountry}
+        onChange={handleChangeContactNumber}
```
ThinuwanW, line 94: 🤖 **Claude** · 🟡 **Suggestion**

Behavioural regression in the company modal: the old `InputField` passed `onBlur={handleBlur}`, and both `AddCompanyModalContent` and `EditCompanyModalContent` use `validateOnChange: false, validateOnBlur: true`, so an invalid contact number surfaced as soon as the field lost focus. `InputPhoneNumber` exposes no `onBlur` prop and none is wired here, so with `validateOnChange: false` the field is never validated until the user clicks Save. Either add an `onBlur` pass-through to `InputPhoneNumber` and wire `handleBlur`, or validate this field on change (e.g. call `setFieldTouched('contactNumber', true, true)` from `handleChangeContactNumber`).

frontend/src/community/crm/components/molecules/CompanyModalForm/CompanyModalForm.tsx
```diff
@@ -43,6 +44,16 @@ const CompanyModalForm: FC<CompanyModalFormProps> = ({
     submitForm
   } = formik;
 
+  const handleChangeCountry = async (code: string) => {
+    await setFieldValue("countryCode", code);
+  };
+
+  const handleChangeContactNumber = async (
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`handleChangeContactNumber` is an `async` wrapper that does nothing but forward to Formik's synchronous `handleChange` — the `async` keyword only exists to satisfy `InputPhoneNumber`'s `onChange?: (e) => Promise<void>` signature, and it makes the returned promise resolve before Formik state has settled. The identical `handleChangeCountry` + `handleChangeContactNumber` pair is duplicated verbatim in `ContactModalForm.tsx:90-98`. Either relax the `InputPhoneNumber` prop type to `(e: ChangeEvent<HTMLInputElement>) => void | Promise<void>` and pass `handleChange` directly, or extract a small shared hook (e.g. `useCrmPhoneFieldHandlers(formik)`) so the two modals don't carry copy-pasted handlers.

frontend/src/community/crm/utils/crmUtil.ts
```diff
@@ -125,6 +132,33 @@ export const groupItemsByPriority = <T extends Id>(
   return { prioritized, deprioritized };
 };
 
+export const splitContactNumber = (
```
ThinuwanW, line 135: 🤖 **Claude** · 🟡 **Suggestion**

`splitContactNumber` and `combineContactNumber` are new pure utilities with non-trivial parsing/formatting behaviour and no unit tests, even though `frontend/src/community/crm/utils/__test__/crmUtil.test.ts` already exists. Please add cases for: empty input, a well-formed `"94 771234567"`, a legacy bare-digit value, a leading-space value, and `combineContactNumber` with an empty number (returns `""`) — the last one matters because `formatPhoneNumber` also returns `""` when `countryCode` is falsy, which would silently drop a user-entered number if `useGetDefaultCountryCode` ever stopped guaranteeing a fallback.

backend/src/test/java/com/skapp/community/crmplanner/util/CrmValidationsTest.java
```diff
@@ -66,18 +66,50 @@ void validateContactNumber_Blank_DoesNotThrow() {
...
+		@DisplayName("Number containing non-digit characters - throws CRM_ERROR_CONTACT_NUMBER_INVALID")
+		void validateContactNumber_NonNumericNumber_ThrowsInvalid() {
+			ModuleException ex = assertThrows(ModuleException.class,
+					() -> CrmValidations.validateContactNumber("94 77-123-4567"));
+			assertEquals(CrmMessageConstant.CRM_ERROR_CONTACT_NUMBER_INVALID, ex.getMessageKey());
+		}
+
+		@Test
+		@DisplayName("Number exceeding max length - throws CRM_ERROR_CONTACT_NUMBER_INVALID")
+		void validateContactNumber_NumberTooLong_ThrowsInvalid() {
```
ThinuwanW, line 102: 🤖 **Claude** · 🟣 **Nit**

Boundary coverage gap on the new regex: there is no test for a number shorter than `PHONE_MIN_LENGTH` (e.g. `"94 123456"` → should throw) and no accepting test at the upper bound (`"94 " + "1".repeat(PHONE_MAX_LENGTH)` → should not throw) or for a 4-digit country code (`"1234 7712345"`). Since the bounds are now interpolated into the regex from constants, these are exactly the cases a future constant change would break.

backend/src/main/java/com/skapp/community/crmplanner/constant/CrmConstants.java
```diff
@@ -50,6 +50,11 @@ public class CrmConstants {
 
 	public static final int PHONE_MAX_LENGTH = 15;
 
+	public static final int PHONE_MIN_LENGTH = 7;
```
ThinuwanW, line 53: 🤖 **Claude** · 🟣 **Nit**

Naming/magic-number: `PHONE_MIN_LENGTH`/`PHONE_MAX_LENGTH` now bound only the subscriber-number portion, not the phone number as a whole (a valid value is up to 20 characters), which will mislead the next reader — consider renaming to `PHONE_NUMBER_MIN_LENGTH`/`PHONE_NUMBER_MAX_LENGTH` or `SUBSCRIBER_*`. Also the `{1,4}` country-code bound is hardcoded inline in the regex while the other two bounds are constants; extract a `COUNTRY_CODE_MAX_LENGTH = 4` for consistency.

frontend/src/community/crm/components/molecules/ContactModalForm/ContactModalForm.tsx
```diff
@@ -23,8 +24,7 @@ import {
 } from "~community/crm/constants/commonConstants";
 import {
   CONTACT_EMAIL_MAX_LENGTH,
-  CONTACT_NAME_MAX_LENGTH,
-  CONTACT_NUMBER_MAX_LENGTH
+  CONTACT_NAME_MAX_LENGTH
```
ThinuwanW, line 27: 🤖 **Claude** · 🟣 **Nit**

`CONTACT_NUMBER_MAX_LENGTH` is dropped from this import and is now unreferenced anywhere in the codebase (only its declaration in `~community/crm/constants/contactConstants.ts:7` remains). Remove the now-dead constant, or reuse it as the explicit `maxLength` for the phone field (see the maxLength mismatch noted above).

backend/src/main/java/com/skapp/community/crmplanner/constant/CrmConstants.java
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

A new bespoke phone regex is being introduced when the codebase already has a shared one, which the backend guidelines explicitly call out ("New validation patterns that ignore the existing Validation utility class"). `com.skapp.community.common.util.Validation` already defines `COUNTRY_CODE_PATTERN = ^[0-9]{1,4}$` and `PHONE_NUMBER_PATTERN = ^(\+)?[0-9 ]{7,15}$`, and `com.skapp.community.peopleplanner.util.Validations.isValidPhoneNumber(String)` (lines 37-50) already implements exactly this contract: it splits on the first space, validates the country code and the subscriber part separately, AND falls back to validating the whole string when there is no space — i.e. it accepts the legacy digits-only values already in `crm_company.contact_number` / `crm_contact.contact_number`. `CrmValidations` already imports that same `Validations` class (line 15) and delegates to it for email, so the precedent is in the very same file. Suggested fix: delete `CONTACT_NUMBER_REGEX` and `PHONE_MIN_LENGTH`, and in `validateContactNumber` call `Validations.isValidPhoneNumber(contactNumber.trim())`, throwing `CRM_ERROR_CONTACT_NUMBER_INVALID` when it returns false. This keeps CRM consistent with how the People module stores and validates the identical format, and removes the backward-compatibility break for free.

frontend/src/community/crm/utils/crmUtil.ts
ThinuwanW, line 144: 🤖 **Claude** · 🔴 **Important**

Guard the blank-phone case explicitly. Both company and contact forms now always carry a country code (defaulted from the org, never empty), while the phone field is optional. If the combine helper unconditionally joins with a space, a company/contact saved with an empty phone number produces `"94 "` instead of `null`/`""` — which the new backend `CONTACT_NUMBER_REGEX` will then reject, so creating or editing a record with no phone number fails with the generic error toast (`toastMessages.add.errorDescription`), and clearing an existing number becomes impossible. The existing shared helper `formatPhoneNumber` in `~community/common/utils/commonUtil.ts:356` already handles this (`if (!countryCode || !phone) return ""`). Ensure the combine helper returns an empty string when the subscriber part is blank, and that the payload builders in `AddCompanyModalContent`, `EditCompanyModalContent`, `AddContactModalContent` and `EditContactModalContent` still coerce that to `null`/`undefined` the way they did before.

frontend/src/community/crm/utils/companyValidations.ts
```diff
@@ -17,22 +17,18 @@ export const addCompanyValidations = (translator: TranslatorFunctionType) =>
...
+      .matches(isValidPhoneNumber(), {
+        message: translator(["validations", "contactNumber"]),
+        excludeEmptyString: true
+      }),
     website: Yup.string()
       .nullable()
       .optional()
       .transform((v) => (v === "" ? null : v))
-      .matches(isValidCompanyWebsiteUrl(), translator(["validations", "website"]))
+      .matches(
```
ThinuwanW, line 29: 🤖 **Claude** · 🔴 **Important**

Missing paired change: the company form's `contactNumber` Yup rule was reworked here for the new subscriber-only value, but `frontend/src/community/crm/utils/contactValidations.ts` was not touched. `addContactValidations` is the schema `ContactModalForm` uses (ContactModalForm.tsx:74) and it still validates `contactNumber` with `.matches(isValidPhoneNumber(), { excludeEmptyString: true })` — i.e. `^[0-9]{7,15}$` — and has no rule at all for the newly added `countryCode` field. Two sibling forms now write the same backend column under different client-side rules, so a value accepted by the contact modal can be rejected by the company modal or by the backend. Apply the same change to `contactValidations.ts` and make both schemas agree with the backend contract (whatever that ends up being after the `Validations.isValidPhoneNumber` reuse above).

frontend/src/community/crm/utils/crmUtil.ts
ThinuwanW, line 135: 🤖 **Claude** · 🟡 **Suggestion**

Wrong module + duplication. The split/combine pair reimplements logic that already exists in shared code: the join is `formatPhoneNumber(countryCode, phone)` in `~community/common/utils/commonUtil.ts:356` (identical `${countryCode.trim()} ${phone.trim()}`), and the split is the same first-token-is-the-country-code parse done inline in `~community/people/components/organisms/PersonDetailsSection/SubSections/ContactDetailsSection.tsx:53-59`. Putting a third copy in a CRM-only util means the three implementations will drift (they already disagree on how a legacy no-space value is interpreted: People treats the whole string as the country code, this one apparently treats it as the subscriber number). Move both helpers into `~community/common/utils/commonUtil.ts` next to `formatPhoneNumber`, have `formatPhoneNumber` be the combine implementation, and refactor `ContactDetailsSection` to use the shared split.

frontend/src/community/crm/components/molecules/CompanyModalForm/CompanyModalForm.tsx
```diff
@@ -43,6 +44,16 @@ const CompanyModalForm: FC<CompanyModalFormProps> = ({
     submitForm
   } = formik;
 
+  const handleChangeCountry = async (code: string) => {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Duplicated Formik handlers. The country-code + phone change handler pair added here (47-56) is repeated almost verbatim in `ContactModalForm.tsx` (90-99), and a third copy already exists as a proper custom hook in `~community/people/hooks/useContactDetailsFormHandlers.tsx:62-76` (`onChangeCountry` -> `setFieldValue("countryCode", ...)`, `onChangePhoneNumber` -> `setFieldValue(<phoneField>, ...)`), which even has its own test file. Per the frontend guideline "2+ hooks with same logic -> extract custom hook", extract a single `useContactNumberFormHandlers({ formik, countryCodeField, numberField })` (or generalize the existing People hook by parameterising the field names) and consume it from both CRM modals.

frontend/src/community/crm/components/molecules/EditCompanyModalContent/EditCompanyModalContent.tsx
```diff
@@ -35,8 +38,8 @@ const EditCompanyModalContent: React.FC = () => {
   const selectedCompany = getCompanyById(selectedCompanyId!);
 
   const initialValues = useMemo(
-    () => getCompanyFormInitialValues(selectedCompany),
-    [selectedCompany]
+    () => getCompanyFormInitialValues(defaultCountryCode, selectedCompany),
```
ThinuwanW, line 41: 🤖 **Claude** · 🟡 **Suggestion**

Threading a store-derived value into `initialValues` interacts badly with `enableReinitialize: true` (line 48 of this file, and the equivalent in the other three modal contents). `useGetDefaultCountryCode` reads `country` from `useCommonStore`, whose initial value is the hardcoded `"Sri Lanka"` in `orgDetailsSlice.ts:6`; the real org country is only written later by the `useGetOrganization` effect in `Drawer.tsx:202-206` / `FullPageContainerLayout.tsx:73-77`. Two consequences: (1) if that query settles while the modal is open, `initialValues` changes identity, `enableReinitialize` fires and Formik silently discards the user's in-progress edits; (2) until it settles, a non-Sri-Lankan org sees `+94` preselected and can save it. Snapshot the code once at mount (`useState(() => defaultCountryCode)` or a `useRef`) rather than letting it flow into `initialValues` on every render.

backend/src/main/java/com/skapp/community/crmplanner/util/CrmValidations.java
```diff
@@ -87,7 +87,7 @@ public static void validateContactNumber(String contactNumber) {
 			return;
 		}
 
-		if (contactNumber.length() > CrmConstants.PHONE_MAX_LENGTH) {
+		if (!contactNumber.matches(CrmConstants.CONTACT_NUMBER_REGEX)) {
```
ThinuwanW, line 90: 🤖 **Claude** · 🟡 **Suggestion**

Inconsistent trimming. Every other pattern-based validator in this class defensively trims before matching — `validateContactName` uses `name.trim().matches(...)` (line 57), `validateContactEmail` passes `email.trim()` (line 68), `validateDealDescription` uses `description.trim().length()` (line 135). The new regex check matches against the raw string, so it only works because `JacksonConfig.StringTrimmingDeserializer` happens to trim every inbound String at the HTTP boundary. Any non-HTTP caller (the enterprise CRM overrides, a future importer, or a direct service-to-service call) passing a padded value now gets a 400. Use `contactNumber.trim().matches(...)` for consistency with the rest of the class.

frontend/src/community/crm/components/molecules/ContactModalForm/ContactModalForm.tsx
```diff
@@ -240,19 +248,16 @@ const ContactModalForm = ({
         />
       )}
 
-      <InputField
-        name="contactNumber"
+      <InputPhoneNumber
```
ThinuwanW, line 245: 🤖 **Claude** · 🟡 **Suggestion**

Design-system mismatch worth a conscious decision. Before this PR the entire `src/community/crm` tree had zero `@mui/material` imports and used only `@rootcodelabs/skapp-ui` + Tailwind; `InputPhoneNumber` is an MUI + `react-phone-input-2` component (`useTheme`, `Stack`, `Typography`) that renders its own label markup, hardcodes colours instead of theme tokens (`color: "black"` at line 143 and `style={{ color: "red" }}` at line 147 of InputPhoneNumber.tsx), and installs a 300 ms `setInterval` that mutates global DOM (`document.querySelector(".country-list")`, `.flag-dropdown input`) for as long as it is mounted. Inside these modals its label/required-marker/error styling will not match the adjacent skapp-ui `InputField`s. Either add a phone input to the skapp-ui set, or at minimum verify the rendered label and error states line up with the sibling fields and that the polling interval is acceptable for a modal.

backend/src/test/java/com/skapp/community/crmplanner/controller/v1/CrmCompanyControllerIntegrationTest.java
```diff
@@ -165,7 +165,7 @@ private CrmCompanyCreateDto createValidPayload() {
 		dto.setIndustry(CrmIndustry.TECHNOLOGY_INFORMATION_AND_MEDIA);
 		dto.setWebsite("https://acme.com");
 		dto.setAddress("123 Main St");
-		dto.setContactNumber("94771234567");
+		dto.setContactNumber("94 771234567");
```
ThinuwanW, line 168: 🤖 **Claude** · 🟡 **Suggestion**

The integration tests only swap the fixture literals to the new spaced format; nothing at the API boundary asserts the new rejection behaviour. Since this is a wire-format change to two public endpoints (`POST/PATCH /v1/crm/companies` and `/v1/crm/contacts`), add tests that a legacy digits-only value and a malformed country code both return 400 with `CRM_ERROR_CONTACT_NUMBER_INVALID`. Those tests are the ones that would have made the backward-compatibility break visible in CI rather than in production.

frontend/src/community/crm/components/molecules/EditContactModalContent/EditContactModalContent.tsx
```diff
@@ -59,10 +64,17 @@ const EditContactModalContent = () => {
     }
   );
 
+  const { countryCode: originalCountryCode, number: originalContactNumber } =
```
ThinuwanW, line 67: 🤖 **Claude** · 🟣 **Nit**

Null-safety in the block being rewritten. `getContactById` is typed `(id: number) => CrmContact | undefined` (`~community/crm/types/StoreTypes.ts:38`) and `selectedContactId!` is non-null-asserted on line 29, so `selectedContact` genuinely can be `undefined` (e.g. the contact is removed from the store while the modal is open). Make sure the new split call tolerates `undefined`/`null` input rather than assuming a string. While you are in this object, note lines 63-64 still assign the possibly-`undefined` `selectedContact?.name` / `?.email` into the non-optional `string` fields of `CrmContactFormValues` — worth adding the `?? ""` fallbacks that the neighbouring lines already use.

backend/src/test/java/com/skapp/community/crmplanner/controller/v1/CrmContactControllerIntegrationTest.java
```diff
@@ -365,15 +365,15 @@ void createContact_WithLeadingTrailingSpaces_SpacesTrimmed() throws Exception {
 		dto.setName("  John Doe  ");
 		dto.setEmail("  john.doe@example.com  ");
 		dto.setCompanyId(companyId);
-		dto.setContactNumber("  9876543210  ");
+		dto.setContactNumber("  94 9876543210  ");
```
ThinuwanW, line 368: 🤖 **Claude** · 🟣 **Nit**

The "leading/trailing spaces trimmed" test now covers a format in which an interior space is semantically significant, but `JacksonConfig.StringTrimmingDeserializer` (line 36) only trims the edges — it does not collapse interior runs. Add a case asserting that a value with a doubled interior space (e.g. two spaces between the country code and the number) is rejected, so the test documents that only edge whitespace is forgiving.

frontend/src/community/crm/utils/crmUtil.ts
dulsara-manakal: That is the expected behaviour according to the requirements. When it comes to legacy numbers, if they are not saved as country code and phone number separately, the whole number will show in the phone number field  and then the update with suitable contrycode is upto the user otherwise the country code will set to org country's code. This is the requirement.

backend/src/main/java/com/skapp/community/crmplanner/constant/CrmConstants.java
dulsara-manakal: According to the requirements, the legacy phone numbers behave the same way, if they are not saved as country code and phone number separately with a space, the whole number will show in the phone number field in edit form and then the update with suitable contrycode is upto the user otherwise the country code will set to org country's code. So migration is no need according to this verdict.

frontend/src/community/crm/components/molecules/CompanyModalForm/CompanyModalForm.tsx
dulsara-manakal: As we discuss now, we only use onBlur for validation purposes.

frontend/src/community/crm/components/molecules/CompanyModalForm/CompanyModalForm.tsx
dulsara-manakal: Cannot change `InputPhoneNumber`  component since it has existing usages and keeps same behaviour and is not going to add a hook for this since it used only two places and has dedicated behavior for each.

frontend/src/community/crm/components/molecules/CompanyModalForm/CompanyModalForm.tsx
dulsara-manakal: `getPhoneNumberMaxLength()` is used to validate for the input field, and it is necessary. So for other places the length of 15 characters is what the whole codebase used and also not wrong according to the global E.164. Therefore, keep the same behaviour.

frontend/src/community/crm/utils/crmUtil.ts
dulsara-manakal: Added test

backend/src/main/java/com/skapp/community/crmplanner/constant/CrmConstants.java
dulsara-manakal: According to the new requirement, the phone number cannot have any non-digit values, but existing ones accept `+`

frontend/src/community/crm/components/molecules/ContactModalForm/ContactModalForm.tsx
dulsara-manakal: Removed

backend/src/main/java/com/skapp/community/crmplanner/util/CrmValidations.java
dulsara-manakal: Do not want to trim since these values are trimmed from global trimmer so this final validation is having only the trimmed value.

backend/src/test/java/com/skapp/community/crmplanner/util/CrmValidationsTest.java
dulsara-manakal: Added test

backend/src/main/java/com/skapp/community/crmplanner/constant/CrmConstants.java
dulsara-manakal: Added new const

frontend/src/community/crm/utils/crmUtil.ts
dulsara-manakal: This is already handled, `combineContactNumber` checks `if (!number) return "";` as its first line, before it ever touches the country code, so a blank phone number never gets glued into something like "94". All four payload builders already do `combineContactNumber(...) || null`, so an empty result is coerced to null before it's sent.

frontend/src/community/crm/utils/companyValidations.ts
dulsara-manakal: This is wrong, both contact and company have the same. but added `.trim()` for consistency

frontend/src/community/crm/utils/crmUtil.ts
dulsara-manakal: The current approach already uses `formatPhoneNumber` with a tiny wrapper to handle legacy numbers specific to CRM.
The split side is genuinely separate, and it disagrees with People's version on legacy no-space values (People puts the whole blob in the country code, ours puts it in the number). That was deliberate: react-phone-input-2 has no real "blank" flag state, so we default the country and leave the raw value in the number field instead of stuffing it into the country code slot. Unifying this properly means touching People's screens too, which is outside this PR's scope.

frontend/src/community/crm/components/molecules/CompanyModalForm/CompanyModalForm.tsx
dulsara-manakal: Added a shared hook.

frontend/src/community/crm/components/molecules/ContactModalForm/ContactModalForm.tsx
dulsara-manakal: This is the required approach, so use the existing components and libraries.

frontend/src/community/crm/components/molecules/EditCompanyModalContent/EditCompanyModalContent.tsx
dulsara-manakal: Fair point, but I think this is low-risk in practice. Hitting it requires the org-details fetch to still be in flight by the time a user has (1) logged in, (2) navigated to the CRM module, (3) opened Contacts/Companies, and (4) opened the Add/Edit modal, several clicks and page loads deep. That fetch kicks off the moment the app mounts, so in virtually all real-world cases it's already resolved long before a user clicks through that many steps to reach this modal. 

and also showing the default country code is happening (1) when there is no specific country code to extract from the existing phone number in the edit form (legacy phone numbers) or (2) in the add form, so the scenario this comment describes doesn't make any impact on that and is good with the requirement perspective as well.

backend/src/test/java/com/skapp/community/crmplanner/controller/v1/CrmCompanyControllerIntegrationTest.java
dulsara-manakal: Added integration test.

frontend/src/community/crm/components/molecules/EditContactModalContent/EditContactModalContent.tsx
dulsara-manakal: Already added a fallback to  split method.

backend/src/test/java/com/skapp/community/crmplanner/controller/v1/CrmContactControllerIntegrationTest.java
dulsara-manakal: The test already covers against the pattern no need to add new test.

frontend/src/community/crm/components/molecules/AddContactModalContent/AddContactModalContent.tsx
```diff
@@ -80,7 +84,9 @@ const AddContactModalContent = () => {
     const payload: CrmContactCreatePayload = {
       name: values.name.trim(),
       email: values.email.trim(),
-      contactNumber: values.contactNumber.trim() || undefined,
+      contactNumber:
+        combineContactNumber(values.countryCode, values.contactNumber.trim()) ||
+        undefined,
```
hasalarootcode, line 89: make it optional without undefined

frontend/src/community/crm/components/molecules/EditCompanyModalContent/EditCompanyModalContent.tsx
hasalarootcode: open for discussion

frontend/src/community/crm/hooks/usePhoneNumberFieldHandlers.ts
```diff
@@ -0,0 +1,28 @@
+import { FormikProps } from "formik";
+import { ChangeEvent } from "react";
+
+interface UsePhoneNumberFieldHandlersParams<T> {
+  formik: FormikProps<T>;
+  countryCodeField: keyof T;
+}
+
+const usePhoneNumberFieldHandlers = <T>({
```
hasalarootcode, line 9: remove this hook

frontend/src/community/crm/utils/companyUtil.ts
```diff
@@ -6,16 +6,26 @@ import {
...
+    company?.contactNumber ?? "",
+    defaultCountryCode
+  );
+
+  return {
+    name: company?.name || "",
+    industry: company?.industry || CrmIndustryEnum.NONE,
+    website: company?.website || "",
+    address: company?.address || "",
+    countryCode,
```
hasalarootcode, line 25: make it consistant for the fallbacks

frontend/src/community/crm/utils/crmUtil.ts
```diff
@@ -125,6 +132,33 @@ export const groupItemsByPriority = <T extends Id>(
...
+  contactNumber: string,
+  defaultCountryCode: string
+): { countryCode: string; number: string } => {
```
hasalarootcode, line 138: interface

frontend/src/community/crm/utils/crmUtil.ts
```diff
@@ -125,6 +132,33 @@ export const groupItemsByPriority = <T extends Id>(
...
+  if (!contactNumber) {
```
hasalarootcode, line 139: parent's responsibility

frontend/src/community/crm/utils/crmUtil.ts
hasalarootcode: Open for discussion

hasalarootcode (changes requested):

frontend/src/community/crm/utils/crmUtil.ts
dulsara-manakal: Moved to parent level.

frontend/src/community/crm/utils/crmUtil.ts
dulsara-manakal: Added  interface

frontend/src/community/crm/components/molecules/CompanyModalForm/CompanyModalForm.tsx
dulsara-manakal: Removed the hook

frontend/src/community/crm/components/molecules/AddContactModalContent/AddContactModalContent.tsx
dulsara-manakal: Removed undefined

frontend/src/community/crm/hooks/usePhoneNumberFieldHandlers.ts
dulsara-manakal: Removed this hook

frontend/src/community/crm/utils/companyUtil.ts
dulsara-manakal: did not add a fallback since changing the type makes an impact on other places as well. And with refactoring, this will be fixed.

hasalarootcode (approved):

backend/src/main/java/com/skapp/community/crmplanner/constant/CrmConstants.java
```diff
@@ -50,6 +50,15 @@ public class CrmConstants {
...
+
+	public static final int COUNTRY_CODE_MIN_LENGTH = 1;
+
+	public static final int COUNTRY_CODE_MAX_LENGTH = 4;
+
+	public static final String CONTACT_NUMBER_REGEX = "^[0-9]{" + COUNTRY_CODE_MIN_LENGTH + ","
```
akilarootcode, line 59: remove this and CONTACT_NUMBER_REGEX  use existing one

frontend/src/community/crm/components/molecules/EditContactModalContent/EditContactModalContent.tsx
```diff
@@ -59,10 +64,17 @@ const EditContactModalContent = () => {
...
+    splitContactNumber({
```
akilarootcode, line 68: need to discuss with thinuwan ayya regarding this. db data corrupted

frontend/src/community/crm/utils/companyValidations.ts
```diff
@@ -15,24 +15,21 @@ export const addCompanyValidations = (translator: TranslatorFunctionType) =>
...
-          if (!inputContactNumber || inputContactNumber === "") {
-            return true;
-          }
-
-          return isValidPhoneNumber().test(inputContactNumber);
-        }
-      ),
+      .matches(isValidPhoneNumber(), {
+        message: translator(["validations", "contactNumber"]),
+        excludeEmptyString: true
```
akilarootcode, line 23: remove excludeEmptyString: true

akilarootcode (changes requested):

PR #1689, opened 2026-07-30
Fix/crm 350 company search relevance

dulsara-manakal:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR is a single-string i18n copy change: the companies table search placeholder becomes "Search by company name". I verified the wording is accurate against the backend — `CrmCompanyRepositoryImpl.getCompanyMetrics` filters only on `CrmCompany_.name` (and the new relevance ordering also ranks on name only), and `CompanyTable.tsx:166` is the sole consumer of `companies.table.search`. JSON stays valid, English is the only locale file so no other translations go stale, and no test asserts the old string. The only gap is that a sibling key in the same file was left with the old wording, producing a user-visible inconsistency in the shared contact modal.

Found **7** new issue(s): 🟡 3 suggestion(s), 🟣 4 nit(s)

frontend/src/community/common/assets/languages/english/crmModule.json
```diff
@@ -290,7 +290,7 @@
     "table": {
       "clearButtonAriaLabel": "Clear search",
       "filterButtonAriaLabel": "Filter companies",
-      "search": "Search company",
+      "search": "Search by company name",
```
ThinuwanW, line 294: 🤖 **Claude** · 🟡 **Suggestion**

Same clarification was not applied to `contacts.editContactModal.placeholders.company`, which still reads "Search company" (line 100), while `contacts.addContactModal.placeholders.company` (line 29) already reads "Search by company name". Both keys feed the *same* rendered element: `ContactModalForm.tsx:210` renders `translateContactText(["placeholders", "company"])`, and the translator scope differs only by parent modal (`AddContactModalContent` scopes to `addContactModal`, `EditContactModalContent.tsx:12-16` scopes to `editContactModal`). So the identical company lookup field shows two different placeholders depending on whether the user is adding or editing a contact. The lookup behind it (`buildPredicatesToFindComapanies` in `CrmCompanyRepositoryImpl`) also matches on company name only, so the new wording is the correct one. Fix: change line 100 to "Search by company name" so all three company-search placeholders match.

ThinuwanW, line 294: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

Copy style is now inconsistent across the three CRM table search placeholders: companies is "Search by company name" (line 293), contacts is "Search contact" (line 174), tasks is "Search tasks" (line 769). These render in visually identical `InputField` search boxes (`CompanyTable.tsx:166`, `ContactTable.tsx:217`). Worth noting that the contacts backend (`CrmContactRepositoryImpl.buildPredicates`) matches on contact name OR owner first/last name, so if the intent of this change is to set accurate expectations about search scope, "Search contact" understates it. Out of strict scope for this PR — flagging so the team can decide whether to align the phrasing pattern in a follow-up.

ThinuwanW, line 294: 🤖 **Claude** · 🟡 **Suggestion**

The new copy promises a company-name search, but the value is never trimmed on the way to the API, so the promise breaks on whitespace. `CompanyTable.tsx:34` debounces the raw `searchTerm` and `:41` passes it straight into `useGetCompanyMetrics`, which forwards it as `searchKeyword` (`CompanyApi.ts:29-42`). The backend metrics predicate does not trim either — `CrmCompanyRepositoryImpl.buildPredicates` builds `%<lowercased keyword>%` with no `.trim()`, unlike the sibling `buildPredicatesToFindComapanies`, which does trim with `Locale.ROOT`. Result: searching " acme" (leading space, very common when pasting) returns zero rows even though "Acme" exists, and "acme " drops the exact-match row entirely; it also defeats the relevance ranking in `buildOrderBy`, whose rank-0 branch compares the full lowercased name to the untrimmed keyword, so an exact hit can never rank first. Secondary effect: `companyQueryKeys.GET_COMPANY_DATA_BY_SEARCH(searchKeyword, limit)` is keyed on the untrimmed string, so "acme" and "acme " create two separate React Query cache entries and two network round trips. Fix on the client with `useDebounce(searchTerm.trim(), COMPANY_NAME_DEBOUNCE_DELAY)` (matching `ContactModalForm.tsx:63-66`, which already trims before debouncing), and ideally trim server-side too so the metrics query matches the filter query's behaviour. Note this is in unchanged code, but it is the exact behaviour the reworded placeholder now advertises and the ticket this branch is named for.

frontend/src/community/common/assets/languages/english/crmModule.json
```diff
@@ -290,7 +290,7 @@
     "table": {
       "clearButtonAriaLabel": "Clear search",
       "filterButtonAriaLabel": "Filter companies",
```
ThinuwanW, line 292: 🤖 **Claude** · 🟡 **Suggestion**

`filterButtonAriaLabel` ("Filter companies") labels an `IconButton` that is hardcoded `disabled` in `CompanyTable.tsx:174-181` — it can never be activated, so both the control and this key are effectively dead. This matters more after this PR: the copy now tells users the box only matches company names, and the only other narrowing affordance on the page is a button that does nothing. Compare `ContactTable.tsx:225-233`, which ships a working company filter `Dropdown`. Either implement the filter, or remove the disabled `IconButton` and this key until it is ready. If it must stay visible as a placeholder affordance, add a `title`/tooltip explaining why it is disabled — a disabled button with only an aria-label gives screen-reader and sighted users no reason for the dead control.

frontend/src/community/common/assets/languages/english/crmModule.json
```diff
@@ -290,7 +290,7 @@
     "table": {
       "clearButtonAriaLabel": "Clear search",
```
ThinuwanW, line 291: 🤖 **Claude** · 🟣 **Nit**

The `table` block defines `clearButtonAriaLabel` and `filterButtonAriaLabel` but no aria label for the search input itself, so the placeholder string changed on line 293 is the input's only accessible name. Placeholder-as-label is a known a11y anti-pattern (the name disappears once text is entered, and some AT/translation setups ignore it), and it is inconsistent with the rest of this module: every form `InputField` passes an explicit `aria-label` (`CompanyModalForm.tsx:77,92,105,118`; `ContactModalForm.tsx:179,199,253`). Add a sibling key here — e.g. `"searchAriaLabel": "Search companies by name"` — and pass `aria-label={translateText(["table", "searchAriaLabel"])}` on the `InputField` at `CompanyTable.tsx:160`. The same gap exists on `ContactTable.tsx:211` and `DealsHeader.tsx:52`, so consider fixing all three for consistency.

frontend/src/community/common/assets/languages/english/crmModule.json
```diff
@@ -290,7 +290,7 @@
...
       "filterButtonAriaLabel": "Filter companies",
-      "search": "Search company",
+      "search": "Search by company name",
       "infiniteScrollLoadingMessage": "Loading companies...",
       "closedDealsLabel": "Deals closed",
       "overdueLabel": "overdue",
```
ThinuwanW, line 297: 🤖 **Claude** · 🟣 **Nit**

Missing paired copy change: now that the placeholder states the search only matches company names, the no-results copy at `table.emptySearchState` ("No results found." / "We couldn't find any results for your search. Please try a different term.") gives no hint about why a website, phone number, or industry query returned nothing. Consider making it specific, e.g. description: "We couldn't find any companies with that name. Try a different name." That keeps the empty state consistent with the narrowed promise the placeholder now makes.

frontend/src/community/common/assets/languages/english/crmModule.json
ThinuwanW, line 294: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

Structural duplication in this locale file makes copy drift like this inevitable: `contacts.addContactModal` (lines 15-67) and `contacts.editContactModal` (lines 86-...) are near-identical key trees — same `labels`, `placeholders`, `ariaLabels`, `buttons`, and `validations` — differing only in title and toast wording. Any single-surface copy edit has to be applied twice by hand. Consider hoisting the shared subtrees into one `contacts.contactModal` namespace (with only `title`/`toastMessages` specialised per mode) and pointing both `AddContactModal` and `EditContactModal` at it via the mode-specific prefix, so a future placeholder change lands in one place. Raising the root cause, not the individual string mismatch.

dulsara-manakal: That is out of this PR scope, and every other translation text is added according to the requirement.

dulsara-manakal: Contact can search with different things, not only by name, unlike company, so again, translation keys are added according to the requirement.

dulsara-manakal: BE has a global trimmer and will trim out the search keyword. And it's out of the scope of this PR.

frontend/src/community/common/assets/languages/english/crmModule.json
dulsara-manakal: This is intentional with the current version and will bring the company filtering in the future version.

frontend/src/community/common/assets/languages/english/crmModule.json
dulsara-manakal: Added a aria label

frontend/src/community/common/assets/languages/english/crmModule.json
dulsara-manakal: Everything was added according to the requirements.

frontend/src/community/common/assets/languages/english/crmModule.json
dulsara-manakal: This is out of scope of this PR, and this duplication will handle in CRM refactoring in the future.

thusala (approved):

akilarootcode (approved):

PR #1690, opened 2026-07-30
fix: add missing bottom layout of kanban view in CRM deal page

dulsara-manakal:
### Summary

- Fix the UI issue in the deal Kanban view
- Align with the Skapp PM setup

## Current layout
Screenshot 2026-08-05 105618

## Expected layout
Screenshot 2026-08-05 105913

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/frontend/src/community/crm/components/organisms/DealsKanbanBoard/DealsKanbanBoard.tsx
+++ b/frontend/src/community/crm/components/organisms/DealsKanbanBoard/DealsKanbanBoard.tsx
@@ -72,7 +72,7 @@ const DealsKanbanBoard: FC<DealsKanbanBoardProps> = ({
         onDragOver={handleDragOver}
         onDragEnd={handleDragEnd}
       >
-        <div className="flex items-stretch h-full overflow-x-auto overflow-y-hidden rounded-lg border border-secondary-accent p-2">
+        <div className="flex h-full overflow-x-auto overflow-y-hidden rounded-lg border border-secondary-accent p-2">
           {boardStages.map((stage) => {
             const stageDeals = stageMap.find((s) => s.stageId === stage.id);
             const deals = stageDeals?.deals ?? [];
--- a/frontend/src/community/crm/components/organisms/DealsKanbanBoard/DealsKanbanBoard.tsx
+++ b/frontend/src/community/crm/components/organisms/DealsKanbanBoard/DealsKanbanBoard.tsx
@@ -72,26 +72,27 @@ const DealsKanbanBoard: FC<DealsKanbanBoardProps> = ({
         onDragOver={handleDragOver}
         onDragEnd={handleDragEnd}
       >
-        <div className="flex items-stretch gap-4 h-full overflow-x-auto overflow-y-hidden rounded-lg border border-secondary-accent p-4">
+        <div className="flex items-stretch h-full overflow-x-auto overflow-y-hidden rounded-lg border border-secondary-accent p-2">
           {boardStages.map((stage) => {
             const stageDeals = stageMap.find((s) => s.stageId === stage.id);
             const deals = stageDeals?.deals ?? [];
 
             return (
-              <DealStageLane
-                key={stage.id}
-                stage={stage}
-                deals={deals}
-                isLoading={isLoading}
-                currentPage={stageDeals?.currentPage ?? 0}
-                hasNextPage={stageDeals?.hasNextPage ?? false}
-                totalCount={stageDeals?.totalCount ?? 0}
-                isOver={overStageId === stage.id}
-                searchKeyword={searchKeyword}
-                onDealClick={handleDealClick}
-                onAddDeal={handleAddDeal}
-                isAddDealDisabled={isCheckingCrmLimit}
-              />
+              <div key={stage.id} className="shrink-0 m-2">
+                <DealStageLane
+                  stage={stage}
+                  deals={deals}
+                  isLoading={isLoading}
+                  currentPage={stageDeals?.currentPage ?? 0}
+                  hasNextPage={stageDeals?.hasNextPage ?? false}
+                  totalCount={stageDeals?.totalCount ?? 0}
+                  isOver={overStageId === stage.id}
+                  searchKeyword={searchKeyword}
+                  onDealClick={handleDealClick}
+                  onAddDeal={handleAddDeal}
+                  isAddDealDisabled={isCheckingCrmLimit}
+                />
+              </div>
             );
           })}
         </div>
--- a/frontend/src/community/crm/components/organisms/DealsKanbanBoard/DealsKanbanBoard.tsx
+++ b/frontend/src/community/crm/components/organisms/DealsKanbanBoard/DealsKanbanBoard.tsx
@@ -72,7 +72,7 @@ const DealsKanbanBoard: FC<DealsKanbanBoardProps> = ({
         onDragOver={handleDragOver}
         onDragEnd={handleDragEnd}
       >
-        <div className="flex items-stretch gap-4 h-full overflow-x-auto py-2">
+        <div className="flex items-stretch gap-4 h-full overflow-x-auto overflow-y-hidden rounded-lg border border-secondary-accent p-4">
           {boardStages.map((stage) => {
             const stageDeals = stageMap.find((s) => s.stageId === stage.id);
             const deals = stageDeals?.deals ?? [];


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR is a one-line, styling-only change to the deals Kanban board container: it swaps `py-2` for `p-4` and adds `overflow-y-hidden rounded-lg border border-secondary-accent`. I verified `border-secondary-accent` is a legitimate Tailwind 4 theme token (`--color-secondary-accent` in frontend/styles/global.css:63) and not a hardcoded color, and I traced the height chain (DealsSection sets an explicit `calc(96vh - offsetTop)` height, DealsKanbanBoard root is `h-full overflow-hidden`, each DealStageLane is `h-full` with its own internal `overflow-y-auto`) — so the added padding/border do not cause clipping under `box-sizing: border-box`, and `overflow-y-hidden` does not break lane scrolling, infinite scroll (IntersectionObserver root is the viewport), or the dnd-kit DragOverlay. No type-safety, React Query, i18n, error-handling, or security concerns are touched. Only minor cosmetic/robustness observations below.

Found **12** new issue(s): 🔴 2 important, 🟡 4 suggestion(s), 🟣 6 nit(s)

frontend/src/community/crm/components/organisms/DealsKanbanBoard/DealsKanbanBoard.tsx
```diff
@@ -72,7 +72,7 @@ const DealsKanbanBoard: FC<DealsKanbanBoardProps> = ({
         onDragOver={handleDragOver}
         onDragEnd={handleDragEnd}
       >
-        <div className="flex items-stretch gap-4 h-full overflow-x-auto py-2">
+        <div className="flex items-stretch gap-4 h-full overflow-x-auto overflow-y-hidden rounded-lg border border-secondary-accent p-4">
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Trailing padding on a horizontally-scrolling flex container is unreliable across browsers. `p-4` gives 16px on all sides, but the inline-end padding of an `overflow-x-auto` flex container is historically dropped by some engines (notably WebKit/Safari), so when the user scrolls to the last stage the final lane can sit flush against the right border with no 16px gap, while the left side keeps its padding — an asymmetry that is exactly what this styling PR is trying to fix. If you need the gap guaranteed, either render a trailing spacer as the last flex child (`<div className="w-4 shrink-0" aria-hidden />`) or use a pseudo-element on the container (`after:block after:w-4 after:shrink-0 after:content-['']`) instead of relying on `padding-right`. Worth a quick check in Safari before merging.

ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`overflow-y-hidden` is a real behavior change, not a no-op. With only `overflow-x-auto` set previously, CSS computed `overflow-y` to `auto`, so any vertical overflow was still reachable by scrolling; now it is silently clipped. This is safe today because every lane is `h-full` (DealStageLaneHeader.tsx:34) and does its own vertical scrolling (DealStageLane.tsx:70), so nothing overflows the row — but it removes the safety net: if a lane ever loses `h-full`, or a non-scrolling element is added as a sibling of the lanes, content becomes invisible and unreachable with no scrollbar to hint at it. Consider adding a short comment noting that vertical scrolling is intentionally delegated to each lane, so a future change does not silently regress. (Note there is precedent for this pairing in ProjectDetailsLayout.tsx:135, so this is consistent with the codebase.)

ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

The new `rounded-lg border border-secondary-accent` on the board container duplicates chrome each lane already draws: DealStageLaneHeader.tsx:34 renders each stage as `rounded-lg ... outline-1 outline-secondary-accent`. The result is a nested frame in the same border color (outer border, 16px gutter, then per-lane outline of the same token), which reads as a double outline. Please confirm with design that the outer frame is intended here; if the goal was just to visually contain the board, using a background token (e.g. `bg-secondary-background`) or dropping the per-lane outline would give a cleaner single-frame result. Also note the horizontal scrollbar now renders inside the new rounded border, which slightly squares off the bottom corners on platforms with classic (non-overlay) scrollbars — worth eyeballing on Windows.

frontend/src/community/crm/components/organisms/DealsKanbanBoard/DealsKanbanBoard.tsx
```diff
@@ -72,7 +72,7 @@ const DealsKanbanBoard: FC<DealsKanbanBoardProps> = ({
         onDragOver={handleDragOver}
         onDragEnd={handleDragEnd}
```
ThinuwanW, line 73: 🤖 **Claude** · 🔴 **Important**

Neither drag mutation invalidates the board's React Query cache. `handleDragEnd` (useKanbanDrag.ts) optimistically writes the new ordering into Zustand and fires `reorderDealWithinStage` / `moveDealToStage`, but `useReorderDealWithinStage` / `useMoveDealBetweenStages` in `api/BoardApi.ts` only wire `onError` — there is no `onSuccess`/`onSettled` invalidation. I grepped the whole frontend: `crmBoardQueryKeys` never appears in an `invalidateQueries` call anywhere, and `useRefreshStageDeals` works around this by re-POSTing through `useFetchMoreStageDeals` instead of invalidating. Consequence: the cached `GET_DEALS_GROUPED_BY_STAGES` entry still holds the pre-drag ordering, so on the next mount of the board within gcTime (e.g. toggling Kanban→List→Kanban in DealsSection.tsx, which unmounts this component) `useBoardData`'s effect re-normalizes that stale response into the store and the dragged deal visibly snaps back to its old stage/position until the background refetch lands. Fix: take `useQueryClient()` in both mutation hooks and call `queryClient.invalidateQueries({ queryKey: crmBoardQueryKeys.GET_DEALS_GROUPED_BY_STAGES(...) })` (or invalidate on the shared `["crm-board-..."]` prefix) in `onSettled`, so the optimistic Zustand state and the server cache converge.

ThinuwanW, line 73: 🤖 **Claude** · 🔴 **Important**

Drag failures revert the board silently with no user-visible feedback. `useKanbanDrag`'s `rollback` is passed as the mutations' only `onError` handler and just restores `dragStartSnapshotRef.current` — no toast, so a user whose move failed sees the card jump back with no explanation and will assume the UI is glitching. The established pattern in this module is `const { setToastMessage } = useToast()` plus a translated title/description (see AddDealSidePanel.tsx:83 and TaskSidePanel.tsx:75); add that to the error path with new keys under `crmModule.deals.kanban` (that namespace currently contains only `addDealBtn`). Second, related defect in the same handler: `dragStartSnapshotRef` is a single shared ref that `handleDragStart` overwrites on every drag, so a late-arriving error from drag #1 rolls the board back to the snapshot captured at the start of drag #2 and silently discards drag #2's already-applied move. Capture the snapshot per mutation instead — pass it as mutation variables/context and read it from `onError(err, vars, context)` — rather than from one mutable ref.

frontend/src/community/crm/components/organisms/DealsKanbanBoard/DealsKanbanBoard.tsx
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

This div is the board's only horizontal scroll surface, and the PR makes that explicit by pinning overflow-y, but the element is not keyboard operable: no `tabIndex`, no role, no accessible name. A keyboard-only or screen-reader user has no way to scroll the board to off-screen stages independently of tabbing through cards, which is the standard axe `scrollable-region-focusable` / WCAG 2.1.1 failure — and it matters more here because the lanes are wide (`w-75` each) so several stages are typically outside the viewport. Add `tabIndex={0}` plus `role="region"` and `aria-label={translateText(["boardAriaLabel"])}`, wiring `useTranslator("crmModule", "deals", "kanban")` into this component and adding the key to crmModule.json (per the guidelines, accessibility labels must go through `t()`, not be hardcoded).

frontend/src/community/crm/components/organisms/DealsKanbanBoard/DealsKanbanBoard.tsx
```diff
@@ -72,7 +72,7 @@ const DealsKanbanBoard: FC<DealsKanbanBoardProps> = ({
...
       >
-        <div className="flex items-stretch gap-4 h-full overflow-x-auto py-2">
+        <div className="flex items-stretch gap-4 h-full overflow-x-auto overflow-y-hidden rounded-lg border border-secondary-accent p-4">
           {boardStages.map((stage) => {
```
ThinuwanW, line 76: 🤖 **Claude** · 🟡 **Suggestion**

The board has no empty state and no error state, and the new border/radius on the container turns that from invisible into a visible defect. `boardStages` comes from `initData?.stages ?? []` in useBoardData.ts:32, so it is `[]` in three cases: during the initial `useGetBoardInitData` fetch, when the tenant has no deal stages configured, and when the request fails — `useBoardData` returns only `isLoading` and drops `isError` entirely, so a failed init-data call renders no message and fires no toast. Because `isLoading` is only forwarded into each lane's skeleton, zero stages means zero children, and with the added chrome the user now sees a bare bordered rectangle rather than nothing. Fix: surface `isError` (and ideally `isFetched`) from `useBoardData`, render a skeleton lane row while init data is pending, and render a translated empty-board / error message (new keys under `crmModule.deals.kanban`) when `boardStages.length === 0`.

frontend/src/community/crm/components/organisms/DealsKanbanBoard/DealsKanbanBoard.tsx
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

The `DragOverlay` block duplicates the eight-prop `DealCard` mapping that already exists verbatim in DraggableDealCard.tsx (id/title/contactName/companyName ?? undefined/owner/amount ?? ""/priority/taskCount). Adding or renaming a field on `CrmBoardDealSliceType` requires editing both sites, and they will silently drift — exactly the duplication the guidelines say to extract. Pull a `toDealCardProps(deal: CrmBoardDealSliceType): DealCardProps` helper (or a small `BoardDealCard` wrapper) into the crm utils/molecules and use it in both places. While there: the overlay card is a non-interactive drag preview but `DealCard` unconditionally renders its title with `role="button" tabIndex={0}` (DealCard.tsx:70-71), so the overlay injects a focusable, clickable-looking node with no `onClick`; pass an `aria-hidden`/non-interactive variant for the overlay.

ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

`handleAddDeal` and `handleDealClick` are re-created on every render and passed to every `DealStageLane`, which is not wrapped in `React.memo` and does real work per render (a `deals.reduce`, `useLoadMoreStageDeals`, `useInfiniteScroll`, a `useMemo` over dealIds). So any board re-render — including every `overStageId` state change during a drag, which fires on each `onDragOver` — re-renders all lanes. Both handlers read their store actions via `useCrmStore.getState()`, so they have no reactive dependencies: wrap them in `useCallback` with `[guardCrmCreate]` / `[]` and wrap `DealStageLane` in `memo`.

ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

Adjacent to the lane chrome this PR is restyling, DealStageLaneHeader has a broken accessible name: the `<section>` sets `aria-labelledby={`crm-stage-${stage.id}`}` (line 39) but the heading it means to reference is `id={String(stage.id)}` (line 48). The reference resolves to nothing, so every lane region is unnamed to screen readers. The bare numeric id is also unsafe — DraggableDealCard.tsx:26 sets `id={String(deal.id)}`, so a deal and a stage sharing a numeric id produce duplicate DOM ids on the same page. Change the heading to `id={`crm-stage-${stage.id}`}` (and consider prefixing the draggable card id similarly).

frontend/src/community/crm/components/organisms/DealsKanbanBoard/DealsKanbanBoard.tsx
```diff
@@ -72,7 +72,7 @@ const DealsKanbanBoard: FC<DealsKanbanBoardProps> = ({
...
             const stageDeals = stageMap.find((s) => s.stageId === stage.id);
```
ThinuwanW, line 77: 🤖 **Claude** · 🟣 **Nit**

`stageMap.find(...)` runs inside the render-time `.map()` over `boardStages`, so lane lookup is O(stages²) and repeats on every board render — including on each `onDragOver`-driven `overStageId` update mid-drag. Stage counts are small today, but this is trivially avoidable: build `const dealsByStage = useMemo(() => new Map(stageMap.map((s) => [s.stageId, s])), [stageMap])` and index it, which also reads more clearly than the chain of `?? 0` / `?? false` fallbacks on the result.

frontend/src/community/crm/components/organisms/DealsKanbanBoard/DealsKanbanBoard.tsx
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`items-stretch` on this line is a no-op — `align-items: stretch` is the flex default and nothing in the ancestor chain overrides it. Since this line is being rewritten anyway, drop the class to keep the utility list meaningful (each lane already forces its own height via `h-full` in DealStageLaneHeader.tsx:34, so removal is behaviour-neutral).

dulsara-manakal: Align with skapp-pm. Added margin to each lane instead of adding padding to the parent.

dulsara-manakal: This won't happen with current approach.

dulsara-manakal: Implement with design requirements to align to  skapp-pm

dulsara-manakal: Removed `items-stretch`

frontend/src/community/crm/components/organisms/DealsKanbanBoard/DealsKanbanBoard.tsx
dulsara-manakal: Currently no bug has been found, as this review comment describes, and also this is out of scope of this PR since this PR is for minor UI change.

dulsara-manakal: Not going to change this behaviour since the PR is only focusing on UI changes.

frontend/src/community/crm/components/organisms/DealsKanbanBoard/DealsKanbanBoard.tsx
dulsara-manakal: This Kanban view is aligned with the existing Skapp PM structure, and the keyboard navigation in scrolling and drag-and-drop is working as required.

frontend/src/community/crm/components/organisms/DealsKanbanBoard/DealsKanbanBoard.tsx
dulsara-manakal: The board is implemented according to the requirements .

frontend/src/community/crm/components/organisms/DealsKanbanBoard/DealsKanbanBoard.tsx
dulsara-manakal: The initial implementation is done according to the requirements. and nothing changed with the behaviour in this PR, only the missing UI change was added.

dulsara-manakal: The initial implementation is done according to the requirements. and nothing changed with the behaviour in this PR, only the missing UI change was added.

dulsara-manakal: First, this is not even touched file in this PR and is obviously out of scope, but it was confirmed with the code, and `aria-labelledby={crm-stage-${stage.id}}` is just for accessibility and screen reader purposes, and `id={String(deal.id)}` is for DOM identification.

frontend/src/community/crm/components/organisms/DealsKanbanBoard/DealsKanbanBoard.tsx
dulsara-manakal: This is pre-existing logic outside the scope of this PR. Without a clear context on why it was implemented this way, it'd rather not risk changing behaviour here.

thusala (approved):

akilarootcode (approved):

PR #1691, opened 2026-07-31
Feat/policy leave type BE

hasalarootcode:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds full CRUD (create, paginated search, get-by-id, patch update, activate/deactivate) for policy leave types, closely mirroring the existing LeavePolicy feature, with good integration-test coverage including role-based access checks. Two issues stand out: the PATCH update path skips the emoji/color validation that create enforces (blank values can be persisted), and the new paginated search endpoint passes unvalidated `page`/`size` query params straight into `PageRequest.of`, which yields a 500 for negative pages and an unbounded query for huge sizes. The remainder are consistency/duplication concerns — no bean validation on the new request DTOs, duplicated request DTOs and message constants, and no DB-level uniqueness backing the application-level name check.

Found **16** new issue(s): 🔴 2 important, 🟡 10 suggestion(s), 🟣 4 nit(s)

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveTypeServiceImpl.java
```diff
@@ -34,4 +48,181 @@ public ResponseEntityDto getPolicyLeaveTypes() {
...
+			PolicyLeaveTypeValidationUtil.validateName(policyLeaveTypeUpdateRequestDto.getName());
+
+			if (policyLeaveTypeDao.existsByNameIgnoreCaseAndIdNot(policyLeaveTypeUpdateRequestDto.getName(), id)) {
+				throw new ModuleException(LeaveMessageConstant.LEAVE_ERROR_POLICY_LEAVE_TYPE_ALREADY_EXISTS);
+			}
+
+			policyLeaveType.setName(policyLeaveTypeUpdateRequestDto.getName());
+		}
+
+		if (policyLeaveTypeUpdateRequestDto.getEmojiCode() != null) {
```
ThinuwanW, line 120: 🤖 **Claude** · 🔴 **Important**

The update path validates `name` (via `PolicyLeaveTypeValidationUtil.validateName`) and re-runs `validateAttachmentSetup`, but never calls `validateAppearance(...)` or `validateMinDuration(...)`. Because the null-check only guards against *absent* fields, a request body of `{"emojiCode": "   "}` or `{"colorCode": ""}` passes the `!= null` gate and is persisted, producing a leave type whose appearance fields are blank — a state `addPolicyLeaveType` explicitly rejects with LEAVE_ERROR_POLICY_LEAVE_TYPE_EMOJI_CODE_REQUIRED / _COLOR_CODE_REQUIRED. Fix: inside each non-null branch, validate the supplied value before assigning, e.g. in the emojiCode branch throw LEAVE_ERROR_POLICY_LEAVE_TYPE_EMOJI_CODE_REQUIRED when the value is blank, and likewise for colorCode; alternatively split `validateAppearance` into `validateEmojiCode(String)` / `validateColorCode(String)` and call them from both create and update. Note `minDuration` is safe today (a null JSON value can't reach the setter) but the same shape applies. Worth adding tests mirroring `updatePolicyLeaveType_BlankName_ReturnsBadRequest` for blank emojiCode/colorCode.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveTypeServiceImpl.java
```diff
@@ -34,4 +48,181 @@ public ResponseEntityDto getPolicyLeaveTypes() {
...
+		return new ResponseEntityDto(false,
+				leaveMapper.policyLeaveTypeToPolicyLeaveTypeDetailResponseDto(policyLeaveType));
+	}
+
+	@Override
+	@Transactional(readOnly = true)
+	public ResponseEntityDto searchPolicyLeaveTypes(PolicyLeaveTypeFilterDto policyLeaveTypeFilterDto) {
+		log.info("searchPolicyLeaveTypes: execution started");
+
+		Pageable pageable = PageRequest.of(policyLeaveTypeFilterDto.getPage(), policyLeaveTypeFilterDto.getSize());
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`PageRequest.of(policyLeaveTypeFilterDto.getPage(), policyLeaveTypeFilterDto.getSize())` uses raw client-supplied values from `PolicyLeaveTypeFilterDto`, which carries no bounds. `GET /v1/leave/policy-leave-types/search?page=-1` (or `size=0`) makes `PageRequest.of` throw `IllegalArgumentException`; GlobalExceptionHandler has no handler for it, so it falls through to `@ExceptionHandler(Exception.class)` and returns HTTP 500 with the raw exception message instead of a 400. Separately, `size` has no upper bound, so `?size=1000000` issues an effectively unbounded query against `lv_leave_type`. Fix: annotate the filter DTO (`@Min(0)` on `page`, `@Min(1) @Max(100)` on `size`) and add `@Valid` to the controller parameter so a bad value becomes a 422 BindException, or clamp defensively in the service (`Math.max(0, page)`, `Math.clamp(size, 1, MAX_PAGE_SIZE)`) before calling `PageRequest.of`.

backend/src/main/java/com/skapp/community/leaveplanner/controller/v1/PolicyLeaveTypeController.java
```diff
@@ -30,4 +37,60 @@ public ResponseEntity<ResponseEntityDto> getPolicyLeaveTypes() {
...
+		ResponseEntityDto response = policyLeaveTypeService.getPolicyLeaveTypeById(id);
+		return new ResponseEntity<>(response, HttpStatus.OK);
+	}
+
+	@Operation(summary = "Create a new policy leave type",
+			description = "Creates a new leave type that can be used when creating leave policies")
+	@PostMapping
+	@PreAuthorize("hasAnyRole('ROLE_LEAVE_ADMIN')")
+	public ResponseEntity<ResponseEntityDto> addPolicyLeaveType(
+			@RequestBody PolicyLeaveTypeRequestDto policyLeaveTypeRequestDto) {
```
ThinuwanW, line 56: 🤖 **Claude** · 🟡 **Suggestion**

`@RequestBody` on both `addPolicyLeaveType` and `updatePolicyLeaveType` is missing `@Valid`, and the three new request DTOs (`PolicyLeaveTypeRequestDto`, `PolicyLeaveTypeUpdateRequestDto`, `PolicyLeaveTypeFilterDto`) carry no bean-validation annotations at all — all required-field checking is deferred to `PolicyLeaveTypeValidationUtil` in the service. The sibling `LeaveTypeController` in this same package already does it the recommended way (`@Valid @RequestBody LeaveTypeRequestDto`, with `@NotNull` on the DTO fields). Suggest adding `@NotBlank` on `name`/`emojiCode`/`colorCode`, `@Size(max = PolicyLeaveTypeConstant.MAX_NAME_LENGTH)` on `name`, `@NotNull` on `minDuration` for the create DTO, and `@Valid` on the controller params, keeping only the genuinely cross-field rule (`validateAttachmentSetup`) in the service.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveTypeServiceImpl.java
```diff
@@ -34,4 +48,181 @@ public ResponseEntityDto getPolicyLeaveTypes() {
...
+		log.info("addPolicyLeaveType: execution started");
+
+		PolicyLeaveTypeValidationUtil.validateName(policyLeaveTypeRequestDto.getName());
+		PolicyLeaveTypeValidationUtil.validateAppearance(policyLeaveTypeRequestDto.getEmojiCode(),
+				policyLeaveTypeRequestDto.getColorCode());
+		PolicyLeaveTypeValidationUtil.validateMinDuration(policyLeaveTypeRequestDto.getMinDuration());
+		PolicyLeaveTypeValidationUtil.validateAttachmentSetup(policyLeaveTypeRequestDto.getIsAttachment(),
+				policyLeaveTypeRequestDto.getIsAttachmentMust());
+
+		if (policyLeaveTypeDao.existsByNameIgnoreCase(policyLeaveTypeRequestDto.getName())) {
```
ThinuwanW, line 77: 🤖 **Claude** · 🟡 **Suggestion**

Name uniqueness is enforced only in application code (`existsByNameIgnoreCase` / `existsByNameIgnoreCaseAndIdNot`), and the `lv_leave_type` table created in `community/db/changelog/ddl-script/leave-ddl-script-v1-create-leave-policy-tables.sql` has no unique index on `name` (the column is `text`, so a plain UNIQUE isn't even possible without a prefix length). Two concurrent POSTs with the same name both pass the check and both commit, permanently breaking the invariant the API advertises. Consider adding a Liquibase changeset introducing `name varchar(100)` (matching `PolicyLeaveTypeConstant.MAX_NAME_LENGTH`) plus a unique index, and translating the resulting constraint violation into LEAVE_ERROR_POLICY_LEAVE_TYPE_ALREADY_EXISTS.

backend/src/main/java/com/skapp/community/leaveplanner/payload/request/PolicyLeaveTypeUpdateRequestDto.java
```diff
@@ -0,0 +1,27 @@
+package com.skapp.community.leaveplanner.payload.request;
+
+import com.skapp.community.leaveplanner.type.LeaveDuration;
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+public class PolicyLeaveTypeUpdateRequestDto {
```
ThinuwanW, line 9: 🤖 **Claude** · 🟡 **Suggestion**

`PolicyLeaveTypeUpdateRequestDto` is field-for-field identical to `PolicyLeaveTypeRequestDto` (same 8 fields, same types, same annotations) — the two classes differ only in name. Two DTOs with identical shape will drift apart accidentally the first time someone adds a field to one. Either collapse them into a single `PolicyLeaveTypeRequestDto` used by both endpoints, or (preferably, since you plan to add bean validation) keep them separate but make the difference meaningful: put `@NotBlank`/`@NotNull` on the create DTO and leave the update DTO's fields optional to express PATCH semantics.

backend/src/main/java/com/skapp/community/leaveplanner/constant/LeaveMessageConstant.java
```diff
@@ -87,7 +87,19 @@ public enum LeaveMessageConstant implements MessageConstant {
 	LEAVE_ERROR_LEAVE_POLICY_NOT_ACCRUAL("api.error.leave.leave-policy-not-accrual"),
 	LEAVE_ERROR_EMPLOYEE_LEAVE_POLICY_NOT_FOUND("api.error.leave.employee-leave-policy-not-found"),
 	LEAVE_ERROR_LEAVE_POLICY_ALREADY_INACTIVE("api.error.leave.leave-policy-already-inactive"),
-	LEAVE_ERROR_LEAVE_POLICY_ALREADY_ACTIVE("api.error.leave.leave-policy-already-active");
+	LEAVE_ERROR_LEAVE_POLICY_ALREADY_ACTIVE("api.error.leave.leave-policy-already-active"),
+	LEAVE_ERROR_POLICY_LEAVE_TYPE_NOT_EXIST("api.error.leave.policy-leave-type-not-exist"),
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Two of the new constants duplicate existing entries in this same enum. (1) `LEAVE_ERROR_POLICY_LEAVE_TYPE_NOT_EXIST` resolves to "Leave type not found", which is byte-identical to the existing `api.error.leave.leave-type-not-found`, and its name is one word away from the pre-existing `LEAVE_ERROR_POLICY_LEAVE_TYPE_NOT_FOUND` ("Leave type not found or inactive", used by LeavePolicyServiceImpl:57) — two near-identically-named constants with subtly different semantics in the same enum is an easy source of picking the wrong one. (2) `LEAVE_ERROR_POLICY_LEAVE_TYPE_UNABLE_TO_MAKE_ATTACHMENT_MANDATORY` has exactly the same message text as the existing `LEAVE_ERROR_LEAVE_TYPE_UNABLE_TO_MAKE_ATTACHMENT_MANDATORY`. Suggest reusing the existing constants where the semantics match, or renaming the new one to something unambiguous (e.g. `LEAVE_ERROR_POLICY_LEAVE_TYPE_ID_NOT_FOUND`) and giving it distinct user-facing text so support can tell the two failure modes apart from the message alone.

backend/src/main/java/com/skapp/community/leaveplanner/util/PolicyLeaveTypeValidationUtil.java
```diff
@@ -0,0 +1,43 @@
...
+		if (name.length() > PolicyLeaveTypeConstant.MAX_NAME_LENGTH) {
+			throw new ModuleException(LeaveMessageConstant.LEAVE_ERROR_POLICY_LEAVE_TYPE_NAME_MAX_LENGTH_EXCEEDED);
+		}
+	}
+
+	public static void validateAppearance(String emojiCode, String colorCode) {
+		if (emojiCode == null || emojiCode.isBlank()) {
+			throw new ModuleException(LeaveMessageConstant.LEAVE_ERROR_POLICY_LEAVE_TYPE_EMOJI_CODE_REQUIRED);
+		}
+		if (colorCode == null || colorCode.isBlank()) {
```
ThinuwanW, line 29: 🤖 **Claude** · 🟡 **Suggestion**

`colorCode` is only checked for non-blank; any string is accepted and stored (the tests only ever send `#FFC107`). The codebase already has `com.skapp.community.common.util.Validation.isValidThemeColor(String)` backed by `ValidationConstant.HEXA_DECIMAL_VALIDATION_PATTERN` (`^#([A-Fa-f0-9]{6,8})$`) for exactly this. Suggest calling it here and throwing a dedicated invalid-format message constant, so a value like `red` or `javascript:...` can't reach the frontend renderer. Same reasoning applies to `emojiCode` if a `U+XXXX` format is expected.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveTypeServiceImpl.java
```diff
@@ -34,4 +48,181 @@ public ResponseEntityDto getPolicyLeaveTypes() {
...
+	public ResponseEntityDto deactivatePolicyLeaveType(Long id) {
+		log.info("deactivatePolicyLeaveType: execution started");
+
+		PolicyLeaveType policyLeaveType = getPolicyLeaveType(id);
+
+		if (!Boolean.TRUE.equals(policyLeaveType.getIsActive())) {
+			throw new ModuleException(LeaveMessageConstant.LEAVE_ERROR_POLICY_LEAVE_TYPE_ALREADY_INACTIVE);
+		}
+
+		policyLeaveType.setIsActive(Boolean.FALSE);
```
ThinuwanW, line 172: 🤖 **Claude** · 🟡 **Suggestion**

Deactivating a leave type only flips `isActive` on the type itself. Leave policies already referencing it keep `LeavePolicyStatus.ACTIVE`, and `EmployeeLeavePolicyServiceImpl` (line 54) gates assignment purely on `policy.getStatus() != ACTIVE` — it never inspects `policy.getLeaveType().getIsActive()`. So after deactivation those policies remain fully assignable to employees, which contradicts the intent implied by the endpoint. `addLeavePolicy` correctly uses `findByIdAndIsActiveTrue`, so the gap is only on the assignment path. Either cascade-deactivate (or block deactivation of) leave types that still have active policies, or add an `isActive` check on the leave type in the assignment flow. If leaving policies active is the intended behaviour, please state it in the `@Operation` description, which currently only says "no longer available for policy creation".

backend/src/main/java/com/skapp/community/leaveplanner/repository/PolicyLeaveTypeDao.java
```diff
@@ -2,16 +2,22 @@
...
+import org.springframework.data.jpa.repository.JpaSpecificationExecutor;
 import org.springframework.stereotype.Repository;
 
 import java.util.List;
 import java.util.Optional;
 
 @Repository
-public interface PolicyLeaveTypeDao extends JpaRepository<PolicyLeaveType, Long> {
+public interface PolicyLeaveTypeDao extends JpaRepository<PolicyLeaveType, Long>,
+		JpaSpecificationExecutor<PolicyLeaveType>, PolicyLeaveTypeRepository {
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`JpaSpecificationExecutor<PolicyLeaveType>` is added but never used — all filtering goes through the new `PolicyLeaveTypeRepository`/`PolicyLeaveTypeRepositoryImpl` Criteria fragment, and grep shows no `Specification` usage against this Dao. It mirrors `LeavePolicyDao`, so it may be intentional boilerplate, but it adds an unused Spring Data fragment. Consider dropping it unless a Specification-based query is coming.

backend/src/main/java/com/skapp/community/leaveplanner/util/PolicyLeaveTypeValidationUtil.java
```diff
@@ -0,0 +1,43 @@
...
+import com.skapp.community.leaveplanner.constant.LeaveMessageConstant;
+import com.skapp.community.leaveplanner.constant.PolicyLeaveTypeConstant;
+import com.skapp.community.leaveplanner.type.LeaveDuration;
+import lombok.experimental.UtilityClass;
+
+@UtilityClass
+public class PolicyLeaveTypeValidationUtil {
+
+	public static void validateName(String name) {
+		if (name == null || name.isBlank()) {
```
ThinuwanW, line 14: 🤖 **Claude** · 🟣 **Nit**

`name` is neither trimmed before validation nor before persistence, so `"  Annual  "` passes `validateName`, does not collide with an existing `"Annual"` in `existsByNameIgnoreCase`, and is stored with the surrounding whitespace — producing two visually identical leave types. Consider normalising with `name.trim()` in the service before the uniqueness check and the `setName` call (also makes the `MAX_NAME_LENGTH` check measure the real name).

backend/src/main/java/com/skapp/community/leaveplanner/payload/request/PolicyLeaveTypeRequestDto.java
```diff
@@ -0,0 +1,27 @@
...
+@Setter
+public class PolicyLeaveTypeRequestDto {
+
+	private String name;
+
+	private String emojiCode;
+
+	private String colorCode;
+
+	private LeaveDuration minDuration;
```
ThinuwanW, line 17: 🤖 **Claude** · 🟣 **Nit**

`minDuration` is a `LeaveDuration` enum bound from the JSON body. An unknown value (`"minDuration": "QUARTER_DAY"`) or any malformed body raises `HttpMessageNotReadableException`, which GlobalExceptionHandler does not handle specifically — it falls through to `@ExceptionHandler(Exception.class)` and returns HTTP 500 plus the raw Jackson message, rather than a 400. This is a pre-existing, app-wide gap rather than something this PR introduces, but these are new client-facing endpoints that accept an enum, so it's newly reachable. A `@ExceptionHandler(HttpMessageNotReadableException.class)` returning 400 in GlobalExceptionHandler would fix it once for every endpoint.

backend/src/test/java/com/skapp/community/leaveplanner/controller/v1/PolicyLeaveTypeControllerIntegrationTest.java
```diff
@@ -56,27 +83,568 @@ private String user2Token() {
...
+		@Test
+		@DisplayName("Leave admin can search leave types regardless of active status")
+		@Sql(statements = { SEED_LEAVE_TYPE, SEED_INACTIVE_LEAVE_TYPE, SEED_SECOND_LEAVE_TYPE })
+		void searchPolicyLeaveTypes_LeaveAdmin_ReturnsAllTypes() throws Exception {
+			mvc.perform(get(SEARCH_ENDPOINT).accept(MediaType.APPLICATION_JSON)
+				.with(SecurityTestUtils.bearerToken(leaveAdminToken())))
+				.andDo(print())
+				.andExpect(status().isOk())
+				.andExpect(jsonPath(STATUS_PATH).value(STATUS_SUCCESSFUL))
+				.andExpect(jsonPath("$.results[0].items", hasSize(3)))
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

The search tests assert absolute totals (`items` hasSize(3), `totalItems` 3, `totalPages` 1/2) that are only correct if `lv_leave_type` is otherwise empty at test time. The tests are `@Transactional` so they roll back, but any future seed data or Liquibase-inserted default leave type will silently break several of these at once. Consider filtering by `searchKeyword="Policy"` in these assertions so the counts are scoped to the rows the test itself seeds, leaving only the dedicated pagination test dependent on the exact total.

backend/src/main/java/com/skapp/community/leaveplanner/repository/PolicyLeaveTypeDao.java
```diff
@@ -2,16 +2,22 @@
...
 
 	Optional<PolicyLeaveType> findByIdAndIsActiveTrue(Long id);
 
 	List<PolicyLeaveType> findAllByIsActive(boolean isActive);
 
+	boolean existsByNameIgnoreCase(String name);
+
```
ThinuwanW, line 19: 🤖 **Claude** · 🟡 **Suggestion**

Missing paired Liquibase change for the new name-uniqueness lookups. In community/db/changelog/ddl-script/leave-ddl-script-v1-create-leave-policy-tables.sql, lv_leave_type declares `name`, `emoji_code` and `color_code` as MySQL `TEXT`, and the PolicyLeaveType entity declares them with no `length`. Two consequences for the new write endpoints: (1) a `TEXT` column cannot take a plain UNIQUE index in MySQL — it requires a prefix length — so the uniqueness guarantee the new `existsByNameIgnoreCase`/`existsByNameIgnoreCaseAndIdNot` checks imply cannot be backed by a DB constraint as the schema stands; (2) nothing bounds the stored size, so an admin can persist arbitrarily large name/emoji/colour values regardless of any application-level length check. Fix: add a Liquibase changeset converting `lv_leave_type.name` to `VARCHAR(n)` (matching whatever MAX_NAME_LENGTH the new PolicyLeaveTypeConstant defines, following the `leave_type.name varchar(20) unique` precedent) plus `VARCHAR` bounds on emoji_code/color_code, then add the unique index on name. Also mirror the bound on the entity with `@Column(name = "name", nullable = false, length = n)`.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveTypeServiceImpl.java
```diff
@@ -34,4 +48,181 @@ public ResponseEntityDto getPolicyLeaveTypes() {
 		return new ResponseEntityDto(false, new PolicyLeaveTypesResponseDto(leaveTypes));
 	}
```
ThinuwanW, line 64: 🤖 **Claude** · 🟡 **Suggestion**

The pre-existing `getPolicyLeaveTypes()` (kept unchanged at this location) calls `policyLeaveTypeDao.findAllByIsActive(true)` with no pagination, and the PR adds a second, paginated+filtered list path over the same resource. That leaves two list endpoints on `/v1/leave/policy-leave-types` with different contracts: the old one returns a `PolicyLeaveTypesResponseDto` wrapper of 4-field `PolicyLeaveTypeResponseDto`s and is unbounded; the new one returns a `PageDto`. Per the repository-patterns guideline (`unbounded queries without pagination`) and the endpoint-symmetry check, decide on one: either drop the legacy method and have callers use the paginated endpoint with a large page size, or keep it explicitly as a lightweight dropdown lookup and document that in its `@Operation` description so the split is deliberate rather than accidental drift.

backend/src/main/java/com/skapp/community/leaveplanner/payload/response/PolicyLeaveTypeDetailResponseDto.java
```diff
@@ -0,0 +1,31 @@
+package com.skapp.community.leaveplanner.payload.response;
```
ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Response-DTO duplication (distinct from the request-DTO duplication already raised on PolicyLeaveTypeUpdateRequestDto). `PolicyLeaveTypeResponseDto` already exists in this same package with id/name/emojiCode/colorCode, and this new detail DTO is a superset of those same four fields plus the configuration flags. Two near-identical response shapes for one entity means every future field addition has to be made twice and it is easy for them to diverge (as `LeaveTypeResponseDto` vs `LeaveTypeBasicDetailsResponseDto` already have). Prefer either making `PolicyLeaveTypeResponseDto` the single response type and letting the list endpoint return it, or having the detail DTO extend the summary DTO so the shared four fields are declared once.

backend/src/main/java/com/skapp/community/leaveplanner/controller/v1/PolicyLeaveTypeController.java
```diff
@@ -30,4 +37,60 @@ public ResponseEntity<ResponseEntityDto> getPolicyLeaveTypes() {
 		return new ResponseEntity<>(response, HttpStatus.OK);
 	}
 
+	@Operation(summary = "Search policy leave types",
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Confirm the role set on the new write endpoints is intentional. The existing GET on this controller is `@PreAuthorize("hasAnyRole('ROLE_LEAVE_ADMIN')")` and `ROLE_SUPER_ADMIN` is not implied by it — the codebase grants it explicitly where it applies (e.g. WorkLocationController uses `hasAnyRole('ROLE_SUPER_ADMIN','ROLE_ATTENDANCE_ADMIN','ROLE_PEOPLE_ADMIN')`). Leave-type configuration is org-level setup, so a super admin currently cannot create, update or deactivate a policy leave type. Either add `ROLE_SUPER_ADMIN` to the new endpoints' `hasAnyRole` list, or note in the PR that leave-type administration is deliberately scoped to ROLE_LEAVE_ADMIN only. Also check the read endpoints match the sibling `LeavePolicyController.getAllLeavePolicies`, which allows `ROLE_LEAVE_ADMIN, ROLE_PEOPLE_ADMIN` — an unexplained mismatch between the two read paths would be drift.

backend/src/main/java/com/skapp/community/leaveplanner/controller/v1/PolicyLeaveTypeController.java
hasalarootcode: We keep request validation in a service-level validation util rather than bean annotations on the DTO — that's the pattern across the modules here, so `PolicyLeaveTypeValidationUtil` is deliberate rather than a gap in coverage. `LeaveTypeController` predates that convention. The real issue you've spotted is that the update path doesn't re-check the appearance fields; fixing that in the util.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveTypeServiceImpl.java
hasalarootcode: We don't add DB unique constraints in this codebase — uniqueness is enforced by the service-level `existsBy...` check, same as the other modules. So the check-then-act race is real but accepted; two admins creating the same leave type name in the same instant is not a workload we're designing against here.

backend/src/main/java/com/skapp/community/leaveplanner/payload/request/PolicyLeaveTypeUpdateRequestDto.java
hasalarootcode: They're identical because the validation lives in the util rather than in annotations, so there's nothing to differentiate them with. I'd rather keep them separate — the create and update payloads have genuinely different semantics (all-required vs partial PATCH) and collapsing them means the Swagger docs stop expressing that. Happy to collapse to one DTO if you'd prefer fewer classes — say which way you want it.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveTypeServiceImpl.java
hasalarootcode: This one needs a product call before I change anything. Deactivating a leave type currently leaves existing policies on it assignable, which as you say contradicts what the endpoint description implies. Should deactivation cascade — block the leave type from being deactivated while active policies reference it, or deactivate those policies with it? I don't want to pick one; let me know which and I'll implement it.

backend/src/main/java/com/skapp/community/leaveplanner/payload/request/PolicyLeaveTypeRequestDto.java
hasalarootcode: Agreed it's a gap, but `GlobalExceptionHandler` is shared by every endpoint in the app, so adding a handler there from this PR changes error responses well outside this feature's blast radius. Better as its own change with its own testing — shall I raise a separate ticket for it?

backend/src/main/java/com/skapp/community/leaveplanner/repository/PolicyLeaveTypeDao.java
hasalarootcode: Two house conventions cut against this: DDL columns are `text` rather than `varchar` here, and we don't use DB unique constraints — uniqueness is a service-level `existsByNameIgnoreCase` check. On the unbounded size point, `validateName` already rejects anything over `PolicyLeaveTypeConstant.MAX_NAME_LENGTH`, so oversized names can't reach the column through the API. Leaving the schema as it is.

backend/src/main/java/com/skapp/community/leaveplanner/payload/response/PolicyLeaveTypeDetailResponseDto.java
hasalarootcode: Fair point on the duplication. `PolicyLeaveTypeResponseDto` is the 4-field shape the existing dropdown lookup returns and the FE is already consuming it, so making the detail DTO extend it would work, but it changes the inheritance on a response type the FE branch is built against. Want me to do that here, or leave it until the FE PR is merged so the two don't move at once?

backend/src/main/java/com/skapp/community/leaveplanner/controller/v1/PolicyLeaveTypeController.java
hasalarootcode: `ROLE_LEAVE_ADMIN` only is intentional — we don't add `SUPER_ADMIN` to endpoints a module admin role already covers, and leave-type configuration sits squarely in the leave module.

The read-path mismatch is worth checking though: `LeavePolicyController.getAllLeavePolicies` allows `PEOPLE_ADMIN` because policies show up on employee profiles. Do people admins need to read leave types too, or is `LEAVE_ADMIN` right for these? If they do I'll widen the three GETs.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveTypeServiceImpl.java
hasalarootcode: Split `validateAppearance` into `validateEmojiCode(String)` / `validateColorCode(String)` and called them from both the create path and inside the non-null branches on update, so a blank `emojiCode` or `colorCode` is now rejected there too. Added `updatePolicyLeaveType_BlankEmojiCode_ReturnsBadRequest` and `_BlankColorCode_ReturnsBadRequest`.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveTypeServiceImpl.java
hasalarootcode: Added `PolicyLeaveTypeValidationUtil.validatePagination(page, size)`, called before `PageRequest.of`, with `MIN_PAGE_NUMBER`/`MIN_PAGE_SIZE`/`MAX_PAGE_SIZE` (0/1/100) in `PolicyLeaveTypeConstant`. Negative page and out-of-range size now return 400 rather than 500. Kept it in the util rather than as DTO annotations to match how the rest of the module validates. Three tests cover `page=-1`, `size=0` and `size=1000000`.

backend/src/main/java/com/skapp/community/leaveplanner/constant/LeaveMessageConstant.java
hasalarootcode: Renamed `LEAVE_ERROR_POLICY_LEAVE_TYPE_NOT_EXIST` to `LEAVE_ERROR_POLICY_LEAVE_TYPE_ID_NOT_FOUND` and gave it distinct text ("No leave type exists with the given id") so it no longer reads identically to `api.error.leave.leave-type-not-found`. Dropped the duplicate `..._POLICY_LEAVE_TYPE_UNABLE_TO_MAKE_ATTACHMENT_MANDATORY` and reused the existing `LEAVE_ERROR_LEAVE_TYPE_UNABLE_TO_MAKE_ATTACHMENT_MANDATORY`. Checked the FE — it only maps `LEAVE_ERROR_POLICY_LEAVE_TYPE_ALREADY_EXISTS`, so neither change breaks it.

backend/src/main/java/com/skapp/community/leaveplanner/util/PolicyLeaveTypeValidationUtil.java
hasalarootcode: `validateColorCode` now calls `Validation.isValidThemeColor` and throws a new `LEAVE_ERROR_POLICY_LEAVE_TYPE_COLOR_CODE_INVALID`, so `red` and friends are rejected. Left `emojiCode` as a non-blank check only — there's no existing validator for the `U+XXXX` form and I'd rather not invent a pattern the FE emoji picker might not match.

backend/src/main/java/com/skapp/community/leaveplanner/repository/PolicyLeaveTypeDao.java
hasalarootcode: Removed.

backend/src/main/java/com/skapp/community/leaveplanner/util/PolicyLeaveTypeValidationUtil.java
hasalarootcode: Trimmed. The service now normalises `name` before `validateName`, the `existsByNameIgnoreCase` check and `setName`, on both create and update, so `"  Annual  "` collides with `"Annual"` as expected and the length check measures the real name. Added tests for the trimmed name and for renaming to a whitespace-padded duplicate.

backend/src/test/java/com/skapp/community/leaveplanner/controller/v1/PolicyLeaveTypeControllerIntegrationTest.java
hasalarootcode: Scoped them. Added a `performScopedSearch` helper that passes `searchKeyword=Policy` (the seed name prefix), and switched the tests asserting absolute totals over to it, so the counts only cover rows the test seeds. Left the case-insensitive and no-match tests on the plain helper since they assert their own keyword.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveTypeServiceImpl.java
hasalarootcode: Kept it as the lightweight dropdown lookup and made that explicit in the `@Operation` description — it now states it's deliberately unpaginated, returns id/name/appearance only, and points at `/search` for the management list view. The FE uses it for the policy-creation dropdown, so dropping it would mean a large page size for what is a 4-field lookup.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveTypeServiceImpl.java
```diff
@@ -34,4 +48,193 @@ public ResponseEntityDto getPolicyLeaveTypes() {
...
+				new PolicyLeaveTypeStatusResponseDto(policyLeaveType.getId(), policyLeaveType.getIsActive()));
+	}
+
+	private PolicyLeaveType getPolicyLeaveType(Long id) {
+		return policyLeaveTypeDao.findById(id)
+			.orElseThrow(
+					() -> new EntityNotFoundException(LeaveMessageConstant.LEAVE_ERROR_POLICY_LEAVE_TYPE_ID_NOT_FOUND));
+	}
+
+	private String normalizeName(String name) {
```
shakila-kamalasena: remove this

backend/src/main/java/com/skapp/community/leaveplanner/util/PolicyLeaveTypeValidationUtil.java
shakila-kamalasena: We do not need trim

shakila-kamalasena (changes requested):

backend/src/main/java/com/skapp/community/leaveplanner/constant/PolicyLeaveTypeConstant.java
```diff
@@ -0,0 +1,16 @@
+package com.skapp.community.leaveplanner.constant;
+
+import lombok.experimental.UtilityClass;
+
+@UtilityClass
+public class PolicyLeaveTypeConstant {
+
+	public static final int MAX_NAME_LENGTH = 100;
+
```
akilarootcode, line 9: remove

backend/src/main/java/com/skapp/community/leaveplanner/constant/PolicyLeaveTypeConstant.java
```diff
@@ -0,0 +1,16 @@
...
+	public static final int MIN_PAGE_NUMBER = 0;
+
+	public static final int MIN_PAGE_SIZE = 1;
+
+	public static final int MAX_PAGE_SIZE = 100;
```
akilarootcode: remov this

backend/src/main/java/com/skapp/community/leaveplanner/controller/v1/PolicyLeaveTypeController.java
```diff
@@ -22,12 +29,70 @@ public class PolicyLeaveTypeController {
...
 	@PreAuthorize("hasAnyRole('ROLE_LEAVE_ADMIN')")
 	public ResponseEntity<ResponseEntityDto> getPolicyLeaveTypes() {
 		ResponseEntityDto response = policyLeaveTypeService.getPolicyLeaveTypes();
 		return new ResponseEntity<>(response, HttpStatus.OK);
 	}
 
+	@Operation(summary = "Search policy leave types",
+			description = "Returns a paginated list of leave types with optional search and active status filter")
+	@GetMapping("/search")
+	@PreAuthorize("hasAnyRole('ROLE_LEAVE_ADMIN')")
```
akilarootcode: remove this and move search to getPolicyLeaveTypes

backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/PolicyLeaveTypeRepositoryImpl.java
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	private List<Predicate> buildPredicates(CriteriaBuilder cb, Root<PolicyLeaveType> root,
+			PolicyLeaveTypeFilterDto filterDto) {
+		List<Predicate> predicates = new ArrayList<>();
+
+		String searchKeyword = filterDto.getSearchKeyword();
+		if (searchKeyword != null && !searchKeyword.isBlank()) {
+			String escaped = StringUtils.escapeLikePattern(searchKeyword.toLowerCase());
+			predicates.add(cb.like(cb.lower(root.get(PolicyLeaveType_.name)), "%" + escaped + "%", '\\'));
```
akilarootcode: remove '\\'

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveTypeServiceImpl.java
```diff
@@ -34,4 +48,193 @@ public ResponseEntityDto getPolicyLeaveTypes() {
...
+	@Override
+	@Transactional
+	public ResponseEntityDto updatePolicyLeaveType(Long id,
+			PolicyLeaveTypeUpdateRequestDto policyLeaveTypeUpdateRequestDto) {
+		log.info("updatePolicyLeaveType: execution started");
+
+		PolicyLeaveType policyLeaveType = getPolicyLeaveType(id);
+
+		if (policyLeaveTypeUpdateRequestDto.getName() != null) {
+			String name = normalizeName(policyLeaveTypeUpdateRequestDto.getName());
```
akilarootcode: remove normalizeName

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveTypeServiceImpl.java
```diff
@@ -34,4 +48,193 @@ public ResponseEntityDto getPolicyLeaveTypes() {
...
+			PolicyLeaveTypeValidationUtil.validateName(name);
+
+			if (policyLeaveTypeDao.existsByNameIgnoreCaseAndIdNot(name, id)) {
```
akilarootcode: move this to first level

backend/src/main/java/com/skapp/community/leaveplanner/util/PolicyLeaveTypeValidationUtil.java
```diff
@@ -0,0 +1,58 @@
...
+		}
+	}
+
+	public static void validateAttachmentSetup(Boolean isAttachment, Boolean isAttachmentMust) {
+		if (!Boolean.TRUE.equals(isAttachment) && Boolean.TRUE.equals(isAttachmentMust)) {
+			throw new ModuleException(LeaveMessageConstant.LEAVE_ERROR_LEAVE_TYPE_UNABLE_TO_MAKE_ATTACHMENT_MANDATORY);
+		}
+	}
+
+	public static void validatePagination(int page, int size) {
```
akilarootcode: remove pagination

backend/src/main/java/com/skapp/community/leaveplanner/util/PolicyLeaveTypeValidationUtil.java
```diff
@@ -0,0 +1,58 @@
...
+
+import com.skapp.community.common.exception.ModuleException;
+import com.skapp.community.common.util.Validation;
+import com.skapp.community.leaveplanner.constant.LeaveMessageConstant;
+import com.skapp.community.leaveplanner.constant.PolicyLeaveTypeConstant;
+import com.skapp.community.leaveplanner.type.LeaveDuration;
+import lombok.experimental.UtilityClass;
+
+@UtilityClass
+public class PolicyLeaveTypeValidationUtil {
```
akilarootcode, line 11: remove util test

akilarootcode (changes requested):

backend/src/main/java/com/skapp/community/leaveplanner/util/PolicyLeaveTypeValidationUtil.java
hasalarootcode: Removed

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveTypeServiceImpl.java
hasalarootcode: Removed

backend/src/main/java/com/skapp/community/leaveplanner/constant/PolicyLeaveTypeConstant.java
hasalarootcode: Fixed

backend/src/main/java/com/skapp/community/leaveplanner/constant/PolicyLeaveTypeConstant.java
hasalarootcode: Fixed

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveTypeServiceImpl.java
hasalarootcode: Removed

backend/src/main/java/com/skapp/community/leaveplanner/util/PolicyLeaveTypeValidationUtil.java
hasalarootcode: removed pagination validations

backend/src/main/java/com/skapp/community/leaveplanner/util/PolicyLeaveTypeValidationUtil.java
hasalarootcode: Mistaken this is not  a test file

backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/PolicyLeaveTypeRepositoryImpl.java
hasalarootcode: Removed

shakila-kamalasena (dismissed):

shakila-kamalasena (dismissed):

backend/src/main/java/com/skapp/community/leaveplanner/controller/v1/PolicyLeaveTypeController.java
hasalarootcode: Removed

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveTypeServiceImpl.java
```diff
@@ -24,14 +37,192 @@ public class PolicyLeaveTypeServiceImpl implements PolicyLeaveTypeService {
 
 	@Override
 	@Transactional(readOnly = true)
-	public ResponseEntityDto getPolicyLeaveTypes() {
+	public ResponseEntityDto getPolicyLeaveTypes(PolicyLeaveTypeFilterDto policyLeaveTypeFilterDto) {
 		log.info("getPolicyLeaveTypes: execution started");
 
-		List<PolicyLeaveTypeResponseDto> leaveTypes = leaveMapper
-			.policyLeaveTypeListToPolicyLeaveTypeResponseDtoList(policyLeaveTypeDao.findAllByIsActive(true));
+		Sort sort = Sort.by(Sort.Order.asc("name").ignoreCase());
```
akilarootcode: use sort key enum

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveTypeServiceImpl.java
```diff
@@ -24,14 +37,192 @@ public class PolicyLeaveTypeServiceImpl implements PolicyLeaveTypeService {
...
+				new PolicyLeaveTypeStatusResponseDto(policyLeaveType.getId(), policyLeaveType.getIsActive()));
+	}
+
+	private PolicyLeaveType getPolicyLeaveType(Long id) {
+		return policyLeaveTypeDao.findById(id)
+			.orElseThrow(
+					() -> new EntityNotFoundException(LeaveMessageConstant.LEAVE_ERROR_POLICY_LEAVE_TYPE_ID_NOT_FOUND));
+	}
+
+	private String normalizeName(String name) {
```
akilarootcode: remove this normalize

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveTypeServiceImpl.java
```diff
@@ -24,14 +37,192 @@ public class PolicyLeaveTypeServiceImpl implements PolicyLeaveTypeService {
...
+	}
+
+	@Override
+	@Transactional
+	public ResponseEntityDto addPolicyLeaveType(PolicyLeaveTypeRequestDto policyLeaveTypeRequestDto) {
+		log.info("addPolicyLeaveType: execution started");
+
+		String name = normalizeName(policyLeaveTypeRequestDto.getName());
+
+		PolicyLeaveTypeValidationUtil.validateName(name);
```
akilarootcode: use request dto directly

akilarootcode (changes requested):

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveTypeServiceImpl.java
hasalarootcode: Fixed

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveTypeServiceImpl.java
hasalarootcode: Fixed

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveTypeServiceImpl.java
hasalarootcode: Fixed

akilarootcode (approved):

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveTypeServiceImpl.java
hasalarootcode: Fixed

akilarootcode (approved):

PR #1692, opened 2026-07-31
Feat/policy leave type FE

hasalarootcode:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR introduces a parallel "policy leave type" flow (API layer, table, form, activation toggle) behind a feature-flag hook, switching the existing Leave Types page over when enabled. The code is well structured and closely mirrors the sibling Leave Policies feature, but the flag is hardcoded to `true` while none of the six endpoints it calls exist in the backend yet, so shipping as-is breaks a live page. I also found a silent create-instead-of-edit path, a form-reset that discards unsaved edits, a backend message key that doesn't exist, and lost enterprise quick-setup/GA behaviour from the legacy form.

Found **30** new issue(s): 🔴 10 important, 🟡 14 suggestion(s), 🟣 6 nit(s)

frontend/src/community/leave/constants/policyLeaveTypeConstants.ts
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * The leave policy feature flag is persisted in organization config and will be
+ * exposed by `GET /v1/leave/policies/config` in a separate PR. Until then this
+ * constant stands in for it, so the old/new leave type switch can be built and
+ * reviewed.
+ *
+ * TODO: remove this constant and back `useIsLeavePoliciesEnabled` with the real
+ * config query once the endpoint is available.
+ */
+export const IS_LEAVE_POLICIES_ENABLED_MOCK = true;
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`IS_LEAVE_POLICIES_ENABLED_MOCK = true` switches EVERY org onto the new flow, but at this commit the backend only exposes `GET /v1/leave/policy-leave-types` (`PolicyLeaveTypeController` has a single `@GetMapping`; `PolicyLeaveTypeService` declares only `getPolicyLeaveTypes()`). All six endpoints this PR adds — `/policy-leave-types/search`, `POST /policy-leave-types`, `GET|PATCH /policy-leave-types/{id}`, `.../activate`, `.../deactivate` — are unimplemented, so the Leave Types page will list nothing (404 → "No leave types yet" empty state) and add/edit/activate will all fail. Separately, `lv_leave_type` (entity `PolicyLeaveType`) is a different table from the legacy `leave_type`, so even once the endpoints land, types managed here are invisible to apply-leave/entitlements/dashboards, which still read `leave_type`. Default the mock to `false` so the legacy flow keeps working, and flip it only when the config endpoint and CRUD endpoints are merged.

frontend/src/community/leave/components/organisms/PolicyLeaveTypeForm/PolicyLeaveTypeForm.tsx
```diff
@@ -0,0 +1,381 @@
...
+      isCommentMust: editingPolicyLeaveType?.isCommentMust ?? false,
+      isAutoApproval: editingPolicyLeaveType?.isAutoApproval ?? false
+    }),
+    [editingPolicyLeaveType]
+  );
+
+  const onSubmit = (formValues: PolicyLeaveTypeFormDataType) => {
+    const payload = mapPolicyLeaveTypeFormToPayload(formValues);
+
+    if (isEditMode && policyLeaveTypeId) {
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

An "edit" URL without an `id` silently becomes a create. `isEditMode` is derived from `slug` only, so the breadcrumb/title say "Edit Leave Type", but `policyLeaveTypeId` is `undefined`, the detail query is disabled, and `if (isEditMode && policyLeaveTypeId)` is false — so submitting falls through to `addPolicyLeaveType(payload)` and creates a brand-new leave type. This is reachable: the legacy `LeaveTypesTable.handleEditLeaveType` pushes `ROUTES.LEAVE.ADD_EDIT_LEAVE_TYPES(EDIT)` with no query param at all, so any bookmark/back-navigation from the old flow lands on `/leave/types/edit` with no id. Guard explicitly: when `isEditMode && !policyLeaveTypeId`, redirect to `ROUTES.LEAVE.LEAVE_TYPES` (or render an error) instead of rendering a submittable form.

frontend/src/community/leave/components/organisms/PolicyLeaveTypeForm/PolicyLeaveTypeForm.tsx
```diff
@@ -0,0 +1,381 @@
...
+      return;
+    }
+
+    addPolicyLeaveType(payload);
+  };
+
+  const formik = useFormik({
+    initialValues,
+    validationSchema: policyLeaveTypeValidationSchema(translateText),
+    enableReinitialize: true,
```
ThinuwanW, line 92: 🤖 **Claude** · 🔴 **Important**

`enableReinitialize: true` combined with `initialValues` memoized on `[editingPolicyLeaveType]` means any refetch of the detail query silently discards the user's unsaved edits and resets `dirty` (which via the effect at line 153 also disarms the unsaved-changes guard in `[slug].tsx`). Two concrete triggers: (1) `PolicyLeaveTypeActivationToggleButton` renders on the same page and its mutation invalidates `policyLeaveTypeQueryKeys.ALL`, which is a prefix of `DETAIL(id)` — so toggling activation while the form has typed changes wipes them; (2) `TanStackProvider` sets no default `refetchOnWindowFocus`/`staleTime`, so React Query's default focus refetch does the same on tab switch. Fix: pass `refetchOnWindowFocus: false` and a `staleTime` on `useGetPolicyLeaveType` (the sibling `useGetLeavePoliciesInfinite`/`useGetLeavePolicyConfig` already do this), and/or scope the toggle's invalidation to the search/list keys rather than `ALL`.

frontend/src/community/common/constants/errorMessageKeys.ts
```diff
@@ -16,6 +16,8 @@ export const LEAVE_ERROR_NUMBER_OF_DAYS_CANNOT_BE_LESS_THAN_USED_DAYS =
   "LEAVE_ERROR_NUMBER_OF_DAYS_CANNOT_BE_LESS_THAN_USED_DAYS";
 export const LEAVE_ERROR_LEAVE_POLICY_ALREADY_EXISTS =
   "LEAVE_ERROR_LEAVE_POLICY_ALREADY_EXISTS";
+export const LEAVE_ERROR_POLICY_LEAVE_TYPE_ALREADY_EXISTS =
```
ThinuwanW, line 19: 🤖 **Claude** · 🔴 **Important**

`LEAVE_ERROR_POLICY_LEAVE_TYPE_ALREADY_EXISTS` does not exist anywhere in the backend. `LeaveMessageConstant` defines `LEAVE_ERROR_LEAVE_TYPE_ALREADY_EXISTS` (and `LEAVE_ERROR_POLICY_LEAVE_TYPE_NOT_FOUND`), but not this one. Because `getPolicyLeaveTypeErrorToastKeys` switches on this key, the duplicate-name case will always fall through to the generic "We couldn't save the leave type" toast — and duplicate-name is the primary error path here, since the new `lv_leave_type.name` column has no UNIQUE constraint and the client-side uniqueness check that `addLeaveTypeValidationSchema` had was not carried over. Confirm the exact `messageKey` the new POST/PATCH endpoints will emit and match it, or reuse `LEAVE_ERROR_LEAVE_TYPE_ALREADY_EXISTS`.

frontend/src/community/leave/components/organisms/PolicyLeaveTypeForm/PolicyLeaveTypeForm.tsx
```diff
@@ -0,0 +1,381 @@
...
+  const { setLeaveTypeFormDirty } = useLeaveStore((state) => ({
+    setLeaveTypeFormDirty: state.setLeaveTypeFormDirty
+  }));
+
+  const [colors, setColors] = useState<string[]>(leaveTypeColors);
+
+  const { data: editingPolicyLeaveType, isLoading: isPolicyLeaveTypeLoading } =
+    useGetPolicyLeaveType(isEditMode ? policyLeaveTypeId : undefined);
+
+  const onMutationSuccess = async (isEdit: boolean) => {
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`onMutationSuccess` drops behaviour that the legacy `handleLeaveTypeApiResponse` performed, so with the flag on these regress: (1) no `sendEvent(GoogleAnalyticsTypes.GA4_LEAVE_TYPE_CREATED / GA4_LEAVE_TYPE_UPDATED)` — leave-type analytics stop firing; (2) no `setQuickSetupModalType(QuickSetupModalTypeEnums.IN_PROGRESS_START_UP)` + `stopAllOngoingQuickSetup()` when `ongoingQuickSetup.SETUP_LEAVE_TYPES` is active — the enterprise onboarding wizard stalls on the leave-types step even though `pages/community/leave/types/index.tsx` still blinks the primary button for it; (3) `useAddPolicyLeaveType` does not invalidate `quickSetupQueryKeys.QUICK_SETUP_PROGRESS`, which `useAddLeaveType` does, so quick-setup progress stays stale. Wire `useGoogleAnalyticsEvent` and `useCommonEnterpriseStore` into this form the same way `LeaveTypeForm` does (including `getBlinkClass` on the Save button).

frontend/src/community/leave/components/molecules/PolicyLeaveTypesTable/PolicyLeaveTypesTable.tsx
```diff
@@ -0,0 +1,155 @@
...
+import { getLeaveTypeDurationTableContent } from "~community/leave/utils/leaveTypes/LeaveTypeUtils";
+
+const PolicyLeaveTypesTable: FC = () => {
+  const translateText = useTranslator("leaveModule", "leaveTypes");
+
+  const router = useRouter();
+
+  const [page, setPage] = useState<number>(0);
+
+  const { data: policyLeaveTypesPage, isLoading } = useSearchPolicyLeaveTypes({
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

No error state anywhere in the new read paths, and the query functions lie about their return type. `searchPolicyLeaveTypes` returns `response.data.results[0]` typed as a non-optional `PolicyLeaveTypesPage`; if the backend ever returns `results: []`, the queryFn resolves to `undefined`, which React Query v5 rejects ("Query data cannot be undefined") — the query fails and this table renders the `noDataState` "No leave types yet" message, indistinguishable from a genuinely empty list. Same for `getPolicyLeaveType`, where a failure leaves `PolicyLeaveTypeForm` rendering a blank form with no indication that loading failed (and only the colour palette has a skeleton — name/emoji show empty then pop). Destructure `isError` here and render an error state, guard with `results?.[0]` and widen the hook return types to `| undefined`, and surface a toast/error state in the edit form.

frontend/src/community/leave/components/molecules/PolicyLeaveTypesTable/PolicyLeaveTypesTable.tsx
```diff
@@ -0,0 +1,155 @@
+import {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

No role gating on the write affordances. `useCanManageLeavePolicies` already exists in this module and the sibling `LeavePoliciesTable` uses it to append the actions column only for super/leave admins. Here the Edit `IconButton` is always rendered, and `PolicyLeaveTypeActivationToggleButton` always renders an enabled toggle. Since the backend gates these operations with `@PreAuthorize("hasAnyRole('ROLE_LEAVE_ADMIN')")`, non-admins get a 403 and an "Access Denied" toast instead of simply not seeing the control. Gate both with `useCanManageLeavePolicies()` for consistency.

frontend/src/community/leave/api/PolicyLeaveTypeApi.ts
```diff
@@ -0,0 +1,187 @@
...
+): Promise<PolicyLeaveTypesPage> => {
+  const response = await authFetch.get<SearchPolicyLeaveTypesResponse>(
+    policyLeaveTypeEndPoints.SEARCH_POLICY_LEAVE_TYPES,
+    { params }
+  );
+
+  return response.data.results[0];
+};
+
+export const useSearchPolicyLeaveTypes = ({
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`searchKeyword` and `isActive` are threaded through `UseSearchPolicyLeaveTypesArgs`, `SearchPolicyLeaveTypesParams` and the query key, but no caller ever sets them — `PolicyLeaveTypesTable` hardcodes `searchKeyword: ""` and omits `isActive`. So the page has neither search nor a status filter, unlike the sibling `LeavePoliciesTable` which pairs an `InputField` + `Dropdown` with `useDebounce(searchTerm, LEAVE_POLICY_SEARCH_DEBOUNCE_MS)`. Either add the search box / status filter (plus a `POLICY_LEAVE_TYPE_SEARCH_DEBOUNCE_MS` constant and `emptyStateType`/`noSearchResultsState` on the Table, as the sibling does) or drop the unused parameters until they're needed.

frontend/src/community/leave/components/molecules/PolicyLeaveTypesTable/PolicyLeaveTypesTable.tsx
```diff
@@ -0,0 +1,155 @@
...
+          />
+        );
+      }
+    },
+    {
+      key: "durations",
+      header: translateText(["durationsHeader"]),
+      render: (value: unknown) => (
+        <div className="flex flex-row flex-wrap gap-2">
+          {getLeaveTypeDurationTableContent(value as LeaveDurationTypes).map(
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Untranslated user-facing strings: `getLeaveTypeDurationTableContent` returns hardcoded English `["Half Day"]` / `["Full Day"]` / `["Full Day", "Half Day"]`, so the ALLOWED DURATION chips bypass i18n entirely — even though `leaveModule.json` already has `leaveTypes.halfDay` ("Half-day") and `leaveTypes.fullDay` ("Full-day"). Map the `LeaveDurationTypes` enum to labels locally, e.g. return the enum members from a helper and render `translateText(["halfDay"])` / `translateText(["fullDay"])`, rather than reusing the legacy string-returning helper. (Note this also makes the chip copy match the form's duration cards.)

frontend/src/community/leave/utils/policyLeaveTypes/policyLeaveTypeUtils.ts
```diff
@@ -0,0 +1,89 @@
+import { AxiosError } from "axios";
```
ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

No unit tests for the new pure logic, even though this module has an established convention for exactly this: `utils/leaveTypes/LeaveTypeUtils.test.ts` covers the equivalent legacy `handleLeaveDurationClick`, and `utils/__test__/validations.test.ts` covers `addLeaveTypeValidationSchema` / `customLeaveAllocationValidation`. Please add tests for `getUpdatedMinDuration` (all four enum transitions, including the HALF_AND_FULL_DAY → single-duration path), `isMinDurationSelected`, `mapPolicyLeaveTypeFormToPayload` (the `isAttachment === false` → `isAttachmentMust: false` coercion and the name `trim()`), `getPolicyLeaveTypeErrorToastKeys` (all three branches plus a malformed error object), and `policyLeaveTypeValidationSchema`.

frontend/src/community/leave/components/molecules/PolicyLeaveTypeActivationToggleButton/PolicyLeaveTypeActivationToggleButton.tsx
```diff
@@ -0,0 +1,137 @@
...
+  const { setToastMessage } = useToast();
+
+  const [isConfirmModalOpen, setIsConfirmModalOpen] = useState(false);
+
+  const { data: policyLeaveType } = useGetPolicyLeaveType(policyLeaveTypeId);
+
+  const isActive = policyLeaveType?.isActive ?? true;
+
+  const { mutate: changeStatus, isPending } = useChangePolicyLeaveTypeStatus(
+    () => {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The error-toast block here is duplicated verbatim from `PolicyLeaveTypeForm.onMutationError` (same `getPolicyLeaveTypeErrorToastKeys` call, same `setToastMessage` shape). Extract a shared helper — e.g. `showPolicyLeaveTypeErrorToast({ error, setToastMessage, translateText })` in `policyLeaveTypeUtils.ts` — so the two call sites can't drift. Also worth noting: activate and deactivate both reuse the generic `editLeaveTypeSuccessToastTitle` / `editLeaveTypeSuccessToastDescription` ("Leave Type Updated" / "successfully updated"), which reads oddly for an activation change; consider dedicated copy given you're already adding activation-specific modal descriptions.

frontend/src/community/leave/components/molecules/PolicyLeaveTypesTable/PolicyLeaveTypesTable.tsx
```diff
@@ -0,0 +1,155 @@
...
+  const handleEditPolicyLeaveType = (
+    policyLeaveType: PolicyLeaveTypeSettingsType
+  ): void => {
+    router.push({
+      pathname: ROUTES.LEAVE.ADD_EDIT_LEAVE_TYPES(LeaveTypeFormTypes.EDIT),
+      query: { id: policyLeaveType.id }
+    });
+  };
+
+  const columns = [
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`columns` is rebuilt on every render (and each `render` closure recaptures `handleEditPolicyLeaveType`, itself recreated every render), while `tableData` right above it is carefully memoized — inconsistent, and it defeats any memoization inside `Table`. Wrap `columns` in `useMemo` keyed on `[translateText, handleEditPolicyLeaveType]` and wrap `handleEditPolicyLeaveType` / `handlePageChange` in `useCallback`.

frontend/src/community/leave/components/molecules/PolicyLeaveTypesTable/PolicyLeaveTypesTable.tsx
```diff
@@ -0,0 +1,155 @@
...
+    <div className="mt-4 flex flex-col gap-4">
+      <Table<TableRow>
+        columns={columns}
+        data={tableData}
+        tableAriaLabel={translateText(["title"])}
+        isLoading={isLoading}
+        noDataState={{ title: translateText(["noLeaveTypesTitle"]) }}
+      />
+      {totalPages > 1 && (
+        <Pagination
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

This is the first use of the skapp-ui `Pagination` anywhere in the repo (the sibling policies table uses infinite scroll instead), so the page-index convention is unverified. `page` state is 0-based and is passed straight to the API, and it's passed straight to `currentPage`. Note the community MUI wrapper at `src/community/common/components/atoms/Pagination/Pagination.tsx` internally does `page={currentPage + 1}`, i.e. it takes a 0-based prop — please confirm the skapp-ui component has the same contract for both `currentPage` and the `onChange` argument, otherwise the list will be off by one page.

frontend/src/community/leave/components/organisms/PolicyLeaveTypeForm/PolicyLeaveTypeForm.tsx
```diff
@@ -0,0 +1,381 @@
...
+  const isEditMode = slug === LeaveTypeFormTypes.EDIT;
+  const policyLeaveTypeId = id ? Number(id) : undefined;
+
+  const { setToastMessage } = useToast();
+
+  const { setLeaveTypeFormDirty } = useLeaveStore((state) => ({
+    setLeaveTypeFormDirty: state.setLeaveTypeFormDirty
+  }));
+
+  const [colors, setColors] = useState<string[]>(leaveTypeColors);
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`colors` state is never meaningfully updated: `handleColorClick` (from the legacy `LeaveTypeUtils`) calls `setColors(colors)` with the identical array, so the only effect is an extra render. Since this form always uses the static `leaveTypeColors`, this can just be `const colors = leaveTypeColors;` and `handleColorSelect` can set `colorCode` / clear its error directly, dropping the `useState` and the `setColors` plumbing.

frontend/src/community/leave/components/organisms/PolicyLeaveTypeForm/DurationOptionCard.tsx
```diff
@@ -0,0 +1,64 @@
...
+  onSelect
+}) => {
+  const handleKeyDown = (event: KeyboardEvent<HTMLDivElement>): void => {
+    if (event.key === "Enter" || event.key === " ") {
+      event.preventDefault();
+      onSelect();
+      return;
+    }
+
+    if (event.key === "ArrowDown" || event.key === "ArrowRight") {
```
ThinuwanW, line 32: 🤖 **Claude** · 🟣 **Nit**

Arrow-key navigation uses `event.currentTarget.nextElementSibling` / `previousElementSibling`, which only works because the two cards happen to be the sole children of the wrapper div at PolicyLeaveTypeForm.tsx:277 — it breaks silently if anything (a hint, a third duration) is inserted between them. Consider driving focus from the parent (refs or a small roving-tabindex list) instead. Two smaller a11y points: `role="checkbox"` is paired with a `RadioButton` visual, which signals single-select to sighted users while announcing multi-select; and the error paragraph at PolicyLeaveTypeForm.tsx:301 isn't linked to the cards via `aria-describedby`, nor is the pair wrapped in a `role="group"` with the section heading as its label.

frontend/src/community/leave/components/organisms/PolicyLeaveTypeForm/PolicyLeaveTypeForm.tsx
```diff
@@ -0,0 +1,381 @@
...
+
+  const { mutate: updatePolicyLeaveType, isPending: isUpdatePending } =
+    useUpdatePolicyLeaveType(() => onMutationSuccess(true), onMutationError);
+
+  const initialValues: PolicyLeaveTypeFormDataType = useMemo(
+    () => ({
+      name: editingPolicyLeaveType?.name ?? "",
+      emoji: getEmoji(editingPolicyLeaveType?.emojiCode ?? ""),
+      emojiCode: editingPolicyLeaveType?.emojiCode ?? "",
+      colorCode: editingPolicyLeaveType?.colorCode ?? leaveTypeColors[0],
```
ThinuwanW, line 78: 🤖 **Claude** · 🟣 **Nit**

In add mode `colorCode` defaults to `leaveTypeColors[0]`, so the `emptyLeaveTypeColorError` branch of `policyLeaveTypeValidationSchema` is unreachable and the user is never required to pick a colour — a behaviour change from `LeaveTypeForm`, which left `colorCode` undefined and enforced an explicit choice. If pre-selecting a default is intended, fine; otherwise default to `""` so the required rule can fire.

frontend/src/community/leave/utils/validations.ts
```diff
@@ -114,6 +115,28 @@ export const addLeaveTypeValidationSchema = (
...
   });
 
+export const policyLeaveTypeValidationSchema = (
+  translateText: TranslatorFunctionType
+) =>
+  Yup.object({
+    name: Yup.string()
+      .trim()
+      .required(translateText(["emptyLeaveTypeNameError"]))
+      .max(
```
ThinuwanW, line 125: 🤖 **Claude** · 🟣 **Nit**

Two things on the name length. (1) The `.max(MAX_POLICY_LEAVE_TYPE_NAME_LENGTH)` rule is unreachable from the UI because `InputField` already sets `maxLength={MAX_POLICY_LEAVE_TYPE_NAME_LENGTH}` — harmless, but it means `leaveTypeNameMaxLengthError` is dead copy unless you drop the `maxLength` attribute. (2) 100 differs from the legacy limit of `characterLengths.LEAVE_TYPE_LENGTH = 20` used by `LeaveTypeForm`; the new `lv_leave_type.name` column is `text` so the DB accepts it, but please confirm the forthcoming POST/PATCH `@Valid` constraints agree on 100, since backend validation is meant to be the source of truth for field limits.

frontend/src/community/leave/api/PolicyLeaveTypeApi.ts
```diff
@@ -0,0 +1,187 @@
+import {
```
ThinuwanW, line 1: 🤖 **Claude** · 🔴 **Important**

📁 **File-level observation**

Missing cross-cache invalidation. Leave-type data is cached in three other places: (1) `leaveTypeQueryKeys.LEAVE_TYPES(params)` — read by `useGetLeaveTypes` / `useGetPreProcessedLeaveTypes` and consumed by apply-leave, entitlements, dashboards, report filters and the legacy `LeaveTypesTable`; (2) `quickSetupQueryKeys.QUICK_SETUP_PROGRESS` — the legacy `useAddLeaveType` invalidates it so the enterprise quick-setup step ticks over after the first leave type is created; (3) `leavePolicyQueryKeys.POLICY_LEAVE_TYPES` — read by `useGetPolicyLeaveTypes`, which feeds the leave-policy wizard's leave-type dropdown (LeavePolicyWizard/BasicInfoStep.tsx:31) and the policies table's leave-type filter (LeavePoliciesTable.tsx:88). Because the new mutations invalidate only the new `policyLeaveType*` key namespace, creating/renaming/activating/deactivating a leave type through the new flow leaves all three stale: a new leave type will not appear in the policy wizard dropdown, a deactivated one keeps showing, quick-setup progress does not advance, and every legacy screen keeps the old list until a hard refresh. Fix: in each mutation's `onSuccess`, additionally invalidate `leaveTypeQueryKeys.ALL` (or `LEAVE_TYPES(params)`), `quickSetupQueryKeys.QUICK_SETUP_PROGRESS`, and `leavePolicyQueryKeys.POLICY_LEAVE_TYPES` — or better, drop the second key namespace and nest the new keys under `leavePolicyQueryKeys.POLICY_LEAVE_TYPES` so one invalidate covers both readers.

frontend/src/community/leave/hooks/useIsLeavePoliciesEnabled.ts
```diff
@@ -0,0 +1,15 @@
+import { IS_LEAVE_POLICIES_ENABLED_MOCK } from "~community/leave/constants/policyLeaveTypeConstants";
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

📁 **File-level observation**

This hook duplicates an existing one that already does the job correctly. `frontend/src/community/leave/hooks/useLeavePoliciesEnabled.ts` reads the real backend config via `useGetLeavePolicyConfig` and returns `{ isLeavePoliciesEnabled, isLoading, isError }`; it is already used by `LeaveConfigurations.tsx:23`. The new hook sits in the same directory under an almost identical name (`useIsLeavePoliciesEnabled` vs `useLeavePoliciesEnabled`), which is a trap for the next reader, and it returns a bare boolean with no loading/error state — so once the mock constant is replaced with the real query the pages will flash the legacy table before swapping to the new one, and a config fetch failure will silently fall back to the legacy UI with no signal. Fix: delete this hook and consume `useLeavePoliciesEnabled` in the two pages, rendering a skeleton (or nothing) while `isLoading`, and treating `isError` explicitly the way `LeaveConfigurations` does.

frontend/src/community/leave/components/molecules/PolicyLeaveTypeActivationToggleButton/PolicyLeaveTypeActivationToggleButton.tsx
```diff
@@ -0,0 +1,137 @@
+import {
```
ThinuwanW, line 1: 🤖 **Claude** · 🔴 **Important**

📁 **File-level observation**

Deactivation now happens immediately, with no confirmation. This component lives in the page header (`customRightContent`), outside the form, and fires its own mutation with its own error toast — so flipping the toggle commits the change instantly. Both established flows in this module require confirmation first: the legacy leave-type flow only stages `isActive` in the Zustand store and routes the save through `ConfirmLeaveTypeStatusUpdateModal`, which shows the `inactivateLeaveTypeModalTitle` / `inactivateLeaveTypeModalDescription` warning that all employee entitlements under the type become invisible; the sibling policy flow uses `DeactivateLeavePolicyModal`. Losing that step on a destructive, entitlement-affecting action is a real regression, and it also orphans the four existing `activateLeaveTypeModalTitle/Description` / `inactivateLeaveTypeModalTitle/Description` translation keys. Fix: gate the deactivate path behind a confirmation modal (reuse `UserPromptModal` + the existing translation keys, mirroring `DeactivateLeavePolicyModal`), and only mutate on confirm.

frontend/pages/community/leave/types/[slug].tsx
```diff
@@ -20,6 +23,8 @@ const LeaveType: NextPage = () => {
   const router = useRouter();
   const { slug } = router.query;
 
+  const isLeavePoliciesEnabled = useIsLeavePoliciesEnabled();
```
Comment (ThinuwanW) [inline frontend/pages/community/leave/types/[slug].tsx]: 🤖 **Claude** · 🔴 **Important**

The unsaved-changes guard on this page becomes dead code in the new flow. The `useEffect` at lines 43-70 blocks `routeChangeStart` based on `isLeaveTypeFormDirty` and opens `UnsavedChangesModal`; `isLeaveTypeFormDirty` is written in exactly one place in the codebase — `LeaveTypeForm.tsx:223` (`setLeaveTypeFormDirty(dirty)`). The new form derives its state from the detail query rather than `useLeaveStore`, so when the flag is on, `isLeaveTypeFormDirty` stays `false`, the guard never fires, and a user with a half-edited leave type navigates away silently losing the edits. The `resetEditingLeaveType` branch and the still-rendered `UnsavedChangesModal` are likewise unreachable on this path. Fix: have the new form mirror the legacy `useEffect` and call `setLeaveTypeFormDirty(dirty)` (and reset it on successful save), or move the dirty-tracking into a shared hook that both forms use.

frontend/src/community/leave/api/utils/ApiEndpoints.ts
```diff
@@ -44,6 +44,19 @@ export const leavePolicyEndPoints = {
   GET_LEAVE_POLICIES: `${moduleAPIPath.LEAVE}/policies`
 };
 
+export const policyLeaveTypeEndPoints = {
```
ThinuwanW, line 50: 🤖 **Claude** · 🟡 **Suggestion**

Duplicate plumbing for one resource, split across two namespaces. The new policy-leave-type CRUD endpoints were added inside the `leavePolicyEndPoints` object, which already owns `GET_POLICY_LEAVE_TYPES` at line 45 — so there are now two constants for the same base path, sitting in the *leave policy* namespace, while every other file this PR adds uses a `policyLeaveType` namespace (`PolicyLeaveTypeApi.ts`, `PolicyLeaveTypeTypes.ts`, `policyLeaveTypeConstants.ts`, `policyLeaveTypeUtils.ts`). The same split exists in `QueryKeys.ts`, where the new factory is added at line 279 immediately above `leavePolicyQueryKeys`, whose `POLICY_LEAVE_TYPES` key already covers this resource; and in `PolicyLeaveTypeApi.ts`, whose list fetcher duplicates `useGetPolicyLeaveTypes` in `LeavePolicyApi.ts:34`. Fix: introduce a dedicated `policyLeaveTypeEndPoints` object (mirroring the existing `leaveTypeEndPoints`), move `GET_POLICY_LEAVE_TYPES` into it, and have `LeavePolicyApi.useGetPolicyLeaveTypes` re-export the new list hook so there is a single fetcher and a single key namespace.

frontend/src/community/leave/types/PolicyLeaveTypeTypes.ts
```diff
@@ -0,0 +1,77 @@
+import { LeaveDurationTypes } from "~community/leave/enums/LeaveTypeEnums";
```
ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Duplicate type definitions for the same backend DTO. `LeavePolicyTypes.ts:73-86` already declares `PolicyLeaveTypeType` (`id`, `name`, `emojiCode`, `colorCode`), `PolicyLeaveTypesResult` and `PolicyLeaveTypesResponse`, and those match the backend `PolicyLeaveTypeResponseDto` / `PolicyLeaveTypesResponseDto` exactly. Declaring a second set in a new file means two names for one contract in the same module, and the two will drift the moment the backend DTO changes (a reader importing the wrong one gets no compile error). Also worth checking against the backend `PolicyLeaveType` entity, where the duration column is `minDuration` (enum `LeaveDuration`) and the attachment/comment flags are `isAttachmentMust` / `isCommentMust` — not the `isAttachmentMandatory` / `isCommentMandatory` names the legacy `LeaveTypeRequestDto` exposes. Fix: import the existing types from `LeavePolicyTypes.ts` (or move them into the new file and re-export from the old one) so there is exactly one definition.

frontend/src/community/leave/components/organisms/PolicyLeaveTypeForm/SettingToggleRow.tsx
```diff
@@ -0,0 +1,30 @@
+import { Toggle } from "@rootcodelabs/skapp-ui";
```
ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Both new sub-components reimplement shared atoms that already exist and are used by the legacy form. `SettingToggleRow` duplicates `~community/common/components/atoms/SwitchRow/SwitchRow`, losing its translated aria-label fallback chain (`arialabelChecked` / `arialabelUnchecked` / `commonAria.components.switch.ariaLabel`), its disabled-label colour handling, and its `role="alert" aria-live` error slot. `DurationOptionCard` duplicates `~community/common/components/atoms/DescribedSelection/DescribedSelection`, losing its `role="button"` + `tabIndex` + shared `shouldActivateButton` keyboard handling and its error styling — which is also the root cause of the fragile sibling-based arrow-key navigation. Separately, the new table should render name+emoji via the existing `LeaveTypeChip` molecule (as `LeavePoliciesTable` does) rather than inline markup. Fix: delete both files and use `SwitchRow` / `DescribedSelection`; if the new visual design genuinely differs, extend those atoms with a variant prop instead of forking them.

frontend/src/community/leave/utils/policyLeaveTypes/policyLeaveTypeUtils.ts
ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

`getLeaveTypeDurationTableContent` here is a verbatim copy of the existing function at `frontend/src/community/leave/utils/leaveTypes/LeaveTypeUtils.ts:62-77`, including the same `LeaveDurationTypes` switch and the same return shapes. Two copies of the same mapping means a future duration enum value (or the i18n fix the first pass called for) has to be applied twice, and one copy will be missed. Fix: import the existing helper from `LeaveTypeUtils`, or move it to a shared location and have both call sites use it — then apply the translation fix once.

frontend/src/community/leave/utils/validations.ts
```diff
@@ -114,6 +115,28 @@ export const addLeaveTypeValidationSchema = (
       })
   });
 
+export const policyLeaveTypeValidationSchema = (
```
ThinuwanW, line 118: 🤖 **Claude** · 🟡 **Suggestion**

The new schema drops the client-side duplicate-name check that `addLeaveTypeValidationSchema` (lines 62-85) performs via the `is-unique-leave-type-name` test against `allLeaveTypes`. In the legacy flow a duplicate name is caught before submit and rendered as an inline field error next to the name input; in the new flow the user submits, waits for a round trip, and gets only a toast — and per the first pass the duplicate message key does not exist in the backend, so they will get the generic "something went wrong" text with no indication which field is wrong. Fix: add an equivalent uniqueness `test` to the new schema, sourcing the existing names from the list query the new table already fetches (compare trimmed + lower-cased on both sides, and exempt the record being edited by id).

frontend/src/community/leave/utils/validations.ts
```diff
@@ -114,6 +115,28 @@ export const addLeaveTypeValidationSchema = (
...
+    colorCode: Yup.string()
+      .transform((value) => (value === null ? "" : value))
+      .required(translateText(["emptyLeaveTypeColorError"])),
+    minDuration: Yup.string().test(
+      "is-min-duration-valid",
+      translateText(["emptyLeaveDurationError"]),
+      (value) => value !== LeaveDurationTypes.NONE
+    )
+  });
+
```
ThinuwanW, line 139: 🤖 **Claude** · 🟡 **Suggestion**

Missing paired test change. `frontend/src/community/leave/utils/__test__/validations.test.ts` already exists and has a dedicated `describe("addLeaveTypeValidationSchema")` block for the sibling schema in this same file, but no case was added for the new schema. Fix: add a `describe` block covering the required-name path, the max-length boundary, the emoji/colour required paths, and the leave-duration `NONE` rejection, following the structure of the existing `addLeaveTypeValidationSchema` block.

frontend/src/community/leave/components/organisms/PolicyLeaveTypeForm/PolicyLeaveTypeForm.tsx
```diff
@@ -0,0 +1,381 @@
+import {
```
ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

At 381 lines this organism is well past the ~200-line guideline and mixes routing, two mutations, toast handling, Formik wiring and the whole JSX tree. The sibling feature already shows the decomposition this module prefers — `LeavePolicyWizard` splits into `BasicInfoStep`, `EntitlementSetupStep`, `SummaryStep`, `WizardSection`. Fix: extract the settings block and the duration block into co-located sub-components (as `DurationOptionCard`/`SettingToggleRow` started to do), and move the create/update mutation + toast wiring into a `usePolicyLeaveTypeFormSubmit` hook so the component is mostly layout.

ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Check the `enabled` guard on the detail query. On a hard load of the edit URL, Next.js renders once with `router.query` empty before hydration fills it, so the `id` is `undefined` on the first render. Without `enabled: Boolean(id)` the detail query fires a request against an `undefined` id (a 400/404 that also lands in the query cache under a key containing `undefined`), and with `enableReinitialize` the form is reinitialised from that failed state. Fix: pass `enabled: Boolean(id)` (matching the pattern `useGetLeavePolicyConfig` uses) and render the form's loading state until the router has resolved and the record has arrived.

frontend/src/community/common/assets/languages/english/leaveModule.json
```diff
@@ -411,7 +415,16 @@
     "activateLeaveTypeModalTitle": "Confirm Activation",
     "activateLeaveTypeModalDescription": "Activating this leave type will make entitlements allocated under it visible to employees. Would you like to proceed?",
     "inactivateLeaveTypeModalTitle": "Confirm Inactivation",
-    "inactivateLeaveTypeModalDescription": "All leave entitlements for employees under this leave type will not be visible once inactivated. Remaining balances will be retained and will become visible again if you reactivate. Would you like to proceed?"
+    "inactivateLeaveTypeModalDescription": "All leave entitlements for employees under this leave type will not be visible once inactivated. Remaining balances will be retained and will become visible again if you reactivate. Would you like to proceed?",
```
ThinuwanW, line 480: 🤖 **Claude** · 🟣 **Nit**

Two i18n housekeeping issues. (1) The ten new keys are inserted between `"enabled"` (line 413) and `"disabled"` (line 414), splitting a logical pair and making the diff read as a bigger change than it is — append them after the existing toast keys instead. (2) The feature's strings are split across two namespaces: four keys go under `leaveModule.leavePolicies` (inserted between `leaveTypeFilterLabel` and `leaveTypeFilterAllOption`) while the rest go under `leaveModule.leaveTypes`, so a translator has to look in two places for one screen and the components need two `useTranslator` scopes. Fix: keep all strings for the new policy-leave-type UI in a single namespace (either entirely under `leaveTypes` or in a new `policyLeaveTypes` block).

frontend/src/community/common/constants/errorMessageKeys.ts
hasalarootcode: The key does exist — `LEAVE_ERROR_POLICY_LEAVE_TYPE_ALREADY_EXISTS` (`api.error.leave.policy-leave-type-already-exists`) is added in `LeaveMessageConstant` on the BE PR #1691, which is still open. It resolves to the specific duplicate-name toast once both sides land, so no change needed here.

frontend/src/community/leave/components/organisms/PolicyLeaveTypeForm/PolicyLeaveTypeForm.tsx
hasalarootcode: Pre-selecting `leaveTypeColors[0]` is intentional — a leave type always renders with a colour, so requiring an explicit pick just adds a validation error for no gain. Keeping the required rule as a backstop for a null `colorCode` coming back from the API.

frontend/src/community/leave/hooks/useIsLeavePoliciesEnabled.ts
hasalarootcode: `useLeavePoliciesEnabled.ts` does not exist on this branch or on develop, and `LeaveConfigurations.tsx` holds the toggle in local `useState` rather than reading a hook — so there is no existing hook being duplicated. `useIsLeavePoliciesEnabled` is a documented temporary stand-in; it gets replaced by the real config query, with the loading/error handling you describe, when `GET /v1/leave/policies/config` lands.

frontend/src/community/leave/components/molecules/PolicyLeaveTypeActivationToggleButton/PolicyLeaveTypeActivationToggleButton.tsx
hasalarootcode: Already gated. `PolicyLeaveTypeActivationToggleButton` opens a `SmallModal` on toggle and only calls `changeStatus` from `handleConfirm`, using `inactivateLeaveTypeModalTitle` plus the new `inactivatePolicyLeaveTypeModalDescription`. That is in as of d27d4e56d, the commit this review ran against.

frontend/pages/community/leave/types/[slug].tsx
Comment (hasalarootcode) [reply frontend/pages/community/leave/types/[slug].tsx]: The new form does write the flag — `setLeaveTypeFormDirty(dirty)` runs in a `useEffect` on `dirty`, and is reset to `false` on a successful save. So `isLeaveTypeFormDirty` tracks the new form and the guard on this page fires as before.

frontend/src/community/leave/api/utils/ApiEndpoints.ts
hasalarootcode: `policyLeaveTypeEndPoints` already exists as its own namespace holding all six CRUD endpoints — the new endpoints are not in `leavePolicyEndPoints`. The remaining overlap (`leavePolicyEndPoints.GET_POLICY_LEAVE_TYPES` and `LeavePolicyApi.useGetPolicyLeaveTypes`) predates this PR and serves a different response shape (`{ leaveTypes: [...] }` for the wizard dropdown), so folding them together is out of scope here. Happy to track it separately.

frontend/src/community/leave/types/PolicyLeaveTypeTypes.ts
hasalarootcode: These are two different DTOs rather than one duplicated. `PolicyLeaveTypeType` is the 4-field projection returned by `GET /policy-leave-types` and feeds the wizard dropdown; `PolicyLeaveTypeSettingsType` is the 10-field detail/search shape the new endpoints return. On the naming point, the new types already use `minDuration` / `isAttachmentMust` / `isCommentMust`, matching the entity.

frontend/src/community/leave/components/organisms/PolicyLeaveTypeForm/SettingToggleRow.tsx
hasalarootcode: `SwitchRow` and `DescribedSelection` are both MUI — `@mui/material`, `SxProps`, `styles.ts`. This feature is built on skapp-ui + Tailwind, so switching to those atoms would reintroduce MUI into the new flow. Keeping the local components and closing the a11y gaps directly instead (see the DurationOptionCard thread). The table already renders name+emoji via `LeaveTypeChip`.

frontend/src/community/leave/utils/validations.ts
hasalarootcode: A dependable client-side check is not available here: the list query is paginated at 20 per page, so the form only ever holds one page of names and would pass a duplicate that lives on another page — worse than no check, since it reads as validated. Leaving uniqueness to the backend, which does map to a specific duplicate-name toast (see the errorMessageKeys thread).

frontend/src/community/leave/components/organisms/PolicyLeaveTypeForm/PolicyLeaveTypeForm.tsx
hasalarootcode: `enabled: Boolean(id)` is already set on `useGetPolicyLeaveType`, so the detail query does not fire while `router.query` is still empty and nothing lands in the cache under an `undefined` id.

frontend/src/community/common/assets/languages/english/leaveModule.json
hasalarootcode: Both points already hold on this commit. The new keys are appended after the existing toast keys — `"enabled"` and `"disabled"` are still adjacent, nothing was inserted between them — and every new component uses the single `useTranslator("leaveModule", "leaveTypes")` scope, with no key for this screen under `leavePolicies`.

frontend/src/community/leave/components/molecules/PolicyLeaveTypesTable/PolicyLeaveTypesTable.tsx
hasalarootcode: Confirmed 0-based on both sides. skapp-ui documents `currentPage` as "Current active page (0-based)" and `onChange` as receiving the new page 0-based, so passing `page` straight through in both directions is correct — no off-by-one. Note this differs from the community MUI wrapper, which takes 0-based but renders `currentPage + 1` internally.

frontend/src/community/leave/utils/validations.ts
hasalarootcode: Backend agrees on 100 — `PolicyLeaveTypeConstant.MAX_NAME_LENGTH = 100` on #1691, enforced in `PolicyLeaveTypeValidationUtil`. Keeping both the `maxLength` attribute and the schema rule: the attribute is the typing guard, the rule still fires for pasted or programmatically set values, and it is now covered by tests.

frontend/src/community/leave/constants/policyLeaveTypeConstants.ts
hasalarootcode: It was removed from a previous pr and now I implemented again

frontend/src/community/leave/components/organisms/PolicyLeaveTypeForm/PolicyLeaveTypeForm.tsx
hasalarootcode: Guarded. `isEditMode && router.isReady && !policyLeaveTypeId` now redirects to `ROUTES.LEAVE.LEAVE_TYPES` via `router.replace` and renders nothing, so an `/leave/types/edit` URL with no id can no longer submit as a create.

frontend/src/community/leave/components/organisms/PolicyLeaveTypeForm/PolicyLeaveTypeForm.tsx
hasalarootcode: not needed

frontend/src/community/leave/utils/policyLeaveTypes/policyLeaveTypeUtils.ts
hasalarootcode: Added

frontend/src/community/leave/components/molecules/PolicyLeaveTypesTable/PolicyLeaveTypesTable.tsx
hasalarootcode: Fixed

frontend/src/community/leave/components/organisms/PolicyLeaveTypeForm/PolicyLeaveTypeForm.tsx
hasalarootcode: Fixed

frontend/src/community/leave/components/organisms/PolicyLeaveTypeForm/DurationOptionCard.tsx
hasalarootcode: Not needed

frontend/src/community/leave/utils/policyLeaveTypes/policyLeaveTypeUtils.ts
hasalarootcode: Fixed

frontend/src/community/leave/utils/validations.ts
hasalarootcode: Fixed

frontend/src/community/leave/components/organisms/PolicyLeaveTypeForm/PolicyLeaveTypeForm.tsx
hasalarootcode: Not valid

frontend/src/community/leave/api/PolicyLeaveTypeApi.ts
hasalarootcode: Fixed

frontend/src/community/leave/api/utils/QueryKeys.ts
```diff
@@ -276,6 +276,28 @@ export const reportsQueryKeys = {
...
+export const policyLeaveTypeQueryKeys = {
+  ALL: ["policy-leave-type-settings"],
+  SEARCH: function (
+    searchKeyword: string,
+    isActive: boolean | undefined,
+    page: number,
+    size: number
+  ) {
+    return [
+      ...(this?.ALL || []),
```
thusala: directly use policy-leave-type-settings

frontend/src/community/leave/api/utils/QueryKeys.ts
```diff
@@ -276,6 +276,28 @@ export const reportsQueryKeys = {
...
+      "search-policy-leave-types",
+      searchKeyword,
+      isActive,
+      page,
+      size
+    ];
+  },
+  DETAIL: function (id: number) {
+    return [...(this?.ALL || []), "policy-leave-type", id];
```
thusala: directly use policy-leave-type-settings

frontend/src/community/leave/api/PolicyLeaveTypeApi.ts
```diff
@@ -0,0 +1,187 @@
...
+): Promise<PolicyLeaveTypeSettingsType> => {
+  const response = await authFetch.get<PolicyLeaveTypeMutationResponse>(
+    policyLeaveTypeEndPoints.GET_POLICY_LEAVE_TYPE(id)
+  );
+
+  return response.data.results[0];
+};
+
+export const useGetPolicyLeaveType = (
+  id: number | undefined
```
thusala: remove undefined in here. it doesnt become undefined, if so please explain

frontend/src/community/leave/api/PolicyLeaveTypeApi.ts
```diff
@@ -0,0 +1,187 @@
...
+): UseQueryResult<PolicyLeaveTypeSettingsType> => {
+  return useQuery({
+    enabled: Boolean(id),
+    queryKey: policyLeaveTypeQueryKeys.DETAIL(id as number),
+    queryFn: () => getPolicyLeaveType(id as number)
```
thusala: remove casting if u remove undefined

frontend/src/community/leave/api/PolicyLeaveTypeApi.ts
```diff
@@ -0,0 +1,187 @@
...
+      queryClient.invalidateQueries({
+        queryKey: leavePolicyQueryKeys.POLICY_LEAVE_TYPES
+      });
+      onSuccess();
+    },
+    onError
+  });
+};
+
+const changePolicyLeaveTypeStatus = ({
```
thusala: shall we keep the methods for deactive and activate?

frontend/src/community/leave/components/molecules/PolicyLeaveTypeActivationToggleButton/PolicyLeaveTypeActivationToggleButton.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+  const canManageLeavePolicies = useCanManageLeavePolicies();
+
+  const { setToastMessage } = useToast();
+
+  const [isConfirmModalOpen, setIsConfirmModalOpen] = useState(false);
+
+  const { data: policyLeaveType } = useGetPolicyLeaveType(policyLeaveTypeId);
+
+  const isActive = policyLeaveType?.isActive ?? true;
```
thusala: do we need fallback?

frontend/src/community/leave/components/molecules/PolicyLeaveTypesTable/PolicyLeaveTypesTable.tsx
```diff
@@ -0,0 +1,171 @@
...
+  const translateText = useTranslator("leaveModule", "leaveTypes");
+
+  const router = useRouter();
+
+  const canManageLeavePolicies = useCanManageLeavePolicies();
+
+  const [page, setPage] = useState<number>(0);
+
+  const { data: policyLeaveTypesPage, isLoading } = useSearchPolicyLeaveTypes({
+    searchKeyword: "",
```
thusala: do not we need to have the search functionality in here? if so we have to use a state right?

frontend/src/community/leave/components/molecules/PolicyLeaveTypesTable/PolicyLeaveTypesTable.tsx
```diff
@@ -0,0 +1,171 @@
...
+                recordName: policyLeaveType.name
+              })}
+            />
+          );
+        }
+      }
+    ];
+  }, [translateText, canManageLeavePolicies, handleEditPolicyLeaveType]);
+
+  const totalPages = policyLeaveTypesPage?.totalPages ?? 0;
```
thusala: remove fallback

frontend/src/community/leave/components/organisms/PolicyLeaveTypeForm/DurationOptionCard.tsx
```diff
@@ -0,0 +1,72 @@
...
+        isError ? "ring-1 ring-semantic-red-accent" : ""
+      }`}
+      onClick={onSelect}
+      onKeyDown={handleKeyDown}
+      role="checkbox"
+      tabIndex={0}
+      aria-checked={isSelected}
+      aria-label={title}
+      aria-describedby={describedBy}
+      {...{ [DURATION_OPTION_ATTRIBUTE]: true }}
```
thusala: why do we need  {...{ [DURATION_OPTION_ATTRIBUTE]: true }}?

frontend/src/community/leave/components/organisms/PolicyLeaveTypeForm/PolicyLeaveTypeForm.tsx
```diff
@@ -0,0 +1,381 @@
...
+import { policyLeaveTypeValidationSchema } from "~community/leave/utils/validations";
+import { useCommonEnterpriseStore } from "~enterprise/common/store/commonStore";
+
+import DurationOptionCard, {
+  DURATION_OPTION_ATTRIBUTE
+} from "./DurationOptionCard";
+import SettingToggleRow from "./SettingToggleRow";
+
+const MIN_DURATION_ERROR_ID = "policy-leave-type-min-duration-error";
+const MIN_DURATION_GROUP_LABEL_ID = "policy-leave-type-min-duration-label";
```
thusala: move to const file

frontend/src/community/leave/constants/policyLeaveTypeConstants.ts
```diff
@@ -0,0 +1,16 @@
+/**
```
thusala: remove comments

frontend/src/community/leave/hooks/useIsLeavePoliciesEnabled.ts
```diff
@@ -0,0 +1,15 @@
...
+
```
thusala: remove comments

frontend/src/community/leave/hooks/usePolicyLeaveTypeFormSubmit.ts
```diff
@@ -0,0 +1,138 @@
...
+interface PolicyLeaveTypeFormSubmit {
+  submitPolicyLeaveType: (formValues: PolicyLeaveTypeFormDataType) => void;
+  isSubmitting: boolean;
+}
+
+/**
+ * Owns the create/update mutations for a policy leave type together with the
+ * toast, analytics and quick setup side effects the legacy leave type flow
+ * performs, so the form component stays layout only.
+ */
```
thusala: remove comments

frontend/src/community/leave/hooks/usePolicyLeaveTypeFormSubmit.ts
```diff
@@ -0,0 +1,138 @@
...
+        isEdit
+          ? "editLeaveTypeSuccessToastDescription"
+          : "addLeaveTypeSuccessToastDescription"
+      ]),
+      isIcon: true
+    });
+
+    setLeaveTypeFormDirty(false);
+
+    await router.push(ROUTES.LEAVE.LEAVE_TYPES);
```
thusala, line 81: why do we need await in here?

frontend/src/community/leave/hooks/usePolicyLeaveTypeFormSubmit.ts
```diff
@@ -0,0 +1,138 @@
...
+    useAddPolicyLeaveType(() => onMutationSuccess(false), onMutationError);
+
+  const { mutate: updatePolicyLeaveType, isPending: isUpdatePending } =
+    useUpdatePolicyLeaveType(() => onMutationSuccess(true), onMutationError);
+
+  const submitPolicyLeaveType = (
+    formValues: PolicyLeaveTypeFormDataType
+  ): void => {
+    const payload: PolicyLeaveTypePayloadType =
+      mapPolicyLeaveTypeFormToPayload(formValues);
```
thusala, line 117: fix formatting issue

frontend/src/community/leave/utils/policyLeaveTypes/policyLeaveTypeUtils.ts
```diff
@@ -0,0 +1,104 @@
...
+  LEAVE_ERROR_POLICY_LEAVE_TYPE_ALREADY_EXISTS
+} from "~community/common/constants/errorMessageKeys";
+import { LeaveDurationTypes } from "~community/leave/enums/LeaveTypeEnums";
+import {
+  PolicyLeaveTypeFormDataType,
+  PolicyLeaveTypePayloadType
+} from "~community/leave/types/PolicyLeaveTypeTypes";
+
+interface PolicyLeaveTypeErrorData {
+  results?: { messageKey?: string }[];
```
thusala: use a proper type

thusala (changes requested):

frontend/src/community/leave/components/organisms/PolicyLeaveTypeForm/PolicyLeaveTypeForm.tsx
hasalarootcode: Fixed

frontend/src/community/leave/components/molecules/PolicyLeaveTypesTable/PolicyLeaveTypesTable.tsx
hasalarootcode: Fixed

frontend/src/community/leave/components/molecules/PolicyLeaveTypesTable/PolicyLeaveTypesTable.tsx
hasalarootcode: Already handled

frontend/src/community/leave/api/PolicyLeaveTypeApi.ts
hasalarootcode: Not needed

frontend/src/community/leave/components/molecules/PolicyLeaveTypesTable/PolicyLeaveTypesTable.tsx
hasalarootcode: Fixed

frontend/src/community/leave/components/molecules/PolicyLeaveTypeActivationToggleButton/PolicyLeaveTypeActivationToggleButton.tsx
hasalarootcode: Not needed

frontend/src/community/leave/api/utils/QueryKeys.ts
hasalarootcode: Fixed

frontend/src/community/leave/api/PolicyLeaveTypeApi.ts
hasalarootcode: Fixed

frontend/src/community/leave/api/PolicyLeaveTypeApi.ts
hasalarootcode: Fixed

frontend/src/community/leave/api/utils/QueryKeys.ts
hasalarootcode: Fixed

frontend/src/community/leave/api/PolicyLeaveTypeApi.ts
hasalarootcode: Fixed

frontend/src/community/leave/components/molecules/PolicyLeaveTypeActivationToggleButton/PolicyLeaveTypeActivationToggleButton.tsx
hasalarootcode: Fixed

frontend/src/community/leave/components/molecules/PolicyLeaveTypesTable/PolicyLeaveTypesTable.tsx
hasalarootcode: Dont need that removed

frontend/src/community/leave/components/molecules/PolicyLeaveTypesTable/PolicyLeaveTypesTable.tsx
hasalarootcode: Removed

frontend/src/community/leave/components/organisms/PolicyLeaveTypeForm/DurationOptionCard.tsx
hasalarootcode: removed

frontend/src/community/leave/components/organisms/PolicyLeaveTypeForm/PolicyLeaveTypeForm.tsx
hasalarootcode: Fixed

frontend/src/community/leave/constants/policyLeaveTypeConstants.ts
hasalarootcode: Removed

frontend/src/community/leave/hooks/useIsLeavePoliciesEnabled.ts
hasalarootcode: removed

frontend/src/community/leave/hooks/usePolicyLeaveTypeFormSubmit.ts
hasalarootcode: Fixed

frontend/src/community/leave/hooks/usePolicyLeaveTypeFormSubmit.ts
hasalarootcode: below two has a squence

frontend/src/community/leave/hooks/usePolicyLeaveTypeFormSubmit.ts
hasalarootcode: already formateed

frontend/src/community/leave/utils/policyLeaveTypes/policyLeaveTypeUtils.ts
hasalarootcode: Fixed

shakila-kamalasena (dismissed):

thusala (dismissed):

frontend/src/community/leave/api/PolicyLeaveTypeApi.ts
```diff
@@ -0,0 +1,216 @@
...
+    queryFn: () =>
+      getPolicyLeaveTypes({
+        isActive,
+        page,
+        size
+      })
+  });
+};
+
+const getActivePolicyLeaveTypes = async (): Promise<PolicyLeaveTypesPage> => {
```
akilarootcode: remove this

frontend/src/community/leave/components/molecules/PolicyLeaveTypeActivationToggleButton/PolicyLeaveTypeActivationToggleButton.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+  const canManageLeavePolicies = useCanManageLeavePolicies();
+
+  const { setToastMessage } = useToast();
+
+  const [isConfirmModalOpen, setIsConfirmModalOpen] = useState(false);
+
+  const { data: policyLeaveType } = useGetPolicyLeaveType(policyLeaveTypeId);
+
+  const isActive = !!policyLeaveType?.isActive;
```
akilarootcode: disable button if page is loading

frontend/src/community/leave/components/molecules/PolicyLeaveTypesTable/PolicyLeaveTypesTable.tsx
```diff
@@ -0,0 +1,168 @@
...
+    ];
+  }, [translateText, canManageLeavePolicies, handleEditPolicyLeaveType]);
+
+  const handlePageChange = useCallback((selectedPage: number): void => {
+    setPage(selectedPage);
+  }, []);
+
+  return (
+    <div className="mt-4 flex flex-col gap-4">
+      <Table<TableRow>
```
akilarootcode: use Deshan's table

frontend/src/community/leave/components/organisms/PolicyLeaveTypeForm/PolicyLeaveTypeForm.tsx
```diff
@@ -0,0 +1,379 @@
...
+    touched,
+    setFieldValue,
+    setFieldError,
+    handleChange,
+    handleSubmit
+  } = formik;
+
+  useEffect(() => {
+    setLeaveTypeFormDirty(dirty);
+  }, [dirty, setLeaveTypeFormDirty]);
```
akilarootcode: remove setLeaveTypeFormDirty

frontend/src/community/leave/constants/policyLeaveTypeConstants.ts
```diff
@@ -0,0 +1,13 @@
+export const MAX_POLICY_LEAVE_TYPE_NAME_LENGTH = 100;
+
+export const POLICY_LEAVE_TYPES_PAGE_SIZE = 20;
+
+// The policy leave type list endpoint returns every record unpaginated when a
```
akilarootcode: remove comment

akilarootcode (changes requested):

frontend/src/community/leave/api/PolicyLeaveTypeApi.ts
hasalarootcode: Fixed

frontend/src/community/leave/components/molecules/PolicyLeaveTypeActivationToggleButton/PolicyLeaveTypeActivationToggleButton.tsx
hasalarootcode: Fixed

frontend/src/community/leave/components/molecules/PolicyLeaveTypesTable/PolicyLeaveTypesTable.tsx
hasalarootcode: Fixed

frontend/src/community/leave/components/organisms/PolicyLeaveTypeForm/PolicyLeaveTypeForm.tsx
hasalarootcode: Fixed

frontend/src/community/leave/constants/policyLeaveTypeConstants.ts
hasalarootcode: Fixed

akilarootcode (approved):

PR #1693, opened 2026-07-31
[Feat] Hikvision ISAPI event receiver for fingerprint attendance

akilarootcode:
## Summary

Adds Hikvision terminal support to fingerprint attendance. Hikvision devices do not speak ZKTeco ADMS, so they get their own authenticated receiver (`POST /hikvision/event`) that parses the ISAPI event notification the terminal pushes and writes the same raw punch row the ZKTeco `/iclock` path writes — so `applyPunches`, end-of-day finalization, the device allow-list, the source enum and the timezone handling are all reused unchanged rather than duplicated.

Target device: **Hikvision DS-K1T804AMF**.

- `TimeMessageConstant` + `community/messages/time-messages.properties` — service-level validation messages for the Hikvision webhook credentials (`TIME_ERROR_ADMS_DEVICE_WEBHOOK_CREDENTIALS_REQUIRED`, `TIME_ERROR_ADMS_DEVICE_WEBHOOK_USERNAME_ALREADY_EXISTS`).
- `attendanceModule.json` — device-type label, vendor names, webhook field labels and helper copy, and three new validation messages for the Fingerprint Devices modal.

## Feature scope (all repos)

This feature spans multiple repos on branch `feat/hikvision-isapi-event-receiver`:

| Repo | What changed |
|------|--------------|
| `SkappHQ/skapp` (super) | Service-level validation messages for webhook credentials; device-type / webhook / vendor labels |
| `rootcodelabs/skapp-ep-be-src` | The receiver, ISAPI event parser, device-credential allow-list, vendor column, `/iclock` vendor gate |
| `rootcodelabs/skapp-ep-be-tests` | Receiver integration tests + device registration validation tests |
| `rootcodelabs/skapp-ep-fe-src` | Device-type dropdown and webhook credential fields in Fingerprint Devices |
| `rootcodelabs/skapp-migrations` | `adms_device` vendor + webhook columns, `adms_attendance_log.event_serial_no` |

## Architecture / flow

```mermaid
flowchart LR
    D["Hikvision DS-K1T804AMF<br/>ISAPI httpHosts push"] -->|"POST /hikvision/event<br/>Basic auth"| C["HikvisionEventController"]
    C --> T["Tenant from host subdomain<br/>tenant.fingerprint.skapp.dev"]
    T --> A{"Registered device?<br/>vendor + webhook_username<br/>+ secret match"}
    A -->|no| E["401 + WWW-Authenticate<br/>device keeps event buffered"]
    A -->|yes| P["HikvisionEventParser<br/>JSON / XML / multipart"]
    P --> H{"AccessControllerEvent<br/>naming a person?"}
    H -->|"heartBeat / no person"| OK1["200, nothing stored"]
    H -->|yes| DUP{"event serialNo<br/>already stored?"}
    DUP -->|yes| OK2["200, duplicate dropped"]
    DUP -->|no| L["adms_attendance_log<br/>shared raw punch log"]
    L --> S["AdmsTimeRecordSyncService.applyPunches<br/>(unchanged, shared with ZKTeco)"]
    S --> R["time_record / time_slot<br/>clock_in_source = FINGERPRINT_DEVICE"]
    R --> F["End-of-day finalization<br/>(unchanged, source-agnostic)"]
```

```mermaid
flowchart LR
    Z["ZKTeco F18"] -->|"/iclock/cdata<br/>serial number, no auth"| ZG["vendor gate:<br/>ZKTECO rows only"]
    ZG --> LOG["adms_attendance_log"]
    HK["Hikvision DS-K1T804AMF"] -->|"/hikvision/event<br/>Basic auth, no serial in body"| LOG
    LOG --> DOM["shared attendance domain"]
```

## Screenshots

None — the UI change is two extra fields plus a dropdown in an existing modal, and there is no physical terminal on hand to capture an end-to-end punch.

- **AcsEvent polling fallback** (`POST /ISAPI/AccessControl/AcsEvent`) — for terminals that cannot push, and to backfill downtime. Follow-up.
- **skapp pushing `httpHosts` to the device** — registration is manual (device web UI or a one-off ISAPI `PUT`) in this slice; automating it needs an outbound Digest client and reachability *to* each device.
- **Honouring `attendanceStatus` / minor event codes** — direction is ignored for now (same call as the ZKTeco path); the codes are stored on the raw log so a later slice can use them.
- **Person enrolment from skapp** and **captured event images** — out of scope.
- **Renaming the `adms_*` tables** to vendor-neutral names — accepted cost of reusing them (see the note below); not worth the migration churn now.
- Unrelated in-progress work sharing the local checkout (PM rich-text, OAuth authorization server, onboarding copy) was deliberately kept out; this branch was cut fresh from `develop`.

- **Submodule hash / pointer bumps** are intentionally excluded. Gitlink updates are done at **deployment time**, not in feature PRs.
- One whitespace-only formatting fix to `people/service/impl/EpProbationInternalServiceImpl.java` is included in the be-src PR **only because `develop` currently violates `spring-javaformat:validate` on that file**, which fails the build before any test runs. It is not feature work.

## Notes for reviewers

- **Why reuse `adms_device` / `adms_attendance_log` instead of new Hikvision tables?** Finalization groups the raw log by pin, and the allow-list, devices UI and `TimeRecordSource.FINGERPRINT_DEVICE` are already vendor-neutral — reusing them means zero new code downstream of event capture. The `adms_` prefix is now a historical name.
- **No vendor backfill.** `adms_device` has no rows, so the column is simply added nullable; a device with no vendor is read as ZKTeco by both the `/iclock` gate and the devices UI, which also covers any row created before this lands.
- **Why Basic auth rather than allow-listing a serial number?** An ISAPI event body carries no device serial; `deviceID` is frequently empty and `macAddress` is firmware-dependent. Credentials also make this receiver authenticated, unlike `/iclock`.
- **Why accept XML and multipart?** Firmwares are documented to ignore the configured `parameterFormatType`, and switch to multipart once picture upload is on. JSON-only would fail silently and totally on such a device.
- `/hikvision/event` is registered in all four filter lists (`EPSecurityConfig`, `EpJwtAuthFilter.PUBLIC_URLS`, `TenantFilter.EXCLUDED_PATHS`, `EpTestSecurityConfig`).
- The receiver carries `@Profile("!ep-prd")`, matching `AdmsController` — a one-line flip when it goes to prod.

## Testing

| Check | Ran? | Result |
|-------|------|--------|
| Unit tests | yes | **pass** — full backend suite 610 tests green, including 7 new receiver integration tests (JSON / XML / multipart punch, wrong + missing credentials → 401, heartbeat dropped, duplicate suppressed) and 6 new device-registration validation tests |
| Formatter (BE only) | yes | **pass** — `spring-javaformat:validate` and `checkstyle:check` clean (after the pre-existing `develop` violation noted above) |
| tsc + lint (BE only) | n/a | Java BE has no tsc; checkstyle covers it. Per convention, formatter/tsc/lint are not run on FE repos |
| e2e (Playwright) | **no** | Not run. The shared suite drives a browser against a running local stack, which would serve a different checkout, so it would not exercise this path. See the gap below |

### Known verification gap

This has **not** been exercised against a physical DS-K1T804AMF — none was available. The protocol shapes come from the Hikvision Europe listening-mode guide, the TPP ISAPI wiki and three independent working implementations against this device family, and are covered by the integration tests above. Remaining gate before rollout: a live smoke test — register a device, point its HTTP listening host at `https://<tenant>.fingerprint.skapp.dev/hikvision/event` with format JSON and auth basic, then punch and confirm the record.

## Related PRs

- [`rootcodelabs/skapp-ep-be-src`#732](https://github.com/rootcodelabs/skapp-ep-be-src/pull/732) — the receiver, ISAPI parser, device allow-list and `/iclock` vendor gate
- [`rootcodelabs/skapp-ep-be-tests`#32](https://github.com/rootcodelabs/skapp-ep-be-tests/pull/32) — receiver integration tests and registration validation tests
- [`rootcodelabs/skapp-ep-fe-src`#764](https://github.com/rootcodelabs/skapp-ep-fe-src/pull/764) — device type and webhook credential fields in Fingerprint Devices
- [`rootcodelabs/skapp-migrations`#198](https://github.com/rootcodelabs/skapp-migrations/pull/198) — `adms_device` vendor + webhook columns, `event_serial_no`
- [`SkappHQ/skapp`#1693](https://github.com/SkappHQ/skapp/pull/1693) — validation messages and UI labels

**Merge order:** migrations (#198) first so the columns exist, then be-src (#732) with be-tests (#32), then fe-src (#764) and the super repo (#1693).


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds two backend error message constants/properties for Hikvision fingerprint-device webhook credentials plus the matching English i18n keys for the vendor dropdown, webhook fields and validation messages. All new keys are actually consumed (AdmsServiceImpl.applyVendorSettings, admsDeviceValidations.ts, FingerprintDeviceForm.tsx, admsDeviceUtils.ts), only the English locale exists so there are no missing translations, and the backend paths are covered by AdmsServiceImplTest. The main concerns are a reachable edit-flow gap where the new 'leave blank to keep current password' placeholder is misleading, and the fact that both new backend messages can never actually be shown to the user because the mutation onError handlers discard the server response.

Found **18** new issue(s): 🔴 6 important, 🟡 7 suggestion(s), 🟣 5 nit(s)

frontend/src/community/common/assets/languages/english/attendanceModule.json
```diff
@@ -41,12 +41,24 @@
       "editTitle": "Edit Fingerprint Device",
       "nameLabel": "Device name",
       "serialLabel": "Serial number",
+      "vendorLabel": "Device type",
+      "webhookUsernameLabel": "Webhook username",
+      "webhookPasswordLabel": "Webhook password",
+      "webhookPasswordUnchangedPlaceholder": "Leave blank to keep the current password",
```
ThinuwanW, line 55: 🤖 **Claude** · 🔴 **Important**

`webhookPasswordUnchangedPlaceholder` ("Leave blank to keep the current password") is rendered whenever `isEdit` is true (FingerprintDeviceForm.tsx:114-121), not only when a stored webhook password actually exists. Concrete broken flow: edit an existing ZKTeco device and switch the vendor dropdown to Hikvision. `isEdit` is true, so (a) this placeholder tells the user they may leave the password blank, and (b) admsDeviceValidations.ts:79-90 skips the `required` check on `webhookPassword` for edits. But the persisted device has no webhook password — AdmsServiceImpl.applyVendorSettings nulls `webhookUsername`/`webhookPassword` for non-Hikvision vendors — so the submit omits `webhookPassword` (FingerprintDeviceModal.tsx:84-92), `StringUtils.defaultIfBlank(null, null)` yields null, and the backend throws the new TIME_ERROR_ADMS_DEVICE_WEBHOOK_CREDENTIALS_REQUIRED. The user sees only the generic "Something went wrong" toast with no indication which field is wrong. Fix: base both the placeholder and the Yup `required` rule on whether the saved device already has Hikvision credentials rather than on `isEdit` alone — e.g. pass a `hasStoredWebhookPassword = isEdit && device?.vendor === DeviceVendor.HIKVISION && Boolean(device?.webhookUsername)` flag from the modal into both the form and `admsDeviceValidations`, and require the password whenever that flag is false.

backend/src/main/resources/community/messages/time-messages.properties
```diff
@@ -32,6 +32,8 @@ api.error.time.time-request-permission.not.found=Current user don't have attenda
 api.error.time.pause-resume-not-available=Pause and resume are not available; only clock in and clock out are allowed
 api.error.time.adms.device.not-found=Fingerprint device not found
 api.error.time.adms.device.serial-already-exists=A fingerprint device with this serial number already exists
+api.error.time.adms.device.webhook-credentials-required=Hikvision devices require a webhook username and password
+api.error.time.adms.device.webhook-username-already-exists=A fingerprint device with this webhook username already exists
```
ThinuwanW, line 36: 🤖 **Claude** · 🔴 **Important**

Neither of the two error messages added by this PR can ever reach a user. `useCreateAdmsDevice`/`useUpdateAdmsDevice` (frontend/src/enterprise/attendance/api/AttendanceApi.ts:116-160) declare `onError: () => void` — no error argument — and FingerprintDeviceModal.showErrorToast always renders the static `fingerprintDeviceModal.toasts.errorDescription` ("Something went wrong. Please try again."). Since the client-side Yup rules catch the normal duplicate-username / missing-username cases, the only paths that actually hit these backend messages (vendor switch on edit, and a concurrent-create race on the webhook username) are exactly the paths where the user gets no actionable text. Fix: widen the `onError` signature to accept the axios error and surface the server message (the ResponseEntityDto/ErrorResponse message from `error.response.data`), or map the returned message key onto the existing `validations.webhookUsernameExists` string plus a new i18n string for the credentials-required case, before falling back to the generic toast.

backend/src/main/resources/community/messages/time-messages.properties
```diff
@@ -32,6 +32,8 @@ api.error.time.time-request-permission.not.found=Current user don't have attenda
 api.error.time.pause-resume-not-available=Pause and resume are not available; only clock in and clock out are allowed
 api.error.time.adms.device.not-found=Fingerprint device not found
 api.error.time.adms.device.serial-already-exists=A fingerprint device with this serial number already exists
+api.error.time.adms.device.webhook-credentials-required=Hikvision devices require a webhook username and password
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`api.error.time.adms.device.webhook-credentials-required=Hikvision devices require a webhook username and password` hardcodes a single vendor name into the message text while the trigger condition is `vendor != DeviceVendor.ZKTECO`-shaped logic against a `DeviceVendor` enum that will grow — the message goes stale the moment a second webhook-based vendor is added. It also lumps two distinct fields into one message, whereas the frontend already distinguishes `webhookUsernameRequired` from `webhookPasswordRequired`. Suggest either a vendor-agnostic wording ("A webhook username and password are required for this device type") or splitting into two constants so the message granularity matches the frontend and the user learns which field is missing.

backend/src/main/java/com/skapp/community/timeplanner/constant/TimeMessageConstant.java
```diff
@@ -19,6 +19,9 @@ public enum TimeMessageConstant implements MessageConstant {
 	TIME_ERROR_ATTENDANCE_CONFIG_NOT_FOUND("api.error.time.attendance.config.not.found"),
 	TIME_ERROR_ADMS_DEVICE_NOT_FOUND("api.error.time.adms.device.not-found"),
 	TIME_ERROR_ADMS_DEVICE_SERIAL_ALREADY_EXISTS("api.error.time.adms.device.serial-already-exists"),
+	TIME_ERROR_ADMS_DEVICE_WEBHOOK_CREDENTIALS_REQUIRED("api.error.time.adms.device.webhook-credentials-required"),
```
ThinuwanW, line 22: 🤖 **Claude** · 🟡 **Suggestion**

These webhook constants describe an enterprise-only capability — the only consumer is com.skapp.enterprise.timeplanner.service.impl.AdmsServiceImpl, and `DeviceVendor`/`AdmsDevice` live entirely under `com.skapp.enterprise`. The enterprise module already has the right home for this: `EpTimeMessageConstant` plus `enterprise/messages/ep-time-messages.properties`, which is only added to `spring.messages.basename` in the ep profiles. Because `community/messages/time-messages` is loaded by the plain community profiles (application-prd.yml / application-non-prd.yml), community-only deployments now ship vendor-specific strings for a feature they don't have. This follows the precedent set by the pre-existing TIME_ERROR_ADMS_DEVICE_* entries rather than introducing the leak, so it is a judgement call — but consider moving the whole ADMS message block (constants + properties lines 33-36 and the three adms success keys) into EpTimeMessageConstant/ep-time-messages.properties instead of extending it.

frontend/src/community/common/assets/languages/english/attendanceModule.json
```diff
@@ -31,7 +31,7 @@
     "fingerprintDevicesError": "Couldn't load fingerprint devices. Please try again.",
     "fingerprintDeviceAvailable": "Available",
     "fingerprintDeviceOffline": "Offline",
-    "fingerprintDeviceLastSynced": "Last synced",
+    "fingerprintDeviceLastSynced": "Last synced ",
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Changing the value to "Last synced " adds a semantically meaningful trailing space to a translation string in order to separate two adjacent JSX expression containers in FingerprintDeviceList.tsx:67-69 (JSX inserts no whitespace between `{a}{b}`). That is fragile: the space is invisible in review and diffs, JSON formatters and translation-management tooling routinely trim it, and translators for future locales have no way to know it is load-bearing. Fix in the component instead — either interpolate (`"fingerprintDeviceLastSynced": "Last synced {{time}}"` with `translateText(["fingerprintDeviceLastSynced"], { time })`) or build one string in JSX, e.g. render `{`${translateText(["fingerprintDeviceLastSynced"])} ${formatDeviceTimestamp(device.lastSyncAt) ?? translateText(["fingerprintDeviceNeverSynced"])}`}` — and revert this value to "Last synced".

frontend/src/community/common/assets/languages/english/attendanceModule.json
```diff
@@ -41,12 +41,24 @@
       "editTitle": "Edit Fingerprint Device",
       "nameLabel": "Device name",
       "serialLabel": "Serial number",
+      "vendorLabel": "Device type",
+      "webhookUsernameLabel": "Webhook username",
+      "webhookPasswordLabel": "Webhook password",
+      "webhookPasswordUnchangedPlaceholder": "Leave blank to keep the current password",
+      "webhookHelperText": "Set these as the HTTP listening host credentials on the device so it can send events to SKAPP.",
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`webhookHelperText` writes the product name as "SKAPP" in all caps. Every other brand reference across the English language files (~64 occurrences in common.json, attendanceModule.json, peopleModule.json, etc.) uses "Skapp" — this is the only all-caps instance. Change "...send events to SKAPP." to "...send events to Skapp." for consistency.

frontend/src/community/common/assets/languages/english/attendanceModule.json
```diff
@@ -41,12 +41,24 @@
       "editTitle": "Edit Fingerprint Device",
       "nameLabel": "Device name",
       "serialLabel": "Serial number",
+      "vendorLabel": "Device type",
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`vendorLabel` is displayed as "Device type", but `AdmsDeviceType` (frontend/src/enterprise/attendance/api/AttendanceApi.ts:29) and `AdmsDevice.deviceType` already have a distinct `deviceType` field holding the device-reported model string. Labelling the vendor selector "Device type" makes the two concepts indistinguishable in UI copy and in future bug reports. Consider "Device vendor" or "Manufacturer" for the label, or rename the key to `deviceTypeLabel` if the wording is deliberate so key and copy at least agree.

frontend/src/community/common/assets/languages/english/attendanceModule.json
```diff
@@ -41,12 +41,24 @@
...
+        "hikvision": "Hikvision"
+      },
       "validations": {
         "nameRequired": "Device name is required",
         "serialRequired": "Serial number is required",
-        "serialExists": "A device with this serial number already exists"
+        "serialExists": "A device with this serial number already exists",
+        "webhookUsernameRequired": "Webhook username is required",
+        "webhookUsernameExists": "A device with this webhook username already exists",
+        "webhookPasswordRequired": "Webhook password is required"
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`webhookPasswordRequired` can never fire for the one case that most needs it, and the resulting save is rejected by the server with no usable feedback. In `frontend/src/enterprise/attendance/utils/admsDeviceValidations.ts:79-90` the password rule is gated on `vendor === HIKVISION && !isEdit`, i.e. it is skipped for *every* edit. But `isEdit` is not the same as "this device already has a stored webhook password": editing a device whose persisted vendor is ZKTECO and switching Device type to Hikvision passes validation with a blank password, `FingerprintDeviceModal.handleSubmit` then omits `webhookPassword` from the payload, and `AdmsServiceImpl.applyVendorSettings` (backend/.../enterprise/timeplanner/service/impl/AdmsServiceImpl.java:187-191) falls back to `device.getWebhookPassword()` which is null and throws `TIME_ERROR_ADMS_DEVICE_WEBHOOK_CREDENTIALS_REQUIRED`. Because `useUpdateAdmsDevice`'s `onError` only shows the generic "Something went wrong" toast, the admin gets no field error and no way to discover the cause — the form is simply unsubmittable. Fix: derive a single flag (e.g. `hasStoredWebhookPassword = isEdit && device?.vendor === DeviceVendor.HIKVISION`) in `FingerprintDeviceModal`, pass it to `admsDeviceValidations` in place of `isEdit`, and use the same flag for the `required` prop on the password field. Cleanest is to have the backend expose `hasWebhookPassword` on `AdmsDeviceResponseDto` so the client stops inferring it from the vendor.

frontend/src/community/common/assets/languages/english/attendanceModule.json
```diff
@@ -31,7 +31,7 @@
     "fingerprintDevicesError": "Couldn't load fingerprint devices. Please try again.",
     "fingerprintDeviceAvailable": "Available",
```
ThinuwanW, line 32: 🤖 **Claude** · 🔴 **Important**

The `fingerprintDeviceAvailable` / `fingerprintDeviceOffline` badge is effectively hard-wired to "Available". `isDeviceAvailable` in `frontend/src/enterprise/attendance/utils/admsDeviceUtils.ts:9-13` returns true whenever `lastOnlineAt` is a valid timestamp that is not in the future — so any device that has *ever* handshaked (ZKTeco) or posted one event (Hikvision) shows as Available forever, even months after being unplugged. "Offline" is only ever shown for a device that has never connected. Since ingestion silently stops when a device dies, this badge is exactly the signal an attendance admin would rely on. Fix: compare against a staleness window instead of `DateTime.now()`, e.g. `lastOnline >= DateTime.now().minus({ minutes: THRESHOLD })` with the threshold in a named constant (Hikvision heartbeats and the ZKTeco poll interval give you a sane value). Note also that `lastOnlineAt` is serialised from a `LocalDateTime` with no offset while the helper parses it as `{ zone: "utc" }`; once a threshold is applied, any non-UTC server clock will skew the result, so consider returning an instant/offset from the API.

frontend/src/community/common/assets/languages/english/attendanceModule.json
```diff
@@ -31,7 +31,7 @@
     "fingerprintDevicesError": "Couldn't load fingerprint devices. Please try again.",
```
ThinuwanW, line 31: 🤖 **Claude** · 🔴 **Important**

`fingerprintDevicesError` is a dead translation key and the query's failure path is unhandled. `FingerprintDevices` (frontend/src/enterprise/configurations/components/organisms/FingerprintDevices/FingerprintDevices.tsx:20) destructures only `data` and `isLoading` from `useGetAdmsDevices`, so when the request fails `devices` is `undefined` and the component renders `fingerprintDevicesEmpty` — telling the admin "No fingerprint devices added yet" when devices may in fact exist and be recording punches. Nothing in the codebase references `fingerprintDevicesError` (grep confirms the only other hit is the test's negative assertion). Fix: also destructure `isError` and render the `fingerprintDevicesError` message (plus a retry) ahead of the empty state. The existing test asserting that the error text is *absent* passes vacuously and should be replaced with an `isError: true` case.

backend/src/main/java/com/skapp/community/timeplanner/constant/TimeMessageConstant.java
ThinuwanW, line 22: 🤖 **Claude** · 🔴 **Important**

📁 **File-level observation**

Unhandled parse failure on the new webhook path returns a 500 (backend checklist item 1). In `com.skapp.enterprise.timeplanner.util.HikvisionEventParser.readAlert` (HikvisionEventParser.java:96) the XML branch is carefully wrapped in try/catch and returns null on failure, but the JSON branch calls `objectMapper.readTree(payload)` bare. With Jackson 3 (`tools.jackson`) every parse failure is an *unchecked* `JacksonException`, so any device that posts a non-JSON/truncated body that happens to not start with `<` propagates the exception out of `HikvisionEventServiceImpl.receiveEvent` and out of the controller as a 500 instead of the intended 200/401 — and a device that retries the same malformed payload will loop. Fix: wrap the `readTree` call in the same try/catch as `readXml` (log and return null) so both branches degrade identically, and add an integration test posting a malformed body alongside the existing JSON/XML/multipart cases in `HikvisionEventControllerIntegrationTest`.

frontend/src/community/common/assets/languages/english/attendanceModule.json
```diff
@@ -41,12 +41,24 @@
...
       "nameLabel": "Device name",
       "serialLabel": "Serial number",
+      "vendorLabel": "Device type",
+      "webhookUsernameLabel": "Webhook username",
+      "webhookPasswordLabel": "Webhook password",
+      "webhookPasswordUnchangedPlaceholder": "Leave blank to keep the current password",
+      "webhookHelperText": "Set these as the HTTP listening host credentials on the device so it can send events to SKAPP.",
       "saveButton": "Save",
       "cancelButton": "Cancel",
+      "vendors": {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The vendor→label mapping is duplicated and silently wrong for any future vendor. `FingerprintDeviceForm` builds `vendorOptions` with the two `vendors.*` keys spelled out, and `admsDeviceUtils.getDeviceVendorLabel` (frontend/src/enterprise/attendance/utils/admsDeviceUtils.ts:15-21) re-implements the same mapping as a ternary that falls back to the `zkteco` label for *anything* that is not `HIKVISION` — so adding a third vendor requires edits in two files and, if one is missed, the list mislabels real devices as ZKTeco. The codebase already has a convention for this: a `Record<Enum, string>` mapper (e.g. `community/leave/utils/leaveTypes/LeaveTypeUtils.ts:8`) or a lowercased enum lookup (`community/crm/components/molecules/SidePanelTaskInfo/SidePanelTaskInfo.tsx:63` uses `translateText(["priorityOptions", value.toLowerCase()])`). Since the `vendors` keys here are already the lowercased enum values, replace both sites with one exported `Record<DeviceVendor, string>` (or a single `getVendorLabelKey` helper) that `vendorOptions` and `getDeviceVendorLabel` both consume.

frontend/src/community/common/assets/languages/english/attendanceModule.json
```diff
@@ -41,12 +41,24 @@
       "editTitle": "Edit Fingerprint Device",
       "nameLabel": "Device name",
       "serialLabel": "Serial number",
+      "vendorLabel": "Device type",
+      "webhookUsernameLabel": "Webhook username",
+      "webhookPasswordLabel": "Webhook password",
```
ThinuwanW, line 54: 🤖 **Claude** · 🟡 **Suggestion**

The webhook password field has no `autoComplete`, which combines badly with the blank-means-unchanged semantics. `FingerprintDeviceForm` renders `type="password"` (FingerprintDeviceForm.tsx:106-128) with no `autoComplete` attribute, so browsers and password managers treat it as a login password: on the edit modal they can autofill the admin's own saved credentials into the field, which makes it non-blank, which makes `handleSubmit` send it, which makes `applyVendorSettings` overwrite the device's stored secret — after which the device's Basic-auth events start being rejected with 401 and punches stop being ingested, with no visible error anywhere. This is the only `type="password"` field in `frontend/src`, and the codebase already uses `autoComplete="off"` where autofill is unwanted (e.g. `common/components/molecules/Form/Form.tsx:10`, `SearchBox.tsx:121`). Set `autoComplete="new-password"` (or `"off"`) on this field, and ideally on the webhook username too.

frontend/src/community/common/assets/languages/english/attendanceModule.json
```diff
@@ -41,12 +41,24 @@
...
+      "webhookPasswordUnchangedPlaceholder": "Leave blank to keep the current password",
+      "webhookHelperText": "Set these as the HTTP listening host credentials on the device so it can send events to SKAPP.",
       "saveButton": "Save",
       "cancelButton": "Cancel",
+      "vendors": {
+        "zkteco": "ZKTeco",
+        "hikvision": "Hikvision"
+      },
       "validations": {
```
ThinuwanW, line 59: 🤖 **Claude** · 🟡 **Suggestion**

None of the new vendor/webhook front-end logic is covered by tests. `FingerprintDevices.test.tsx` is the only test file under `frontend/src/enterprise` and it explicitly mocks away `FingerprintDeviceModal` and `FingerprintDeviceList`, so the branchy new code — the vendor-conditional Yup schema in `admsDeviceValidations.ts`, the uniqueness `test()` callbacks behind these `webhookUsernameExists`/`webhookPasswordRequired` messages, `getDeviceVendorLabel`, and `isDeviceAvailable` — ships untested while the backend equivalents got thorough coverage in `AdmsServiceImplTest`. The two pure util modules are the cheapest, highest-value gap: add `admsDeviceValidations` unit tests (Hikvision requires username/password on create, password optional when a password is already stored, duplicate username rejected, credentials not required for ZKTeco) and an `admsDeviceUtils` test for the availability threshold and vendor label.

backend/src/main/resources/community/messages/time-messages.properties
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

The device webhook secret is stored reversibly rather than hashed. `AdmsDevice.webhookPassword` (backend/.../enterprise/timeplanner/model/AdmsDevice.java:52) uses `FieldEncryptionConverter`, so the plaintext secret is recoverable by anything holding the app's encryption key, and `AdmsServiceImpl.applyVendorSettings` decrypts and re-persists it on every update just to implement keep-existing. Nothing needs the plaintext: the device presents it on every request, and `FingerprintDeviceUtils.matchesSecret` only ever compares it. Storing a `PasswordEncoder` (BCrypt) hash and switching `matchesSecret` to `passwordEncoder.matches` would remove the recoverable-secret exposure at no functional cost (the field is already never returned in `AdmsDeviceResponseDto`, which is good). If reversible storage is deliberate — e.g. a planned "reveal credentials" UI — please note that in the field's javadoc so the next reader doesn't 'fix' it.

frontend/src/community/common/assets/languages/english/attendanceModule.json
```diff
@@ -41,12 +41,24 @@
       "editTitle": "Edit Fingerprint Device",
       "nameLabel": "Device name",
       "serialLabel": "Serial number",
+      "vendorLabel": "Device type",
+      "webhookUsernameLabel": "Webhook username",
```
ThinuwanW, line 53: 🤖 **Claude** · 🟣 **Nit**

`webhookHelperText` is rendered as a bare paragraph above the two webhook inputs (FingerprintDeviceForm.tsx:89-91) with no programmatic association, so screen-reader users focusing the `webhookUsernameLabel`/`webhookPasswordLabel` fields never hear the instruction that these must match the device's HTTP listening host credentials — which is the one piece of information needed to configure the device correctly. Give the paragraph an id and reference it from both inputs via `aria-describedby` (or pass it through the `InputField` helper/description prop if the design-system component exposes one).

frontend/src/community/common/assets/languages/english/attendanceModule.json
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

The vendor Dropdown wires up an error message that can never render (dead code, checklist item 6). `FingerprintDeviceForm` passes `errorMessage={errors?.vendor}` and `required` for the `vendorLabel` dropdown, but `admsDeviceValidations` defines no rule for `vendor` and the field always has a default value, so `errors.vendor` is permanently `undefined`. Additionally, the established usage of this `Dropdown` also sets `variant="primary-error"` when showing an error (see `community/leave/components/organisms/LeavePolicyWizard/EntitlementSetupStep.tsx:96-102`), which is missing here — so even if a rule were added the field would not visibly turn red. Either drop the `errorMessage` prop or add the missing `variant` toggle if a vendor rule is intended.

backend/src/main/resources/community/messages/time-messages.properties
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

`AdmsServiceImpl` has no logger, so every rejection driven by these two new message keys is silent server-side. The class carries no `@Slf4j` and none of its methods have the conventional `log.info("methodName: execution started")` entry line (backend guidelines: logging section + checklist item 8), while the sibling `HikvisionEventServiceImpl` does have `@Slf4j` and logs rejected events. Since the front end collapses all of these into one generic toast, a support engineer investigating "I can't save my Hikvision device" has nothing at all to go on. Add `@Slf4j` plus entry logging on `addDevice`/`updateDevice`/`deleteDevice` and a `log.warn` (device id + username, never the password) at each throw site in `applyVendorSettings` and `resolveUniqueSerialNumber`.

frontend/src/community/common/assets/languages/english/attendanceModule.json
akilarootcode: Fixed. The placeholder and the Yup `required` rule now key off `hasStoredWebhookPassword` (`isEdit && device.vendor === HIKVISION`) instead of `isEdit`, and the vendor dropdown is locked after creation so the flag can no longer disagree with what is actually stored.

backend/src/main/resources/community/messages/time-messages.properties
akilarootcode: Fixed. `useCreateAdmsDevice`/`useUpdateAdmsDevice` now take `(error: ErrorResponse)` and the toast renders `error.response.data.results[0].message` when the server sends one, falling back to the generic copy.

backend/src/main/resources/community/messages/time-messages.properties
akilarootcode: Fixed. Reworded to "A webhook username and password are required for this device vendor" so it does not go stale as `DeviceVendor` grows. Kept it as one message because the form already flags the specific missing field inline.

backend/src/main/java/com/skapp/community/timeplanner/constant/TimeMessageConstant.java
akilarootcode: Not addressing. The pre-existing `TIME_ERROR_ADMS_DEVICE_*` block already lives here, so moving only the two new keys would split the group across two modules. Relocating the whole ADMS block is worth its own PR.

frontend/src/community/common/assets/languages/english/attendanceModule.json
akilarootcode: Fixed. Reverted the value to "Last synced" and `FingerprintDeviceList` now builds the whole sentence in one template literal, so no whitespace is load-bearing in the JSON.

frontend/src/community/common/assets/languages/english/attendanceModule.json
akilarootcode: Fixed. Now reads "…send events to Skapp." to match every other brand reference.

frontend/src/community/common/assets/languages/english/attendanceModule.json
akilarootcode: Fixed. Relabelled "Device vendor" so it no longer reads as the same thing as the device-reported `deviceType` field.

frontend/src/community/common/assets/languages/english/attendanceModule.json
akilarootcode: Fixed. The password rule is now gated on `hasStoredWebhookPassword` rather than `isEdit`, and the vendor is immutable after creation, so the ZKTeco-to-Hikvision-on-edit path that produced the unsubmittable form no longer exists.

frontend/src/community/common/assets/languages/english/attendanceModule.json
akilarootcode: Fixed. `isDeviceAvailable` now compares against `DEVICE_OFFLINE_THRESHOLD_MINUTES` (15 minutes, i.e. 30 missed polls at the ADMS `Delay=30`). Left `lastOnlineAt` as a `LocalDateTime` for now — changing the wire format is a separate API change.

frontend/src/community/common/assets/languages/english/attendanceModule.json
akilarootcode: Fixed. `FingerprintDevices` destructures `isError` and renders `fingerprintDevicesError` ahead of the empty state, so a failed fetch no longer reads as "no devices".

backend/src/main/java/com/skapp/community/timeplanner/constant/TimeMessageConstant.java
akilarootcode: Fixed. `readTree` moved behind a `readJson` helper with the same try/catch shape as `readXml`, so both content types degrade to a clean 200. Added endpoint tests for malformed JSON, a non-JSON/non-XML body and malformed XML.

frontend/src/community/common/assets/languages/english/attendanceModule.json
akilarootcode: Fixed. Both sites now read one `DEVICE_VENDOR_LABEL_KEYS: Record<DeviceVendor, string[]>`, so adding a vendor is a compile error rather than a silent mislabel.

frontend/src/community/common/assets/languages/english/attendanceModule.json
akilarootcode: Fixed. `autoComplete="new-password"` on the password field and `autoComplete="off"` on the username; `InputField` extends `InputHTMLAttributes` so both forward to the input.

frontend/src/community/common/assets/languages/english/attendanceModule.json
akilarootcode: Not addressing. The rulebook is integration-API-tests-only, so the new behaviour is pinned in `AdmsServiceImplTest` and `HikvisionEventControllerIntegrationTest` instead of in FE unit tests.

backend/src/main/resources/community/messages/time-messages.properties
akilarootcode: Not changing. The column is already `FieldEncryptionConverter`-encrypted at rest, and BCrypt verification on every punch and heartbeat would add real per-request cost on a high-frequency device endpoint. Skipping the javadoc per the no-comments rule.

frontend/src/community/common/assets/languages/english/attendanceModule.json
akilarootcode: Fixed. The helper paragraph now carries an id and both webhook inputs reference it via `aria-describedby`.

frontend/src/community/common/assets/languages/english/attendanceModule.json
akilarootcode: Fixed. Dropped the dead `errorMessage` prop and added `ariaLabel`. The disabled-on-edit state uses the repo's `variant="primary-disabled"` pattern, since this `Dropdown` exposes no `disabled` prop.

backend/src/main/resources/community/messages/time-messages.properties
akilarootcode: Not addressing. The rulebook rules out routine entry logging, and the actual gap you identified — the admin having nothing to go on — is fixed by surfacing the server message in the toast.

akilarootcode:
### Scope added: fingerprint devices are now linked to work locations

This PR has grown to include the work-location feature (it was briefly a separate stacked PR, now folded in here so it ships together rather than waiting behind this one).

**What that adds**
- A **Work location** selector in the device modal, for both ZKTeco and Hikvision. Optional, so a device without one behaves exactly as before.
- The device list shows the location next to vendor and serial.
- When a punch becomes the clock-in or the clock-out, the punching device's work location is stamped onto that attendance record.

**Storage** — `att_time_record_location` already existed as the 1:1 "where" table for a `TimeRecord`, so it gains `clock_in_work_location_id` and `clock_out_work_location_id`. Recording both ends separately means clocking in at one site and out at another is captured properly, instead of one location per day overwriting the first. No column was added to `adms_attendance_log`: the punch already carries `device_id`, and the historically meaningful location is the one captured on the attendance record at sync time.

**By design** — the end-of-day `finalizeTenantDay` scheduler records an automatic clock-out, not a real punch at a device, so it deliberately leaves the work location unset.

**Verified against a running local stack** — a ZKTeco punch at 08:00 from a Colombo device produced `clock_in 08:00 at Colombo Office`, and a Hikvision punch at 18:30 from a Kandy device produced `clock_out 18:30 at Kandy Office`. An unknown work location is rejected with 400. 57 backend tests and 4 end-to-end UI flows pass.

yuhanga-rtc (approved):

hasalarootcode (approved):

shakila-kamalasena (approved):

PR #1694, opened 2026-07-31
feat: enhance CRM contact lookup with companyId filtering and update related components

dulsara-manakal:
### Summary

When adding a deal from a **Company** page, the contact dropdown showed **all contacts** in the system, not just the contacts of that company. This PR fixes that.

## What we changed

**Backend**
- Added a new `companyId` filter to the contact lookup API (`GET /v1/crm/contact/lookup`).
- Now you can search contacts and filter by company at the same time.

**Frontend**
- When you open a company's side panel and click "Add Deal," the contact dropdown now only shows contacts from that company.
- Cleaned up the `useGetCrmContacts` hook to take one simple object of params instead of many separate arguments (easier to read, follows our other hooks' pattern).
- Updated the "Add Deal" form's validation to check fields on blur, matching our other forms.

### How to test

- [ ] Open a company panel → Add Deal → contact dropdown only shows that company's contacts.
- [ ] Search still works when filtering by company.
- [ ] Adding a deal from other places (Contacts page, Tasks, global Add Deal) still shows contacts as before.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

The core change — adding an optional `companyId` filter to the CRM contact lookup (DTO → Criteria predicate → typed `CrmContactLookupParams` on the frontend) — is implemented cleanly and consistently: the predicate reuses the existing LEFT-joined company alias so contacts without a company are correctly excluded, the count query applies the same predicates, all four `useGetCrmContacts` call sites were migrated to the new params-object signature, and the params-object query key matches the existing `GET_DEALS`/`RELATED_TASKS_BY_PARAMS` convention. Deal creation derives `deal.company` from `contact.company`, so deals added from the company side panel will correctly appear in that company's deals list. My main concerns are an unrelated Formik validation-mode flip in `SidePanelAddDeal` that leaves stale validation errors on screen, and the absence of an integration test for the new filter despite an existing test for the sibling `dealId` filter.

Found **11** new issue(s): 🔴 1 important, 🟡 7 suggestion(s), 🟣 3 nit(s)

frontend/src/community/crm/components/molecules/SidePanelAddDeal/SidePanelAddDeal.tsx
```diff
@@ -129,8 +135,8 @@ const SidePanelAddDeal: FC<Props> = ({ onClose, defaultContact }) => {
   const formik = useFormik<CrmInlineDealAddFormTypes>({
     initialValues,
     validationSchema: inlineAddDealValidations(translateText),
-    validateOnChange: true,
-    validateOnBlur: false,
+    validateOnChange: false,
```
ThinuwanW, line 142: 🤖 **Claude** · 🔴 **Important**

Flipping to `validateOnChange: false` / `validateOnBlur: true` makes validation errors sticky, and the blur half is inert. Nothing in this component (or in `AddDealContactSearch`/`SubTaskInput`) is wired to `formik.handleBlur`/`setFieldTouched`, so `validateOnBlur: true` can never fire — the only remaining validation trigger is `submitForm`. Concretely: after a failed save, `errors.contactId` is set and `touched.contactId` is true, so the contact field renders in the error state; `handleContactChange` then calls `setFieldValue("contactId", ...)` whose `shouldValidate` defaults to `validateOnChange` (now false), so the field stays red even after a valid contact is picked. Same for the deal name: `errorMessage={formik.errors.name}` keeps showing "required/invalid chars" while the user types a valid name. Note the sibling form `AddDealSidePanel.tsx` still uses `validateOnChange: true, validateOnBlur: false`, so this also introduces an inconsistency, and the change is unrelated to this PR's companyId scope. Fix: either revert to `validateOnChange: true, validateOnBlur: false`, or — if the intent was to stop typing from wiping the server-side DEAL_EXISTS error set via `setFieldError("name", ...)` — keep `validateOnChange: false` but clear the specific field error in the change handlers (e.g. a `clearError` helper calling `formik.setFieldError(field, undefined)`, as `TaskModalForm` already does) and drop `validateOnBlur: true` or actually wire an `onBlur` that calls `setFieldTouched(field, true)`.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
```diff
@@ -163,6 +163,10 @@ private <T> List<Predicate> buildLookupPredicates(CriteriaBuilder cb, CriteriaQu
 			predicates.add(contact.get(CrmContact_.id).in(dealSub));
 		}
 
+		if (filterDto.getCompanyId() != null) {
```
ThinuwanW, line 166: 🤖 **Claude** · 🟡 **Suggestion**

New `companyId` lookup filter has no test coverage. `CrmContactControllerIntegrationTest` already has a dedicated test for the sibling filter (`getContactsLookup_FilterByDealId_ReturnsOnlyLinkedContact`) plus tests asserting that company-less contacts are returned and counted consistently. Add an equivalent `getContactsLookup_FilterByCompanyId_ReturnsOnlyCompanyContacts` case that seeds two companies plus a contact with no company, calls `/lookup?companyId=...`, and asserts both `totalItems` (exercises `getLookupTotalCount`, which builds the join separately) and the returned items — the company-less contact must be excluded, which is the behavior most at risk if someone later changes the LEFT fetch join.

frontend/src/community/crm/components/molecules/SidePanelAddDeal/SidePanelAddDeal.tsx
```diff
@@ -48,9 +50,13 @@ const SidePanelAddDeal: FC<Props> = ({ onClose, defaultContact }) => {
...
+  const contactLookupParams: CrmContactLookupParams = {
+    searchKeyword: debouncedContactSearch,
+    size: DEFAULT_LOOKUP_PAGE_SIZE,
+    companyId
+  };
   const { data: contactLookupData } = useGetCrmContacts(
-    debouncedContactSearch,
-    DEFAULT_LOOKUP_PAGE_SIZE,
+    contactLookupParams,
     true
```
ThinuwanW, line 64: 🤖 **Claude** · 🟡 **Suggestion**

`enabled` is hardcoded `true`, so the lookup always fires even when it cannot be used. When this component is rendered from `ContactSidePanel` with `defaultContact` set, `AddDealContactSearch` renders a read-only `InputField` (no dropdown at all), yet a `/contact/lookup` request still goes out on mount with an empty keyword. Now that the params object exists, gate it on runtime state instead, e.g. `useGetCrmContacts(contactLookupParams, !selectedContact)` — the query then activates only when the search dropdown is actually visible (including after the user clears the pre-selected contact).

backend/src/main/java/com/skapp/community/crmplanner/payload/request/CrmContactFilterDto.java
```diff
@@ -15,4 +15,6 @@ public class CrmContactFilterDto {
 
 	private Long dealId;
 
+	private Long companyId;
```
ThinuwanW, line 18: 🤖 **Claude** · 🟣 **Nit**

With `companyId` added, `CrmContactFilterDto` (page, size, searchKeyword, dealId, companyId) is now a strict superset of `CrmContactMetricRequestDto` (page, size, searchKeyword, companyId) — the two request DTOs are duplicates apart from `dealId`. Consider extracting a shared base (e.g. `CrmContactBaseFilterDto` holding page/size/searchKeyword/companyId) that both extend, so future filter fields and any future validation annotations (`@Min`/`@Max` on `size`, which is currently unbounded on both endpoints) only need to be added once.

frontend/src/community/crm/components/organisms/CompanySidePanel/CompanySidePanel.tsx
```diff
@@ -218,6 +218,7 @@ const CompanySidePanel: FC = () => {
         return (
           <SidePanelDealSection
             deals={selectedCompany?.deals ?? []}
+            companyId={selectedCompanyId}
```
ThinuwanW, line 221: 🤖 **Claude** · 🟣 **Nit**

Scoping the contact dropdown to `selectedCompanyId` means that for a company with zero contacts the inline add-deal form becomes a dead end: the dropdown only renders `noResultsText` and `contactId` is required, so the user can never save and gets no explanation of why (previously any contact was selectable, which was the bug being fixed). Consider surfacing a company-specific empty message or a link to create a contact for this company when `companyId` is set and the lookup returns no items.

frontend/src/community/crm/api/utils/QueryKeys.ts
```diff
@@ -49,11 +50,10 @@ export const contactQueryKeys = {
     "crm-owners-lookup",
     searchKeyword
   ],
-  CONTACT_LOOKUP: (
-    searchKeyword: string,
-    size: number,
-    dealId?: number | null
-  ) => ["crm-contacts-lookup", searchKeyword, size, dealId],
+  CONTACT_LOOKUP: (params: CrmContactLookupParams) => [
```
ThinuwanW, line 53: 🤖 **Claude** · 🟡 **Suggestion**

The new object-based key is not normalized, so semantically identical lookups hash to different cache entries. TaskModalForm always sets `dealId` (it is `null` when no deal is chosen) while AddDealSidePanel and DealPropertiesSidebar omit the field entirely; React Query's `hashKey` drops `undefined` properties but keeps explicit `null`, so `{searchKeyword,size,dealId:null}` and `{searchKeyword,size}` become two cache entries that issue the *identical* HTTP request (both drop `dealId` via the `dealId != null` guard in `fetchContactsLookup`). That is a duplicate fetch whenever both screens are mounted. Fix by normalizing inside the factory, e.g. build the key from a stripped object: drop keys whose value is `null`/`undefined` before returning `["crm-contacts-lookup", normalized]`. Also worth noting the sibling `crmDealQueryKeys.DEAL_LOOKUP` in this same file kept positional args, so the two lookup factories now follow different conventions — consider migrating both or neither.

frontend/src/community/crm/api/ContactApi.ts
```diff
@@ -207,24 +208,28 @@ export const useGetOwnerLookup = (
...
+  });
+  return response?.data?.results?.[0];
+};
+
 export const useGetCrmContacts = (
-  searchKeyword: string,
-  size: number,
-  enabled: boolean,
-  dealId?: number | null
+  params: CrmContactLookupParams,
```
ThinuwanW, line 227: 🤖 **Claude** · 🟡 **Suggestion**

`useGetCrmContacts(params, enabled)` now mixes a params object with a trailing positional boolean, which is unlike every other hook in this file (`useGetOwnerLookup`, `useGetCrmOwners`, `useGetCompanyLookup`, `useGetContactById` all take flat args, and `enabled` is always last for a reason). More importantly, the refactor pushed the same four-line `contactLookupParams` literal into four components (SidePanelAddDeal, AddDealSidePanel, DealPropertiesSidebar, TaskModalForm), so each call site is now more verbose than before and the `searchKeyword`/`size` defaults are repeated. Consider either folding `enabled` into the argument object (`useGetCrmContacts({ searchKeyword, size, companyId, enabled })`) or exposing a small `useContactLookup({ searchTerm, companyId, dealId, enabled })` hook that owns the debounce + `DEFAULT_LOOKUP_PAGE_SIZE` default, so the four components just pass what actually varies.

frontend/src/community/crm/components/molecules/SidePanelDealSection/SidePanelDealSection.tsx
```diff
@@ -26,6 +26,7 @@ import DealAccordionItemSkeleton from "./DealAccordionItemSkeleton";
 interface Props {
   deals: DetailPanelDealResponseType[];
   defaultContact?: CrmContactLookup;
+  companyId?: number | null;
```
ThinuwanW, line 29: 🤖 **Claude** · 🟡 **Suggestion**

The new `companyId` prop is only supplied by one of the two consumers that enable the add-deal action. `CompanySidePanel` passes `selectedCompanyId`, but `ContactSidePanel` renders this same component with `defaultContact={contact}` and no `companyId`, even though the contact's company is available (`contact?.company?.id`). Result: the identical inline add-deal form scopes the contact dropdown to one company in the company panel but stays global in the contact panel, so if the user clears the prefilled contact there they can attach the deal to a contact from any company. (`TaskSidePanel` passes `showAddDealAction={false}`, so it is correctly unaffected.) Either pass `companyId={contact?.company?.id}` from `ContactSidePanel` for consistent behaviour, or add a short comment on the prop explaining that scoping is intentionally company-panel-only.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
ThinuwanW, line 166: 🤖 **Claude** · 🟡 **Suggestion**

This companyId predicate is now character-for-character identical to the one already in `buildPredicates` (lines 92-95, used by `findContacts`/`getContactTotalCount`). Two independent copies of the same filter in the same class will drift — e.g. if the company filter later needs to also exclude soft-deleted companies or fall back to contacts with no company, one builder will be updated and the other missed. Extract a tiny private helper (e.g. `private Optional<Predicate> companyIdPredicate(CriteriaBuilder cb, Join<CrmContact, CrmCompany> company, Long companyId)`) and call it from both builders. This is separate from the DTO-level overlap: it is the predicate construction, not the request objects.

backend/src/main/java/com/skapp/community/crmplanner/payload/request/CrmContactFilterDto.java
ThinuwanW, line 18: 🤖 **Claude** · 🟡 **Suggestion**

This filter DTO carries no bean-validation annotations and the controller binds it without `@Valid`, so the newly added `companyId` accepts any value including negatives (silently returning an empty page rather than a 400), and the pre-existing `size` is unbounded in both directions. `?size=0` or `?size=-1` reaches `PageRequest.of(filterDto.getPage(), filterDto.getSize())` in `CrmContactServiceImpl.getContactsLookup`, which throws `IllegalArgumentException`; that is not handled specifically in `GlobalExceptionHandler`, so it falls through to `@ExceptionHandler(Exception.class)` and returns 500 to the client instead of a 4xx. An unbounded upper limit also lets a caller request the whole contact table in one page. Add `@Positive` on `companyId` and `dealId`, `@Min(0)` on `page`, `@Min(1) @Max(<sane cap>)` on `size`, and `@Valid` on the `getContactsLookup` parameter so failures surface as the standard 422 `BindException` response.

frontend/src/community/crm/api/ContactApi.ts
```diff
@@ -207,24 +208,28 @@ export const useGetOwnerLookup = (
   });
 };
 
+const fetchContactsLookup = async (
```
ThinuwanW, line 211: 🤖 **Claude** · 🟣 **Nit**

The newly extracted `fetchContactsLookup` declares `Promise<CrmContactLookupResponseType>` but returns `response?.data?.results?.[0]`, which is `undefined` whenever the envelope is empty or the shape differs — the optional chaining is itself an admission that the value may be missing. The declared type therefore lies to consumers; it happens to be safe today only because all four call sites write `contactLookupData?.items ?? []`. Since this function is new code, type it `Promise<CrmContactLookupResponseType | undefined>` (and let `UseQueryResult<CrmContactLookupResponseType | undefined>` flow through `useGetCrmContacts`) so the optionality is enforced rather than assumed. Same applies to the neighbouring extracted fetchers if you touch them.

frontend/src/community/crm/components/molecules/SidePanelAddDeal/SidePanelAddDeal.tsx
dulsara-manakal: Valid but  `validateOnBlur` relies on each field component (SubTaskInput, AddDealContactSearch) calling formik's blur handler internally.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
dulsara-manakal: Test added

frontend/src/community/crm/components/molecules/SidePanelAddDeal/SidePanelAddDeal.tsx
dulsara-manakal: We need enable to be true always.

backend/src/main/java/com/skapp/community/crmplanner/payload/request/CrmContactFilterDto.java
dulsara-manakal: Keep these two separate since the usage is two different thing.

frontend/src/community/crm/components/organisms/CompanySidePanel/CompanySidePanel.tsx
dulsara-manakal: The contact dropdown explicitly shows no result test in the dropdown, so according to the requirement, this is correct and can not link create contact to dropdown.

frontend/src/community/crm/api/utils/QueryKeys.ts
dulsara-manakal: `AddDealSidePanel` has no concept of dealId at all it's not that the field is empty, it just doesn't exist there.

frontend/src/community/crm/api/ContactApi.ts
dulsara-manakal: This is intentional. The component knows its own context (search text, whether a company/deal is selected, when the query should even run), that's why it builds the params object itself.

frontend/src/community/crm/components/molecules/SidePanelDealSection/SidePanelDealSection.tsx
dulsara-manakal: No companyId here on purpose, if the user clears the prefilled contact, they should be free to pick a contact from any company.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
dulsara-manakal: These two are for separate endpoints (`/metrics` and `/lookup`) that happen to filter by company the same way today, but they're independent use cases and could reasonably extend later

backend/src/main/java/com/skapp/community/crmplanner/payload/request/CrmContactFilterDto.java
dulsara-manakal: No bean-validation annotations here on purpose, none of the other CRM filter DTOs (CrmCompanyFilterDto, CrmDealFilterDto, CrmTaskFilterDto, CrmContactMetricRequestDto, etc.) validate page/size/companyId either, so this stays consistent with the rest of the CRM module.

frontend/src/community/crm/api/ContactApi.ts
dulsara-manakal: It actually guards with an empty array fallback in the component. but the type mismatch is valid and will fix in CRM refactor.

frontend/src/community/crm/api/ContactApi.ts
```diff
@@ -207,24 +208,28 @@ export const useGetOwnerLookup = (
...
+  params: CrmContactLookupParams
+): Promise<CrmContactLookupResponseType> => {
+  const { searchKeyword, size, dealId, companyId } = params;
+  const response = await authFetch.get(contactEndpoints.CONTACT_LOOKUP, {
+    params: {
+      searchKeyword,
+      size,
+      ...(dealId != null && { dealId }),
+      ...(companyId != null && { companyId })
```
thusala: handle these from parent? we do not add conditions on api layer

frontend/src/community/crm/components/molecules/SidePanelAddDeal/SidePanelAddDeal.tsx
thusala: if it always true why do we use enable Prop? just remove it?

thusala (changes requested):

frontend/src/community/crm/components/molecules/SidePanelAddDeal/SidePanelAddDeal.tsx
dulsara-manakal: Removed it and make enable optional

frontend/src/community/crm/api/ContactApi.ts
dulsara-manakal: No need to check this dynamically, so remove this part.

thusala (dismissed):

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
dulsara-manakal: Remove this since relevant BE changes are already done and merged with the previous PR.

backend/src/main/java/com/skapp/community/crmplanner/payload/request/CrmContactFilterDto.java
dulsara-manakal: Remove this since relevant BE changes are already done and merged with the previous PR.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
dulsara-manakal: Remove this since relevant BE changes are already done and merged with the previous PR.

backend/src/main/java/com/skapp/community/crmplanner/payload/request/CrmContactFilterDto.java
dulsara-manakal: Remove this since relevant BE changes are already done and merged with the previous PR.

frontend/src/community/crm/types/CommonTypes.ts
```diff
@@ -124,6 +124,13 @@ export interface CrmContactLookupResponseType {
   totalPages: number;
 }
 
+export interface CrmContactLookupParams {
+  searchKeyword: string;
+  size: number;
+  dealId?: number | null;
```
akilarootcode: remove null

akilarootcode (changes requested):

frontend/src/community/crm/types/CommonTypes.ts
dulsara-manakal: Removed

akilarootcode (approved):

ThinuwanW (approved):

PR #1695, opened 2026-07-31
feat: accept add/remove deltas on the role restrictions endpoint

anusham-rc:
## What

Phase 2 of the Module Roles Restriction alignment — the write side. `PATCH /v1/roles/restrictions` now takes an **add/remove delta** and writes **only** the new `module_roles_restriction` table. The legacy `module_role_restriction` write is dropped.

> Base is `feat/skapp-role-restriction` (Phase 1, #1677) so this diff shows only Phase 2 work. Retarget to `develop` once the parent merges.

## Changes

- **`ModuleRoleRestrictionRequestDto`** — replaced `isAdmin` / `isManager` / `restrictions` (and all `@Schema`) with `module`, `add`, `remove`.
- **`RolesServiceImpl.updateRoleRestrictions`** — validate → read current baseline → apply `remove` → apply `add` → persist. Read-modify-write stays inside the existing `@Transactional`. Legacy `moduleRoleRestrictionDao.save(...)` write and the boolean derivation removed.
- **`PeopleMapper`** — deleted `roleRestrictionRequestDtoToRestrictRole` (only fed the legacy write).
- **`PeopleUtil.toRestrictionsString`** — serialize the restricted-role set to the stored string (write-side inverse of `parseRestrictions`).
- **`PeopleMessageConstant` + `people-messages.properties`** — four new validation messages.

## Validation (all → 400 via `ModuleException`)

- `module` null or `COMMON`
- a level in `add` that is not restrictable for the module
- `null` entries in either list
- duplicates within a list
- `add ∩ remove ≠ ∅`

Removal is intentionally **not** checked against restrictable roles, so a value stored before this validation existed (or by the sync job's lossy derivation) can always be cleared. `remove` applies before `add`, matching the People-skills precedent.

## Tests

Added `UpdateRoleRestrictionsTests` (14 cases): add/remove/remove-before-add, idempotent add & remove, lenient removal of a non-restrictable stored value, and each validation rule (400) plus the super-admin `@PreAuthorize` (403). Full suite: **27/27 pass**.

## Notes / follow-ups (not in this PR)

- `peopleMapper` and `moduleRoleRestrictionDao` are now unused fields on `RolesServiceImpl`; kept until Phase 5 so the `EpRolesServiceImpl` constructor keeps compiling.
- **Retiring the `skapp-migrations` sync job is step 1 of Phase 2** (separate repo) and must land before this deploys — otherwise a later sync run overwrites correct values with stale derived ones.
- Phase 2 **frontend** (delta modal) ships separately.


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR flips the role-restriction read path from the legacy `module_role_restriction` boolean table to `module_roles_restriction.restrictions` and reshapes the update contract from a full replace (`isAdmin`/`isManager`/`restrictions`) into a delta (`add`/`remove`), with new validation, message constants, and a solid 399-line integration test. The refactor itself is clean and the new tests are well targeted, but three issues need attention before production: the community frontend still sends the old payload and will silently no-op with a success toast, `PeopleUtil.parseRestrictions` can throw an unhandled `IllegalArgumentException` that turns into a 500 on several hot paths, and there is no Liquibase backfill from the legacy table so restrictions configured before the dual-write release stop being enforced. Remaining findings are dead fields, a lost-update race, missing bean validation, and lost Swagger docs.

Found **23** new issue(s): 🔴 5 important, 🟡 11 suggestion(s), 🟣 7 nit(s)

backend/src/main/java/com/skapp/community/peopleplanner/payload/request/ModuleRoleRestrictionRequestDto.java
```diff
@@ -2,30 +2,19 @@
...
-@Schema(description = "Data transfer object for restricting role assignments in a specific module.")
 public class ModuleRoleRestrictionRequestDto {
 
-	@Schema(description = "The module to which the role restriction is applied.", example = "ATTENDANCE")
 	private ModuleType module;
 
-	@Schema(description = "Flag indicating whether the Admin role is restricted for the specified module.",
-			example = "true")
-	private Boolean isAdmin;
+	private List<RoleLevel> add;
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Breaking API contract change with a silent no-op failure mode. This DTO drops `isAdmin`, `isManager` and `restrictions` and now only accepts `add`/`remove`, but the community frontend was not updated in this PR: `frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx` (handleSubmit) still builds `{ module, isAdmin, isManager }`, and `frontend/src/community/configurations/types/UserRolesTypes.ts` still declares `UserRoleRestrictionsType` with exactly those three fields. Spring Boot's default Jackson setup does not fail on unknown properties, so `isAdmin`/`isManager` are dropped, both `add` and `remove` deserialize to null, `resolveRestrictedRoles` returns the stored set unchanged, and the endpoint returns 200 `successful`. The UI then shows the success toast while nothing was persisted — users will believe they have restricted (or un-restricted) a role when they have not. Note this is worse than a 400 because it is invisible. Fix: ship the frontend change in the same PR (send `add`/`remove` derived from the checkbox deltas, and drive the checkbox list from the new `restrictableRoles` response field), or keep accepting the legacy fields for one release and translate them into a full replace inside `updateRoleRestrictions`.

backend/src/main/java/com/skapp/community/peopleplanner/util/PeopleUtil.java
```diff
@@ -91,6 +98,39 @@ public static List<EmployeeManager> filterManagersByAttendanceRoles(List<Employe
...
+		if (restrictions == null || restrictions.isBlank()) {
+			return roleLevels;
+		}
+
+		for (String restriction : restrictions.split(RESTRICTION_SEPARATOR)) {
+			String roleLevel = restriction.trim();
+			if (roleLevel.isEmpty()) {
+				continue;
+			}
+			roleLevels.add(RoleLevel.valueOf(roleLevel));
```
ThinuwanW, line 111: 🤖 **Claude** · 🔴 **Important**

`RoleLevel.valueOf(roleLevel)` throws an unhandled `IllegalArgumentException` for any token that is not a current enum constant. `restrictions` is a free-text `VARCHAR(255)` column (see `common-ddl-script-v1-create-table-module-roles-restriction.sql`), so a value written by a different edition/version, a hand-edited row, or a `RoleLevel` constant that is later renamed or removed will blow up here. `IllegalArgumentException` has no dedicated handler in `GlobalExceptionHandler`, so it falls through to `@ExceptionHandler(Exception.class)` and returns HTTP 500. The blast radius is large because `getRestrictedRoleLevels` is now on three hot paths: `getRestrictedRoleByModule`, `getAllowedRoles`/`processModuleRoles` (called for every module on every role-dropdown load), and `validateRestrictedRoleAssignment` (called 5-6 times on every employee create/update). One malformed row bricks all of them. Per the review checklist, this parse must be defensive: either skip unknown tokens with a `log.warn` (matching the existing lenient treatment of blanks and whitespace two lines above) or catch `IllegalArgumentException` and rethrow as a `ModuleException` with a new message constant.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
```diff
@@ -107,77 +106,108 @@ public EmployeeRole assignRolesToEmployee(EmployeeSystemPermissionsDto roleReque
...
+						new String[] { roleLevel.name() });
 			}
 		}
+	}
 
-		ModuleRolesRestriction moduleRolesRestriction = new ModuleRolesRestriction();
-		moduleRolesRestriction.setModule(module);
-		moduleRolesRestriction.setRestrictions(restrictedRoles.isEmpty() ? null : String.join(",", restrictedRoles));
-		return moduleRolesRestriction;
+	protected Set<RoleLevel> getRestrictedRoleLevels(ModuleType module) {
```
ThinuwanW, line 159: 🤖 **Claude** · 🔴 **Important**

Read path switched to `module_roles_restriction` with no data migration from the legacy `module_role_restriction` table. Before this PR, `getRestrictedRoleByModule` and `processModuleRoles` read `module_role_restriction` while `module_roles_restriction` was write-only (populated only by `updateRoleRestrictions` since the dual-write release). This PR makes the new table the sole source and there is no Liquibase changeset backfilling `is_admin`/`is_manager` into `restrictions` — `community/db/changelog` only contains the `CREATE TABLE` (changeset 00081). Every restriction configured before the dual-write release silently stops being enforced: `validateRestrictedRoleAssignment` returns false, so People Admins regain the ability to assign the restricted ADMIN/MANAGER roles. This is a silent security regression, and the new test `getRestrictions_LegacyTableRowOnly_ReturnsNoRestrictions` codifies it as expected behaviour. Add a Liquibase DML changeset that inserts into `module_roles_restriction (module, restrictions)` from `module_role_restriction`, mapping `is_admin=true` to `ADMIN` and `is_manager=true` to the module's secondary role (`SENDER` for ESIGN, `SALES_MANAGER` for CRM, else `MANAGER`) — the same mapping as `PeopleUtil.getSecondaryRestrictionRole`. Related behaviour change worth confirming: the old `processModuleRoles` used `Boolean.FALSE.equals(getIsAdmin())`, so a legacy row with a NULL `is_admin` blocked the ADMIN role; the new `!restrictedRoleLevels.contains(ADMIN)` allows it. Also note `updateRoleRestrictions` no longer writes the legacy table at all, so a rollback to the previous release would resurrect stale restrictions.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
```diff
@@ -107,77 +106,108 @@ public EmployeeRole assignRolesToEmployee(EmployeeSystemPermissionsDto roleReque
...
-			moduleRoleRestrictionRequestDto.setIsManager(
-					restrictions.contains(getSecondaryRestrictionRole(moduleRoleRestrictionRequestDto.getModule())));
-		}
+		validateRoleRestrictionRequest(moduleRoleRestrictionRequestDto);
 
-		ModuleRoleRestriction moduleRoleRestriction = peopleMapper
-			.roleRestrictionRequestDtoToRestrictRole(moduleRoleRestrictionRequestDto);
-		moduleRoleRestrictionDao.save(moduleRoleRestriction);
+		ModuleType module = moduleRoleRestrictionRequestDto.getModule();
+		Set<RoleLevel> restrictedRoles = resolveRestrictedRoles(moduleRoleRestrictionRequestDto);
```
ThinuwanW, line 107: 🤖 **Claude** · 🟡 **Suggestion**

Lost-update race on the read-modify-write in `updateRoleRestrictions`. `resolveRestrictedRoles` reads the current row via `getRestrictedRoleLevels` and then `save()` writes the merged set. Under the default READ_COMMITTED isolation, two concurrent PATCHes for the same module can each read the same baseline and the second write silently discards the first. This matters more than it did with the old full-replace contract, precisely because the request is now a delta on top of stored state (e.g. two admins each adding a different role, one add is lost). Fix by taking a pessimistic write lock on the row (a `@Lock(LockModeType.PESSIMISTIC_WRITE)` `findById` variant on `ModuleRolesRestrictionDao`) or adding a `@Version` column to `ModuleRolesRestriction` so the loser gets an optimistic-lock failure instead of silent data loss.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
```diff
@@ -107,77 +106,108 @@ public EmployeeRole assignRolesToEmployee(EmployeeSystemPermissionsDto roleReque
...
 		log.info("updateRoleRestrictions: execution started");
 
-		if (moduleRoleRestrictionRequestDto.getRestrictions() != null
-				&& !moduleRoleRestrictionRequestDto.getRestrictions().isEmpty()) {
-			List<RoleLevel> restrictions = moduleRoleRestrictionRequestDto.getRestrictions();
-			moduleRoleRestrictionRequestDto.setIsAdmin(restrictions.contains(RoleLevel.ADMIN));
-			moduleRoleRestrictionRequestDto.setIsManager(
-					restrictions.contains(getSecondaryRestrictionRole(moduleRoleRestrictionRequestDto.getModule())));
-		}
+		validateRoleRestrictionRequest(moduleRoleRestrictionRequestDto);
```
ThinuwanW, line 104: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Two injected fields are now dead. `peopleMapper` (line 70) was only used by the removed `roleRestrictionRequestDtoToRestrictRole` / `restrictRoleToRestrictRoleResponseDto` calls, and `moduleRoleRestrictionDao` (line 72) was only used by the removed legacy save and `findById`. Grep confirms neither is referenced anywhere else in this class, and `moduleRoleRestrictionDao` is now the last reference to the legacy table in production code (only the new integration test still uses it). Leaving them keeps two unused beans wired through `EpRolesServiceImpl`'s constructor and `super(...)` call (EpRolesServiceImpl.java:58-64), which is misleading about what the service depends on. Remove both fields and the corresponding constructor parameters in the enterprise submodule — unless you intend to keep writing the legacy table for rollback safety, in which case keep the DAO and restore the dual write (see the migration finding).

backend/src/main/java/com/skapp/community/peopleplanner/payload/request/ModuleRoleRestrictionRequestDto.java
```diff
@@ -2,30 +2,19 @@
...
 
 import java.util.List;
 
 @Getter
 @Setter
-@Schema(description = "Data transfer object for restricting role assignments in a specific module.")
 public class ModuleRoleRestrictionRequestDto {
 
-	@Schema(description = "The module to which the role restriction is applied.", example = "ATTENDANCE")
 	private ModuleType module;
```
ThinuwanW, line 14: 🤖 **Claude** · 🟡 **Suggestion**

No bean validation on the request DTO, and `RolesController.updateRoleRestrictions` (RolesController.java:50-51) has no `@Valid` on the `@RequestBody`. The null/COMMON module check is done manually in `validateRoleRestrictionRequest`, which the guidelines call out as service-layer validation duplicating what bean validation should catch. Add `@NotNull` to `module` here and `@Valid` to the controller's `@RequestBody` parameter, then keep only the COMMON-module rejection in the service. That also makes the constraint visible in the generated OpenAPI spec instead of only discoverable by triggering a 400.

backend/src/main/java/com/skapp/community/peopleplanner/payload/request/ModuleRoleRestrictionRequestDto.java
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

All `@Schema` documentation was removed from this DTO and none was added for the new fields, and `ModuleRoleRestrictionResponseDto`'s new `restrictions`/`restrictableRoles` are undocumented too. The new semantics are considerably less obvious than the old boolean pair and are not inferable from the field names: `add`/`remove` are deltas applied against the stored set (remove first, then add), omitting both is a no-op rather than a clear, the same role in both lists is a 400, and only the roles returned in `restrictableRoles` may appear in `add` (while `remove` accepts any role level). Re-add class- and field-level `@Schema(description=..., example=...)` capturing that contract so API consumers do not have to read the service to use the endpoint.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
```diff
@@ -107,77 +106,108 @@ public EmployeeRole assignRolesToEmployee(EmployeeSystemPermissionsDto roleReque
...
+			.orElseGet(() -> EnumSet.noneOf(RoleLevel.class));
 	}
 
-	private RoleLevel getSecondaryRestrictionRole(ModuleType module) {
-		return switch (module) {
-			case ESIGN -> RoleLevel.SENDER;
-			case CRM -> RoleLevel.SALES_MANAGER;
-			default -> RoleLevel.MANAGER;
-		};
+	protected List<RoleLevel> getRestrictableRoles(ModuleType module) {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`getRestrictableRoles` derives the allowlist from `initializeRolesForModule()`, which is the "role levels a user can be assigned in this module" map, not a "modules that support restriction" map — the two are not the same thing. In the community edition that map has no ESIGN entry, so `getRestrictableRoles(ESIGN)` returns `List.of()` and any `{"module":"ESIGN","add":["ADMIN"]}` now fails with 400, where the old endpoint accepted it and persisted the restriction. The community UI still renders ESIGN admin/sender checkboxes (`RestrictedUserRolesModal.tsx` getRestrictableRoles maps ESIGN to `["isAdmin", "isSender"]`), so this is a reachable regression, not a theoretical one. Either confirm ESIGN restriction is genuinely enterprise-only and drop it from the community UI, or source the restrictable set from a dedicated module→restrictable-roles definition rather than reusing the assignable-roles map.

backend/src/test/java/com/skapp/community/peopleplanner/controller/v1/RolesControllerIntegrationTest.java
```diff
@@ -0,0 +1,399 @@
+package com.skapp.community.peopleplanner.controller.v1;
```
ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Good coverage of the two endpoints, but three gaps in exactly the areas this PR changed. (1) `processModuleRoles` / `GET /v1/roles` (`getAllowedRoles`) is the largest behavioural change in the diff — it now derives `isAdminAllowed`/`isManagerAllowed` from the parsed restriction set instead of the legacy booleans, including a semantic flip for NULL `is_admin` — and has no test at all. Add a case that stores `ADMIN` for PEOPLE and asserts the PEOPLE admin role is filtered out of the allowed-roles response for a non-super-admin. (2) No test that a stored value which is not a valid `RoleLevel` (e.g. `"ADMIN,BOGUS"`) is handled gracefully — today it produces a 500 (see the `parseRestrictions` finding); a test here would lock in the fix. (3) No test that restrictions actually block role assignment end to end via `validateRestrictedRoleAssignment`, which is the security-relevant consumer of this data.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
```diff
@@ -107,77 +106,108 @@ public EmployeeRole assignRolesToEmployee(EmployeeSystemPermissionsDto roleReque
...
-		}
+		Set<RoleLevel> restrictedRoleLevels = getRestrictedRoleLevels(module);
 
-		ModuleRoleRestriction moduleRoleRestriction = restrictedRole.get();
-		ModuleRoleRestrictionResponseDto moduleRoleRestrictionResponseDto = peopleMapper
-			.restrictRoleToRestrictRoleResponseDto(moduleRoleRestriction);
+		ModuleRoleRestrictionResponseDto moduleRoleRestrictionResponseDto = new ModuleRoleRestrictionResponseDto();
+		moduleRoleRestrictionResponseDto.setModule(module);
+		moduleRoleRestrictionResponseDto.setRestrictions(List.copyOf(restrictedRoleLevels));
+		moduleRoleRestrictionResponseDto.setRestrictableRoles(getRestrictableRoles(module));
```
ThinuwanW, line 193: 🤖 **Claude** · 🟣 **Nit**

`getRestrictedRoleByModule` is doing double duty as the public read endpoint and as an internal helper for `validateRestrictedRoleAssignment` (line 562), and it got heavier in this PR: it now also calls `getRestrictableRoles`, which rebuilds the whole `EnumMap` via `initializeRolesForModule()` on every invocation. `validateRoles` calls `validateRestrictedRoleAssignment` five times (six in the enterprise override), so every employee create/update now performs 5-6 `findById` queries plus 5-6 full module-map constructions purely to read two booleans it then discards. Have `validateRestrictedRoleAssignment` call `getRestrictedRoleLevels(moduleType)` directly and check `contains(RoleLevel.ADMIN)` / `contains(PeopleUtil.getSecondaryRestrictionRole(moduleType))`, leaving `getRestrictedRoleByModule` as a pure response-DTO builder.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
```diff
@@ -107,77 +106,108 @@ public EmployeeRole assignRolesToEmployee(EmployeeSystemPermissionsDto roleReque
...
+		if (requestDto.getAdd() == null || requestDto.getRemove() == null) {
+			return;
+		}
+
+		Set<RoleLevel> overlap = EnumSet.noneOf(RoleLevel.class);
+		overlap.addAll(requestDto.getAdd());
+		overlap.retainAll(requestDto.getRemove());
+		if (!overlap.isEmpty()) {
+			throw new ModuleException(PeopleMessageConstant.PEOPLE_ERROR_RESTRICTION_ADD_REMOVE_OVERLAP,
+					new String[] { overlap.toString() });
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Two rough edges in the new error messages, both user-visible since `ModuleException` messages are returned in the `ErrorResponse` body. Here `overlap.toString()` on an `EnumSet` renders as `[ADMIN]`, so the client sees "[ADMIN] cannot be added and removed in the same request" with literal brackets — prefer `overlap.stream().map(RoleLevel::name).collect(joining(", "))`. Separately, at line 167-169 a null list entry is reported through `PEOPLE_ERROR_INVALID_RESTRICTION_ROLE_LEVEL` with `String.valueOf(roleLevel)`, producing "null is not a restrictable role for the PEOPLE module", which reads as a bug report rather than a validation message; a dedicated constant (or an `@NotNull` element constraint on the list, e.g. `List<@NotNull RoleLevel>`) would be clearer.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
```diff
@@ -107,77 +106,108 @@ public EmployeeRole assignRolesToEmployee(EmployeeSystemPermissionsDto roleReque
...
+		}
+
+		Set<RoleLevel> restrictableRoles = EnumSet.of(RoleLevel.ADMIN, PeopleUtil.getSecondaryRestrictionRole(module));
+		restrictableRoles.retainAll(moduleRoles);
+
+		return List.copyOf(restrictableRoles);
 	}
 
 	@Override
 	public ModuleRoleRestrictionResponseDto getRestrictedRoleByModule(ModuleType module) {
```
ThinuwanW, line 185: 🤖 **Claude** · 🟣 **Nit**

Pre-existing but on the path this PR reworks and worth fixing while you are here: `RolesController.getRestrictedRolesByModule` (RolesController.java:60-62) calls `ModuleType.valueOf(module)` on the raw path variable with no try/catch, so `GET /v1/roles/restrictions/NOT_A_MODULE` throws `IllegalArgumentException` and returns HTTP 500 via the generic `Exception` handler instead of a 400. The `@Valid` on the `@PathVariable String` there does nothing since there is no constraint annotation. Either declare the path variable as `@PathVariable ModuleType module` (Spring's converter produces a 400 via `MethodArgumentTypeMismatchException`) or wrap the `valueOf` and throw `ModuleException(PEOPLE_ERROR_INVALID_RESTRICTION_MODULE)` — the constant this PR just added is a natural fit.

backend/src/main/java/com/skapp/community/peopleplanner/constant/PeopleMessageConstant.java
```diff
@@ -145,6 +145,10 @@ public enum PeopleMessageConstant implements MessageConstant {
 	PEOPLE_ERROR_LEAVE_RESTRICTED_ROLE_ACCESS("api.error.people.leave-restricted-role-access"),
 	PEOPLE_ERROR_INVOICE_RESTRICTED_ROLE_ACCESS("api.error.people.invoice-restricted-role-access"),
```
ThinuwanW, line 147: 🤖 **Claude** · 🔴 **Important**

`PEOPLE_ERROR_INVOICE_RESTRICTED_ROLE_ACCESS("api.error.people.invoice-restricted-role-access")` has NO matching entry in `community/messages/people-messages.properties`. Its four siblings (attendance/people/leave/crm) are all present at properties lines 107-110, and this PR added four more keys immediately below them at 111-114 without noticing the hole. Consequence: `ValidationException`/`ModuleException` resolve their message eagerly in the constructor via `MessageUtil.getMessage(...)` -> `MessageSource.getMessage(code, args, locale)` with no default, and `spring.messages.use-code-as-default-message` is not set in application-non-prd.yml / application-prd.yml / config/application-ep-*.yml. So the throw at RolesServiceImpl:524 (`PEOPLE_ERROR_INVOICE_RESTRICTED_ROLE_ACCESS`) raises `NoSuchMessageException` instead of constructing the exception, which falls through to the catch-all `@ExceptionHandler(Exception.class)` in GlobalExceptionHandler and returns HTTP 500 rather than the intended 400. This is reachable today: restrict the INVOICE admin role, then have a People Admin (non-super-admin) try to assign `INVOICE_ADMIN`/`INVOICE_MANAGER`. Fix: add `api.error.people.invoice-restricted-role-access=People admins are restricted from assigning invoice {0} role` alongside the other four, and consider a test or a startup assertion that every `PeopleMessageConstant` key resolves, since eager resolution makes a missing key a 500 rather than a degraded message.

backend/src/main/java/com/skapp/community/peopleplanner/payload/request/ModuleRoleRestrictionRequestDto.java
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Invalid enum tokens in the request body produce HTTP 500, not 400. `module`, `add` and `remove` are enum-typed, so `{"module":"PEOPLE","add":["SUPER_ADMIN"]}` or `{"module":"FOO"}` fails during Jackson deserialization -> `HttpMessageNotReadableException`. I grepped the whole backend: there is no `@ExceptionHandler(HttpMessageNotReadableException.class)` and `GlobalExceptionHandler` does not extend `ResponseEntityExceptionHandler`, so it lands on `@ExceptionHandler(Exception.class)` -> 500 with a raw Jackson message in the body. This is distinct from the missing bean validation: bean validation runs after deserialization and can never see these values. The codebase already has the idiom for this — `com.skapp.community.peopleplanner.util.deserializer.AccountStatusDeserializer` (and Ethnicity/BloodGroup/Eeo/EmployeeType/MaritalStatus/EmploymentAllocation) wrap `IllegalArgumentException` from `valueOf` into a `ModuleException` with a message constant so the client gets a 400. Fix: add a `RoleLevel` (and `ModuleType`) `StdDeserializer` following that pattern and annotate these fields with `@JsonDeserialize(using = ...)` / `@JsonDeserialize(contentUsing = ...)`, or add a global `HttpMessageNotReadableException` handler that returns 400. Worth adding an integration test for a bogus role token too — the new test class covers null and duplicate entries but not an unparseable one.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
```diff
@@ -235,12 +265,11 @@ private AllowedModuleRolesResponseDto processModuleRoles(Map.Entry<ModuleType, L
 		ModuleType module = entry.getKey();
 		List<RoleLevel> prebuiltRoles = entry.getValue();
 
-		ModuleRoleRestriction moduleRoleRestriction = moduleRoleRestrictionDao.findById(module).orElse(null);
+		Set<RoleLevel> restrictedRoleLevels = getRestrictedRoleLevels(module);
 
-		boolean isAdminAllowed = isSuperAdmin
-				|| (moduleRoleRestriction == null || Boolean.FALSE.equals(moduleRoleRestriction.getIsAdmin()));
+		boolean isAdminAllowed = isSuperAdmin || !restrictedRoleLevels.contains(RoleLevel.ADMIN);
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The storage model was generalised to a set of role levels, but all three enforcement/read paths immediately collapse it back to two booleans keyed on ADMIN + `getSecondaryRestrictionRole(module)`: here (lines 270-272), in `getRestrictedRoleByModule` (lines 206-208), and in `validateRestrictedRoleAssignment` (lines 561-575, which reads only `getIsAdmin()`/`getIsManager()` off the response DTO). Two consequences. (1) Any value in the `restrictions` column outside {ADMIN, secondary} — from the legacy sync, a manual DB edit, or a future role level — is faithfully echoed by `GET /v1/roles/restrictions/{module}` and is even removable via `remove` (the new test `update_RemoveNonRestrictableStoredRole_ClearsIt` proves the write path accepts it), yet it is silently never enforced anywhere. The API reports a restriction that does nothing. (2) The same two-boolean derivation is duplicated in the two places noted above. Fix: have `isRoleAllowed`/`processModuleRoles` test `restrictedRoleLevels.contains(roleLevel)` directly instead of routing through `isAdminAllowed`/`isManagerAllowed`, and have `validateRestrictedRoleAssignment` map the incoming `Role` to its `RoleLevel` and check set membership; keep `isAdmin`/`isManager` purely as a legacy response projection computed in one shared helper.

backend/src/main/java/com/skapp/community/peopleplanner/payload/response/ModuleRoleRestrictionResponseDto.java
```diff
@@ -14,4 +17,8 @@ public class ModuleRoleRestrictionResponseDto {
 
 	private Boolean isManager;
 
+	private List<RoleLevel> restrictions;
+
+	private List<RoleLevel> restrictableRoles;
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The new `restrictableRoles` field has no consumer and contradicts the shipped UI for two modules. `frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx:94-110` hardcodes its own `getRestrictableRoles(module)` switch and never reads this field. Comparing the two: INVOICE -> frontend `["isAdmin"]` vs backend `[ADMIN, MANAGER]` (because `initializeRolesForModule()` maps INVOICE to `[ADMIN, MANAGER]`), and OKR -> frontend falls to `default: return []` (no checkboxes at all) vs backend `[ADMIN, MANAGER]`. Both modules are rendered by the config UI (`utils/userRoles/apiUtils.ts:45-46`, `ModuleRolesTable.tsx:122`), so the backend is now advertising restrictable roles the UI will not show. Decide which is authoritative: either delete the frontend switch and drive the checkboxes off `restrictableRoles` (which is presumably the intent of adding it), or correct the backend list. Either way, shipping a response field with zero consumers and a known disagreement invites drift.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
```diff
@@ -107,77 +106,108 @@ public EmployeeRole assignRolesToEmployee(EmployeeSystemPermissionsDto roleReque
...
-		if (restrictedRole.isEmpty()) {
-			ModuleRoleRestrictionResponseDto newRestrictRole = new ModuleRoleRestrictionResponseDto();
-			newRestrictRole.setModule(module);
-			newRestrictRole.setIsAdmin(false);
-			newRestrictRole.setIsManager(false);
+		log.info("getRestrictedRoleByModule: execution started");
 
-			return newRestrictRole;
-		}
+		Set<RoleLevel> restrictedRoleLevels = getRestrictedRoleLevels(module);
```
ThinuwanW, line 188: 🤖 **Claude** · 🟡 **Suggestion**

Read/write validation asymmetry on the same resource. `updateRoleRestrictions` now rejects a null or COMMON module with a 400 via `validateRoleRestrictionRequest`, but `getRestrictedRoleByModule` applies no module validation at all: `GET /v1/roles/restrictions/COMMON` returns HTTP 200 with `restrictions: []`, `restrictableRoles: []`, `isAdmin: false`, `isManager: false` — a meaningless success payload for a module that can never hold restrictions. The paired endpoints for one resource should agree on what a valid module is. Extract the module check into a small helper (e.g. `validateRestrictableModule(module)`) and call it from both, so the read path returns the same `PEOPLE_ERROR_INVALID_RESTRICTION_MODULE` 400. This also gives the GET endpoint a defined answer for modules absent from `initializeRolesForModule()` instead of a silently-empty `restrictableRoles`.

backend/src/main/java/com/skapp/community/peopleplanner/util/PeopleUtil.java
```diff
@@ -91,6 +98,39 @@ public static List<EmployeeManager> filterManagersByAttendanceRoles(List<Employe
 		return filterManagersByRoles(managers, List.of(Role.ATTENDANCE_ADMIN, Role.ATTENDANCE_MANAGER));
 	}
 
+	public static RoleLevel getSecondaryRestrictionRole(ModuleType module) {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Three new public static methods with real branching/parsing logic were added to `PeopleUtil` and none of them has a unit test — there is no `PeopleUtilTest` anywhere in `backend/src/test`. `parseRestrictions` in particular is the piece with edge cases (null, blank, untrimmed tokens, empty segments between separators, unknown tokens, case sensitivity) and it is currently only exercised indirectly through MockMvc integration tests, which means the failure modes are asserted at HTTP-status granularity rather than at the function boundary. `getSecondaryRestrictionRole` and `toRestrictionsString` are pure functions and trivial to pin down (including the deliberate `null` return for an empty set, which callers rely on to write SQL NULL). Add a plain JUnit test class for these three — it is much cheaper than the integration tests and covers cases the integration tests structurally cannot reach.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
```diff
@@ -107,77 +106,108 @@ public EmployeeRole assignRolesToEmployee(EmployeeSystemPermissionsDto roleReque
...
+		validateRoleRestrictionRequest(moduleRoleRestrictionRequestDto);
 
-		ModuleRoleRestriction moduleRoleRestriction = peopleMapper
-			.roleRestrictionRequestDtoToRestrictRole(moduleRoleRestrictionRequestDto);
-		moduleRoleRestrictionDao.save(moduleRoleRestriction);
+		ModuleType module = moduleRoleRestrictionRequestDto.getModule();
+		Set<RoleLevel> restrictedRoles = resolveRestrictedRoles(moduleRoleRestrictionRequestDto);
 
-		ModuleRolesRestriction moduleRolesRestriction = buildModuleRolesRestriction(moduleRoleRestrictionRequestDto);
+		ModuleRolesRestriction moduleRolesRestriction = new ModuleRolesRestriction();
```
ThinuwanW, line 109: 🤖 **Claude** · 🟡 **Suggestion**

`ModuleRolesRestriction` is now the single authoritative store for a security control (which roles a People Admin may not assign), and it does not extend `Auditable<String>` — so there is no record of who changed a restriction or when. Every other entity in the codebase extends `Auditable<String>` per the project convention, and the write here is a blind upsert of the whole row, so the previous value is not recoverable either. The legacy `module_role_restriction` table had the same gap, but this PR is what promotes this table to authoritative and removes the second copy of the data, so it is the right moment to fix it: make `ModuleRolesRestriction` extend `Auditable<String>` and add a Liquibase changeset for the audit columns (following the existing `community/db/changelog/` pattern, alongside changeset `00081_create_table_module_roles_restriction`).

backend/src/main/java/com/skapp/community/peopleplanner/util/PeopleUtil.java
```diff
@@ -91,6 +98,39 @@ public static List<EmployeeManager> filterManagersByAttendanceRoles(List<Employe
...
+		};
+	}
+
+	public static Set<RoleLevel> parseRestrictions(String restrictions) {
+		Set<RoleLevel> roleLevels = EnumSet.noneOf(RoleLevel.class);
+		if (restrictions == null || restrictions.isBlank()) {
+			return roleLevels;
+		}
+
+		for (String restriction : restrictions.split(RESTRICTION_SEPARATOR)) {
```
ThinuwanW, line 106: 🤖 **Claude** · 🟣 **Nit**

`RESTRICTION_SEPARATOR` is used in two incompatible ways: as a regex in `restrictions.split(RESTRICTION_SEPARATOR)` here, and as a literal in `Collectors.joining(RESTRICTION_SEPARATOR)` at line 131. It happens to work because `,` is not a regex metacharacter, but the shared constant makes the two look interchangeable — changing it to `|`, `.` or `+` would silently break parsing while serialisation kept working. Either hold a `private static final Pattern RESTRICTION_PATTERN = Pattern.compile(Pattern.quote(RESTRICTION_SEPARATOR))` and call `RESTRICTION_PATTERN.split(restrictions)`, or add a comment stating the constant must stay regex-safe.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
```diff
@@ -107,77 +106,108 @@ public EmployeeRole assignRolesToEmployee(EmployeeSystemPermissionsDto roleReque
...
+
+	private void validateRestrictionRoleLevels(List<RoleLevel> roleLevels, ModuleType module,
+			List<RoleLevel> restrictableRoles) {
+		if (roleLevels == null) {
+			return;
+		}
+
+		Set<RoleLevel> seen = EnumSet.noneOf(RoleLevel.class);
+		for (RoleLevel roleLevel : roleLevels) {
+			if (roleLevel == null || (restrictableRoles != null && !restrictableRoles.contains(roleLevel))) {
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Two rough edges in `validateRestrictionRoleLevels`. (1) The `restrictableRoles` parameter doubles as a control flag — callers pass `null` to mean "skip the allowlist check" (line 144 for `remove`). That is not obvious at the call site and a future caller passing an accidentally-null list would silently disable validation. Prefer an explicit `boolean enforceAllowlist` parameter, or split into `validateAddedRoleLevels` / `validateRemovedRoleLevels`. (2) A `null` entry reuses the `PEOPLE_ERROR_INVALID_RESTRICTION_ROLE_LEVEL` message, so `{"module":"PEOPLE","remove":[null]}` returns the user-visible text "null is not a restrictable role for the PEOPLE module" (via `String.valueOf(roleLevel)`). Separate the null check into its own branch with a dedicated message key such as `api.error.people.restriction-role-level-required`.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
```diff
@@ -107,77 +106,108 @@ public EmployeeRole assignRolesToEmployee(EmployeeSystemPermissionsDto roleReque
...
 
 		log.info("updateRoleRestrictions: execution ended");
 		return new ResponseEntityDto(false, messageUtil.getMessage(PeopleMessageConstant.PEOPLE_SUCCESS_ROLE_RESTRICT));
 	}
 
-	private ModuleRolesRestriction buildModuleRolesRestriction(
-			ModuleRoleRestrictionRequestDto moduleRoleRestrictionRequestDto) {
-		ModuleType module = moduleRoleRestrictionRequestDto.getModule();
+	private Set<RoleLevel> resolveRestrictedRoles(ModuleRoleRestrictionRequestDto requestDto) {
+		Set<RoleLevel> restrictedRoles = EnumSet.noneOf(RoleLevel.class);
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`resolveRestrictedRoles` allocates an empty `EnumSet.noneOf(RoleLevel.class)` and then `addAll`s the result of `getRestrictedRoleLevels(...)`, but `getRestrictedRoleLevels` already returns a freshly-built, mutable `EnumSet` in both branches (`PeopleUtil.parseRestrictions` creates one, and the `orElseGet` creates one). The defensive copy is redundant. Simplify to `Set<RoleLevel> restrictedRoles = getRestrictedRoleLevels(requestDto.getModule());` — or, if the copy is intentional insurance against a subclass returning a shared/immutable set, use `EnumSet.copyOf(...)` guarded for the empty case and say so in a comment, since as written the intent is ambiguous.

backend/src/test/java/com/skapp/community/peopleplanner/controller/v1/RolesControllerIntegrationTest.java
```diff
@@ -0,0 +1,399 @@
...
+				.andExpect(jsonPath(RESULTS_0_PATH + IS_ADMIN_FIELD).value(false))
+				.andExpect(jsonPath(RESULTS_0_PATH + IS_MANAGER_FIELD).value(false));
+		}
+
+		@Test
+		@DisplayName("Get restrictions for manager module - Returns admin and manager as restrictable")
+		void getRestrictions_ManagerModule_ReturnsAdminAndManagerAsRestrictable() throws Exception {
+			getRestrictionsByModule(ModuleType.PEOPLE).andExpect(status().isOk())
+				.andExpect(jsonPath(RESULTS_0_PATH + RESTRICTABLE_ROLES_FIELD, hasSize(2)))
+				.andExpect(jsonPath(RESULTS_0_PATH + RESTRICTABLE_ROLES_FIELD + "[0]").value(RoleLevel.ADMIN.name()))
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

These assertions pin the API payload to the declaration order of the `RoleLevel` enum. `restrictableRoles` is built as an `EnumSet` and copied with `List.copyOf` (RolesServiceImpl:190-193), so iteration order is enum-ordinal order, not a documented API contract; the same applies to `restrictions` (RolesServiceImpl:204). Asserting `restrictableRoles[0] == ADMIN` and `[1] == MANAGER`/`SALES_MANAGER` means that reordering constants in `RoleLevel` — a change that looks harmless because everything is persisted as `EnumType.STRING` — silently breaks these tests and reorders the JSON for clients. The sibling tests in this class already use `containsInAnyOrder` for `restrictions`; use it here too (`containsInAnyOrder(ADMIN.name(), MANAGER.name())` with `hasSize(2)`), or make the ordering explicit in the service and document it.

backend/src/main/java/com/skapp/community/peopleplanner/payload/request/ModuleRoleRestrictionRequestDto.java
anusham-rc: separate frontend pr has raised to handle this change

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
anusham-rc: Migration API is already available to handle this issue

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
anusham-rc: invalid with the current code practices. we do not use row locks

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
anusham-rc: removed those unwanted deal code from both community and ep

backend/src/main/java/com/skapp/community/peopleplanner/payload/request/ModuleRoleRestrictionRequestDto.java
anusham-rc: invalid, does not match with the current code practice

backend/src/main/java/com/skapp/community/peopleplanner/payload/request/ModuleRoleRestrictionRequestDto.java
anusham-rc: invalid, does not match with the current code practice

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
anusham-rc: No change needed.
ESIGN restriction is enterprise-only: the community role map has no ESIGN, and only the enterprise override adds it. The community restriction modal isn't wired into any screen, so the ESIGN path isn't reachable there.

backend/src/test/java/com/skapp/community/peopleplanner/controller/v1/RolesControllerIntegrationTest.java
anusham-rc: updated the tests

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
anusham-rc: Done, updated validateRestrictedRoleAssignment to use getRestrictedRoleLevels directly

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
anusham-rc: invalid, arrays always guard from frontend

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
anusham-rc: invalid, not related to this PR

backend/src/main/java/com/skapp/community/peopleplanner/constant/PeopleMessageConstant.java
anusham-rc: The key isn't missing, api.error.people.invoice-restricted-role-access exists at people-messages.properties:110, right alongside the other four. So this resolves fine and returns the intended 400, not a 500.

backend/src/main/java/com/skapp/community/peopleplanner/payload/request/ModuleRoleRestrictionRequestDto.java
anusham-rc: This endpoint is super-admin-only and the client always sends valid enum values from a fixed set, so a bad token isn't reachable in normal use, the 500 only occurs on a hand-crafted payload. Treating this as low-priority hardening rather than fixing it in this PR; can follow up with the global deserializer approach if we want the clean 400 later.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
anusham-rc: this is a generalization to support future role levels (e.g. sales rep) and de-duplicate the two-boolean logic, not a bug in the current flow. Since restrictions can't be set to non-restrictable values through the API now, I'll handle this in a separate PR rather than expanding the scope here.

backend/src/main/java/com/skapp/community/peopleplanner/payload/response/ModuleRoleRestrictionResponseDto.java
anusham-rc: invalid, separate frontend PR raised to handle this

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
anusham-rc: will raise a separate pr to fix this module restrictions

backend/src/main/java/com/skapp/community/peopleplanner/util/PeopleUtil.java
anusham-rc: Invalid, backend has only integration test as a practice

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
anusham-rc: will handle this in separate pr if needed the auditing

backend/src/main/java/com/skapp/community/peopleplanner/util/PeopleUtil.java
anusham-rc: invalid, the design is to use , as the separator.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
anusham-rc: Addressed.

1. Gave the `null` entry its own branch with a dedicated `api.error.people.restriction-role-level-required` message ("A role level must be provided"), so `{"module":"PEOPLE","remove":[null]}` no longer returns "null is not a restrictable role for the PEOPLE module".
2. Replaced the `null`-as-skip-flag with an explicit `boolean enforceAllowlist` parameter — `add` passes `true`, `remove` passes `false` — so the allowlist skip is obvious at the call site instead of being inferred from a null list.

Behaviour is otherwise unchanged; the existing integration tests (null entry → 400, non-restrictable add → 400, remove non-restrictable stored role → clears it) still pass.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
anusham-rc: Done. Removed the redundant copy and assign the result directly — `getRestrictedRoleLevels` already returns a fresh mutable set. Tests pass.

backend/src/test/java/com/skapp/community/peopleplanner/controller/v1/RolesControllerIntegrationTest.java
anusham-rc: Done. Switched all three `restrictableRoles` assertions to `containsInAnyOrder` (with `hasSize`), matching the sibling `restrictions` tests, so reordering the `RoleLevel` enum no longer breaks them. Tests pass.

backend/src/main/java/com/skapp/community/peopleplanner/util/PeopleUtil.java
anusham-rc: Leaving `parseRestrictions` as-is. The `restrictions` column is only ever written by two controlled paths — the validated PATCH API (which writes `RoleLevel` names via `toRestrictionsString`) and the linked sync migration — and both only produce valid enum tokens, so an unparseable value cannot reach the read path. There are no manual/out-of-band or cross-edition writers on this table. A future `RoleLevel` rename/removal would be handled via migration alongside the code change. We can add the defensive skip later if an unvalidated write path is ever introduced.

shakila-kamalasena (dismissed):

backend/src/main/java/com/skapp/community/peopleplanner/payload/request/ModuleRoleRestrictionRequestDto.java
akilarootcode: addedRoles removedRoles

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
```diff
@@ -107,48 +100,75 @@ public EmployeeRole assignRolesToEmployee(EmployeeSystemPermissionsDto roleReque
...
+		if (roleLevels == null) {
+			return;
+		}
+
+		Set<RoleLevel> seen = EnumSet.noneOf(RoleLevel.class);
+		for (RoleLevel roleLevel : roleLevels) {
+			if (roleLevel == null) {
+				throw new ModuleException(PeopleMessageConstant.PEOPLE_ERROR_RESTRICTION_ROLE_LEVEL_REQUIRED);
+			}
+			if (enforceAllowlist && !restrictableRoles.contains(roleLevel)) {
```
akilarootcode: remove this

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
```diff
@@ -107,48 +100,75 @@ public EmployeeRole assignRolesToEmployee(EmployeeSystemPermissionsDto roleReque
...
+		return restrictedRoles;
+	}
+
+	private void validateRoleRestrictionRequest(ModuleRoleRestrictionRequestDto requestDto) {
+		ModuleType module = requestDto.getModule();
+		if (module == null || module == ModuleType.COMMON) {
+			throw new ModuleException(PeopleMessageConstant.PEOPLE_ERROR_INVALID_RESTRICTION_MODULE);
+		}
+
+		validateRestrictionRoleLevels(requestDto.getAdd(), module, getRestrictableRoles(module), true);
```
akilarootcode: remove this and hardcode this

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
```diff
@@ -107,48 +100,75 @@ public EmployeeRole assignRolesToEmployee(EmployeeSystemPermissionsDto roleReque
...
+		validateRestrictionRoleLevels(requestDto.getRemove(), module, null, false);
+
+		if (requestDto.getAdd() == null || requestDto.getRemove() == null) {
```
akilarootcode: move this logic to top

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
```diff
@@ -107,48 +100,75 @@ public EmployeeRole assignRolesToEmployee(EmployeeSystemPermissionsDto roleReque
...
+			return;
+		}
+
+		Set<RoleLevel> overlap = EnumSet.noneOf(RoleLevel.class);
```
akilarootcode: check to use dto level Set instead remove duplicates inside service level

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
```diff
@@ -107,48 +100,75 @@ public EmployeeRole assignRolesToEmployee(EmployeeSystemPermissionsDto roleReque
...
+
+	private void validateRestrictionRoleLevels(List<RoleLevel> roleLevels, ModuleType module,
+			List<RoleLevel> restrictableRoles, boolean enforceAllowlist) {
+		if (roleLevels == null) {
+			return;
+		}
+
+		Set<RoleLevel> seen = EnumSet.noneOf(RoleLevel.class);
+		for (RoleLevel roleLevel : roleLevels) {
+			if (roleLevel == null) {
```
akilarootcode: check this needed or not

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
```diff
@@ -107,48 +100,75 @@ public EmployeeRole assignRolesToEmployee(EmployeeSystemPermissionsDto roleReque
...
+				throw new ModuleException(PeopleMessageConstant.PEOPLE_ERROR_RESTRICTION_ROLE_LEVEL_REQUIRED);
+			}
+			if (enforceAllowlist && !restrictableRoles.contains(roleLevel)) {
+				throw new ModuleException(PeopleMessageConstant.PEOPLE_ERROR_INVALID_RESTRICTION_ROLE_LEVEL,
+						new String[] { roleLevel.name(), module.name() });
 			}
-			if (Boolean.TRUE.equals(moduleRoleRestrictionRequestDto.getIsManager())) {
-				restrictedRoles.add(PeopleUtil.getSecondaryRestrictionRole(module).name());
+			if (!seen.add(roleLevel)) {
```
akilarootcode: if added and removed has same values throw error

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
```diff
@@ -535,16 +555,16 @@ protected boolean hasOnlyPeopleAdminPermissions(User currentUser) {
...
 		if (role == Role.PEOPLE_ADMIN || role == Role.ATTENDANCE_ADMIN || role == Role.LEAVE_ADMIN
 				|| role == Role.INVOICE_ADMIN || role == Role.CRM_ADMIN) {
-			return Boolean.TRUE.equals(restrictedRole.getIsAdmin());
+			return restrictedRoleLevels.contains(RoleLevel.ADMIN);
 		}
 
 		if (role == Role.PEOPLE_MANAGER || role == Role.ATTENDANCE_MANAGER || role == Role.LEAVE_MANAGER
 				|| role == Role.INVOICE_MANAGER || role == Role.CRM_SALES_MANAGER) {
-			return Boolean.TRUE.equals(restrictedRole.getIsManager());
+			return restrictedRoleLevels.contains(PeopleUtil.getSecondaryRestrictionRole(moduleType));
```
akilarootcode: check getSecondaryRestrictionRole

akilarootcode (changes requested):

backend/src/main/java/com/skapp/community/peopleplanner/payload/request/ModuleRoleRestrictionRequestDto.java
anusham-rc: renamed both and used Set instead List

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
anusham-rc: enforceAllowlist  removed and now check the validation on both add and remove senarions

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
anusham-rc: removed and getRestrictableRoles now handle inside the validateRestrictionRoleLevels

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
anusham-rc: now dto use Set instead of List

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
anusham-rc: refactored this, now no **getSecondaryRestrictionRole** check, just flat roleLevel to check as RestrictionableRoles

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
anusham-rc: remove and simplified the validations

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
anusham-rc: removed, now handles without for loop

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/RolesServiceImpl.java
anusham-rc: refactored this check

akilarootcode (approved):

PR #1696, opened 2026-07-31
feat/12: add team to the project

yuhanga-rtc:
### Summary

- https://github.com/rootcodelabs/skapp-ui/pull/360/changes
- https://github.com/SkappHQ/skapp/pull/1696/changes
- https://github.com/rootcodelabs/skapp-pm/pull/1201/changes
- https://github.com/rootcodelabs/skapp-ep-be-src/pull/733/changes

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/backend/src/main/java/com/skapp/community/peopleplanner/service/impl/TeamServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/peopleplanner/service/impl/TeamServiceImpl.java
@@ -11,11 +11,11 @@ import com.skapp.community.peopleplanner.mapper.PeopleMapper;
 import com.skapp.community.peopleplanner.model.Employee;
 import com.skapp.community.peopleplanner.model.EmployeeTeam;
 import com.skapp.community.peopleplanner.model.Team;
+import com.skapp.community.peopleplanner.payload.EmployeeTeamIdDto;
 import com.skapp.community.peopleplanner.payload.request.TeamPatchRequestDto;
 import com.skapp.community.peopleplanner.payload.request.TeamRequestDto;
 import com.skapp.community.peopleplanner.payload.request.TeamsRequestDto;
 import com.skapp.community.peopleplanner.payload.request.TransferTeamMembersDto;
-import com.skapp.community.peopleplanner.payload.EmployeeTeamIdDto;
 import com.skapp.community.peopleplanner.payload.response.EmployeeTransferableTeamsResponseDto;
 import com.skapp.community.peopleplanner.payload.response.TeamBasicDetailsResponseDto;
 import com.skapp.community.peopleplanner.payload.response.TeamResponseDto;
@@ -99,6 +99,7 @@ public class TeamServiceImpl implements TeamService {
 		setEmployeeTeam(employeeTeamSet, teamMembers, team, false);
 		team.setEmployees(employeeTeamSet);
 		teamDao.save(team);
+		invalidateTeamsCache();
 
 		log.info("addNewTeam: execution successfully finished by saving team : {}", team.getTeamId());
 		return new ResponseEntityDto(false, peopleMapper.teamToTeamResponseDto(team));
@@ -177,6 +178,7 @@ public class TeamServiceImpl implements TeamService {
 			team.setEmployees(employeeTeamSet);
 			employeeTeamDao.deleteAllInBatch(employeeTeamsTobeRemoved);
 			teamDao.save(team);
+			invalidateTeamsCache();
 
 			log.info("updateTeam: execution successfully finished by updating team : {}", team.getTeamId());
 			return new ResponseEntityDto(false, peopleMapper.teamToTeamResponseDto(team));
@@ -221,8 +223,10 @@ public class TeamServiceImpl implements TeamService {
 			})
 			.toList();
 
-		if (!teams.isEmpty())
+		if (!teams.isEmpty()) {
 			teamDao.saveAll(teams);
+			invalidateTeamsCache();
+		}
 
 		log.info("addTeams: execution successfully finished by saving teams by adding {} team(s)", teams.size());
 		return new ResponseEntityDto(false, peopleMapper.teamListToTeamResponseDtoList(teams));
@@ -276,6 +280,7 @@ public class TeamServiceImpl implements TeamService {
 
 			employeeTeamDao.deleteEmployeeTeamByTeamId(id);
 			teamDao.deleteTeamById(id);
+			invalidateTeamsCache();
 		}
 		else {
 			throw new EntityNotFoundException(PeopleMessageConstant.PEOPLE_ERROR_TEAM_NOT_FOUND);
@@ -317,6 +322,10 @@ public class TeamServiceImpl implements TeamService {
 
 		log.info("getTeamsForCurrentUser: Successfully finished");
 		return new ResponseEntityDto(false, response);
+	}
+
+	protected void invalidateTeamsCache() {
+		// This method is a placeholder for enterprise cache invalidation logic
 	}
 
 	private void setUpNewTeam(List<EmployeeTeam> employeeTeamList, TransferTeamMembersDto transfer) {


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds a `protected void invalidateTeamsCache()` template hook to `TeamServiceImpl` and calls it from the four team write paths (`addNewTeam`, `updateTeam`, `addTeams`, `transferMembersAndDeleteTeam`), mirroring the existing `invalidateJobsCache`/`invalidateUserCache` pattern; the enterprise `EpTeamServiceImpl` overrides it to evict the `teams_cache:all_teams` Redis key (7-day TTL). Coverage of the write paths inside `TeamServiceImpl` itself is complete and correctly guarded, but two correctness gaps remain: team membership is also mutated from `PeopleServiceImpl` (employee create/update/bulk-import/deactivate, supervisor transfer) without invalidating the cache even though the cached DTO carries `employeeIds`, and in the two `@Transactional` methods the eviction runs before commit, which allows a concurrent read to re-cache the pre-commit snapshot. Both are fixable with small, localized changes.

Found **8** new issue(s): 🔴 3 important, 🟡 3 suggestion(s), 🟣 2 nit(s)

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/TeamServiceImpl.java
```diff
@@ -319,6 +324,10 @@ public ResponseEntityDto getTeamsForCurrentUser() {
 		return new ResponseEntityDto(false, response);
 	}
 
+	protected void invalidateTeamsCache() {
```
ThinuwanW, line 327: 🤖 **Claude** · 🔴 **Important**

Invalidation is incomplete: team membership is also mutated outside `TeamServiceImpl`, and those paths do not evict the cache. The cached payload (`EpTeamResponseDto`, produced by `EpTeamServiceImpl.getAllTeams()` and stored under `TENANT_ALL_TEAMS_CACHE_KEY` with a 7-day TTL) includes `employeeIds` per team, so any employee↔team change must invalidate it. Uncovered writers in `PeopleServiceImpl`: `processEmployeeTeams(...)` (reached from `createEmployee` and `updateEmployee` via `createEmployeeEntity`, ~line 469/738 — clears and rebuilds `employee.getEmployeeTeams()`), `saveEmployeeTeams(...)` (~line 2201, bulk employee import), `processTeamSupervisorTransfer(...)` (~line 1532-1540, can insert a brand-new `EmployeeTeam` row), and the deactivate/delete path (~line 2694-2697, clears all `employeeTeams`) which invalidates only `invalidateUserCache()`/`invalidateUserAuthPicCache()`. Result: the internal API consumer of `/internal/v1/ep/teams` can see a team whose member list is missing/containing employees for up to the TTL. Fix: add the same `protected void invalidateTeamsCache()` hook to `PeopleServiceImpl` (overridden in `EpPeopleServiceImpl`, alongside the existing `invalidateUserCache`) and call it from those four paths — or, cleaner, replace the subclass hook with an application event (see the separate comment) that both services publish.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/TeamServiceImpl.java
```diff
@@ -177,6 +178,7 @@ public ResponseEntityDto updateTeam(Long id, TeamPatchRequestDto teamPatchReques
 			team.setEmployees(employeeTeamSet);
 			employeeTeamDao.deleteAllInBatch(employeeTeamsTobeRemoved);
 			teamDao.save(team);
+			invalidateTeamsCache();
```
ThinuwanW, line 181: 🤖 **Claude** · 🔴 **Important**

The eviction runs *inside* the transaction, before commit. `updateTeam` (line 181) and `transferMembersAndDeleteTeam` (line 283) are `@Transactional`, so at the moment `invalidateTeamsCache()` deletes the Redis key the new rows are not yet committed. A concurrent `EpTeamServiceImpl.getAllTeams()` (which runs `@Transactional(readOnly = true)` in its own transaction and unconditionally `put`s the result back) can land in that window, read the *pre-commit* state, and re-populate the cache with stale data — which then survives for the full 7-day TTL because nothing evicts it again. The mirror problem: if the transaction subsequently rolls back, the cache was evicted for a change that never happened (benign, but the current code cannot distinguish the two). Fix: defer the eviction to after commit. The enterprise module already has exactly this helper — wrap the body of `EpTeamServiceImpl.invalidateTeamsCache()` in `TransactionUtil.runAfterCommit(() -> cacheService.invalidate(cacheKey.getKey()))` (it falls back to running immediately when no transaction is active, so the non-transactional `addNewTeam`/`addTeams` paths keep working unchanged). Note the same latent issue exists in `PeopleServiceImpl.updateEmployee` → `invalidateUserCache()`, so fixing it in the enterprise override fixes it consistently for all hooks.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/TeamServiceImpl.java
```diff
@@ -99,6 +99,7 @@ public ResponseEntityDto addNewTeam(TeamRequestDto teamRequestDto) {
 		setEmployeeTeam(employeeTeamSet, teamMembers, team, false);
 		team.setEmployees(employeeTeamSet);
 		teamDao.save(team);
+		invalidateTeamsCache();
```
ThinuwanW, line 102: 🤖 **Claude** · 🟡 **Suggestion**

`addNewTeam` (line 102) and `addTeams` (line 228) are not annotated `@Transactional`, unlike `updateTeam` and `transferMembersAndDeleteTeam`. Each performs reads-then-write and now also triggers a cache side effect, so the DB write and the eviction are two independent units of work with no atomicity guarantee: `addNewTeam` in particular relies on a single cascading `teamDao.save(team)` for both the `Team` and its `EmployeeTeam` children, and if the eviction throws (e.g. Redis unavailable) the request fails with a 500 after the team was already committed. Consider adding `@Transactional` to both methods for consistency with the other two write paths and with the guideline on writes requiring `@Transactional` — combined with the after-commit deferral suggested above, that makes commit-then-evict the uniform behaviour across all four call sites.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/TeamServiceImpl.java
```diff
@@ -319,6 +324,10 @@ public ResponseEntityDto getTeamsForCurrentUser() {
 		return new ResponseEntityDto(false, response);
 	}
 
+	protected void invalidateTeamsCache() {
+		// This method is a placeholder for enterprise cache invalidation logic
+	}
```
ThinuwanW, line 329: 🤖 **Claude** · 🟡 **Suggestion**

This is now the third no-op `protected void invalidateXCache()` template hook in the community layer with a byte-identical placeholder comment (`PeopleServiceImpl.invalidateUserCache`/`invalidateUserAuthPicCache` at lines 343-349, `JobServiceImpl.invalidateJobsCache` at line 330). The pattern forces enterprise to subclass the community `@Service` and re-declare its entire constructor (`EpTeamServiceImpl` already duplicates all six super constructor args, so any new `TeamServiceImpl` dependency breaks the enterprise build), and it offers no place to express commit ordering. Consider migrating to the event mechanism already present in this module — `applicationEventPublisher.publishEvent(new TeamsChangedEvent(this, teamId))` (compare `UserCreatedEvent`/`UserDeactivatedEvent` in `PeopleServiceImpl`) with an enterprise `@TransactionalEventListener(phase = AFTER_COMMIT)` handler. That removes the subclass coupling, lets `PeopleServiceImpl` publish the same event for the membership changes noted above, and gets after-commit semantics for free. At minimum, add a Javadoc on this method documenting that it is an extension point overridden by `EpTeamServiceImpl` and that overrides must be safe to call inside an active transaction.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/TeamServiceImpl.java
```diff
@@ -11,11 +11,11 @@
 import com.skapp.community.peopleplanner.model.Employee;
 import com.skapp.community.peopleplanner.model.EmployeeTeam;
 import com.skapp.community.peopleplanner.model.Team;
+import com.skapp.community.peopleplanner.payload.EmployeeTeamIdDto;
```
ThinuwanW, line 14: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

No test coverage for the new behaviour: there is no `TeamServiceImplUnitTest` anywhere under `backend/src/test` (only `JobServiceImplUnitTest`, which also does not assert its cache hook), and `EpTeamServiceImpl` has no test either. Since the community implementation is a no-op, the only way a regression here is caught is by asserting the call. Suggest a `TeamServiceImplUnitTest` using `Mockito.spy`/a test subclass that verifies `invalidateTeamsCache()` is invoked exactly once by `addNewTeam`, `updateTeam`, `addTeams` and `transferMembersAndDeleteTeam`, and — importantly for the guards added in this PR — that it is *not* invoked when `addTeams` filters every requested name out as an existing team (empty `teams` list), when `updateTeam`/`transferMembersAndDeleteTeam` throw `EntityNotFoundException` for an unknown id, and when `addNewTeam` throws on a duplicate team name or invalid supervisor/member ids.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/TeamServiceImpl.java
```diff
@@ -276,6 +280,7 @@ public ResponseEntityDto transferMembersAndDeleteTeam(Long id,
 
 			employeeTeamDao.deleteEmployeeTeamByTeamId(id);
 			teamDao.deleteTeamById(id);
+			invalidateTeamsCache();
```
ThinuwanW, line 283: 🤖 **Claude** · 🔴 **Important**

Cache invalidation is now a hard dependency of team writes. `invalidateTeamsCache()` is called inside the `@Transactional` bodies of `transferMembersAndDeleteTeam` (here) and `updateTeam`, and the enterprise override (`EpTeamServiceImpl.invalidateTeamsCache` -> `EpCacheServiceImpl.invalidate`) only catches `RedisConnectionFailureException`. Other Spring-translated Redis failures are not caught: command timeouts surface as `QueryTimeoutException`, and MISCONF/OOM/READONLY-after-failover surface as `RedisSystemException`. Both escape the hook, propagate out of the service method, and roll back the transaction - so a Redis blip makes an admin's team update or team deletion fail with a 500 even though the DB work had succeeded. In the non-transactional `addNewTeam`/`addTeams` the rows are already committed, so the same failure returns a 500 for an operation that actually persisted, leaving the client's view inconsistent. Fix: make invalidation non-fatal. Either wrap each call site (or the hook body) in `try { invalidateTeamsCache(); } catch (Exception e) { log.error("failed to invalidate teams cache", e); }`, or broaden the enterprise catch to `org.springframework.dao.DataAccessException`. Document the hook contract as "must never throw" in the placeholder comment so future overrides do not reintroduce this.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/TeamServiceImpl.java
ThinuwanW, line 327: 🤖 **Claude** · 🟡 **Suggestion**

This template-hook mechanism only works if the enterprise subclass is actually registered as a bean, and that invariant is silently broken elsewhere in the codebase: `com.skapp.enterprise.people.service.impl.EpJobServiceImpl` (which overrides `JobServiceImpl.invalidateJobsCache()`) carries no `@Service`/`@Component`/`@Primary` and is not referenced by any `@Bean` method or config, so it is never instantiated and the jobs cache is never invalidated - a whole feature that silently does nothing. `EpTeamServiceImpl` currently gets this right (`@Service @Primary`), and no caller injects the concrete `TeamServiceImpl`, so the override does resolve today; but nothing guards against regression, and a plain unit test of the community class cannot detect it. Add a context/integration assertion that the injected `TeamService` bean is an instance of the enterprise subclass (e.g. `assertThat(AopProxyUtils.ultimateTargetClass(teamService)).isEqualTo(EpTeamServiceImpl.class)`), and while here, wire up or delete the dead `EpJobServiceImpl` so the jobs cache is not left permanently stale.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/TeamServiceImpl.java
ThinuwanW, line 14: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

Efficiency of the path this change makes hotter: every invalidation forces the external consumer to re-trigger the cache-population read, and that read is an N+1. `teamDao.findByIsActive(true)` (used by both `getAllTeamDetails` here and `EpTeamServiceImpl.getAllTeams`) is a derived query with no fetch join, while `Team.employees` is a lazy `@OneToMany` and `EmployeeTeam.employee` is a lazy `@ManyToOne` - so mapping teams to DTOs issues one query per team plus one per membership row. With four write paths now evicting the cache, that fan-out runs after every create/update/bulk-create/delete instead of once per TTL window. Consider an `@EntityGraph(attributePaths = {"employees", "employees.employee"})` on the `findByIsActive` query (or a JOIN FETCH variant in `TeamRepositoryImpl`) so the list-all-teams path is a single query.

ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds a `invalidateTeamsCache()` template-method hook to `TeamServiceImpl` and calls it from the four team write paths (addNewTeam, updateTeam, addTeams, transferMembersAndDeleteTeam), overridden in the enterprise `EpTeamServiceImpl` to delete the Redis `teams_cache:all_teams` key. The pattern itself mirrors the existing `invalidateUserCache()` hook in `PeopleServiceImpl`, so it is consistent with the codebase. Two real correctness gaps remain: invalidation happens inside the open transaction (before commit) in the two `@Transactional` methods, and the other code paths that mutate team membership (employee create/update/bulk-upload/termination in `PeopleServiceImpl`) still never invalidate this cache — with a 7-day TTL, both can leave stale team data served to consumers for a week.

Found **5** new issue(s): 🔴 2 important, 🟡 2 suggestion(s), 🟣 1 nit(s)
_6 finding(s) already posted in a previous review._

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/TeamServiceImpl.java
ThinuwanW, line 14: 🤖 **Claude** · 🔴 **Important**

📁 **File-level observation**

Missing paired change: the diff contains only this community file and no enterprise submodule pointer bump. The behaviour this PR is named for lives entirely in the submodule — `EpTeamServiceImpl.invalidateTeamsCache()` (the override) and `EpCacheKeys.TENANT_ALL_TEAMS_CACHE_KEY`. Without a gitlink update for `backend/src/main/java/com/skapp/enterprise`, merging this PR adds four calls to an empty method and changes nothing in production. Please confirm the enterprise override is already merged on the submodule commit that `main` points at; if it is not, the submodule bump belongs in this PR so the two halves land atomically. Related: `TeamService` (the interface) does not expose any invalidation entry point, so the hook is reachable only via subclassing `TeamServiceImpl`. If any caller outside this class ever needs to invalidate (see the `EpPeopleServiceImpl.invalidateAllUserCaches()` precedent, which is public precisely for that reason), that will require another change.

ThinuwanW, line 14: 🤖 **Claude** · 🔴 **Important**

📁 **File-level observation**

`updateTeam` (the method that now invalidates the shared teams cache at line 181) has no input validation and can corrupt team data. `TeamPatchRequestDto` carries zero bean-validation annotations, while the create counterpart `TeamRequestDto` has `@NotNull`/`@NotEmpty`/`@Size(min=1,max=50)` — so the `@Valid` on the controller's `@RequestBody` (TeamsController.java:67) is a no-op. Three concrete failures on a PATCH that omits fields: (1) `teamSupervisors` omitted -> NPE at line 136 (`teamSupervisorIds.stream()`) -> 500 rather than a 400 through `GlobalExceptionHandler`; (2) `teamMembers` omitted -> NPE at line 165 (`updatedTeamMemberIds.contains(...)`); (3) `teamName` omitted -> `Objects.equals(existingName, null)` is false at line 127, `teamNameCheck(null)` finds nothing, so line 132 sets the team name to null, silently wiping it. There is also no length bound, while `team.team_name` is `length = 50`. This matters more now than before: `EpTeamResponseDto` caches `teamName`, so a nulled name is republished to every consumer of `teams_cache:all_teams` on the next `getAllTeams()`. Fix: add `@NotBlank @Size(min = 1, max = 50)` to `TeamPatchRequestDto.teamName` and `@NotNull` to both list fields, mirroring `TeamRequestDto`; or, if PATCH is genuinely meant to allow partial bodies, guard each block with a null check so an omitted field means "unchanged" instead of "clear".

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/TeamServiceImpl.java
```diff
@@ -177,6 +178,7 @@ public ResponseEntityDto updateTeam(Long id, TeamPatchRequestDto teamPatchReques
 			team.setEmployees(employeeTeamSet);
```
ThinuwanW, line 178: 🤖 **Claude** · 🟡 **Suggestion**

`team.setEmployees(employeeTeamSet)` replaces the managed collection with only the newly added supervisors and members (`supervisorsToBeAdded` + `membersToBeAdded`), discarding every member that was already on the team and is staying. `Team.employees` is `@OneToMany(mappedBy = "team", cascade = ALL)` with no `orphanRemoval`, so nothing is deleted in the DB — but the in-memory entity is now wrong, and it is mapped straight into the response at line 184. `TeamResponseDto.employees` therefore lists only the newly added people, so a PATCH that adds one member to a ten-member team returns a team with one member. This also means any code reading `team.getEmployees()` later in the same request sees a truncated roster. Fix: add the new `EmployeeTeam` rows to the existing collection (`team.getEmployees().addAll(employeeTeamSet)` after removing the deleted ones) instead of replacing it, or re-read the team via `teamDao.findByTeamIdAndIsActive(id, true)` after the flush and map that for the response.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/TeamServiceImpl.java
```diff
@@ -276,6 +280,7 @@ public ResponseEntityDto transferMembersAndDeleteTeam(Long id,
 
 			employeeTeamDao.deleteEmployeeTeamByTeamId(id);
```
ThinuwanW, line 281: 🤖 **Claude** · 🟡 **Suggestion**

Redundant deletes in `transferMembersAndDeleteTeam`. Line 281 unconditionally deletes every `employee_team` row for team `id`, which fully subsumes the three earlier scoped deletes: line 245 (`deleteEmployeeTeamByTeamId(id)` in the empty-list branch — literally the same call), line 267 (`deleteEmployeeTeamByTeamIdAndEmployeeIds(id, employeesToUnassignIds)`), and line 272 (same call for the transfer list). The new rows written at line 277 always belong to a different team — `getFilteredTransferTeamMembers` throws if any transfer targets `team.getTeamId()` — so line 281 cannot delete them. Net effect is one to three extra DELETE statements per request and three code paths a reader has to reason about. Suggest dropping lines 245, 267 and 272 and relying on the single delete at 281, or, if the scoped deletes are intentional, dropping 281 and covering the remaining-members case explicitly. Worth a quick verification against the transfer flow before removing, since this is pre-existing logic.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/TeamServiceImpl.java
```diff
@@ -221,8 +223,10 @@ public ResponseEntityDto addTeams(TeamsRequestDto teamsRequestDto) {
 			})
 			.toList();
 
-		if (!teams.isEmpty())
+		if (!teams.isEmpty()) {
```
ThinuwanW, line 226: 🤖 **Claude** · 🟣 **Nit**

Two small issues in the block this PR touched. (1) `addTeams` treats already-existing team names as a non-error: line 211 logs at `error` level and execution continues, so a bulk create where some names collide returns 201 with only the subset that was created and no indication of what was skipped. The caller cannot distinguish "all created" from "half created". Consider returning the skipped names in the response or throwing `PEOPLE_ERROR_TEAM_NAME_ALREADY_EXISTS`, consistent with how `addNewTeam` (line 71) handles the same collision. (2) The log message at line 211 is prefixed `addNewTeams:` but the method is `addTeams` — the convention elsewhere in this class is `<methodName>: ...`, so this breaks log grepping by method name.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/TeamServiceImpl.java
shakila-kamalasena: Not needed here. PeopleServiceImpl isn't part of this PR; adding the membership-eviction hook there is a separate change. Out of scope for this file.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/TeamServiceImpl.java
shakila-kamalasena: Not needed here. The community hook is a no-op; the real eviction and its after-commit handling live in EpTeamServiceImpl in the enterprise submodule, not this file.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/TeamServiceImpl.java
shakila-kamalasena: Not needed. The missing @Transactional on those methods is pre-existing, and the community hook is a no-op so nothing new can fail here.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/TeamServiceImpl.java
shakila-kamalasena: Not needed. The template-hook matches the existing invalidateUserCache/invalidateJobsCache pattern; switching to application events is a broader refactor beyond this PR.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/TeamServiceImpl.java
shakila-kamalasena: Not needed. We don't add tests here, and the community implementation is a no-op.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/TeamServiceImpl.java
shakila-kamalasena: Not needed here. The catch behavior is in the enterprise override (EpTeamServiceImpl/EpCacheServiceImpl), which isn't part of this PR.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/TeamServiceImpl.java
shakila-kamalasena: Not needed. EpTeamServiceImpl registration and EpJobServiceImpl are enterprise submodule files, not part of this PR.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/TeamServiceImpl.java
shakila-kamalasena: Not related to this PR. findByIsActive and the lazy mapping are pre-existing; this PR adds only the invalidation hook, not the read path.

shakila-kamalasena: Not needed. The enterprise override lands via its own PR (#733) on the enterprise submodule; per repo convention this community PR doesn't bump the submodule pointer, so the two halves are merged separately.

shakila-kamalasena: Not related to this PR. The updateTeam validation gap and TeamPatchRequestDto are pre-existing and in other files. Adding bean-validation annotations also conflicts with our convention (no entity/DTO validation annotations; validation is service-layer).

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/TeamServiceImpl.java
shakila-kamalasena: Not related to this PR. The setEmployees logic in updateTeam is pre-existing; this PR only adds the cache-invalidation call.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/TeamServiceImpl.java
shakila-kamalasena: Not related to this PR. These deletes are pre-existing logic in transferMembersAndDeleteTeam, unchanged by this PR.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/TeamServiceImpl.java
shakila-kamalasena: Not related to this PR. The addTeams collision handling and the addNewTeams: log prefix at line 211 are pre-existing; this PR only added the cache-invalidation call in that method.

akilarootcode (approved):

ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds a `invalidateTeamsCache()` template-method hook to the community `TeamServiceImpl` and calls it after each team mutation, mirroring the existing `invalidateJobsCache()` pattern in `JobServiceImpl`; the enterprise `EpTeamServiceImpl` override invalidates the `teams_cache:all_teams` Redis key (7-day TTL). The hook placement inside `TeamServiceImpl` is correct and consistent with the established pattern, but the invalidation coverage is incomplete: team membership is also mutated outside this service (employee create/update, bulk import, supervisor transfer), and the cached DTO carries `employeeIds`, so those paths still leave a stale cache for up to 7 days. Secondary concerns: invalidation runs inside the transaction rather than after commit, two of the four mutating methods are not `@Transactional`, and there is no test coverage for the new hook.

Found **1** new issue(s): 🟡 1 suggestion(s)
_10 finding(s) already posted in a previous review._

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/TeamServiceImpl.java
```diff
@@ -319,6 +324,10 @@ public ResponseEntityDto getTeamsForCurrentUser() {
 		return new ResponseEntityDto(false, response);
 	}
 
+	protected void invalidateTeamsCache() {
+		// This method is a placeholder for enterprise cache invalidation logic
```
ThinuwanW, line 328: 🤖 **Claude** · 🟡 **Suggestion**

The key this hook invalidates has no read-through path inside the backend, so the new calls buy nothing in-process. Grepping TENANT_ALL_TEAMS_CACHE_KEY shows exactly two uses: the put in EpTeamServiceImpl.getAllTeams and the invalidate in the enterprise override of this hook — there is no cacheService.get for it anywhere. Every other cached read in the codebase (UserVersionServiceImpl, SystemVersionServiceImpl, EpOrganizationServiceImpl.getOrganizationTimeZone, EpGuestUserCacheServiceImpl, AITokenServiceImpl) follows get -> if null load -> put, but getAllTeams unconditionally queries the DB and overwrites the key, so the entry is only ever consumed by an out-of-process reader hitting Redis directly. Please confirm that consumer exists and that it repopulates (it can only do so via the ROLE_INTERNAL_API endpoint) — otherwise a deleted key reads as "no teams" until someone calls that API. The complete fix is to make getAllTeams read-through like the rest of the codebase, at which point this invalidation becomes meaningful for in-process reads too.

PR #1697, opened 2026-07-31
feat: send add/remove role restriction changes FE

anusham-rc:
## What

Frontend counterpart to #1695. `PATCH /v1/roles/restrictions` now takes an **add/remove delta** of role levels instead of the `isAdmin` / `isManager` booleans, so the restricted-user-roles modal now submits the changes the user made rather than the full state.

> Base is `feat/skapp-role-restriction-fe` (Phase 1 FE) so this diff shows only the write-side change. Retarget once the parent merges. Depends on backend #1695.

## Changes

- **`UserRoleRestrictionsUpdateType`** — replaced `isAdmin` / `isManager` with `add: RoleLevel[]` / `remove: RoleLevel[]`.
- **`getRestrictionChanges`** (new util) — diffs the current checkbox selection against the initially-restricted roles: `add` = newly checked, `remove` = newly unchecked.
- **`RestrictedUserRolesModal`** — `handleSubmit` now sends `{ module, add, remove }`; dropped the `SECONDARY_ROLE_LEVELS` set and the `isManager` derivation.

## Tests

Added 6 `getRestrictionChanges` cases (add, remove, swap, no-change, from-empty, clear-all). Full util suite: **18/18 pass**.


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR migrates the role-restriction modal from the legacy `isAdmin`/`isManager` boolean pair to a `RoleLevel[]` model, driving the checkbox list from a server-provided `restrictableRoles` list and submitting an `add`/`remove` diff. The refactor itself is clean and well tested at the util level (order-insensitive change detection is a genuine improvement over Formik's `dirty`, which would have false-positived on toggle order). However, the frontend now depends on a backend contract that does not exist in this repository: the GET response DTO returns only `module`/`isAdmin`/`isManager` (no `restrictions`/`restrictableRoles`), which will crash the modal on open, and the PATCH DTO accepts `restrictions` rather than `add`/`remove`, so the new payload will be silently dropped by Jackson and effectively wipe/invert the module's restrictions.

Found **21** new issue(s): 🔴 5 important, 🟡 8 suggestion(s), 🟣 8 nit(s)

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
```diff
@@ -64,52 +83,40 @@ const RestrictedUserRolesModal = ({ initialData }: Props) => {
...
-      case Modules.INVOICE:
-      case Modules.PM:
-        return ["isAdmin"];
-      case Modules.CRM:
-        return ["isAdmin", "isManager"];
-      default:
-        return [];
-    }
-  };
+  const restrictableRoles = initialData.restrictableRoles;
```
ThinuwanW, line 110: 🤖 **Claude** · 🔴 **Important**

Runtime crash: `initialData.restrictableRoles` and `initialData.restrictions` are not returned by the API. The GET endpoint backing `getUserRoleRestrictions` is `GET /v1/roles/restrictions/{module}` (RolesController), which returns `ModuleRoleRestrictionResponseDto` — that DTO only declares `module`, `isAdmin`, `isManager`. There is no `restrictions` or `restrictableRoles` field, and no enterprise override of that endpoint exists in this repo. So at runtime `initialData.restrictions` is `undefined`, which makes `hasSelectionChanged(undefined, undefined)` throw on `.length` (line 112) and `restrictableRoles.map(...)` throw (line 139) — the modal will fail to render as soon as it is opened. The previous code was explicitly defensive here (`initialData !== undefined ? initialData?.isAdmin : false`) and that safety net was removed along with the optional prop type. Fix by landing the backend response change (add `restrictions: List<RoleLevel>` and `restrictableRoles: List<RoleLevel>` to `ModuleRoleRestrictionResponseDto` and populate them in `RolesServiceImpl.getRestrictedRoleByModule`, including the empty-restriction branch) before this frontend change, and in the meantime harden the frontend: type the fields optional in `UserRoleRestrictionsType` and normalize in the API layer, e.g. return `{ ...results[0], restrictions: results[0]?.restrictions ?? [], restrictableRoles: results[0]?.restrictableRoles ?? [] }` from `getUserRoleRestrictions`, so a missing field degrades to an empty checkbox list instead of a white screen.

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
```diff
@@ -64,52 +83,40 @@ const RestrictedUserRolesModal = ({ initialData }: Props) => {
...
+      values.selected,
+      initialData.restrictions
+    );
+
+    const payload: UserRoleRestrictionsUpdateType = {
       module: moduleType,
-      isAdmin: values.isAdmin,
-      isManager:
-        moduleType === Modules.ESIGN ? values.isSender : values.isManager
+      add,
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Request contract mismatch: the new payload `{ module, add, remove }` does not match the endpoint it is PATCHed to. `PATCH /v1/roles/restrictions` binds `ModuleRoleRestrictionRequestDto`, which declares `module`, `isAdmin`, `isManager`, and `restrictions` — there are no `add`/`remove` properties. Spring Boot leaves `FAIL_ON_UNKNOWN_PROPERTIES` disabled (nothing in JacksonConfig or application properties re-enables it), so this will not 400; `add`/`remove` are silently discarded and the server sees `restrictions == null`, `isAdmin == null`, `isManager == null`. `RolesServiceImpl.updateRoleRestrictions` then persists a `ModuleRoleRestriction` row with `isAdmin = null` / `isManager = null` and a `ModuleRolesRestriction` with `restrictions = null`. That is worse than a no-op: `processModuleRoles` computes `isAdminAllowed = isSuperAdmin || (restriction == null || Boolean.FALSE.equals(restriction.getIsAdmin()))`, and `Boolean.FALSE.equals(null)` is `false`, so after any save from this modal the Admin and Manager roles become un-grantable for every non-super-admin in that module — the opposite of what the user selected. Either land the backend `add`/`remove` support first, or keep sending the shape the current backend understands (`{ module, restrictions: values.selected }`) until it does.

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
```diff
@@ -5,18 +5,37 @@ import { useFormik } from "formik";
...
+} from "~community/configurations/types/UserRolesTypes";
+import {
+  getRestrictionChanges,
+  hasSelectionChanged,
+  toggleRoleLevel
+} from "~community/configurations/utils/userRoles/roleRestrictionUtils";
 
 import styles from "./styles";
 
+const ROLE_LEVEL_LABEL_KEYS: Record<RoleLevel, string> = {
```
ThinuwanW, line 26: 🤖 **Claude** · 🟡 **Suggestion**

`ROLE_LEVEL_LABEL_KEYS` advertises UI support for all eight `RoleLevel` values (EMPLOYEE, GUEST, NONE, SALES_REPRESENTATIVE included), but the backend can only enforce two restrictions per module: `ModuleRoleRestriction` stores just `isAdmin`/`isManager`, `RolesServiceImpl.isRoleAllowed` returns `true` by default for anything other than ADMIN/MANAGER/SENDER/SALES_MANAGER, and `validateRestrictedRoleAssignment` only reads those two booleans. `getSecondaryRestrictionRole` collapses the whole selection down to a single secondary role. So if the backend ever returns EMPLOYEE/GUEST/NONE/SALES_REPRESENTATIVE in `restrictableRoles`, the user will be able to tick those checkboxes, get a success toast, and have the restriction silently ignored at enforcement time. Confirm the enforcement side supports arbitrary role levels before exposing them, or restrict the map/UI to the role levels that are actually enforced. Also consider moving this map next to `roleRestrictionUtils.ts` so it is reusable and unit-testable rather than living inline in the component.

frontend/src/community/configurations/api/userRolesApi.ts
```diff
@@ -49,7 +50,7 @@ export const useUpdateUserRoleRestrictions = (
   onError: () => void
 ) => {
   return useMutation({
-    mutationFn: (payload: UserRoleRestrictionsType) => {
+    mutationFn: (payload: UserRoleRestrictionsUpdateType) => {
```
ThinuwanW, line 55: 🤖 **Claude** · 🟡 **Suggestion**

`useUpdateUserRoleRestrictions` still has no `queryClient.invalidateQueries()`. This mutation changes data that is cached under two keys: `userRolesQueryKeys.USER_ROLE_RESTRICTIONS(module)` (consumed by `useGetUserRoleRestrictions`) and `userRolesQueryKeys.ALLOWED_GRANTABLE_PERMISSIONS`, which feeds the role dropdowns in `SystemPermissionForm` and `useSystemPermissionFormHandlers` under People. After a super admin changes restrictions, those dropdowns keep serving stale allowed-role lists for the rest of the session. Add `const queryClient = useQueryClient()` and invalidate both keys in `onSuccess` before calling the caller's `onSuccess`. This matters more now than before: the new flow computes `add`/`remove` by diffing against the last-read restrictions, so any staleness in that baseline produces a wrong diff rather than just a stale display.

frontend/src/community/configurations/types/UserRolesTypes.ts
```diff
@@ -27,6 +27,14 @@ export interface UserRoleRestrictionsType {
   module: Modules;
   isAdmin: boolean;
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`isAdmin` and `isManager` on `UserRoleRestrictionsType` are now dead: grepping the frontend shows no remaining reader of either field after this PR removed the Formik `isAdmin`/`isManager`/`isSender` values. They are still what the backend actually returns, so leaving them is not harmful, but the interface is now half-old/half-new and reads as if both models are live. Either drop them (and rely on `restrictions`) or add a short comment marking them as the deprecated pre-migration shape so the next reader knows which pair is authoritative.

frontend/src/community/configurations/utils/userRoles/roleRestrictionUtils.ts
```diff
@@ -0,0 +1,24 @@
...
+  roleLevel: RoleLevel
+): RoleLevel[] =>
+  selected.includes(roleLevel)
+    ? selected.filter((selectedRole) => selectedRole !== roleLevel)
+    : [...selected, roleLevel];
+
+export const hasSelectionChanged = (
+  selected: RoleLevel[],
+  initialSelected: RoleLevel[]
+): boolean =>
```
ThinuwanW, line 14: 🤖 **Claude** · 🟡 **Suggestion**

`hasSelectionChanged` re-implements the same set-difference logic as `getRestrictionChanges`, and the two can drift. Since `toggleRoleLevel` guarantees no duplicates, `hasSelectionChanged` is exactly `add.length > 0 || remove.length > 0`. Consider deriving it: `const { add, remove } = getRestrictionChanges(selected, initialSelected); return add.length > 0 || remove.length > 0;`. That also removes the one place where the two differ today — `hasSelectionChanged` leans on a `length` comparison, which would give a wrong answer if the server ever returns a duplicate role level in `restrictions` (the backing column is a comma-joined string, so duplicates are not structurally prevented), whereas the set-difference form is duplicate-tolerant.

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
```diff
@@ -129,34 +136,15 @@ const RestrictedUserRolesModal = ({ initialData }: Props) => {
...
-            )}
-            {restrictableRoles.includes("isSender") && (
+            {restrictableRoles.map((roleLevel) => (
               <Checkbox
-                label={translateText(["senderRoleLabel"])}
-                name="isSender"
-                checked={values.isSender}
-                onChange={() => setFieldValue("isSender", !values.isSender)}
+                key={roleLevel}
+                label={translateText([ROLE_LEVEL_LABEL_KEYS[roleLevel]])}
```
ThinuwanW, line 142: 🤖 **Claude** · 🟣 **Nit**

`ROLE_LEVEL_LABEL_KEYS[roleLevel]` is type-safe only because `restrictableRoles` is typed `RoleLevel[]`; the value actually comes from an untrusted API response. If the backend adds a new role level before the frontend enum is updated, the lookup yields `undefined`, `translateText([undefined])` builds the key `configurations.userRoles.` and i18next echoes that raw key back as the checkbox label. Add a fallback, e.g. `const labelKey = ROLE_LEVEL_LABEL_KEYS[roleLevel]` and skip (or render the raw role level) when it is missing.

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
```diff
@@ -64,52 +83,40 @@ const RestrictedUserRolesModal = ({ initialData }: Props) => {
...
+      remove
     };
 
     updateUserRoleRestrictions(payload);
   };
 
-  // TODO: For ESIGN, the backend uses isManager to represent Sender role.
-  // This mapping will be removed after the backend's migration is completed.
-  const { values, dirty, setFieldValue, resetForm } = useFormik({
+  const { values, setFieldValue, resetForm } = useFormik<{
```
ThinuwanW, line 100: 🤖 **Claude** · 🟣 **Nit**

Formik is now reduced to a plain value container: there is no validation schema, no field-level `onChange` wiring (every checkbox goes through `setFieldValue` with a string path, so a typo in `"selected"` would not be caught by the compiler), and `onSubmit: handleSubmit` is dead code because the Save button calls `handleSubmit` directly rather than Formik's `handleSubmit`. `dirty` — the one Formik feature this component genuinely used — has been replaced by `hasSelectionChanged`. Either wire the button to Formik's `handleSubmit` so `onSubmit`/`isSubmitting` are real, or drop Formik here in favour of `useState<RoleLevel[]>`, which would also make the selection state strongly typed.

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
```diff
@@ -170,7 +158,7 @@ const RestrictedUserRolesModal = ({ initialData }: Props) => {
             <ButtonV2
               variant={"primary"}
               onClick={handleSubmit}
-              disabled={!dirty}
+              disabled={!isSelectionChanged}
```
ThinuwanW, line 161: 🤖 **Claude** · 🟣 **Nit**

The Save button is not disabled while the mutation is in flight, so a double-click fires two PATCHes. Pre-existing, but the new `add`/`remove` diff semantics make a duplicate submit slightly more surprising than the old idempotent full-replacement payload. Destructure `isPending` from `useUpdateUserRoleRestrictions` and fold it into the `disabled` prop (`disabled={!isSelectionChanged || isPending}`).

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
```diff
@@ -5,18 +5,37 @@ import { useFormik } from "formik";
 import Checkbox from "~community/common/components/atoms/Checkbox/Checkbox";
 import Icon from "~community/common/components/atoms/Icon/Icon";
 import Tooltip from "~community/common/components/atoms/Tooltip/Tooltip";
-import { Modules } from "~community/common/enums/CommonEnums";
+import { Modules, RoleLevel } from "~community/common/enums/CommonEnums";
```
ThinuwanW, line 8: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

The util tests are thorough (toggle, order-insensitive change detection, add/remove diffing, immutability — good coverage), but the component itself was substantially rewritten and has no test. The parts most likely to regress are exactly the parts not covered by `roleRestrictionUtils.test.ts`: rendering one checkbox per `restrictableRoles` entry with the right translated label, pre-checking the entries present in `restrictions`, the Save-disabled state, and the exact `{ module, add, remove }` payload handed to the mutation. Consider a `RestrictedUserRolesModal.test.tsx` using `@testing-library/react` with the `MockTheme` wrapper and `screen.getByRole("checkbox", { name: ... })`.

frontend/src/community/configurations/api/userRolesApi.ts
```diff
@@ -7,7 +7,8 @@ import { userRolesQueryKeys } from "~community/configurations/api/utils/QueryKey
 import {
   AllUserRolesResponseType,
   AllowedGrantableRolesType,
-  UserRoleRestrictionsType
+  UserRoleRestrictionsType,
```
ThinuwanW, line 10: 🤖 **Claude** · 🔴 **Important**

📁 **File-level observation**

Unhandled promise rejection / permanently stuck loading state on the fetch that feeds this modal. `getUserRoleRestrictions` does `return data.data.results[0]` with no guard, and its only caller (`pages/community/configurations/user-roles/[module].tsx`, `onPrimaryButtonClick`) does `const data = await getUserRoleRestrictions(formattedModule); setInitialData(data); setIsPending(false);` with no try/catch. Any rejection means `setIsPending(false)` never runs: the "Set Restrictions" button spins forever, no error toast is shown, and the rejection surfaces only as an unhandled promise rejection in the console. This is reachable today: the GET is `@PreAuthorize("hasAnyRole('ROLE_SUPER_ADMIN')")`, so a non-super-admin gets 403; and `mapApiModuleToEnum` returns `Modules.NONE` (empty string) for any unrecognised route param, producing `GET /v1/roles/restrictions/` which will 404/405. This PR makes it worse because the modal's `initialData` prop is now non-optional and it dereferences `.restrictions`/`.restrictableRoles`, so the page is the only remaining place that can absorb a bad response. Fix: wrap the await in try/catch, call `setIsPending(false)` in a `finally`, and surface an error toast (the module already has `useToast` plus `errorToastTitle`/`errorToastDescription` keys). Also guard the accessor, e.g. `return data?.data?.results?.[0]`, consistent with `getSuperAdminCount` which already uses `results?.[0]`.

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
```diff
@@ -64,52 +83,40 @@ const RestrictedUserRolesModal = ({ initialData }: Props) => {
...
+    selected: RoleLevel[];
+  }>({
     initialValues: {
-      isAdmin: initialData !== undefined ? initialData?.isAdmin : false,
-      isManager: initialData !== undefined ? initialData?.isManager : false,
-      isSender:
-        initialData !== undefined && moduleType === Modules.ESIGN
-          ? initialData?.isManager
-          : false
+      selected: initialData.restrictions
```
ThinuwanW, line 104: 🤖 **Claude** · 🔴 **Important**

Null-safety on the new fields, specifically for the empty case. This is not just "the API doesn't return them yet" — the backend deliberately persists an empty restriction list as NULL: in `RolesServiceImpl.buildModuleRolesRestriction` the code is `moduleRolesRestriction.setRestrictions(restrictedRoles.isEmpty() ? null : String.join(",", restrictedRoles))`. So once the GET response is extended, the most common state (a module with no restrictions configured) will very plausibly serialize `restrictions` as `null`, not `[]`, and `initialData.restrictions` flows straight into Formik `initialValues.selected`, into `hasSelectionChanged` (`selected.length`), and into `getRestrictionChanges` (`.filter`) — all of which throw on null. Same for `restrictableRoles` at line 139 (`.map`). Fix: declare both as optional in `UserRoleRestrictionsType` (`restrictions?: RoleLevel[] | null`) so TypeScript forces the handling, and normalise once at the top of the component, e.g. `const restrictions = initialData.restrictions ?? []` / `const restrictableRoles = initialData.restrictableRoles ?? []`, then use those locals everywhere instead of dereferencing `initialData` in three separate places (lines 88, 104, 110, 114).

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
```diff
@@ -129,34 +136,15 @@ const RestrictedUserRolesModal = ({ initialData }: Props) => {
...
-                    ? translateText(["salesManagerRoleLabel"])
-                    : translateText(["managerRoleLabel"])
-                }
-                name="isManager"
-                checked={values.isManager}
-                onChange={() => setFieldValue("isManager", !values.isManager)}
-              />
-            )}
-            {restrictableRoles.includes("isSender") && (
+            {restrictableRoles.map((roleLevel) => (
```
ThinuwanW, line 139: 🤖 **Claude** · 🔴 **Important**

`selected` is seeded from `initialData.restrictions` but the checkboxes are rendered from `initialData.restrictableRoles`, and the two are never reconciled — under the new delta payload this makes a stale restriction permanently unremovable. Concretely: if the server reports a restriction for a role that is not in `restrictableRoles` (module config changed, a role was retired, or the two lists are computed from different sources — the backend keeps two tables, `ModuleRoleRestriction` and `ModuleRolesRestriction`), no checkbox is rendered for it, so the user cannot untick it; it stays in `values.selected` forever; `getRestrictionChanges` therefore puts it in neither `add` nor `remove`; and the restriction silently persists with no way to clear it from the UI. The old code was immune to this because the checkbox set came from a hardcoded per-module allowlist and the payload always sent an explicit boolean for every restrictable role (full replace), so unrenderable state was still overwritten. Fix: seed the form with the intersection, `selected: (initialData.restrictions ?? []).filter((role) => (initialData.restrictableRoles ?? []).includes(role))` — that way any non-restrictable stored role lands in `remove` on the first save and gets cleaned up. If instead these roles are meant to stay, render checkboxes over the union of the two lists so the user can see and clear them.

frontend/src/community/configurations/api/userRolesApi.ts
ThinuwanW, line 10: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

`useGetUserRoleRestrictions` (lines 38-46) is dead code — grepping the whole frontend, the only reference is its own definition. The page fetches imperatively with `await getUserRoleRestrictions(module)` and stores the result in `useState`, so `userRolesQueryKeys.USER_ROLE_RESTRICTIONS(module)` is never populated. Two consequences worth acting on in this PR, since it is the PR that reshapes this data flow: (1) adding `queryClient.invalidateQueries({ queryKey: userRolesQueryKeys.USER_ROLE_RESTRICTIONS(module) })` to the mutation would be a no-op for the restrictions data as long as the page bypasses React Query — the real fix is to consume the hook; (2) per the review guidelines, component-level data fetching into `useState` instead of a React Query hook is the pattern to avoid. Recommended: delete the unused hook, or (better) switch `[module].tsx` to `useGetUserRoleRestrictions(formattedModule)` with an `enabled` flag driven by the modal-open state, which gets you loading/error states, cache invalidation after the mutation, and removes the hand-rolled `isPending` state and the missing try/catch in one move.

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
```diff
@@ -64,52 +83,40 @@ const RestrictedUserRolesModal = ({ initialData }: Props) => {
...
-    // TODO: For ESIGN, the backend uses isManager to represent Sender role.
-    // This mapping will be removed after the backend's migration is completed.
-    const payload: UserRoleRestrictionsType = {
+    const { add, remove } = getRestrictionChanges(
+      values.selected,
+      initialData.restrictions
+    );
+
+    const payload: UserRoleRestrictionsUpdateType = {
       module: moduleType,
```
ThinuwanW, line 92: 🤖 **Claude** · 🟡 **Suggestion**

The payload's `module` and the diff base now come from two different sources of truth: `module` is read from the Zustand store (`moduleType`), while `add`/`remove` are diffed against `initialData.restrictions` (the fetched object, which itself carries a `module` field). Nothing guarantees they agree, and if they ever diverge the delta computed for module A is applied to module B — a silent, hard-to-diagnose data corruption rather than an error. `moduleType` can also legitimately be `Modules.NONE` (empty string), in which case the PATCH is sent with an empty module. Now that `UserRoleRestrictionsType` includes `module`, use `initialData.module` for the payload so the module and the diff base are provably the same, and keep the store purely for modal open/close state.

frontend/src/community/configurations/utils/userRoles/roleRestrictionUtils.ts
ThinuwanW, line 14: 🤖 **Claude** · 🟡 **Suggestion**

`hasSelectionChanged` is incorrect when either list contains duplicates, and duplicates are reachable because the backend stores restrictions as a comma-joined string (`String.join(",", restrictedRoles)`) with no uniqueness constraint, then splits it back out. Example: `hasSelectionChanged([ADMIN, ADMIN], [ADMIN, MANAGER])` returns false (lengths match, and every element of `selected` is a member of `initialSelected`), so the Save button stays disabled even though MANAGER should be removed. `getRestrictionChanges` has the mirror problem: duplicated input produces duplicated entries in `add`/`remove`. Fix by normalising through sets, e.g. build `const initial = new Set(initialSelected)` / `const current = new Set(selected)` and compare `current.size !== initial.size || [...current].some((r) => !initial.has(r))`, and derive `add`/`remove` from the same sets. This also turns both helpers from O(n²) `includes`-in-`filter` scans into O(n) lookups. Consider having `toggleRoleLevel` dedupe on the add branch too, so the form state can never accumulate duplicates from a dirty API payload.

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
ThinuwanW, line 26: 🤖 **Claude** · 🟡 **Suggestion**

`ROLE_LEVEL_LABEL_KEYS` is a domain mapping (backend enum → i18n key) but it lives unexported inside a component file, so it cannot be reused or unit-tested. `ModuleRolesTable` in the same module already carries its own role-label keys (`salesManagerHeader`, `salesRepresentativeHeader`, and the per-row `label` values from the `data/*.json` files), which is exactly the duplication the guidelines call out. Move this map into a shared location — `~community/configurations/constants/` or next to the new helpers in `utils/userRoles/roleRestrictionUtils.ts` — and export it, so there is one authoritative RoleLevel→label mapping. Tightening the value type from `string` to a literal union of the label keys would also catch typos at compile time.

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
```diff
@@ -129,34 +136,15 @@ const RestrictedUserRolesModal = ({ initialData }: Props) => {
...
-            {restrictableRoles.includes("isSender") && (
+            {restrictableRoles.map((roleLevel) => (
               <Checkbox
-                label={translateText(["senderRoleLabel"])}
-                name="isSender"
-                checked={values.isSender}
-                onChange={() => setFieldValue("isSender", !values.isSender)}
+                key={roleLevel}
+                label={translateText([ROLE_LEVEL_LABEL_KEYS[roleLevel]])}
+                name={roleLevel}
```
ThinuwanW, line 143: 🤖 **Claude** · 🟣 **Nit**

`name={roleLevel}` no longer corresponds to anything: the form has exactly one field, `selected`, and the change handler calls `setFieldValue("selected", ...)` directly. Before this PR the `name` matched the Formik field (`isAdmin`/`isManager`/`isSender`), so it was meaningful; now it renders a DOM input named e.g. `ADMIN` that Formik knows nothing about, which is misleading to the next reader and would break if anyone later wired `handleChange` to it. Either pass `name="selected"` (with the role as `value`) or drop the prop if the `Checkbox` atom can be made to accept it as optional. While here, the atom supports an `ariaLabel` prop that is not being used — the visible `label` covers most cases, but passing it makes the accessible name explicit.

frontend/src/community/common/enums/CommonEnums.ts
```diff
@@ -41,6 +41,17 @@ export enum Modules {
   NONE = ""
 }
 
+export enum RoleLevel {
```
ThinuwanW, line 44: 🤖 **Claude** · 🟣 **Nit**

Verified: these eight values match `com.skapp.community.common.type.RoleLevel` exactly (name and set), so the backend contract is right. The concern is taxonomy sprawl — the frontend already models roles in `~community/common/types/AuthTypes.ts` via `AdminTypes`, `ManagerTypes`, `EmployeeTypes`, `SenderTypes`, and `RepresentativeTypes` (the `ROLE_*_*` granted-authority strings), and `RoleLevel` is now a second, overlapping vocabulary in a different file with no mapping between them. The backend bridges the two in `getRoleForModuleAndLevel(ModuleType, RoleLevel)`; the frontend has no equivalent. Consider colocating `RoleLevel` with the other role enums in `AuthTypes.ts`, or adding a short comment stating that it is the module-agnostic role *level* (mirroring the backend enum) as opposed to the module-scoped granted authorities, so a third taxonomy doesn't appear later.

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
ThinuwanW, line 110: 🤖 **Claude** · 🟣 **Nit**

`const restrictableRoles = initialData.restrictableRoles;` is a leftover shim from the deleted `getRestrictableRoles(moduleType)` call — a plain alias that adds a name without adding meaning. Either inline it at the single use site (line 139) or, better, turn it into the normalised local suggested for the null-safety fix (`initialData.restrictableRoles ?? []`) so the indirection earns its keep. Similarly, `onRoleLevelChange` is re-created every render and passed down to each `Checkbox`; wrapping it in `useCallback` is optional here (the atom is not memoised so it would not save a re-render), but the derived `isSelectionChanged` and the role list are cheap and fine as-is — no `useMemo` needed.

frontend/src/community/configurations/utils/userRoles/roleRestrictionUtils.test.ts
```diff
@@ -0,0 +1,133 @@
...
+      true
+    );
+  });
+
+  it("should report a change when everything is cleared", () => {
+    expect(hasSelectionChanged([], [RoleLevel.ADMIN])).toBe(true);
+  });
+});
+
+describe("getRestrictionChanges", () => {
```
ThinuwanW, line 92: 🤖 **Claude** · 🟣 **Nit**

The suite covers the happy paths well but misses the two cases most likely to bite in production, both of which are driven by real API shapes rather than hypotheticals. Add: (1) duplicate entries in either argument — `hasSelectionChanged([ADMIN, ADMIN], [ADMIN, MANAGER])` currently returns `false`, which is wrong, and `getRestrictionChanges` currently emits duplicates in `add`/`remove`; these will lock in the set-based fix. (2) A role present in `restrictions` but absent from `restrictableRoles`, asserting it ends up in `remove` (or is filtered out of the initial selection) so the invisible-unremovable-restriction case has a regression test. Given the helpers are pure and the diff logic is the whole contract with the backend, both are cheap to add.

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
anusham-rc: Outdated — backend PR #1695 (with Phase 1 #1677) added `restrictions` and `restrictableRoles` to the `GET /v1/roles/restrictions/{module}` response, so both fields are returned now. No missing-field crash.

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
anusham-rc: Resolved — backend PR #1695 replaced the request DTO with an `addedRoles`/`removedRoles` delta, and the FE payload was renamed to match (`addedRoles`/`removedRoles`). The contract aligns now.

frontend/src/community/configurations/types/UserRolesTypes.ts
anusham-rc: Fixed — `isAdmin`/`isManager` removed from `UserRoleRestrictionsType`. Backend PR #1695 dropped them from the response DTO too, so `restrictions` is the single authoritative field.

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
anusham-rc: Outdated — the new `module_roles_restriction` table stores a set of `RoleLevel` (not the two booleans), and the UI only renders the `restrictableRoles` the API returns, so it can't over-advertise roles the backend doesn't consider restrictable.

frontend/src/community/configurations/utils/userRoles/roleRestrictionUtils.ts
anusham-rc: Skipping — the duplicate concern is moot now: the response is a `Set<RoleLevel>` and PR #1695 rejects intra-list duplicates with a 400. The DRY tidy-up is fine but not worth the churn here.

anusham-rc: Same as above — with the Set-typed response and backend duplicate validation, duplicates aren't reachable from the API, so skipping the set-based rewrite.

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
anusham-rc: Skipping — `RoleLevel` matches the backend enum exactly (confirmed in the CommonEnums comment), so an undefined label isn't reachable today. Can add a fallback later if the enums ever drift.

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
anusham-rc: Agreed Formik is overkill here, but swapping to `useState<RoleLevel[]>` is a pure refactor with no behavior change — leaving it out of this PR.

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
anusham-rc: Low risk — the delta is idempotent on the backend (add-existing / remove-missing are no-ops), so a double submit is harmless. Disabling on in-flight is nice UX but optional; skipping for now.

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
anusham-rc: Noted — the pure util logic is covered; a component test is worthwhile but out of scope for this PR.

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
anusham-rc: Organization nit — leaving the map inline for now; can extract to a shared exported constant in a follow-up if it gets reused.

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
anusham-rc: Cosmetic and tied to the Formik question — skipping unless we drop Formik, in which case this goes away with it.

frontend/src/community/common/enums/CommonEnums.ts
anusham-rc: Values confirmed to match the backend `RoleLevel` enum. Consolidating the role taxonomies is a broader refactor beyond this PR — leaving as-is.

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
anusham-rc: This folds into the null-safety fix — the alias becomes the normalized local (`restrictableRoles ?? []`), so it'll earn its keep there.

frontend/src/community/configurations/utils/userRoles/roleRestrictionUtils.test.ts
anusham-rc: The non-restrictable-role case will get a test alongside the intersection fix. The duplicate case isn't reachable (Set-typed response + backend duplicate validation), so skipping that one.

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
anusham-rc: `restrictions` serializes as `[]` (not `null`) for the empty case, so the arrays are always safe. No null-handling needed.

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
anusham-rc: With PR #1695 the write path validates `addedRoles` against the module's restrictable roles and the sync job is retired, so `restrictions` stays a subset of `restrictableRoles` — the unremovable case can't be produced. Skipping; if the set of restrictable roles ever changes for a module, that's a data-migration concern for that release rather than silent UI reconciliation.

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
anusham-rc: The payload sourced `module` from the store (`moduleType`) before this PR too, and both `moduleType` and `initialData.module` are set from the same `formattedModule` in the page, so they always agree. Out of scope for this PR's delta change; can revisit as a separate cleanup if  needed

frontend/src/community/configurations/api/userRolesApi.ts
anusham-rc: Fixed — `[module].tsx` now uses `useGetUserRoleRestrictions` (registered with `enabled: false`, fetched on click via `refetch()`), so the hook is the actual data source and the imperative `getUserRoleRestrictions` + hand-rolled `useState`/`isPending` are gone. Aligns with the codebase's React Query convention.

anusham-rc: Fixed — the on-demand `refetch()` result is checked (`isError`/no data) and shows an error toast with an early return, so a 403/404 no longer leaves the button spinning. The button's loading state is now driven by React Query's `isFetching`, and the accessor is guarded (`results?.[0]`), consistent with `getSuperAdminCount`.

frontend/src/community/configurations/api/userRolesApi.ts
anusham-rc: Not needed with this app's config. The restrictions query is fetched on demand (`enabled: false` + `refetch`), so invalidating its key is a no-op. `ALLOWED_GRANTABLE_PERMISSIONS` is consumed only on a separate page (People permission form), and with the QueryClient's default `staleTime: 0` + `refetchOnMount: true`, that view already refetches fresh on mount — so an explicit invalidation here would be redundant.

shakila-kamalasena (dismissed):

frontend/pages/community/configurations/user-roles/[module].tsx
```diff
@@ -24,23 +24,33 @@ const Module: NextPage = () => {
...
-  >();
+  const {
+    data: initialData,
+    isFetching,
+    refetch
+  } = useGetUserRoleRestrictions(formattedModule, false);
 
   const onPrimaryButtonClick = async () => {
-    setIsPending(true);
+    const { data, isError } = await refetch();
```
Comment (akilarootcode) [inline frontend/pages/community/configurations/user-roles/[module].tsx:38]: get data and isError from hook

akilarootcode (approved):

frontend/src/community/configurations/utils/userRoles/roleRestrictionUtils.ts
```diff
@@ -14,3 +14,11 @@ export const hasSelectionChanged = (
 ): boolean =>
   selected.length !== initialSelected.length ||
   selected.some((roleLevel) => !initialSelected.includes(roleLevel));
+
+export const getRestrictionChanges = (
+  selected: RoleLevel[],
+  initialSelected: RoleLevel[]
+): { add: RoleLevel[]; remove: RoleLevel[] } => ({
```
akilarootcode, line 21: add interface { add: RoleLevel[]; remove: RoleLevel[] }

frontend/pages/community/configurations/user-roles/[module].tsx
Comment (akilarootcode) [inline frontend/pages/community/configurations/user-roles/[module].tsx:38]: remove refetch and fix hook issue

akilarootcode (changes requested):

frontend/pages/community/configurations/user-roles/[module].tsx
Comment (anusham-rc) [reply frontend/pages/community/configurations/user-roles/[module].tsx]: handled properly

frontend/src/community/configurations/utils/userRoles/roleRestrictionUtils.ts
anusham-rc: added interface

frontend/pages/community/configurations/user-roles/[module].tsx
Comment (anusham-rc) [reply frontend/pages/community/configurations/user-roles/[module].tsx]: done, handled properly

frontend/pages/community/configurations/user-roles/[module].tsx
```diff
@@ -24,23 +24,31 @@ const Module: NextPage = () => {
...
-  const [initialData, setInitialData] = useState<
-    UserRoleRestrictionsType | undefined
-  >();
+  const { setIsUserRoleModalOpen, setModuleType } = useConfigurationStore();
 
-  const onPrimaryButtonClick = async () => {
-    setIsPending(true);
+  const {
+    data: initialData,
+    isFetching,
```
Comment (akilarootcode) [inline frontend/pages/community/configurations/user-roles/[module].tsx:33]: get from tanstack

akilarootcode (changes requested):

frontend/pages/community/configurations/user-roles/[module].tsx
Comment (anusham-rc) [reply frontend/pages/community/configurations/user-roles/[module].tsx:33]: Switched to the hook's built-in `isLoading` from TanStack Query instead of `isFetching`. `isLoading` covers only the initial fetch (no cached data yet), whereas `isFetching` also fires on background refetches — so the button now reflects just the first load.

frontend/src/community/configurations/api/userRolesApi.ts
anusham-rc: Done — `[module].tsx` now consumes `useGetUserRoleRestrictions` instead of the imperative `getUserRoleRestrictions` + `useState`. It's disabled on mount and fired on click via `refetch()`; `data`/`isError` from the result drive the modal vs. error toast, and `isFetching` drives the spinner. The cache key is now populated, so `invalidateQueries` on the mutation refreshes it.

frontend/pages/community/configurations/user-roles/[module].tsx
Comment (anusham-rc) [reply frontend/pages/community/configurations/user-roles/[module].tsx:33]: Done — the data now comes from the TanStack Query hook (`useGetUserRoleRestrictions`), not the imperative fetch + `useState`. The query is disabled on mount and triggered on click via `refetch()`, with `data`/`isError` from the result driving the modal vs. error toast and `isFetching` driving the button spinner. I kept `refetch()` so the fetch resolves before the modal opens (lets us toast on failure instead of opening an empty modal) — happy to switch to an `enabled`-flag reactive version if you'd prefer to drop it.

akilarootcode (approved):

PR #1698, opened 2026-07-31
feat: add breadcrumbs and common translations to CRM pages

dulsara-manakal:
## What
Added a static breadcrumb ("CRM / Contacts", "CRM / Companies", "CRM / Deals", "CRM / Tasks") to the top of each CRM list view.

## Why
Other modules (People, Leave, Timesheet, etc.) already show a breadcrumb so users know where they are. CRM didn't have one.

## Changes
- Added the `"crm": "CRM"` label under `common` in `crmModule.json`.
- Added a `breadcrumbs` prop to `ContentLayout` in `contacts.tsx`, `companies.tsx`, `deals.tsx`, `tasks.tsx`.
- Both segments are plain static text (no links), matching how other simple list-view breadcrumbs work across the app.

## Notes
- CRM has no detail pages (just side panels), so no entity names in the breadcrumb — kept it simple on purpose.
- Not tested in browser locally (needs backend + login), but code follows the exact same pattern used elsewhere (e.g. People Directory, Leave Requests).

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/frontend/pages/community/crm/companies.tsx
+++ b/frontend/pages/community/crm/companies.tsx
@@ -16,7 +16,7 @@ import { CrmLimitResource } from "~enterprise/crm/types/CrmLimitTypes";
 
 const Companies: NextPage = () => {
   const translateText = useTranslator("crmModule", "companies");
-  const translateCommonText = useTranslator("crmModule", "common");
+  const translateBreadcrumb = useTranslator("crmModule", "breadcrumbs");
   const { guardCrmCreate, isCheckingCrmLimit } = useCrmLimitGuard();
 
   const { setIsCompanyModalOpen, setCompanyModalType, selectedCompanyId } =
@@ -36,7 +36,7 @@ const Companies: NextPage = () => {
   return (
     <ContentLayout
       breadcrumbs={[
-        { label: translateCommonText(["crm"]) },
+        { label: translateBreadcrumb(["crm"]) },
         { label: translateText(["title"]) }
       ]}
       pageHead={translateText(["pageHead"])}
--- a/frontend/pages/community/crm/companies.tsx
+++ b/frontend/pages/community/crm/companies.tsx
@@ -15,8 +15,7 @@ import useCrmLimitGuard from "~enterprise/crm/hooks/useCrmLimitGuard";
 import { CrmLimitResource } from "~enterprise/crm/types/CrmLimitTypes";
 
 const Companies: NextPage = () => {
-  const translateText = useTranslator("crmModule", "companies");
-  const translateBreadcrumb = useTranslator("crmModule", "breadcrumbs");
+  const translateText = useTranslator("crmModule");
   const { guardCrmCreate, isCheckingCrmLimit } = useCrmLimitGuard();
 
   const { setIsCompanyModalOpen, setCompanyModalType, selectedCompanyId } =
@@ -36,12 +35,12 @@ const Companies: NextPage = () => {
   return (
     <ContentLayout
       breadcrumbs={[
-        { label: translateBreadcrumb(["crm"]) },
-        { label: translateText(["title"]) }
+        { label: translateText(["breadcrumbs.crm"]) },
+        { label: translateText(["companies.title"]) }
       ]}
-      pageHead={translateText(["pageHead"])}
-      title={translateText(["title"])}
-      primaryButtonText={translateText(["addCompanyBtn"])}
+      pageHead={translateText(["companies.pageHead"])}
+      title={translateText(["companies.title"])}
+      primaryButtonText={translateText(["companies.addCompanyBtn"])}
       primaryBtnIconName={IconName.ADD_ICON}
       onPrimaryButtonClick={onPrimaryButtonClick}
       isPrimaryBtnLoading={isCheckingCrmLimit}
--- a/frontend/pages/community/crm/companies.tsx
+++ b/frontend/pages/community/crm/companies.tsx
@@ -35,12 +35,12 @@ const Companies: NextPage = () => {
   return (
     <ContentLayout
       breadcrumbs={[
-        { label: translateText(["breadcrumbs.crm"]) },
-        { label: translateText(["companies.title"]) }
+        { label: translateText(["breadcrumbs", "crm"]) },
+        { label: translateText(["companies", "title"]) }
       ]}
-      pageHead={translateText(["companies.pageHead"])}
-      title={translateText(["companies.title"])}
-      primaryButtonText={translateText(["companies.addCompanyBtn"])}
+      pageHead={translateText(["companies", "pageHead"])}
+      title={translateText(["companies", "title"])}
+      primaryButtonText={translateText(["companies", "addCompanyBtn"])}
       primaryBtnIconName={IconName.ADD_ICON}
       onPrimaryButtonClick={onPrimaryButtonClick}
       isPrimaryBtnLoading={isCheckingCrmLimit}
--- a/frontend/pages/community/crm/companies.tsx
+++ b/frontend/pages/community/crm/companies.tsx
@@ -16,6 +16,7 @@ import { CrmLimitResource } from "~enterprise/crm/types/CrmLimitTypes";
 
 const Companies: NextPage = () => {
   const translateText = useTranslator("crmModule", "companies");
+  const translateCommonText = useTranslator("crmModule", "common");
   const { guardCrmCreate, isCheckingCrmLimit } = useCrmLimitGuard();
 
   const { setIsCompanyModalOpen, setCompanyModalType, selectedCompanyId } =
@@ -34,6 +35,10 @@ const Companies: NextPage = () => {
 
   return (
     <ContentLayout
+      breadcrumbs={[
+        { label: translateCommonText(["crm"]) },
+        { label: translateText(["title"]) }
+      ]}
       pageHead={translateText(["pageHead"])}
       title={translateText(["title"])}
       primaryButtonText={translateText(["addCompanyBtn"])}
--- a/frontend/pages/community/crm/tasks.tsx
+++ b/frontend/pages/community/crm/tasks.tsx
@@ -16,6 +16,7 @@ import { CrmLimitResource } from "~enterprise/crm/types/CrmLimitTypes";
 
 const Tasks: NextPage = () => {
   const translateText = useTranslator("crmModule", "tasks");
+  const translateCommonText = useTranslator("crmModule", "common");
   const containerRef = useRef<HTMLDivElement>(null);
   const { guardCrmCreate, isCheckingCrmLimit } = useCrmLimitGuard();
 
@@ -54,6 +55,10 @@ const Tasks: NextPage = () => {
 
   return (
     <ContentLayout
+      breadcrumbs={[
+        { label: translateCommonText(["crm"]) },
+        { label: translateText(["title"]) }
+      ]}
       pageHead={translateText(["pageHead"])}
       title={translateText(["title"])}
       primaryButtonText={translateText(["addTaskBtn"])}
--- a/frontend/pages/community/crm/tasks.tsx
+++ b/frontend/pages/community/crm/tasks.tsx
@@ -15,8 +15,7 @@ import useCrmLimitGuard from "~enterprise/crm/hooks/useCrmLimitGuard";
 import { CrmLimitResource } from "~enterprise/crm/types/CrmLimitTypes";
 
 const Tasks: NextPage = () => {
-  const translateText = useTranslator("crmModule", "tasks");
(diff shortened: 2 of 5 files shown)


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds a two-level breadcrumb trail ("CRM" → page title) to the four community CRM pages and introduces the `crmModule.common.crm` translation key. I verified the `BreadcrumbItem` shape against `skapp-ui`/`breadcrumbSlice`, confirmed `useTranslator("crmModule", "common")(["crm"])` resolves to the newly added key, checked that English is the only locale file for `crmModule`, confirmed all four CRM pages under `pages/**/crm` are covered (no enterprise CRM pages exist), and confirmed the pattern matches existing breadcrumb usage in timesheet/invoice/sign/people pages. Leaving the first crumb without `onClick` is correct here since `ROUTES.CRM.BASE` (`/crm`) has no index page. No correctness or security issues found — only minor consistency/maintainability observations.

frontend/src/community/common/assets/languages/english/crmModule.json
```diff
@@ -1,5 +1,6 @@
 {
   "common": {
+    "crm": "CRM",
```

The new breadcrumb label is placed at `crmModule.common.crm`, which reads oddly (a key named after its own module nested under `common`) and diverges from how other modules namespace breadcrumb labels — `eSignatureModule` uses a dedicated `breadcrumbs` section (`eSignatureModule.breadcrumbs.sign`, consumed via `useTranslator("eSignatureModule", "breadcrumbs")`). Consider moving it to a `"breadcrumbs": { "crm": "CRM" }` section and using `useTranslator("crmModule", "breadcrumbs")` in the pages, so the key follows the `module.section.key` convention and leaves room for future CRM breadcrumb labels (e.g. detail pages) without polluting `common`. Note the existing `common` section only holds `ownerPopupSearch`, so this is a low-cost change today.

frontend/pages/community/crm/companies.tsx
```diff
@@ -16,6 +16,7 @@ import { CrmLimitResource } from "~enterprise/crm/types/CrmLimitTypes";
 
 const Companies: NextPage = () => {
   const translateText = useTranslator("crmModule", "companies");
+  const translateCommonText = useTranslator("crmModule", "common");
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

The exact same breadcrumb array shape — `[{ label: translateCommonText(["crm"]) }, { label: translateText(["title"]) }]` — plus the extra `useTranslator("crmModule", "common")` call is duplicated verbatim across all four CRM pages (companies.tsx:38-41, contacts.tsx:38-41, deals.tsx:38-41, tasks.tsx:58-61), with only the second label's namespace differing. Consider a small shared helper in the CRM module (e.g. `~community/crm/utils/getCrmBreadcrumbs.ts` exporting `getCrmBreadcrumbs(t, pageTitle): BreadcrumbItem[]`, or a `useCrmBreadcrumbs(pageTitle)` hook that owns the `common` translator) so the parent crumb's label/route is defined once. This keeps future changes (e.g. making the CRM crumb navigable, or adding a third level for detail panels) to a single edit instead of four. Note the repo currently inlines breadcrumbs on every page, so this is optional if you prefer to stay consistent with that convention.

frontend/pages/community/crm/companies.tsx
```diff
@@ -34,6 +35,10 @@ const Companies: NextPage = () => {
 
   return (
     <ContentLayout
+      breadcrumbs={[
```
ThinuwanW, line 37: 🤖 **Claude** · 🟣 **Nit**

Pre-existing `ContentLayout` behaviour these breadcrumbs now depend on: `ContentLayout.tsx:241-243` pushes the `breadcrumbs` prop into the Zustand store inside `useEffect(() => setBreadcrumbs(breadcrumbs), [])` with an empty dependency array. Because the array is built inline from `t()` calls on every render, the store only ever receives the labels captured at mount — if the active i18n language changes while a CRM page is open, the page `title` re-renders in the new language while the AppBar breadcrumb keeps the old strings until a route change remounts the layout. Not introduced by this PR (it affects every breadcrumb page), and harmless with the current static English-only bundle, but worth fixing centrally by depending on the serialized labels (e.g. `[JSON.stringify(breadcrumbs)]`) or on `i18n.language` in that effect.

frontend/pages/community/crm/companies.tsx
```diff
@@ -34,6 +35,10 @@ const Companies: NextPage = () => {
...
+        { label: translateCommonText(["crm"]) },
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The breadcrumb trail is entirely inert on all four CRM pages — neither crumb carries an `onClick`, so the parent "CRM" crumb renders as decorative text a user cannot activate. This matters here because `/crm` is a genuinely navigable, role-gated route: `ROUTES.CRM.BASE` is rewritten to `/community/crm/contacts` in `next.config.mjs`, and it is the exact path `middleware.ts` gates for `CRM_ADMIN`, `CRM_SALES_MANAGER`, and `CRM_SALES_REPRESENTATIVE`. The repo's convention for a non-terminal crumb that maps to a real route is to attach a router push — see the dashboard, timesheet, configurations, invoice-customers, and projects pages, which all give their parent crumb an `onClick` that calls `router.push(...)` with the corresponding `ROUTES` constant. Fix: give the first crumb an `onClick` that pushes `ROUTES.CRM.BASE` (importing `useRouter` and `ROUTES`), and apply it consistently to contacts.tsx, deals.tsx, and tasks.tsx. Note this is a judgement call rather than a hard rule — the People and Timesheet module-root crumbs are also inert — so if the team's intent is that module-root crumbs are always non-interactive, leaving these as-is is defensible; the trail is then purely a location indicator.

frontend/src/community/common/assets/languages/english/crmModule.json
```diff
@@ -1,5 +1,6 @@
...
     "ownerPopupSearch": {
```
ThinuwanW, line 3: 🤖 **Claude** · 🟣 **Nit**

Dead translation content sitting in the exact block this PR edits. The `ownerPopupSearch` object (`placeholder`, `searchPlaceholder`, `noResults`) that the new key is inserted above has zero references anywhere in the repository — a repo-wide search for `ownerPopupSearch` returns only this JSON definition itself, and no consumer builds the key dynamically (nothing calls `useTranslator("crmModule", "common", ...)` other than the four lines added by this PR, and no `translateText` call passes `"common"`/`"ownerPopupSearch"` as a suffix). Since you are already touching this section, either delete the block or wire it up to whatever owner-search popup it was written for. Leaving it means translators will be asked to localise strings that are never rendered.

frontend/src/community/common/assets/languages/english/crmModule.json

Duplication of the module display name across namespaces (distinct from where the key is placed): "CRM" as a user-facing label now exists in four separate translation files — `configurations.json` (twice: `userRoles.crm` and a second `crm` entry), `peopleModule.json`, `commonEnterprise.json` as `modules.crmModuleText` (resolved via the `useGetModuleTextFromEnum` hook), and now `crmModule.common.crm`. A rename or rebrand of the module therefore requires four coordinated edits, and the copies can silently drift. The enterprise `crmModuleText` entry cannot be reused from a community page, so the cleanest fix is a single canonical module-name group in the shared community `common.json` (which `english.ts` spreads at the root, e.g. a `modules` section) that all four call sites read from. Low priority given the label is a fixed acronym, but worth noting since this PR adds the fourth copy.

frontend/pages/community/crm/tasks.tsx
```diff
@@ -54,6 +55,10 @@ const Tasks: NextPage = () => {
 
   return (
     <ContentLayout
+      breadcrumbs={[
```
ThinuwanW, line 57: 🤖 **Claude** · 🟣 **Nit**

Possible one-frame layout mismeasurement unique to this page. Tasks pins its scroll container by reading `getBoundingClientRect().top` in a mount effect with an empty dependency array and writing `height: calc(100vh - <offsetTop>px)`, re-running only on window `resize`. The breadcrumbs added here do not render inside `ContentLayout` — `ContentLayout` pushes them into the Zustand `breadcrumbSlice`, and `AppBar` renders them via `PageHeader`'s `left` slot. `AppBar` sits in normal document flow above the page content inside `ContentWithDrawer`, and the store write happens in `ContentLayout`'s passive effect, which React flushes *before* the parent Tasks effect but whose resulting `AppBar` re-render commits *after* Tasks has already measured. So `offsetTop` is captured while the app bar's left slot is still empty; if painting the crumbs increases the app bar's height, the container ends up too tall and nothing recomputes it because no resize event fires. Before this PR the array was always empty on CRM pages, so no post-measure change was possible. This is likely benign — the app bar's right slot always renders a 2.5rem avatar or same-height skeleton, which probably drives the header height — hence nit rather than important. If you want it robust regardless, observe the container with a `ResizeObserver` (or observe the app bar element) instead of relying on a single mount measurement plus `resize`.

frontend/src/community/common/assets/languages/english/crmModule.json
dulsara-manakal: Make a dedicated key as breadcrumbs.

frontend/pages/community/crm/companies.tsx
dulsara-manakal: Checked eSign, Leave, People, Timesheet, Invoice - none extract a shared breadcrumb hook, all inline it per page. Keeping CRM consistent with that convention for now.

frontend/pages/community/crm/companies.tsx
dulsara-manakal: Yes, root cumb is non-interactive according to the requirement.

frontend/pages/community/crm/companies.tsx
dulsara-manakal: As comment said this is a pre-existing issue, and since we only have English for now, this won't be reachable.

frontend/src/community/common/assets/languages/english/crmModule.json
dulsara-manakal: Acknowledge these dead translation keys since we have a separate task to refactor these translation texts and keys keep this for now.

frontend/src/community/common/assets/languages/english/crmModule.json
dulsara-manakal: Acknowledge this duplication. Since we have a separate task to refactor these translation texts and keys keep this for now.

frontend/pages/community/crm/tasks.tsx
dulsara-manakal: Confirmed. `PageHeader` uses a fixed h-[60px], not min-h-, so the app bar's height is constant regardless of breadcrumb content. No race condition possible.

frontend/pages/community/crm/companies.tsx
```diff
@@ -16,6 +16,7 @@ import { CrmLimitResource } from "~enterprise/crm/types/CrmLimitTypes";
 
 const Companies: NextPage = () => {
   const translateText = useTranslator("crmModule", "companies");
+  const translateBreadcrumb = useTranslator("crmModule", "breadcrumbs");
```
hasalarootcode: use one

hasalarootcode (changes requested):

frontend/pages/community/crm/companies.tsx
dulsara-manakal: Used one hook call

hasalarootcode (dismissed):

Deshan1100 (dismissed):

frontend/pages/community/crm/companies.tsx
```diff
@@ -34,9 +34,13 @@ const Companies: NextPage = () => {
 
   return (
     <ContentLayout
-      pageHead={translateText(["pageHead"])}
-      title={translateText(["title"])}
-      primaryButtonText={translateText(["addCompanyBtn"])}
+      breadcrumbs={[
+        { label: translateText(["breadcrumbs.crm"]) },
```
akilarootcode: add , "" breadcrumbs crm

akilarootcode (changes requested):

frontend/pages/community/crm/companies.tsx
dulsara-manakal: added comma-separation

akilarootcode (approved):

Deshan1100 (approved):

hasalarootcode (approved):

PR #1699, opened 2026-07-31
fix: add prefix error indication for contact ID in SidePanelAddDeal

dulsara-manakal:
## What this PR does

Fixes a UI alignment bug in the **Add Deal** side panel (CRM). When the contact field showed a "Contact is required" error, the contact box and the deal name box (and the Save/Cancel buttons) no longer lined up.

## The fix

Passes a new `hasPrefixError` prop to `SubTaskInput`, set to `true` when the contact field has a validation error:

```tsx
hasPrefixError={!!formik.errors.contactId}
```

This tells the component to line up the boxes at the top instead of centering them, so they stay aligned even when only one box shows an error.

## Depends on

This needs the matching change in `skapp-ui` (adds the `hasPrefixError` prop to `SubTaskInput`), published as a separate PR:

👉 **skapp-ui PR:** [link](https://pr.skapp.dev/view/843b9087-7536-4d2a-9111-45f61f49a42c)

This PR won't work correctly until that one is merged and published, and the `@rootcodelabs/skapp-ui` version is bumped here.

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This one-line change adds `hasError` to the `SubTaskInput` in SidePanelAddDeal, deriving the deal-name input's error styling from the `contactId` error as well as the `name` error. The intent (surfacing the missing-contact state) is reasonable, but the condition is not gated by `touched` and is not paired with a matching `errorMessage`, so the name field renders in an error state — with no message — as soon as the user types a valid name while no contact is selected. This also diverges from the `touched && error` gating used two lines above for the contact dropdown and from the `state`/`errorMessage` pairing used consistently across the CRM module.

Found **9** new issue(s): 🔴 1 important, 🟡 4 suggestion(s), 🟣 4 nit(s)

frontend/src/community/crm/components/molecules/SidePanelAddDeal/SidePanelAddDeal.tsx
```diff
@@ -190,6 +190,7 @@ const SidePanelAddDeal: FC<Props> = ({ onClose, defaultContact }) => {
         maxLength={DEAL_NAME_MAX_LENGTH}
         required
         errorMessage={formik.errors.name}
+        hasError={!!formik.errors.name || !!formik.errors.contactId}
```
ThinuwanW, line 193: 🤖 **Claude** · 🔴 **Important**

`hasError` turns on for a `contactId` error without any `touched` check, and `errorMessage` (line 192) still only reflects `formik.errors.name`. Concrete flow with no `defaultContact`: the user types a valid deal name → `validateOnChange` runs the whole schema → `errors.name` clears but `errors.contactId` is set → `hasError` becomes true while `errorMessage` is `undefined`. Result: the deal-name input is styled as invalid with no explanation, even though the name is valid, and the contact dropdown itself shows nothing (its `isInvalid` at line 165-167 is gated by `formik.touched.contactId`, and `setFieldValue` does not mark fields touched, so contact only turns red after `submitForm`). The two fields therefore show inconsistent states for the same underlying error. Fix by gating the cross-field part the same way the dropdown does, e.g. `hasError={!!formik.errors.name || !!(formik.touched.contactId && formik.errors.contactId)}`, or keep the name input's state tied to its own field (`hasError={!!formik.errors.name}`) and rely on the dropdown to surface the contact error. If the intent is to show a message on the name input for the missing contact, also feed it through `errorMessage` so the error state is never message-less.

ThinuwanW, line 193: 🤖 **Claude** · 🟡 **Suggestion**

This line now makes `formik.errors.contactId` user-visible (as an error style), but that error is produced by `contactId: Yup.string().required()` in `~community/crm/utils/dealValidations.ts` (`inlineAddDealValidations`) with no message, so the error string is Yup's untranslated default ("contactId is a required field"). A translated key already exists and is used by `AddDealContactSearch` — pass it into the schema: `contactId: Yup.string().required(translator(["inlineAddDeal", "validations", "contactRequired"]))`. That also lets the panel render the real message instead of a hardcoded one in `AddDealContactSearch`, and prevents the untranslated string leaking if it is ever surfaced via `errorMessage`. The adjacent `.max(DEAL_NAME_MAX_LENGTH)` in the same schema is likewise missing a translated message.

frontend/src/community/crm/components/molecules/SidePanelAddDeal/SidePanelAddDeal.tsx
```diff
@@ -190,6 +190,7 @@ const SidePanelAddDeal: FC<Props> = ({ onClose, defaultContact }) => {
         maxLength={DEAL_NAME_MAX_LENGTH}
         required
         errorMessage={formik.errors.name}
```
ThinuwanW, line 192: 🤖 **Claude** · 🟣 **Nit**

`hasError` is now the only usage of that prop name in the repo, and `@rootcodelabs/skapp-ui` (1.1.1-beta.4) is not installed locally, so I could not verify the prop exists on `SubTaskInput`'s props type or how it interacts with `errorMessage` — please confirm the type-check/build passes and that `hasError` is not simply derived from `errorMessage` internally (in which case this prop would be redundant for the name case). Note the rest of the CRM module derives the visual state from the same field's message (`state={errors.x ? "error" : "default"}` next to `errorMessage={errors.x}`, e.g. ContactModalForm.tsx:174-175, TaskModalForm.tsx:240-241); keeping `hasError` and `errorMessage` driven by the same expression avoids the two drifting apart later.

frontend/src/community/crm/components/molecules/SidePanelAddDeal/SidePanelAddDeal.tsx
ThinuwanW, line 193: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

React Query `enabled` misuse (lines 51-55): `useGetCrmContacts(debouncedContactSearch, DEFAULT_LOOKUP_PAGE_SIZE, true)` hardcodes `enabled: true`. This is the only call site in the codebase that does so — `DealPropertiesSidebar` passes `debouncedContactSearchTerm.length > 0`, `AddDealSidePanel` passes `isOpen`, and `TaskModalForm` passes `isContactSearchEnabled`. Consequences: (a) a contact-lookup request fires on every mount of the inline add-deal row with an empty `searchKeyword`, including the `defaultContact` case where `AddDealContactSearch` renders the read-only `InputField` branch and the dropdown is never shown, so the request is pure waste; (b) because that empty-keyword response populates `contacts`, and the dropdown is rendered with `isOpenOnFocus={true}`, merely focusing the contact field pops a list of arbitrary unrelated contacts rather than search results. Fix: pass `!selectedContact && debouncedContactSearch.length > 0` as the `enabled` argument (and consider a `staleTime` since lookup results are reused across panel opens).

ThinuwanW, line 193: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Null-safety in the submit payload (lines 115-127) — this PR is about surfacing form errors, but the worst error path here is silent. `stageId: initialStageId!` uses a non-null assertion, yet `useGetMappedDealStages` derives `initialStageId` via `stages.find(s => s.stageType === INITIAL)?.id`, so it is legitimately `undefined` when the stages fetch fails or no INITIAL stage exists. Likewise `ownerId: Number(currentUser?.employeeId)` yields `NaN` (serialized as `null`) because `useGetUserPersonalDetails` has a `select` that returns `undefined` when `results` is empty — and in both failure cases `isLoading` is already `false`, so `isFormDisabled` no longer blocks save. The user gets only the generic "Failed to add deal" toast for what is really a client-side precondition failure. Fix: include `isError` from `useGetMappedDealStages` (it is returned but currently ignored) in `isFormDisabled`, and guard inside `handleSubmit` — e.g. `if (initialStageId == null || currentUser?.employeeId == null) { show error toast; return; }` — then drop the `!` assertion, which the strict-mode guidelines flag as silencing a real type.

ThinuwanW, line 193: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

No test coverage for this component or its sibling. The repo has 119 `.test.tsx` files, but `find src/community/crm -name "*.test.tsx"` returns zero — so this behavioural change to error visibility (and the interaction between `hasError`, `errorMessage`, and the dropdown's `touched`-gated `isInvalid`) ships unverified. Add `SidePanelAddDeal.test.tsx` covering: (1) typing a deal name with no contact selected, (2) an empty name with a contact selected, (3) clearing a `defaultContact`, and (4) the `DEAL_EXISTS` server-error path that sets `formik.errors.name`. Wrap in `MockTheme` per the testing guidelines and assert via `screen.getByRole` rather than test ids.

ThinuwanW, line 193: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

Type mismatch on the mutation error handler: `handleCreateDealError` is declared as `(error: ErrorResponse) => void` (line 82) but `useCreateDeal` declares `onError: (error: AxiosError) => void` (crmDealApi.ts:80), and `ErrorResponse` (CommonTypes.ts:39) declares `response` as **required** while an Axios network/timeout failure has `response === undefined`. This is the only place in `src/community/crm` that types a mutation error handler as `ErrorResponse`. Runtime behaviour is currently safe thanks to the optional chaining on line 83, but the declared type is inaccurate and will mislead the next edit into dropping the `?.`. Fix: type the parameter as `AxiosError<...>` (matching the hook contract) or make `response` optional on `ErrorResponse`.

ThinuwanW, line 193: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

Dead i18n keys plus a visual-only disabled state. `crmModule.json` defines `deals.sidePanel.inlineAddDeal.ariaLabels.saving` and `...ariaLabels.loading` (lines 451-452), but neither key is referenced anywhere in `src` — evidence that a busy/announced state was intended and never wired. Meanwhile `isFormDisabled` (lines 112-113, 154-156) is expressed only as `opacity-50 pointer-events-none`, which conveys nothing to assistive tech and does not block keyboard input, so the name field stays editable while the mutation is in flight. Fix: either wire the keys up (`aria-busy={isPending}` / `aria-disabled` on the wrapper, or a `disabled` prop on the inputs) or delete the two unused keys.

ThinuwanW, line 193: 🤖 **Claude** · 🟣 **Nit**

Duplication: `formik.errors.contactId` is now consumed by two sibling expressions in the same JSX tree — line 165-167 for the dropdown's `isInvalid` and line 193 for the name field's `hasError` — each with its own inline condition. Regardless of which condition is ultimately correct, having the same error consumed by two independently written boolean expressions is what allowed them to drift in the first place. Fix: hoist a single derived value near the handlers (e.g. `const isContactInvalid = !!(formik.touched.contactId && formik.errors.contactId);` and `const isNameInvalid = !!formik.errors.name;`) and reference those in both places, so the two fields can only ever agree on when the contact error is considered active.

dulsara-manakal: This won't happen. The errorMessage is passed to the component to show as an InputField error, and hasError is only for UI handling in the error state of both prefix node and inputfield.

dulsara-manakal: Added

frontend/src/community/crm/components/molecules/SidePanelAddDeal/SidePanelAddDeal.tsx
dulsara-manakal: Confirmed everything checked, and both props are intentional.

frontend/src/community/crm/components/molecules/SidePanelAddDeal/SidePanelAddDeal.tsx
dulsara-manakal: This will be handled in this [PR](https://pr.skapp.dev/view/65cbef61-bd03-4e38-990e-903acea093a0).  We can't use `!selectedContact` standalone and not for debounce length.

dulsara-manakal: Keep it like this since moving to a constant doesn't make a big impact

dulsara-manakal: Will not reach to this point since we already guarded it with , `const isFormDisabled = isPending || isStagesLoading || isUserLoading || isCheckingCrmLimit;`

dulsara-manakal: not writing a test for the UI component. Check everything against the design and behavioural requirements; everything works as expected.

dulsara-manakal: Currently we work with AxiosError as well since we are deriving the message key with the ErrorResponse and will implement the existence-checking endpoint as well, and then this will not be happen.

dulsara-manakal: Removed that dead translation keys.

shakila-kamalasena (approved):

akilarootcode (approved):

PR #1700, opened 2026-08-03
fix:leave entitlements update

Deshan1100:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/frontend/src/community/leave/components/molecules/CustomLeaveAllocationFilterBody/CustomLeaveAllocationFilterBody.tsx
+++ b/frontend/src/community/leave/components/molecules/CustomLeaveAllocationFilterBody/CustomLeaveAllocationFilterBody.tsx
@@ -26,7 +26,7 @@ const CustomLeaveAllocationFilterBody: FC<Props> = ({
 
   const { data: leaveTypes } = useGetLeaveTypes();
 
-  const [selectedLeaveTypes, setSelectedLeaveTypes] =
+  const [selectedLeaveTypeIds, setSelectedLeaveTypeIds] =
     useState<string[]>(appliedLeaveTypeIds);
 
   const leaveTypeOptions = (leaveTypes ?? []).map((leaveType) => ({
@@ -35,7 +35,7 @@ const CustomLeaveAllocationFilterBody: FC<Props> = ({
   }));
 
   const handleApply = () => {
-    onApply(selectedLeaveTypes);
+    onApply(selectedLeaveTypeIds);
     onClose();
   };
 
@@ -49,7 +49,7 @@ const CustomLeaveAllocationFilterBody: FC<Props> = ({
       title={translateFilterText(["title"])}
       resetButtonProps={{
         onClick: handleReset,
-        disabled: selectedLeaveTypes.length === 0,
+        disabled: selectedLeaveTypeIds.length === 0,
         children: translateFilterText(["resetBtn"])
       }}
       applyButtonProps={{
@@ -61,9 +61,9 @@ const CustomLeaveAllocationFilterBody: FC<Props> = ({
       <SelectableItemList
         title={translateText(["filterButtonTitle"])}
         items={leaveTypeOptions}
-        selectedValues={selectedLeaveTypes}
+        selectedValues={selectedLeaveTypeIds}
         onChipClick={(leaveTypeId) =>
-          setSelectedLeaveTypes((previous) =>
+          setSelectedLeaveTypeIds((previous) =>
             toggleFilterValue(previous, leaveTypeId)
           )
         }
--- a/frontend/src/community/leave/components/molecules/CustomLeaveAllocationFilterBody/CustomLeaveAllocationFilterBody.tsx
+++ b/frontend/src/community/leave/components/molecules/CustomLeaveAllocationFilterBody/CustomLeaveAllocationFilterBody.tsx
@@ -2,7 +2,7 @@ import {
   BasicFilterStructure,
   SelectableItemList
 } from "@rootcodelabs/skapp-ui";
-import { FC, useMemo, useState } from "react";
+import { FC, useState } from "react";
 
 import { useTranslator } from "~community/common/hooks/useTranslator";
 import { toggleFilterValue } from "~community/common/utils/commonUtil";
@@ -29,14 +29,10 @@ const CustomLeaveAllocationFilterBody: FC<Props> = ({
   const [selectedLeaveTypes, setSelectedLeaveTypes] =
     useState<string[]>(appliedLeaveTypeIds);
 
-  const leaveTypeOptions = useMemo(
-    () =>
-      (leaveTypes ?? []).map((leaveType) => ({
-        label: leaveType.name,
-        value: leaveType.typeId.toString()
-      })),
-    [leaveTypes]
-  );
+  const leaveTypeOptions = (leaveTypes ?? []).map((leaveType) => ({
+    label: leaveType.name,
+    value: leaveType.typeId.toString()
+  }));
 
   const handleApply = () => {
     onApply(selectedLeaveTypes);
--- a/frontend/src/community/leave/components/molecules/CustomLeaveAllocationFilterBody/CustomLeaveAllocationFilterBody.tsx
+++ b/frontend/src/community/leave/components/molecules/CustomLeaveAllocationFilterBody/CustomLeaveAllocationFilterBody.tsx
@@ -1,0 +1,80 @@
+import {
+  BasicFilterStructure,
+  SelectableItemList
+} from "@rootcodelabs/skapp-ui";
+import { FC, useMemo, useState } from "react";
+
+import { useTranslator } from "~community/common/hooks/useTranslator";
+import { toggleFilterValue } from "~community/common/utils/commonUtil";
+import { useGetLeaveTypes } from "~community/leave/api/LeaveTypesApi";
+
+interface Props {
+  appliedLeaveTypeIds: string[];
+  onApply: (leaveTypeIds: string[]) => void;
+  onReset: () => void;
+  onClose: () => void;
+}
+
+const CustomLeaveAllocationFilterBody: FC<Props> = ({
+  appliedLeaveTypeIds,
+  onApply,
+  onReset,
+  onClose
+}) => {
+  const translateText = useTranslator("leaveModule", "customLeave");
+  const translateFilterText = useTranslator("commonComponents", "filterButton");
+
+  const { data: leaveTypes } = useGetLeaveTypes();
+
+  const [selectedLeaveTypes, setSelectedLeaveTypes] =
+    useState<string[]>(appliedLeaveTypeIds);
+
+  const leaveTypeOptions = useMemo(
+    () =>
+      (leaveTypes ?? []).map((leaveType) => ({
+        label: leaveType.name,
+        value: leaveType.typeId.toString()
+      })),
+    [leaveTypes]
+  );
+
+  const handleApply = () => {
+    onApply(selectedLeaveTypes);
+    onClose();
+  };
+
+  const handleReset = () => {
+    onReset();
+    onClose();
+  };
+
+  return (
+    <BasicFilterStructure
+      title={translateFilterText(["title"])}
+      resetButtonProps={{
+        onClick: handleReset,
+        disabled: selectedLeaveTypes.length === 0,
+        children: translateFilterText(["resetBtn"])
+      }}
+      applyButtonProps={{
+        onClick: handleApply,
+        children: translateFilterText(["applyBtn"]),
+        "aria-label": translateFilterText(["applyBtn"])
+      }}
+    >
+      <SelectableItemList
+        title={translateText(["filterButtonTitle"])}
+        items={leaveTypeOptions}
+        selectedValues={selectedLeaveTypes}
+        onChipClick={(leaveTypeId) =>
+          setSelectedLeaveTypes((previous) =>
+            toggleFilterValue(previous, leaveTypeId)
+          )
+        }
+        className="max-h-full"
+      />
+    </BasicFilterStructure>
+  );
+};
+
+export default CustomLeaveAllocationFilterBody;
--- a/frontend/src/community/leave/components/molecules/CustomLeaveAllocationsTable/CustomLeaveAllocationsTable.tsx
+++ b/frontend/src/community/leave/components/molecules/CustomLeaveAllocationsTable/CustomLeaveAllocationsTable.tsx
@@ -34,9 +34,6 @@ import {
   LeaveAllocation
 } from "~community/leave/types/CustomLeaveAllocationTypes";
 
-const chipClassName =
-  "inline-flex w-fit items-center gap-2 rounded-full bg-tertiary-background px-4 py-2";
-
 interface Props {
   searchTerm?: string;
   onSearchTermChange: (searchTerm: string) => void;
@@ -70,7 +67,7 @@ const CustomLeaveAllocationsTable: React.FC<Props> = ({
   const { data: customLeaveData, isLoading } = useGetCustomLeaves(
     currentPage,
     5,
-    searchTerm?.trim(),
+    searchTerm,
     Number(selectedYear),
     leaveTypes
   );
@@ -151,12 +148,22 @@ const CustomLeaveAllocationsTable: React.FC<Props> = ({
             chipStyles={{
               display: "flex",
               justifyContent: "start",
-              maxWidth: "fit-content"
+              maxWidth: "fit-content",
+              backgroundColor: "var(--color-tertiary-background)"
             }}
           />
         ),
         duration: (
-          <div className={chipClassName}>
+          <div
+            style={{
+              display: "inline-flex",
+              alignItems: "center",
+              gap: "0.5rem",
+              backgroundColor: "var(--color-tertiary-background)",
+              borderRadius: "9.375rem",
+              padding: "0.5rem 1rem"
+            }}
+          >
             {leaveAllocation.totalDaysAllocated === 0.5
               ? translateText(["halfDayChip"])
               : `${leaveAllocation.totalDaysAllocated} ${
@@ -167,7 +174,16 @@ const CustomLeaveAllocationsTable: React.FC<Props> = ({
           </div>
         ),
         type: (
-          <div className={chipClassName}>
+          <div
+            style={{
+              display: "inline-flex",
+              alignItems: "center",
+              gap: "0.5rem",
+              backgroundColor: "var(--color-tertiary-background)",
+              borderRadius: "9.375rem",
+              padding: "0.5rem 1rem"
+            }}
+          >
             <span role="img" aria-hidden="true">
               {getEmoji(leaveAllocation.leaveType?.emojiCode || "")}
             </span>
@@ -254,6 +270,13 @@ const CustomLeaveAllocationsTable: React.FC<Props> = ({
     <TableView
       className="body2"
       tableName={TableNames.CUSTOM_LEAVE_ALLOCATIONS}
+      ariaLabel={{
+        regionAriaLabel: translateAria(["tableRegion"]),
+        paginationAriaLabel: translateAria(["pagination"]),
+        previousPageLabel: translateAria(["previousPage"]),
+        nextPageLabel: translateAria(["nextPage"]),
+        getPageAriaLabel: (page) => translateAria(["page"], { page })
+      }}
       headers={tableHeaders}
       rows={transformToTableRows()}
       isLoading={isLoading}
--- a/frontend/src/community/leave/components/molecules/CustomLeaveAllocationsTable/CustomLeaveAllocationsTable.tsx
+++ b/frontend/src/community/leave/components/molecules/CustomLeaveAllocationsTable/CustomLeaveAllocationsTable.tsx
@@ -81,7 +81,7 @@ const CustomLeaveAllocationsTable: React.FC<Props> = ({
     if (customLeaveData?.items) {
       setCustomLeaveAllocations(customLeaveData.items);
     }
-  }, [customLeaveData?.items, setCustomLeaveAllocations]);
+  }, [customLeaveData?.items]);
 
   const handleEdit = useCallback(
     (leaveAllocation: LeaveAllocation) => {
@@ -257,12 +257,12 @@ const CustomLeaveAllocationsTable: React.FC<Props> = ({
       : undefined
   };
 
-  const renderFilterContent = ({ close }: TableViewFilterContentArgs) => (
+  const renderFilterContent = ({ onClose }: TableViewFilterContentArgs) => (
     <CustomLeaveAllocationFilterBody
       appliedLeaveTypeIds={selectedLeaveTypes}
       onApply={handleApplyFilters}
       onReset={handleResetFilters}
-      onClose={close}
+      onClose={onClose}
     />
   );
 
--- a/frontend/src/community/leave/components/molecules/CustomLeaveAllocationsTable/CustomLeaveAllocationsTable.tsx
+++ b/frontend/src/community/leave/components/molecules/CustomLeaveAllocationsTable/CustomLeaveAllocationsTable.tsx
@@ -1,21 +1,32 @@
-import { Box, Theme, useTheme } from "@mui/material";
-import { SelectableItemList } from "@rootcodelabs/skapp-ui";
-import React, { useCallback, useEffect, useMemo, useState } from "react";
+import { EditIcon, IconButton } from "@rootcodelabs/skapp-ui";
+import React, {
+  ChangeEvent,
+  useCallback,
+  useEffect,
+  useMemo,
+  useState
+} from "react";
 
 import AvatarChip from "~community/common/components/molecules/AvatarChip/AvatarChip";
-import FilterButton from "~community/common/components/molecules/FilterButton/FilterButton";
-import RoundedSelect from "~community/common/components/molecules/RoundedSelect/RoundedSelect";
-import Table from "~community/common/components/molecules/Table/Table";
+import TableView from "~community/common/components/organisms/TableView/TableView";
+import type {
+  GridHeader,
+  GridRow,
+  TableViewFilterContentArgs
+} from "~community/common/components/organisms/TableView/types";
 import { TableNames } from "~community/common/enums/Table";
 import { useTranslator } from "~community/common/hooks/useTranslator";
-import { getEmoji } from "~community/common/utils/commonUtil";
+import {
+  getEmoji,
+  removeSpecialCharacters
+} from "~community/common/utils/commonUtil";
 import {
   currentYear,
   getAdjacentYearsWithCurrent,
   nextYear
 } from "~community/common/utils/dateTimeUtils";
 import { useGetCustomLeaves } from "~community/leave/api/LeaveApi";
-import { useGetLeaveTypes } from "~community/leave/api/LeaveTypesApi";
+import CustomLeaveAllocationFilterBody from "~community/leave/components/molecules/CustomLeaveAllocationFilterBody/CustomLeaveAllocationFilterBody";
 import { useLeaveStore } from "~community/leave/store/store";
 import {
   CustomLeaveAllocationModalTypes,
@@ -23,21 +34,18 @@ import {
   LeaveAllocation
 } from "~community/leave/types/CustomLeaveAllocationTypes";
 
-import {
-  tableContainerStyles,
-  tableHeaderCellStyles,
-  tableHeaderRowStyles
-} from "./styles";
+const chipClassName =
+  "inline-flex w-fit items-center gap-2 rounded-full bg-tertiary-background px-4 py-2";
 
 interface Props {
   searchTerm?: string;
-  setHasFilteredData: (hasData: boolean) => void;
+  onSearchTermChange: (searchTerm: string) => void;
   setHasEmptyFilterResults: (isEmpty: boolean) => void;
 }
 
 const CustomLeaveAllocationsTable: React.FC<Props> = ({
   searchTerm,
-  setHasFilteredData,
+  onSearchTermChange,
   setHasEmptyFilterResults
 }) => {
   const translateText = useTranslator("leaveModule", "customLeave");
@@ -46,7 +54,6 @@ const CustomLeaveAllocationsTable: React.FC<Props> = ({
     "entitlement",
     "customLeaveAllocationTable"
   );
-  const theme: Theme = useTheme();
 
   const {
     selectedYear,
@@ -60,37 +67,35 @@ const CustomLeaveAllocationsTable: React.FC<Props> = ({
   } = useLeaveStore((state) => state);
 
   const [selectedLeaveTypes, setSelectedLeaveTypes] = useState<string[]>([]);
-  const [tempSelectedLeaveTypes, setTempSelectedLeaveTypes] = useState<
-    string[]
-  >([]);
 
   const leaveTypes = selectedLeaveTypes.join(",");
   const { data: customLeaveData, isLoading } = useGetCustomLeaves(
     currentPage,
     5,
-    searchTerm,
+    searchTerm?.trim(),
     Number(selectedYear),
     leaveTypes
   );
 
+  const hasRows = (customLeaveData?.items?.length ?? 0) > 0;
+  const isFiltering = !!searchTerm || selectedLeaveTypes.length > 0;
+  const searchPlaceholder = translateText([
+    "CustomLeaveAllocationsSectionSearchBarPlaceholder"
+  ]);
+
   useEffect(() => {
     if (customLeaveData?.items) {
       setCustomLeaveAllocations(customLeaveData.items);
-      setHasFilteredData(customLeaveData.items.length > 0);
       setHasEmptyFilterResults(
-        customLeaveData.items.length === 0 &&
-          (!!searchTerm || selectedLeaveTypes.length > 0)
+        customLeaveData.items.length === 0 && isFiltering
       );
     } else {
-      setHasFilteredData(false);
       setHasEmptyFilterResults(false);
     }
   }, [
     customLeaveData?.items,
-    searchTerm,
-    selectedLeaveTypes,
+    isFiltering,
     setCustomLeaveAllocations,
-    setHasFilteredData,
     setHasEmptyFilterResults
   ]);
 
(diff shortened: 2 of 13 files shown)


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR migrates the custom leave allocations table from the legacy `Table` component to the shared `TableView`/`TableV2` stack, extracts the leave-type filter into a new `CustomLeaveAllocationFilterBody` molecule, and moves the search input from the parent organism into the table toolbar. The migration closely follows the established patterns in `LeaveTypesTable`/`ManagerLeaveRequest`, translation keys all resolve, and pagination is correctly 0-based. The main concerns are the search bar's visibility being tied to in-flight query data (it unmounts when the user clears the input), and pagination not being reset when the search term or year changes, which surfaces false "no results" states.

Found **21** new issue(s): 🔴 4 important, 🟡 8 suggestion(s), 🟣 9 nit(s)

frontend/src/community/leave/components/molecules/CustomLeaveAllocationsTable/CustomLeaveAllocationsTable.tsx
```diff
@@ -291,84 +235,77 @@ const CustomLeaveAllocationsTable: React.FC<Props> = ({
...
+      skeletonRows={5}
+      emptyState={emptyState}
+      pagination={{
+        totalPages: customLeaveData?.totalPages,
+        currentPage,
+        onPageChange: setCurrentPage
+      }}
+      toolbar={{
+        searchBar:
+          hasRows || isFiltering
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The search bar unmounts while the user is interacting with it. `toolbar.searchBar` is rendered only when `hasRows || isFiltering`, where `hasRows` is derived from the live `customLeaveData` and `isFiltering` is `!!searchTerm || selectedLeaveTypes.length > 0`. `useGetCustomLeaves` has no `placeholderData`/`keepPreviousData` (and the QueryClient sets no default), so every keystroke produces a new query key and `customLeaveData` becomes `undefined` while the request is in flight. Repro: type a search term, then delete the last character. `searchTerm` becomes `""` → `isFiltering` is false, and the new query key `[... , "", ...]` has no cache entry (the initial mount used `undefined`, a different key) → `hasRows` is false → the search input disappears mid-typing and reappears (unfocused) once the response lands. The same flicker happens when the last leave-type filter is removed. Two fixes worth combining: (1) normalize the cleared term back to `undefined` so the key matches the initial cached query (e.g. `onSearchTermChange(sanitized || undefined)` with the prop typed accordingly, or `searchTerm?.trim() || undefined` at the query call), and (2) decouple the toolbar from in-flight data, e.g. keep the bar once shown via a ref/state flag, or at minimum `hasRows || isFiltering || isLoading`.

frontend/src/community/leave/components/molecules/CustomLeaveAllocationsTable/CustomLeaveAllocationsTable.tsx
```diff
@@ -126,161 +131,100 @@ const CustomLeaveAllocationsTable: React.FC<Props> = ({
...
   };
 
-  const handleLeaveTypeFilter = (leaveType: { id: string; text: string }) => {
-    setTempSelectedLeaveTypes((prev) =>
-      prev.includes(leaveType.id)
-        ? prev.filter((i) => i !== leaveType.id)
-        : [...prev, leaveType.id]
-    );
+  const handleSearchChange = (event: ChangeEvent<HTMLInputElement>) => {
+    onSearchTermChange(removeSpecialCharacters(event.target.value.trimStart()));
```
ThinuwanW, line 228: 🤖 **Claude** · 🔴 **Important**

`handleSearchChange` (and the year dropdown's `onChange: setSelectedYear` at line 298) do not reset `currentPage`, while `handleApplyFilters`/`handleResetFilters` correctly call `setCurrentPage(0)`. Repro: with more than one page of allocations, navigate to page 2, then type a search term that matches a single employee — the request goes out with `page: 1`, the backend returns an empty page, and the user sees the "No results found" empty state even though a match exists on page 0. Same for switching the year while on a later page. Since the search handler now lives in this component, add `setCurrentPage(0)` inside `handleSearchChange` and wrap the dropdown handler (e.g. `onChange: (year) => { setSelectedYear(year); setCurrentPage(0); }`).

frontend/src/community/leave/components/organisms/CustomLeaveAllocationContent/CustomLeaveAllocationContent.tsx
```diff
@@ -46,26 +43,21 @@ const CustomLeaveAllocationContent: FC = () => {
       customLeaveAllocationSearchTerm.trim() !== "";
 
     if (hasEmptyFilterResults) {
-      setShowSearchBox(true);
       setShowAddButton(false);
       return;
     }
 
     if (!hasData && !hasSearchTerm && !isFilterActive) {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Dead logic left behind by the refactor. In the second branch `!hasData` is necessarily `true`, so `setShowAddButton(true)` is identical to the fall-through `setShowAddButton(!hasData)`. That makes `hasSearchTerm` and `isFilterActive` have no effect on the outcome at all — and `isFilterActive` is only ever set to `true` (in `handleSearchTermChange`), never back to `false`, so it is doubly inert. The whole effect reduces to `setShowAddButton(!hasEmptyFilterResults && !hasData)`, which can be a derived value instead of state + effect. Suggest collapsing it and deleting the `isFilterActive` state and the now-unused deps.

frontend/src/community/leave/components/molecules/CustomLeaveAllocationsTable/CustomLeaveAllocationsTable.tsx
```diff
@@ -1,43 +1,51 @@
...
   CustomLeaveAllocationType,
   LeaveAllocation
 } from "~community/leave/types/CustomLeaveAllocationTypes";
 
-import {
-  tableContainerStyles,
-  tableHeaderCellStyles,
-  tableHeaderRowStyles
-} from "./styles";
+const chipClassName =
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The `chipClassName` literal is now copy-pasted verbatim in three table components (this file, `ManagerLeaveRequest.tsx:34`, `LeaveTypesTable.tsx:20`). Extract it once — e.g. a `TABLE_CHIP_CLASS` constant in `~community/common/constants/` or a small shared `TableChip` atom — so the chip styling can't drift between tables.

frontend/src/community/leave/components/molecules/CustomLeaveAllocationFilterBody/CustomLeaveAllocationFilterBody.tsx
```diff
@@ -0,0 +1,80 @@
...
+    onClose();
+  };
+
+  const handleReset = () => {
+    onReset();
+    onClose();
+  };
+
+  return (
+    <BasicFilterStructure
```
ThinuwanW, line 48: 🤖 **Claude** · 🟣 **Nit**

`handleReset` clears the parent's applied filters via `onReset()` but never clears the local `selectedLeaveTypes`, which is seeded from `appliedLeaveTypeIds` only on mount. This is only correct if the skapp-ui `Popper` unmounts its children when closed — `TableView` always renders `filter.filterContent({ close })` and relies on `Popper`'s `open` prop. If children are kept mounted, reopening the popover shows the chips still selected (and the Reset button still enabled) while the table is unfiltered. Cheap hardening: also call `setSelectedLeaveTypes([])` in `handleReset`. The sibling filter bodies share this assumption, so verify it once for all of them.

frontend/src/community/leave/components/molecules/CustomLeaveAllocationsTable/CustomLeaveAllocationsTable.tsx
```diff
@@ -126,161 +131,100 @@ const CustomLeaveAllocationsTable: React.FC<Props> = ({
...
+                ...leaveAllocation,
+                employee: {
+                  ...leaveAllocation.employee,
+                  employeeId: Number(leaveAllocation.employee.employeeId)
+                },
+                validTo: leaveAllocation.validTo || "",
+                validFrom: leaveAllocation.validFrom || ""
+              })
+            }
+            aria-label={translateText(["editButton.label"], {
```
ThinuwanW, line 207: 🤖 **Claude** · 🟣 **Nit**

The migration drops the per-row `ariaDescription.editButton` (and the row-level `ariaLabel`) that the old `Table` action column consumed; only `aria-label` survives on the `IconButton`. That leaves `leaveModule.customLeave.editButton.description` orphaned in `leaveModule.json`. Either wire the description through (e.g. `aria-description`) as `ManagerTimesheetRequestTable` still does, or delete the now-unused translation key. Note `ManagerLeaveRequest` keeps a row-level `ariaLabel`, so this table is now the odd one out.

frontend/src/community/leave/components/molecules/CustomLeaveAllocationsTable/CustomLeaveAllocationsTable.tsx
```diff
@@ -126,161 +131,100 @@ const CustomLeaveAllocationsTable: React.FC<Props> = ({
...
-      }) || []
-    );
-  }, [customLeaveData?.items, translateText]);
+              recordName
+            })}
+          />
+        )
+      };
+    });
+  }, [customLeaveData?.items, handleEdit, translateText]);
```
ThinuwanW, line 215: 🤖 **Claude** · 🟣 **Nit**

The memoization here is a no-op: `useTranslator` returns a freshly-created closure on every render, so `translateText` changes identity each render and both `useMemo([translateText])` for `tableHeaders` and `useCallback([..., translateText])` for `transformToTableRows` recompute every time. `transformToTableRows` is also invoked immediately at the call site, so `useCallback` buys nothing over `useMemo` on the resulting array. Relatedly, `getAdjacentYearsWithCurrent().map(...)` at line 292 allocates a new options array on every render. Either memoize on stable inputs (`customLeaveData?.items` only, hoisting the year options) or drop the memo wrappers so the code doesn't imply a guarantee it doesn't provide.

frontend/src/community/leave/components/molecules/CustomLeaveAllocationFilterBody/CustomLeaveAllocationFilterBody.tsx
```diff
@@ -0,0 +1,80 @@
...
+  const translateText = useTranslator("leaveModule", "customLeave");
+  const translateFilterText = useTranslator("commonComponents", "filterButton");
+
+  const { data: leaveTypes } = useGetLeaveTypes();
+
+  const [selectedLeaveTypes, setSelectedLeaveTypes] =
+    useState<string[]>(appliedLeaveTypeIds);
+
+  const leaveTypeOptions = useMemo(
+    () =>
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`useGetLeaveTypes()` destructures only `data`; there is no loading or error handling. While the request is in flight the popover renders an empty `SelectableItemList` with a title and no chips, and a failed fetch renders the same silently with no user feedback. Consider pulling `isLoading`/`isError` and rendering a skeleton or an inline error message.

frontend/src/community/leave/components/molecules/CustomLeaveAllocationsTable/CustomLeaveAllocationsTable.tsx
```diff
@@ -291,84 +235,77 @@ const CustomLeaveAllocationsTable: React.FC<Props> = ({
...
+    <TableView
+      className="body2"
+      tableName={TableNames.CUSTOM_LEAVE_ALLOCATIONS}
+      headers={tableHeaders}
+      rows={transformToTableRows()}
+      isLoading={isLoading}
+      skeletonRows={5}
+      emptyState={emptyState}
+      pagination={{
+        totalPages: customLeaveData?.totalPages,
```
ThinuwanW, line 286: 🤖 **Claude** · 🟣 **Nit**

Behavioral difference in the refactor: the old `tableFoot.pagination` was gated by `isEnabled: (items?.length ?? 0) > 0` and defaulted `totalPages` to `1`. The new code always passes the `pagination` object with `totalPages: customLeaveData?.totalPages` (i.e. `undefined` while loading and possibly `0` on an empty result). Please confirm `TableV2` hides the pager for `undefined`/`0` totals, otherwise an empty pager will render under the empty state. If it doesn't, mirror `ManagerTimesheetRequestTable` and pass `pagination` conditionally.

frontend/src/community/leave/components/molecules/CustomLeaveAllocationFilterBody/CustomLeaveAllocationFilterBody.tsx
```diff
@@ -0,0 +1,80 @@
+import {
```
ThinuwanW, line 1: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

New component ships without a `CustomLeaveAllocationFilterBody.test.tsx`. Per the testing guideline, new components should have a colocated test (rendered with the `MockTheme` wrapper, asserting via `screen.getByRole`) covering chip toggling, Apply calling `onApply`/`onClose` with the selected ids, and Reset being disabled when nothing is selected. Noting that the sibling filter bodies also lack tests, so this may be a broader gap rather than a blocker for this PR.

frontend/src/community/leave/components/molecules/CustomLeaveAllocationsTable/CustomLeaveAllocationsTable.tsx
```diff
@@ -291,84 +235,77 @@ const CustomLeaveAllocationsTable: React.FC<Props> = ({
...
-      <Table
-        tableName={TableNames.CUSTOM_LEAVE_ALLOCATIONS}
-        headers={tableHeaders}
-        isLoading={isLoading}
-        rows={transformToTableRows()}
-        tableHead={{
-          customStyles: {
-            row: tableHeaderRowStyles(theme),
-            cell: tableHeaderCellStyles(theme)
+  const emptyState = {
```
ThinuwanW, line 243: 🤖 **Claude** · 🔴 **Important**

`emptyState` is built as an untyped local `const` and then passed as `emptyState={emptyState}`. Because it is not a fresh object literal at the prop site, TypeScript's excess-property check is disabled, so an unknown/misspelled key is accepted silently. That matters here: `actions: [{ label, onClick }]` is used by no other `TableView` call site in the repo — `ManagerLeaveRequest`, `LeaveRequests`, `EmployeeTimesheetRequestTable` and `ManagerTimesheetRequestTable` all pass only `title`/`description`, and the code this replaced used `tableBody.emptyState.noData.button = { label, onClick }`. If `actions` is not the actual `TableV2` empty-state prop name, the "Add custom allocation" CTA disappears from the empty state with no compile error, and combined with `CustomLeaveAllocationContent` hiding the header button when there is no data (`setShowAddButton(!hasData)` → `{!showAddButton && <ButtonV2>}`), a first-time user would be left with no way to create an allocation at all. Fix: either inline the object directly into the JSX prop, or annotate it (`const emptyState: TableV2Props["emptyState"] = { ... }`) so the compiler validates the shape, and confirm against the skapp-ui 1.1.1-beta.5 typings that the CTA key is `actions` and not `button`/`onButtonClick`.

frontend/src/community/leave/components/molecules/CustomLeaveAllocationsTable/CustomLeaveAllocationsTable.tsx
```diff
@@ -126,161 +131,100 @@ const CustomLeaveAllocationsTable: React.FC<Props> = ({
...
-          ariaDescription: {
-            editButton: translateText(["editButton.description"], {
+      return {
+        id: leaveAllocation.entitlementId,
+        employee: (
+          <AvatarChip
+            firstName={leaveAllocation.employee?.firstName}
+            lastName={leaveAllocation.employee?.lastName}
+            avatarUrl={leaveAllocation.employee?.authPic}
+            chipStyles={{
```
ThinuwanW, line 148: 🤖 **Claude** · 🔴 **Important**

Missing paired change. This PR fixes the AvatarChip layout in `ManagerLeaveRequest.tsx` (adds `width: "fit-content"`, `justifyContent: "flex-start"`, and it already had `backgroundColor: "var(--color-tertiary-background)"`) but the equivalent AvatarChip in this refactored table was left with `maxWidth: "fit-content"`, `justifyContent: "start"` and no background override. Two consequences: (1) `AvatarChip/styles.ts` defaults the chip to `backgroundColor: "common.white"`, so the employee chip now renders white while the `duration` and `type` chips in the same row moved from `theme.palette.common.white` to the Tailwind `bg-tertiary-background` — the row is visually inconsistent; (2) `maxWidth: fit-content` without `width: fit-content` is not the same shrink behaviour that was just deemed necessary next door, and `"start"` vs `"flex-start"` is gratuitously inconsistent. Apply the same `chipStyles` (`width: "fit-content"`, `justifyContent: "flex-start"`, `backgroundColor: "var(--color-tertiary-background)"`) here, and ideally hoist the shared chipStyles object next to the shared `chipClassName` so the two tables cannot drift again.

frontend/src/community/leave/components/molecules/CustomLeaveAllocationsTable/CustomLeaveAllocationsTable.tsx
```diff
@@ -60,37 +67,35 @@ const CustomLeaveAllocationsTable: React.FC<Props> = ({
   } = useLeaveStore((state) => state);
 
   const [selectedLeaveTypes, setSelectedLeaveTypes] = useState<string[]>([]);
-  const [tempSelectedLeaveTypes, setTempSelectedLeaveTypes] = useState<
-    string[]
-  >([]);
 
   const leaveTypes = selectedLeaveTypes.join(",");
   const { data: customLeaveData, isLoading } = useGetCustomLeaves(
```
ThinuwanW, line 67: 🤖 **Claude** · 🟡 **Suggestion**

No error state. `useGetCustomLeaves` destructures only `data` and `isLoading`; `isError` is ignored. When the request fails, `customLeaveData` is `undefined`, `rows` is `[]`, `isLoading` is false and `isFiltering` is typically false, so the user sees the `emptyCustomLeaveScreen` copy — "No allocations available / No custom leave allocations have been added yet" — for what is actually a server failure, and they may then try to re-create allocations that already exist. Pull `isError` out of the hook and render a distinct error affordance (or at minimum swap the empty-state title/description to an error message with a retry action) instead of reusing the no-data state. The same gap applies to the `pagination`/`toolbar` blocks, which stay rendered against a failed response.

frontend/src/community/leave/components/molecules/CustomLeaveAllocationFilterBody/CustomLeaveAllocationFilterBody.tsx
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Duplicate leave-types fetch and duplicated option-mapping. There are two `useGetLeaveTypes` hooks in the leave module hitting the same `${LEAVE}/types` endpoint: the one imported here from `~community/leave/api/LeaveTypesApi` uses query key `[leaveTypeQueryKeys.LEAVE_TYPES(params)]` (an array wrapping the key array) with `filterByInUse: false, isCarryForward: false`, while `~community/leave/api/LeaveApi`'s version uses the un-nested `leaveTypeQueryKeys.LEAVE_TYPES(params)`. The page that hosts this component (`pages/community/leave/entitlements/leave-entitlements.tsx:24`) already calls the `LeaveApi` variant and pushes the result into the leave store via `setLeaveTypes`, so opening this popover issues a second, separately-cached request for data already in memory. On top of that, the `leaveTypeOptions` mapping duplicates `useManagerLeaveRequestFilters` (`frontend/src/community/leave/hooks/useManagerLeaveRequestFilters.ts:21`), which is the established pattern the sibling `ManagerLeaveRequestFilterBody`/`TimesheetRequestFilterBody` use. Extract a `useCustomLeaveAllocationFilters` hook (or reuse the shared option mapping) and settle on one `useGetLeaveTypes` so both call sites share a single cache entry.

frontend/src/community/leave/components/organisms/CustomLeaveAllocationContent/CustomLeaveAllocationContent.tsx
```diff
@@ -112,22 +104,11 @@ const CustomLeaveAllocationContent: FC = () => {
...
-          />
-        )}
-        <Box sx={{ marginTop: 2 }}>
-          <CustomLeaveAllocationsTable
-            searchTerm={customLeaveAllocationSearchTerm}
-            setHasFilteredData={setHasFilteredData}
-            setHasEmptyFilterResults={setHasEmptyFilterResults}
-          />
-        </Box>
+        <CustomLeaveAllocationsTable
```
ThinuwanW, line 70: 🤖 **Claude** · 🟡 **Suggestion**

The search term now round-trips through the parent for no reason. The input is rendered by `CustomLeaveAllocationsTable` (`toolbar.searchBar`), which calls `onSearchTermChange` up to this component, which stores it in `customLeaveAllocationSearchTerm` and passes the same value straight back down as `searchTerm`. The parent no longer renders anything that depends on it (the `SearchBox` was removed and `hasSearchTerm` only feeds the now-degenerate branch above). Move the search-term state into `CustomLeaveAllocationsTable` — or into the leave store beside `currentPage`/`selectedYear`, which the table already owns — and delete both the `searchTerm` and `onSearchTermChange` props. That removes a prop-drilling round trip and, as a side benefit, puts the search state in the same place as `currentPage` so resetting the page on search becomes local and obvious.

frontend/src/community/leave/components/organisms/CustomLeaveAllocationContent/CustomLeaveAllocationContent.tsx
```diff
@@ -46,26 +43,21 @@ const CustomLeaveAllocationContent: FC = () => {
...
       return;
     }
 
-    setShowSearchBox(true);
     setShowAddButton(!hasData);
-    setHasFilteredData(hasData);
   }, [
     customLeaveAllocations,
     customLeaveAllocationSearchTerm,
     isFilterActive,
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`isFilterActive` is now write-only state. Its only writer is `handleSearchTermChange`, which sets it to `true` unconditionally — including when the user *clears* the search box, since the change handler fires on every keystroke — so once the user types a single character the flag can never return to `false`. Its only reader is the `!hasData && !hasSearchTerm && !isFilterActive` branch, whose body is now equivalent to the fall-through. Net effect: the whole effect reduces to `setShowAddButton(!hasEmptyFilterResults && !hasData)`. Delete `isFilterActive`/`setIsFilterActive` and the `hasSearchTerm` local (and drop `customLeaveAllocationSearchTerm` from the dependency array), or, if the flag is meant to represent "a filter is applied", give it a real reset path and derive it from the table's applied leave-type filter rather than from the search input.

frontend/src/community/leave/components/molecules/CustomLeaveAllocationsTable/CustomLeaveAllocationsTable.tsx
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Orphaned styles module. Dropping the `tableContainerStyles`/`tableHeaderCellStyles`/`tableHeaderRowStyles` import leaves `CustomLeaveAllocationsTable/styles.ts` unused by its own component; its only remaining consumer is `CustomAllocationsReportTable/CustomAllocationsReportTable.tsx:32`, which reaches in via the deep relative import `"../CustomLeaveAllocationsTable/styles"` — exactly the cross-module relative import the path-alias convention forbids, and now with no colocated component to justify the file's location. Move those three helpers (plus `typographyStyles`, which is also unused) into `CustomAllocationsReportTable/styles.ts` or a shared leave-table styles module, update that import, and delete this file so the refactor doesn't leave a dangling MUI-theme styles file in a directory that is now Tailwind-only.

frontend/src/community/leave/components/molecules/CustomLeaveAllocationsTable/CustomLeaveAllocationsTable.tsx
```diff
@@ -60,37 +67,35 @@ const CustomLeaveAllocationsTable: React.FC<Props> = ({
   } = useLeaveStore((state) => state);
```
ThinuwanW, line 62: 🤖 **Claude** · 🟡 **Suggestion**

`useLeaveStore((state) => state)` subscribes to the entire leave store, so every unrelated write — modal flags, `leaveRequestParams`, `leaveRequestsFilter`, sort changes elsewhere in the module — re-renders this component and re-runs `transformToTableRows()` over the page's rows. The effect's own `setCustomLeaveAllocations(customLeaveData.items)` also re-renders this same subscriber. The sibling table touched by this PR does it correctly (`ManagerLeaveRequest.tsx:74` selects an explicit object). Now that this component also owns the controlled search input, the extra render churn is more visible. Replace with an explicit selector listing only `selectedYear`, `currentPage` and the five setters it actually uses.

frontend/src/community/leave/components/molecules/CustomLeaveAllocationsTable/CustomLeaveAllocationsTable.tsx
```diff
@@ -1,43 +1,51 @@
...
-  tableHeaderRowStyles
-} from "./styles";
+const chipClassName =
+  "inline-flex w-fit items-center gap-2 rounded-full bg-tertiary-background px-4 py-2";
 
 interface Props {
   searchTerm?: string;
-  setHasFilteredData: (hasData: boolean) => void;
+  onSearchTermChange: (searchTerm: string) => void;
   setHasEmptyFilterResults: (isEmpty: boolean) => void;
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Props naming is inconsistent within one interface: `onSearchTermChange` follows the event-callback convention while `setHasEmptyFilterResults` exposes the parent's `useState` setter directly. If the `setHasEmptyFilterResults` channel survives (see the note about moving search state into the table), rename it to `onEmptyFilterResultsChange` so the child's contract is uniformly callback-shaped and doesn't leak that the parent happens to back it with `useState`.

frontend/src/community/leave/components/molecules/CustomLeaveAllocationsTable/CustomLeaveAllocationsTable.tsx
```diff
@@ -291,84 +235,77 @@ const CustomLeaveAllocationsTable: React.FC<Props> = ({
...
+    <CustomLeaveAllocationFilterBody
+      appliedLeaveTypeIds={selectedLeaveTypes}
+      onApply={handleApplyFilters}
+      onReset={handleResetFilters}
+      onClose={close}
+    />
+  );
+
+  return (
+    <TableView
```
ThinuwanW, line 270: 🤖 **Claude** · 🟣 **Nit**

`TableView` accepts an `ariaLabel` group (`regionAriaLabel`, `paginationAriaLabel`, `previousPageLabel`, `nextPageLabel`, `getPageAriaLabel`) and the sibling table migrated in the same effort supplies `regionAriaLabel` (`ManagerLeaveRequest.tsx:272`); this call site supplies none, so the table region and the newly-added pagination controls are unlabelled. Also worth confirming: the removed `SearchBox` wrapped its input in `<Box role="search" aria-label={placeHolder}>` and set `role="searchbox"` on the input, whereas the new `toolbar.searchBar` only gets `aria-label`. Add an `ariaLabel={{ regionAriaLabel: ... }}` (a `leaveAria.entitlement.customLeaveAllocationTable` key would fit the existing namespace) and verify skapp-ui's toolbar search bar still exposes a search landmark.

frontend/src/community/leave/components/molecules/CustomLeaveAllocationsTable/CustomLeaveAllocationsTable.tsx
```diff
@@ -126,161 +131,100 @@ const CustomLeaveAllocationsTable: React.FC<Props> = ({
...
     setCurrentPage(0);
   };
 
-  const handleLeaveTypeFilter = (leaveType: { id: string; text: string }) => {
-    setTempSelectedLeaveTypes((prev) =>
-      prev.includes(leaveType.id)
-        ? prev.filter((i) => i !== leaveType.id)
-        : [...prev, leaveType.id]
-    );
+  const handleSearchChange = (event: ChangeEvent<HTMLInputElement>) => {
```
ThinuwanW, line 227: 🤖 **Claude** · 🟣 **Nit**

The search input is undebounced: every keystroke updates the parent state, which changes the `leaveQueryKeys.CUSTOM_LEAVES(...)` key and fires a fresh request. This carried over from the `SearchBox` version, but since the search is being rewritten here it is a good moment to adopt the established pattern — `WorkLocationsTable` wraps its input with `useDebounce(search, WORK_LOCATION_SEARCH_DEBOUNCE_MS)`. Keep `searchTerm` as the immediate value for the controlled input and feed a debounced copy to `useGetCustomLeaves`.

frontend/src/community/leave/components/molecules/CustomLeaveAllocationFilterBody/CustomLeaveAllocationFilterBody.tsx
Deshan1100: no need test

frontend/src/community/leave/components/molecules/CustomLeaveAllocationFilterBody/CustomLeaveAllocationFilterBody.tsx
Deshan1100: no need a loader

Deshan1100: kept as previous it always gives leave types

frontend/src/community/leave/components/molecules/CustomLeaveAllocationFilterBody/CustomLeaveAllocationFilterBody.tsx
Deshan1100: its clear when popper unmount

frontend/src/community/leave/components/molecules/CustomLeaveAllocationsTable/CustomLeaveAllocationsTable.tsx
Deshan1100: removed kept with previous styles

Deshan1100: kept previous styles

frontend/src/community/leave/components/molecules/CustomLeaveAllocationsTable/CustomLeaveAllocationsTable.tsx
Deshan1100: No need to change this is for onSearchTermChange

frontend/src/community/leave/components/molecules/CustomLeaveAllocationsTable/CustomLeaveAllocationsTable.tsx
Deshan1100: No need kept as previous

frontend/src/community/leave/components/molecules/CustomLeaveAllocationsTable/CustomLeaveAllocationsTable.tsx
Deshan1100: didn't any functional changes

frontend/src/community/leave/components/molecules/CustomLeaveAllocationsTable/CustomLeaveAllocationsTable.tsx
Deshan1100: styles updated

frontend/src/community/leave/components/molecules/CustomLeaveAllocationsTable/CustomLeaveAllocationsTable.tsx
Deshan1100: no need

frontend/src/community/leave/components/molecules/CustomLeaveAllocationsTable/CustomLeaveAllocationsTable.tsx
Deshan1100: didn't any functional changes

frontend/src/community/leave/components/molecules/CustomLeaveAllocationsTable/CustomLeaveAllocationsTable.tsx
Deshan1100: no need

frontend/src/community/leave/components/molecules/CustomLeaveAllocationsTable/CustomLeaveAllocationsTable.tsx
Deshan1100: its working fine checked

frontend/src/community/leave/components/molecules/CustomLeaveAllocationsTable/CustomLeaveAllocationsTable.tsx
Deshan1100: empty state handled correctly checked

frontend/src/community/leave/components/molecules/CustomLeaveAllocationsTable/CustomLeaveAllocationsTable.tsx
Deshan1100: now its handled in UI

frontend/src/community/leave/components/molecules/CustomLeaveAllocationsTable/CustomLeaveAllocationsTable.tsx
Deshan1100: updated

frontend/src/community/leave/components/molecules/CustomLeaveAllocationsTable/CustomLeaveAllocationsTable.tsx
Deshan1100: updated need to display search bar always

frontend/src/community/leave/components/organisms/CustomLeaveAllocationContent/CustomLeaveAllocationContent.tsx
Deshan1100: kept as previous no need to update

frontend/src/community/leave/components/molecules/CustomLeaveAllocationFilterBody/CustomLeaveAllocationFilterBody.tsx
```diff
@@ -0,0 +1,76 @@
...
+  onApply,
+  onReset,
+  onClose
+}) => {
+  const translateText = useTranslator("leaveModule", "customLeave");
+  const translateFilterText = useTranslator("commonComponents", "filterButton");
+
+  const { data: leaveTypes } = useGetLeaveTypes();
+
+  const [selectedLeaveTypes, setSelectedLeaveTypes] =
```
hasalarootcode: rename to mean ids

frontend/src/community/leave/components/molecules/CustomLeaveAllocationFilterBody/CustomLeaveAllocationFilterBody.tsx
hasalarootcode: But onReset is coming from the parent, which means the state is somewhere else. Then this will be an issue

frontend/src/community/leave/components/molecules/CustomLeaveAllocationsTable/CustomLeaveAllocationsTable.tsx
```diff
@@ -73,26 +72,16 @@ const CustomLeaveAllocationsTable: React.FC<Props> = ({
     leaveTypes
   );
 
+  const isFiltering = !!searchTerm || selectedLeaveTypes.length > 0;
```
hasalarootcode: The purpose of this?

frontend/src/community/leave/components/molecules/CustomLeaveAllocationsTable/CustomLeaveAllocationsTable.tsx
```diff
@@ -291,84 +238,81 @@ const CustomLeaveAllocationsTable: React.FC<Props> = ({
...
-        headers={tableHeaders}
-        isLoading={isLoading}
-        rows={transformToTableRows()}
-        tableHead={{
-          customStyles: {
-            row: tableHeaderRowStyles(theme),
-            cell: tableHeaderCellStyles(theme)
+  const emptyState = {
+    title: isFiltering
+      ? translateText(["emptySearchResult", "title"])
```
hasalarootcode, line 245: why filtering showing empty state content

hasalarootcode (changes requested):

frontend/src/community/leave/components/organisms/CustomLeaveAllocationContent/CustomLeaveAllocationContent.tsx
Deshan1100: updated

frontend/src/community/leave/components/molecules/CustomLeaveAllocationFilterBody/CustomLeaveAllocationFilterBody.tsx
Deshan1100: updated

frontend/src/community/leave/components/molecules/CustomLeaveAllocationFilterBody/CustomLeaveAllocationFilterBody.tsx
Deshan1100: checked skapp-ui Popper unmounts its children when closed

frontend/src/community/leave/components/molecules/CustomLeaveAllocationsTable/CustomLeaveAllocationsTable.tsx
Deshan1100: updated variable properly isEmptySearchResult

frontend/src/community/leave/components/molecules/CustomLeaveAllocationsTable/CustomLeaveAllocationsTable.tsx
Deshan1100: this is for isEmptySearchResult

frontend/src/community/leave/components/organisms/CustomLeaveAllocationContent/CustomLeaveAllocationContent.tsx
Deshan1100: search bar updated with table tool bar

hasalarootcode (dismissed):

frontend/src/community/leave/components/molecules/CustomLeaveAllocationsTable/CustomLeaveAllocationsTable.tsx
```diff
@@ -126,161 +115,119 @@ const CustomLeaveAllocationsTable: React.FC<Props> = ({
...
-      }) || []
-    );
-  }, [customLeaveData?.items, translateText]);
+              recordName
+            })}
+          />
+        )
+      };
+    });
+  }, [customLeaveData?.items, handleEdit, translateText]);
```
akilarootcode, line 215: remove callback

frontend/src/community/leave/components/molecules/CustomLeaveAllocationsTable/CustomLeaveAllocationsTable.tsx
```diff
@@ -291,84 +238,81 @@ const CustomLeaveAllocationsTable: React.FC<Props> = ({
...
-            rightButton: filterButton
-          }
-        }}
-      />
-    </Box>
+        ]
+      : undefined
+  };
+
+  const renderFilterContent = ({ close }: TableViewFilterContentArgs) => (
```
akilarootcode: rename to onClose

frontend/src/community/leave/components/molecules/CustomLeaveAllocationsTable/CustomLeaveAllocationsTable.tsx
```diff
@@ -73,26 +72,16 @@ const CustomLeaveAllocationsTable: React.FC<Props> = ({
...
     }
-  }, [
-    customLeaveData?.items,
-    searchTerm,
-    selectedLeaveTypes,
-    setCustomLeaveAllocations,
-    setHasFilteredData,
-    setHasEmptyFilterResults
-  ]);
+  }, [customLeaveData?.items, setCustomLeaveAllocations]);
```
akilarootcode: remove setter

akilarootcode (changes requested):

frontend/src/community/leave/components/molecules/CustomLeaveAllocationsTable/CustomLeaveAllocationsTable.tsx
Deshan1100: used this for pass table row data to parent

frontend/src/community/leave/components/molecules/CustomLeaveAllocationsTable/CustomLeaveAllocationsTable.tsx
Deshan1100: updated

frontend/src/community/leave/components/molecules/CustomLeaveAllocationsTable/CustomLeaveAllocationsTable.tsx
Deshan1100: updated

akilarootcode (approved):

hasalarootcode (approved):

PR #1701, opened 2026-08-03
feat: implement StageLabel component and update related components for better stage representation

YasinduRC:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR extracts the repeated stage color-dot + name markup into a new `StageLabel` atom and adopts it in the deals table and the two stage dropdowns, adds a `title` tooltip to the Kanban lane heading, and marks the contact field on the deal edit sidebar as required. The refactor is behavior-preserving for the table (the removed `getStageByName` dep is stable, so the narrowed `useMemo` deps are safe) and correctly reuses existing `required`/`ariaRequired` props that already exist on `PropertyRow`/`ContactPopupSearch`. Main issues found: the Kanban lane's `aria-labelledby` still points at an id the touched `<h2>` does not expose, the new atom double-applies the stage-name translation for the two dropdown call sites (their names come pre-translated from `useGetMappedDealStages`), and the newly-"required" contact field has no invalid state and a clear affordance that silently no-ops.

Found **13** new issue(s): 🔴 2 important, 🟡 6 suggestion(s), 🟣 5 nit(s)

frontend/src/community/crm/components/molecules/DealStageLane/DealStageLaneHeader/DealStageLaneHeader.tsx
```diff
@@ -45,7 +45,11 @@ const DealStageLaneHeader: FC<DealStageLaneHeaderProps> = ({
 
       <div className="flex items-center justify-between gap-2 px-3 pt-3">
         <div className="min-w-0">
-          <h2 id={String(stage.id)} className="subtitle1 truncate capitalize">
+          <h2
+            id={String(stage.id)}
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The `<h2>` you just rewrote still sets `id={String(stage.id)}`, but the wrapping `<section>` (line 39, unchanged) declares `aria-labelledby={`crm-stage-${stage.id}`}`. I grepped the whole crm module: no element anywhere renders an id of the form `crm-stage-<n>`, so the reference dangles and the lane section ends up with no accessible name at all — screen readers announce an unlabelled region. Since this PR is otherwise about accessibility on this exact element, fix it here: change the heading to `id={`crm-stage-${stage.id}`}`. That also removes the secondary problem of emitting a bare numeric DOM id (e.g. `id="3"`), which is globally collision-prone across lanes/other widgets and is not a valid CSS identifier.

frontend/src/community/crm/components/atoms/StageLabel/StageLabel.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+interface StageLabelProps {
+  name: string;
+  color: CrmDealStageColorsEnum;
+}
+
+const StageLabel: FC<StageLabelProps> = ({ name, color }) => {
+  const { getStageByName } = useStageNameMapper();
+
+  const stageName = getStageByName(name);
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`StageLabel` calls `getStageByName(name)` internally, but two of the three call sites feed it names that have *already* been mapped. `useGetMappedDealStages` (hooks/useGetMappedDealStages.ts:19-22) returns `dealStages` as `stages.map(stage => ({...stage, name: getStageByName(stage.name)}))`, and both `DealNameStageSection` and `DealPropertiesSidebar` build their options from `dealStages`. So the translation is applied twice. It happens to be idempotent today only because `getStageByName` guards on `Object.values(DefaultStageNameEnum).includes(name)` and the already-translated value ("Lead") no longer matches the enum key ("LEAD") — a fragile invariant that breaks the moment a locale's translated value collides with an enum key, and it makes the `name` prop contract ambiguous (raw backend key vs. display string). Pick one owner of the mapping: either pass the raw `stages` array (already returned by `useGetMappedDealStages` alongside `dealStages`) at the two dropdown call sites so `StageLabel` is the single translator, or make `StageLabel` a purely presentational atom that takes an already-resolved display name and keep `getStageByName` in `DealsTable`'s `useMemo` as before.

frontend/src/community/crm/components/organisms/DealSidePanel/DealPropertiesSidebar.tsx
```diff
@@ -136,13 +128,14 @@ const DealPropertiesSidebar: FC<DealPropertiesSidebarProps> = ({
       )}
 
       <div className="border border-secondary-accent rounded-lg p-3 flex flex-col gap-2 w-full">
-        <PropertyRow label={translateText(["contact"])}>
+        <PropertyRow label={translateText(["contact"])} required>
```
ThinuwanW, line 131: 🤖 **Claude** · 🟡 **Suggestion**

Marking contact as `required` here (plus `ariaRequired`) makes the field's stated contract stricter than its actual behavior, and it diverges from the equivalent add-deal field. Two gaps: (1) `ariaInvalid` is not passed and no error text is rendered, whereas `AddDealSidePanel/DealPropertiesSection.tsx:91-98` passes `ariaInvalid={!!errors.contactId}` and renders the message — for a deal whose `contactId` is null the sidebar will show a required field with the "none" placeholder and no invalid state at all. (2) `ContactPopupSearch` hardcodes `clearable` on `DropdownWithSearchablePopup` (ContactPopupSearch.tsx:100), and `handleContactChange` (line 95-99) drops the `null` case on the floor, so pressing clear on a now-required field does nothing observable — the value silently snaps back from the store. Either surface an invalid/required state for the empty case, or add a `clearable?: boolean` prop to `ContactPopupSearch` and pass `clearable={false}` here so the affordance matches the requirement.

frontend/src/community/crm/components/organisms/DealsTable/DealsTable.tsx
```diff
@@ -170,17 +168,7 @@ const DealsTable: FC<Props> = ({
...
-              <div
-                className="size-2 rounded-full shrink-0"
-                style={{
-                  backgroundColor: STAGE_COLOR_MAP[deal.stage.color]
-                }}
-              />
-              <span className="body2">{getStageByName(deal.stage.name)}</span>
-            </div>
-          ),
+          stage: <StageLabel name={deal.stage.name} color={deal.stage.color} />,
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Moving `useStageNameMapper` inside `StageLabel` means one `useTranslator`/`useTranslation` hook instance per table row instead of one per table. `DealsTable` is an infinite-scroll list (`hasNextPage`/`onLoadMore`), so with a few hundred accumulated deals that is a few hundred react-i18next subscriptions created and torn down for a value that is identical across rows. If you take the "presentational atom" option from the `StageLabel` comment, this goes away for free; otherwise it is worth a quick check on a large pipeline before merging.

frontend/src/community/crm/components/atoms/StageLabel/StageLabel.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+  return (
+    <div className="flex min-w-0 max-w-full items-center gap-2">
+      <div
+        className="size-2 rounded-full shrink-0"
+        style={{ backgroundColor: STAGE_COLOR_MAP[color] }}
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`color` is typed as `CrmDealStageColorsEnum`, but `STAGE_COLOR_MAP` is declared as `Record<string, string>` (constants/stageConstants.ts:1), so TypeScript cannot verify the map covers every enum member and an unmapped value silently produces `backgroundColor: undefined` — an invisible dot with no fallback. Now that lookup is centralized in one atom, tighten it: declare the constant as `Record<CrmDealStageColorsEnum, string>` (which makes any future enum addition a compile error) and/or add a fallback such as `STAGE_COLOR_MAP[color] ?? "var(--color-secondary-accent)"`. Note `crmUtil.ts:133` derives `dealStageColors` from `Object.entries(STAGE_COLOR_MAP)`, so the stricter type keeps that in sync too.

frontend/src/community/crm/components/organisms/AddDealSidePanel/DealNameStageSection.tsx
```diff
@@ -64,15 +64,7 @@ const DealNameStageSection: FC<DealNameStageSectionProps> = ({
...
-        label: (
-          <div className="inline-flex items-center gap-2.5">
-            <div
-              className="size-2 rounded-full shrink-0"
-              style={{ backgroundColor: STAGE_COLOR_MAP[s.color] }}
-            />
-            <span className="body2">{s.name}</span>
-          </div>
-        )
+        label: <StageLabel name={s.name} color={s.color} />
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Small unintended visual delta in the two dropdowns: the replaced markup used `inline-flex ... gap-2.5`, while `StageLabel` renders `flex ... gap-2` with `max-w-full`. Going from `inline-flex` to a block-level `flex` makes the option/trigger label stretch to the full container width instead of shrink-wrapping, which can shift alignment relative to the dropdown chevron, and the dot/text gap tightens from 0.625rem to 0.5rem. The table cell wanted the truncating full-width variant, but the dropdowns did not. Worth an eyeball in both the add panel and the edit sidebar; if the dropdowns need the old behavior, expose a `className` (or an `inline` flag) on `StageLabel` rather than reverting.

frontend/src/community/crm/components/atoms/StageLabel/StageLabel.tsx
```diff
@@ -0,0 +1,30 @@
+import { FC } from "react";
```
ThinuwanW, line 1: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

Two follow-ups on the extraction: (1) `DealAccordionItemBadge.tsx:13-24` still hand-rolls the same pairing (`useStageNameMapper` + `STAGE_COLOR_MAP` dot) as a `Chip` prefix and was not migrated — either reuse `StageLabel` there or note why the Chip variant must stay separate, otherwise the duplication this PR set out to remove partially survives. (2) No `StageLabel.test.tsx` was added. The crm module currently has zero test files, so this is not out of line with its neighbours, but this atom has real branching worth pinning (default-stage key gets translated, custom stage name passes through unchanged, dot color resolved from the map) and `~community/common/components/atoms/*` has an established pattern to copy.

ThinuwanW, line 1: 🤖 **Claude** · 🔴 **Important**

StageLabel bakes in both its typography (`body2`) and its layout box (`flex ... gap-2`) with no `min-w-0`, no `truncate`, no `title`, and no className passthrough — but it is being dropped into three places that previously relied on the *parent* to constrain the text.

Concrete regression in DealStageLaneHeader: the lane title lives in `<h2 className="subtitle1 truncate capitalize">` inside a fixed-width `w-75` lane with a `min-w-0` wrapper. `truncate` is `overflow:hidden; text-overflow:ellipsis; white-space:nowrap` — it only ellipsizes *inline text content*. Once the h2's only child is a `display:flex` box, ellipsis no longer applies and a long custom stage name (stage names are user-created via configurations > deal stages, only the 7 DefaultStageNameEnum values are fixed) will clip hard or blow out the lane instead of showing '…'. The `subtitle1` class on the h2 is also overridden by StageLabel's own `body2`, so the kanban lane heading silently drops a typography step.

Same exposure in DealsTable: the stage column is pinned at `minWidth: 140` and every other cell in that table guards overflow (`block w-full truncate` + `title={...}`); the stage cell now has neither.

Fix: inside StageLabel render the text as `<span className="min-w-0 truncate" title={label}>` and put `min-w-0` on the flex wrapper, and either accept an optional `className`/`labelClassName` prop or drop the hardcoded `body2` so callers like the `<h2>` keep their own type scale.

frontend/src/community/crm/components/molecules/DealStageLane/DealStageLaneHeader/DealStageLaneHeader.tsx
```diff
@@ -45,7 +45,11 @@ const DealStageLaneHeader: FC<DealStageLaneHeaderProps> = ({
 
       <div className="flex items-center justify-between gap-2 px-3 pt-3">
         <div className="min-w-0">
-          <h2 id={String(stage.id)} className="subtitle1 truncate capitalize">
+          <h2
```
ThinuwanW, line 50: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Dead code left behind by the h2 rewrite. The diff for this file is a single hunk covering lines 48-52 only — nothing above line 45 was touched — so `import useStageNameMapper from "~community/crm/hooks/useStageNameMapper";` (line 5) and `const { getStageByName } = useStageNameMapper();` (line 29) are still present. If the rewritten h2 now renders `<StageLabel>` (which resolves the name internally), `getStageByName` has no remaining reference in this file: `unused-imports/no-unused-vars` will warn on the destructured binding, and you are paying for an extra `useTranslation()` subscription per lane for nothing. Remove both lines. (If instead the h2 still calls `getStageByName` and only the dot markup was inlined by hand, then this file is the one stage-name site that was *not* migrated to the new atom — please make it consistent either way.)

ThinuwanW, line 50: 🤖 **Claude** · 🟡 **Suggestion**

Duplicate stage-colour indicator. This lane already renders the stage colour as a full-width bar at lines 41-44 (`style={{ backgroundColor: STAGE_COLOR_MAP[stage.color] }}`). Putting StageLabel in the heading adds a second swatch — a coloured dot — for the same datum, roughly 20px below the bar. That is redundant visually and adds a decorative element inside an `<h2>` that is also the `aria-labelledby` target for the section. Consider either dropping the bar, or giving StageLabel a `showDot`/`hideIndicator` prop (default true) and passing false here so the lane header reuses only the name-resolution half of the atom.

frontend/src/community/crm/components/atoms/StageLabel/StageLabel.tsx
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Stale-translation risk, now systemic. StageLabel resolves its text through `useStageNameMapper`, whose `getStageByName` is wrapped in `useCallback(..., [])` (hooks/useStageNameMapper.ts:13-19) while closing over `translateText`. `useTranslator` (common/hooks/useTranslator.tsx) returns a brand-new closure over `t` on every render and is not memoized, so the empty dep array permanently pins the very first `t` captured on mount. Before this PR that staleness was spread across a handful of call sites; StageLabel now makes it the single funnel for every stage name in the table, the kanban, and both stage dropdowns (via useGetMappedDealStages), so a language switch can leave the whole deals surface showing the previous locale's stage names. Fix at the source: either memoize the function returned by `useTranslator` with `useMemo`/`useCallback` keyed on `t`, or have `useStageNameMapper` depend on `t` directly (`const { t } = useTranslation(); const getStageByName = useCallback(..., [t])`). Note that naively adding `translateText` to the existing dep array will not work — it changes identity every render and would destroy the memoization that DealsTable's `tableRows` relies on.

frontend/src/community/crm/components/atoms/StageLabel/StageLabel.tsx
ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

No `StageLabel.test.tsx` accompanies the new atom. The review guidelines require a colocated test for new components, and this one has branching worth pinning: an unknown `color` key falls through `STAGE_COLOR_MAP` to `undefined` (transparent dot), a `name` matching `DefaultStageNameEnum` renders the translated string, and an arbitrary custom stage name renders verbatim. Worth adding a small test with the `MockTheme` wrapper and `screen.getByText` asserting those three cases — it would also lock in the fix for the truncation/typography issue above. (Context: the crm module currently has util tests only, no component tests, so this is a convention gap rather than a break in local practice.)

frontend/src/community/crm/components/organisms/DealsTable/DealsTable.tsx
```diff
@@ -203,7 +191,7 @@ const DealsTable: FC<Props> = ({
           )
         };
       }),
-    [allDeals, getStageByName]
+    [allDeals]
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

The `tableRows` dependency array was just edited to drop `getStageByName`, but it still omits `translateText`, which the callback uses for the row's `aria-label` (`translateText(["openDealDetails"], { name: deal.name })`). `react-hooks/exhaustive-deps` is enabled as `warn` in .eslintrc.json, so this will surface in lint. Practically: because the memo returns the identical element references, React bails out of reconciling those rows, so on a language change the row aria-labels keep the old locale even though StageLabel (which subscribes to i18n itself) updates. Adding `translateText` directly would defeat the memo since `useTranslator` returns a new function every render — the clean fix is the `useTranslator` memoization described in the StageLabel finding, after which `translateText` can be listed here honestly. `onDealClick` is also missing but is benign today: `handleDealOnClick` in DealsSection.tsx:62 only closes over stable Zustand setters.

frontend/src/community/crm/components/molecules/DealStageLane/DealStageLaneHeader/DealStageLaneHeader.tsx
YasinduRC: Fixed  the `<section>` was pointing at crm-stage-3 but the `<h2> `had id="3", so they never matched and the lane had no name for screen readers. Changed the heading to id={crm-stage-${stage.id}}.

frontend/src/community/crm/components/atoms/StageLabel/StageLabel.tsx
YasinduRC: Currently we are not writing tests for atoms

frontend/src/community/crm/components/organisms/DealsTable/DealsTable.tsx
YasinduRC: getStageByName was added since the now the parent translate the text and passed in to the component

frontend/src/community/crm/components/atoms/StageLabel/StageLabel.tsx
YasinduRC: 1)  Since the badge couldn't use StageLabel because Chip's label prop is typed string, not ReactNode  the Chip needs the text and the icon as two separate things. extracted the dot into its own StageDot atom, one level below StageLabel.
2)Tests skipped (we don't test atoms),

YasinduRC: * `StageLabel` already includes `min-w-0`, `truncate`, and `title`, so the overflow handling is already in place.
* `DealStageLaneHeader` wasn't migrated to use `StageLabel`, so the typography/ellipsis concern doesn't apply there.
* `DealsTable` still gets truncation and `title` via `StageLabel`; they're just provided internally now.
* A `className`/`labelClassName` prop could be added later if needed, but none of the current call sites require it, so I'd prefer to keep the API minimal for now.

frontend/src/community/crm/components/atoms/StageLabel/StageLabel.tsx
YasinduRC: The dropdowns were using dealStages while StageLabel was translating again, so the mapping happened twice. I've made StageLabel presentational, moved the mapping back to the callers, and renamed the prop to label to make it clear that it expects display text.

YasinduRC: Agreed, this is a real issue, but the proper fix is in useTranslator rather than here. Since useTranslator is a shared hook used across the app, making it stable is a broader behavioral change

frontend/src/community/crm/components/atoms/StageLabel/StageLabel.tsx
YasinduRC: updated STAGE_COLOR_MAP to Record<CrmDealStageColorsEnum, string>.

frontend/src/community/crm/components/molecules/DealStageLane/DealStageLaneHeader/DealStageLaneHeader.tsx
YasinduRC: This component wasn't migrated intentionally because it's rendering a different visual pattern. It still shares the stage-name translation logic (getStageByName); only the color rendering is different because this uses a lane bar instead of an inline dot.

YasinduRC: StageLabel isn't actually used in the lane header,the h2 still renders the stage name as plain text, so there's no duplicate color indicator here.

frontend/src/community/crm/components/organisms/DealSidePanel/DealPropertiesSidebar.tsx
YasinduRC: -The contact selector doesn't expose a clear action, so users can't clear the value. Because of that, the second concern doesn't apply here.
This sidebar is an inline-edit dropdown rather than a validated form. 

-Changes are applied immediately, so there's no validation cycle or error state to drive aria-invalid or an error message as there is in the Formik-based create/edit flow.

frontend/src/community/crm/components/organisms/AddDealSidePanel/DealNameStageSection.tsx
YasinduRC: The inline-flex → flex change doesn't affect rendering here, and the gap change is intentional to keep StageLabel styling consistent across all usages.

frontend/src/community/crm/components/organisms/DealsTable/DealsTable.tsx
YasinduRC: I updated StageLabel to be a presentational component, so it no longer owns the translation hook.

shakila-kamalasena (dismissed):

frontend/src/community/crm/components/organisms/DealsTable/DealsTable.tsx
```diff
@@ -171,15 +172,10 @@ const DealsTable: FC<Props> = ({
...
-              <div
-                className="size-2 rounded-full shrink-0"
-                style={{
-                  backgroundColor: STAGE_COLOR_MAP[deal.stage.color]
-                }}
-              />
-              <span className="body2">{getStageByName(deal.stage.name)}</span>
-            </div>
+            <StageLabel
+              label={getStageByName(deal.stage.name)}
```
akilarootcode: add optional chaining

frontend/src/community/crm/components/organisms/DealSidePanel/DealPropertiesSidebar.tsx
```diff
@@ -81,15 +81,7 @@ const DealPropertiesSidebar: FC<DealPropertiesSidebarProps> = ({
...
-        label: (
-          <div className="inline-flex items-center gap-2.5">
-            <div
-              className="size-2 rounded-full shrink-0"
-              style={{ backgroundColor: STAGE_COLOR_MAP[stage.color] }}
-            />
-            <span className="body2">{stage.name}</span>
-          </div>
-        )
+        label: <StageLabel label={stage.name} color={stage.color} />
```
akilarootcode: optional chaining

akilarootcode (changes requested):

frontend/src/community/crm/components/organisms/DealSidePanel/DealPropertiesSidebar.tsx
YasinduRC: added optional chaning

frontend/src/community/crm/components/organisms/DealsTable/DealsTable.tsx
YasinduRC: added optional chaning

shakila-kamalasena (dismissed):

akilarootcode (dismissed):

akilarootcode (approved):

PR #1703, opened 2026-08-03
Fix/manual time entry email time mismatch

SanithuGunasena07:
Bug: The time shown in the email body for a manual time entry request does not match the actual requested time, in both the supervisor's approval email and the requester's own email.

Root cause: The Time Entry Requests and Requests Awaiting Approval tables display times converted to the organization's configured time zone. However, the epochMillisToAmPmString() method in DateTimeUtils.java which is used when generating manual time entry request emails has the time zone hardcoded to UTC instead of using the organization's configured time zone. As a result, the time shown in the email body does not match the actual requested time or the time shown in the timesheet tables as the email have the UTC time.

### DIFF
--- a/backend/src/main/java/com/skapp/community/common/service/OrganizationService.java
+++ b/backend/src/main/java/com/skapp/community/common/service/OrganizationService.java
@@ -24,6 +24,6 @@ public interface OrganizationService {
 
 	String getOrganizationTimeZone();
 
-	ZoneId getOrganizationTimeZoneInZoneId();
+	ZoneId getOrganizationZoneId();
 
 }
--- a/backend/src/main/java/com/skapp/community/common/service/OrganizationService.java
+++ b/backend/src/main/java/com/skapp/community/common/service/OrganizationService.java
@@ -6,6 +6,8 @@ import com.skapp.community.common.payload.request.UpdateOrganizationRequestDto;
 import com.skapp.community.common.payload.response.EmailServerConfigResponseDto;
 import com.skapp.community.common.payload.response.ResponseEntityDto;
 
+import java.time.ZoneId;
+
 public interface OrganizationService {
 
 	ResponseEntityDto saveOrganization(OrganizationDto organizationDto);
@@ -21,5 +23,7 @@ public interface OrganizationService {
 	ResponseEntityDto updateOrganization(UpdateOrganizationRequestDto organizationDto);
 
 	String getOrganizationTimeZone();
+
+	ZoneId getOrganizationTimeZoneInZoneId();
 
 }
--- a/backend/src/main/java/com/skapp/community/common/service/impl/OrganizationServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/common/service/impl/OrganizationServiceImpl.java
@@ -36,6 +36,7 @@ import tools.jackson.databind.node.ArrayNode;
 import tools.jackson.databind.node.ObjectNode;
 
 import java.time.DayOfWeek;
+import java.time.ZoneId;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
@@ -225,6 +226,11 @@ public class OrganizationServiceImpl implements OrganizationService {
 		return organizationDao.findTopByOrderByOrganizationIdDesc()
 			.map(Organization::getOrganizationTimeZone)
 			.orElse("UTC");
+	}
+
+	@Override
+	public ZoneId getOrganizationTimeZoneInZoneId() {
+		return ZoneId.of(getOrganizationTimeZone());
 	}
 
 	public void getDefaultTimeConfigs() {
--- a/backend/src/main/java/com/skapp/community/common/service/impl/OrganizationServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/common/service/impl/OrganizationServiceImpl.java
@@ -19,6 +19,7 @@ import com.skapp.community.common.service.EncryptionDecryptionService;
 import com.skapp.community.common.service.OrganizationService;
 import com.skapp.community.common.service.UserService;
 import com.skapp.community.common.type.OrganizationConfigType;
+import com.skapp.community.common.util.DateTimeUtils;
 import com.skapp.community.common.util.MessageUtil;
 import com.skapp.community.crmplanner.service.CrmConfigService;
 import com.skapp.community.leaveplanner.service.LeaveCycleService;
@@ -229,8 +230,8 @@ public class OrganizationServiceImpl implements OrganizationService {
 	}
 
 	@Override
-	public ZoneId getOrganizationTimeZoneInZoneId() {
-		return ZoneId.of(getOrganizationTimeZone());
+	public ZoneId getOrganizationZoneId() {
+		return DateTimeUtils.resolveZoneId(getOrganizationTimeZone());
 	}
 
 	public void getDefaultTimeConfigs() {
--- a/backend/src/main/java/com/skapp/community/common/util/DateTimeUtils.java
+++ b/backend/src/main/java/com/skapp/community/common/util/DateTimeUtils.java
@@ -2,10 +2,12 @@ package com.skapp.community.common.util;
 
 import com.skapp.community.common.constant.CommonMessageConstant;
 import com.skapp.community.common.exception.ModuleException;
+import lombok.extern.slf4j.Slf4j;
 
 import java.text.DateFormat;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
+import java.time.DateTimeException;
 import java.time.DayOfWeek;
 import java.time.Instant;
 import java.time.LocalDate;
@@ -26,8 +28,10 @@ import java.util.Set;
 import static org.aspectj.bridge.Version.SIMPLE_DATE_FORMAT;
 
 /**
- * Utility class for handling UTC date and time operations.
+ * Utility class for handling date and time operations, including both UTC-based and
+ * zone-parameterised formatting/conversion.
  */
+@Slf4j
 public class DateTimeUtils {
 
 	public static final int JANUARY = 1;
@@ -455,6 +459,25 @@ public class DateTimeUtils {
 		}
 		Set<String> validIDs = ZoneId.getAvailableZoneIds();
 		return validIDs.contains(timeZone);
+	}
+
+	/**
+	 * Resolves a time zone string to a ZoneId, falling back to UTC if the value is
+	 * null, blank, or not a valid zone ID.
+	 * @param timezone The time zone ID string to resolve.
+	 * @return The resolved ZoneId, or UTC if the input is invalid.
+	 */
+	public static ZoneId resolveZoneId(String timezone) {
+		if (timezone == null) {
+			return UTC_ZONE_ID;
+		}
+		try {
+			return ZoneId.of(timezone);
+		}
+		catch (DateTimeException e) {
+			log.warn("resolveZoneId: Invalid time zone '{}', falling back to UTC", timezone);
+			return UTC_ZONE_ID;
+		}
 	}
 
 	/**
--- a/backend/src/main/java/com/skapp/community/common/util/DateTimeUtils.java
+++ b/backend/src/main/java/com/skapp/community/common/util/DateTimeUtils.java
@@ -627,11 +627,7 @@ public class DateTimeUtils {
 	}
 
 	public static String epochMillisToAmPmString(Long epochMillis) {
-		if (epochMillis == null) {
-			throw new ModuleException(CommonMessageConstant.COMMON_ERROR_EPOCH_MILLIS_CANNOT_BE_NULL);
-		}
-		LocalTime time = Instant.ofEpochMilli(epochMillis).atZone(ZoneOffset.UTC).toLocalTime();
-		return time.format(AM_PM_FORMATTER);
+		return epochMillisToAmPmString(epochMillis, UTC_ZONE_ID);
 	}
 
 	public static String epochMillisToAmPmString(Long epochMillis, ZoneId zoneId) {
--- a/backend/src/main/java/com/skapp/community/common/util/DateTimeUtils.java
+++ b/backend/src/main/java/com/skapp/community/common/util/DateTimeUtils.java
@@ -634,6 +634,14 @@ public class DateTimeUtils {
 		return time.format(AM_PM_FORMATTER);
 	}
 
+	public static String epochMillisToAmPmString(Long epochMillis, ZoneId zoneId) {
+		if (epochMillis == null) {
+			throw new ModuleException(CommonMessageConstant.COMMON_ERROR_EPOCH_MILLIS_CANNOT_BE_NULL);
+		}
+		LocalTime time = Instant.ofEpochMilli(epochMillis).atZone(zoneId).toLocalTime();
+		return time.format(AM_PM_FORMATTER);
+	}
+
 	public static String formatDateWithSuffix(LocalDate date) {
 		int dayOfMonth = date.getDayOfMonth();
 		DateTimeFormatter monthFormatter = DateTimeFormatter.ofPattern("MMM", Locale.ENGLISH);
--- a/backend/src/main/java/com/skapp/community/common/util/DateTimeUtils.java
+++ b/backend/src/main/java/com/skapp/community/common/util/DateTimeUtils.java
@@ -28,10 +28,8 @@ import java.util.Set;
 import static org.aspectj.bridge.Version.SIMPLE_DATE_FORMAT;
 
 /**
- * Utility class for handling date and time operations, including both UTC-based and
- * zone-parameterised formatting/conversion.
+ * Utility class for handling UTC date and time operations.
  */
-@Slf4j
 public class DateTimeUtils {
 
 	public static final int JANUARY = 1;
@@ -462,8 +460,8 @@ public class DateTimeUtils {
 	}
 
 	/**
-	 * Resolves a time zone string to a ZoneId, falling back to UTC if the value is
-	 * null, blank, or not a valid zone ID.
+	 * Resolves a time zone string to a ZoneId, falling back to UTC if the value is null,
+	 * blank, or not a valid zone ID.
 	 * @param timezone The time zone ID string to resolve.
 	 * @return The resolved ZoneId, or UTC if the input is invalid.
 	 */
@@ -475,7 +473,6 @@ public class DateTimeUtils {
 			return ZoneId.of(timezone);
 		}
 		catch (DateTimeException e) {
-			log.warn("resolveZoneId: Invalid time zone '{}', falling back to UTC", timezone);
 			return UTC_ZONE_ID;
 		}
 	}
--- a/backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeEmailServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeEmailServiceImpl.java
@@ -316,8 +316,9 @@ public class TimeEmailServiceImpl implements TimeEmailService {
 
 	private void setTimeEntryFields(AttendanceEmailDynamicFields fields, TimeRequest request) {
 		ZoneId zoneId = organizationService.getOrganizationZoneId();
-		fields.setTimeEntryDate(
-				DateTimeUtils.epochMillisToUtcLocalDateTime(request.getRequestedStartTime(), zoneId).toLocalDate().toString());
+		fields.setTimeEntryDate(DateTimeUtils.epochMillisToUtcLocalDateTime(request.getRequestedStartTime(), zoneId)
+			.toLocalDate()
+			.toString());
 		fields.setStartTime(DateTimeUtils.epochMillisToAmPmString(request.getRequestedStartTime(), zoneId));
 		fields.setEndTime(DateTimeUtils.epochMillisToAmPmString(request.getRequestedEndTime(), zoneId));
 	}
--- a/backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeEmailServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeEmailServiceImpl.java
@@ -172,15 +172,10 @@ public class TimeEmailServiceImpl implements TimeEmailService {
 	@Override
 	public void sendTimeEntryRequestSubmittedEmployeeEmail(TimeRequest timeRequest) {
 		AttendanceEmailDynamicFields emailDynamicFields = new AttendanceEmailDynamicFields();
-		ZoneId zoneId = organizationService.getOrganizationTimeZoneInZoneId();
 
 		emailDynamicFields.setEmployeeOrManagerName(
 				timeRequest.getEmployee().getFirstName() + " " + timeRequest.getEmployee().getLastName());
-		emailDynamicFields
-			.setTimeEntryDate(DateTimeUtils.epochMillisToUtcLocalDate(timeRequest.getRequestedStartTime()).toString());
-		emailDynamicFields
-			.setStartTime(DateTimeUtils.epochMillisToAmPmString(timeRequest.getRequestedStartTime(), zoneId));
-		emailDynamicFields.setEndTime(DateTimeUtils.epochMillisToAmPmString(timeRequest.getRequestedEndTime(), zoneId));
+		setTimeEntryFields(emailDynamicFields, timeRequest);
 
 		User user = timeRequest.getEmployee().getUser();
 
@@ -191,15 +186,10 @@ public class TimeEmailServiceImpl implements TimeEmailService {
 	@Override
 	public void sendReceivedTimeEntryRequestManagerEmail(TimeRequest timeRequest) {
 		AttendanceEmailDynamicFields emailDynamicFields = new AttendanceEmailDynamicFields();
-		ZoneId zoneId = organizationService.getOrganizationTimeZoneInZoneId();
 
 		emailDynamicFields
 			.setEmployeeName(timeRequest.getEmployee().getFirstName() + " " + timeRequest.getEmployee().getLastName());
-		emailDynamicFields
-			.setTimeEntryDate(DateTimeUtils.epochMillisToUtcLocalDate(timeRequest.getRequestedStartTime()).toString());
-		emailDynamicFields
-			.setStartTime(DateTimeUtils.epochMillisToAmPmString(timeRequest.getRequestedStartTime(), zoneId));
-		emailDynamicFields.setEndTime(DateTimeUtils.epochMillisToAmPmString(timeRequest.getRequestedEndTime(), zoneId));
+		setTimeEntryFields(emailDynamicFields, timeRequest);
 
 		List<EmployeeManager> managers = employeeManagerDao.findByEmployee(timeRequest.getEmployee());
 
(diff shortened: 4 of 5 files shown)


ThinuwanW (commented):
## 🤖 Claude Code Review

The PR correctly identifies that time-entry emails were rendering UTC clock times and threads an organization ZoneId through `epochMillisToAmPmString`; since the frontend stores `requestedStartTime/EndTime` as true UTC instants (`convertToUtc` in TimeUtils.ts), converting to the org zone for display is the right fix. Two problems remain: the accompanying `timeEntryDate` in the same emails is still computed in UTC, so the date and the times can disagree by a day, and the new `ZoneId.of(getOrganizationTimeZone())` is unguarded against the unvalidated, free-form `organization_time_zone` column, which turns a bad/empty value into an unchecked DateTimeException that rolls back the enclosing transactional operation. Smaller items: the 3-line date/time block is now duplicated 10x with a per-call DB lookup, the new overload lacks the null-zone fallback its sibling has, and the retained UTC-default overload leaves in-app timesheet times inconsistent with emails.

Found **14** new issue(s): 🔴 4 important, 🟡 6 suggestion(s), 🟣 4 nit(s)

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeEmailServiceImpl.java
```diff
@@ -168,13 +172,15 @@ public void sendNonWorkingDayMultiDayApprovedLeaveRequestRevokedManagerEmail(Lea
 	@Override
 	public void sendTimeEntryRequestSubmittedEmployeeEmail(TimeRequest timeRequest) {
 		AttendanceEmailDynamicFields emailDynamicFields = new AttendanceEmailDynamicFields();
+		ZoneId zoneId = organizationService.getOrganizationTimeZoneInZoneId();
 
 		emailDynamicFields.setEmployeeOrManagerName(
 				timeRequest.getEmployee().getFirstName() + " " + timeRequest.getEmployee().getLastName());
 		emailDynamicFields
 			.setTimeEntryDate(DateTimeUtils.epochMillisToUtcLocalDate(timeRequest.getRequestedStartTime()).toString());
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Incomplete fix: start/end times are now rendered in the organization time zone, but `timeEntryDate` in the same email still uses `DateTimeUtils.epochMillisToUtcLocalDate(...)` (UTC). The email template renders them together (`Date: {{timeEntryDate}}`, `From: {{startTime}}`, `To: {{endTime}}` in attendance-module-time-entry-request-submitted-employee.html), so the date and the times can refer to different days. Example: org zone America/New_York, employee logs 8:00 PM–9:00 PM on Aug 4 → stored UTC is Aug 5 00:00–01:00 → email now reads "Date: 2026-08-05, From: 08:00 PM, To: 09:00 PM". Same class of error in the other direction for Asia/Colombo entries after 6:30 PM UTC. Fix: derive the date in the same zone, e.g. `DateTimeUtils.epochMillisToUtcLocalDateTime(timeRequest.getRequestedStartTime(), zoneId).toLocalDate().toString()` (or add an `epochMillisToLocalDate(Long, ZoneId)` overload alongside the new AM/PM one), and apply it to all 10 methods changed here. Note AttendanceNotificationServiceImpl sends the in-app notification for the same events with a UTC `timeEntryDate` too, so that file needs the same treatment to keep email and notification consistent.

backend/src/main/java/com/skapp/community/common/service/impl/OrganizationServiceImpl.java
```diff
@@ -227,6 +228,11 @@ public String getOrganizationTimeZone() {
 			.orElse("UTC");
 	}
 
+	@Override
+	public ZoneId getOrganizationTimeZoneInZoneId() {
+		return ZoneId.of(getOrganizationTimeZone());
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`ZoneId.of(getOrganizationTimeZone())` is unguarded against invalid input and throws unchecked `DateTimeException`/`ZoneRulesException` (not `ModuleException`), so it bypasses the message-constant/ErrorResponse convention and surfaces as a 500. The source value is not trustworthy: `Organization.organizationTimeZone` is a nullable free-form `varchar(100)` with no bean validation on the way in (`OrganizationDto.organizationTimeZone` has no `@NotNull`/zone-id check, and `setup-organization.tsx` initialises it to ""), and `getOrganizationTimeZone()` only falls back to "UTC" when there is no organization row — an empty string or a value like "IST" is returned as-is and blows up here. Because these email methods run inside `@Transactional` service methods (e.g. `TimeServiceImpl.addManualEntryRequest` line 626 calls `sendTimeEntryRequestSubmittedEmployeeEmail`), a single bad config value would fail the entire time-entry request, not just the email — a regression versus the previous UTC-constant path, which could not throw. The codebase already has the safe pattern: `EpDateTimeUtils.resolveZoneId` (backend/src/main/java/com/skapp/enterprise/common/util/EpDateTimeUtils.java:41) catches `DateTimeException` and returns `ZoneOffset.UTC`. Move that helper into the community `DateTimeUtils` (having `EpDateTimeUtils.resolveZoneId` delegate to it, to avoid two implementations) and call it here, ideally with a `log.warn` recording the rejected value. Separately, adding a zone-id validation check when the organization is saved would fix the root cause.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeEmailServiceImpl.java
```diff
@@ -168,13 +172,15 @@ public void sendNonWorkingDayMultiDayApprovedLeaveRequestRevokedManagerEmail(Lea
 	@Override
 	public void sendTimeEntryRequestSubmittedEmployeeEmail(TimeRequest timeRequest) {
 		AttendanceEmailDynamicFields emailDynamicFields = new AttendanceEmailDynamicFields();
+		ZoneId zoneId = organizationService.getOrganizationTimeZoneInZoneId();
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The identical block — resolve the org zone, then set `timeEntryDate`/`startTime`/`endTime` from a `TimeRequest` — is now repeated in 10 methods, and this PR duplicated the zone lookup line 10 more times. Extract a private helper such as `private void setTimeEntryFields(AttendanceEmailDynamicFields fields, TimeRequest request)` that resolves the zone once and sets all three fields; each call site then becomes one line, and the date/zone fix from the finding above only has to be made in one place. This also matters for cost: `getOrganizationTimeZoneInZoneId()` executes `organizationDao.findTopByOrderByOrganizationIdDesc()` on every invocation, and community has no caching (only the enterprise `EpOrganizationServiceImpl.getOrganizationTimeZone()` override caches), so a flow like `addManualEntryRequest` — which sends employee + manager (+ auto-approval) emails — now issues one extra organization query per email method.

backend/src/main/java/com/skapp/community/common/util/DateTimeUtils.java
```diff
@@ -627,10 +627,14 @@ public static DayOfWeek getDayOfWeek(LocalDate date) {
...
 	public static String epochMillisToAmPmString(Long epochMillis) {
+		return epochMillisToAmPmString(epochMillis, UTC_ZONE_ID);
+	}
+
+	public static String epochMillisToAmPmString(Long epochMillis, ZoneId zoneId) {
 		if (epochMillis == null) {
 			throw new ModuleException(CommonMessageConstant.COMMON_ERROR_EPOCH_MILLIS_CANNOT_BE_NULL);
 		}
-		LocalTime time = Instant.ofEpochMilli(epochMillis).atZone(ZoneOffset.UTC).toLocalTime();
+		LocalTime time = Instant.ofEpochMilli(epochMillis).atZone(zoneId).toLocalTime();
```
ThinuwanW, line 657: 🤖 **Claude** · 🟡 **Suggestion**

The new overload has no null guard on `zoneId`, so a null zone yields a raw `NullPointerException` from `Instant.atZone(null)` instead of a `ModuleException`. This is inconsistent with the sibling zone-aware method in the same class, `epochMillisToUtcLocalDateTime(Long, ZoneId)` (line 499), which explicitly falls back to `ZoneOffset.UTC` when `zoneId == null`. Mirror that behaviour (fall back to `UTC_ZONE_ID`) or throw a `ModuleException` with a message constant. While here, add the Javadoc block that the neighbouring public methods in this class all carry, documenting the zone parameter and the null-epoch `ModuleException`.

backend/src/main/java/com/skapp/community/common/util/DateTimeUtils.java
```diff
@@ -627,10 +627,14 @@ public static DayOfWeek getDayOfWeek(LocalDate date) {
 	}
 
 	public static String epochMillisToAmPmString(Long epochMillis) {
+		return epochMillisToAmPmString(epochMillis, UTC_ZONE_ID);
```
ThinuwanW, line 650: 🤖 **Claude** · 🟡 **Suggestion**

Keeping the no-zone overload defaulting to UTC is a reasonable compatibility shim, but its one remaining production caller is a display path that arguably has the same bug this PR is fixing: `TimeAnalyticsServiceImpl` lines 317-319 format `clockInTime`/`clockOutTime` for the clock-in summary response with the UTC overload, so the in-app timesheet still shows UTC clock times while emails now show organization-local times. That same class already resolves the org zone at line 396 (`isLateArrival`) with its own raw `ZoneId.of(organizationService.getOrganizationTimeZone())`. Consider passing the org zone at lines 317-319 and switching line 396 to the new `getOrganizationTimeZoneInZoneId()` so there is a single, safe conversion point; if the analytics path is intentionally out of scope for this PR, say so in the description so the divergence is a known decision rather than an oversight.

backend/src/main/java/com/skapp/community/common/service/OrganizationService.java
```diff
@@ -22,4 +24,6 @@ public interface OrganizationService {
 
 	String getOrganizationTimeZone();
 
+	ZoneId getOrganizationTimeZoneInZoneId();
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`getOrganizationTimeZoneInZoneId()` reads awkwardly and encodes the return type in the name. `getOrganizationZoneId()` is clearer and still matches the `get*` verb convention; the return type already communicates "in ZoneId".

backend/src/main/java/com/skapp/community/common/service/impl/OrganizationServiceImpl.java
```diff
@@ -227,6 +228,11 @@ public String getOrganizationTimeZone() {
 			.orElse("UTC");
 	}
 
+	@Override
+	public ZoneId getOrganizationTimeZoneInZoneId() {
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Per the logging convention (`log.info("methodName: execution started")`), this new service method has no entry logging. Given it is now invoked on every attendance email, `log.info` here would be noisy — `log.debug` is probably the right level, and the more valuable log is a `log.warn` on the invalid-zone fallback described above. Flagging so the omission is a deliberate choice.

backend/src/main/java/com/skapp/community/common/util/DateTimeUtils.java
ThinuwanW, line 650: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

No tests accompany this behaviour change. There is no `DateTimeUtilsTest` and no unit test for `TimeEmailServiceImpl` in `backend/src/test`, so nothing pins the new zone-aware formatting. Worth adding a small unit test for `epochMillisToAmPmString(Long, ZoneId)` covering: a positive-offset zone (Asia/Colombo) where the AM/PM value differs from UTC, a negative-offset zone (America/New_York) crossing midnight, a DST-transition instant, the retained no-zone overload still returning UTC, and the null-epoch `ModuleException` path.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeEmailServiceImpl.java
```diff
@@ -168,13 +172,15 @@ public void sendNonWorkingDayMultiDayApprovedLeaveRequestRevokedManagerEmail(Lea
...
 
 		emailDynamicFields.setEmployeeOrManagerName(
 				timeRequest.getEmployee().getFirstName() + " " + timeRequest.getEmployee().getLastName());
 		emailDynamicFields
 			.setTimeEntryDate(DateTimeUtils.epochMillisToUtcLocalDate(timeRequest.getRequestedStartTime()).toString());
-		emailDynamicFields.setStartTime(DateTimeUtils.epochMillisToAmPmString(timeRequest.getRequestedStartTime()));
-		emailDynamicFields.setEndTime(DateTimeUtils.epochMillisToAmPmString(timeRequest.getRequestedEndTime()));
+		emailDynamicFields
+			.setStartTime(DateTimeUtils.epochMillisToAmPmString(timeRequest.getRequestedStartTime(), zoneId));
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Missing paired frontend change — the web UI and the email will now disagree on the same value. `requestedStartTime`/`requestedEndTime` are returned to the browser as raw epoch millis (`EmployeeTimeRequestResponseDto`, `ManagerTimeRequestResponseDto`), and the frontend preprocessor converts them with `convertUnixTimestampToISO` (`frontend/src/community/attendance/utils/TimeUtils.ts`), which is `DateTime.fromMillis(...)` — i.e. the **browser's** zone — then re-parses with `convertTo24HourByDateString` in `getCurrentTimeZone()` (`DateTime.local().zoneName`). So `EmployeeTimesheetRequestTable`/`ManagerTimesheetRequestTable` render these times in each viewer's local zone, while after this PR the email renders them in the organization zone. Nothing in the frontend reads the organization time zone for attendance display (it is only used in org-settings/signup screens). Concretely: org zone = Asia/Colombo, an employee viewing from Europe/London sees `03:30` in the timesheet request table but receives an email saying `09:00 AM` for the same request. Either update the frontend attendance display to convert with the organization time zone (fetch it alongside the request list and use `DateTime.fromMillis(ts, { zone: orgTimeZone })`), or return a pre-formatted org-zone string from the backend so both surfaces share one formatting decision. This PR should not ship with only the email half changed.

backend/src/main/java/com/skapp/community/common/service/impl/OrganizationServiceImpl.java
```diff
@@ -227,6 +228,11 @@ public String getOrganizationTimeZone() {
 			.orElse("UTC");
```
ThinuwanW, line 229: 🤖 **Claude** · 🔴 **Important**

The write side of `organizationTimeZone` is completely unvalidated, and once a bad value is stored there is no API to correct it — which turns the new zone lookup into a permanent failure. (a) `OrganizationDto.organizationTimeZone` (line 28) has no `@NotNull`, no `@Pattern`, not even a `@Schema` — unlike every other field in that DTO — and `saveOrganization` performs no check, so `POST /v1/organization` will happily persist `""`, `"Asia/Colombo "` (trailing space), `"IST"`, or any typo into `organization.organization_time_zone varchar(100)` (nullable). (b) `.orElse("UTC")` here only covers a missing organization row or a NULL column — an empty or malformed non-null string is returned verbatim to the caller. (c) `UpdateOrganizationRequestDto` does not contain `organizationTimeZone` and `setOrganizationDetails` never sets it, so the `PATCH /v1/organization` endpoint cannot fix it; the only remedy is a manual DB update. Net effect: one bad value at org setup permanently breaks every time-entry-request email for that tenant. Fix: validate at write time — add a `isValidTimeZone` helper to `com.skapp.community.common.util.Validation` backed by `ZoneId.getAvailableZoneIds().contains(value)`, reject in `saveOrganization` with a new `CommonMessageConstant` key plus its `*-messages.properties` entry, and add `organizationTimeZone` to `UpdateOrganizationRequestDto`/`setOrganizationDetails` so it is correctable. Also harden the read here to `filter(tz -> tz != null && !tz.isBlank())` before `.orElse("UTC")`.

backend/src/main/java/com/skapp/community/common/service/impl/OrganizationServiceImpl.java
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

A safe zone-resolution helper already exists and should have been consolidated here rather than adding a sixth raw `ZoneId.of()` call site. `EpDateTimeUtils.resolveZoneId(String)` (enterprise, lines 41-51) already implements exactly the contract this method needs — null → `ZoneOffset.UTC`, `DateTimeException` → `ZoneOffset.UTC` — and is used by `AdmsTimeRecordSyncServiceImpl`. Meanwhile raw `ZoneId.of(...)` on the org timezone is duplicated in `TimeAnalyticsServiceImpl:396`, `NotificationServiceImpl:372`, `CommonModuleUtils:46`, `EpLeaveInsightInternalServiceImpl:51`, `AnnouncementServiceImpl:63/85` and `EpProbationInternalServiceImpl:63`. Since this new method is intended to be the single accessor for the org zone, move `resolveZoneId`'s logic down into `com.skapp.community.common.util.DateTimeUtils` (community), delegate to it from here, and then migrate those six call sites to `organizationService.getOrganizationTimeZoneInZoneId()`. That deletes the duplication, removes the enterprise/community fork of the same helper, and gives one place where the invalid-zone fallback is defined.

backend/src/main/java/com/skapp/community/common/service/impl/OrganizationServiceImpl.java
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

This lookup hits the database on every call in community, whereas enterprise deliberately caches it — so the new per-email invocation is a community-only regression. `getOrganizationTimeZone()` here executes `organizationDao.findTopByOrderByOrganizationIdDesc()` with no caching and no `@Transactional(readOnly = true)`; `EpOrganizationServiceImpl.getOrganizationTimeZone()` (lines 426-437) wraps the same call in `EpCacheKeys.ORGANIZATION_TIMEZONE_CACHE_KEY`, which is evidence the team already considered this a hot read. After this PR every one of the ten email methods issues its own query, and a single approve action in `TimeServiceImpl` calls two of them (`sendTimeEntryRequestApprovedEmployeeEmail` + `sendTimeEntryRequestApprovedOtherManagerEmail`), so the auto-approval scheduler pays 2 extra queries per request processed. Also, `ZoneId.of()` is re-parsed each time. Either add the same caching to the community implementation, or resolve the zone once per business operation in `TimeServiceImpl` and pass the `ZoneId` into the email methods as a parameter (which would also collapse the ten duplicated lookup lines).

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeEmailServiceImpl.java
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Whatever zone decision is finally made for `timeEntryDate` must be applied to the in-app notification path in the same PR, or notifications and emails for the same event will state different dates. `AttendanceNotificationServiceImpl` renders `timeEntryDate` for the identical ten attendance events via `DateTimeUtils.epochMillisToUtcLocalDate(timeRequest.getRequestedStartTime())` at lines 37, 48, 63, 74, 85, 97, 112, 124, 260 and 276, and `community/templates/notification/notification-templates.json` interpolates `{{timeEntryDate}}` in 10 places. That service does not inject `OrganizationService` at all. So once the email date is moved to the org zone, an evening time entry near midnight will show one date in the user's notification bell and a different date in their inbox for the same request. Fix in lockstep: inject `OrganizationService` into `AttendanceNotificationServiceImpl` and derive the date with the same zone (ideally via a shared helper, since both classes populate the same `AttendanceEmailDynamicFields` DTO).

backend/src/main/java/com/skapp/community/common/util/DateTimeUtils.java
```diff
@@ -627,10 +627,14 @@ public static DayOfWeek getDayOfWeek(LocalDate date) {
...
+	}
+
+	public static String epochMillisToAmPmString(Long epochMillis, ZoneId zoneId) {
```
ThinuwanW, line 653: 🤖 **Claude** · 🟣 **Nit**

The new public overload is undocumented while its direct neighbours are, leaving callers no way to tell which of the two overloads is correct for their case. `epochMillisToUtcLocalDateTime(Long, ZoneId)` at lines 492-498, `getDateStringByTimeZone` at 476-481 and `epochMillisToUtcLocalTime` at 533-538 all carry `@param`/`@return`/`@throws` Javadoc; both `epochMillisToAmPmString` overloads have none. Given the whole point of this PR is that the wrong overload produces wrong times in customer-facing email, add Javadoc to the two-arg version stating that it formats in the supplied zone and throws `ModuleException` when `epochMillis` is null, and to the no-arg version stating that it formats in UTC and is retained for compatibility (so new callers are steered to the explicit-zone form). Related: the class-level comment at lines 28-30 still reads "Utility class for handling UTC date and time operations", which is no longer accurate now that the class exposes zone-parameterised formatting — worth updating in the same edit.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeEmailServiceImpl.java
SanithuGunasena07: fixed, timeEntryDate in both TimeEmailServiceImpl and AttendanceNotificationServiceImpl now derives the date using the same org zoneId, so date and time stay consistent.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeEmailServiceImpl.java
SanithuGunasena07: extracted a private setTimeEntryFields(fields, request) helper that resolves the zone and sets date/start/end time in one place. Each of the 10 methods now just calls it once.

backend/src/main/java/com/skapp/community/common/util/DateTimeUtils.java
SanithuGunasena07: now throws ModuleException(COMMON_ERROR_TIME_ZONE_CANNOT_BE_NULL) on a null zone

backend/src/main/java/com/skapp/community/common/util/DateTimeUtils.java
SanithuGunasena07: intentionally out of scope for this PR

backend/src/main/java/com/skapp/community/common/service/OrganizationService.java
SanithuGunasena07: renamed

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeEmailServiceImpl.java
SanithuGunasena07: not necessary

backend/src/main/java/com/skapp/community/common/util/DateTimeUtils.java
SanithuGunasena07: reverted as not needed

backend/src/main/java/com/skapp/community/common/service/impl/OrganizationServiceImpl.java
SanithuGunasena07: Added resolveZoneId(String) to community DateTimeUtils (catches DateTimeException, falls back to UTC, with log.warn on the rejected value), and getOrganizationZoneId() now uses it instead of raw ZoneId.of(...). Also removed EpDateTimeUtils.resolveZoneId and updated its 3 call sites to call the community version directly

backend/src/main/java/com/skapp/community/common/service/impl/OrganizationServiceImpl.java
SanithuGunasena07: resolved

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeEmailServiceImpl.java
SanithuGunasena07: fixed

backend/src/main/java/com/skapp/community/common/service/impl/OrganizationServiceImpl.java
SanithuGunasena07: Not part of the bug fix

backend/src/main/java/com/skapp/community/common/service/impl/OrganizationServiceImpl.java
SanithuGunasena07: done

backend/src/main/java/com/skapp/community/common/service/impl/OrganizationServiceImpl.java
SanithuGunasena07: Not part of the bug fix

backend/src/main/java/com/skapp/community/common/util/DateTimeUtils.java
SanithuGunasena07: not required

backend/src/main/java/com/skapp/community/common/util/DateTimeUtils.java
SanithuGunasena07: not required

backend/src/main/java/com/skapp/community/common/util/DateTimeUtils.java
```diff
@@ -26,8 +28,10 @@
 import static org.aspectj.bridge.Version.SIMPLE_DATE_FORMAT;
 
 /**
- * Utility class for handling UTC date and time operations.
```
yuhanga-rtc, line 29: Keep previous comments only.

backend/src/main/java/com/skapp/community/common/util/DateTimeUtils.java
```diff
@@ -26,8 +28,10 @@
...
+ * Utility class for handling date and time operations, including both UTC-based and
+ * zone-parameterised formatting/conversion.
  */
+@Slf4j
```
yuhanga-rtc: Remove logs

backend/src/main/java/com/skapp/community/common/util/DateTimeUtils.java
```diff
@@ -457,6 +461,25 @@ public static boolean isValidTimeZone(String timeZone) {
...
+	 */
+	public static ZoneId resolveZoneId(String timezone) {
+		if (timezone == null) {
+			return UTC_ZONE_ID;
+		}
+		try {
+			return ZoneId.of(timezone);
+		}
+		catch (DateTimeException e) {
+			log.warn("resolveZoneId: Invalid time zone '{}', falling back to UTC", timezone);
```
yuhanga-rtc: Remove

yuhanga-rtc (changes requested):

backend/src/main/java/com/skapp/community/common/util/DateTimeUtils.java
SanithuGunasena07: reverted

backend/src/main/java/com/skapp/community/common/util/DateTimeUtils.java
SanithuGunasena07: removed

backend/src/main/java/com/skapp/community/common/util/DateTimeUtils.java
SanithuGunasena07: removed

yuhanga-rtc (dismissed):

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeEmailServiceImpl.java
```diff
@@ -343,6 +314,15 @@ public void sendTimeEntryRequestDeclinedOtherManagerEmail(TimeRequest timeReques
 				EmailBodyTemplates.ATTENDANCE_MODULE_TIME_ENTRY_REQUEST_DECLINED_OTHER_MANAGER);
 	}
 
+	private void setTimeEntryFields(AttendanceEmailDynamicFields fields, TimeRequest request) {
```
akilarootcode: use proper meaningful name

akilarootcode (changes requested):

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeEmailServiceImpl.java
SanithuGunasena07: renamed to setTimeEntryDisplayFields

akilarootcode (approved):

PR #1704, opened 2026-08-03
Feature/birthday notification fe implementation

gaindurootcode:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds a birthday-notification modal system (global provider + queue hook + modals) plus a new People tab in Configurations. The util layer is well tested and the module structure follows conventions, but the orchestration has several correctness problems: the server's `lastViewedDate` is fetched yet never used to suppress the modal, a failed/slow mark-viewed PATCH causes the modal to re-appear on the next navigation, and the wrapper is mounted in BaseLayout for every page (including public/unauthenticated ones) where it fires an unconditional `/me` request. The config screen can also get stuck on its skeleton forever when the GET resolves to null, and one modal dereferences `firstName` without the null-guard used everywhere else.

Found **39** new issue(s): 🔴 9 important, 🟡 18 suggestion(s), 🟣 12 nit(s)

frontend/src/community/people/hooks/useBirthdayNotifications.ts
```diff
@@ -0,0 +1,167 @@
...
+  useEffect(() => {
+    if (!data) return;
+    if (seededDataRef.current === data) return;
+    if (!shouldEvaluate) return;
+    if (isShowingRef.current) return;
+    if (!isCurrentEmployeeResolved) return;
+
+    seededDataRef.current = data;
+
+    if (data.lastViewedDate) {
```
ThinuwanW, line 73: 🤖 **Claude** · 🔴 **Important**

The server's `lastViewedDate` is fetched but never used to suppress display. When it equals `today` (the user already dismissed today), the code only seeds the local cache via `markBirthdayViewedDate` and then falls through to build the queue and open the modals anyway — the `setViewedCache` state update lands after this effect body has already set `isShowingRef.current = true` and `setQueue(...)`, so the modal renders. This is reachable without multiple devices: `SpecialNotificationProvider` calls `clearViewedCache()` whenever `sessionStatus === "unauthenticated"`, so logging out and back in on the same day wipes the local record and the modal replays. Same for a second browser/device or cleared localStorage. Fix: after seeding the cache, bail out when the server already recorded today — e.g. `if (data.lastViewedDate === today) return;` before `buildBirthdayQueue`.

frontend/src/community/people/hooks/useBirthdayNotifications.ts
```diff
@@ -0,0 +1,167 @@
...
+    isCurrentEmployeeResolved,
+    markViewed,
+    markBirthdayViewedDate
+  ]);
+
+  const finishSequence = useCallback(() => {
+    isShowingRef.current = false;
+    setQueue([]);
+    setCursor(0);
+    markViewed();
```
ThinuwanW, line 110: 🤖 **Claude** · 🔴 **Important**

`finishSequence` clears `isShowingRef` and calls `markViewed()`, but the local viewed-cache is only written from the mutation's `onSuccess` (and only when the response contains a `lastViewedDate`). So if the PATCH fails, times out (the timeout is a tight 3000ms), or returns no `lastViewedDate`, `isViewedToday` stays false → `shouldEvaluate` stays true → the `evaluationTick` effect refetches on the next route change → a brand-new `data` object !== `seededDataRef.current` → the whole modal sequence replays. There is also a race even on success: the user can navigate within the PATCH round-trip and re-trigger the sequence. Fix: persist locally and optimistically at dismiss time — call `persistViewedDate(SpecialNotificationType.BIRTHDAY, today)` inside `finishSequence` before/independently of `markViewed()`, and add an `onError` to `useMarkBirthdayNotificationsViewedToday` (currently it has none) so the failure is at least not silently self-repeating.

frontend/src/community/common/components/templates/BaseLayout/BaseLayout.tsx
```diff
@@ -109,6 +111,9 @@ const BaseLayout = ({ children }: Props) => {
     <>
       {renderComponent}
       <CommonModalController />
+      <SpecialNotificationProvider>
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`BaseLayout` wraps every page via `_app.tsx`, including public routes (sign-in, sign-up, reset-password). `BirthdayNotificationWrapper` is rendered unconditionally, so `useBirthdayNotifications` runs there too — and it calls `useGetUserPersonalDetails()`, which has no `enabled` flag, so a `GET /people/me` fires on every public page for unauthenticated visitors (401). Note that this `/me` call is only a fallback: `currentEmployeeId` almost always resolves from `useSessionData().employeeDetails?.employeeId`. Fix: gate the mount (e.g. only render the provider/wrapper when `sessionStatus === "authenticated"`, and ideally behind `next/dynamic`), and/or add an `enabled` argument to `useGetUserPersonalDetails` so it only runs when the session employee id is missing.

frontend/src/community/people/components/molecules/ColleagueBirthdayModal/ColleagueBirthdayModal.tsx
```diff
@@ -0,0 +1,39 @@
...
+  onDismiss
+}) => {
+  const translateText = useTranslator("peopleModule", "birthdayNotifications");
+
+  return (
+    <BirthdayModalShell
+      id={MODAL_ID}
+      employee={employee}
+      heading={translateText(["colleague", "heading"], {
+        name: employee.firstName.trim()
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`employee.firstName.trim()` is dereferenced without a guard, while `getFullName` in birthdayNotificationUtils.ts deliberately uses `employee.firstName?.trim()` — i.e. the codebase does not trust this field to be non-null at runtime (compare `EmployeeDetails.firstName: string | null` in EmployeeTypes.tsx). An employee whose `firstName` is null but `lastName` is set passes the `hasDisplayableName` filter and reaches this line, throwing a TypeError that bubbles to the app-level ErrorBoundary (full-page error). Fix: use `employee.firstName?.trim()` (or reuse `getFullName`) here for consistency.

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
```diff
@@ -0,0 +1,205 @@
...
+        {subOptionsAriaMessage}
+      </div>
+
+      {isError && (
+        <p className="body1 text-secondary-text">
+          {translateText([SECTION, "loadError"])}
+        </p>
+      )}
+
+      {!isError && (isLoading || !config) && (
```
ThinuwanW, line 107: 🤖 **Claude** · 🔴 **Important**

`getBirthdayNotificationConfig` returns `result?.data?.results?.[0] ?? null`, so a successful-but-empty response (or a 401, which the response interceptor in TanStackProvider resolves with `undefined` rather than rejecting) yields `data === null`. Then `isLoading` is false, `isError` is false, and the seeding effect returns early on `!data`, so `config` stays null and this branch (`isLoading || !config`) renders the loading skeleton forever with no way out. Fix: treat a null/absent payload as an error state (throw in the queryFn, or add `data === null` to the error branch) so the `loadError` message is shown.

frontend/src/community/people/assets/images/BirthdayCelebration.tsx
```diff
@@ -0,0 +1,616 @@
+const BirthdayCelebration = () => {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

This inline SVG is ~23.7 KB of JSX (53 `<rect>` confetti shapes plus dozens of paths) and is statically reachable from BaseLayout (BaseLayout → BirthdayNotificationWrapper → BirthdayModalController → modals → BirthdayModalShell → BirthdayCelebrationPanel → here), so it ships in the shared bundle downloaded on every page load, including the login page, for a modal most users see once a year. Recommend lazy-loading the birthday tree with `next/dynamic({ ssr: false })` (at the wrapper or panel level), or moving the illustration to a static asset in `public/` referenced by ``. The `<svg>` is also purely decorative and should carry `aria-hidden="true"` so assistive tech skips it.

frontend/src/community/people/api/PeopleApi.ts
```diff
@@ -926,3 +939,106 @@ export const useReassignSupervisorsAndTerminateOrDeleteEmployee = (
...
+        { timeout: SPECIAL_NOTIFICATION_REQUEST_TIMEOUT_MS, signal }
+      );
+
+      const payload = result?.data?.results?.[0];
+
+      if (!payload) {
+        throw new Error("Birthday notifications response was empty");
+      }
+
+      return {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The response is unvalidated and cast into shape: `payload` is implicitly `any` and `payload.employeeBirthdays ?? []` is asserted `as EmployeeBirthdayType[]`, so a non-array or malformed entry flows straight into `buildBirthdayQueue`/the modals (see the `firstName` crash noted separately). Per the guidelines, prefer fixing the type over asserting it — type the axios response (`authFetch.get<{ results: BirthdayNotificationPayloadType[] }>(...)`) and defensively coerce with `Array.isArray(payload.employeeBirthdays) ? ... : []`. Also, the three new hooks (`useMarkBirthdayNotificationsViewedToday`, `useGetBirthdayNotificationConfig`, `useUpdateBirthdayNotificationConfig`) have no explicit return type annotations, unlike the surrounding hooks in this file which annotate `UseQueryResult<T>`.

frontend/src/community/people/api/PeopleApi.ts
```diff
@@ -926,3 +939,106 @@ export const useReassignSupervisorsAndTerminateOrDeleteEmployee = (
...
+    queryKey: peopleQueryKeys.BIRTHDAY_NOTIFICATIONS_TODAY,
+    queryFn: async ({ signal }) => {
+      const result = await authFetch.get(
+        peoplesEndpoints.GET_TODAYS_BIRTHDAY_NOTIFICATIONS,
+        { timeout: SPECIAL_NOTIFICATION_REQUEST_TIMEOUT_MS, signal }
+      );
+
+      const payload = result?.data?.results?.[0];
+
+      if (!payload) {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Throwing `new Error("Birthday notifications response was empty")` turns a legitimately empty `results` array into a query error. Because `retry: false` and there is no error handling downstream, `data` stays `undefined`, the seeding effect never runs, `markViewed()` is never called, and `shouldEvaluate` stays true — so the `evaluationTick` effect re-hits this endpoint on *every* route change for the rest of the day (including shallow `router.replace` tab switches on the Configurations page, which still emit `routeChangeComplete`). Consider returning `{ lastViewedDate: null, employeeBirthdays: [] }` for an empty payload instead of throwing, and surface real transport failures separately.

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
```diff
@@ -0,0 +1,205 @@
...
+        <>
+          <div className="flex flex-col gap-6">
+            <SwitchRow
+              label={translateText([SECTION, "mainToggleLabel"])}
+              labelId="birthday-notification"
+              arialabel={translateText([SECTION, "aria", "mainToggle"])}
+              checked={config.isTurnedOn}
+              onChange={(checked) => handleToggleChange("isTurnedOn", checked)}
+            />
+            {config.isTurnedOn && (
```
ThinuwanW, line 136: 🤖 **Claude** · 🟡 **Suggestion**

`isTeamWide` and `isOrganizationWide` read as mutually exclusive audiences ("team only" vs "entire organisation") but are independent switches with no validation, so the user can save both true or both false — and `isTurnedOn: true` with both false means notifications are enabled with no audience. Additionally, when the main toggle is switched off the sub-values are only hidden, not reset, so stale `isTeamWide`/`isOrganizationWide` values are still PATCHed. Recommend making the audience a single radio/segmented choice (or adding validation before enabling Save), and normalising the payload when `isTurnedOn` is false.

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
```diff
@@ -0,0 +1,205 @@
...
+  const { data, isLoading, isError } = useGetBirthdayNotificationConfig();
+
+  useEffect(() => {
+    if (!data || config !== null) return;
+    setConfig(data);
+    setInitialConfig(data);
+  }, [data, config]);
+
+  const handleSuccess = (savedConfig: BirthdayNotificationConfigPatchType) => {
+    setInitialConfig(savedConfig as BirthdayNotificationConfigType);
```
ThinuwanW, line 41: 🤖 **Claude** · 🟡 **Suggestion**

`setInitialConfig(savedConfig as BirthdayNotificationConfigType)` uses an `as` assertion to widen the mutation's `Partial<...>` variable back to the full type — the guidelines call out assertions used to silence types rather than fixing them. Since the only caller (`handleSave`) always sends a complete `BirthdayNotificationConfigType`, type `updateBirthdayNotificationConfig`/`useUpdateBirthdayNotificationConfig` with `BirthdayNotificationConfigType` and drop `BirthdayNotificationConfigPatchType` (or keep the patch type only if partial updates are actually a supported call path), removing the cast.

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
```diff
@@ -0,0 +1,205 @@
...
+
+  const [config, setConfig] = useState<BirthdayNotificationConfigType | null>(
+    null
+  );
+  const [initialConfig, setInitialConfig] =
+    useState<BirthdayNotificationConfigType | null>(null);
+
+  const { data, isLoading, isError } = useGetBirthdayNotificationConfig();
+
+  useEffect(() => {
```
ThinuwanW, line 34: 🤖 **Claude** · 🟡 **Suggestion**

The `if (!data || config !== null) return;` guard means local state is seeded exactly once and never resynced. Since `useUpdateBirthdayNotificationConfig` invalidates this query on success, the refetched server state (e.g. values the backend normalised, such as forcing the audience flags false when `isTurnedOn` is false) is fetched and then discarded — the UI keeps showing what was submitted. Consider resyncing when the server data changes and the form is not dirty, or seeding `initialConfig`/`config` from the fresh `data` in the mutation's success path instead of from the submitted payload.

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
```diff
@@ -0,0 +1,205 @@
...
+              onClick={handleCancel}
+            >
+              {translateText(["buttons", "cancel"])}
+            </ButtonV2>
+            <ButtonV2
+              variant="primary"
+              size="md"
+              icon={<SaveIcon />}
+              iconPosition="end"
+              disabled={!isFormChanged || isPending}
```
ThinuwanW, line 178: 🤖 **Claude** · 🟡 **Suggestion**

During save the button is only disabled — there is no in-flight indication. `ButtonV2` supports `isLoading` (see CrmConfigurations.tsx:198 using it for `isCheckingCrmLimit`); add `isLoading={isPending}` to the Save button so the async operation has a visible loading state, per the loading-state guideline.

frontend/src/community/people/components/atoms/BirthdayAvatar/BirthdayAvatar.tsx
```diff
@@ -0,0 +1,66 @@
+import { FC } from "react";
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

This re-implements logic that already exists in `~community/common/components/molecules/Avatar/Avatar.tsx` (image resolution via `useGetUploadedImage`/S3) and `~community/common/components/atoms/DefaultAvatar/DefaultAvatar.tsx` (initials fallback). The three-component split (`BirthdayAvatar` → `UploadedBirthdayAvatar` → `ResolvedBirthdayAvatar`) exists only to avoid calling `useGetImageUrl` when `authPic` is null, and has the side effect that an authPic null→non-null transition swaps component identity and remounts. Consider extending the existing `Avatar`/`DefaultAvatar` with the size/border styling this needs instead, or at minimum reuse `DefaultAvatar`'s initials logic rather than duplicating `getInitials`.

frontend/src/community/people/components/organisms/BirthdayNotificationWrapper/BirthdayNotificationWrapper.tsx
```diff
@@ -0,0 +1,20 @@
+import { FC } from "react";
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

None of the seven new components (BirthdayNotificationWrapper, BirthdayModalController, BirthdayModalShell, SelfBirthdayModal, ColleagueBirthdayModal, BirthdayAvatar, BirthdayCelebrationPanel, PeopleConfigurations) has a `.test.tsx`, and neither does the new `SpecialNotificationProvider`/`useBirthdayNotifications` orchestration — which is where the queue/dismiss/mark-viewed sequencing bugs live. The util-level tests are good but cover only pure functions. Please add at least: a controller test asserting self-first-then-colleagues sequencing and dismiss-through-queue behaviour, and a PeopleConfigurations test for the loading/error/dirty-state branches (with the `MockTheme` wrapper and `screen.getByRole` queries). The new `tooltip` prop on SwitchRow is also uncovered by the existing SwitchRow.test.tsx.

frontend/src/community/common/assets/languages/english/configurations.json
```diff
@@ -362,5 +363,35 @@
...
     }
+  },
+  "people": {
+    "buttons": {
+      "cancel": "Cancel",
+      "save": "Save changes"
+    },
+    "birthdayNotificationSection": {
+      "title": "Birthday Notification",
+      "description": "Control whether birthday notification are sent across the organisation",
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Grammar: "Control whether birthday notification are sent across the organisation" → "birthday notifications are sent". Also note this block mixes British "organisation" in copy with the American "isOrganizationWide"/`organizationToggle` keys; worth aligning the user-facing spelling with the rest of the product copy.

frontend/src/community/people/hooks/useBirthdayNotifications.ts
```diff
@@ -0,0 +1,167 @@
...
+      markBirthdayViewedDate(data.lastViewedDate);
+    }
+
+    const birthdayQueue = buildBirthdayQueue(
+      data.employeeBirthdays,
+      currentEmployeeId
+    );
+
+    if (birthdayQueue.length === 0) {
+      if (data.lastViewedDate !== today) {
```
ThinuwanW, line 83: 🤖 **Claude** · 🟣 **Nit**

`today` comes from `getDateForPeriod("day", "start")`, which is the *browser's* local ISO date, while `lastViewedDate` is produced server-side. For users whose timezone offset crosses midnight relative to the server, `data.lastViewedDate !== today` can be wrong by a day, causing either a redundant mark-viewed PATCH or (combined with the suppression fix) a missed/duplicated modal. Consider comparing against a server-supplied "today" from the same payload rather than a locally computed date.

frontend/src/community/common/providers/SpecialNotificationProvider.tsx
```diff
@@ -0,0 +1,145 @@
...
+      isEligible,
+      today,
+      evaluationTick,
+      isViewedToday,
+      persistViewedDate
+    }),
+    [isEligible, today, evaluationTick, isViewedToday, persistViewedDate]
+  );
+
+  return (
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

In BaseLayout the provider wraps only `<BirthdayNotificationWrapper />`, not `{children}`, so `useSpecialNotifications()` throws for any other consumer — the context abstraction currently has exactly one child and buys nothing over passing values directly. If the intent is to support future special-notification types (as the `SpecialNotificationType` enum and the `OTHER_TYPE` test placeholder suggest), move the provider up to wrap `renderComponent` as well; otherwise consider collapsing it into a hook.

frontend/src/community/common/components/atoms/SwitchRow/SwitchRow.tsx
```diff
@@ -65,6 +68,7 @@ const SwitchRow: FC<SwitchComponentProps> = ({
         >
           {label}
           {!!icon && <Icon name={icon} />}
+          {!!tooltip && <Tooltip title={tooltip} isDisabled={disabled} />}
```
ThinuwanW, line 71: 🤖 **Claude** · 🟣 **Nit**

The Tooltip is rendered inside `<Typography component="label">`, and Tooltip renders a `<span tabIndex={0}>` trigger. Nesting an independently focusable, Enter-activated element inside a `<label>` is fragile for keyboard/screen-reader users (and would activate the labelled control if a `htmlFor` association is ever added — note `labelId` is currently set but never referenced by the `Toggle`, which only receives `ariaLabel`). Consider rendering the tooltip as a sibling of the label inside the `Stack` instead of a child of it.

frontend/src/community/common/assets/languages/english/peopleModule.json
```diff
@@ -1148,5 +1148,16 @@
     "crmSalesManagerLimitationDescription": "Only one CRM Sales Manager account is allowed.",
     "crmSalesRepresentativeLimitationTitle": "CRM Sales Representative Limit Reached",
     "crmSalesRepresentativeLimitationDescription": "Only one CRM Sales Representative account is allowed."
+  },
+
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Stray blank line inserted between the previous object and the new `birthdayNotifications` key; no other top-level key in this file is separated this way. Harmless for JSON parsing but inconsistent with the file's formatting.

frontend/src/community/people/hooks/useBirthdayNotifications.ts
```diff
@@ -0,0 +1,167 @@
...
+  const {
+    isEligible,
+    today,
+    evaluationTick,
+    isViewedToday,
+    persistViewedDate
+  } = useSpecialNotifications();
+
+  const { data: currentEmployee, isPending: isCurrentEmployeePending } =
+    useGetUserPersonalDetails();
```
ThinuwanW, line 33: 🤖 **Claude** · 🔴 **Important**

`useGetUserPersonalDetails()` is called with no `enabled` gate, while every other data path in this hook is carefully gated by `isEligible`/`shouldEvaluate`. Because `BirthdayNotificationWrapper` mounts from `BaseLayout` on *every* route, this now issues a `GET /v1/people/me` on public pages (sign-in, sign-up, reset-password) and for sessions that can never show a modal (super-admin-only, non-drawer routes). Previously `/me` was only fetched from `AppBar`, which lives inside `ContentWithDrawer` and therefore never rendered on public pages — this is new behaviour. It is also nearly always redundant: `useSessionData().employeeDetails.employeeId` already comes from the JWT `employee` claim. Fix: add an `enabled` parameter to `useGetUserPersonalDetails` and pass `shouldEvaluate` (or `isEligible`), or drop the `/me` fallback entirely and rely on the session claim.

frontend/src/community/people/hooks/useBirthdayNotifications.ts
```diff
@@ -0,0 +1,167 @@
...
+
+  const currentEmployeeId =
+    normalizeEmployeeId(employeeDetails?.employeeId) ??
+    normalizeEmployeeId(currentEmployee?.employeeId);
+
+  const isCurrentEmployeeResolved =
+    currentEmployeeId !== undefined || !isCurrentEmployeePending;
```
ThinuwanW, line 40: 🤖 **Claude** · 🔴 **Important**

`isCurrentEmployeeResolved` is satisfied by `!isCurrentEmployeePending` even when `currentEmployeeId` is still `undefined`. That is reachable: if the JWT has no `employee` claim and the `/me` call resolves without data (the shared `authFetch` response interceptor returns `undefined` for HTTP 401 rather than rejecting, so the query settles as *success* with `data === undefined`), the effect proceeds with `currentEmployeeId === undefined`. `buildBirthdayQueue`/`partitionBirthdays` then classify everyone — including the viewer — as `COLLEAGUE` (the util test 'treats everyone as a colleague when the viewer id is unavailable' encodes exactly this), so the user is shown "Today is <their own name>'s birthday!" about themselves. Fix: require `currentEmployeeId !== undefined` before seeding the queue (bail out of the effect otherwise), so an unresolvable viewer means 'show nothing' rather than 'show self as colleague'.

frontend/src/community/people/api/PeopleApi.ts
```diff
@@ -926,3 +939,106 @@ export const useReassignSupervisorsAndTerminateOrDeleteEmployee = (
...
+export const useGetBirthdayNotificationConfig = () => {
+  return useQuery({
+    queryKey: peopleConfigQueryKeys.BIRTHDAY_NOTIFICATION_CONFIG(),
+    queryFn: getBirthdayNotificationConfig,
+    retry: false,
+    refetchOnWindowFocus: false
+  });
+};
+
+const updateBirthdayNotificationConfig = async (
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`updateBirthdayNotificationConfig` awaits `authFetch.patch(...)` and discards the response without validating it. The shared response interceptor in `TanStackProvider.tsx:75-77` does `if (error?.response?.status === 401) { return; }` — i.e. it resolves with `undefined` instead of rejecting. A PATCH that fails with 401 therefore resolves, `onSuccess` fires, `PeopleConfigurations` shows the "Settings Updated" toast and moves `initialConfig` to the unsaved values, and the form reports no pending changes — a silent false success for a write the server rejected. Fix: inspect the resolved value in `mutationFn` (e.g. `const result = await authFetch.patch(...); if (!result?.data?.results) throw new Error(...)`) so the mutation rejects and `onError` runs. The same guard is worth adding to `useMarkBirthdayNotificationsViewedToday`.

frontend/src/community/people/hooks/useBirthdayNotifications.ts
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    const birthdayQueue = buildBirthdayQueue(
+      data.employeeBirthdays,
+      currentEmployeeId
+    );
+
+    if (birthdayQueue.length === 0) {
+      if (data.lastViewedDate !== today) {
+        markViewed();
```
ThinuwanW, line 84: 🤖 **Claude** · 🔴 **Important**

`markViewed()` is fired when `birthdayQueue.length === 0`, i.e. when nothing was displayed. Two consequences: (1) every user issues a PATCH write on their first page load of every day, purely to record "there was nothing to see" — the local viewed-cache already suppresses re-evaluation for the rest of the session, so the server write buys nothing here; (2) it permanently marks the day as viewed server-side, so if birthdays become visible later that same day (a new employee is added, an admin turns the feature or an audience toggle on mid-day, or the backend recomputes for a different timezone) the user will never be shown them on any device. Fix: only call `markViewed()` from `finishSequence()` after a modal has actually been shown; for the empty case, record suppression locally only (and re-evaluate on later route changes).

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
```diff
@@ -0,0 +1,205 @@
...
+            )}
+          </div>
+
+          <div className="flex flex-row gap-4">
+            <ButtonV2
+              variant="tertiary"
+              size="md"
+              icon={<CloseIcon />}
+              iconPosition="end"
+              disabled={!isFormChanged}
```
ThinuwanW, line 168: 🤖 **Claude** · 🟡 **Suggestion**

The Cancel button is only `disabled={!isFormChanged}` while Save is `disabled={!isFormChanged || isPending}`. `CrmConfigurations.tsx:222,231` disables both on the in-flight flag. Because Cancel stays live during the PATCH, a user can click Cancel mid-save: `setConfig(initialConfig)` reverts the local state, then `handleSuccess` sets `initialConfig` to the *saved* (new) values — so the form ends up showing the reverted values, `isFormChanged` flips back to true, and a "Settings Updated" toast appears for values the user just discarded. Add `|| isPending` to the Cancel button's `disabled`.

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
```diff
@@ -0,0 +1,205 @@
...
+  ) => {
+    setConfig((previousConfig) =>
+      previousConfig ? { ...previousConfig, [key]: checked } : previousConfig
+    );
+  };
+
+  const handleCancel = () => setConfig(initialConfig);
+
+  const handleSave = () => {
+    if (!config || isPending) return;
```
ThinuwanW, line 81: 🤖 **Claude** · 🟡 **Suggestion**

`handleSave` performs no validation before sending. Two reachable bad states: (a) `isTurnedOn: true` with both `isTeamWide` and `isOrganizationWide` false — birthday notifications are 'on' but have no audience, so nobody is ever notified, and the UI gives no indication anything is wrong; (b) when the master toggle is turned off the two sub-toggles are unmounted but their previous values are still sent in the payload, so the persisted record keeps audience flags that the user can no longer see. Fix: block save (or surface an inline error) when `isTurnedOn` is true and neither audience flag is set, and decide explicitly whether the sub-flags should be reset or preserved when `isTurnedOn` is false rather than leaving it implicit.

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
```diff
@@ -0,0 +1,205 @@
...
+      <div className="flex flex-col gap-3">
+        <h2 className="subtitle2 text-black">
+          {translateText([SECTION, "title"])}
+        </h2>
+        <p className="body1 text-secondary-text">
+          {translateText([SECTION, "description"])}
+        </p>
+      </div>
+
+      <div className="sr-only" role="status" aria-live="polite">
```
ThinuwanW, line 103: 🤖 **Claude** · 🟡 **Suggestion**

This `role="status" aria-live="polite"` region is populated from `config.isTurnedOn` as soon as the query resolves, not only when the user flips the toggle. Live regions announce content *changes*, so on first load screen-reader users hear "Team and organisation notification options are now hidden" (or "…now available") unprompted, which is confusing since nothing changed. Fix: track whether the toggle has been changed by the user (e.g. set the message inside `handleToggleChange` and clear it on load/cancel) so the region is empty until a real change occurs.

frontend/src/community/common/utils/specialNotificationUtils.ts
```diff
@@ -0,0 +1,65 @@
...
+        return null;
+
+      return cachedValue as SpecialNotificationViewedCacheType;
+    } catch {
+      removeDataFromLocalStorage(SPECIAL_NOTIFICATION_VIEWED_CACHE_KEY);
+      return null;
+    }
+  };
+
+export const writeViewedCache = (
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`writeViewedCache` (and `clearViewedCache` below it, plus the `removeDataFromLocalStorage` call inside `readViewedCache`'s catch) invoke `setDataToLocalStorage`/`removeDataFromLocalStorage`, which are declared `async` in `accessLocalStorage.ts`. `localStorage.setItem` can throw — quota exceeded, or Safari private browsing — and because the helper is async that becomes a rejected promise that nothing awaits or catches, i.e. an unhandled rejection rather than a handled failure. `readViewedCache` already guards its read with try/catch; the write path is inconsistent. Fix: wrap the write/remove calls in try/catch (or `.catch(() => {})`) so a storage failure degrades to 'no cache' instead of an unhandled rejection.

frontend/src/community/people/components/atoms/BirthdayAvatar/BirthdayAvatar.tsx
```diff
@@ -0,0 +1,66 @@
...
+  className = ""
+}) => (
+  <div
+    id={id}
+    role="img"
+    aria-label={alt}
+    className={`${AVATAR_CLASS} ${className}`}
+  >
+    {src ? (
+      <img src={src} alt="" aria-hidden className="h-full w-full object-cover" />
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The raw `` has no `onError` handler, so a failed image load renders the browser's broken-image glyph instead of falling back to the initials. This is a behavioural regression versus the component being reimplemented: MUI's `Avatar` (used by `~community/common/components/molecules/Avatar/Avatar.tsx`) automatically falls back to its children when the image errors. The failure is reachable — in community mode `useGetImageUrl` does `if (logoUrl) setImage(logoUrl); else if (src) setImage(src)`, so when the signed URL isn't available yet it sets `src` to the raw storage filename, which will 404. Fix: add local state cleared by `onError` on the `` and render the initials branch when it fires.

frontend/src/community/people/components/molecules/BirthdayModals/SelfBirthdayModal.tsx
```diff
@@ -0,0 +1,40 @@
+import { FC } from "react";
```
ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

`SelfBirthdayModal` and `ColleagueBirthdayModal` are byte-for-byte identical apart from the modal id, the translation sub-key (`self` vs `colleague`) and the name passed to interpolation (`getFullName(employee)` vs `employee.firstName.trim()`). `BirthdayModalController` then exists purely to branch between them. Per the duplication guideline, collapse these into one component parameterised by `BirthdayModalVariant` (deriving the id, sub-key and display name from the variant) and let `BirthdayNotificationWrapper` render it directly — that removes two of the three files and the controller's branch.

frontend/src/community/people/api/utils/ApiEndpoints.ts
```diff
@@ -38,7 +38,9 @@ export const peoplesEndpoints = {
   REASSIGN_SUPERVISORS_AND_TERMINATE_OR_DELETE_EMPLOYEE: (userId: number) =>
     `${moduleAPIPath.PEOPLE}/user/${userId}/reassign-supervisors-and-terminate-or-delete`,
   GET_SKILLS: `${moduleAPIPath.PEOPLE}/skills`,
-  CREATE_CUSTOM_SKILLS: `${moduleAPIPath.PEOPLE}/skills/custom`
+  CREATE_CUSTOM_SKILLS: `${moduleAPIPath.PEOPLE}/skills/custom`,
+  GET_TODAYS_BIRTHDAY_NOTIFICATIONS: `${moduleAPIPath.PEOPLE}/birthday-notifications/today`,
```
ThinuwanW, line 44: 🤖 **Claude** · 🟡 **Suggestion**

None of the three endpoints this PR introduces exist in the `backend/` module of this repo: grepping the backend for `birthday-notifications`, `people-config` or any birthday controller/DTO returns nothing (only unrelated `dateOfBirth` matches). So `GET /v1/people/birthday-notifications/today`, `PATCH /v1/people/birthday-notifications/mark-viewed-today` and `GET|PATCH /v1/people-config/birthday-notifications` will all 404 as shipped: the modal never appears and the new People configurations tab renders its permanent loading skeleton. Either land the paired backend change in the same PR or gate the feature (env flag / feature toggle) until the API is available. Separately, `peopleConfigEndpoints` hardcodes the `/people-config` prefix as a literal while every sibling group builds paths from `moduleAPIPath` — add a `PEOPLE_CONFIG` member to that enum for consistency.

frontend/src/community/common/providers/SpecialNotificationProvider.tsx
```diff
@@ -0,0 +1,145 @@
+import { useRouter } from "next/router";
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

This provider re-implements a pattern that already exists: `~enterprise/common/providers/AnnouncementProvider.tsx` does the same localStorage-cache + `routeChangeComplete` + `isSuperAdminOnlySession` gating for its own modal queue. Worse, there is no coordination between them — in enterprise mode `_app.tsx` renders `<AnnouncementWrapper />` and `BaseLayout` renders `<BirthdayNotificationWrapper />`, both of which mount `LargeModal` with `isOpen` on first authenticated load, so an announcement and a birthday modal can stack on top of each other with two dark backdrops and ambiguous focus/Escape handling. The generic `SpecialNotification*` naming suggests this was meant to be the single coordinator: either route announcements through it too, or add mutual exclusion so only one special notification is presented at a time.

frontend/src/community/common/constants/specialNotificationConstants.ts
```diff
@@ -0,0 +1,4 @@
+export const SPECIAL_NOTIFICATION_VIEWED_CACHE_KEY =
+  "specialNotificationsViewed";
+
+export const SPECIAL_NOTIFICATION_REQUEST_TIMEOUT_MS = 3000;
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`SPECIAL_NOTIFICATION_REQUEST_TIMEOUT_MS = 3000` is applied to both the GET and the `mark-viewed-today` PATCH. 3s is aggressive for this backend (cold JVM start, tenant lookup) and the axios timeout fires before React Query's `retry: 1` can help. For the read path a timeout just means no modal, which is fine; for the write path a timeout means the viewed date is never persisted, so the modal sequence replays on the next navigation. Consider raising the timeout (or dropping it on the PATCH and letting the default apply) and increasing `retry` on the mutation.

frontend/src/community/people/assets/images/BirthdayCelebration.tsx
```diff
@@ -0,0 +1,616 @@
...
+  return (
+    <svg
```
ThinuwanW, line 5: 🤖 **Claude** · 🟣 **Nit**

Three small things on this file: (1) the `<svg>` is purely decorative but has no `aria-hidden="true"` / `role="presentation"`, so assistive tech may announce it as a graphic inside the modal — add `aria-hidden`; (2) the component has no explicit type or return annotation, unlike the `FC`-typed components elsewhere in the PR; (3) `assets/images/` in this codebase otherwise contains only binary assets (`src/community/common/assets/images/*.png`), so a 616-line JSX component is misplaced — a real `.svg` under `~public/` (or the component under `components/atoms/`) matches the existing structure.

frontend/src/community/people/components/molecules/BirthdayCelebrationPanel/BirthdayCelebrationPanel.tsx
```diff
@@ -0,0 +1,11 @@
+import { FC } from "react";
+
+import BirthdayCelebration from "~community/people/assets/images/BirthdayCelebration";
+
+const BirthdayCelebrationPanel: FC = () => (
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

This component adds nothing but a `relative h-full w-full bg-white` wrapper around `BirthdayCelebration`, yet costs a directory, a file and an extra indirection between `BirthdayModalShell` and the illustration. Move the wrapper div into `BirthdayCelebration` (or inline it at the `image={...}` call site in `BirthdayModalShell`) and delete this file.

frontend/src/community/common/assets/languages/english/configurations.json
```diff
@@ -362,5 +363,35 @@
...
+      "organizationToggleTooltip": "Notifies all employees in the organisation, regardless of team",
+      "loadError": "We couldn't load birthday notification settings. Please refresh the page and try again.",
+      "toastMessages": {
+        "successTitle": "Settings Updated",
+        "successDescription": "Birthday notification settings have been updated successfully.",
+        "errorTitle": "Oops! Something went wrong",
+        "errorDescription": "Failed to update settings. Please try again."
+      },
+      "aria": {
+        "mainToggle": "Birthday notification",
```
ThinuwanW, line 391: 🤖 **Claude** · 🟣 **Nit**

`aria.mainToggle`, `aria.teamToggle` and `aria.organizationToggle` are character-for-character identical to `mainToggleLabel`, `teamToggleLabel` and `organizationToggleLabel`, and `SwitchRow.getAriaLabel()` already falls back to `label` when `arialabel` is undefined (`SwitchRow.tsx:53`). These three keys and the corresponding `arialabel` props in `PeopleConfigurations` are dead weight that translators will have to keep in sync twice. Drop them and let the label fallback do its job; keep only aria strings that genuinely differ from the visible text.

frontend/src/community/common/providers/SpecialNotificationProvider.tsx
```diff
@@ -0,0 +1,145 @@
...
+import {
+  clearViewedCache,
+  isViewedToday as isViewedTodayForType,
+  readViewedCache,
+  withViewedDate,
+  writeViewedCache
+} from "~community/common/utils/specialNotificationUtils";
+
+const AUTHENTICATED = "authenticated";
+const UNAUTHENTICATED = "unauthenticated";
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`const AUTHENTICATED = "authenticated"` / `const UNAUTHENTICATED = "unauthenticated"` are declared here and `UNAUTHENTICATED` is declared again in `useBirthdayNotifications.ts:27`. These mirror the untyped string literals returned by `useSessionData` (`"loading" | "authenticated" | "unauthenticated"`). Introduce a shared `SessionStatus` enum (or exported const union) in `~community/common/enums` and have `useSessionData` return it, so the two new files stop duplicating magic strings and comparisons are type-checked.

frontend/src/community/common/components/atoms/SwitchRow/SwitchRow.tsx
```diff
@@ -19,6 +20,7 @@ interface SwitchComponentProps {
   error?: string;
   wrapperStyles?: SxProps<Theme>;
   icon?: IconName;
+  tooltip?: string;
```
ThinuwanW, line 23: 🤖 **Claude** · 🟣 **Nit**

The new `tooltip` prop is a shared-atom change with two gaps. (1) `SwitchRow.test.tsx` already exists with five cases but was not extended, so nothing covers the tooltip branch. (2) `Tooltip` is rendered without `ariaLabel`, and its trigger is a bare focusable `<span>` (`Tooltip.tsx:107-121`) whose only accessible content is the MUI `title` shown on hover/focus — screen-reader users reach an unnamed control and the explanatory text ('Notifies only the direct team members…') is not associated with the toggle it describes. Pass an `ariaLabel` (or wire the tooltip text to the Toggle via `aria-describedby`) and add a test asserting the tooltip renders only when the prop is supplied.

frontend/src/community/people/types/BirthdayNotificationTypes.ts
```diff
@@ -0,0 +1,26 @@
...
+export type EmployeeIdType = number | string | null | undefined;
+
+export interface EmployeeBirthdayType {
+  employeeId: number;
+  firstName: string;
+  lastName: string;
+  authPic: string | null;
+  email: string;
+}
+
```
ThinuwanW, line 10: 🤖 **Claude** · 🟣 **Nit**

`email` is declared required on `EmployeeBirthdayType` but is never read anywhere in the feature — not in `BirthdayModalShell`, the two modals, `birthdayNotificationUtils`, or the hook; it only shows up in test fixtures. Either drop it from the type (and ideally from the backend response, since it means every user receives colleagues' email addresses for no purpose) or use it. If it is being kept for a planned follow-up, mark it optional so the contract doesn't overstate what the client needs.

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
```diff
@@ -0,0 +1,77 @@
...
+  return (
+    <LargeModal
+      id={id}
+      isOpen
+      className={BIRTHDAY_MODAL_CLASS}
+      imagePosition="left"
+      backdropVariant="dark"
+      onClose={onDismiss}
+      closeButtonAriaLabel={translateAria(["closeButton"])}
+      ariaLabel={heading}
```
ThinuwanW, line 41: 🤖 **Claude** · 🟣 **Nit**

`ariaLabel` is passed to `LargeModal` here and nowhere else in the codebase — every other `LargeModal` usage (`UpgradeToCoreModal`, `AnnouncementModal`, `PolicyTypeSelectionModal`, `GeofenceSelectorModal`, `StartQuickSetupModal`) supplies `modalHeader` instead, and this modal deliberately omits `modalHeader`. Worth confirming against `@rootcodelabs/skapp-ui@1.1.1-beta.5` that `ariaLabel` is actually forwarded onto the dialog element; if it is a loosely-typed passthrough the dialog ends up with no accessible name at all. Also note `ariaLabel={heading}` duplicates the `<h2>` a few lines below, so where it does work the heading is announced twice.

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
```diff
@@ -0,0 +1,205 @@
...
+      toastType: ToastType.ERROR,
+      title: translateText([SECTION, "toastMessages", "errorTitle"]),
+      description: translateText([SECTION, "toastMessages", "errorDescription"])
+    });
+  };
+
+  const { mutate: updateConfig, isPending } =
+    useUpdateBirthdayNotificationConfig(handleSuccess, handleError);
+
+  const isFormChanged =
```
hasalarootcode, line 62: try to simplify

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
```diff
@@ -0,0 +1,205 @@
...
+        <p className="body1 text-secondary-text">
+          {translateText([SECTION, "description"])}
+        </p>
+      </div>
+
+      <div className="sr-only" role="status" aria-live="polite">
+        {subOptionsAriaMessage}
+      </div>
+
+      {isError && (
```
hasalarootcode: check and remove

frontend/src/community/people/api/utils/QueryKeys.ts
```diff
@@ -133,3 +134,10 @@ export const getPeopleDashboardQueryKeys = {
     );
   }
 };
+
+export const peopleConfigQueryKeys = {
+  ALL: ["people-config"],
+  BIRTHDAY_NOTIFICATION_CONFIG: function () {
+    return [...(this?.ALL || []), "birthday-notification-config"];
```
hasalarootcode: remove this.ALL

frontend/src/community/people/api/PeopleApi.ts
```diff
@@ -926,3 +939,106 @@ export const useReassignSupervisorsAndTerminateOrDeleteEmployee = (
...
+
+export const useGetTodaysBirthdayNotifications = (
+  enabled: boolean
+): UseQueryResult<BirthdayNotificationPayloadType> => {
+  return useQuery({
+    queryKey: peopleQueryKeys.BIRTHDAY_NOTIFICATIONS_TODAY,
+    queryFn: async ({ signal }) => {
+      const result = await authFetch.get(
+        peoplesEndpoints.GET_TODAYS_BIRTHDAY_NOTIFICATIONS,
+        { timeout: BIRTHDAY_NOTIFICATION_REQUEST_TIMEOUT_MS, signal }
```
hasalarootcode: remove timeout

frontend/src/community/people/api/PeopleApi.ts
```diff
@@ -926,3 +939,106 @@ export const useReassignSupervisorsAndTerminateOrDeleteEmployee = (
...
+      );
+
+      const payload = result?.data?.results?.[0];
```
hasalarootcode: remove this also and redirectly return [0] objewct

frontend/src/community/people/api/PeopleApi.ts
```diff
@@ -926,3 +939,106 @@ export const useReassignSupervisorsAndTerminateOrDeleteEmployee = (
...
+      }
+
+      return {
+        lastViewedDate: (payload.lastViewedDate ?? null) as string | null,
+        employeeBirthdays: (payload.employeeBirthdays ??
+          []) as EmployeeBirthdayType[]
+      };
+    },
+    enabled,
+    retry: false,
```
hasalarootcode: remove these unnessary attributes

frontend/src/community/people/api/PeopleApi.ts
```diff
@@ -926,3 +939,106 @@ export const useReassignSupervisorsAndTerminateOrDeleteEmployee = (
...
+export const useUpdateBirthdayNotificationConfig = (
+  onSuccess: (config: BirthdayNotificationConfigPatchType) => void,
+  onError: () => void
+) => {
+  const queryClient = useQueryClient();
+
+  return useMutation({
+    mutationFn: updateBirthdayNotificationConfig,
+    onSuccess: (_data, config) => {
+      queryClient
```
hasalarootcode: try to refactor these things by refering leavepolicyapi file

frontend/src/community/people/components/molecules/BirthdayCelebrationPanel/BirthdayCelebrationPanel.tsx
hasalarootcode: move inside your modal directory

frontend/src/community/people/components/organisms/BirthdayNotificationWrapper/BirthdayNotificationWrapper.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+import BirthdayModalController from "~community/people/components/organisms/BirthdayModalController/BirthdayModalController";
+import useBirthdayNotifications from "~community/people/hooks/useBirthdayNotifications";
+
+const BirthdayNotificationWrapper: FC = () => {
```
hasalarootcode: remove this and use this inside the controller

frontend/src/community/people/constants/birthdayNotificationConstants.ts
```diff
@@ -0,0 +1,11 @@
+export const BIRTHDAY_MODAL_CLASS =
```
hasalarootcode: use these inline and remove this

frontend/src/community/people/enums/BirthdayNotificationEnums.ts
```diff
@@ -0,0 +1,4 @@
+export enum BirthdayModalVariant {
+  SELF = "self",
```
hasalarootcode: make the value uppercase

frontend/src/community/people/hooks/useBirthdayNotifications.ts
```diff
@@ -0,0 +1,151 @@
...
+    evaluationTick,
+    isViewedToday,
+    cacheLastViewedDate
+  } = useBirthdayNotificationContext();
+
+  const { data: currentEmployee, isPending: isCurrentEmployeePending } =
+    useGetUserPersonalDetails();
+
+  const currentEmployeeId =
+    normalizeEmployeeId(employeeDetails?.employeeId) ??
```
hasalarootcode: try to do this without normalized util seems to be possible without this

frontend/src/community/people/hooks/useBirthdayNotifications.ts
```diff
@@ -0,0 +1,151 @@
...
+    normalizeEmployeeId(currentEmployee?.employeeId);
+
+  const isCurrentEmployeeResolved =
+    currentEmployeeId !== undefined || !isCurrentEmployeePending;
+
+  const [queue, setQueue] = useState<BirthdayQueueEntryType[]>([]);
+  const [cursor, setCursor] = useState(0);
+
+  const seededDataRef = useRef<BirthdayNotificationPayloadType | null>(null);
+  const isShowingRef = useRef(false);
```
hasalarootcode, line 46: try to do it whithout ref by using onCLose of modal

frontend/src/community/people/hooks/useBirthdayNotifications.ts
```diff
@@ -0,0 +1,151 @@
...
+
+  const shouldEvaluate = isEligible && !isViewedToday;
+
+  const { mutate: markViewed } =
+    useMarkBirthdayNotificationsViewedToday(cacheLastViewedDate);
+
+  const { data, refetch } = useGetTodaysBirthdayNotifications(shouldEvaluate);
+
+  useEffect(() => {
+    if (!data) return;
```
hasalarootcode, line 65: this should simplify with  above comment

frontend/src/community/people/hooks/useBirthdayViewedCache.ts
```diff
@@ -0,0 +1,64 @@
...
+  const { sessionStatus, userId } = useSessionData();
+
+  const [hasHydrated, setHasHydrated] = useState(false);
+  const [viewedCache, setViewedCache] =
+    useState<BirthdayNotificationViewStateType | null>(null);
+
+  const viewedCacheRef = useRef<BirthdayNotificationViewStateType | null>(null);
+
+  useEffect(() => {
+    const hydratedCache = readViewedCache();
```
hasalarootcode: rename

frontend/src/community/people/providers/BirthdayNotificationProvider.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+import { useAuth } from "~community/auth/providers/AuthProvider";
+import { IsAProtectedUrlWithDrawer } from "~community/auth/utils/authUtils";
+import useSessionData from "~community/common/hooks/useSessionData";
+import { isSuperAdminOnlySession } from "~community/common/utils/commonUtil";
+import { getDateForPeriod } from "~community/common/utils/dateTimeUtils";
+import useBirthdayViewedCache from "~community/people/hooks/useBirthdayViewedCache";
+import { BirthdayNotificationContextType } from "~community/people/types/BirthdayNotificationTypes";
+
+const AUTHENTICATED = "authenticated";
```
hasalarootcode: check just user existing wihtout using this attribute

frontend/src/community/people/utils/birthdayNotificationCacheUtils.ts
```diff
@@ -0,0 +1,44 @@
...
+import { BIRTHDAY_NOTIFICATION_VIEW_STATE_CACHE_KEY } from "~community/people/constants/birthdayNotificationConstants";
+import { BirthdayNotificationViewStateType } from "~community/people/types/BirthdayNotificationTypes";
+
+export const readViewedCache = (): BirthdayNotificationViewStateType | null => {
+  try {
+    const cachedValue = getDataFromLocalStorage(
+      BIRTHDAY_NOTIFICATION_VIEW_STATE_CACHE_KEY
+    );
+
+    if (!cachedValue || typeof cachedValue !== "object") return null;
```
hasalarootcode, line 15: if you need to do valdations, do it using a single if

frontend/src/community/people/utils/birthdayNotificationCacheUtils.ts
```diff
@@ -0,0 +1,44 @@
...
+export const clearViewedCache = (): void => {
+  removeDataFromLocalStorage(BIRTHDAY_NOTIFICATION_VIEW_STATE_CACHE_KEY);
+};
+
+export const isViewedToday = (
+  cache: BirthdayNotificationViewStateType | null,
+  today: string | null,
+  currentUserId: number | undefined
+): boolean => {
+  if (!cache || !today || currentUserId === undefined) return false;
```
hasalarootcode, line 37: sinplify if condition

frontend/src/community/people/utils/birthdayNotificationUtils.ts
```diff
@@ -0,0 +1,71 @@
+import { BirthdayModalVariant } from "~community/people/enums/BirthdayNotificationEnums";
+import {
+  BirthdayQueueEntryType,
+  EmployeeBirthdayType,
+  EmployeeIdType,
+  PartitionedBirthdaysType
+} from "~community/people/types/BirthdayNotificationTypes";
+
+export const getFullName = (
```
hasalarootcode: use existing

frontend/src/community/people/utils/birthdayNotificationUtils.ts
```diff
@@ -0,0 +1,71 @@
...
+  employee: Pick<EmployeeBirthdayType, "firstName" | "lastName">
+): string =>
+  [employee.firstName?.trim(), employee.lastName?.trim()]
+    .filter(Boolean)
+    .join(" ");
+
+export const hasDisplayableName = (employee: EmployeeBirthdayType): boolean =>
```
hasalarootcode: remove this

frontend/src/community/people/utils/birthdayNotificationUtils.ts
```diff
@@ -0,0 +1,71 @@
...
+  getFullName(employee).length > 0;
+
+export const normalizeEmployeeId = (
```
hasalarootcode, line 8: remove this

frontend/src/community/people/utils/birthdayNotificationUtils.ts
```diff
@@ -0,0 +1,71 @@
...
+  if (typeof employeeId === "string" && employeeId.trim().length === 0) {
+    return undefined;
+  }
+
+  const numericEmployeeId = Number(employeeId);
+
+  return Number.isFinite(numericEmployeeId) ? numericEmployeeId : undefined;
+};
+
+export const partitionBirthdays = (
```
hasalarootcode, line 24: check the logic

hasalarootcode (changes requested):

frontend/src/community/people/components/molecules/BirthdayCelebrationPanel/BirthdayCelebrationPanel.tsx
gaindurootcode: moved

frontend/src/community/people/constants/birthdayNotificationConstants.ts
gaindurootcode: removed

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
gaindurootcode: removed

frontend/src/community/people/api/utils/QueryKeys.ts
gaindurootcode: removed

frontend/src/community/people/api/PeopleApi.ts
gaindurootcode: removed

frontend/src/community/people/api/PeopleApi.ts
gaindurootcode: fixed

frontend/src/community/people/api/PeopleApi.ts
gaindurootcode: fixed

frontend/src/community/people/api/PeopleApi.ts
gaindurootcode: fixd

frontend/src/community/people/components/organisms/BirthdayNotificationWrapper/BirthdayNotificationWrapper.tsx
gaindurootcode: removed the wrapper and called the controller directly

frontend/src/community/people/enums/BirthdayNotificationEnums.ts
gaindurootcode: updated

frontend/src/community/people/hooks/useBirthdayViewedCache.ts
gaindurootcode: renamed

frontend/src/community/people/providers/BirthdayNotificationProvider.tsx
gaindurootcode: fixed

frontend/src/community/people/utils/birthdayNotificationUtils.ts
gaindurootcode: used existing util

frontend/src/community/people/api/utils/ApiEndpoints.ts
```diff
@@ -96,3 +98,7 @@ export const peopleDashboardEndpoints = {
   GET_PEOPLE_DASHBOARD_ANALYTICS_DATA: (): string =>
     `/people/analytics/dashboard-summary`
 };
+
+export const peopleConfigEndpoints = {
+  BIRTHDAY_NOTIFICATION_CONFIG: `/people-config/birthday-notifications`
```
akilarootcode: GET_

frontend/src/community/people/api/PeopleApi.ts
```diff
@@ -971,3 +986,105 @@ export const useReassignSupervisorsAndTerminateOrDeleteEmployee = (
...
+export const useMarkBirthdayNotificationsViewedToday = (
+  onViewedDatePersisted: (lastViewedDate: string) => void
+): UseMutationResult<
+  AxiosResponse<MarkBirthdayNotificationsViewedResponse>,
+  AxiosError,
+  void
+> => {
+  return useMutation({
+    mutationFn: markBirthdayNotificationsViewedToday,
+    retry: 1,
```
akilarootcode: remove  retry: 1,

frontend/src/community/people/api/PeopleApi.ts
```diff
@@ -971,3 +986,105 @@ export const useReassignSupervisorsAndTerminateOrDeleteEmployee = (
...
+    return response.data.results[0];
+  };
+
+export const useGetBirthdayNotificationConfig = (): UseQueryResult<
+  BirthdayNotificationConfigType
+> => {
+  return useQuery({
+    queryKey: peopleConfigQueryKeys.BIRTHDAY_NOTIFICATION_CONFIG,
+    queryFn: getBirthdayNotificationConfig,
+    retry: false,
```
akilarootcode: remove retry

frontend/src/community/people/api/PeopleApi.ts
```diff
@@ -971,3 +986,105 @@ export const useReassignSupervisorsAndTerminateOrDeleteEmployee = (
...
+  const queryClient = useQueryClient();
+
+  return useMutation({
+    mutationFn: updateBirthdayNotificationConfig,
+    onSuccess: (_data, config) => {
+      queryClient
+        .invalidateQueries({
+          queryKey: peopleConfigQueryKeys.BIRTHDAY_NOTIFICATION_CONFIG
+        })
+        .catch(() => {});
```
akilarootcode: remove catch

frontend/src/community/people/api/PeopleApi.ts
```diff
@@ -971,3 +986,105 @@ export const useReassignSupervisorsAndTerminateOrDeleteEmployee = (
...
+  onViewedDatePersisted: (lastViewedDate: string) => void
+): UseMutationResult<
+  AxiosResponse<MarkBirthdayNotificationsViewedResponse>,
+  AxiosError,
+  void
+> => {
+  return useMutation({
+    mutationFn: markBirthdayNotificationsViewedToday,
+    retry: 1,
+    onSuccess: (response) => {
```
akilarootcode: move onSuccess on parent component

frontend/src/community/people/api/PeopleApi.ts
```diff
@@ -971,3 +986,82 @@ export const useReassignSupervisorsAndTerminateOrDeleteEmployee = (
...
+export const useGetTodaysBirthdayNotifications = (
+  enabled: boolean
+): UseQueryResult<BirthdayNotificationPayloadType> => {
+  return useQuery({
+    queryKey: peopleQueryKeys.GET_TODAYS_BIRTHDAY_NOTIFICATIONS,
+    queryFn: async () =>
+      await authFetch.get<BirthdayNotificationTodayResponse>(
+        peoplesEndpoints.GET_TODAYS_BIRTHDAY_NOTIFICATIONS
+      ),
+    select: (data) => data?.data?.results[0],
```
akilarootcode: remove select on ALL

frontend/src/community/people/hooks/useBirthdayViewedCache.ts
```diff
@@ -0,0 +1,64 @@
...
+  writeViewedCache
+} from "~community/people/utils/birthdayNotificationCacheUtils";
+
+interface BirthdayViewedCacheType {
+  isCacheLoaded: boolean;
+  isViewedToday: boolean;
+  cacheLastViewedDate: (lastViewedDate: string) => void;
+}
+
+const UNAUTHENTICATED = "unauthenticated";
```
akilarootcode: remove this

frontend/src/community/people/providers/BirthdayNotificationProvider.tsx
```diff
@@ -0,0 +1,82 @@
...
+    <BirthdayNotificationContext.Provider value={value}>
+      {children}
+    </BirthdayNotificationContext.Provider>
+  );
+};
+
+const useBirthdayNotificationContext = (): BirthdayNotificationContextType => {
+  const context = useContext(BirthdayNotificationContext);
+
+  if (!context) {
```
akilarootcode, line 83: remove this

akilarootcode (changes requested):

akilarootcode (approved):

hasalarootcode (approved):

PR #1705, opened 2026-08-04
add location filter when getting holidays.

erandi-rootcode:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR correctly scopes the holiday lookup in `getAllActiveSlots` to the employee's work location, reusing the existing `LEFT JOIN + isEmpty(workLocations)` predicate pattern already established elsewhere in `HolidayRepositoryImpl`, and cleanly removes the now-unused `HolidayDao.findAllByIsActiveTrueAndWorkLocationsWorkLocationId` (verified no remaining callers). Null-safety on `getWorkLocation()` is handled and the only caller (`getActiveTimeSlot`) is `@Transactional(readOnly = true)`, so the new LAZY `workLocation` dereference is safe today. Remaining concerns are near-verbatim query duplication, naming inconsistency in the custom repository interface, the semantics of the no-work-location fallback, and sibling holiday lookups in the same attendance flow that were left unfiltered.

Found **11** new issue(s): 🔴 1 important, 🟡 6 suggestion(s), 🟣 4 nit(s)

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/HolidayRepositoryImpl.java
```diff
@@ -102,6 +101,27 @@ public List<Holiday> findAllActiveHolidaysByWorkLocationId(Long workLocationId)
 		return entityManager.createQuery(criteriaQuery).getResultList();
 	}
 
+	@Override
```
ThinuwanW, line 101: 🤖 **Claude** · 🟡 **Suggestion**

`findAllByIsActiveTrueAndDateAndWorkLocationId` is a verbatim copy of `findAllActiveHolidaysByWorkLocationId` (lines 84-102) with a single extra date predicate — same builder setup, same LEFT JOIN, same OR/isEmpty predicate, same distinct. The same work-location OR-predicate also appears a third time in `buildPredicates` (lines 155-158). Extract a shared private helper, e.g. `private Predicate workLocationPredicate(CriteriaBuilder cb, Root<Holiday> root, Long workLocationId)` returning the `or(equal(join.get(workLocationId), id), isEmpty(root.get(workLocations)))` predicate, and have all three call sites use it. Better still, collapse the two public methods into one implementation that takes a nullable `LocalDate date` and adds the date predicate only when non-null, so the "holiday applies to all locations when workLocations is empty" rule lives in exactly one place and cannot drift.

backend/src/main/java/com/skapp/community/peopleplanner/repository/HolidayRepository.java
```diff
@@ -17,4 +16,6 @@ public interface HolidayRepository {
 
 	List<Holiday> findAllActiveHolidaysByWorkLocationId(Long workLocationId);
 
+	List<Holiday> findAllByIsActiveTrueAndDateAndWorkLocationId(LocalDate date, Long workLocationId);
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Naming is inconsistent with the rest of this interface and misleading. `HolidayDao` extends both `JpaRepository` and this custom `HolidayRepository` fragment, and `findAllByIsActiveTrueAndDateAndWorkLocationId` is written in Spring Data derived-query style — it reads exactly like the derived methods declared right next to it in `HolidayDao` (`findAllByIsActiveTrueAndDate`, `findAllByIsActiveTrueAndDateBetween`), yet it is hand-implemented via Criteria API and its actual semantics are NOT what the name implies (it also returns holidays with no work locations at all, i.e. an OR, not an AND). The two existing custom methods here are named `findAllActiveHolidaysByDateWithWorkLocations` and `findAllActiveHolidaysByWorkLocationId`. Rename to something like `findAllActiveHolidaysByDateAndWorkLocationId` to match the fragment's convention and to stop implying a plain derived AND-query.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeServiceImpl.java
```diff
@@ -1271,9 +1271,15 @@ private ResponseEntityDto getAllActiveSlotsNoLeaveDay(TimeConfig currentDayConfi
...
-	private ResponseEntityDto getAllActiveSlots(LocalDate currentDate, TimeConfig currentDayConfig, float morningHours,
-			float eveningHours) {
-		List<Holiday> holidayList = holidayDao.findAllByIsActiveTrueAndDate(currentDate);
+	private ResponseEntityDto getAllActiveSlots(User currentUser, LocalDate currentDate, TimeConfig currentDayConfig,
+			float morningHours, float eveningHours) {
+
+		Long employeeWorkLocationId = currentUser.getEmployee().getWorkLocation() != null
+				? currentUser.getEmployee().getWorkLocation().getWorkLocationId() : null;
+
+		List<Holiday> holidayList = employeeWorkLocationId == null
```
ThinuwanW, line 1280: 🤖 **Claude** · 🟡 **Suggestion**

When the employee has no work location, the fallback `findAllByIsActiveTrueAndDate(currentDate)` returns EVERY active holiday on that date, including holidays that are explicitly scoped to specific work locations. Combined with `CLOCK_IN_ON_COMPANY_HOLIDAYS = false`, an employee with no work location will be blocked from clocking in because of a holiday that belongs only to some other office — the exact scenario this PR is trying to prevent. The more defensible fallback is "only holidays that apply to all locations", i.e. holidays whose `workLocations` set is empty. Consider passing `null` through to the criteria method and skipping the work-location OR-branch in favour of a plain `isEmpty(workLocations)` predicate when the id is null, rather than branching to the unfiltered derived query. If the current behaviour is deliberate (it does mirror `LeaveServiceImpl.getHolidaysForEmployee`, which falls back to `findAllByIsActiveTrue()`), a short comment stating that employees without a work location intentionally observe all holidays would prevent this from being re-litigated.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeServiceImpl.java
```diff
@@ -1271,9 +1271,15 @@ private ResponseEntityDto getAllActiveSlotsNoLeaveDay(TimeConfig currentDayConfi
 		return null;
 	}
 
-	private ResponseEntityDto getAllActiveSlots(LocalDate currentDate, TimeConfig currentDayConfig, float morningHours,
-			float eveningHours) {
-		List<Holiday> holidayList = holidayDao.findAllByIsActiveTrueAndDate(currentDate);
+	private ResponseEntityDto getAllActiveSlots(User currentUser, LocalDate currentDate, TimeConfig currentDayConfig,
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The work-location filter was applied to only one of several holiday lookups in the same attendance flow, so the same employee can now see contradictory holiday information. In particular `getRequestedDateTimeAvailability` (line 456) calls `holidayDao.findAllByIsActiveTrueAndDateBetween(requestDto.getDate(), requestDto.getDate())` for the current user's own date and returns it as `response.setHoliday(...)` — that is the direct analogue of this change and is still unfiltered, so the availability screen will report a holiday for another location while `getActiveTimeSlot` (correctly) will not. `getEmployeesDailyTimeRecords` (line 372) and the time-record summary at line 1569 have the same gap, and `TimeAnalyticsServiceImpl` lines 336/427 use `findByIsActiveTrueAndDate` unfiltered. Either extend the filter to these call sites in this PR or record explicitly why they are out of scope, so the inconsistency is a known state rather than an oversight.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeServiceImpl.java
```diff
@@ -1209,7 +1209,7 @@ public ResponseEntityDto checkLeaveOrHolidayOrNonWorkingDay() {
 			if (activeTimeSlotResponseDto1 != null)
 				return activeTimeSlotResponseDto1;
 
-			return getAllActiveSlots(currentDate, currentDayConfig, morningHours, eveningHours);
+			return getAllActiveSlots(currentUser, currentDate, currentDayConfig, morningHours, eveningHours);
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`checkLeaveOrHolidayOrNonWorkingDay()` is `public`, carries no `@Transactional`, and is not declared on the `TimeService` interface (no `@Override`). It only worked before because its single caller `getActiveTimeSlot()` is `@Transactional(readOnly = true)`. This change makes it newly dependent on an active persistence context: `Employee.workLocation` is `@ManyToOne(fetch = FetchType.LAZY)` (Employee.java:173-175), so `currentUser.getEmployee().getWorkLocation()` now forces initialisation of the Employee proxy and the WorkLocation association — previously only `getEmployeeId()` was touched, which a proxy can serve without a session. Any future caller (or an enterprise bean invoking it outside a transaction) would get a `LazyInitializationException` rather than a clean failure. Since it has exactly one in-class caller, make it `private`; if it must stay public, add `@Transactional(readOnly = true)` and declare it on the interface.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeServiceImpl.java
```diff
@@ -1271,9 +1271,15 @@ private ResponseEntityDto getAllActiveSlotsNoLeaveDay(TimeConfig currentDayConfi
...
-		List<Holiday> holidayList = holidayDao.findAllByIsActiveTrueAndDate(currentDate);
+	private ResponseEntityDto getAllActiveSlots(User currentUser, LocalDate currentDate, TimeConfig currentDayConfig,
+			float morningHours, float eveningHours) {
+
+		Long employeeWorkLocationId = currentUser.getEmployee().getWorkLocation() != null
+				? currentUser.getEmployee().getWorkLocation().getWorkLocationId() : null;
+
+		List<Holiday> holidayList = employeeWorkLocationId == null
+				? holidayDao.findAllByIsActiveTrueAndDate(currentDate)
+				: holidayDao.findAllByIsActiveTrueAndDateAndWorkLocationId(currentDate, employeeWorkLocationId);
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

No test covers the new behaviour. There is no `TimeServiceImplUnitTest` in `backend/src/test/java/com/skapp/community/timeplanner/service/impl/` (only `TimeAnalyticsServiceImplUnitTest` and `AttendanceConfigServiceImplUnitTest`), and `TimeControllerIntegrationTest` fully mocks `timeService.getActiveTimeSlot()`, so neither the work-location branch nor the null fallback is exercised. Because this decides whether an employee is allowed to clock in, it is worth at least a `@DataJpaTest`-style test on `HolidayRepositoryImpl.findAllByIsActiveTrueAndDateAndWorkLocationId` asserting the three cases: holiday scoped to the employee's location (returned), holiday scoped only to a different location (excluded), and holiday with an empty `workLocations` set (returned).

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/HolidayRepositoryImpl.java
```diff
@@ -102,6 +101,27 @@ public List<Holiday> findAllActiveHolidaysByWorkLocationId(Long workLocationId)
...
+		Root<Holiday> root = criteriaQuery.from(Holiday.class);
+
+		List<Predicate> predicates = new ArrayList<>();
+		predicates.add(criteriaBuilder.equal(root.get(Holiday_.isActive), true));
+		predicates.add(criteriaBuilder.equal(root.get(Holiday_.date), date));
+
+		Join<Holiday, WorkLocation> workLocationJoin = root.join(Holiday_.workLocations, JoinType.LEFT);
+		predicates.add(criteriaBuilder.or(
+				criteriaBuilder.equal(workLocationJoin.get(WorkLocation_.workLocationId), workLocationId),
+				criteriaBuilder.isEmpty(root.get(Holiday_.workLocations))));
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

This predicate makes "holiday has an empty `workLocations` set" mean "applies to every location", and this PR promotes that inference into the attendance clock-in gate (TimeServiceImpl.getAllActiveSlots). The problem is that an empty set can be reached accidentally, and the blast radius is now org-wide:

1. `shr_com_work_location_ppl_holiday` declares `work_location_id` FK with `ON DELETE CASCADE` (people-ddl-script-v1-create_shr_com_work_location_ppl_holiday_table.sql:10). `WorkLocationServiceImpl.deleteWorkLocation` (lines 105-118) calls `clearWorkLocationFromEmployees(id)` and then `workLocationDao.delete(workLocation)` — it cleans up employees but never holidays. So deleting a location silently strips it from every holiday row; any holiday that was scoped only to that location ends up with zero locations and is immediately reinterpreted by this query as a company-wide holiday, blocking clock-in for all employees at all sites.
2. `HolidayServiceImpl.resolveWorkLocations` (lines 547-553) skips names it cannot resolve; if none of the supplied names resolve, the holiday is persisted with an empty set. A typo in a bulk holiday upload therefore creates an org-wide clock-in block rather than a per-location one.

Fix options: (a) make "all locations" explicit rather than inferred — e.g. an `applies_to_all_locations` boolean column on `holiday` (with a Liquibase changeset) that this predicate checks instead of `isEmpty`; or (b) at minimum, handle holiday fallout in `deleteWorkLocation` (reassign/deactivate holidays that would be left with no locations, or block the delete with a `ModuleException` when such holidays exist) and make `resolveWorkLocations` throw a `ModuleException` with a message constant when a supplied location name cannot be resolved instead of silently dropping it.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeServiceImpl.java
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

This is a third inline copy of "resolve the holidays that apply to this employee, falling back to all holidays when the employee has no work location". `LeaveServiceImpl.getHolidaysForEmployee(Employee)` (LeaveServiceImpl:1037-1043) already does exactly this — and it guards one level deeper (`employee.getWorkLocation() != null && employee.getWorkLocation().getWorkLocationId() != null`) and falls back to `findAllByIsActiveTrue()`, so the two copies can drift. Extract a single resolver (e.g. a `getHolidaysForEmployee(Employee, LocalDate)` method on `HolidayService`, or a helper in `CommonModuleUtils`) and have both call sites use it, so the null-work-location policy is defined in one place.

Separately, this private helper only needs the work-location id, but it now takes the whole `User` and re-navigates `User -> Employee -> WorkLocation` twice on lines 1277-1278. Prefer passing `Long employeeWorkLocationId` (resolved by the caller, which already holds `currentUser`) or the already-resolved `List<Holiday>`; that keeps the helper unit-testable without a full `User` graph and avoids the duplicate dereference. Also note the method name `getAllActiveSlots` no longer describes what it does (it decides whether today is a holiday for this employee) — `resolveHolidayTimeSlotResponse` or similar would be more accurate.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeServiceImpl.java
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Missing paired change in the same module: `TimeAnalyticsServiceImpl` still resolves "is this date a holiday" with the single-result derived query `holidayDao.findByIsActiveTrueAndDate(date)` — at `TimeAnalyticsServiceImpl.getHoliday()` (line 336, feeds the clock-in summary response) and `isHolidayOrNoTimeConfig()` (line 427, drives expected-clock-in / late-arrival dashboard numbers). A single-entity return means Hibernate throws `NonUniqueResultException` (surfaced as `IncorrectResultSizeDataAccessException` -> 500 via `GlobalExceptionHandler`) as soon as two active holidays share a date. Per-work-location holidays make exactly that shape normal data (e.g. one holiday row per site on the same day), so these endpoints are a latent 500 that this PR's model makes routine. Change both to `List<Holiday>` (or reuse the new `findAllByIsActiveTrueAndDateAndWorkLocationId` with the employee's work location, which is also the semantically correct fix for the per-employee summary) and drop the `findByIsActiveTrueAndDate` single-result method from `HolidayDao`.

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/HolidayRepositoryImpl.java
```diff
@@ -102,6 +101,27 @@ public List<Holiday> findAllActiveHolidaysByWorkLocationId(Long workLocationId)
...
+		CriteriaQuery<Holiday> criteriaQuery = criteriaBuilder.createQuery(Holiday.class);
+		Root<Holiday> root = criteriaQuery.from(Holiday.class);
+
+		List<Predicate> predicates = new ArrayList<>();
+		predicates.add(criteriaBuilder.equal(root.get(Holiday_.isActive), true));
+		predicates.add(criteriaBuilder.equal(root.get(Holiday_.date), date));
+
+		Join<Holiday, WorkLocation> workLocationJoin = root.join(Holiday_.workLocations, JoinType.LEFT);
+		predicates.add(criteriaBuilder.or(
+				criteriaBuilder.equal(workLocationJoin.get(WorkLocation_.workLocationId), workLocationId),
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

The new method has no input contract for `workLocationId`, and both edge values fail silently rather than loudly:

- `workLocationId == null`: `criteriaBuilder.equal(path, null)` renders `= null`, which is UNKNOWN in SQL, so the OR collapses to just the `isEmpty` branch and the caller gets only globally-scoped holidays instead of all holidays. The current caller guards against null before calling (TimeServiceImpl:1280), but nothing on the interface signals that requirement — note this PR also removed the `lombok.NonNull` import from `HolidayRepository`, leaving no null annotation anywhere on it. Add `org.springframework.lang.NonNull` on the parameter (the convention used in `EmployeeRepositoryImpl:715`) or handle null explicitly by omitting the join predicate.
- `workLocationId == -1L`: `buildPredicates` (line 154) deliberately treats `PeopleConstants.HOLIDAY_ALL_WORK_LOCATIONS_ID` as "no location filter", and `HolidayServiceImpl.normalizeWorkLocations` (line 562) hands that same sentinel back to clients in responses. This new method — and `findAllActiveHolidaysByWorkLocationId` above it — ignore the sentinel, so a caller round-tripping a value from the API gets only globally-scoped holidays. Handle the sentinel consistently (skip the work-location predicate when it matches) so all three code paths agree.

backend/src/main/java/com/skapp/community/peopleplanner/repository/HolidayDao.java
```diff
@@ -13,8 +13,6 @@ public interface HolidayDao extends JpaRepository<Holiday, Long>, JpaSpecificati
 
 	List<Holiday> findAllByIsActiveTrueAndDate(LocalDate date);
```
ThinuwanW, line 14: 🤖 **Claude** · 🟣 **Nit**

While pruning this interface, the two semantically identical derived queries left behind are worth consolidating: `findAllByIsActiveTrueAndDate(LocalDate)` (line 14) and `findAllByDateAndIsActiveTrue(LocalDate)` (line 22) generate the same query under different names, and both are in active use (`HolidayServiceImpl:144` / `LeaveAnalyticsServiceImpl:437,1466,1846` use the first; `LeaveServiceImpl:555` uses the second). Keep one and update the callers — otherwise the next work-location-scoping change has to be applied twice, which is the same class of drift this PR is already exposed to.

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/HolidayRepositoryImpl.java
erandi-rootcode: Not Valid.

Empty workLocations = all locations is the app's existing convention, not a new inference — resolveWorkLocations saves the "all locations" option as an empty set (HolidayServiceImpl:543), and the isEmpty → all read semantic already predates this PR (HolidayRepositoryImpl:96, LeaveServiceImpl:1040). This PR just reuses it.

The delete-collapse is a business requirement — a holiday scoped only to a deleted location is meant to become an all-location holiday.

The bulk-typo case can't happen — validateWorkLocations rejects unknown names before save (HolidayServiceImpl:461,478), so resolveWorkLocations never gets an unresolvable name.
This PR narrows clock-in blocking (previously every holiday blocked everyone regardless of location), so it reduces the blast radius rather than expanding it.

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/HolidayRepositoryImpl.java
erandi-rootcode: Valid and Fixed.

Extracted buildWorkLocationPredicate(cb, root, id); all 3 call sites use it. Kept the 2 public methods separate (didn't collapse to nullable-date).

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeServiceImpl.java
erandi-rootcode: Valid and Fixed. 

Null branch now calls findAllByIsActiveTrueAndDateAndWorkLocationsIsEmpty (all-location holidays only)

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeServiceImpl.java
erandi-rootcode: Not Applicable in this scope.

backend/src/main/java/com/skapp/community/peopleplanner/repository/HolidayRepository.java
erandi-rootcode: Valid, Minor, and Fixed. 

Renamed findAllByIsActiveTrueAndDateAndWorkLocationId → findAllActiveHolidaysByDateAndWorkLocationId

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeServiceImpl.java
erandi-rootcode: Made it private as it has one in-class caller

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeServiceImpl.java
erandi-rootcode: Helper now takes Long employeeWorkLocationId (no User, no double-deref). 
Shared resolver with LeaveServiceImpl not merged (different fallback)

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeServiceImpl.java
erandi-rootcode: This was flagged and noted. but out of scope in this context. impact analysis is in progress.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeServiceImpl.java
erandi-rootcode: Integration test updated to assert HOLIDAY periodType passthrough. Unit/@DataJpaTest not added —  unit tests Not applicable.

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/HolidayRepositoryImpl.java
erandi-rootcode: Null is no longer important as the contract was changed.

backend/src/main/java/com/skapp/community/peopleplanner/repository/HolidayDao.java
erandi-rootcode: consolidated findAllByIsActiveTrueAndDate / findAllByDateAndIsActiveTrue.

shakila-kamalasena (approved):

akilarootcode (approved):

PR #1706, opened 2026-08-04
feat/crm table update

Deshan1100:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR migrates the CRM Company and Contact tables from the legacy `Table` component to the shared `TableView`/`TableV2` wrapper, and relaxes `TableView`'s filter contract so a filter button can be rendered in a disabled state without `filterContent`. The migration is largely faithful, but it silently drops several accessibility affordances that the old `Table` API carried (per-column aria labels, the infinite-scroll loading message, `tableName`/region aria label), removes the `hasNextPage && !isFetchingNextPage` guard around `fetchNextPage`, and loses the compile-time link between column keys and the entity type. The `TableView` filter contract change also introduces a silently-dead filter button state when `filterContent` is omitted without `isDisabled`.

Found **20** new issue(s): 🔴 3 important, 🟡 3 suggestion(s), 🟣 4 nit(s)

frontend/src/community/crm/components/organisms/CompanyTable/CompanyTable.tsx
```diff
@@ -61,158 +56,131 @@ export const CompanyTable: FC = () => {
...
+        description: hasActiveFilters
+          ? translateText(["table", "emptySearchState", "description"])
+          : translateText(["table", "emptyDataState", "description"])
+      }}
+      onRowClick={handleRowClick}
+      infiniteScroll={{
+        isEnabled: true,
+        height: "34.5rem",
+        hasMore: hasNextPage,
+        isFetchingNextPage,
```
ThinuwanW, line 158: 🤖 **Claude** · 🔴 **Important**

The infinite-scroll loading message is lost in this migration. The old `Table` received `infiniteScrollLoadingMessage={translateText(["table","infiniteScrollLoadingMessage"])}` ("Loading companies..." / "Loading contacts..."), which is both a visible affordance and a screen-reader announcement while the next page loads. The new `infiniteScroll` object passes no equivalent — and it cannot: `TableViewInfiniteScroll.loadingMessage` is declared in `TableView/types.ts` but `TableView.tsx` never forwards it to `TableV2` (it forwards only `height`, `hasMore`, `isFetchingNextPage`, `onLoadMore`, `scrollThreshold`). Fix by adding the pass-through in `TableView.tsx` (e.g. `infiniteScrollLoadingMessage={isInfiniteScroll ? infiniteScroll?.loadingMessage : undefined}`) and then passing `loadingMessage: translateText(["table","infiniteScrollLoadingMessage"])` here and in `ContactTable.tsx` (line 212). Otherwise the existing `crmModule.json` keys at `contacts.table.infiniteScrollLoadingMessage` and `companies.table.infiniteScrollLoadingMessage` become dead and should be deleted.

frontend/src/community/crm/components/organisms/CompanyTable/CompanyTable.tsx
```diff
@@ -61,158 +56,131 @@ export const CompanyTable: FC = () => {
...
+        onLoadMore: () => {
```
ThinuwanW, line 159: 🤖 **Claude** · 🔴 **Important**

The `loadMore` guard was dropped. The previous implementation was `if (hasNextPage && !isFetchingNextPage) await fetchNextPage()`; the new `onLoadMore: () => { void fetchNextPage(); }` fires unconditionally. React Query v5's `fetchNextPage` defaults to `cancelRefetch: true`, so repeated scroll-threshold events while a page request is already in flight will abort and re-issue the request, producing redundant/cancelled network calls and visible stutter. Restore the guard: `onLoadMore: () => { if (hasNextPage && !isFetchingNextPage) void fetchNextPage(); }`. Same issue in `ContactTable.tsx` line 217.

frontend/src/community/crm/components/organisms/CompanyTable/CompanyTable.tsx
```diff
@@ -61,158 +56,131 @@ export const CompanyTable: FC = () => {
     setSearchTerm(value);
   };
 
-  const columns: TableColumn<CrmCompany>[] = [
+  const columns = [
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Per-column accessibility labels were dropped. Every column previously carried `columnAriaLabel` (e.g. `table.columns.nameAriaLabel`, `phoneAriaLabel`, `tasksAriaLabel`, `pipelineAriaLabel`, `accountValueAriaLabel`), and the same for all six contact columns. The new `columns`/`tableHeaders` mapping only carries `id`, `label`, `width`, `align` (and `className` in ContactTable), so the header cells now expose only the uppercase display label. Either map the aria label onto the `GridHeader` (if `GridHeader` supports an aria/`ariaLabel` field in skapp-ui 1.1.1-beta.5) or confirm with design/a11y that the visible label is sufficient — and if so, delete the now-unused `*AriaLabel` keys from `crmModule.json` (`contacts.table.columns.*AriaLabel`, `companies.table.columns.*AriaLabel`) so they don't rot.

frontend/src/community/crm/components/organisms/CompanyTable/CompanyTable.tsx
```diff
@@ -61,158 +56,131 @@ export const CompanyTable: FC = () => {
...
-            "description"
-          ])
-        }}
-        onRowClick={(row) => {
-          setSelectedCompanyId(row.id);
-          openCrmSidePanel(CrmSidePanelTypes.COMPANY_SIDE_PANEL);
-        }}
-      />
-    </div>
+    <TableView
```
ThinuwanW, line 139: 🤖 **Claude** · 🟡 **Suggestion**

`tableName` and `ariaLabel.regionAriaLabel` are not passed. Every other `TableView` consumer in the repo passes `tableName` (LeaveRequests, LeaveTypesTable, ManagerLeaveRequest, ManagerTimesheetRequestTable, EmployeeTimesheetRequestTable), and most pass `ariaLabel.regionAriaLabel`. Add CRM entries to the `TableNames` enum in `~community/common/enums/Table.ts` (e.g. `CRM_COMPANIES = "Companies"`, `CRM_CONTACTS = "Contacts"`) and pass `tableName` here and in `ContactTable.tsx` line 197, plus a `regionAriaLabel` so the table region is announced.

frontend/src/community/crm/components/organisms/ContactTable/ContactTable.tsx
```diff
@@ -68,203 +67,176 @@ export const ContactTable: FC = () => {
...
+      value: String(company.id)
     }))
   ];
 
-  const columns: TableColumn<CrmContact>[] = [
+  const handleSearchChange = (event: ChangeEvent<HTMLInputElement>) => {
+    setSearchTerm(event.target.value);
+  };
+
+  const columns = [
```
ThinuwanW, line 84: 🤖 **Claude** · 🟡 **Suggestion**

Type-safety regression: `columns` was `TableColumn<CrmContact>[]` (so `key` was checked against the entity) and is now an untyped object-literal array whose `field` strings must line up by convention with the keys produced in `transformToTableRows()`. A typo in either place (e.g. `closedDealValue` vs `closedDealsValue`) compiles cleanly and renders a silently blank column. Tighten it by declaring a shared key union — e.g. `type ContactColumnKey = "name" | "email" | "contactNumber" | "closedDealValue" | "openTasksCount" | "owner";` — then type `columns` as `{ field: ContactColumnKey; headerName: string; width: string; align?: "right"; className?: string }[]` and the row objects as `GridRow & Record<ContactColumnKey, ReactNode>`. Same applies to `CompanyTable.tsx` line 59.

frontend/src/community/common/components/organisms/TableView/TableView.tsx
```diff
@@ -70,9 +75,10 @@ const TableView: FC<TableViewProps> = ({
             isFilterEnabled
               ? {
                   onClick: togglePopover,
+                  disabled: filter?.isDisabled,
```
ThinuwanW, line 78: 🤖 **Claude** · 🟡 **Suggestion**

Making `filterContent` optional introduces a silently dead button state. If a consumer passes a `filter` object with neither `filterContent` nor `isDisabled: true`, the toolbar renders an *enabled* filter button (`disabled: filter?.isDisabled` → `undefined`), but `togglePopover` early-returns and the `Popper` is never rendered — clicking does nothing, with no visual cue. Make the disabled state derive from interactivity: `disabled: filter?.isDisabled ?? !filter?.filterContent` (or better, model `TableViewFilter` as a discriminated union: either `{ isDisabled: true }` or `{ filterContent: ...; popoverId: string }`) so the type system rules out the dead configuration.

frontend/src/community/common/components/organisms/TableView/types.ts
```diff
@@ -32,11 +32,12 @@ export interface TableViewFilterContentArgs {
...
 export interface TableViewFilter {
   filterCount: number;
-  filterContent: (args: TableViewFilterContentArgs) => ReactNode;
+  filterContent?: (args: TableViewFilterContentArgs) => ReactNode;
+  isDisabled?: boolean;
   filterButtonAriaLabel?: string;
   popoverAriaLabel?: string;
   popoverAriaLabelledBy?: string;
-  popoverId: string;
+  popoverId?: string;
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`popoverId` was relaxed from required to optional to accommodate the disabled-filter case, which also lets a consumer that *does* provide `filterContent` omit it — `Popper` then renders with `id={undefined}`, breaking any `aria-controls`/`aria-labelledby` wiring that depends on a stable popover id. If you keep the flat interface, at least document that `popoverId` is required whenever `filterContent` is supplied; a discriminated union (see the `TableView.tsx` comment) would enforce it.

frontend/src/community/crm/components/organisms/ContactTable/ContactTable.tsx
```diff
@@ -68,203 +67,176 @@ export const ContactTable: FC = () => {
...
-          </div>
-        );
-      },
-      className: "text-right pr-[3%]",
-      width: "10%"
+      field: "closedDealValue",
+      headerName: translateText(["table", "columns", "closedValueHeader"]),
+      width: "10%",
+      align: "right" as const,
+      className: "pr-[3%]"
```
ThinuwanW, line 105: 🤖 **Claude** · 🟣 **Nit**

`className: "pr-[3%]"` / `"pl-[6%]"` is now mapped onto `GridHeader.className` (line 129), i.e. the header cell only. In the previous `TableColumn` API these classes sat alongside `text-right` on the column definition; if `TableColumn.className` was also applied to body cells, the data cells have lost that inset and will no longer line up with their headers. Please verify visually against the old rendering (and confirm `GridHeader` in skapp-ui 1.1.1-beta.5 actually accepts `className` — if it doesn't, this will fail typecheck).

frontend/src/community/crm/components/organisms/CompanyTable/CompanyTable.tsx
```diff
@@ -61,158 +56,131 @@ export const CompanyTable: FC = () => {
...
-      await fetchNextPage();
-    }
+  const tableHeaders: GridHeader[] = columns.map((col) => ({
+    id: col.field,
+    label: col.headerName,
+    width: col.width,
+    align: col.align
+  }));
+
+  const transformToTableRows = (): GridRow[] =>
```
ThinuwanW, line 89: 🤖 **Claude** · 🟣 **Nit**

`columns`, `tableHeaders` and `transformToTableRows()` are all rebuilt on every render, and `transformToTableRows()` is invoked inline in JSX (line 148), so every row's JSX subtree is recreated on each keystroke in the search box. Wrap `tableHeaders` and the row array in `useMemo` (deps: `translateText` / `companies`) and pass the memoized array. Same in `ContactTable.tsx` lines 124/132.

frontend/src/community/crm/components/organisms/ContactTable/ContactTable.tsx
```diff
@@ -1,15 +1,15 @@
 import {
-  Dropdown,
-  InputField,
   Label,
   ProjectTableSkeletonLoader,
-  SearchIcon,
-  Table,
-  TableColumn
+  SearchIcon
```
ThinuwanW, line 4: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

`CompanyTable` and `ContactTable` now share a near-identical `TableView` configuration block: the same `emptyState` ternary on `hasActiveFilters`, the same `infiniteScroll` shape, the same `toolbar.searchBar` props, the same `loader`. Consider extracting a small `CrmTableView` wrapper (or at least a `buildCrmEmptyState(translateText, hasActiveFilters)` helper) so the empty-state/search-bar conventions stay in sync as more CRM tables (DealsTable, SidePanelCompanyContacts) are migrated off the legacy `Table`.

frontend/src/community/common/components/organisms/TableView/TableView.tsx
```diff
@@ -31,6 +31,9 @@ const TableView: FC<TableViewProps> = ({
 
   const isOpen = anchorEl !== null;
   const isFilterEnabled = filter;
+  const isFilterInteractive = Boolean(
```
ThinuwanW, line 34: 🤖 **Claude** · 💬 **Note**

📁 **File-level observation**

`TableViewInfiniteScroll.loadingMessage` is declared in types.ts but TableView never forwards it to `TableV2` — the JSX only passes height/hasMore/isFetchingNextPage/onLoadMore/scrollThreshold. This matters because it makes the lost infinite-scroll loading message unfixable at the consumer level: even if CompanyTable/ContactTable pass `infiniteScroll.loadingMessage`, TableView silently drops it. Same for `TableViewProps.footer` — declared in the props interface but never destructured or rendered. Fix: forward `loadingMessage` to TableV2's corresponding prop (and render `footer`), or delete both from the type so the API doesn't advertise capabilities that don't exist.

frontend/src/community/common/components/organisms/TableView/TableView.tsx
```diff
@@ -108,7 +114,7 @@ const TableView: FC<TableViewProps> = ({
         onPageChange={isPaginated ? pagination?.onPageChange : undefined}
       />
 
-      {isFilterEnabled && (
+      {isFilterInteractive && (
```
ThinuwanW, line 117: 🤖 **Claude** · 💬 **Note**

Stale popover state when `isFilterInteractive` flips to false while the popover is open. The Popper is now gated on `isFilterInteractive`, but `anchorEl` is never cleared — if a consumer toggles `filter.isDisabled` to true (or swaps `filterContent` to undefined) while the panel is open, the Popper unmounts without `closePopover()` running, `anchorEl` stays set, and re-enabling the filter re-opens the panel spontaneously with no user action. The new early-return in `togglePopover` also means the user cannot close it in the meantime. Fix: add `useEffect(() => { if (!isFilterInteractive) setAnchorEl(null); }, [isFilterInteractive]);`.

frontend/src/community/crm/components/organisms/ContactTable/ContactTable.tsx
```diff
@@ -68,203 +67,176 @@ export const ContactTable: FC = () => {
...
+          placeholder: translateText(["table", "search"]),
+          "aria-label": translateText(["table", "search"]),
+          ariaLabelClearButton: translateText(["table", "clearButtonAriaLabel"])
+        },
+        dropdown: {
+          id: "crm-contacts-company-filter",
+          options: companyOptions,
+          value: String(selectedCompany ?? ""),
+          onChange: (value) =>
+            setSelectedCompany(value ? Number(value) : ALL_COMPANIES),
```
ThinuwanW, line 230: 🤖 **Claude** · 💬 **Note**

Fragile string↔number coercion in the company filter. `setSelectedCompany(value ? Number(value) : ALL_COMPANIES)` uses truthiness to detect the "All companies" sentinel, and `ALL_COMPANIES` is literally `undefined`, so the expression reads as if a real sentinel exists when it does not. Any falsy option value collapses to "all", and a non-numeric value would silently produce `NaN` and be sent to `useGetContactMetrics` as the company filter. Also note the first option's `id` changed from `ALL_COMPANIES` to `""`, which is now an empty-string element id/key. Fix: introduce an explicit named constant for the sentinel option value (e.g. `ALL_COMPANIES_OPTION_VALUE = "all"`), compare against it directly rather than by truthiness, and guard the parse: `const parsed = Number(value); setSelectedCompany(Number.isNaN(parsed) ? ALL_COMPANIES : parsed)`.

frontend/src/community/crm/components/organisms/CompanyTable/CompanyTable.tsx
```diff
@@ -61,158 +56,131 @@ export const CompanyTable: FC = () => {
...
-          openCrmSidePanel(CrmSidePanelTypes.COMPANY_SIDE_PANEL);
-        }}
-      />
-    </div>
+    <TableView
+      headers={tableHeaders}
+      rows={transformToTableRows()}
+      isLoading={isLoading}
+      loader={<ProjectTableSkeletonLoader rowCount={8} />}
+      emptyState={{
```
ThinuwanW, line 144: 🤖 **Claude** · 💬 **Note**

No error state. `useGetCompanyMetrics` / `useGetContactMetrics` expose `isError`, but neither table reads it, so a failed request falls through to `emptyState` and the user sees "No companies available. Create a company to get started." — actively misleading, since it suggests the dataset is empty rather than that the fetch failed. The guidelines require loading AND error states to be rendered. Fix: destructure `isError`/`error` from the hook and either render a distinct error state (or surface a toast via `useToast`) instead of the empty state when the query fails. Applies identically to ContactTable.

frontend/src/community/common/components/organisms/TableView/TableView.tsx
ThinuwanW, line 34: 🤖 **Claude** · 💬 **Note**

No test coverage for the new conditional logic. There is no `TableView.test.tsx` anywhere in the repo, and this PR adds three new branches to a shared organism used by seven consumers: `isFilterInteractive`, the `togglePopover` early return, and the disabled/undefined aria attributes. Add a `TableView.test.tsx` (with the `MockTheme` wrapper, `@testing-library/react`, `screen.getByRole`) covering: filter with `filterContent` renders an enabled button that opens the popover; `isDisabled: true` renders a disabled button and no popover; `filter` with neither `filterContent` nor `isDisabled` (see the first-pass dead-button finding).

frontend/src/community/common/components/organisms/TableView/TableView.tsx
ThinuwanW, line 78: 🤖 **Claude** · 💬 **Note**

The filter count badge is not reconciled with the new `isDisabled` flag. The badge block still renders `count: filter?.filterCount ?? 0, show: (filter?.filterCount ?? 0) > 0` regardless of `isDisabled`, so a consumer that disables the filter while filters are applied (a plausible case: filter temporarily unavailable during a refetch) renders a count badge on a button the user cannot interact with. Either suppress the badge when `isDisabled` (`show: !filter?.isDisabled && (filter?.filterCount ?? 0) > 0`) or document on `TableViewFilter` that `isDisabled` implies `filterCount: 0`.

frontend/src/community/crm/components/organisms/ContactTable/ContactTable.tsx
```diff
@@ -68,203 +67,176 @@ export const ContactTable: FC = () => {
...
-          className="rounded-full"
-          menuWidth="content"
-          options={companyOptions}
-          value={selectedCompany}
-          variant="secondary"
-          onChange={(value) => setSelectedCompany(value)}
+  const transformToTableRows = (): GridRow[] =>
+    (contacts ?? []).map((contact) => ({
+      id: contact.id,
+      ariaLabel: contact.name,
```
ThinuwanW, line 131: 🤖 **Claude** · 💬 **Note**

Row-level `ariaLabel` is just the raw entity name. Rows are clickable (`onRowClick` opens the side panel), and every other TableView consumer builds a descriptive row label via a dedicated helper in `utils/accessibilityUtils` (e.g. `generateMyLeaveRequestAriaLabel`, `generateManagerLeaveRequestAriaLabel`). Here a screen-reader user activating a row hears only "Acme Corp" with no indication of what activation does or what the row contains. Add a `generateCrmContactAriaLabel` / `generateCrmCompanyAriaLabel` helper under `~community/crm/utils/accessibilityUtils` that composes name + company + open tasks + the action, and use it for `ariaLabel`. Same in CompanyTable.

frontend/src/community/crm/components/organisms/CompanyTable/CompanyTable.tsx
```diff
@@ -61,158 +56,131 @@ export const CompanyTable: FC = () => {
...
+        onLoadMore: () => {
+          void fetchNextPage();
+        }
+      }}
+      toolbar={{
+        searchBar: {
+          value: searchTerm,
+          onChange: handleSearchChange,
+          placeholder: translateText(["table", "search"]),
+          "aria-label": translateText(["table", "search"]),
```
ThinuwanW: 🤖 **Claude** · 💬 **Note**

The search input's `aria-label` reuses the placeholder key `table.search` ("Search company" / "Search contact"), so the accessible name and the visible placeholder are the identical string — redundant, and it means no dedicated a11y copy exists for this control. The codebase convention is dedicated aria namespaces (`leaveAria`, `attendanceAria`, `commonAria`, loaded via `useTranslator("leaveAria", ...)`); the CRM module has none. Add a `crmAria` namespace (or at minimum distinct `...AriaLabel` keys under `companies.table` / `contacts.table`) and use it for the search field, dropdown and filter button labels. Applies to ContactTable too.

frontend/src/community/crm/components/organisms/ContactTable/ContactTable.tsx
ThinuwanW, line 4: 🤖 **Claude** · 💬 **Note**

📁 **File-level observation**

Missing paired change: orphaned i18n keys. This PR removes the last references to `companies.table.columns.{name,phone,tasks,pipeline,accountValue}AriaLabel`, `contacts.table.columns.{name,email,phone,closedValue,tasks,contactOwner}AriaLabel`, and `{companies,contacts}.table.infiniteScrollLoadingMessage` in `src/community/common/assets/languages/english/crmModule.json`, but the keys are left behind. Either restore their usage (which is what the dropped column aria labels and infinite-scroll message findings call for) or delete the now-dead keys so the translation file doesn't accumulate unused entries for translators.

frontend/src/community/common/components/organisms/TableView/TableView.tsx
```diff
@@ -31,6 +31,9 @@ const TableView: FC<TableViewProps> = ({
 
   const isOpen = anchorEl !== null;
   const isFilterEnabled = filter;
```
ThinuwanW, line 33: 🤖 **Claude** · 💬 **Note**

`isFilterEnabled = filter` holds the filter object rather than a boolean, so `isToolbarVisible` is typed `boolean | TableViewFilter | undefined` and the `isFilterEnabled ? ... : undefined` ternary at line 75 relies on object truthiness. The PR adds a correctly `Boolean()`-wrapped sibling (`isFilterInteractive`) right next to it, which makes the inconsistency stand out. Change to `const isFilterEnabled = Boolean(filter);` so both flags are genuine booleans and `isToolbarVisible` is `boolean`.

Deshan1100: need this bool to check whether filter need to display or not

frontend/src/community/common/components/organisms/TableView/TableView.tsx
Deshan1100: no need loading handle through the table compoent

Deshan1100: no need test

frontend/src/community/common/components/organisms/TableView/TableView.tsx
Deshan1100: need this handled in component level

Deshan1100: no need count if disable

frontend/src/community/common/components/organisms/TableView/TableView.tsx
Deshan1100: need to check filter enable

frontend/src/community/common/components/organisms/TableView/types.ts
Deshan1100: updated

frontend/src/community/crm/components/organisms/CompanyTable/CompanyTable.tsx
Deshan1100: updated

frontend/src/community/crm/components/organisms/CompanyTable/CompanyTable.tsx
Deshan1100: updated

frontend/src/community/crm/components/organisms/CompanyTable/CompanyTable.tsx
Deshan1100: no need kept as previous

frontend/src/community/crm/components/organisms/CompanyTable/CompanyTable.tsx
Deshan1100: now loading handled in component level

frontend/src/community/crm/components/organisms/CompanyTable/CompanyTable.tsx
Deshan1100: checked works fine

frontend/src/community/crm/components/organisms/CompanyTable/CompanyTable.tsx
Deshan1100: kept as previous

frontend/src/community/crm/components/organisms/ContactTable/ContactTable.tsx
Deshan1100: kept ads previous

frontend/src/community/crm/components/organisms/ContactTable/ContactTable.tsx
Deshan1100: kept as previous

frontend/src/community/crm/components/organisms/ContactTable/ContactTable.tsx
Deshan1100: checked

frontend/src/community/crm/components/organisms/ContactTable/ContactTable.tsx
Deshan1100: kept as previous

frontend/src/community/crm/components/organisms/ContactTable/ContactTable.tsx
Deshan1100: kept as previous

frontend/src/community/crm/components/organisms/CompanyTable/CompanyTable.tsx
Deshan1100: out of scope kept as previous

frontend/src/community/crm/components/organisms/ContactTable/ContactTable.tsx
Deshan1100: kept as previous

Deshan1100: kept as previous

anusham-rc (dismissed):

anusham-rc (dismissed):

shakila-kamalasena (dismissed):

frontend/src/community/common/components/organisms/TableView/TableView.tsx
```diff
@@ -117,7 +123,7 @@ const TableView: FC<TableViewProps> = ({
           containerClassName="rounded-4 shadow-lg"
           {...popperProps}
         >
-          {filter?.filterContent({ onClose: closePopover })}
+          {filter?.filterContent?.({ close: closePopover })}
```
akilarootcode: rename onClose

akilarootcode: remove optional

frontend/src/community/crm/components/organisms/ContactTable/ContactTable.tsx
```diff
@@ -66,205 +61,163 @@ export const ContactTable: FC = () => {
...
+          placeholder: translateText(["table", "search"]),
+          "aria-label": translateText(["table", "search"]),
+          ariaLabelClearButton: translateText(["table", "clearButtonAriaLabel"])
+        },
+        dropdown: {
+          id: "crm-contacts-company-filter",
+          options: companyOptions,
+          value: String(selectedCompany ?? ""),
+          onChange: (value) =>
+            setSelectedCompany(value ? Number(value) : ALL_COMPANIES),
```
akilarootcode: remove fallback  : ALL_COMPANIES

frontend/src/community/crm/components/organisms/ContactTable/ContactTable.tsx
```diff
@@ -66,205 +61,163 @@ export const ContactTable: FC = () => {
...
-  const hasActiveFilters =
+  const isEmptyFilterState =
     debouncedSearch.trim() !== "" || selectedCompany !== undefined;
-  const emptyStateType = hasActiveFilters
-    ? EmptyStateTypeEnum.NO_SEARCH_RESULTS
-    : EmptyStateTypeEnum.NO_DATA;
 
   const companyOptions = [
     {
-      id: ALL_COMPANIES,
```
akilarootcode, line 77: check why all companies not showing with ID

frontend/src/community/crm/components/organisms/CrmTableView/CrmTableView.tsx
```diff
@@ -0,0 +1,70 @@
...
+  scrollHeight: string;
+  hasNextPage?: boolean;
+  isFetchingNextPage: boolean;
+  onLoadMore: () => void;
+  onRowClick: (row: GridRow) => void;
+  toolbar: TableViewProps["toolbar"];
+  filter?: TableViewFilter;
+}
+
+const CrmTableView: FC<Props> = ({
```
akilarootcode: remove this and call directly

akilarootcode (changes requested):

frontend/src/community/common/components/organisms/TableView/TableView.tsx
Deshan1100: updated

Deshan1100: updated

frontend/src/community/crm/components/organisms/CrmTableView/CrmTableView.tsx
Deshan1100: updated

frontend/src/community/crm/components/organisms/ContactTable/ContactTable.tsx
Deshan1100: updated

akilarootcode (approved):

PR #1707, opened 2026-08-04
fix/invoice all invoice table update with pageSize

Deshan1100:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/frontend/src/community/common/components/organisms/TableView/TableView.tsx
+++ b/frontend/src/community/common/components/organisms/TableView/TableView.tsx
@@ -109,14 +109,8 @@ const TableView: FC<TableViewProps> = ({
         totalPages={isPaginated ? pagination?.totalPages : undefined}
         currentPage={isPaginated ? pagination?.currentPage : undefined}
         onPageChange={isPaginated ? pagination?.onPageChange : undefined}
-        pageSize={isPaginated ? pagination?.pageSize : undefined}
-        pageSizeOptions={isPaginated ? pagination?.pageSizeOptions : undefined}
-        onPageSizeChange={
-          isPaginated ? pagination?.onPageSizeChange : undefined
-        }
-        pageSizeLabel={isPaginated ? pagination?.pageSizeLabel : undefined}
-        isPageSizeDisabled={
-          isPaginated ? pagination?.isPageSizeDisabled : undefined
+        pageSizeSelector={
+          isPaginated ? pagination?.pageSizeSelector : undefined
         }
       />
 
--- a/frontend/src/community/common/components/organisms/TableView/TableView.tsx
+++ b/frontend/src/community/common/components/organisms/TableView/TableView.tsx
@@ -12,6 +12,7 @@ const TableView: FC<TableViewProps> = ({
   heading,
   tableName,
   ariaLabel,
+  variant,
   headers,
   rows,
   isLoading,
@@ -56,7 +57,8 @@ const TableView: FC<TableViewProps> = ({
     paginationAriaLabel: ariaLabel?.paginationAriaLabel,
     previousPageLabel: ariaLabel?.previousPageLabel,
     nextPageLabel: ariaLabel?.nextPageLabel,
-    getPageAriaLabel: ariaLabel?.getPageAriaLabel
+    getPageAriaLabel: ariaLabel?.getPageAriaLabel,
+    pageSizeAriaLabel: ariaLabel?.pageSizeAriaLabel
   };
 
   return (
@@ -86,6 +88,7 @@ const TableView: FC<TableViewProps> = ({
       <TableV2
         tableName={tableName}
         ariaLabel={mergedAriaLabel}
+        variant={variant}
         headers={headers}
         rows={rows}
         isLoading={isLoading}
@@ -106,6 +109,15 @@ const TableView: FC<TableViewProps> = ({
         totalPages={isPaginated ? pagination?.totalPages : undefined}
         currentPage={isPaginated ? pagination?.currentPage : undefined}
         onPageChange={isPaginated ? pagination?.onPageChange : undefined}
+        pageSize={isPaginated ? pagination?.pageSize : undefined}
+        pageSizeOptions={isPaginated ? pagination?.pageSizeOptions : undefined}
+        onPageSizeChange={
+          isPaginated ? pagination?.onPageSizeChange : undefined
+        }
+        pageSizeLabel={isPaginated ? pagination?.pageSizeLabel : undefined}
+        isPageSizeDisabled={
+          isPaginated ? pagination?.isPageSizeDisabled : undefined
+        }
       />
 
       {isFilterEnabled && (
--- a/frontend/src/community/common/components/organisms/TableView/types.ts
+++ b/frontend/src/community/common/components/organisms/TableView/types.ts
@@ -11,7 +11,8 @@ import type { ReactNode } from "react";
 export type {
   GridHeader,
   GridRow,
-  PageSizeSelectorProps
+  PageSizeSelectorProps,
+  TableV2Variant
 } from "@rootcodelabs/skapp-ui";
 
 /** Static pagination — 0-based, passed straight through to TableV2. */
@@ -55,7 +56,6 @@ export interface TableViewAriaLabels {
   pageSizeAriaLabel?: string;
 }
 
-/** Footer region rendered below the table. */
 export interface TableViewFooter {
   left?: ReactNode;
   right?: ReactNode;
--- a/frontend/src/community/common/components/organisms/TableView/types.ts
+++ b/frontend/src/community/common/components/organisms/TableView/types.ts
@@ -1,24 +1,25 @@
 import type {
   GridHeader,
   GridRow,
+  PageSizeSelectorProps,
   TableToolBarProps,
   TableV2Props,
   TableV2Variant
 } from "@rootcodelabs/skapp-ui";
 import type { ReactNode } from "react";
 
-export type { GridHeader, GridRow } from "@rootcodelabs/skapp-ui";
+export type {
+  GridHeader,
+  GridRow,
+  PageSizeSelectorProps
+} from "@rootcodelabs/skapp-ui";
 
 /** Static pagination — 0-based, passed straight through to TableV2. */
 export interface TableViewPagination {
   totalPages?: number;
   currentPage?: number;
   onPageChange?: (page: number) => void;
-  pageSize?: number;
-  pageSizeOptions?: number[];
-  onPageSizeChange?: (pageSize: number) => void;
-  pageSizeLabel?: string;
-  isPageSizeDisabled?: boolean;
+  pageSizeSelector?: PageSizeSelectorProps;
 }
 
 /** Infinite scroll — requires a fixed `height` on the scroll container. */
--- a/frontend/src/community/common/components/organisms/TableView/types.ts
+++ b/frontend/src/community/common/components/organisms/TableView/types.ts
@@ -2,7 +2,8 @@ import type {
   GridHeader,
   GridRow,
   TableToolBarProps,
-  TableV2Props
+  TableV2Props,
+  TableV2Variant
 } from "@rootcodelabs/skapp-ui";
 import type { ReactNode } from "react";
 
@@ -13,6 +14,11 @@ export interface TableViewPagination {
   totalPages?: number;
   currentPage?: number;
   onPageChange?: (page: number) => void;
+  pageSize?: number;
+  pageSizeOptions?: number[];
+  onPageSizeChange?: (pageSize: number) => void;
+  pageSizeLabel?: string;
+  isPageSizeDisabled?: boolean;
 }
 
 /** Infinite scroll — requires a fixed `height` on the scroll container. */
@@ -45,6 +51,7 @@ export interface TableViewAriaLabels {
   previousPageLabel?: string;
   nextPageLabel?: string;
   getPageAriaLabel?: (page: number) => string;
+  pageSizeAriaLabel?: string;
 }
 
 /** Footer region rendered below the table. */
@@ -57,6 +64,7 @@ export interface TableViewProps {
   heading?: ReactNode;
   tableName?: string;
   ariaLabel?: TableViewAriaLabels;
+  variant?: TableV2Variant;
   headers: GridHeader[];
   rows: GridRow[];
   isLoading?: boolean;


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR threads a `variant` prop and five page-size pagination props (plus a `pageSizeAriaLabel`) through the shared `TableView` wrapper down to `TableV2` from `@rootcodelabs/skapp-ui`. The change is additive and correctly gated behind the existing `isPaginated` guard, so the three existing paginated consumers (LeaveRequests, ManagerLeaveRequest, EmployeeTimesheetRequestTable) and the infinite-scroll path are unaffected — no behavioral regression found. Remaining concerns are type-contract fidelity against the pinned UI library version, hand-rolled types that can silently drift from `TableV2Props`, and growing repetition in the prop-forwarding block.

Found **15** new issue(s): 🔴 3 important, 🟡 8 suggestion(s), 🟣 4 nit(s)

frontend/src/community/common/components/organisms/TableView/types.ts
```diff
@@ -2,7 +2,8 @@ import type {
   GridHeader,
   GridRow,
   TableToolBarProps,
-  TableV2Props
+  TableV2Props,
+  TableV2Variant
```
ThinuwanW, line 7: 🤖 **Claude** · 🔴 **Important**

This PR imports the new type `TableV2Variant` and (via TableView.tsx) passes five new props to `TableV2` (`pageSize`, `pageSizeOptions`, `onPageSizeChange`, `pageSizeLabel`, `isPageSizeDisabled`) plus a new `ariaLabel.pageSizeAriaLabel` key — but `frontend/package.json` / `package-lock.json` still pin `@rootcodelabs/skapp-ui` at `1.1.1-beta.5` and the PR does not bump it. `skipLibCheck` in tsconfig does not protect call-site usage, so if `TableV2Variant` is not exported by that exact version the `import type` fails and `tsc`/`next build` breaks. I could not verify locally (the package is on a private GitHub registry and node_modules is not installed). Please confirm the pinned version exports `TableV2Variant` and that `TableV2Props` declares all five page-size props; if page-size support landed in a later beta, bump the dependency and lockfile in this PR so the build is reproducible.

frontend/src/community/common/components/organisms/TableView/TableView.tsx
```diff
@@ -56,7 +57,8 @@ const TableView: FC<TableViewProps> = ({
     paginationAriaLabel: ariaLabel?.paginationAriaLabel,
     previousPageLabel: ariaLabel?.previousPageLabel,
     nextPageLabel: ariaLabel?.nextPageLabel,
-    getPageAriaLabel: ariaLabel?.getPageAriaLabel
+    getPageAriaLabel: ariaLabel?.getPageAriaLabel,
+    pageSizeAriaLabel: ariaLabel?.pageSizeAriaLabel
```
ThinuwanW, line 61: 🤖 **Claude** · 🟡 **Suggestion**

`mergedAriaLabel` is an untyped object literal assigned to a `const` before being passed to `TableV2`, so TypeScript's excess-property check does not apply — if `TableV2`'s `ariaLabel` type does not have `pageSizeAriaLabel` (or renames it, e.g. to `pageSizeSelectAriaLabel`), the new key is silently dropped and the page-size control ships with no accessible name, with no compile error to catch it. Annotate the object so the compiler validates the key set: `const mergedAriaLabel: TableV2Props["ariaLabel"] = { ... }` (`TableV2Props` is already imported in ./types and can be imported here too). This applies to the whole aria-label merge, not just the new key.

frontend/src/community/common/components/organisms/TableView/types.ts
```diff
@@ -13,6 +14,11 @@ export interface TableViewPagination {
   totalPages?: number;
   currentPage?: number;
   onPageChange?: (page: number) => void;
+  pageSize?: number;
+  pageSizeOptions?: number[];
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The five new page-size fields are hand-written (`pageSize?: number`, `pageSizeOptions?: number[]`, `onPageSizeChange?: (pageSize: number) => void`, ...) rather than derived from the library contract, which diverges from the pattern already used two lines below in this same file (`emptyState?: TableV2Props["emptyState"]`). Hand-rolled duplicates drift silently on the next skapp-ui upgrade — and `number[]` in particular will fail to assign if the library declares `readonly number[]` or a narrower literal union. Prefer deriving each field, e.g. `pageSize?: TableV2Props["pageSize"]; pageSizeOptions?: TableV2Props["pageSizeOptions"]; onPageSizeChange?: TableV2Props["onPageSizeChange"];` — same for `pageSizeLabel` / `isPageSizeDisabled` and for `pageSizeAriaLabel` on `TableViewAriaLabels`.

frontend/src/community/common/components/organisms/TableView/TableView.tsx
```diff
@@ -106,6 +109,15 @@ const TableView: FC<TableViewProps> = ({
         totalPages={isPaginated ? pagination?.totalPages : undefined}
         currentPage={isPaginated ? pagination?.currentPage : undefined}
         onPageChange={isPaginated ? pagination?.onPageChange : undefined}
+        pageSize={isPaginated ? pagination?.pageSize : undefined}
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The `isPaginated ? pagination?.x : undefined` idiom is now repeated eight times (five added here), and each new pagination prop has to remember the gate — an easy thing to miss on the next addition, which would then leak pagination props into infinite-scroll mode. Also, inside the truthy branch `pagination` is guaranteed defined (`isPaginated = !isInfiniteScroll && pagination`), so the `?.` is dead. Consider building the group once and spreading it: `const paginationProps = isPaginated ? { totalPages: pagination.totalPages, currentPage: pagination.currentPage, onPageChange: pagination.onPageChange, pageSize: pagination.pageSize, pageSizeOptions: pagination.pageSizeOptions, onPageSizeChange: pagination.onPageSizeChange, pageSizeLabel: pagination.pageSizeLabel, isPageSizeDisabled: pagination.isPageSizeDisabled } : {};` then `<TableV2 ... {...paginationProps} />` (mirroring this for the infinite-scroll block would be consistent).

frontend/src/community/common/components/organisms/TableView/types.ts
```diff
@@ -13,6 +14,11 @@ export interface TableViewPagination {
   totalPages?: number;
   currentPage?: number;
   onPageChange?: (page: number) => void;
+  pageSize?: number;
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

All five page-size fields are independently optional, so invalid combinations type-check cleanly: passing `onPageSizeChange` without `pageSizeOptions`/`pageSize` compiles but renders a selector with no options, and passing `onPageSizeChange` without `ariaLabel.pageSizeAriaLabel`/`pageSizeLabel` yields an interactive control with no accessible name. Since the page-size selector is only meaningful as a unit, consider grouping it so the compiler enforces completeness, e.g. `pageSizeControl?: { value: number; options: number[]; onChange: (pageSize: number) => void; label?: string; ariaLabel?: string; isDisabled?: boolean }`, and forward that group in TableView.tsx. If the flat shape is preferred for symmetry with TableV2's own API, that's a reasonable call — but then it's worth a JSDoc note on the interface stating the four fields must be supplied together.

frontend/src/community/common/components/organisms/TableView/types.ts
```diff
@@ -45,6 +51,7 @@ export interface TableViewAriaLabels {
   previousPageLabel?: string;
   nextPageLabel?: string;
   getPageAriaLabel?: (page: number) => string;
+  pageSizeAriaLabel?: string;
 }
 
 /** Footer region rendered below the table. */
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Pre-existing dead code in the interface this PR extends: `footer?: TableViewFooter` (and the `TableViewFooter` type itself) plus `TableViewInfiniteScroll.loadingMessage` are declared here but never read in TableView.tsx — I grepped the repo and no consumer passes `footer` to TableView, and `loadingMessage` appears only in this file. Since you are already reshaping this contract, either wire them into the rendered output or delete them, so the public props of the shared wrapper stay an accurate description of what it actually does.

frontend/src/community/common/components/organisms/TableView/types.ts
ThinuwanW, line 7: 🤖 **Claude** · 🔴 **Important**

This change depends on design-system capabilities that are not accounted for in the dependency pin. It imports a new type `TableV2Variant` and forwards six new props to `TableV2` (`variant`, `pageSize`, `pageSizeOptions`, `onPageSizeChange`, `pageSizeLabel`, `isPageSizeDisabled`), but neither `frontend/package.json` nor `frontend/package-lock.json` is touched — both still pin `@rootcodelabs/skapp-ui` at the exact version `1.1.1-beta.5` (no caret, so no float). Please confirm that this exact published version exports `TableV2Variant` and that `TableV2` accepts all six props. If the page-size selector / list variant landed in a later beta, the build fails on the missing type export, or (worse) the props are silently ignored at runtime while still type-checking through the widened `mergedAriaLabel` object. Fix: bump the pin (and regenerate the lockfile) in the same PR so the wrapper and the library ship together.

frontend/src/community/common/components/organisms/TableView/types.ts
```diff
@@ -45,6 +51,7 @@ export interface TableViewAriaLabels {
   previousPageLabel?: string;
   nextPageLabel?: string;
   getPageAriaLabel?: (page: number) => string;
+  pageSizeAriaLabel?: string;
```
ThinuwanW, line 56: 🤖 **Claude** · 🔴 **Important**

The new `pageSizeAriaLabel` is dead on arrival for its only consumer. `frontend/src/enterprise/invoice/components/molecules/InvoiceTable/InvoiceTable.tsx` populates it with `translateAria(["invoiceTable.itemsPerPage"])` where `translateAria = useTranslator("invoiceModuleAria", "components")`. `useTranslator` builds the key `invoiceModuleAria.components.invoiceTable.itemsPerPage`, but the namespace `invoiceModuleAria` is not registered in `frontend/src/enterprise/common/assets/languages/english/english.ts` (only `commonEnterpriseAria`, `leaveEnterpriseAria`, `onboardingEnterpriseAria`, `settingEnterpriseAria`, `eSignatureModuleAria`, `pmModuleAria` are spread into the resource bundle), and no `invoiceModuleAria` JSON file exists anywhere under `frontend/src`. i18next therefore returns the key verbatim, so the page-size dropdown's accessible name renders as the literal string `invoiceModuleAria.components.invoiceTable.itemsPerPage` for screen-reader users. Fix: add an `aria/invoiceModule.json` with an `components.invoiceTable.itemsPerPage` entry and register it as `invoiceModuleAria` in `english.ts` (mirroring how `eSignatureModuleAria` is wired), or point the consumer at an existing registered namespace. Note the same broken namespace already silently affects `InvoiceTable.rowDescription` and the filter aria labels in that file.

frontend/src/community/common/components/organisms/TableView/TableView.tsx
```diff
@@ -106,6 +109,15 @@ const TableView: FC<TableViewProps> = ({
...
+        pageSizeOptions={isPaginated ? pagination?.pageSizeOptions : undefined}
+        onPageSizeChange={
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`onPageSizeChange` is forwarded raw, so the wrapper gives no guarantee that the current page is reset when the page size changes. With `currentPage` documented as 0-based and controlled by the caller, a user sitting on page 9 who switches from 10 to 50 rows keeps `currentPage: 9`, which is now out of range — the consumer fires a backend request for a page that no longer exists and the table renders empty until the user manually pages back. The invoice consumer works around this by hand (`onPageSizeChange: (size) => { setSize(size); setPage(0); }`), which means every future consumer has to remember the same dance. Fix: wrap the handler in `TableView` so the reset is part of the contract, e.g. pass `onPageSizeChange={isPaginated ? (size) => { pagination?.onPageSizeChange?.(size); pagination?.onPageChange?.(0); } : undefined}` — and at minimum document the reset expectation in the `TableViewPagination` doc comment.

frontend/src/community/common/components/organisms/TableView/TableView.tsx
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Adding a user-controllable `pageSize` makes the existing `skeletonRows = 4` default wrong for any paginated table: pick 50 rows per page and the loading state still shows 4 skeleton rows, so the table visibly collapses and then jumps on every refetch. The invoice consumer patches this by passing `skeletonRows={itemsPerPage}` separately, i.e. the same value has to be threaded in twice. Fix: derive the default inside the wrapper — change the destructured default to something like `skeletonRows` falling back to `pagination?.pageSize ?? 4` (computed after destructuring, since a default parameter cannot reference a sibling), so callers only supply the page size once.

frontend/src/community/common/components/organisms/TableView/TableView.tsx
ThinuwanW, line 61: 🤖 **Claude** · 🟡 **Suggestion**

The page-size feature is configured across two unrelated prop groups with inconsistent gating. `pageSizeLabel` (the visible label) lives on `pagination`, while `pageSizeAriaLabel` (its accessible name) lives on `ariaLabel` — a caller enabling the selector has to touch two separate objects for one control. Worse, the gating differs: every other page-size prop is guarded by `isPaginated ? … : undefined`, but `pageSizeAriaLabel` is merged unconditionally, so when `infiniteScroll.isEnabled` is true (which suppresses all pagination props) the page-size aria label still reaches `TableV2` describing a control that is not rendered. Fix: either move `pageSizeAriaLabel` alongside `pageSizeLabel` on `TableViewPagination`, or gate it the same way the rest are (`isPaginated ? ariaLabel?.pageSizeAriaLabel : undefined`) so the aria contract matches what is actually rendered.

frontend/src/community/common/components/organisms/TableView/types.ts
```diff
@@ -2,7 +2,8 @@ import type {
   GridHeader,
   GridRow,
   TableToolBarProps,
-  TableV2Props
+  TableV2Props,
```
ThinuwanW, line 6: 🤖 **Claude** · 🟡 **Suggestion**

`types.ts` deliberately re-exports the design-system types consumers need (`export type { GridHeader, GridRow } from "@rootcodelabs/skapp-ui"`) so feature modules never import from the library directly — see `InvoiceTable.tsx` and `CustomerTable.tsx`, which both pull `GridHeader`/`GridRow` from this module. `TableV2Variant` is imported here for the new `variant` prop but is not added to that re-export list, so any consumer that wants to type its own `variant` pass-through prop must break the abstraction and import from `@rootcodelabs/skapp-ui` itself. Fix: add `TableV2Variant` to the existing `export type { … }` line. (Related: this folder still has no `index.tsx` barrel, which the module convention calls for — consumers deep-import `.../TableView/TableView`.)

frontend/src/community/common/components/organisms/TableView/TableView.tsx
```diff
@@ -56,7 +57,8 @@ const TableView: FC<TableViewProps> = ({
     paginationAriaLabel: ariaLabel?.paginationAriaLabel,
     previousPageLabel: ariaLabel?.previousPageLabel,
     nextPageLabel: ariaLabel?.nextPageLabel,
-    getPageAriaLabel: ariaLabel?.getPageAriaLabel
+    getPageAriaLabel: ariaLabel?.getPageAriaLabel,
```
ThinuwanW, line 60: 🤖 **Claude** · 🟣 **Nit**

`mergedAriaLabel` is rebuilt as a fresh object literal on every render, so `TableV2` receives a new `ariaLabel` reference each time and any memoization inside it is defeated; this PR grows that object rather than fixing it. Wrap it in `useMemo` keyed on the individual `ariaLabel?.*` fields (or on `ariaLabel` itself, if callers pass a stable object). The same referential-instability concern now applies to the new `pageSizeOptions` array — the invoice consumer builds it inline with `Object.values(ItemsPerPage).filter(...)`, producing a new array identity every render that flows straight into `TableV2`; worth calling out in the `TableViewPagination` doc comment that callers should hoist or memoize it.

frontend/src/community/common/components/organisms/TableView/TableView.tsx
```diff
@@ -12,6 +12,7 @@ const TableView: FC<TableViewProps> = ({
   heading,
   tableName,
   ariaLabel,
+  variant,
```
ThinuwanW, line 15: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

There is no `TableView.test.tsx` anywhere in the repo, and this PR adds five more conditionally-forwarded props to a component whose entire job is conditional prop forwarding. `jest.config.js:30` excludes `**/organisms/**` from coverage collection, but sibling organisms are still tested (`organisms/Forms/SignInForm/SignInForm.test.tsx`, etc.), so tests here are not against convention. A small test asserting that (a) the page-size props reach the table when `pagination` is set, (b) they are dropped when `infiniteScroll.isEnabled` is true even though `pagination` is also supplied, and (c) `variant` is forwarded, would lock in the gating rules cheaply — remember the `MockTheme` wrapper and `screen.getByRole` patterns used by the existing organism tests.

frontend/src/community/common/components/organisms/TableView/types.ts
ThinuwanW, line 6: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

This PR introduces a first-class page-size selector in the shared `TableView`, which now duplicates at least two hand-rolled implementations already in the codebase: `frontend/src/enterprise/sign/components/molecules/EnvelopeTable/EnvelopeTable.tsx` (~line 391) and `frontend/src/enterprise/invoice/components/molecules/CustomerProjects/CustomerProjectsTable.tsx` (~line 198) both render their own `rowsPerPage` label plus a `DropdownList` fed by an `ItemsPerPage` enum with a manually wired `ariaLabel`. Leaving three implementations in place means the label copy, option list, and accessible naming will drift. Not necessarily in-scope for this PR, but please file/track the follow-up to migrate those two tables onto the new `pagination.pageSize*` API and delete the bespoke dropdowns — otherwise the shared abstraction never actually consolidates anything.

frontend/src/community/common/components/organisms/TableView/TableView.tsx
Deshan1100: need this for set list varient no need test

frontend/src/community/common/components/organisms/TableView/TableView.tsx
Deshan1100: need this its a mergedAriaLabel checked

frontend/src/community/common/components/organisms/TableView/TableView.tsx
Deshan1100: checkd kept as previous

Deshan1100: need this checked

frontend/src/community/common/components/organisms/TableView/types.ts
Deshan1100: removed

frontend/src/community/common/components/organisms/TableView/types.ts
Deshan1100: updated skapp ui

Deshan1100: updated skapp ui

frontend/src/community/common/components/organisms/TableView/types.ts
Deshan1100: updated

Deshan1100: updated to result pare page

frontend/src/community/common/components/organisms/TableView/types.ts
Deshan1100: out of scope kept previous translations

frontend/src/community/common/components/organisms/TableView/types.ts
Deshan1100: page size props updated

frontend/src/community/common/components/organisms/TableView/TableView.tsx
Deshan1100: page size props updated

frontend/src/community/common/components/organisms/TableView/types.ts
Deshan1100: page size props updated

frontend/src/community/common/components/organisms/TableView/TableView.tsx
Deshan1100: updated

frontend/src/community/common/components/organisms/TableView/TableView.tsx
Deshan1100: page size props updated

shakila-kamalasena (approved):

akilarootcode (approved):

PR #1708, opened 2026-08-04
Feat/2853 business unit selection

shakila-kamalasena:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds a Business Unit CRUD feature (backend controller/service/DAO/entity + Liquibase, frontend configuration section with add/edit/delete modals, plus a business-unit field on the employee employment form and bulk upload). The implementation closely mirrors the existing WorkLocation feature and is generally consistent with it, but two issues will cause real problems in production: the business-unit list query is fired on every user's own account page while the endpoint is restricted to SUPER_ADMIN/PEOPLE_ADMIN (403 + blank read-only field), and the deletion-impact employee count includes terminated/deleted employees, so the delete dialog and toast report wrong numbers. Several secondary concerns around PATCH semantics, missing bean validation, unvalidated transfer target, duplication between the Add/Edit modals, and missing tests are listed below.

Found **31** new issue(s): 🔴 5 important, 🟡 16 suggestion(s), 🟣 10 nit(s)

frontend/src/community/people/hooks/useEmployeeDetailsFormHandler.tsx
```diff
@@ -98,6 +99,8 @@ const useEmployeeDetailsFormHandler = ({
 
   const { data: workLocations } = useGetAllWorkLocations();
 
+  const { data: businessUnits } = useGetBusinessUnits();
```
ThinuwanW, line 102: 🤖 **Claude** · 🔴 **Important**

`useGetBusinessUnits()` is called unconditionally here, but `GET /v1/com/business-unit` is gated to `ROLE_SUPER_ADMIN`/`ROLE_PEOPLE_ADMIN`. This hook is used by `EmploymentFormSection/SubSections/EmploymentDetailsSection`, which is rendered on `/community/account` (via PeopleAccount → AccountSectionWrapper → PeopleAccountSection → EmploymentDetailsForm) for EVERY logged-in user, and on `/community/people/individual/[id]` (via PeopleIndividualSection, read-only) for people/leave/attendance managers. Two consequences: (1) every non-admin user triggers a 403 on their own account page; (2) because the read-only Business unit dropdown resolves its display label by looking `values.businessUnitId` up in the fetched `businessUnits` list, those users see an EMPTY Business unit field even when the employee is assigned — `businessUnitId` is returned by `mapEmploymentBasicDetails` without any access-level gate, so the data is there but unresolvable. Contrast `useGetAllWorkLocations`, whose `/work-location/all` endpoint permits `ROLE_PEOPLE_EMPLOYEE` precisely so read-only views work. Fix by either (a) broadening the `@PreAuthorize` on `getAllBusinessUnits` to include `ROLE_PEOPLE_EMPLOYEE` like the work-location `/all` endpoint, or (b) returning the business unit *name* on `EmployeeEmploymentBasicDetailsDto` for display and adding `enabled: isSuperAdmin || isPeopleAdmin` to this query so non-admins never call it.

backend/src/main/java/com/skapp/community/common/service/impl/BusinessUnitServiceImpl.java
```diff
@@ -0,0 +1,189 @@
...
+	@Override
+	@Transactional(readOnly = true)
+	public ResponseEntityDto getBusinessUnitDeletionImpact(Long id) {
+		log.info("getBusinessUnitDeletionImpact: execution started");
+
+		businessUnitDao.findById(id)
+			.orElseThrow(() -> new ModuleException(CommonMessageConstant.COMMON_ERROR_BUSINESS_UNIT_NOT_FOUND));
+
+		BusinessUnitDeletionImpactResponseDto responseDto = new BusinessUnitDeletionImpactResponseDto();
+		responseDto.setAssignedEmployeeCount(employeeDao.countByBusinessUnitBusinessUnitId(id));
```
ThinuwanW, line 88: 🤖 **Claude** · 🔴 **Important**

`employeeDao.countByBusinessUnitBusinessUnitId(id)` (and `findByBusinessUnitBusinessUnitId(id)` on line 111) match on ALL account statuses. Employee deletion in this codebase is a soft delete (`updateUserStatus(userId, AccountStatus.DELETED, true)` in PeopleServiceImpl), and terminated employees also stay as rows, so the deletion-impact count includes deleted/terminated people. The user is then told e.g. "5 employees are currently assigned to X" and the success toast reports "5 employees transferred" when only 2 are real. The equivalent work-location code (`WorkLocationServiceImpl.clearWorkLocationFromEmployees`) deliberately filters via `findByWorkLocationWorkLocationIdAndAccountStatusIn(id, Set.of(ACTIVE, PENDING))`. Add matching status-filtered derived queries (e.g. `countByBusinessUnitBusinessUnitIdAndAccountStatusIn` / `findByBusinessUnitBusinessUnitIdAndAccountStatusIn`) and use them for both the count and the transfer/unassign loop, so the number shown and the rows actually touched agree.

backend/src/main/java/com/skapp/community/common/service/impl/BusinessUnitServiceImpl.java
```diff
@@ -0,0 +1,189 @@
...
+	@Override
+	@Transactional
+	public ResponseEntityDto deleteBusinessUnit(Long id, Long transferToBusinessUnitId) {
+		log.info("deleteBusinessUnit: execution started");
+
+		BusinessUnit businessUnit = businessUnitDao.findById(id)
+			.orElseThrow(() -> new ModuleException(CommonMessageConstant.COMMON_ERROR_BUSINESS_UNIT_NOT_FOUND));
+
+		BusinessUnit transferTarget = null;
+		if (transferToBusinessUnitId != null) {
```
ThinuwanW, line 105: 🤖 **Claude** · 🟡 **Suggestion**

`deleteBusinessUnit` never validates that `transferToBusinessUnitId` differs from `id`. If a client passes the same id, every assigned employee is re-pointed at the unit about to be deleted; Hibernate flushes those updates and then the DELETE, and the FK's `ON DELETE SET NULL` silently nulls them instead of raising an error — so the request "succeeds" while the caller was told the employees were transferred. The UI happens to exclude the current unit from the dropdown, but the endpoint is directly callable. Add an explicit guard, e.g. `if (id.equals(transferToBusinessUnitId)) throw new ModuleException(CommonMessageConstant.COMMON_ERROR_BUSINESS_UNIT_TRANSFER_TARGET_NOT_FOUND);` (or a dedicated message constant) before resolving the target.

backend/src/main/java/com/skapp/community/common/service/impl/BusinessUnitServiceImpl.java
```diff
@@ -0,0 +1,189 @@
...
+		validateBusinessUnitRequest(businessUnitRequestDto);
+
+		String name = businessUnitRequestDto.getName();
+
+		if (!name.equals(businessUnit.getName()) && existsByNameCaseSensitiveExcludingId(name, id)) {
+			throw new ModuleException(CommonMessageConstant.COMMON_ERROR_BUSINESS_UNIT_NAME_ALREADY_EXISTS);
+		}
+
+		businessUnit.setName(name);
+		businessUnit.setDescription(businessUnitRequestDto.getDescription());
```
ThinuwanW, line 71: 🤖 **Claude** · 🟡 **Suggestion**

`updateBusinessUnit` is mapped as `@PatchMapping` but behaves as a full replace: it calls `validateBusinessUnitRequest` (which throws if `name` is null/blank) and then unconditionally does `setDescription(dto.getDescription())`. A PATCH that sends only `{"description": "..."}` gets a 400, and a PATCH that sends only `{"name": "..."}` silently wipes the existing description. `WorkLocationServiceImpl.updateWorkLocation` handles this correctly by null-checking each field before assigning. Either apply fields only when non-null (matching the work-location precedent) or change the mapping to `@PutMapping` to make the replace semantics explicit.

backend/src/main/java/com/skapp/community/common/payload/request/BusinessUnitRequestDto.java
```diff
@@ -0,0 +1,14 @@
+package com.skapp.community.common.payload.request;
+
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+public class BusinessUnitRequestDto {
+
+	private String name;
```
ThinuwanW, line 10: 🤖 **Claude** · 🟡 **Suggestion**

The request DTO carries no bean-validation annotations and `BusinessUnitController` omits `@Valid` on both `@RequestBody` parameters, so all constraints live in `validateBusinessUnitRequest`. Per the review guidelines, String fields that must be non-empty need `@NotBlank` and length limits need `@Size`, with `@Valid` on the `@RequestBody` to trigger them. Add `@NotBlank(message = ...)` + `@Size(max = 100)` on `name` and `@Size(max = 250)` on `description`, add `@Valid` in the controller, and drop the now-duplicated manual checks (guidelines also flag "custom validation in service layer that duplicates what bean validation already catches"). Note `WorkLocationController` imports `jakarta.validation.Valid` but never uses it — the same gap, worth not replicating.

backend/src/main/java/com/skapp/community/common/service/impl/BusinessUnitServiceImpl.java
```diff
@@ -0,0 +1,189 @@
...
+
+		BusinessUnit businessUnit = businessUnitDao.findById(id)
+			.orElseThrow(() -> new ModuleException(CommonMessageConstant.COMMON_ERROR_BUSINESS_UNIT_NOT_FOUND));
+
+		log.info("getBusinessUnitById: execution ended");
+
+		return new ResponseEntityDto(false, mapToResponseDto(businessUnit));
+	}
+
+	private boolean existsByNameCaseSensitive(String name) {
```
ThinuwanW, line 153: 🤖 **Claude** · 🟡 **Suggestion**

Name uniqueness is enforced only in application code and only case-sensitively: `existsByNameCaseSensitive` fetches `findByNameIgnoreCase(name)` and then filters with `.equals(name)`, so "Sales" and "sales" can both be created. There is also no unique index — the new DDL declares `name` as `text NOT NULL` with no constraint, and the entity omits the `unique = true` that `WorkLocation.name` carries — so two concurrent creates can both pass the check and persist duplicates. This diverges from work location, which uses a genuinely case-insensitive `existsByNameIgnoreCase`. It also leaks into bulk upload: `validateBusinessUnitInBulk`/`setBulkEmployeeBusinessUnit` require an exact-case match, so a CSV containing "sales" for a unit named "Sales" is rejected with "Business unit 'sales' does not exist", whereas the work-location column is case-insensitive. Decide on one rule (case-insensitive is both the existing convention and the friendlier one for spreadsheet input) and back it with a DB-level unique index — which will require changing `name` from `text` to `varchar(100)` in the create-table script, matching the 100-char application limit.

backend/src/main/java/com/skapp/community/common/service/impl/BusinessUnitServiceImpl.java
```diff
@@ -0,0 +1,189 @@
...
+			.orElseThrow(() -> new ModuleException(CommonMessageConstant.COMMON_ERROR_BUSINESS_UNIT_NOT_FOUND));
+
+		BusinessUnit transferTarget = null;
+		if (transferToBusinessUnitId != null) {
+			transferTarget = businessUnitDao.findById(transferToBusinessUnitId)
+				.orElseThrow(() -> new ModuleException(
+						CommonMessageConstant.COMMON_ERROR_BUSINESS_UNIT_TRANSFER_TARGET_NOT_FOUND));
+		}
+
+		List<Employee> assignedEmployees = employeeDao.findByBusinessUnitBusinessUnitId(id);
```
ThinuwanW, line 111: 🤖 **Claude** · 🟡 **Suggestion**

`findByBusinessUnitBusinessUnitId(id)` loads every assigned employee entity into memory and `saveAll` issues one UPDATE per row. On a large tenant where a business unit covers thousands of employees this is a slow, memory-heavy transaction for what is a single-column reassignment. The guidelines call out unbounded `findAll`-style queries. Consider a `@Modifying @Query` bulk update on `EmployeeDao` (e.g. `update Employee e set e.businessUnit = :target where e.businessUnit.businessUnitId = :id`) and note that the FK already declares `ON DELETE SET NULL`, so the unassign path (`transferToBusinessUnitId == null`) does not need the explicit loop at all.

backend/src/main/java/com/skapp/community/common/controller/v1/BusinessUnitController.java
```diff
@@ -0,0 +1,82 @@
...
+	@GetMapping
+	public ResponseEntity<ResponseEntityDto> getAllBusinessUnits() {
+		ResponseEntityDto response = businessUnitService.getAllBusinessUnits();
+		return new ResponseEntity<>(response, HttpStatus.OK);
+	}
+
+	@Operation(summary = "Get a business unit by ID")
+	@PreAuthorize("hasAnyRole('ROLE_SUPER_ADMIN')")
+	@GetMapping(value = "/{id}")
+	public ResponseEntity<ResponseEntityDto> getBusinessUnitById(@PathVariable Long id) {
```
ThinuwanW, line 40: 🤖 **Claude** · 🟡 **Suggestion**

`getBusinessUnitById` (plus `BusinessUnitService.getBusinessUnitById` and its impl) has no caller — `BusinessUnitApi.ts` only uses list, create, update, deletion-impact and delete. It is also role-inconsistent with the list endpoint (`ROLE_SUPER_ADMIN` only vs `ROLE_SUPER_ADMIN` + `ROLE_PEOPLE_ADMIN`), which suggests the role lists were not deliberately reasoned through. Either remove it as dead code or document why it exists and align its `@PreAuthorize` with `getAllBusinessUnits`.

frontend/src/community/configurations/components/molecules/AddBusinessUnitModal/AddBusinessUnitModal.tsx
```diff
@@ -0,0 +1,139 @@
+import {
```
ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

`AddBusinessUnitModal` and `EditBusinessUnitModal` are near-identical: same `BusinessUnitFormValues` interface (declared twice), same InputField/TextArea markup, same button config, same toast-then-reset-then-close flow, differing only in the modal title, the mutation hook, initial values and the unchanged-check. The guidelines call for extracting a shared component when 2+ components have near-identical markup. Extract a single `BusinessUnitFormModal` taking `{ title, initialValues, isPending, onSubmit }` (and move the duplicated `BusinessUnitFormValues` type into `~community/common/types/BusinessUnitTypes.ts`), with the two callers supplying only the mutation wiring.

frontend/src/community/common/api/BusinessUnitApi.ts
```diff
@@ -0,0 +1,133 @@
...
+  AxiosResponse<BusinessUnit>,
+  AxiosError,
+  BusinessUnitDeleteVariables
+> => {
+  const queryClient = useQueryClient();
+
+  return useMutation({
+    mutationFn: deleteBusinessUnit,
+    onSuccess: () => {
+      queryClient.invalidateQueries({ queryKey: businessUnitQueryKeys.ALL });
```
ThinuwanW, line 128: 🤖 **Claude** · 🟡 **Suggestion**

All three mutations invalidate only `businessUnitQueryKeys.ALL`, never `DELETION_IMPACT`. Deleting a unit with transfer changes the assigned-employee count of the *target* unit, so a cached `DELETION_IMPACT(targetId)` entry becomes wrong. With React Query v5 defaults (`staleTime: 0`) a cached entry still resolves the query to `success` immediately, so `isLoading` is false and `DeleteBusinessUnitModal` renders the stale count — and `handleDeleteSuccess` builds its toast from that stale `impact.assignedEmployeeCount`. Add `queryClient.invalidateQueries({ queryKey: ["business-unit-deletion-impact"] })` in the delete mutation's `onSuccess` (a prefix match over the whole family), and consider doing the same wherever an employee's business unit is reassigned. While here: none of the four `useGetBusinessUnits()` call sites set `staleTime`, so this near-static list is refetched on every mount of the employment form and of each modal — a modest `staleTime` would remove that.

frontend/src/community/configurations/components/organisms/BusinessUnitsSection/BusinessUnitsSection.tsx
```diff
@@ -0,0 +1,103 @@
...
+        <div className="grid grid-cols-1 gap-6 sm:grid-cols-2 lg:grid-cols-3">
+          {skeletonKeys.map((key) => (
+            <BusinessUnitCardSkeleton key={key} />
+          ))}
+        </div>
+      </div>
+    );
+  }
+
+  if (!businessUnits) {
```
ThinuwanW, line 52: 🤖 **Claude** · 🟡 **Suggestion**

`if (!businessUnits) return null;` collapses the error case into a blank area with no message and no retry — a failed `GET /v1/com/business-unit` is indistinguishable from "nothing here". The guidelines require loading AND error states to be rendered. Destructure `isError` from `useGetBusinessUnits()` and render an error state (or a toast) with a retry, keeping `null` only for the genuinely-undefined transitional case.

frontend/src/community/configurations/components/molecules/AddBusinessUnitModal/AddBusinessUnitModal.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+  const { data: businessUnits } = useGetBusinessUnits();
+
+  const handleCreateSuccess = () => {
+    setToastMessage({
+      open: true,
+      toastType: ToastType.SUCCESS,
+      title: translateText(["toasts", "createSuccess", "title"]),
+      description: translateText(["toasts", "createSuccess", "description"], {
+        name: formik.values.name.trim()
```
ThinuwanW, line 42: 🤖 **Claude** · 🟡 **Suggestion**

When the backend rejects a create/update — most likely `api.error.common.business-unit-name-already-exists`, which the client-side Yup `name-unique` test cannot catch reliably because `businessUnits` may be stale or empty (it defaults to `[]` in `businessUnitValidation`) — `handleCreateError` discards the `AxiosError` and shows the generic "Failed to save business unit. Please try again." The user gets no indication that the name is the problem. Read the message from the error response and either surface it in the toast or map the duplicate-name case to `formik.setFieldError("name", ...)` so the error appears on the offending field. The same applies to `EditBusinessUnitModal.handleUpdateError`.

frontend/src/community/configurations/components/molecules/AddBusinessUnitModal/AddBusinessUnitModal.tsx
ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

None of the six new frontend components (AddBusinessUnitModal, EditBusinessUnitModal, DeleteBusinessUnitModal, DeleteBusinessUnitTransferContent, BusinessUnitCard, BusinessUnitsSection) nor `BusinessUnitServiceImpl` ship tests, and the only test changes in this PR are field additions to existing bulk-upload fixtures. The branching worth covering is real and easy to get wrong: the three-way delete flow (count === 0 / other units exist / no other units), the transfer-target dropdown excluding the current unit, and on the backend the duplicate-name and transfer/unassign paths. Add at least a `DeleteBusinessUnitModal.test.tsx` (with the `MockTheme` wrapper and `screen.getByRole` queries per the guidelines) and a `BusinessUnitServiceImplTest`.

backend/src/main/resources/community/messages/common-messages.properties
```diff
@@ -129,4 +129,11 @@ api.error.common.work-location-not-found=Work location not found
...
 api.error.common.work-location-name-length-exceeded=Work location name must not exceed {0} characters
-api.error.common.work-location-employee-assignment-conflict=Cannot specify both all employees and specific employee IDs
\ No newline at end of file
+api.error.common.work-location-employee-assignment-conflict=Cannot specify both all employees and specific employee IDs
+
+api.error.common.business-unit-not-found=Business unit not found
+api.error.common.business-unit-name-already-exists=A business unit with this name already exists
+api.error.common.business-unit-name-required=Please enter a business unit name
+api.error.common.business-unit-name-length-exceeded=Business unit name must be 100 characters or fewer
+api.error.common.business-unit-description-length-exceeded=Description must be 250 characters or fewer
```
ThinuwanW, line 138: 🤖 **Claude** · 🟣 **Nit**

`business-unit-name-length-exceeded` and `business-unit-description-length-exceeded` hardcode "100" and "250" in the message text, duplicating `CommonConstants.BUSINESS_UNIT_NAME_MAX_LENGTH` / `BUSINESS_UNIT_DESCRIPTION_MAX_LENGTH` (and again in `businessUnitValidations.ts`, which defines its own `BUSINESS_UNIT_NAME_MAX_LENGTH = 100`). Changing a limit now means editing three places. `ModuleException` already has an `(MessageConstant, Object[])` constructor, so use `{0}` in the properties file and pass the constant — the adjacent `work-location-name-length-exceeded` key already uses `{0}` (though its throw site never supplies the arg, so it currently renders the literal placeholder — worth not copying). Also note this file still ends without a trailing newline.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
```diff
@@ -495,6 +506,29 @@ private Employee createEmployeeEntity(Employee employee, CreateEmployeeRequestDt
 		return employee;
 	}
 
+	private void processBusinessUnit(CreateEmployeeRequestDto requestDto, Employee employee) {
+		if (requestDto == null || requestDto.getEmployment() == null
```
ThinuwanW, line 510: 🤖 **Claude** · 🟣 **Nit**

The guard at the top of `processBusinessUnit` re-checks exactly what the caller on lines 433-435 already checked (`requestDto`, `getEmployment()`, `getEmploymentDetails()` non-null) and adds `employee == null`, which can never be true — `createEmployeeEntity` is only ever called with a constructed entity. Drop the duplicate condition at the call site and keep the guard inside the method (or vice versa). Separately, `findBusinessUnitByExactName` here duplicates `existsByNameCaseSensitive`/`existsByNameCaseSensitiveExcludingId` in `BusinessUnitServiceImpl`; a single lookup method exposed on `BusinessUnitService` would keep the case-matching rule in one place.

frontend/src/community/configurations/components/molecules/EditBusinessUnitModal/EditBusinessUnitModal.tsx
```diff
@@ -0,0 +1,164 @@
...
+      name === businessUnit.name.trim() &&
+      description === (businessUnit.description ?? "").trim();
+    if (isUnchanged) {
+      onClose();
+      return;
+    }
+
+    updateBusinessUnit({
+      id: businessUnit.businessUnitId,
+      payload: { name, description }
```
ThinuwanW, line 83: 🤖 **Claude** · 🟣 **Nit**

Two small inconsistencies in how the description is handled. (1) Edit sends `description` as a possibly-empty string while Add sends `values.description.trim() || null`, so clearing a description on edit stores `""` rather than NULL — send `description || null` here to match. (2) Both modals validate with Yup `.max()` against the untrimmed field value but submit the trimmed value, so a 100-character name followed by a space is rejected as too long even though the payload would be valid; apply the max check to the trimmed value (or `.trim()` in a transform) so client and server agree.

frontend/src/community/configurations/components/molecules/AddBusinessUnitModal/AddBusinessUnitModal.tsx
```diff
@@ -0,0 +1,139 @@
...
+}
+
+const AddBusinessUnitModal: FC<Props> = ({ isOpen, onClose }) => {
+  const translateText = useTranslator("configurations", "businessUnit");
+  const { setToastMessage } = useToast();
+
+  const { data: businessUnits } = useGetBusinessUnits();
+
+  const handleCreateSuccess = () => {
+    setToastMessage({
```
ThinuwanW, line 37: 🤖 **Claude** · 🟣 **Nit**

`handleCreateSuccess` reads the name for the toast from live form state (`formik.values.name.trim()`) rather than from the values that were submitted. The InputField is not disabled while `isPending` (only the buttons are), so a user who keeps typing during the request sees a toast naming something that was never created. Capture the name in `handleSubmit` and pass it through (e.g. via a ref or by using the mutation's `variables`). `EditBusinessUnitModal.handleUpdateSuccess` has the same pattern.

frontend/src/community/common/components/organisms/CommonModalController/CommonModalController.tsx
```diff
@@ -68,6 +71,29 @@ const CommonModalController = (): JSX.Element => {
...
+          isOpen={isCommonModalOpen}
+          onClose={closeCommonModal}
+        />
+      );
+    case CommonModalType.EDIT_BUSINESS_UNIT:
+      return (
+        <EditBusinessUnitModal
+          isOpen={isCommonModalOpen}
+          onClose={closeCommonModal}
+          businessUnit={commonModalData?.businessUnit!}
```
ThinuwanW, line 86: 🤖 **Claude** · 🟣 **Nit**

`commonModalData?.businessUnit!` uses a non-null assertion to satisfy the required `businessUnit` prop — the guidelines flag assertions used to silence type errors rather than fixing the underlying type. It is safe today only because `closeCommonModal` resets `commonModalType` and `commonModalData` in the same `set()` call, and it matches the existing lines above, so this is consistency-with-precedent rather than a new bug. If you touch this file again, prefer an explicit early return (`if (!commonModalData?.businessUnit) return <></>;`) over `!`.

frontend/src/community/configurations/components/molecules/BusinessUnitCard/BusinessUnitCard.tsx
```diff
@@ -0,0 +1,69 @@
+import {
```
ThinuwanW, line 1: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

`description` is collected in both modals, validated to 250 chars, persisted and returned by the API, but never rendered anywhere except as a prefilled edit field — the card shows only the name. Either display it on the card (truncated) or reconsider whether the field earns its place in the form. Also `configurations.json` adds `businessUnit.toasts.deleteError.title`, which is unused: `handleDeleteError` uses `toasts.error.title` with `toasts.deleteError.description`. And the toast copy mixes quote styles — `createSuccess`/`updateSuccess` use escaped double quotes around `{{name}}` while the three delete toasts use single quotes.

frontend/src/community/people/components/organisms/EmploymentFormSection/SubSections/EmploymentDetailsSection.tsx
```diff
@@ -651,6 +656,44 @@ const EmploymentDetailsSection = forwardRef<FormMethods, Props>(
                 readOnly={isReadOnly || isProfileView}
               />
             </Grid>
+
```
ThinuwanW, line 659: 🤖 **Claude** · 🟣 **Nit**

The Business unit dropdown was added only to this edit/view section, not to `AddNewResourceFlow/EmploymentDetailsSection`, so a business unit cannot be assigned while creating an employee even though `createEmployeeEntity` and the bulk-upload path both support it. Work location has the same gap in the add flow, so this is consistent with the existing behaviour rather than a regression — flagging so the omission is a deliberate scope decision rather than an oversight.

frontend/src/community/people/utils/userBulkUploadUtils.ts
```diff
@@ -128,6 +128,7 @@ export const convertUserBulkCsvHeaders = (header: string) => {
     "Probation End Date": "probationEndDate",
     "Work Time Zone": "timeZone",
     "Work Location": "workLocation",
+    "Business Unit": "businessUnit",
```
ThinuwanW, line 131: 🤖 **Claude** · 🔴 **Important**

Missing paired change that breaks CSV bulk upload. `convertUserBulkCsvHeaders` now maps the "Business Unit" column to `businessUnit`, the DTOs/converter/test fixtures were updated, and the .xlsx template binary was regenerated in this same commit — but `USER_BULK_HEADERS` in `frontend/src/community/people/utils/constants/constants.ts` was NOT updated. `useUserBulkValidation.isCsvValid` compares the parsed header list against that constant with a strict length equality check (`headerList.length === USER_BULK_HEADERS.length`) AND a contains-all check. So if the regenerated template now carries a Business Unit column, every upload of the official template is rejected with the generic "templateError" message and the whole bulk-upload feature is blocked. If the template does NOT carry the column, then `businessUnit` is never populated and the backend plumbing added in this PR is dead code. Fix: add "businessUnit" to `USER_BULK_HEADERS` (immediately after "workLocation" so the order matches the template), and update the exact expected-headers array in `constants.test.ts` (it asserts `toEqual` on the full list, so it will fail otherwise). Please also verify the template column header text is exactly "Business Unit" so the header transform resolves.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
```diff
@@ -1810,6 +1844,20 @@ public void setBulkEmployeeWorkLocation(EmployeeBulkDto employeeBulkDto, Employe
...
+		if (businessUnit != null && !businessUnit.isBlank()) {
+			findBusinessUnitByExactName(businessUnit).ifPresentOrElse(employee::setBusinessUnit, () -> {
+				throw new EntityNotFoundException(
+						PeopleMessageConstant.PEOPLE_ERROR_VALIDATION_BUSINESS_UNIT_NOT_FOUND);
+			});
+		}
+	}
+
+	private Optional<BusinessUnit> findBusinessUnitByExactName(String name) {
+		return businessUnitDao.findByNameIgnoreCase(name).stream().filter(bu -> bu.getName().equals(name)).findFirst();
```
ThinuwanW, line 1858: 🤖 **Claude** · 🔴 **Important**

`findBusinessUnitByExactName` resolves the CSV value case-sensitively and without trimming (it queries ignore-case, then filters with `equals` on the raw string). Both `setBulkEmployeeBusinessUnit` and `validateBusinessUnitInBulk` use it, so a spreadsheet value of "sales" or "Sales " (trailing space — extremely common in exported CSVs; papaparse does not trim cell values and no `transform` is configured on the frontend parser) fails to match an existing "Sales" unit and the whole row is rejected. This is inconsistent with the sibling Work Location column in the very same file, which matches with `findByNameIgnoreCase` and trims in the validator — so the same CSV tolerates casing/whitespace in one column and rejects it in the next. Fix: trim the incoming value and match case-insensitively (i.e. return the first result of `businessUnitDao.findByNameIgnoreCase(name.trim())`), or at minimum trim before the exact comparison in both the validate and the set path.

backend/src/main/java/com/skapp/community/common/service/impl/BusinessUnitServiceImpl.java
```diff
@@ -0,0 +1,189 @@
...
+		validateBusinessUnitRequest(businessUnitRequestDto);
+
+		String name = businessUnitRequestDto.getName();
+
+		if (existsByNameCaseSensitive(name)) {
+			throw new ModuleException(CommonMessageConstant.COMMON_ERROR_BUSINESS_UNIT_NAME_ALREADY_EXISTS);
+		}
+
+		BusinessUnit businessUnit = new BusinessUnit();
+		businessUnit.setName(name);
```
ThinuwanW, line 45: 🤖 **Claude** · 🔴 **Important**

Neither `createBusinessUnit` nor `updateBusinessUnit` trims the incoming name/description before persisting — the raw request value is stored. `validateBusinessUnitRequest` only rejects null/blank, so " Sales" is accepted and stored with its padding. Consequences: (1) it silently defeats the uniqueness check, since " Sales" and "Sales" are different strings and both can be created; (2) the bulk-import lookup (`findBusinessUnitByExactName`) can then never match the padded unit from a CSV; (3) the name renders with leading/trailing space in the dropdown and cards. The frontend trims before sending, so this is only reachable via direct API use today, but the service is the contract boundary and should not depend on client-side trimming. Fix: trim name (and description) inside `validateBusinessUnitRequest`/before assignment, and enforce the max-length check against the trimmed value so the limit matches what is actually stored.

backend/src/main/resources/community/messages/people-messages.properties
```diff
@@ -170,6 +170,7 @@ api.error.people.transfer-supervisors.primary-record-not-found=The specified emp
 api.error.people.transfer-supervisors.team-record-not-found=The specified user is not a supervisor of this team
 api.error.people.employee-termination-or-deletion-action-required=Employee termination or deletion action is required
 api.error.people.validation-work-location-not-found=Work location not found
+api.error.people.validation-business-unit-not-found=Business unit ''{0}'' does not exist
```
ThinuwanW, line 173: 🤖 **Claude** · 🟡 **Suggestion**

The new message uses MessageFormat-escaped quotes and a `{0}` placeholder, but it is consumed from two places with different arg handling. `validateBusinessUnitInBulk` passes an args array, so MessageFormat runs and it renders correctly. `setBulkEmployeeBusinessUnit` throws `EntityNotFoundException` with the constant and NO args — `MessageUtil.getMessage(MessageConstant)` passes null args, and Spring's `AbstractMessageSource` skips MessageFormat entirely when args are null (`always-use-message-format` is not enabled in application-prd.yml/application-non-prd.yml). That path therefore surfaces the raw text, doubled quotes and literal `{0}` included, into the per-row bulk error shown in the upload summary. It is only reachable when a unit is deleted between validation and the async row creation, but it is still user-visible. Also note every other parameterized message in this file uses plain `{0}` without the doubled quotes. Fix: pass the business unit name as an args array when constructing the `EntityNotFoundException`, and drop the doubled single quotes to match the rest of the file.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
```diff
@@ -495,6 +506,29 @@ private Employee createEmployeeEntity(Employee employee, CreateEmployeeRequestDt
...
+		}
+
+		Set<String> userRoles = userService.getCurrentUserRoles();
+		boolean canModifyBusinessUnit = userRoles.contains(AuthUtil.withRolePrefix(Role.SUPER_ADMIN))
+				|| userRoles.contains(AuthUtil.withRolePrefix(Role.PEOPLE_ADMIN));
+		if (!canModifyBusinessUnit) {
+			return;
+		}
+
+		employee.setBusinessUnit(businessUnitDao.findById(businessUnitId)
```
ThinuwanW, line 527: 🤖 **Claude** · 🟡 **Suggestion**

`businessUnitId` has no counterpart in `EmployeeValidationServiceImpl.validateCreateEmployeeRequestEmploymentDetails`, unlike `workLocationId` which is pre-validated there via `Validations.validateWorkLocation` and throws `ValidationException` (HTTP 400) before any entity mutation. Business unit is instead only resolved deep inside `createEmployeeEntity`, throwing `EntityNotFoundException`, which `GlobalExceptionHandler` maps to HTTP 404 — so the same class of bad input on the same request body returns 400 for one field and 404 for the other, and the failure happens after the employee entity has already been partially mutated. Fix: add a `Validations.validateBusinessUnit(businessUnitId, businessUnitDao)` helper alongside `validateWorkLocation` and call it from `EmployeeValidationServiceImpl` so both fields fail fast with the same status and error shape.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
```diff
@@ -495,6 +506,29 @@ private Employee createEmployeeEntity(Employee employee, CreateEmployeeRequestDt
...
+
+		Long businessUnitId = requestDto.getEmployment().getEmploymentDetails().getBusinessUnitId();
+		if (businessUnitId == null) {
+			return;
+		}
+
+		Set<String> userRoles = userService.getCurrentUserRoles();
+		boolean canModifyBusinessUnit = userRoles.contains(AuthUtil.withRolePrefix(Role.SUPER_ADMIN))
+				|| userRoles.contains(AuthUtil.withRolePrefix(Role.PEOPLE_ADMIN));
+		if (!canModifyBusinessUnit) {
```
ThinuwanW, line 523: 🤖 **Claude** · 🟡 **Suggestion**

When the current user is neither SUPER_ADMIN nor PEOPLE_ADMIN, `processBusinessUnit` silently returns and discards the submitted `businessUnitId`, and the request still succeeds with HTTP 200 — the caller gets no signal that part of their payload was dropped. Note `UserServiceImpl.getCurrentUserRoles()` returns an empty set when there is no `Authentication`, so any non-request-scoped invocation path would also silently drop the value rather than fail loudly. Fix: throw a `ValidationException`/403 with a dedicated message constant when a caller who cannot modify the business unit actually supplies one, and only no-op when the field is absent. That keeps the write behaviour explicit and testable.

backend/src/main/java/com/skapp/community/common/service/impl/BusinessUnitServiceImpl.java
```diff
@@ -0,0 +1,189 @@
+package com.skapp.community.common.service.impl;
```
ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

No backend tests were added for the new controller or service. The repo has an established `*ControllerIntegrationTest` convention under `backend/src/test/.../controller/v1/` (CRM board/company/contact/deal/deal-stage/task, several leave controllers, people, holiday, job), and this PR introduces a full CRUD surface plus a destructive delete-with-transfer operation and a role-gated write path in `PeopleServiceImpl`. At minimum, cover: create/update name-uniqueness rejection, delete with a transfer target (employees reassigned) vs. without (employees unassigned), deletion-impact counts, and the role gate in `processBusinessUnit` (admin sets the unit, non-admin's value is ignored/rejected). These are exactly the behaviours the other first-pass findings are about, so tests would lock them down as they get fixed.

frontend/src/community/common/api/BusinessUnitApi.ts
```diff
@@ -0,0 +1,133 @@
+import {
```
ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Module placement diverges from the established split for the analogous work-location feature. There, configuration-only CRUD lives in `~community/configurations/api/WorkLocationApi.ts` + `~community/configurations/api/utils/ApiEndpoints.ts`/`QueryKeys.ts` + `~community/configurations/types/WorkLocationTypes.ts`, and only the shared read-only `all` hook lives in `~community/common/api/WorkLocationApi.ts`. This PR puts everything — including create/update/delete/deletion-impact, which are used exclusively by the configurations module — in `~community/common/`. Separately, this PR creates `~community/common/constants/businessUnitConstants.ts` (holding only the unassign sentinel) but then declares `BUSINESS_UNIT_NAME_MAX_LENGTH`/`BUSINESS_UNIT_DESCRIPTION_MAX_LENGTH` in `~community/configurations/utils/businessUnitValidations.ts`, so business-unit constants are split across two files in two modules. Fix: move the CRUD hooks/endpoints/query keys into the configurations module (keeping only the list hook in common, since the people module consumes it), and consolidate the length constants into the single `businessUnitConstants.ts` file.

frontend/src/community/common/api/BusinessUnitApi.ts
```diff
@@ -0,0 +1,133 @@
...
+  BusinessUnitRequestPayload,
+  BusinessUnitUpdateVariables
+} from "~community/common/types/BusinessUnitTypes";
+import authFetch from "~community/common/utils/axiosInterceptor";
+
+const getBusinessUnits = async (): Promise<BusinessUnit[]> => {
+  const response = await authFetch.get(
+    businessUnitEndpoints.GET_BUSINESS_UNITS
+  );
+  return response.data.results;
```
ThinuwanW, line 25: 🤖 **Claude** · 🟣 **Nit**

`getBusinessUnits` returns the results array with no fallback, unlike the established `useGetAllWorkLocations` in the same folder which uses a `select` with a `?? []` guard. If the payload ever lacks `results` (error envelope, proxy/interceptor edge case), `businessUnits` becomes `undefined`/`null`. The default parameter in `businessUnitValidation(translateText, businessUnits = [])` only kicks in for `undefined`, not `null`, so a null would throw inside the Yup `name-unique` test at submit time rather than degrading gracefully. Fix: return `response.data?.results ?? []` (or mirror the `select` pattern used by the work-location hook).

frontend/src/community/common/assets/languages/english/configurations.json
```diff
@@ -241,6 +241,80 @@
...
+      "deleteAndUnassignSuccess": {
+        "title": "Business Unit Deleted",
+        "description_one": "'{{name}}' has been deleted. {{count}} employee is now unassigned.",
+        "description_other": "'{{name}}' has been deleted. {{count}} employees are now unassigned."
+      },
+      "error": {
+        "title": "Oops! Something went wrong",
+        "description": "Failed to save business unit. Please try again."
+      },
+      "deleteError": {
```
ThinuwanW, line 312: 🤖 **Claude** · 🟣 **Nit**

Three of the new translation keys are dead and the quoting is inconsistent. `toasts.deleteError.title`, `toasts.deleteAndTransferSuccess.title` and `toasts.deleteAndUnassignSuccess.title` are never read — `DeleteBusinessUnitModal` always uses `toasts.error.title` for failures and `toasts.deleteSuccess.title` for all three success variants. Also, the create/update toast descriptions wrap the unit name in escaped double quotes while all four delete-related descriptions use single quotes, so the same name renders differently depending on which toast fires. Fix: delete the unused title keys (or use `deleteError.title` in the delete failure toast, which is clearly what was intended) and pick one quote style across the whole `toasts` block.

frontend/src/community/configurations/components/organisms/BusinessUnitsSection/BusinessUnitsSection.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+  const handleDeleteUnit = (businessUnit: BusinessUnit) => {
+    openCommonModal(CommonModalType.DELETE_BUSINESS_UNIT, { businessUnit });
+  };
+
+  if (isLoading) {
+    return (
+      <div className="flex flex-col gap-4">
+        <div className="flex items-center justify-between">
+          <h1 className="h2">{translateText(["title"])}</h1>
```
ThinuwanW, line 41: 🤖 **Claude** · 🟣 **Nit**

The section title is rendered as an `h1` (with an `h2` utility class) in both the loading branch and the loaded branch. Every sibling configuration organism (`LeaveConfigurations`, `CrmConfigurations`) uses an `h2` element for its section heading, and this component is mounted inside a tab panel on a page that already has a page-level heading — so this introduces a second `h1` and breaks the heading hierarchy for screen readers. The identical header block is also duplicated across the two return branches. Fix: change the element to `h2` (keep the `h2` class for styling) and lift the shared header markup out of the loading/loaded branches so it is declared once.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
shakila-kamalasena: No need

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
shakila-kamalasena: Fixed

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
shakila-kamalasena: No need to trim

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
shakila-kamalasena: Fixed

backend/src/main/resources/community/messages/people-messages.properties
shakila-kamalasena: Fixed

frontend/src/community/people/components/organisms/EmploymentFormSection/SubSections/EmploymentDetailsSection.tsx
shakila-kamalasena: No need

frontend/src/community/people/utils/userBulkUploadUtils.ts
shakila-kamalasena: Fixed

frontend/src/community/people/hooks/useEmployeeDetailsFormHandler.tsx
shakila-kamalasena: Fixed

backend/src/main/java/com/skapp/community/common/service/impl/BusinessUnitServiceImpl.java
shakila-kamalasena: Not related to this PR

backend/src/main/java/com/skapp/community/common/service/impl/BusinessUnitServiceImpl.java
shakila-kamalasena: Not Related to this PR, this PR only contains assigning a business unit to an employee

backend/src/main/java/com/skapp/community/common/service/impl/BusinessUnitServiceImpl.java
shakila-kamalasena: Not Related to this PR, this PR only contains assigning a business unit to an employee

backend/src/main/java/com/skapp/community/common/payload/request/BusinessUnitRequestDto.java
shakila-kamalasena: Not Related to this PR, this PR only contains assigning a business unit to an employee

backend/src/main/java/com/skapp/community/common/service/impl/BusinessUnitServiceImpl.java
shakila-kamalasena: Not Related to this PR, this PR only contains assigning a business unit to an employee

backend/src/main/java/com/skapp/community/common/service/impl/BusinessUnitServiceImpl.java
shakila-kamalasena: Not Related to this PR, this PR only contains assigning a business unit to an employee

backend/src/main/java/com/skapp/community/common/controller/v1/BusinessUnitController.java
shakila-kamalasena: Not Related to this PR, this PR only contains assigning a business unit to an employee

frontend/src/community/configurations/components/molecules/AddBusinessUnitModal/AddBusinessUnitModal.tsx
shakila-kamalasena: Not Related to this PR, this PR only contains assigning a business unit to an employee

frontend/src/community/common/api/BusinessUnitApi.ts
shakila-kamalasena: Not Related to this PR, this PR only contains assigning a business unit to an employee

frontend/src/community/configurations/components/organisms/BusinessUnitsSection/BusinessUnitsSection.tsx
shakila-kamalasena: Not Related to this PR, this PR only contains assigning a business unit to an employee

frontend/src/community/configurations/components/molecules/AddBusinessUnitModal/AddBusinessUnitModal.tsx
shakila-kamalasena: Not Related to this PR, this PR only contains assigning a business unit to an employee

frontend/src/community/configurations/components/molecules/AddBusinessUnitModal/AddBusinessUnitModal.tsx
shakila-kamalasena: Not Related to this PR, this PR only contains assigning a business unit to an employee

backend/src/main/resources/community/messages/common-messages.properties
shakila-kamalasena: Not Related to this PR, this PR only contains assigning a business unit to an employee

frontend/src/community/configurations/components/molecules/EditBusinessUnitModal/EditBusinessUnitModal.tsx
shakila-kamalasena: Not Related to this PR, this PR only contains assigning a business unit to an employee

frontend/src/community/configurations/components/molecules/AddBusinessUnitModal/AddBusinessUnitModal.tsx
shakila-kamalasena: Not Related to this PR, this PR only contains assigning a business unit to an employee

frontend/src/community/common/components/organisms/CommonModalController/CommonModalController.tsx
shakila-kamalasena: Not Related to this PR, this PR only contains assigning a business unit to an employee

frontend/src/community/configurations/components/molecules/BusinessUnitCard/BusinessUnitCard.tsx
shakila-kamalasena: Not Related to this PR, this PR only contains assigning a business unit to an employee

backend/src/main/java/com/skapp/community/common/service/impl/BusinessUnitServiceImpl.java
shakila-kamalasena: Not Related to this PR, this PR only contains assigning a business unit to an employee

backend/src/main/java/com/skapp/community/common/service/impl/BusinessUnitServiceImpl.java
shakila-kamalasena: Not Related to this PR, this PR only contains assigning a business unit to an employee

frontend/src/community/common/api/BusinessUnitApi.ts
shakila-kamalasena: Not Related to this PR, this PR only contains assigning a business unit to an employee

frontend/src/community/common/api/BusinessUnitApi.ts
shakila-kamalasena: Not Related to this PR, this PR only contains assigning a business unit to an employee

frontend/src/community/common/assets/languages/english/configurations.json
shakila-kamalasena: Not Related to this PR, this PR only contains assigning a business unit to an employee

frontend/src/community/configurations/components/organisms/BusinessUnitsSection/BusinessUnitsSection.tsx
shakila-kamalasena: Not Related to this PR, this PR only contains assigning a business unit to an employee

thusala (dismissed):

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
```diff
@@ -1965,6 +1992,19 @@ public void setBulkEmployeeWorkLocation(EmployeeBulkDto employeeBulkDto, Employe
 		}
 	}
 
+	public void setBulkEmployeeBusinessUnit(EmployeeBulkDto employeeBulkDto, Employee employee) {
+		String businessUnit = employeeBulkDto.getBusinessUnit();
+		if (businessUnit != null && !businessUnit.isBlank()) {
+			businessUnitDao.findByName(businessUnit)
+				.filter(unit -> unit.getName().equals(businessUnit))
+				.ifPresentOrElse(employee::setBusinessUnit, () -> {
+					throw new EntityNotFoundException(
```
akilarootcode: ModuleException

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
```diff
@@ -1965,6 +1992,19 @@ public void setBulkEmployeeWorkLocation(EmployeeBulkDto employeeBulkDto, Employe
...
+							PeopleMessageConstant.PEOPLE_ERROR_VALIDATION_BUSINESS_UNIT_NOT_FOUND,
+							new Object[] { businessUnit });
```
akilarootcode: String or number or cast

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
```diff
@@ -2084,6 +2124,16 @@ public void validateWorkLocationInBulk(String workLocation, List<String> errors)
 		}
 	}
 
+	public void validateBusinessUnitInBulk(String businessUnit, List<String> errors) {
+		if (businessUnit != null && !businessUnit.isBlank()
+				&& businessUnitDao.findByName(businessUnit)
+					.filter(unit -> unit.getName().equals(businessUnit))
+					.isEmpty()) {
+			errors.add(messageUtil.getMessage(PeopleMessageConstant.PEOPLE_ERROR_VALIDATION_BUSINESS_UNIT_NOT_FOUND,
+					new Object[] { businessUnit }));
```
akilarootcode: Use proper String[]

akilarootcode (changes requested):

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java

akilarootcode (approved):

PR #1709, opened 2026-08-04
fix/added a new filter, to check employee work location 

umeshaliyanage:
### Summary

Employees with no assigned work location get every location's holidays applied to their leave calculations, instead of only company wide holidays in apply leave.

So we added a filter applyEmployeeWorkLocationFilter to filter to see if the employee has a work location and is used to handle the situation where an employee who doesn't have a location not get other location's holidays 

applyEmployeeWorkLocationFilter -
If workLocationId is absent from the incoming HolidayFilterDto, the method populates it from the authenticated user's context before the query runs

### How to test

1.Make a user and have them set to no location
2. Have holidays set to specific locations
3. When applying for leaves should not be able to see location specific holidays

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/backend/src/main/java/com/skapp/community/peopleplanner/constant/PeopleConstants.java
+++ b/backend/src/main/java/com/skapp/community/peopleplanner/constant/PeopleConstants.java
@@ -13,6 +13,8 @@ public class PeopleConstants {
 
 	public static final String HOLIDAY_ALL_WORK_LOCATIONS = "All locations";
 
+	public static final Long HOLIDAY_NO_WORK_LOCATION_ID = 0L;
+
 	public static final String FIELD_EMPLOYMENT_TYPE = "Employment Type";
 
 	public static final String FIELD_JOB_FAMILY = "Job Family";
--- a/backend/src/main/java/com/skapp/community/peopleplanner/constant/PeopleConstants.java
+++ b/backend/src/main/java/com/skapp/community/peopleplanner/constant/PeopleConstants.java
@@ -13,8 +13,6 @@ public class PeopleConstants {
 
 	public static final String HOLIDAY_ALL_WORK_LOCATIONS = "All locations";
 
-	public static final Long HOLIDAY_NO_WORK_LOCATION_ID = 0L;
-
 	public static final String FIELD_EMPLOYMENT_TYPE = "Employment Type";
 
 	public static final String FIELD_JOB_FAMILY = "Job Family";
--- a/backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
@@ -549,8 +549,9 @@ public class HolidayServiceImpl implements HolidayService {
 		Employee currentEmployee = currentUser.getEmployee();
 		WorkLocation workLocation = currentEmployee != null ? currentEmployee.getWorkLocation() : null;
 
-		holidayFilterDto.setWorkLocationId(
-				workLocation != null ? workLocation.getWorkLocationId() : PeopleConstants.HOLIDAY_NO_WORK_LOCATION_ID);
+		if (workLocation != null) {
+			holidayFilterDto.setWorkLocationId(workLocation.getWorkLocationId());
+		}
 	}
 
 	private Set<WorkLocation> resolveWorkLocations(List<String> workLocationNames,
--- a/backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
@@ -1,12 +1,14 @@
 package com.skapp.community.peopleplanner.service.impl;
 
 import com.skapp.community.common.exception.ModuleException;
+import com.skapp.community.common.model.User;
 import com.skapp.community.common.model.WorkLocation;
 import com.skapp.community.common.payload.response.BulkStatusSummary;
 import com.skapp.community.common.payload.response.PageDto;
 import com.skapp.community.common.payload.response.ResponseEntityDto;
 import com.skapp.community.common.repository.WorkLocationDao;
 import com.skapp.community.common.service.OrganizationService;
+import com.skapp.community.common.service.UserService;
 import com.skapp.community.common.type.Role;
 import com.skapp.community.common.util.CommonModuleUtils;
 import com.skapp.community.common.util.DateTimeUtils;
@@ -97,6 +99,8 @@ public class HolidayServiceImpl implements HolidayService {
 
 	private final EmployeeDao employeeDao;
 
+	private final UserService userService;
+
 	@Override
 	public ResponseEntityDto getAllHolidays(HolidayFilterDto holidayFilterDto) {
 		log.info("getAllHolidays: execution started");
@@ -109,6 +113,8 @@ public class HolidayServiceImpl implements HolidayService {
 
 		Pageable pageable = PageRequest.of(holidayFilterDto.getPage(), pageSize,
 				Sort.by(holidayFilterDto.getSortOrder(), holidayFilterDto.getSortKey().toString()));
+
+		applyEmployeeWorkLocationFilter(holidayFilterDto);
 
 		Page<Holiday> holidays = holidayDao.findAllHolidays(holidayFilterDto, pageable);
 		PageDto pageDto = pageTransformer.transform(holidays);
@@ -532,6 +538,19 @@ public class HolidayServiceImpl implements HolidayService {
 	private boolean canDeleteHoliday(Holiday holiday) {
 		LocalDate currentDate = DateTimeUtils.getCurrentUtcDate();
 		return holiday.getDate().isAfter(currentDate);
+	}
+
+	private void applyEmployeeWorkLocationFilter(HolidayFilterDto holidayFilterDto) {
+		if (holidayFilterDto.getWorkLocationId() != null) {
+			return;
+		}
+
+		User currentUser = userService.getCurrentUser();
+		Employee currentEmployee = currentUser.getEmployee();
+		WorkLocation workLocation = currentEmployee != null ? currentEmployee.getWorkLocation() : null;
+
+		holidayFilterDto.setWorkLocationId(
+				workLocation != null ? workLocation.getWorkLocationId() : PeopleConstants.HOLIDAY_NO_WORK_LOCATION_ID);
 	}
 
 	private Set<WorkLocation> resolveWorkLocations(List<String> workLocationNames,
--- a/backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
@@ -1,14 +1,12 @@
 package com.skapp.community.peopleplanner.service.impl;
 
 import com.skapp.community.common.exception.ModuleException;
-import com.skapp.community.common.model.User;
 import com.skapp.community.common.model.WorkLocation;
 import com.skapp.community.common.payload.response.BulkStatusSummary;
 import com.skapp.community.common.payload.response.PageDto;
 import com.skapp.community.common.payload.response.ResponseEntityDto;
 import com.skapp.community.common.repository.WorkLocationDao;
 import com.skapp.community.common.service.OrganizationService;
-import com.skapp.community.common.service.UserService;
 import com.skapp.community.common.type.Role;
 import com.skapp.community.common.util.CommonModuleUtils;
 import com.skapp.community.common.util.DateTimeUtils;
@@ -98,8 +96,6 @@ public class HolidayServiceImpl implements HolidayService {
 	private final WorkLocationDao workLocationDao;
 
 	private final EmployeeDao employeeDao;
-
-	private final UserService userService;
 
 	@Override
 	public ResponseEntityDto getAllHolidays(HolidayFilterDto holidayFilterDto) {
--- a/backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
@@ -114,8 +114,6 @@ public class HolidayServiceImpl implements HolidayService {
 		Pageable pageable = PageRequest.of(holidayFilterDto.getPage(), pageSize,
 				Sort.by(holidayFilterDto.getSortOrder(), holidayFilterDto.getSortKey().toString()));
 
-		applyEmployeeWorkLocationFilter(holidayFilterDto);
-
 		Page<Holiday> holidays = holidayDao.findAllHolidays(holidayFilterDto, pageable);
 		PageDto pageDto = pageTransformer.transform(holidays);
 		List<HolidayResponseDto> list = peopleMapper
@@ -538,20 +536,6 @@ public class HolidayServiceImpl implements HolidayService {
 	private boolean canDeleteHoliday(Holiday holiday) {
 		LocalDate currentDate = DateTimeUtils.getCurrentUtcDate();
 		return holiday.getDate().isAfter(currentDate);
-	}
-
-	private void applyEmployeeWorkLocationFilter(HolidayFilterDto holidayFilterDto) {
-		if (holidayFilterDto.getWorkLocationId() != null) {
-			return;
-		}
-
-		User currentUser = userService.getCurrentUser();
-		Employee currentEmployee = currentUser.getEmployee();
-		WorkLocation workLocation = currentEmployee != null ? currentEmployee.getWorkLocation() : null;
-
-		if (workLocation != null) {
-			holidayFilterDto.setWorkLocationId(workLocation.getWorkLocationId());
-		}
 	}
 
 	private Set<WorkLocation> resolveWorkLocations(List<String> workLocationNames,
--- a/backend/src/main/java/com/skapp/community/leaveplanner/service/impl/LeaveServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/leaveplanner/service/impl/LeaveServiceImpl.java
@@ -1039,7 +1039,7 @@ public class LeaveServiceImpl implements LeaveService {
 				&& employee.getWorkLocation().getWorkLocationId() != null) {
 			return holidayDao.findAllActiveHolidaysByWorkLocationId(employee.getWorkLocation().getWorkLocationId());
 		}
-		return holidayDao.findAllByIsActiveTrue();
+		return holidayDao.findAllByIsActiveTrueAndWorkLocationsIsEmpty();
 	}
 
 	private float getRemainingTotalHours(List<LeaveEntitlement> leaveEntitlements, LeaveRequest leaveRequest,
--- a/backend/src/main/java/com/skapp/community/peopleplanner/repository/HolidayDao.java
+++ b/backend/src/main/java/com/skapp/community/peopleplanner/repository/HolidayDao.java
@@ -17,6 +17,8 @@ public interface HolidayDao extends JpaRepository<Holiday, Long>, JpaSpecificati
 
 	List<Holiday> findAllByIsActiveTrue();
 
+	List<Holiday> findAllByIsActiveTrueAndWorkLocationsIsEmpty();
+
 	List<Holiday> findAllByIsActiveTrueAndDateBetween(LocalDate startDate, LocalDate endDate);
 
 	Holiday findByIsActiveTrueAndDate(LocalDate currentDate);
--- a/backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/HolidayRepositoryImpl.java
+++ b/backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/HolidayRepositoryImpl.java
@@ -131,7 +131,10 @@ public class HolidayRepositoryImpl implements HolidayRepository {
 			predicates.add(datePredicate);
 
 			Long workLocationId = holidayFilterDto.getWorkLocationId();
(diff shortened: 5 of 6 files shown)


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR makes `GET /v1/holiday` default the `workLocationId` filter to the calling user's own work location when the parameter is absent, adding a `HOLIDAY_NO_WORK_LOCATION_ID = 0L` sentinel for users with no work location. The core mechanism works with the existing Criteria predicate (0 is treated as a real id, so only holidays with an empty work-location set match), but the unconditional, role-agnostic default silently narrows results for every caller that omits the parameter — including admin/manager shared views and, most importantly, the very common case of employees whose `work_location_id` is NULL (the column is nullable and defaults to NULL). That case is also inconsistent with `LeaveServiceImpl.getHolidaysForEmployee`, which returns *all* active holidays when an employee has no work location, and the existing holiday integration test looks like it will now fail against the seeded data.

Found **12** new issue(s): 🔴 5 important, 🟡 4 suggestion(s), 🟣 3 nit(s)

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
```diff
@@ -534,6 +540,19 @@ private boolean canDeleteHoliday(Holiday holiday) {
...
+	private void applyEmployeeWorkLocationFilter(HolidayFilterDto holidayFilterDto) {
+		if (holidayFilterDto.getWorkLocationId() != null) {
+			return;
+		}
+
+		User currentUser = userService.getCurrentUser();
+		Employee currentEmployee = currentUser.getEmployee();
+		WorkLocation workLocation = currentEmployee != null ? currentEmployee.getWorkLocation() : null;
+
+		holidayFilterDto.setWorkLocationId(
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Employees with no work location now see an almost-empty holiday list. `employee.work_location_id` is nullable and defaults to NULL (see people-ddl-script-v1-alter-table-employee-add-work-location-id.sql), so on any existing deployment most employees have no work location until one is explicitly assigned. For those users this sets `workLocationId = 0`, and `HolidayRepositoryImpl.buildPredicates` (line 133-140) treats it as a real id: the predicate becomes `join.workLocationId = 0 OR holiday.workLocations IS EMPTY`, i.e. only holidays that target *no* location at all are returned. Every location-scoped holiday disappears from the calendar and from the leave-application modal. This also contradicts the established rule in `LeaveServiceImpl.getHolidaysForEmployee` (LeaveServiceImpl.java:1037-1042), which falls back to `findAllByIsActiveTrue()` (all holidays) when the employee has no work location — so leave duration calculations will still deduct holidays the employee can no longer see. Fix: when the employee has no work location, leave the filter unset (or set `PeopleConstants.HOLIDAY_ALL_WORK_LOCATIONS_ID`, which the repository explicitly skips) so all holidays are returned, matching the leave-planner behaviour; then drop the 0 sentinel. If the product decision really is "only global holidays", change `LeaveServiceImpl.getHolidaysForEmployee` in the same PR so the two paths agree.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
```diff
@@ -110,6 +114,8 @@ public ResponseEntityDto getAllHolidays(HolidayFilterDto holidayFilterDto) {
 		Pageable pageable = PageRequest.of(holidayFilterDto.getPage(), pageSize,
 				Sort.by(holidayFilterDto.getSortOrder(), holidayFilterDto.getSortKey().toString()));
 
+		applyEmployeeWorkLocationFilter(holidayFilterDto);
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The default is applied unconditionally for every caller and every role, which silently narrows results in views that are not "my holidays". Concrete case: `useGetHoliday()` (frontend/src/community/common/hooks/useGetHoliday.ts) calls `useGetAllHolidays()` with no parameters and feeds `EmployeeTimeRecordsTable`, which is rendered by `TimesheetAnalytics` in manager mode — a manager/admin in location A inspecting an employee in location B will now have holidays resolved from the *viewer's* location instead of the subject's (previously all holidays were returned, so the highlight was at least never missing). The same applies to any non-UI API client (integrations, exports with `isExport=true`) that omits `workLocationId`: its response set changes without any version bump or response field indicating a filter was auto-applied. Recommended fix: make the default role-aware — skip it for admin roles using `userService.getCurrentUserRoles()` / `AuthUtil.withRolePrefix(...)` as `AttendanceConfigServiceImpl.java:97` already does — and/or have the affected frontend callers pass `ALL_LOCATIONS_ID` explicitly (the admin holiday table already does via `holidayDataFilterSlice`). At minimum, document the new default on the `@Operation` description of `HolidayController.getAllHolidays` since it is an API contract change.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
```diff
@@ -534,6 +540,19 @@ private boolean canDeleteHoliday(Holiday holiday) {
 		return holiday.getDate().isAfter(currentDate);
 	}
 
+	private void applyEmployeeWorkLocationFilter(HolidayFilterDto holidayFilterDto) {
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

No test changes accompany this behaviour change, and an existing test looks like it will now fail. `HolidayControllerIntegrationTest.getAllHolidays_ReturnsSuccessful` performs a GET with no `workLocationId` as user1@gmail.com; in src/test/resources/data.sql the employee rows are inserted without `work_location_id` (so it is NULL) while every seeded holiday is mapped in `shr_com_work_location_ppl_holiday`. The new default therefore yields `workLocationId = 0`, no holiday matches, the count query returns 0, and the assertions on `items[0]['id']` / `['date']` / `['name']` will fail. Please run the holiday integration tests, and add coverage for the three new branches: (a) explicit `workLocationId` is honoured untouched, (b) employee with a work location gets that location plus global holidays, (c) employee with no work location — asserting whatever behaviour you settle on after the fallback question above.

backend/src/main/java/com/skapp/community/peopleplanner/constant/PeopleConstants.java
```diff
@@ -13,6 +13,8 @@ public class PeopleConstants {
 
 	public static final String HOLIDAY_ALL_WORK_LOCATIONS = "All locations";
 
+	public static final Long HOLIDAY_NO_WORK_LOCATION_ID = 0L;
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`HOLIDAY_NO_WORK_LOCATION_ID = 0L` is a magic sentinel injected into a field (`HolidayFilterDto.workLocationId`) that otherwise holds real `com_work_location.id` values, and it only produces the intended result because `HolidayRepositoryImpl.buildPredicates` happens to OR in `holiday.workLocations IS EMPTY`. Two problems: the meaning is invisible at the query site (someone tightening that predicate later would silently break this feature), and the name reads like an id when it is really a mode. Prefer an explicit intent flag on the filter DTO (e.g. `Boolean onlyGlobalHolidays`) handled as its own predicate in the repository, or reuse `HOLIDAY_ALL_WORK_LOCATIONS_ID` semantics if the fallback becomes "return everything". If the sentinel is kept, at least add a comment stating it relies on work location ids starting at 1.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
```diff
@@ -97,6 +99,8 @@ public class HolidayServiceImpl implements HolidayService {
 
 	private final EmployeeDao employeeDao;
 
+	private final UserService userService;
+
 	@Override
 	public ResponseEntityDto getAllHolidays(HolidayFilterDto holidayFilterDto) {
```
ThinuwanW, line 101: 🤖 **Claude** · 🟣 **Nit**

`getAllHolidays` is not transactional, and the new helper dereferences the LAZY `@ManyToOne` `Employee.workLocation` (Employee.java:173-175) on an entity loaded by `userService.getCurrentUser()` in its own transaction. This only works because `spring.jpa.open-in-view` is left at its Spring default of true — note that the enterprise configs set it explicitly (`application-ep-prd.yml`, `application-ep-non-prd.yml`) while the community profiles do not, so anyone disabling OSIV would get a `LazyInitializationException` here rather than in an obviously related place. Adding `@Transactional(readOnly = true)` to `getAllHolidays` (consistent with the read-only guidance) removes the dependency.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
```diff
@@ -534,6 +540,19 @@ private boolean canDeleteHoliday(Holiday holiday) {
 		return holiday.getDate().isAfter(currentDate);
 	}
 
+	private void applyEmployeeWorkLocationFilter(HolidayFilterDto holidayFilterDto) {
+		if (holidayFilterDto.getWorkLocationId() != null) {
+			return;
+		}
+
+		User currentUser = userService.getCurrentUser();
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

The helper silently rewrites the caller's filter with no trace: nothing is logged and the response (`PageDto`) does not report the `workLocationId` that was actually applied. When a user reports "my holidays disappeared", there is no way to tell from the logs whether the default kicked in and with which value. Add an entry/decision log in line with the module convention, e.g. `log.info("applyEmployeeWorkLocationFilter: defaulting workLocationId to {}", resolvedId)`, and consider echoing the effective `workLocationId` back in the response so clients can distinguish "no holidays" from "filtered to your location".

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The no-work-location fallback contradicts the rule that already exists in this codebase for the same concept. LeaveServiceImpl.getHolidaysForEmployee (LeaveServiceImpl.java:1037-1043) resolves holidays for an employee with the identical logic — if employee != null && workLocation != null && workLocationId != null, query by work location — but when there is no work location it falls back to holidayDao.findAllByIsActiveTrue(), i.e. ALL active holidays. The new helper falls back to a sentinel that matches no work location, i.e. only all-location holidays. Result: for any employee without a work location, leave duration/entitlement calculations count location-specific holidays as non-working days while the holiday list endpoint hides those very holidays, so the number of leave days deducted cannot be reconciled with what the employee can see. Two fixes are needed: (1) make the fallback agree with the leave module — either leave workLocationId null or set PeopleConstants.HOLIDAY_ALL_WORK_LOCATIONS_ID so the no-location case means "all holidays", and (2) stop duplicating the resolution rule — expose one shared method (e.g. on HolidayService, or a small util in peopleplanner/util) that returns the effective work-location id for an employee and have both LeaveServiceImpl and this class call it, so the two can never drift again.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
```diff
@@ -534,6 +540,19 @@ private boolean canDeleteHoliday(Holiday holiday) {
 		return holiday.getDate().isAfter(currentDate);
 	}
 
+	private void applyEmployeeWorkLocationFilter(HolidayFilterDto holidayFilterDto) {
+		if (holidayFilterDto.getWorkLocationId() != null) {
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Because the helper returns early whenever the caller already supplied a workLocationId, the new behaviour is a default, not a restriction — it enforces nothing. A ROLE_PEOPLE_EMPLOYEE token can call GET /v1/holiday?workLocationId=<any other location id> to read another location's holidays, or ?workLocationId=-1 (HOLIDAY_ALL_WORK_LOCATIONS_ID, which buildPredicates treats as "no location predicate") to read every holiday in the org, fully bypassing the intended scoping. Decide which of the two this is and make it explicit: if it is genuine per-employee scoping, resolve the caller's roles with userService.getCurrentUserRoles() and for non-admin roles OVERWRITE workLocationId with the employee's own value (ignoring/rejecting the request parameter) rather than only filling it in when absent; if it is purely a UX default for the "my holidays" view, it does not belong in the service at all — the caller should pass the value, and the current unconditional mutation is what causes the admin-view narrowing.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

This call guarantees holidayFilterDto.getWorkLocationId() is never null from here on, which permanently enables a query path in HolidayRepositoryImpl.buildPredicates (HolidayRepositoryImpl.java:133-140) that was previously taken only when a client explicitly filtered by location. That branch calls root.join(Holiday_.workLocations, LEFT) while the data query has already called root.fetch(Holiday_.workLocations, LEFT) at HolidayRepositoryImpl.java:55. Hibernate does not merge a fetch join with a separate join on the same attribute, so every list and export request now produces two independent left joins over shr_com_work_location_ppl_holiday — an N x N row explosion per holiday that DISTINCT then has to collapse, and it is worst on the isExport path where pageSize is set to the full holiday count. Fix by not adding a second join: either cast and reuse the existing fetch join, e.g. Join<Holiday, WorkLocation> join = (Join<Holiday, WorkLocation>) root.fetch(Holiday_.workLocations, JoinType.LEFT) and build the predicate on that alias, or replace the join + OR-isEmpty predicate with an exists/not-exists subquery over the join table so the root query stays single-row-per-holiday.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
```diff
@@ -1,12 +1,14 @@
 package com.skapp.community.peopleplanner.service.impl;
 
 import com.skapp.community.common.exception.ModuleException;
+import com.skapp.community.common.model.User;
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Missing paired changes: work-location scoping was added to getAllHolidays only, while every other holiday read path still returns holidays for all locations. Inside this same class, getHolidaysByDate (line 202) calls findAllActiveHolidaysByDateWithWorkLocations(date) with no location predicate. Outside it, TimeServiceImpl:372/456/1163/1276/1563, TimeAnalyticsServiceImpl:238/336/427, LeaveAnalyticsServiceImpl (many, e.g. 334/437/891) and EpLeaveCalendarServiceImpl:126/297 all use findAllByIsActiveTrue / findAllByIsActiveTrueAndDate(Between). The user-visible contradiction is concrete: a holiday declared only for location B will still be treated as a non-working day in an employee-of-location-A's timesheet and attendance analytics, and will still appear in leave-calendar data, but that employee can no longer see it in the holiday list — so they cannot explain the gap in their own timesheet. Either scope the calculation paths the same way (they all have the employee in hand, so they can reuse the shared resolver suggested above), or explicitly document in the PR which reads are intentionally org-wide, and confirm getHolidaysByDate staying unscoped is deliberate (it is admin-only today, so that one is defensible).

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Three extra database round-trips are now added to every holiday list request just to read a single id: userService.getCurrentUser() does a userDao.findByEmail lookup (UserServiceImpl.java:27-33), User.employee is a @OneToOne(mappedBy="user") which defaults to EAGER and costs a secondary select, and Employee.workLocation is @ManyToOne(fetch = LAZY) (Employee.java:173-175) so reading its id initialises the proxy with a third select. This endpoint is polled by the calendar/dashboard hooks (useGetHoliday -> useGetAllHolidays with no arguments), so it is a hot path. Prefer a single projection: add a derived query to EmployeeDao such as Optional<Long> findWorkLocationWorkLocationIdByEmployeeId(Long employeeId) (EmployeeDao already has findByWorkLocationWorkLocationId variants, so the property path is proven) and resolve the id in one query — that also removes this method's dependence on an open persistence session for lazy proxy initialisation.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Naming and placement make the side effect easy to miss. The name applyEmployeeWorkLocationFilter reads as if it applies a filter to a query, but the method actually mutates the caller's request DTO and returns nothing; a reviewer of getAllHolidays sees a void call with no visible effect on the following line. Prefer a pure resolver, e.g. Long effectiveWorkLocationId = resolveEffectiveWorkLocationId(holidayFilterDto.getWorkLocationId()), and pass the resolved value explicitly (or set it in one obvious place). Also move the call above the pageable construction at line 114 so all filter normalisation happens before any pagination/sort setup, keeping the method's phases in a readable order.

backend/src/main/java/com/skapp/community/peopleplanner/constant/PeopleConstants.java
umeshaliyanage: Removed Long HOLIDAY_NO_WORK_LOCATION_ID = 0L from constants instead removed its usage and just check and sets work location to workLocation.getWorkLocationId()

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
umeshaliyanage: Changed the data.sql seed to accompany the change we made by making holiday 6 a global holiday

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
umeshaliyanage: Needed to add the filter in predicate method too

LeaveServiceImpl.getHolidaysForEmployee no longer falls back to findAllByIsActiveTrue(); it uses findAllByIsActiveTrueAndWorkLocationsIsEmpty(), so duration calculations deduct exactly the holidays the employee sees in the calendar and leave modal 

 Dropped constant entirely and remove the default fallback
 
HolidayRepositoryImpl.buildPredicates handles null explicitly: workLocations IS EMPTY only. Non-null keeps the existing location = id OR workLocations IS EMPTY, so employees who do have a work location are unaffected.     
                                                       
HOLIDAY_ALL_WORK_LOCATIONS_ID (-1) still means "all locations" for the admin view and is unchanged.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
umeshaliyanage: Not needed

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
umeshaliyanage: removed service level filter

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
umeshaliyanage: Removed service level filter

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
umeshaliyanage: removed

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
umeshaliyanage: removed

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
umeshaliyanage: removed

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
umeshaliyanage: removed

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
umeshaliyanage: removed

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
umeshaliyanage: out of scope

hasalarootcode (approved):

shakila-kamalasena (approved):

akilarootcode (approved):

PR #1712, opened 2026-08-05
fix: signup redirect subdomain bug

SanithuGunasena07:
Bug: Sign up redirect uses previous tenant's subdomain

Root cause: The code assumed the current subdomain was always app, but nothing guaranteed that so stale tenant subdomains leaked into the new tenant's URL.

Fix: Added a check in Next.js middleware that intercepts every request to /signup before the page renders. It reads the Host header of the incoming request, extracts the current subdomain, and if it isn't app, strips that subdomain and issues an HTTP redirect to app.<basedomain>/signup instead. If it's already app, the request just passes through. The logic itself lives in a small utility function (getSignUpTenantRedirect), and middleware.ts just calls it for that one route.

### DIFF
--- a/frontend/middleware.ts
+++ b/frontend/middleware.ts
@@ -18,11 +18,7 @@ import {
   SenderTypes,
   SuperAdminType
 } from "~community/common/types/AuthTypes";
-import {
-  checkRestrictedRoutesAndRedirect,
-  getSignUpTenantRedirect,
-  isEnterpriseMode
-} from "~community/common/utils/commonUtil";
+import { checkRestrictedRoutesAndRedirect } from "~community/common/utils/commonUtil";
 import { TenantStatusEnums } from "~enterprise/common/enums/Common";
 import { isCoreOrProTier } from "~enterprise/common/utils/commonUtil";
 
@@ -202,18 +198,12 @@ const allowedRoutes: Record<
 };
 
 export function middleware(request: NextRequest) {
-  const currentPath = request.nextUrl.pathname;
-
-  if (isEnterpriseMode() && currentPath === ROUTES.AUTH.SIGNUP) {
-    const signUpRedirect = getSignUpTenantRedirect(request);
-    if (signUpRedirect) return signUpRedirect;
-    return NextResponse.next();
-  }
-
   // Get accessToken from cookies
   const token = request.cookies.get("accessToken")?.value;
 
   const claims = extractClaimsFromToken(token || "");
+
+  const currentPath = request.nextUrl.pathname;
 
   if (
     currentPath === ROUTES.SIGN.DOCUMENT_ACCESS ||
@@ -388,8 +378,6 @@ export const config = {
   matcher: [
     // All community routes
     "/community/:path*",
-    // Auth routes
-    "/signup",
     // Super admin routes
     "/setup-organization/:path*",
     "/module-selection",
--- a/frontend/middleware.ts
+++ b/frontend/middleware.ts
@@ -20,7 +20,8 @@ import {
 } from "~community/common/types/AuthTypes";
 import {
   checkRestrictedRoutesAndRedirect,
-  getSignUpTenantRedirect
+  getSignUpTenantRedirect,
+  isEnterpriseMode
 } from "~community/common/utils/commonUtil";
 import { TenantStatusEnums } from "~enterprise/common/enums/Common";
 import { isCoreOrProTier } from "~enterprise/common/utils/commonUtil";
@@ -201,16 +202,18 @@ const allowedRoutes: Record<
 };
 
 export function middleware(request: NextRequest) {
+  const currentPath = request.nextUrl.pathname;
+
+  if (isEnterpriseMode() && currentPath === ROUTES.AUTH.SIGNUP) {
+    const signUpRedirect = getSignUpTenantRedirect(request);
+    if (signUpRedirect) return signUpRedirect;
+    return NextResponse.next();
+  }
+
   // Get accessToken from cookies
   const token = request.cookies.get("accessToken")?.value;
 
   const claims = extractClaimsFromToken(token || "");
-
-  const currentPath = request.nextUrl.pathname;
-
-  if (currentPath.startsWith(ROUTES.AUTH.SIGNUP)) {
-    return getSignUpTenantRedirect(request);
-  }
 
   if (
     currentPath === ROUTES.SIGN.DOCUMENT_ACCESS ||
--- a/frontend/middleware.ts
+++ b/frontend/middleware.ts
@@ -18,7 +18,10 @@ import {
   SenderTypes,
   SuperAdminType
 } from "~community/common/types/AuthTypes";
-import { checkRestrictedRoutesAndRedirect } from "~community/common/utils/commonUtil";
+import {
+  checkRestrictedRoutesAndRedirect,
+  getSignUpTenantRedirect
+} from "~community/common/utils/commonUtil";
 import { TenantStatusEnums } from "~enterprise/common/enums/Common";
 import { isCoreOrProTier } from "~enterprise/common/utils/commonUtil";
 
@@ -205,6 +208,10 @@ export function middleware(request: NextRequest) {
 
   const currentPath = request.nextUrl.pathname;
 
+  if (currentPath.startsWith(ROUTES.AUTH.SIGNUP)) {
+    return getSignUpTenantRedirect(request);
+  }
+
   if (
     currentPath === ROUTES.SIGN.DOCUMENT_ACCESS ||
     currentPath.startsWith(ROUTES.SIGN.SIGN) ||
@@ -378,6 +385,8 @@ export const config = {
   matcher: [
     // All community routes
     "/community/:path*",
+    // Auth routes
+    "/signup",
     // Super admin routes
     "/setup-organization/:path*",
     "/module-selection",
--- a/frontend/src/community/common/utils/commonUtil.ts
+++ b/frontend/src/community/common/utils/commonUtil.ts
@@ -3,12 +3,14 @@ import { NextRequest, NextResponse } from "next/server";
 
 import {
   APP,
+  LOCALHOST,
   characterLengths
 } from "~community/common/constants/stringConstants";
 import { HOURS_PER_DAY } from "~community/common/constants/timeConstants";
 import {
   alphaNumericNamePatternWithSpecialCharacters,
   containsUnicode,
+  isValidIPv4Address,
   matchInvalidEmailCharactersSearchPattern,
   removeNonAlphaNumericCharactersPattern
 } from "~community/common/regex/regexPatterns";
@@ -526,11 +528,21 @@ export const checkRestrictedRoutesAndRedirect = (
 
 export const getSignUpTenantRedirect = (
   request: NextRequest
-): NextResponse => {
-  const host = request.headers.get("host") || "";
+): NextResponse | null => {
+  const host = (request.headers.get("host") || "").toLowerCase();
+  const hostname = host.split(":")[0];
+
+  if (
+    !host.includes(".") ||
+    hostname === LOCALHOST ||
+    isValidIPv4Address().test(hostname)
+  ) {
+    return null;
+  }
+
   const tenant = host.split(".")[0];
 
-  if (tenant === APP) return NextResponse.next();
+  if (tenant === APP) return null;
 
   const baseHost = host.split(".").slice(1).join(".");
   const url = request.nextUrl.clone();
--- a/frontend/src/community/common/utils/commonUtil.ts
+++ b/frontend/src/community/common/utils/commonUtil.ts
@@ -1,16 +1,11 @@
 import { SxProps, Theme } from "@mui/material";
 import { NextRequest, NextResponse } from "next/server";
 
-import {
-  APP,
-  LOCALHOST,
-  characterLengths
-} from "~community/common/constants/stringConstants";
+import { characterLengths } from "~community/common/constants/stringConstants";
 import { HOURS_PER_DAY } from "~community/common/constants/timeConstants";
 import {
   alphaNumericNamePatternWithSpecialCharacters,
   containsUnicode,
-  isValidIPv4Address,
   matchInvalidEmailCharactersSearchPattern,
   removeNonAlphaNumericCharactersPattern
 } from "~community/common/regex/regexPatterns";
@@ -524,31 +519,6 @@ export const checkRestrictedRoutesAndRedirect = (
     );
   }
   return null;
-};
-
-export const getSignUpTenantRedirect = (
-  request: NextRequest
-): NextResponse | null => {
-  const host = (request.headers.get("host") || "").toLowerCase();
-  const hostname = host.split(":")[0];
-
-  if (
-    !host.includes(".") ||
-    hostname === LOCALHOST ||
-    isValidIPv4Address().test(hostname)
-  ) {
-    return null;
-  }
-
-  const tenant = host.split(".")[0];
-
-  if (tenant === APP) return null;
-
-  const baseHost = host.split(".").slice(1).join(".");
-  const url = request.nextUrl.clone();
-  url.host = `${APP}.${baseHost}`;
-
-  return NextResponse.redirect(url);
 };
 
 export const getLabelForReadOnlyChip = (
--- a/frontend/src/community/common/utils/commonUtil.ts
+++ b/frontend/src/community/common/utils/commonUtil.ts
@@ -1,7 +1,10 @@
 import { SxProps, Theme } from "@mui/material";
 import { NextRequest, NextResponse } from "next/server";
 
-import { characterLengths } from "~community/common/constants/stringConstants";
+import {
+  APP,
+  characterLengths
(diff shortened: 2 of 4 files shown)


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds a middleware rule that forces `/signup` onto the `app.<base-domain>` subdomain by parsing the raw `Host` header. The core idea is sound for enterprise multi-tenant deployments, but the helper has no guards: it runs in community/self-hosted mode too, produces bogus targets for hosts with fewer than three labels (`localhost:3000` → `app.`, `skapp.com` → `app.com`), and builds the redirect target from an unvalidated, attacker-controllable `Host` header. It also duplicates the existing `getAppSubdomainUrl` helper in the enterprise utils.

Found **14** new issue(s): 🔴 5 important, 🟡 5 suggestion(s), 🟣 4 nit(s)

frontend/middleware.ts
```diff
@@ -205,6 +208,10 @@ export function middleware(request: NextRequest) {
 
   const currentPath = request.nextUrl.pathname;
 
+  if (currentPath.startsWith(ROUTES.AUTH.SIGNUP)) {
```

This tenant redirect is applied unconditionally, but `/signup` exists in community (self-hosted) mode too — `next.config.mjs` rewrites `/signup` to `/community/signup`, and `pages/index.tsx` sends fresh installs there via `handleCommunityNavigation` when `!setupStatus.isSignUpCompleted`. Community deployments have no `app.` subdomain (`example.env` doesn't even set `NEXT_PUBLIC_MODE`, so community is the default), so a self-hosted instance on `example.com` or `localhost:3000` will now bounce the very first onboarding screen to a non-existent host. Gate the branch on mode, e.g. `if (isEnterpriseMode() && currentPath.startsWith(ROUTES.AUTH.SIGNUP))` using the existing `isEnterpriseMode()` from `~community/common/utils/commonUtil`, or move the helper to `~enterprise/common/utils/commonUtil` with a no-op fallback stub in `src/fallback/` (the same pattern already used for `isCoreOrProTier` in this file).

frontend/src/community/common/utils/commonUtil.ts
```diff
@@ -521,6 +524,21 @@ export const checkRestrictedRoutesAndRedirect = (
...
 
+export const getSignUpTenantRedirect = (
+  request: NextRequest
+): NextResponse => {
+  const host = request.headers.get("host") || "";
+  const tenant = host.split(".")[0];
+
+  if (tenant === APP) return NextResponse.next();
+
+  const baseHost = host.split(".").slice(1).join(".");
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`host.split('.').slice(1).join('.')` silently produces garbage for hosts with fewer than three labels. `localhost:3000` → baseHost `""` → `url.host = "app."` → redirect to `http://app./signup` (breaks all local dev and any single-label host). Worse, an apex domain like `skapp.com` → baseHost `"com"` → `url.host = "app.com"`, redirecting the user to an unrelated third-party domain and leaking the signup query string (`plan`, `tier`, `partnerId`) with it. Add an early bail-out before building the target: return `NextResponse.next()` when `host.split('.').length < 3`, when the hostname is `localhost` (the `LOCALHOST` constant is already defined in `stringConstants.ts` and currently unused), or when the host is a bare IP.

frontend/src/community/common/utils/commonUtil.ts
```diff
@@ -521,6 +524,21 @@ export const checkRestrictedRoutesAndRedirect = (
   return null;
 };
 
+export const getSignUpTenantRedirect = (
+  request: NextRequest
+): NextResponse => {
+  const host = request.headers.get("host") || "";
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The redirect target is derived entirely from the raw, unvalidated `Host` header. A request with `Host: x.attacker.com` yields a 307 to `https://app.attacker.com/signup` — a host-header-injection open redirect from a public, unauthenticated route. There is also an inconsistency: the tenant/base host is parsed from `request.headers.get("host")` while the URL being mutated comes from `request.nextUrl.clone()`, which resolves through `x-forwarded-host`/`x-forwarded-proto`; behind a proxy where those differ, the two disagree and the redirect goes somewhere neither expected. Fix by reading `request.nextUrl.hostname` for consistency and validating the derived base domain against an allowlist (e.g. a `NEXT_PUBLIC_BASE_DOMAIN` env var), returning `NextResponse.next()` when it doesn't match.

frontend/src/community/common/utils/commonUtil.ts
```diff
@@ -521,6 +524,21 @@ export const checkRestrictedRoutesAndRedirect = (
...
+  const tenant = host.split(".")[0];
+
+  if (tenant === APP) return NextResponse.next();
+
+  const baseHost = host.split(".").slice(1).join(".");
+  const url = request.nextUrl.clone();
+  url.host = `${APP}.${baseHost}`;
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

This duplicates `getAppSubdomainUrl` in `~enterprise/common/utils/commonUtil.ts` (line 59), which computes the identical `host.split('.').slice(1).join('.')` + `${APP}.${host}` string on the client. Two copies of the subdomain-derivation rule will drift — and the guards added for the issues above would have to be written twice. Extract a single `buildAppSubdomainHost(host: string): string | null` helper (returning `null` when the host isn't eligible) and have both this function and `getAppSubdomainUrl` call it.

frontend/src/community/common/utils/commonUtil.ts
```diff
@@ -521,6 +524,21 @@ export const checkRestrictedRoutesAndRedirect = (
   return null;
 };
 
+export const getSignUpTenantRedirect = (
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Naming and return contract don't match the established pattern. `getSignUpTenantRedirect` returns a `NextResponse` even in the no-redirect case (`NextResponse.next()`), whereas its sibling directly above, `checkRestrictedRoutesAndRedirect`, returns `NextResponse | null` and lets the caller decide how to continue. Prefer `NextResponse | null` here too (return `null` when the tenant is already `app`) and have `middleware.ts` do `const signUpRedirect = getSignUpTenantRedirect(request); if (signUpRedirect) return signUpRedirect; return NextResponse.next();` — that also makes the "bail out" cases from the findings above express themselves naturally as `null`.

frontend/src/community/common/utils/commonUtil.ts
```diff
@@ -521,6 +524,21 @@ export const checkRestrictedRoutesAndRedirect = (
...
+  const tenant = host.split(".")[0];
+
+  if (tenant === APP) return NextResponse.next();
+
+  const baseHost = host.split(".").slice(1).join(".");
+  const url = request.nextUrl.clone();
+  url.host = `${APP}.${baseHost}`;
+
+  return NextResponse.redirect(url);
+};
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

No test coverage for this helper, even though `src/community/common/utils/__test__/commonUtil.test.ts` already exists and the host-parsing logic here is exactly the kind of thing that needs it. Add cases for: host already on `app.` (expect pass-through), tenant subdomain `acme.skapp.com` (expect redirect to `app.skapp.com` with path and query preserved), apex domain `skapp.com`, `localhost:3000`, port preservation on `acme.localhost:3000`, and a missing/empty `Host` header.

frontend/middleware.ts
```diff
@@ -378,6 +385,8 @@ export const config = {
   matcher: [
     // All community routes
     "/community/:path*",
+    // Auth routes
+    "/signup",
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

The matcher entry is the exact path `"/signup"` while the guard at line 211 uses `startsWith(ROUTES.AUTH.SIGNUP)` — the broader check can never fire on anything but `/signup`, so use `currentPath === ROUTES.AUTH.SIGNUP` to keep the two in sync (or add `"/signup/:path*"` if sub-paths are intended). Separately, note the enforcement is bypassable: `/community/signup` is the rewrite destination and is already covered by the `"/community/:path*"` matcher, but it doesn't start with `/signup`, so it skips this branch entirely and falls through to the role gate.

Missing paired change: `config.matcher` is not private to the middleware. `src/community/auth/utils/authUtils.ts` imports `{ config }` from `../../../../middleware` and `IsAProtectedUrlWithDrawer()` treats every matcher entry's first path segment as a "protected route with drawer". Adding "/signup" makes `IsAProtectedUrlWithDrawer("/signup")` return true (it is not covered by `drawerHiddenProtectedRoutes`, and `ROUTES.SIGN.SIGN` is "/sign/sign" so the prefix check does not catch it). `BaseLayout` consumes that value for every page: when `sessionStatus === "authenticated"` (a user who still holds an `accessToken` cookie and lands on `app.<domain>/signup`, e.g. via the SignUpLayout link or a partner URL), the sign-up page will now render inside `ContentWithDrawer` — i.e. wrapped in the full app navigation shell instead of the auth layout — and the `useEffect` in BaseLayout will additionally fire `setDeviceToken(token)` on the sign-up page. Fix: add `ROUTES.AUTH.SIGNUP` to `drawerHiddenProtectedRoutes` in `src/community/auth/constants/routeConfigs.ts` (minimal fix), or better, stop deriving the drawer route list from `config.matcher` so middleware matcher changes can't leak into layout behaviour.

frontend/src/community/common/utils/commonUtil.ts
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The tenant is read from the raw `host` header while the redirect target is built from `request.nextUrl`, which are two different sources. Next.js resolves `nextUrl` using the forwarded headers (`x-forwarded-host` / `x-forwarded-proto`), so in any deployment where a proxy/ALB forwards the public host in `x-forwarded-host` and rewrites `host` to the internal service name, `host` becomes e.g. `internal-svc:3000` → `tenant` is "internal-svc:3000" (never equal to `APP`, so it always redirects) and `baseHost` becomes "local:3000"/"" → sign-up redirects to a non-existent host and is completely broken in production. Related: `nextUrl.protocol` can be `http` behind TLS termination, so the 307 can downgrade the scheme (contrast the client-side twin `getAppSubdomainUrl`, which correctly uses `window.location.protocol`). Fix: derive both the tenant and the base host from a single proxy-aware source — e.g. `const currentHost = request.headers.get("x-forwarded-host") ?? request.headers.get("host") ?? ""` (or simply `request.nextUrl.host`) — and set the protocol from `x-forwarded-proto` when present.

frontend/middleware.ts

The enforcement is path-shaped and therefore bypassable/inconsistent with the pages it protects. `/signup` is only a `next.config.mjs` rewrite onto `/enterprise/signup` (or `/community/signup`), and both real page paths are still directly reachable: `/enterprise/signup` is not in `config.matcher` at all, so a user visiting `https://<tenant>.<domain>/enterprise/signup` renders the sign-up page on a tenant subdomain with the tenant redirect entirely skipped — exactly the state this PR is trying to prevent (and the sign-up API calls would then carry the wrong tenant context). Conversely `/community/signup` *is* matched but the new guard does not recognise it, so it falls through to the role gate and unauthenticated users are bounced to `/signin`. Fix: either add the concrete page paths (`/enterprise/signup`, `/community/signup`) to the matcher and include them in the guard's condition, or check against a small array of sign-up paths rather than only `ROUTES.AUTH.SIGNUP`.

frontend/src/community/common/utils/commonUtil.ts
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

No self-redirect guard. The `URL.host` setter silently ignores values it cannot parse (it does not throw), so if `${APP}.${baseHost}` is not a valid host — which happens whenever the incoming `Host` header is malformed or unusual — `url` is left byte-identical to `request.nextUrl` and the middleware returns a 307 to the exact same URL, producing an infinite redirect loop (`ERR_TOO_MANY_REDIRECTS`) on the public sign-up page. Add a cheap guard before returning, e.g. `if (!baseHost || url.host === request.nextUrl.host) return NextResponse.next();` so a failed/no-op host rewrite degrades to serving the page rather than looping.

frontend/src/community/common/utils/commonUtil.ts
```diff
@@ -521,6 +524,21 @@ export const checkRestrictedRoutesAndRedirect = (
   return null;
 };
 
+export const getSignUpTenantRedirect = (
+  request: NextRequest
+): NextResponse => {
+  const host = request.headers.get("host") || "";
+  const tenant = host.split(".")[0];
+
+  if (tenant === APP) return NextResponse.next();
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`tenant === APP` is a case-sensitive comparison against a case-insensitive value. Per RFC, the `Host` header is case-insensitive and, while browsers normally lower-case it, proxies and non-browser clients do not guarantee it. `Host: App.skapp.com` would fail the equality check and incur a pointless extra redirect hop. Normalise once at the top: `const host = (request.headers.get("host") ?? "").toLowerCase();`.

frontend/src/community/common/utils/commonUtil.ts
```diff
@@ -1,7 +1,10 @@
 import { SxProps, Theme } from "@mui/material";
 import { NextRequest, NextResponse } from "next/server";
 
-import { characterLengths } from "~community/common/constants/stringConstants";
+import {
+  APP,
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

This PR adds the first consumer of the `APP` constant from `stringConstants.ts`, but its sibling `LOCALHOST` (line 85 of the same file, mirrored as `""` in `src/fallback/common/constants/stringConstants.ts`) has zero usages anywhere in the frontend — it appears to have been added for exactly this kind of host parsing and is now dead code. Either use it here to short-circuit local/dev hosts (`if (host.startsWith(LOCALHOST)) return NextResponse.next();`), which also removes the need for special-casing hostless dev URLs, or delete both the community and fallback declarations.

frontend/middleware.ts
```diff
@@ -205,6 +208,10 @@ export function middleware(request: NextRequest) {
 
   const currentPath = request.nextUrl.pathname;
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Ordering: the sign-up guard is placed after `extractClaimsFromToken(token || "")` on line 207, so every request to the public `/signup` route pays a base64-decode + `JSON.parse` of the JWT whose `claims` the guard never reads. Since this branch is fully independent of authentication, move the `currentPath.startsWith(ROUTES.AUTH.SIGNUP)` check above the cookie read and token decode (right after `const currentPath = ...`) so the public route does not depend on token handling at all.

frontend/middleware.ts
SanithuGunasena07: The redirect check now only runs when isEnterpriseMode() is true

frontend/src/community/common/utils/commonUtil.ts
SanithuGunasena07: Fixed the bare localhost:3000, localhost, and raw IP cases those now skip the redirect instead of building a broken URL.

frontend/src/community/common/utils/commonUtil.ts
SanithuGunasena07: not needed

frontend/src/community/common/utils/commonUtil.ts
SanithuGunasena07: matched the sibling function's pattern: returns null instead of NextResponse.next(), caller handles the fallback.

frontend/middleware.ts
SanithuGunasena07: tightened the check to currentPath === ROUTES.AUTH.SIGNUP to match the matcher's exact-match behavior

frontend/src/community/common/utils/commonUtil.ts
SanithuGunasena07: not needed

frontend/middleware.ts
SanithuGunasena07: Added ROUTES.AUTH.SIGNUP to drawerHiddenProtectedRoutes in routeConfigs.ts

frontend/src/community/common/utils/commonUtil.ts
SanithuGunasena07: not needed

frontend/src/community/common/utils/commonUtil.ts
SanithuGunasena07: host is now lowercased

frontend/src/community/common/utils/commonUtil.ts
SanithuGunasena07: Already used at line 537 (hostname === LOCALHOST)

frontend/middleware.ts
SanithuGunasena07: Moved the currentPath declaration and the signup check above the token/claims lines

PR #1717, opened 2026-08-07
refactor/pre-midnight-clock-out-alert-modal

Deshan1100:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR migrates PreMidnightClockOutAlertModal from MUI Box/Stack/Typography + styles.ts to plain Tailwind markup, and moves the modal title into the SmallModal `modalHeader` via a new branch in TimeWidgetPopupController.getModalTitle(). The refactor is directionally correct and matches the newer Tailwind modal pattern used elsewhere (e.g. EnableLeavePoliciesConfirmModal, UpgradeModal), but it breaks the existing co-located unit test, drops some styling that was previously applied, and the new title branch is ordered ahead of the auto-clock-out branch in a state where both flags can legitimately be true at once.

Found **13** new issue(s): 🔴 3 important, 🟡 4 suggestion(s), 🟣 6 nit(s)

frontend/src/community/attendance/components/molecules/AttendanceModals/PreMidnightClockOutAlertModal/PreMidnightClockOutAlertModal.tsx
```diff
@@ -1,54 +1,35 @@
...
-import styles from "./styles";
-
 interface Props {
   closeModal: () => void;
 }
 
 const PreMidnightClockOutAlertModal: FC<Props> = ({ closeModal }) => {
   const translateText = useTranslator("attendanceModule", "timeWidget");
-  const classes = styles();
+
```
ThinuwanW, line 14: 🤖 **Claude** · 🔴 **Important**

📁 **File-level observation**

Breaks the existing co-located test. `PreMidnightClockOutAlertModal.test.tsx` (line 28) asserts `expect(screen.getByText("clockOutAlert")).toBeInTheDocument();`, but this PR removes the `clockOutAlert` Typography from the component (the title now lives in SmallModal's `modalHeader`, rendered by TimeWidgetPopupController). The test file was not touched in this PR, so `getByText` will throw and the suite will fail in CI. Fix: remove that assertion from the component test (the component no longer owns the title) and, if title coverage is still wanted, assert it in a TimeWidgetPopupController-level test instead.

frontend/src/community/attendance/components/organisms/TimeWidgetPopupController/TimeWidgetPopupController.tsx
```diff
@@ -111,6 +111,9 @@ const TimeWidgetPopupController = (): JSX.Element => {
     if (isAttendanceModalOpen && popupType === AttendancePopupTypes.CLOCK_IN) {
       return translateText(["clockInConfirmation"]);
     }
+    if (isPreMidnightClockOutAlertOpen) {
```
ThinuwanW, line 114: 🤖 **Claude** · 🔴 **Important**

The new `isPreMidnightClockOutAlertOpen` branch is placed *before* the `isAutoClockOutMidnightModalOpen` branch, but both flags can be true simultaneously. The pre-midnight alert is opened at 23:50 (line 76) and nothing clears `isPreMidnightClockOutAlertOpen` when the auto clock-out fires at 23:59:58 (line 100) — verified: the only writers of that flag are `setIsPreMidnightClockOutAlertOpen` in attendanceSlice and `handleClosePreMidnightModal`. If the user does not dismiss the alert within those ~8 minutes (precisely the away-from-desk scenario the alert exists for), `modalContent()` renders BOTH PreMidnightClockOutAlertModal and AutoClockOutMidnightModal stacked in one SmallModal, and the header now reads "Clock out alert !" instead of "Clocked out" — a behavioural regression versus before this PR, where the auto-clock-out title won. Fix: call `setIsPreMidnightClockOutAlertOpen(false)` inside the 23:59:58 timeout before `setIsAutoClockOutMidnightModalOpen(true)` (which also fixes the stacked-content bug), and/or move the `isAutoClockOutMidnightModalOpen` check above the new pre-midnight check so the more recent modal wins.

frontend/src/community/attendance/components/molecules/AttendanceModals/PreMidnightClockOutAlertModal/PreMidnightClockOutAlertModal.tsx
```diff
@@ -1,54 +1,35 @@
...
-              iconPosition="end"
-            >
-              {translateText(["ok"])}
-            </ButtonV2>
-          </Stack>
-        </Box>
-      </Box>
-    </>
+    <div>
+      <p className="body2 pb-4">{translateText(["clockOutAlertMessage"])}</p>
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Styling lost in the migration. The deleted `styles.ts` `bodyText` applied `color: "#71717A"` and `textAlign: "center"`; the replacement `className="body2"` only carries font-size 14px / weight 400 (see styles/global.css `.body2`). The message now renders in the inherited (primary/black) colour and left-aligned. If the greyed, centred treatment was intended to be preserved, use `className="body2 text-secondary-text text-center"` — `--color-secondary-text: #52525b` is the closest existing theme token (global.css:59) and is the token other migrated modals use (e.g. GeofenceMap, AllFeaturesTable). If the colour/alignment change is a deliberate design update, ignore this.

frontend/src/community/attendance/components/molecules/AttendanceModals/PreMidnightClockOutAlertModal/PreMidnightClockOutAlertModal.tsx
```diff
@@ -1,54 +1,35 @@
...
+      <div className="flex justify-end gap-2 mt-4">
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Button layout changed and is now inconsistent with its sibling modal. Previously the ButtonV2 sat in `<Stack spacing={2}>` (column flex, default `alignItems: stretch`), so it rendered full-width. `flex justify-end` makes it right-aligned and auto-width. AutoClockOutMidnightModal — which renders inside the *same* SmallModal from the same controller — still uses the full-width stacked buttons, so the two midnight modals will now look different. Either keep it full-width (`<div className="mt-4"><ButtonV2 .../></div>`) or migrate AutoClockOutMidnightModal in the same PR. Note also that SmallModal exposes a `buttons` prop (see EnableLeavePoliciesConfirmModal) which is the established way to render modal actions; it isn't usable here because TimeWidgetPopupController owns the SmallModal and multiplexes four different contents, so hand-rolling the footer inside `content` is acceptable — worth a comment so the next reader doesn't 'fix' it.

frontend/src/community/attendance/components/molecules/AttendanceModals/PreMidnightClockOutAlertModal/PreMidnightClockOutAlertModal.tsx
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Double spacing between the message and the button: `pb-4` (1rem) on the `<p>` plus `mt-4` (1rem) on the button row collapses to 2rem of gap, whereas the previous `bodyText.pb: "1rem"` produced 1rem. Drop one of them — e.g. `className="body2"` on the `<p>` and keep `mt-4`, or use a single `flex flex-col gap-4` wrapper as EnableLeavePoliciesConfirmModal does.

frontend/src/community/attendance/components/molecules/AttendanceModals/PreMidnightClockOutAlertModal/PreMidnightClockOutAlertModal.tsx
```diff
@@ -1,54 +1,35 @@
...
-              icon={<Icon name={IconName.CHECK_ICON} />}
-              iconPosition="end"
-            >
-              {translateText(["ok"])}
-            </ButtonV2>
-          </Stack>
-        </Box>
-      </Box>
-    </>
+    <div>
```
ThinuwanW, line 20: 🤖 **Claude** · 🟣 **Nit**

The deleted `styles.ts` `bodyContainer` had `maxHeight: "50vh"; overflow: "auto"`, which is not reproduced in the new markup. The English string is short so this is unlikely to bite today, but longer translations (this key is user-facing and localised) on a short viewport could now overflow the modal instead of scrolling. Consider `className="max-h-[50vh] overflow-auto"` on the message wrapper if the scroll guard was intentional.

ThinuwanW, line 20: 🤖 **Claude** · 🟣 **Nit**

The outer `<div>` carries no classes and no semantics — it only exists to satisfy the single-root requirement, which the `<p>` + button row already need. Either give it the layout classes (e.g. `flex flex-col gap-4`, which would also resolve the double-spacing note above) or replace it with a `<>...</>` fragment.

frontend/src/community/attendance/components/organisms/TimeWidgetPopupController/TimeWidgetPopupController.tsx
```diff
@@ -111,6 +111,9 @@ const TimeWidgetPopupController = (): JSX.Element => {
...
+      return translateText(["clockOutAlert"]);
+    }
     if (isAutoClockOutMidnightModalOpen) {
```
ThinuwanW, line 117: 🤖 **Claude** · 🔴 **Important**

Beyond which *title* wins, the underlying state problem is that `isPreMidnightClockOutAlertOpen` and `isAutoClockOutMidnightModalOpen` are never made mutually exclusive, and this PR makes the consequences visible. The 23:59:58 effect calls `setIsAutoClockOutMidnightModalOpen(true)` without clearing the pre-midnight flag, so from 23:59:58 both flags are true and `modalContent()` renders BOTH `<PreMidnightClockOutAlertModal>` and `<AutoClockOutMidnightModal>` stacked inside the single `SmallModal` — the user sees "You will be clocked out in 10 minutes..." directly above "You have been clocked out...", plus three action buttons. Reordering the title branches does not fix this. Worse, `handleCloseModal` uses an if/else-if chain that clears only the first matching flag, so clicking Okay or the close (X) icon dismisses the alert but leaves `isAutoClockOutMidnightModalOpen` true and the `SmallModal` still open with residual content. Fix: in the second `useEffect` (the 23:59:58 timer), call `setIsPreMidnightClockOutAlertOpen(false)` immediately before `setIsAutoClockOutMidnightModalOpen(true)`; and change `handleCloseModal` to reset all three flags unconditionally rather than else-if, so the close handler can never leave the modal half-open. Same applies to the overlap with `isAttendanceModalOpen` — if the user has the clock-out confirmation open at 23:50 the alert body appears underneath it with no heading at all, since the title branches for `isAttendanceModalOpen` win.

frontend/src/community/attendance/components/molecules/AttendanceModals/PreMidnightClockOutAlertModal/PreMidnightClockOutAlertModal.tsx
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The migration hand-rolls the modal footer inside `content` instead of using `SmallModal`'s `buttons` prop, which is the established convention for every other migrated modal in the codebase (see `SetHireDateModal.tsx:88`, `FolderNotEmptyModal.tsx:35`, `EnableLeavePoliciesConfirmModal.tsx:59`, `PolicyLeaveTypeActivationToggleButton.tsx:132` — all pass `buttons={{ buttonRight: { variant, onClick, icon, iconPosition, children } }}`). The consequence is that the exact class string `flex justify-end gap-2 mt-4` is now duplicated verbatim in two files (this line and `ClockOutModal.tsx:104`), and the footer will not pick up spacing/divider/responsive changes made in the design system. Because `TimeWidgetPopupController` owns a single shared `SmallModal`, moving to `buttons` requires the controller to build the button config per active modal (e.g. a `getModalButtons()` sibling to `getModalTitle()`); if that refactor is out of scope for this PR, at minimum extract the repeated footer row into a shared `ModalFooter` molecule so the two copies cannot drift.

frontend/src/community/attendance/components/molecules/AttendanceModals/PreMidnightClockOutAlertModal/PreMidnightClockOutAlertModal.tsx
```diff
@@ -1,54 +1,35 @@
...
+      <p className="body2 pb-4">{translateText(["clockOutAlertMessage"])}</p>
+      <div className="flex justify-end gap-2 mt-4">
+        <ButtonV2
+          onClick={handleOkay}
+          aria-label={translateText(["ok"])}
+          icon={<Icon name={IconName.CHECK_ICON} />}
```
ThinuwanW, line 26: 🤖 **Claude** · 🟡 **Suggestion**

Incomplete migration: the component now has zero MUI dependencies, but still routes its icon through the legacy `~community/common/components/atoms/Icon/Icon` + `IconName.CHECK_ICON` indirection. `Icon.tsx` is a 375-line barrel that eagerly `import`s ~200 individual icon components at module scope, so pulling it into an otherwise design-system-only component drags the entire icon set into this chunk — the exact pattern the guidelines call out under bundle size. Peer modals already migrated to `SmallModal` import the icon directly from the design system (`import { CheckIcon } from "@rootcodelabs/skapp-ui"`, see `PolicyLeaveTypeActivationToggleButton.tsx:2,146`). Replace `icon={<Icon name={IconName.CHECK_ICON} />}` with `icon={<CheckIcon />}` and drop the `Icon` / `IconName` imports.

frontend/src/community/attendance/components/molecules/AttendanceModals/PreMidnightClockOutAlertModal/PreMidnightClockOutAlertModal.tsx
```diff
@@ -1,54 +1,35 @@
...
-          </Stack>
-        </Box>
-      </Box>
-    </>
+    <div>
+      <p className="body2 pb-4">{translateText(["clockOutAlertMessage"])}</p>
+      <div className="flex justify-end gap-2 mt-4">
+        <ButtonV2
+          onClick={handleOkay}
+          aria-label={translateText(["ok"])}
```
ThinuwanW, line 25: 🤖 **Claude** · 🟣 **Nit**

Accessibility: `aria-label={translateText(["ok"])}` duplicates the button's own visible text content (`{translateText(["ok"])}` as children), so the accessible name is computed twice from the same string with no benefit — and the aria-label silently overrides the visible label, which is a WCAG 2.5.3 (Label in Name) hazard if the two ever diverge. The icon is purely decorative and should be hidden from the accessibility tree. Suggest dropping `aria-label` entirely and adding `aria-hidden="true"` to the icon element, matching `EnableLeavePoliciesConfirmModal.tsx:38`.

frontend/src/community/attendance/components/molecules/AttendanceModals/PreMidnightClockOutAlertModal/PreMidnightClockOutAlertModal.tsx
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

This modal appears unprompted on a timer (the 23:50 `setTimeout` in `TimeWidgetPopupController`) rather than in response to a user action, so screen-reader users get no announcement unless `SmallModal` moves focus into the dialog. Since the component is now plain markup rather than MUI (which handled some of this via the dialog role), consider adding `role="alert"` to the message paragraph so the warning is announced when it appears — the codebase already uses this pattern for time-sensitive text (`LeavePolicyWizard/BasicInfoStep.tsx:83` uses `<p role="alert" className="body2 ...">`). Flagging as a nit because it depends on what focus management `SmallModal` already provides.

frontend/src/community/attendance/components/molecules/AttendanceModals/PreMidnightClockOutAlertModal/PreMidnightClockOutAlertModal.tsx
```diff
@@ -1,54 +1,35 @@
...
-
 interface Props {
   closeModal: () => void;
 }
 
 const PreMidnightClockOutAlertModal: FC<Props> = ({ closeModal }) => {
   const translateText = useTranslator("attendanceModule", "timeWidget");
-  const classes = styles();
+
   const handleOkay = (): void => {
```
ThinuwanW, line 15: 🤖 **Claude** · 🟣 **Nit**

`handleOkay` is a zero-value wrapper — it takes no arguments and does nothing but call `closeModal()`. A refactor PR whose stated purpose is simplifying this component is the right moment to remove it and pass `onClick={closeModal}` directly, eliminating a new function identity on every render for no behavioural difference. (Peer modals such as `ClockOutModal.handleUndoClockOut` have the same redundancy, so this is consistent with existing code — hence a nit rather than a must-fix.)

frontend/src/community/attendance/components/molecules/AttendanceModals/PreMidnightClockOutAlertModal/PreMidnightClockOutAlertModal.tsx
Deshan1100: No need test

frontend/src/community/attendance/components/molecules/AttendanceModals/PreMidnightClockOutAlertModal/PreMidnightClockOutAlertModal.tsx
Deshan1100: Kept as previous

frontend/src/community/attendance/components/molecules/AttendanceModals/PreMidnightClockOutAlertModal/PreMidnightClockOutAlertModal.tsx
Deshan1100: No need this styles checked

Deshan1100: div updated align with new small modal

frontend/src/community/attendance/components/molecules/AttendanceModals/PreMidnightClockOutAlertModal/PreMidnightClockOutAlertModal.tsx
Deshan1100: No need to change it align with new design

Deshan1100: No need checkd

Deshan1100: kept as previous

frontend/src/community/attendance/components/organisms/TimeWidgetPopupController/TimeWidgetPopupController.tsx
Deshan1100: checked isPreMidnightClockOutAlertOpen modal first

frontend/src/community/attendance/components/organisms/TimeWidgetPopupController/TimeWidgetPopupController.tsx
Deshan1100: need this title checked

frontend/src/community/attendance/components/molecules/AttendanceModals/PreMidnightClockOutAlertModal/PreMidnightClockOutAlertModal.tsx
Deshan1100: changed as per the design need this icon

frontend/src/community/attendance/components/molecules/AttendanceModals/PreMidnightClockOutAlertModal/PreMidnightClockOutAlertModal.tsx
Deshan1100: kept as previous

frontend/src/community/attendance/components/molecules/AttendanceModals/PreMidnightClockOutAlertModal/PreMidnightClockOutAlertModal.tsx
Deshan1100: checked need this UI changes as per the new small modal

Deshan1100: need this styls for new small modal

shakila-kamalasena (approved):

frontend/src/community/attendance/components/molecules/AttendanceModals/PreMidnightClockOutAlertModal/PreMidnightClockOutAlertModal.tsx
shakila-kamalasena, line 26: Use skapp-ui's CheckTaskIcon

frontend/src/community/attendance/components/molecules/AttendanceModals/PreMidnightClockOutAlertModal/PreMidnightClockOutAlertModal.tsx
shakila-kamalasena: Do as this since its clean

shakila-kamalasena (changes requested):

frontend/src/community/attendance/components/molecules/AttendanceModals/PreMidnightClockOutAlertModal/PreMidnightClockOutAlertModal.tsx
Deshan1100: icon size not match as default not size prop exposed

frontend/src/community/attendance/components/molecules/AttendanceModals/PreMidnightClockOutAlertModal/PreMidnightClockOutAlertModal.tsx
Deshan1100: updated

shakila-kamalasena (approved):

akilarootcode (approved):

PR #1718, opened 2026-08-07
Fix crm 559 route restriction for crm none role

YasinduRC:
### Summary

<html>
<body>
<html><head></head><body><p>Here's a ready-to-paste PR description covering the full branch:</p>
<hr>
<h2>Summary</h2>
<p>Fixes <strong>CRM-559</strong>: when an admin revokes a user's module access (e.g. sets CRM to <strong>None</strong>), the user could keep viewing that module's pages (<code>/crm</code>) until their access token expired (~1 hour) or they manually signed out. This PR makes revocation take effect promptly by repairing the token-refresh path and adding a client-side module access guard.</p>
<h2>Root cause</h2>
<p>Access revocation relies on a 4-step chain. Step 3 was broken, which silently disabled step 4:</p>
<ol>
<li><strong>Backend</strong> invalidates the stale token → next API call returns <code>USER_VERSION_MISMATCH</code>. ✅</li>
<li><strong>Frontend interceptor</strong> catches the mismatch and requests a token refresh. ✅</li>
<li><strong>Token refresh had no endpoint URL.</strong> ❌ — The refresh call reads <code>authenticationEndpoints.REFRESH_TOKEN</code>, imported from <code>~enterprise/...</code>. In community builds the enterprise submodule isn't present, so the import falls back to <code>src/fallback/common/api/utils/ApiEndpoints.ts</code>, which was literally <code>export const authenticationEndpoints = {}</code>. <code>REFRESH_TOKEN</code> was <code>undefined</code>, so <strong>every refresh in community builds failed</strong>.</li>
<li><strong>The failure was swallowed.</strong> <code>getNewAccessToken()</code> returned <code>null</code>, but the caller ignored it and called <code>checkAuth()</code>, which re-read the <strong>same stale cookie</strong> and restored the <strong>old roles</strong>. Since roles never changed, the client had no signal to re-evaluate route access, so the user stayed on the module.</li>
</ol>
<p>Additionally, route access was only enforced by <code>middleware.ts</code>, which runs on document requests. Client-side navigations never reach the edge, so nothing re-checked access while the user stayed inside the app.</p>
<h2>Fix</h2>
<p><strong>1. Restore the community auth endpoints</strong> (<code>fallback/.../ApiEndpoints.ts</code>)
Populated the fallback with the session (cookie) based endpoints matching <code>CREDENTIAL_SIGN_IN</code>:</p>
<ul>
<li><code>REFRESH_TOKEN: /v1/auth/session/refresh-token</code></li>
<li><code>SIGNOUT: /v1/auth/session/sign-out</code></li>
</ul>
<p><strong>2. Fail safe on a failed refresh</strong> (<code>TanStackProvider.tsx</code>)
If the refresh cannot mint a new token (<code>getNewAccessToken()</code> returns <code>null</code>), the session is ended instead of silently retaining stale permissions:</p>
<pre><code class="language-ts">const refreshedAccessToken = await getNewAccessToken();
if (!refreshedAccessToken) { await signOut(); return; }
</code></pre>
<p><strong>3. Client-side module access guard</strong> (<code>useModuleAccessGuard</code>, <code>moduleGuardedRoutes</code>, <code>isModuleRouteRestricted</code>)
A new hook, mounted in <code>BaseLayout</code>, re-checks module access whenever <code>user.roles</code> changes. If the user is on a route they no longer have the required role for, it redirects to the dashboard. The route→role table mirrors the gates <code>getDrawerRoutes</code> already applies (CRM, Invoice, Projects, Leave, Timesheet), so the guard is never stricter than the sidebar.</p>
<h3>Resulting behaviour</h3>

Refresh outcome | Result
-- | --
Succeeds (properly configured origin) | Roles update → guard redirects to dashboard
Fails (e.g. refresh token gone/expired) | User is signed out

<p>Either way, the user stops viewing a module they no longer have access to.</p>
<h2>Testing</h2>
<ul>
<li>Added <code>isModuleRouteRestricted.test.ts</code> (9 cases): CRM downgrade, higher CRM roles inheriting the representative role, super-admin without a CRM role, and the other four module gates. All passing.</li>
<li>Verified live against the running backend: after downgrading a user's CRM role, <code>POST /v1/auth/session/refresh-token</code> returns <code>200</code> with a token whose <code>roles</code> no longer include CRM.</li>
<li>Verified the role model is cumulative (an admin/manager token also carries the base role), so the guard cannot wrongly redirect a legitimate admin/manager.</li>
</ul>
<h2>Risk / breaking changes</h2>
<ul>
<li><strong>No change to any working flow.</strong> The refresh success path is unchanged; new behaviour only occurs on the refresh <em>failure</em> path (which was already broken), where the app now fails safe (sign out) instead of failing open (retain revoked access).</li>
<li><strong>Enterprise unaffected by the endpoints change</strong> — <code>~enterprise/*</code> resolves to the enterprise sources first; the fallback is used only in community builds. The guard and sign-out handler run in both community and enterprise, and mirror existing drawer gates.</li>
</ul>
<h2>Ticket</h2>
<p>CRM-559</p>
<hr>
<p>Want me to tailor the tone (shorter/more concise), add a screenshots/GIF section, or save this to a file?</p></body></html>
</body>
</html>

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/frontend/src/community/common/components/templates/BaseLayout/BaseLayout.tsx
+++ b/frontend/src/community/common/components/templates/BaseLayout/BaseLayout.tsx
@@ -8,6 +8,7 @@ import CommonModalController from "~community/common/components/organisms/Common
 import ContentWithDrawer from "~community/common/components/organisms/ContentWithDrawer/ContentWithDrawer";
 import ContentWithoutDrawer from "~community/common/components/organisms/ContentWithoutDrawer/ContentWithoutDrawer";
 import { appModes } from "~community/common/constants/configs";
+import useModuleAccessGuard from "~community/common/hooks/useModuleAccessGuard";
 import useSessionData from "~community/common/hooks/useSessionData";
 import { tenantID } from "~community/common/utils/axiosInterceptor";
 import { setDeviceToken } from "~enterprise/common/api/setDeviceTokenApi";
@@ -25,6 +26,8 @@ const BaseLayout = ({ children }: Props) => {
   const { asPath } = useRouter();
 
   const { sessionStatus } = useSessionData();
+
+  useModuleAccessGuard();
 
   const { token } = useFcmToken();
 
--- a/frontend/src/community/common/constants/routes.ts
+++ b/frontend/src/community/common/constants/routes.ts
@@ -193,14 +193,6 @@ export const managerRestrictedRoutes = [ROUTES.PEOPLE.ADD];
 
 export const userRolesRestrictedRoutes = [ROUTES.CONFIGURATIONS.USER_ROLES];
 
-/**
- * The base role each module's routes require, mirroring the module gates
- * `getDrawerRoutes` already applies when it hides a module from the drawer.
- *
- * `middleware.ts` cannot enforce these while the user stays inside the app, since
- * client side navigations never reach the edge. `useModuleAccessGuard` applies them
- * whenever the roles change.
- */
 export const moduleGuardedRoutes: {
   routes: string[];
   requiredRole: EmployeeTypes | ManagerTypes | RepresentativeTypes;
--- a/frontend/src/community/common/constants/routes.ts
+++ b/frontend/src/community/common/constants/routes.ts
@@ -1,3 +1,9 @@
+import {
+  EmployeeTypes,
+  ManagerTypes,
+  RepresentativeTypes
+} from "~community/common/types/AuthTypes";
+
 const ROUTES = {
   AUTH: {
     SIGNUP: "/signup",
@@ -186,3 +192,41 @@ export const employeeRestrictedRoutes = [
 export const managerRestrictedRoutes = [ROUTES.PEOPLE.ADD];
 
 export const userRolesRestrictedRoutes = [ROUTES.CONFIGURATIONS.USER_ROLES];
+
+/**
+ * The base role each module's routes require, mirroring the module gates
+ * `getDrawerRoutes` already applies when it hides a module from the drawer.
+ *
+ * `middleware.ts` cannot enforce these while the user stays inside the app, since
+ * client side navigations never reach the edge. `useModuleAccessGuard` applies them
+ * whenever the roles change.
+ */
+export const moduleGuardedRoutes: {
+  routes: string[];
+  requiredRole: EmployeeTypes | ManagerTypes | RepresentativeTypes;
+}[] = [
+  {
+    routes: [ROUTES.CRM.BASE],
+    requiredRole: RepresentativeTypes.CRM_SALES_REPRESENTATIVE
+  },
+  {
+    routes: [ROUTES.INVOICE.BASE],
+    requiredRole: ManagerTypes.INVOICE_MANAGER
+  },
+  {
+    routes: [
+      ROUTES.PROJECTS.BASE,
+      ROUTES.PROJECTS.GUESTS,
+      ROUTES.PROJECTS.GUEST_REQUESTS
+    ],
+    requiredRole: EmployeeTypes.PM_EMPLOYEE
+  },
+  {
+    routes: [ROUTES.LEAVE.BASE],
+    requiredRole: EmployeeTypes.LEAVE_EMPLOYEE
+  },
+  {
+    routes: [ROUTES.TIMESHEET.BASE],
+    requiredRole: EmployeeTypes.ATTENDANCE_EMPLOYEE
+  }
+];
--- a/frontend/src/community/common/hooks/useModuleAccessGuard.ts
+++ b/frontend/src/community/common/hooks/useModuleAccessGuard.ts
@@ -1,0 +1,39 @@
+import { useRouter } from "next/router";
+import { useEffect } from "react";
+
+import { useAuth } from "~community/auth/providers/AuthProvider";
+import ROUTES from "~community/common/constants/routes";
+import { isModuleRouteRestricted } from "~community/common/utils/commonUtil";
+
+/**
+ * Re-checks module access on the client whenever the user's roles change.
+ *
+ * `middleware.ts` is the only place route access is enforced, and it only runs on
+ * document requests. Client side navigations never reach the edge, so a user who is
+ * already inside a module keeps viewing it after an admin removes their role: the
+ * drawer recomputes from `user.roles` and hides the module, the page does not.
+ *
+ * A hard refresh does not recover either, because the middleware authorises from the
+ * roles inside the access token cookie, and that cookie is only replaced later, once
+ * an API call returns a user version mismatch.
+ */
+const useModuleAccessGuard = (): void => {
+  const router = useRouter();
+
+  const { isAuthenticated, user } = useAuth();
+
+  useEffect(() => {
+    const roles = user?.roles;
+
+    if (!isAuthenticated || !roles) return;
+
+    // `asPath` holds the public url, `pathname` holds the rewritten page path
+    const currentPath = router.asPath.split(/[?#]/)[0];
+
+    if (isModuleRouteRestricted(currentPath, roles)) {
+      void router.replace(ROUTES.DASHBOARD.BASE);
+    }
+  }, [isAuthenticated, user?.roles, router]);
+};
+
+export default useModuleAccessGuard;
--- a/frontend/src/community/common/hooks/useModuleAccessGuard.ts
+++ b/frontend/src/community/common/hooks/useModuleAccessGuard.ts
@@ -5,18 +5,6 @@ import { useAuth } from "~community/auth/providers/AuthProvider";
 import ROUTES from "~community/common/constants/routes";
 import { isModuleRouteRestricted } from "~community/common/utils/commonUtil";
 
-/**
- * Re-checks module access on the client whenever the user's roles change.
- *
- * `middleware.ts` is the only place route access is enforced, and it only runs on
- * document requests. Client side navigations never reach the edge, so a user who is
- * already inside a module keeps viewing it after an admin removes their role: the
- * drawer recomputes from `user.roles` and hides the module, the page does not.
- *
- * A hard refresh does not recover either, because the middleware authorises from the
- * roles inside the access token cookie, and that cookie is only replaced later, once
- * an API call returns a user version mismatch.
- */
 const useModuleAccessGuard = (): void => {
   const router = useRouter();
 
@@ -27,7 +15,6 @@ const useModuleAccessGuard = (): void => {
 
     if (!isAuthenticated || !roles) return;
 
-    // `asPath` holds the public url, `pathname` holds the rewritten page path
     const currentPath = router.asPath.split(/[?#]/)[0];
 
     if (isModuleRouteRestricted(currentPath, roles)) {
--- a/frontend/src/community/common/providers/TanStackProvider.tsx
+++ b/frontend/src/community/common/providers/TanStackProvider.tsx
@@ -34,7 +34,13 @@ const TanStackProvider = ({ children }: { children: ReactNode }) => {
 
   useEffect(() => {
     const handleTokenRefresh = async () => {
-      await getNewAccessToken();
+      const refreshedAccessToken = await getNewAccessToken();
+
+      if (!refreshedAccessToken) {
+        await signOut();
+        return;
+      }
+
       await checkAuth();
       queryClient.invalidateQueries();
     };
--- a/frontend/src/community/common/utils/__test__/isModuleRouteRestricted.test.ts
+++ b/frontend/src/community/common/utils/__test__/isModuleRouteRestricted.test.ts
@@ -1,0 +1,160 @@
+/**
+ * `commonUtil` pulls in `next/server`, which needs the node environment
+ * @jest-environment node
+ */
+import ROUTES from "~community/common/constants/routes";
+import {
+  AdminTypes,
+  EmployeeTypes,
+  ManagerTypes,
+  ROLE_SUPER_ADMIN,
+  RepresentativeTypes
+} from "~community/common/types/AuthTypes";
+import { isModuleRouteRestricted } from "~community/common/utils/commonUtil";
+
+// Roles an employee keeps regardless of their module roles
+const baseRoles: string[] = [
+  EmployeeTypes.PEOPLE_EMPLOYEE,
+  EmployeeTypes.LEAVE_EMPLOYEE,
+  EmployeeTypes.ATTENDANCE_EMPLOYEE
+];
+
+describe("isModuleRouteRestricted", () => {
+  describe("CRM downgraded to none", () => {
+    it("restricts every CRM route", () => {
+      expect(isModuleRouteRestricted(ROUTES.CRM.BASE, baseRoles)).toBe(true);
+      expect(isModuleRouteRestricted(ROUTES.CRM.CONTACTS, baseRoles)).toBe(
+        true
+      );
+      expect(isModuleRouteRestricted(ROUTES.CRM.COMPANIES, baseRoles)).toBe(
+        true
+      );
+      expect(isModuleRouteRestricted(ROUTES.CRM.DEALS, baseRoles)).toBe(true);
+      expect(isModuleRouteRestricted(ROUTES.CRM.TASKS, baseRoles)).toBe(true);
+      expect(isModuleRouteRestricted("/crm/deals/123", baseRoles)).toBe(true);
+    });
+
+    it("restricts CRM for a super admin without a CRM role, matching the middleware", () => {
+      expect(
+        isModuleRouteRestricted(ROUTES.CRM.CONTACTS, [
+          ...baseRoles,
+          ROLE_SUPER_ADMIN
+        ])
+      ).toBe(true);
+    });
+
+    it("leaves the user's remaining modules alone", () => {
+      expect(isModuleRouteRestricted(ROUTES.DASHBOARD.BASE, baseRoles)).toBe(
+        false
+      );
+      expect(isModuleRouteRestricted(ROUTES.LEAVE.MY_REQUESTS, baseRoles)).toBe(
+        false
+      );
+      expect(
+        isModuleRouteRestricted(ROUTES.TIMESHEET.MY_TIMESHEET, baseRoles)
+      ).toBe(false);
+      expect(isModuleRouteRestricted(ROUTES.PEOPLE.DIRECTORY, baseRoles)).toBe(
+        false
+      );
+      expect(isModuleRouteRestricted(ROUTES.SETTINGS.BASE, baseRoles)).toBe(
+        false
+      );
+      expect(isModuleRouteRestricted(ROUTES.NOTIFICATIONS, baseRoles)).toBe(
+        false
+      );
+    });
+  });
+
+  describe("CRM granted", () => {
+    const crmRoles = [
+      ...baseRoles,
+      RepresentativeTypes.CRM_SALES_REPRESENTATIVE
+    ];
+
+    it("allows every CRM route", () => {
+      expect(isModuleRouteRestricted(ROUTES.CRM.BASE, crmRoles)).toBe(false);
+      expect(isModuleRouteRestricted(ROUTES.CRM.DEALS, crmRoles)).toBe(false);
+      expect(isModuleRouteRestricted("/crm/deals/123", crmRoles)).toBe(false);
+    });
+
+    it("allows CRM for the higher CRM roles, which inherit the representative role", () => {
+      expect(
+        isModuleRouteRestricted(ROUTES.CRM.DEALS, [
+          ...crmRoles,
+          ManagerTypes.CRM_SALES_MANAGER
+        ])
+      ).toBe(false);
+      expect(
+        isModuleRouteRestricted(ROUTES.CRM.DEALS, [
+          ...crmRoles,
+          AdminTypes.CRM_ADMIN
+        ])
+      ).toBe(false);
+    });
+  });
+
+  describe("the other module gates", () => {
+    it("restricts invoice without the invoice manager role", () => {
+      expect(isModuleRouteRestricted(ROUTES.INVOICE.BASE, baseRoles)).toBe(
+        true
+      );
+      expect(
+        isModuleRouteRestricted(ROUTES.INVOICE.ALL_INVOICES, [
+          ...baseRoles,
+          ManagerTypes.INVOICE_MANAGER
+        ])
+      ).toBe(false);
+    });
+
+    it("restricts projects without the pm employee role", () => {
+      expect(isModuleRouteRestricted(ROUTES.PROJECTS.BASE, baseRoles)).toBe(
+        true
+      );
+      expect(isModuleRouteRestricted(ROUTES.PROJECTS.GUESTS, baseRoles)).toBe(
+        true
+      );
+      expect(
+        isModuleRouteRestricted(ROUTES.PROJECTS.BASE, [
+          ...baseRoles,
+          EmployeeTypes.PM_EMPLOYEE
+        ])
+      ).toBe(false);
+    });
+
+    it("restricts leave and timesheet only when the base role is gone", () => {
+      const noLeaveOrAttendance = [EmployeeTypes.PEOPLE_EMPLOYEE];
+
+      expect(
+        isModuleRouteRestricted(ROUTES.LEAVE.MY_REQUESTS, noLeaveOrAttendance)
+      ).toBe(true);
+      expect(
+        isModuleRouteRestricted(
+          ROUTES.TIMESHEET.MY_TIMESHEET,
+          noLeaveOrAttendance
+        )
+      ).toBe(true);
+      expect(isModuleRouteRestricted(ROUTES.LEAVE.MY_REQUESTS, baseRoles)).toBe(
+        false
+      );
+    });
+  });
+
+  describe("routes outside the module gates", () => {
+    it("never restricts them, whatever the roles", () => {
+      [
+        ROUTES.AUTH.SIGNIN,
+        ROUTES.AUTH.UNAUTHORIZED,
+        ROUTES.AUTH.RESET_PASSWORD,
+        ROUTES.DASHBOARD.BASE,
+        ROUTES.CONFIGURATIONS.BASE,
+        ROUTES.PEOPLE.DIRECTORY,
+        ROUTES.SIGN.INBOX,
+        ROUTES.NOTIFICATIONS,
+        ROUTES.SETTINGS.BASE
+      ].forEach((route) => {
+        expect(isModuleRouteRestricted(route, [])).toBe(false);
+        expect(isModuleRouteRestricted(route, baseRoles)).toBe(false);
+      });
+    });
+  });
+});
--- a/frontend/src/community/common/utils/commonUtil.ts
+++ b/frontend/src/community/common/utils/commonUtil.ts
@@ -22,7 +22,7 @@ import { JobFamilies } from "~community/people/types/JobRolesTypes";
 import { getShortDayName } from "~community/people/utils/holidayUtils/commonUtils";
 
 import { appModes } from "../constants/configs";
-import ROUTES from "../constants/routes";
+import ROUTES, { moduleGuardedRoutes } from "../constants/routes";
 
 export const getBlinkClass = (shouldBlink: boolean): string =>
   shouldBlink ? "animate-pulse" : "";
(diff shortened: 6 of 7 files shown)


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds a client-side module access guard (`useModuleAccessGuard` + `isModuleRouteRestricted` + `moduleGuardedRoutes`) wired into `BaseLayout`, forces sign-out when a token refresh fails, and fills in the previously-empty community fallback `authenticationEndpoints`. The fallback endpoints fix is a genuine bug fix — in community builds `authenticationEndpoints.REFRESH_TOKEN`/`SIGNOUT` resolved to `undefined`, so refresh and sign-out were POSTing to the bare API base URL. The guard itself is reasonable and the role mapping matches what the backend `AuthorityServiceImpl` role hierarchy produces, but it fails open when the `roles` claim is absent, only redirects after the restricted page has already mounted and fired its queries, diverges from the middleware's redirect target, omits the eSign module, and introduces a fourth copy of the module→role mapping. One test is vacuous and does not verify what its name claims.

Found **23** new issue(s): 🔴 5 important, 🟡 13 suggestion(s), 🟣 5 nit(s)

frontend/src/community/common/hooks/useModuleAccessGuard.ts
```diff
@@ -0,0 +1,26 @@
...
+
+const useModuleAccessGuard = (): void => {
+  const router = useRouter();
+
+  const { isAuthenticated, user } = useAuth();
+
+  useEffect(() => {
+    const roles = user?.roles;
+
+    if (!isAuthenticated || !roles) return;
```
ThinuwanW, line 16: 🤖 **Claude** · 🔴 **Important**

Fail-open on a missing `roles` claim. `if (!isAuthenticated || !roles) return;` skips the check entirely when `user.roles` is `undefined` — `extractUserFromToken` sets `roles: claims?.roles`, so any token without a `roles` claim (or a malformed one) leaves the user sitting on a restricted module page with no guard at all. An access guard should deny, not allow, on missing data. Fix: `const roles = user?.roles ?? [];` and then only early-return on `!isAuthenticated`. With `[]`, `isModuleRouteRestricted` correctly treats every guarded module as restricted.

frontend/src/community/common/hooks/useModuleAccessGuard.ts
```diff
@@ -0,0 +1,26 @@
...
+
+    const currentPath = router.asPath.split(/[?#]/)[0];
+
+    if (isModuleRouteRestricted(currentPath, roles)) {
+      void router.replace(ROUTES.DASHBOARD.BASE);
+    }
+  }, [isAuthenticated, user?.roles, router]);
```
ThinuwanW, line 23: 🤖 **Claude** · 🟡 **Suggestion**

The effect reads `router.asPath` but the dependency array only contains `router`. This happens to work today because Next's pages-router `RouterContext` value is rebuilt by `makePublicRouterInstance()` on each render, so the object identity changes on navigation — but that is an implementation detail, not a contract. Depend on the value you actually read: `}, [isAuthenticated, user?.roles, router.asPath, router]);` (keep `router` for the `router.replace` call). That makes re-evaluation on client-side navigation explicit rather than incidental.

frontend/src/community/common/hooks/useModuleAccessGuard.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  useEffect(() => {
+    const roles = user?.roles;
+
+    if (!isAuthenticated || !roles) return;
+
+    const currentPath = router.asPath.split(/[?#]/)[0];
+
+    if (isModuleRouteRestricted(currentPath, roles)) {
+      void router.replace(ROUTES.DASHBOARD.BASE);
```
ThinuwanW, line 21: 🤖 **Claude** · 🟡 **Suggestion**

Redirect target diverges from the middleware and may itself be inaccessible. `middleware.ts:305-312` sends a CRM-restricted user to `ROUTES.AUTH.UNAUTHORIZED`, while this guard sends them to `ROUTES.DASHBOARD.BASE` — the same violation produces two different destinations depending on whether it was a hard or soft navigation. Worse, `/dashboard` is not universally reachable: `middleware.ts:269-280` bounces users without `LEAVE_EMPLOYEE`/`PEOPLE_MANAGER`/`ATTENDANCE_MANAGER` to `/timesheet/my-timesheet`, and `allowedRoutes` does not grant `/dashboard` to manager-only role sets (e.g. a user whose only role is `INVOICE_MANAGER`, since the invoice hierarchy in `EpAuthorityServiceImpl` has no employee tier), so those users get a second redirect to `/unauthorized`. Recommend redirecting to `ROUTES.AUTH.UNAUTHORIZED` to match the middleware, or at minimum documenting why `/dashboard` was chosen.

frontend/src/community/common/constants/routes.ts
```diff
@@ -186,3 +192,33 @@ export const employeeRestrictedRoutes = [
 export const managerRestrictedRoutes = [ROUTES.PEOPLE.ADD];
 
 export const userRolesRestrictedRoutes = [ROUTES.CONFIGURATIONS.USER_ROLES];
+
+export const moduleGuardedRoutes: {
```
ThinuwanW, line 196: 🤖 **Claude** · 🟡 **Suggestion**

`moduleGuardedRoutes` omits the eSign module. The backend can disable eSign (`EpAuthorityServiceImpl` filters authorities on `moduleConfig.isEsignModule()`), `useSessionData` already exposes `isEsignatureModuleEnabled` keyed on `EmployeeTypes.ESIGN_EMPLOYEE`, and `getDrawerRoutes` hides the Sign nav on the same role — so a user whose eSign module is turned off while they are sitting on `/sign/inbox` will not be redirected, unlike every other module. Adding it is not a trivial one-liner though: `/sign/document/access`, `/sign/sign` and `/sign/info` are public external-signer routes (middleware explicitly short-circuits them at lines 204-210) and must stay ungated. Either add an eSign entry scoped to the authenticated sign routes (`/sign/inbox`, `/sign/sent`, `/sign/template`, `/sign/folders`, `/sign/contacts`, `/sign/create`) or add a comment stating eSign is deliberately out of scope.

frontend/src/community/common/constants/routes.ts
```diff
@@ -186,3 +192,33 @@ export const employeeRestrictedRoutes = [
...
+  routes: string[];
+  requiredRole: EmployeeTypes | ManagerTypes | RepresentativeTypes;
```
ThinuwanW, line 198: 🤖 **Claude** · 🟡 **Suggestion**

This is now the fourth copy of the module→required-role mapping: `useSessionData` (`isLeaveModuleEnabled`/`isAttendanceModuleEnabled`/`isInvoiceModuleEnabled`/`isPmModuleEnabled`/`isCrmModuleEnabled`), `getDrawerRoutes` (the per-route `hasPMAccess`/`isInvoiceManager`/`hasCRMAccess` blocks), the hardcoded CRM gate in `middleware.ts:305-312`, and now `moduleGuardedRoutes`. They agree today, but a future module role change has to be made in four places or the drawer, the middleware and the guard will disagree. Consider exporting a single `MODULE_ACCESS` map (module → required role → route prefixes) from one place and deriving all four from it — in particular, `middleware.ts` could call `isModuleRouteRestricted` instead of its bespoke CRM check, which would also make the server and client agree on which modules are gated.

frontend/src/community/common/utils/commonUtil.ts
```diff
@@ -521,6 +521,16 @@ export const checkRestrictedRoutesAndRedirect = (
   return null;
 };
 
+export const isModuleRouteRestricted = (
+  pathname: string,
+  roles: string[]
+): boolean =>
+  moduleGuardedRoutes.some(
+    ({ routes, requiredRole }) =>
+      routes.some((url) => pathname.startsWith(url)) &&
```
ThinuwanW, line 530: 🤖 **Claude** · 🟡 **Suggestion**

`pathname.startsWith(url)` has no path-segment boundary, so `/leave` also matches a hypothetical `/leave-analytics-export`, and `/crm` matches `/crmx`. More concretely, the Projects entry uses `ROUTES.PROJECTS.BASE` = `/projects/list` rather than the module root `/projects`, so only the three enumerated leaf pages are covered while `middleware.ts` guards the whole `/projects/:path*` subtree — any new page added under `/projects` silently bypasses this guard. Suggest boundary-aware matching, e.g. `routes.some((url) => pathname === url || pathname.startsWith(`${url}/`))`, and listing `/projects` as the guarded prefix instead of `/projects/list`.

frontend/src/community/common/providers/TanStackProvider.tsx
```diff
@@ -34,7 +34,13 @@ const TanStackProvider = ({ children }: { children: ReactNode }) => {
 
   useEffect(() => {
     const handleTokenRefresh = async () => {
-      await getNewAccessToken();
+      const refreshedAccessToken = await getNewAccessToken();
+
+      if (!refreshedAccessToken) {
```
ThinuwanW, line 39: 🤖 **Claude** · 🟡 **Suggestion**

Two issues with the new sign-out-on-null branch. (1) `getNewAccessToken` swallows every failure in a bare `catch { return null; }` — a 5xx or a network timeout on the refresh endpoint is indistinguishable from an actually-invalid refresh token, so a transient blip during a version-mismatch retry now hard-navigates the user to `/signin` and loses any unsaved form state. Consider only signing out when the refresh genuinely fails auth (surface the status/messageKey from `getNewAccessToken`) and surfacing a toast otherwise. (2) `authInterceptor.ts:42-47` already calls `signOut()` when the refresh response carries `COMMON_ERROR_INVALID_REFRESH_TOKEN`/`COMMON_ERROR_MISSING_COOKIE_IN_TOKEN`; in that path `getNewAccessToken` then returns `null` and this block calls `signOut()` a second time, producing a duplicate `POST /v1/auth/session/sign-out` and a second `window.location.href` assignment. Guard against the double sign-out (e.g. an `isSigningOut` flag in `authUtils`, mirroring the existing `isRefreshing` flag).

frontend/src/community/common/components/templates/BaseLayout/BaseLayout.tsx
```diff
@@ -26,6 +27,8 @@ const BaseLayout = ({ children }: Props) => {
 
   const { sessionStatus } = useSessionData();
 
+  useModuleAccessGuard();
```
ThinuwanW, line 30: 🤖 **Claude** · 🟡 **Suggestion**

The guard only redirects; it does not block rendering. `useModuleAccessGuard()` runs its redirect inside `useEffect`, so on a restricted route the page still mounts, its React Query hooks still fire (the backend will reject them, adding noise and error toasts), and the unauthorized content flashes before `router.replace` lands. Since `BaseLayout` already has a `FullScreenLoader` branch for `sessionStatus === "loading"`, consider having the hook return the restricted boolean and rendering `<FullScreenLoader />` instead of `children` while the redirect is in flight.

frontend/src/community/common/utils/__test__/isModuleRouteRestricted.test.ts
```diff
@@ -0,0 +1,160 @@
...
+      RepresentativeTypes.CRM_SALES_REPRESENTATIVE
+    ];
+
+    it("allows every CRM route", () => {
+      expect(isModuleRouteRestricted(ROUTES.CRM.BASE, crmRoles)).toBe(false);
+      expect(isModuleRouteRestricted(ROUTES.CRM.DEALS, crmRoles)).toBe(false);
+      expect(isModuleRouteRestricted("/crm/deals/123", crmRoles)).toBe(false);
+    });
+
+    it("allows CRM for the higher CRM roles, which inherit the representative role", () => {
```
ThinuwanW, line 80: 🤖 **Claude** · 🟡 **Suggestion**

This test does not verify what its name claims. `crmRoles` already contains `RepresentativeTypes.CRM_SALES_REPRESENTATIVE`, so spreading `CRM_SALES_MANAGER`/`CRM_ADMIN` on top proves nothing — the assertion passes purely because of the representative role and would still pass if `isModuleRouteRestricted` ignored manager/admin roles entirely (which it does: it is a flat `roles.includes(requiredRole)` check with no hierarchy awareness; the flattening happens server-side in `AuthorityServiceImpl.addRoleHierarchy`). Either drop the test, or rename it to make the real contract explicit and assert it: `isModuleRouteRestricted(ROUTES.CRM.DEALS, [...baseRoles, ManagerTypes.CRM_SALES_MANAGER])` returns `true`, documenting that the util depends on the backend emitting the inherited representative role in the JWT.

frontend/src/fallback/common/api/utils/ApiEndpoints.ts
```diff
@@ -1 +1,6 @@
-export const authenticationEndpoints = {};
\ No newline at end of file
+import { ApiVersions } from "~community/common/constants/configs";
+
+export const authenticationEndpoints = {
```
ThinuwanW, line 3: 🤖 **Claude** · 🟡 **Suggestion**

Good catch fixing the `undefined` endpoints — but note these two paths are plain community endpoints (`/v1/auth/session/...`, no `ep/` prefix), so the literals are now duplicated verbatim between `src/enterprise/common/api/utils/ApiEndpoints.ts` and this fallback, and will drift. Cleaner: add `REFRESH_TOKEN`/`SIGNOUT` to `~community/common/api/utils/ApiEndpoints.ts` and have `src/community/auth/utils/authUtils.ts` use `communityAuthEndpoints` for them (it already imports that module at line 1), dropping the `~enterprise/common/api/utils/ApiEndpoints` import entirely. That removes the community→enterprise dependency that caused this bug in the first place.

frontend/src/community/common/utils/__test__/isModuleRouteRestricted.test.ts
```diff
@@ -0,0 +1,160 @@
+/**
```
ThinuwanW, line 1: 🤖 **Claude** · 🟣 **Nit**

The `@jest-environment node` docblock and its justification look unnecessary: the sibling `src/community/common/utils/__test__/commonUtil.test.ts` imports the same `commonUtil` module (and therefore the same `next/server` import) under the default jsdom environment without a pragma. Unless this was added to fix an actual failure, drop it — and consider folding these cases into the existing `commonUtil.test.ts` rather than adding a second test file for the same module.

frontend/src/community/common/hooks/useModuleAccessGuard.ts
```diff
@@ -0,0 +1,26 @@
+import { useRouter } from "next/router";
+import { useEffect } from "react";
+
+import { useAuth } from "~community/auth/providers/AuthProvider";
+import ROUTES from "~community/common/constants/routes";
+import { isModuleRouteRestricted } from "~community/common/utils/commonUtil";
+
+const useModuleAccessGuard = (): void => {
```
ThinuwanW, line 8: 🤖 **Claude** · 🟣 **Nit**

No test for the hook itself. `isModuleRouteRestricted` is well covered, but the hook's own behaviour — early return when unauthenticated, stripping the query/hash from `asPath`, calling `router.replace` exactly once — is untested. `src/community/common/hooks/useBlockBackNavigation/useBlockBackNavigation.test.tsx` is the existing precedent for hook tests with `renderHook`.

frontend/src/community/common/constants/routes.ts
ThinuwanW, line 196: 🤖 **Claude** · 🔴 **Important**

The guard matches only the public (rewritten) paths, so the underlying Next page paths bypass it. `next.config.mjs` rewrites are aliases, not redirects: `pages/enterprise/invoice/allInvoices.tsx` is still directly served at `/enterprise/invoice/allInvoices`, and likewise `/enterprise/projects/list`, `/enterprise/projects/guests`, `/enterprise/projects/guest-requests`. Those paths do not start with `/invoice` or `/projects/list`, so `isModuleRouteRestricted` returns false — and `middleware.ts` config.matcher has no `/enterprise/:path*` entry either, so the edge check never runs on them. Net effect: a user with the invoice or PM module disabled can load the full page by typing the `/enterprise/...` URL. (The `/community/...` twin is incidentally covered because middleware runs on `/community/:path*` and no allowedRoutes prefix matches it, so everyone gets bounced to /unauthorized.) Fix: normalise the path before matching — strip a leading `/enterprise` or `/community` segment inside `isModuleRouteRestricted` — and add `/enterprise/:path*` to the middleware matcher so the server-side gate covers the enterprise pages too.

frontend/src/community/common/hooks/useModuleAccessGuard.ts
```diff
@@ -0,0 +1,26 @@
...
+  const { isAuthenticated, user } = useAuth();
+
+  useEffect(() => {
+    const roles = user?.roles;
+
+    if (!isAuthenticated || !roles) return;
+
+    const currentPath = router.asPath.split(/[?#]/)[0];
+
+    if (isModuleRouteRestricted(currentPath, roles)) {
```
ThinuwanW, line 20: 🤖 **Claude** · 🔴 **Important**

Missing paired change: the authoritative gate in `middleware.ts` was not touched, so the hole this PR targets is still open server-side. `middleware.ts:35` puts `ROUTES.PROJECTS.BASE` into `commonRoutes`, and `commonRoutes` is spread into every entry of `employeeRoutes`, so *every* authenticated role is still granted `/projects/list` at the edge; the only thing stopping them now is a client `useEffect`. That is not an access-control boundary — it does not apply to direct HTML requests before hydration, and it is trivially defeated. Please make the same module→role table drive the middleware (remove `ROUTES.PROJECTS.BASE` from `commonRoutes` and run the `moduleGuardedRoutes` check there, mirroring the existing `checkRestrictedRoutesAndRedirect` calls), and keep this hook only as the client-side UX backstop.

frontend/src/community/common/constants/routes.ts
```diff
@@ -186,3 +192,33 @@ export const employeeRestrictedRoutes = [
...
+}[] = [
+  {
+    routes: [ROUTES.CRM.BASE],
+    requiredRole: RepresentativeTypes.CRM_SALES_REPRESENTATIVE
+  },
+  {
+    routes: [ROUTES.INVOICE.BASE],
```
ThinuwanW, line 205: 🤖 **Claude** · 🔴 **Important**

Behavioural divergence for super-admin-only sessions on invoice and projects. `middleware.ts:39-64` explicitly grants `ROLE_SUPER_ADMIN` access to `ROUTES.INVOICE.BASE`, `ROUTES.INVOICE.ALL_INVOICES`, `ROUTES.INVOICE.CUSTOMERS.BASE`, `ROUTES.PROJECTS.BASE`, `ROUTES.PROJECTS.GUESTS` and `ROUTES.PROJECTS.GUEST_REQUESTS` without requiring any module role. Backend `AuthorityServiceImpl.addSuperAdminAuthority` adds `ROLE_SUPER_ADMIN` independently of module roles, and `isSuperAdminOnlySession()` (commonUtil.ts:48) confirms `roles === [ROLE_SUPER_ADMIN]` is a real state, so such a user has neither `ROLE_INVOICE_MANAGER` nor `ROLE_PM_EMPLOYEE` and is now bounced off pages the middleware just allowed. Note the CRM entry is *not* in the same position — middleware.ts:305-312 already blocks super admins without a CRM role, which is why the test can legitimately assert that. Decide explicitly: either add a `ROLE_SUPER_ADMIN` bypass to `isModuleRouteRestricted`, or delete the invoice/projects entries from `superAdminRoutes` so both layers agree.

frontend/src/fallback/common/api/utils/ApiEndpoints.ts
```diff
@@ -1 +1,6 @@
...
+  REFRESH_TOKEN: `${ApiVersions.V1}/auth/session/refresh-token`,
```
ThinuwanW, line 4: 🤖 **Claude** · 🔴 **Important**

This fix makes `/v1/auth/session/refresh-token` reachable in community mode for the first time (previously `authenticationEndpoints.REFRESH_TOKEN` was `undefined`, so `getNewAccessToken` always POSTed to the API root and always returned `null`), which in turn makes the new `if (!refreshedAccessToken) await signOut()` branch in TanStackProvider live. That exposes an unpaired gap: the backend endpoint is `refreshAccessTokenFromCookie` (AuthController.java:79-83), i.e. it reads the refresh token from an http-only cookie, but `communitySignIn`/`communitySignUp` (authUtils.ts:268, 291) post *without* `{ withCredentials: true }`, unlike `getNewAccessToken` and `clearCookies`. If `NEXT_PUBLIC_API_URL` points at a different origin than the frontend, the Set-Cookie from sign-in is discarded, refresh returns null, and users get force-signed-out on the first version-mismatch response. Please verify the community sign-in response actually persists the refresh cookie, and add `{ withCredentials: true }` to `communitySignIn`/`communitySignUp` if not.

frontend/src/community/common/constants/routes.ts
```diff
@@ -186,3 +192,33 @@ export const employeeRestrictedRoutes = [
...
+  {
+    routes: [
+      ROUTES.PROJECTS.BASE,
+      ROUTES.PROJECTS.GUESTS,
+      ROUTES.PROJECTS.GUEST_REQUESTS
+    ],
+    requiredRole: EmployeeTypes.PM_EMPLOYEE
+  },
+  {
+    routes: [ROUTES.LEAVE.BASE],
```
ThinuwanW, line 217: 🤖 **Claude** · 🟡 **Suggestion**

The table only gates each module's own path prefix, so module-scoped pages living outside it stay reachable when the module is disabled. Concretely: `/dashboard/leave/resource-availability` renders leave data but is only gated by `/leave`; `/dashboard/attendance/clock-in-summary` and `/dashboard/attendance/late-arrivals-summary` are only gated by `/timesheet` — and `middleware.ts` grants `ROUTES.DASHBOARD.BASE` to every employee role via `commonRoutes`, so nothing stops a user without `ROLE_LEAVE_EMPLOYEE`/`ROLE_ATTENDANCE_EMPLOYEE` from opening them. Same class of gap for `/configurations/invoice`, `/configurations/esign` and `/configurations/attendance`: `adminRoutes` gives `CONFIGURATIONS.BASE` to `ATTENDANCE_ADMIN`, `ESIGN_ADMIN`, `INVOICE_ADMIN`, `CRM_ADMIN` and super admin, so e.g. an attendance admin can open `/configurations/invoice` with the invoice module off. Add these paths to the corresponding `routes` arrays.

frontend/src/community/common/constants/routes.ts
```diff
@@ -186,3 +192,33 @@ export const employeeRestrictedRoutes = [
...
+    routes: [ROUTES.CRM.BASE],
+    requiredRole: RepresentativeTypes.CRM_SALES_REPRESENTATIVE
+  },
+  {
+    routes: [ROUTES.INVOICE.BASE],
+    requiredRole: ManagerTypes.INVOICE_MANAGER
+  },
+  {
+    routes: [
+      ROUTES.PROJECTS.BASE,
```
ThinuwanW, line 210: 🤖 **Claude** · 🟡 **Suggestion**

`ROUTES.PROJECTS.BASE` is `"/projects/list"`, not `"/projects"` (routes.ts:143), so unlike every other entry here this one does not gate the module root — it enumerates three exact page paths. Any other `/projects/*` route added later is silently ungated even though the middleware matcher already covers `/projects/:path*`. Replace the three-element array with the single prefix `"/projects"` (or add a `ROUTES.PROJECTS.ROOT = "/projects"` constant and use that) so the gate is closed by default for new project pages.

frontend/src/community/common/providers/TanStackProvider.tsx
```diff
@@ -34,7 +34,13 @@ const TanStackProvider = ({ children }: { children: ReactNode }) => {
...
+        await signOut();
```
ThinuwanW, line 40: 🤖 **Claude** · 🟡 **Suggestion**

This `signOut()` duplicates one that already fires on the same failure. `authInterceptor.ts:42-48` registers a response interceptor on the very `authAxios` instance that `getNewAccessToken` uses, and it already calls `await signOut()` when the refresh response carries `COMMON_ERROR_MISSING_COOKIE_IN_TOKEN` or `COMMON_ERROR_INVALID_REFRESH_TOKEN` — the normal expired/invalid-refresh-token path. So that path now POSTs `/v1/auth/session/sign-out` twice and assigns `window.location.href` twice. It also fans out: `getNewAccessToken` shares one `refreshPromise` across concurrent callers (authUtils.ts:95-97), so every concurrently-failing request resolves to the same `null` and each independently calls `signOut()`. Either narrow this branch to the case the existing interceptor does not cover, or add an idempotency guard (a module-level `isSigningOut` flag) inside `signOut` itself.

frontend/src/community/common/components/templates/BaseLayout/BaseLayout.tsx
ThinuwanW, line 30: 🤖 **Claude** · 🟡 **Suggestion**

`BaseLayout` wraps *every* page from `_app.tsx` (lines 106 and 126) — including `pages/404.tsx`, `_error.tsx`, `/signin`, `/verify/guest-otp` and the public `/sign/document/access` flow that `middleware.ts:204-210` deliberately lets through. Consequence: an authenticated user without `ROLE_CRM_SALES_REPRESENTATIVE` who hits a non-existent URL under `/crm/...` is silently redirected to `/dashboard` instead of seeing the 404, and the same applies to error pages under any guarded prefix. Also note BaseLayout already calls `useSessionData()` on line 28, which derives `isCrmModuleEnabled`/`isInvoiceModuleEnabled`/`isPmModuleEnabled` from exactly the same `user.roles`; the new hook re-reads `useAuth()` independently. Consider scoping the guard to the drawer-protected routes (the existing `IsAProtectedUrlWithDrawer(asPath)` memo on line 56 is right there) rather than running it globally.

frontend/src/community/common/hooks/useModuleAccessGuard.ts
ThinuwanW, line 21: 🤖 **Claude** · 🟣 **Nit**

`void router.replace(...)` discards the promise without a rejection handler. Next's pages-router `replace` rejects when a navigation is cancelled or superseded (a real possibility here, since the effect can fire while another transition is in flight), which surfaces as an unhandled promise rejection. Use `router.replace(...).catch(() => {})`, or follow the established pattern in the sibling hook `useOrgSetupRedirect.tsx:24-26`, which calls `router.replace(targetPath)` plainly after a `router.pathname !== targetPath` check.

frontend/src/community/common/constants/routes.ts
ThinuwanW, line 198: 🤖 **Claude** · 🟣 **Nit**

Type modelling of the new table is looser than the rest of the auth layer. (1) `requiredRole` is typed `EmployeeTypes | ManagerTypes | RepresentativeTypes`, omitting `AdminTypes`, `SenderTypes` and `SuperAdminType` — so the table cannot express an admin- or sender-gated module (relevant if the eSign gate is added, which needs `SenderTypes.ESIGN_SENDER`/`AdminTypes.ESIGN_ADMIN` variants) without editing the type. Reuse the same union as `User["roles"]` in `authUtils.ts:70-77`. (2) The inline object literal type should be extracted to an exported named type (e.g. `ModuleGuardedRoute`) per the interface-definition convention, and the array marked `readonly`/`as const` — it is exported mutable configuration today. (3) `isModuleRouteRestricted(pathname: string, roles: string[])` widens `User["roles"]` to `string[]`, so a typo'd role literal would type-check; accept the role union instead.

frontend/src/community/common/utils/__test__/isModuleRouteRestricted.test.ts
```diff
@@ -0,0 +1,160 @@
...
+      expect(isModuleRouteRestricted(ROUTES.CRM.BASE, baseRoles)).toBe(true);
+      expect(isModuleRouteRestricted(ROUTES.CRM.CONTACTS, baseRoles)).toBe(
+        true
+      );
+      expect(isModuleRouteRestricted(ROUTES.CRM.COMPANIES, baseRoles)).toBe(
+        true
+      );
+      expect(isModuleRouteRestricted(ROUTES.CRM.DEALS, baseRoles)).toBe(true);
+      expect(isModuleRouteRestricted(ROUTES.CRM.TASKS, baseRoles)).toBe(true);
+      expect(isModuleRouteRestricted("/crm/deals/123", baseRoles)).toBe(true);
```
ThinuwanW, line 34: 🤖 **Claude** · 🟣 **Nit**

The dynamic-route cases assert on paths the app cannot serve, while the real ones are untested. `pages/community/crm/` contains only `companies.tsx`, `contacts.tsx`, `deals.tsx` and `tasks.tsx` — there is no `deals/[id].tsx` and no `index.tsx`, so both `"/crm/deals/123"` (here and at line 77) and `ROUTES.CRM.BASE` are 404s. Meanwhile the dynamic routes that do exist and that this guard must cover get no coverage: `/invoice/create/1`, `/invoice/view/1`, `/invoice/customers/customer-details/1`, `/invoice/customers/customer-details/projects/1`. Swapping in those cases would also pin down the `startsWith` prefix behaviour where it actually matters.

PR #1719, opened 2026-08-07
Fix/2341 people admin organization config access

shakila-kamalasena:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/frontend/middleware.ts
+++ b/frontend/middleware.ts
@@ -69,7 +69,10 @@ const superAdminRoutes = {
 };
 
 const adminRoutes = {
-  [AdminTypes.PEOPLE_ADMIN]: [ROUTES.PEOPLE.BASE],
+  [AdminTypes.PEOPLE_ADMIN]: [
+    ROUTES.PEOPLE.BASE,
+    ROUTES.CONFIGURATIONS.BASE
+  ],
   [AdminTypes.LEAVE_ADMIN]: [ROUTES.LEAVE.BASE],
   [AdminTypes.ATTENDANCE_ADMIN]: [
     ROUTES.TIMESHEET.BASE,


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR grants `PEOPLE_ADMIN` access to `/configurations` in the Next.js middleware route map. The change is correct and closes a real gap: the sidebar (`src/community/common/utils/data/routes.ts:423`) and the "organization" configuration tab (`src/community/configurations/utils/configurationTabsUtil.tsx:20`) already list `PEOPLE_ADMIN`, and the backend `WorkLocationController` already authorizes `ROLE_PEOPLE_ADMIN` for work-location CRUD, so without this change the nav item dead-ended at `/unauthorized`. Super-admin-only `/configurations/user-roles` remains protected via `userRolesRestrictedRoutes`, so no privilege escalation is introduced. The one notable issue is that the identical gap still exists for `LEAVE_ADMIN`, which this PR leaves unfixed in the very block it edits.

Found **9** new issue(s): 🔴 1 important, 🟣 1 nit(s)

frontend/middleware.ts
```diff
@@ -69,7 +69,10 @@ const superAdminRoutes = {
 };
 
 const adminRoutes = {
-  [AdminTypes.PEOPLE_ADMIN]: [ROUTES.PEOPLE.BASE],
+  [AdminTypes.PEOPLE_ADMIN]: [
+    ROUTES.PEOPLE.BASE,
+    ROUTES.CONFIGURATIONS.BASE
+  ],
   [AdminTypes.LEAVE_ADMIN]: [ROUTES.LEAVE.BASE],
```
ThinuwanW, line 76: 🤖 **Claude** · 🔴 **Important**

`LEAVE_ADMIN` has the exact same broken flow this PR fixes for `PEOPLE_ADMIN`, and it is left unfixed one line below the change. `src/community/common/utils/data/routes.ts:424` lists `AdminTypes.LEAVE_ADMIN` in the `requiredAuthLevel` for the "Configurations" sidebar item, and `src/community/configurations/utils/configurationTabsUtil.tsx:42` defines a `leave` tab whose `requiredRoles` are `[SUPER_ADMIN, LEAVE_ADMIN]`. But `allowedRoutes[LEAVE_ADMIN]` is only `["/leave"]`, and `/configurations` does not start with `/leave`, so a user whose only admin role is `LEAVE_ADMIN` (their other roles being `*_EMPLOYEE`, whose `commonRoutes` also omit `/configurations`) sees the nav item, clicks it, and gets redirected to `/unauthorized` — their leave configuration tab is unreachable. Unless this is deliberate, add `ROUTES.CONFIGURATIONS.BASE` to the `LEAVE_ADMIN` entry as well: `[AdminTypes.LEAVE_ADMIN]: [ROUTES.LEAVE.BASE, ROUTES.CONFIGURATIONS.BASE]`. If it is deliberate, remove `AdminTypes.LEAVE_ADMIN` from the Configurations `requiredAuthLevel` in `data/routes.ts` and drop the `leave` tab's `LEAVE_ADMIN` role so the nav does not advertise an inaccessible page.

frontend/middleware.ts
```diff
@@ -69,7 +69,10 @@ const superAdminRoutes = {
 };
 
 const adminRoutes = {
-  [AdminTypes.PEOPLE_ADMIN]: [ROUTES.PEOPLE.BASE],
+  [AdminTypes.PEOPLE_ADMIN]: [
+    ROUTES.PEOPLE.BASE,
+    ROUTES.CONFIGURATIONS.BASE
```
ThinuwanW, line 74: 🤖 **Claude** · 🟣 **Nit**

`isAllowed` uses `pathname.startsWith(url)`, so adding `ROUTES.CONFIGURATIONS.BASE` (`/configurations`) grants `PEOPLE_ADMIN` every current and future `/configurations/*` subroute, not just the "organization" tab it needs. Today this is safe — the only rewritten subroutes in `next.config.mjs` are `/configurations/user-roles*` (blocked for non-super-admins by `userRolesRestrictedRoutes`) and `/configurations/work-location/*` (intentionally allowed; backend `WorkLocationController` permits `ROLE_PEOPLE_ADMIN`). It matches the existing pattern for `ATTENDANCE_ADMIN`/`ESIGN_ADMIN`/`INVOICE_ADMIN`/`CRM_ADMIN`, so no change is required now, but be aware that any new `/configurations/<module>` page becomes reachable by every admin type by default and must be explicitly added to a restricted-routes list. Worth a short comment above `adminRoutes` noting that the entries are prefix matches so future subroutes need explicit denial.

frontend/middleware.ts
```diff
@@ -69,7 +69,10 @@ const superAdminRoutes = {
 };
 
 const adminRoutes = {
-  [AdminTypes.PEOPLE_ADMIN]: [ROUTES.PEOPLE.BASE],
+  [AdminTypes.PEOPLE_ADMIN]: [
```
ThinuwanW, line 72: 🤖 **Claude** · 💬 **Note**

📁 **File-level observation**

Dead code that silently violates the declared type (line 200, inside the `allowedRoutes` merge this PR extends). `commonRoutes` is an *array*, so `...commonRoutes` inside the object literal produces numeric keys `"0"`, `"1"`, … whose values are plain strings, not `string[]`. No role identifier is ever a numeric string, so `allowedRoutes[role]?.some(...)` never reads them — the spread contributes nothing. TypeScript does not catch it because spread properties bypass excess-property checking, so the `Record<AdminTypes | ... , string[]>` annotation is silently wrong (it now also has a numeric index of `string`). The practical hazard is that it reads as if every role inherits the common routes; in fact only the `employeeRoutes` entries explicitly spread `...commonRoutes`, so the `adminRoutes` entries this PR edits get none of `/dashboard`, `/settings`, `/account`, `/notifications`. That is masked today only because the backend role hierarchy (`AuthorityServiceImpl.addRoleHierarchy`) always issues `ROLE_PEOPLE_EMPLOYEE` alongside `ROLE_PEOPLE_ADMIN`. Fix: delete `...commonRoutes` from the merge, and if the intent was truly "everyone gets these", express it in the check instead — `const isAllowed = commonRoutes.some((url) => pathname.startsWith(url)) || roles.some(...)`.

frontend/middleware.ts
ThinuwanW, line 74: 🤖 **Claude** · 💬 **Note**

Who may reach `/configurations` is now hand-maintained in three unconnected places, which is why it drifted in the first place: (1) this `adminRoutes` table, (2) the drawer entry `id: "8"` `requiredAuthLevel` in `src/community/common/utils/data/routes.ts:418-426`, and (3) `requiredRoles` on each tab in `src/community/configurations/utils/configurationTabsUtil.tsx` / `src/enterprise/configurations/utils/configurationTabsUtil.tsx`. Nothing links them, so a role can be shown the nav item and have a visible tab while the middleware denies the route (or, worse in the other direction, be routed to a page with zero visible tabs and an empty body, since `pages/community/configurations/index.tsx` renders `visibleTabs.find(...)?.component` with no empty-state). Fix: export one constant (e.g. `CONFIGURATION_ROLES` derived from the union of `requiredRoles` across the tab factories, or at minimum a shared `CONFIGURATION_ADMIN_ROLES` array) and consume it from all three sites, so adding a configurations tab for a role automatically grants the route and the drawer item.

frontend/middleware.ts
ThinuwanW, line 72: 🤖 **Claude** · 💬 **Note**

📁 **File-level observation**

The role gate this PR widens is driven by an unverified, non-expiry-checked token (lines 205-207, 289-291). `extractClaimsFromToken` → `decodeJWTToken` (`src/community/auth/utils/authUtils.ts:59-64`) only base64-decodes the payload segment — no signature verification and no `exp` check — even though `isTokenExpired` lives in the same module and is used elsewhere. Two concrete consequences: (a) a user whose access token has expired still passes every role check here and is routed into `/configurations`, then hits 401s from the API instead of being sent to sign-in; (b) the payload is fully client-editable, so this table is a UX affordance, not an authorization boundary. The backend is the real boundary and it is correct for this change (`WorkLocationController` allows `ROLE_PEOPLE_ADMIN` on all six endpoints), so this is defense-in-depth rather than an exploit — but please add at least `if (isTokenExpired(token)) return NextResponse.redirect(new URL(ROUTES.AUTH.SIGNIN, request.url));` before the role evaluation, and add a comment stating the table is not an authorization boundary so future PRs don't move real access decisions here.

ThinuwanW, line 72: 🤖 **Claude** · 💬 **Note**

📁 **File-level observation**

Unhandled-exception risk in the gate (lines 226 and 289-291). `claims` is typed `Record<string, any>`, so `claims?.roles || []` is an unchecked `any` cast into a typed role array — the `|| []` only defends against `null`/`undefined`, not against a wrong shape. A token whose `roles` claim decodes to a non-array (e.g. a single string) makes `roles.some(...)` / `roles.includes(...)` throw a `TypeError`. `extractClaimsFromToken` swallows parse errors, but `middleware()` itself has no try/catch, and an uncaught throw in Next.js middleware returns a 500 for *every* path in `config.matcher`, not just the offending request. Fix: `const roles = Array.isArray(claims?.roles) ? (claims.roles as Role[]) : [];` and consider wrapping the body in a try/catch that falls back to redirecting to `ROUTES.AUTH.SIGNIN`.

ThinuwanW, line 72: 🤖 **Claude** · 💬 **Note**

📁 **File-level observation**

Duplicate key across the merged route objects: `SenderTypes.ESIGN_SENDER` is declared in both `managerRoutes` (lines 122-133) and `senderRoutes` (lines 171-182) with byte-identical 10-element arrays. Because `...senderRoutes` is spread last in `allowedRoutes`, the `managerRoutes` copy is silently discarded — so editing the `managerRoutes` one (the natural place to look, since it sits with the other manager roles) would have no effect at all. Delete the `ESIGN_SENDER` entry from `managerRoutes` and keep `senderRoutes` as the single definition.

ThinuwanW, line 72: 🤖 **Claude** · 💬 **Note**

📁 **File-level observation**

No test coverage for the role→route matrix. There is no `middleware.test.ts` anywhere under `frontend/`, despite the codebase having extensive Jest coverage elsewhere (including `src/community/common/utils/__test__/commonUtil.test.ts`, which sits next to `checkRestrictedRoutesAndRedirect` — the helper this file depends on — and has its JWT-decode tests commented out). This table is security-adjacent, edited one role at a time by hand, and has demonstrably been wrong for at least one role. Add a table-driven test that feeds a mocked `NextRequest` plus a stubbed `extractClaimsFromToken` and asserts allow/deny per `(role, path)` pair — at minimum covering `PEOPLE_ADMIN` → `/configurations` allowed and `PEOPLE_ADMIN` → `/configurations/user-roles` redirected to `/unauthorized`, so the `userRolesRestrictedRoutes` guard that is now the only thing protecting that sub-route cannot regress unnoticed.

ThinuwanW, line 72: 🤖 **Claude** · 💬 **Note**

📁 **File-level observation**

Nit — inconsistent identifiers for the same role string in the file the PR touches. `ROLE_SUPER_ADMIN` (the bare constant) is used at lines 41, 235 and 343, while `AdminTypes.SUPER_ADMIN` (the enum member, same `"ROLE_SUPER_ADMIN"` value) is used at line 325. They are interchangeable only by coincidence of value; if the enum member is ever renamed or re-valued, `superAdminRoutes` and the `checkRestrictedRoutesAndRedirect` calls would diverge without a type error. Standardise on `AdminTypes.SUPER_ADMIN` throughout (keeping `ROLE_SUPER_ADMIN` only where the `SuperAdminType` literal type is required for the `allowedRoutes` key).

frontend/middleware.ts
shakila-kamalasena: Only people admin should have access

frontend/middleware.ts
shakila-kamalasena: This is a change that need to be done

frontend/middleware.ts
shakila-kamalasena: Not related to this PR

frontend/middleware.ts
shakila-kamalasena: No need

frontend/middleware.ts
shakila-kamalasena: No need

shakila-kamalasena: No need

shakila-kamalasena: Not related to this PR

shakila-kamalasena: No need tests

shakila-kamalasena: No need

hasalarootcode (approved):

akilarootcode (approved):

PR #1720, opened 2026-08-07
Fix/2438 work location delete future holidays

shakila-kamalasena:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds cleanup of future, location-specific holidays when a work location is deleted, wiring `HolidayDao` into `WorkLocationServiceImpl`. The core intent is correct and the constructor change is already mirrored in the enterprise subclass, but the new helper hard-deletes holidays while bypassing every guard the canonical `HolidayServiceImpl` delete paths enforce (leave-request conflict check, soft delete), and it leaves past/inactive single-location holidays behind — which the join table's `ON DELETE CASCADE` then silently promotes to organization-wide holidays. There are also efficiency (N+1 + unbounded query), logging, notification-asymmetry, and test-coverage gaps.

Found **15** new issue(s): 🔴 4 important, 🟡 8 suggestion(s), 🟣 3 nit(s)

backend/src/main/java/com/skapp/community/common/service/impl/WorkLocationServiceImpl.java
```diff
@@ -231,6 +238,22 @@ else if (requestDto.getEmployeeIds() != null && !requestDto.getEmployeeIds().isE
...
+
+		List<Holiday> holidaysToDelete = locationHolidays.stream()
+			.filter(holiday -> holiday.getDate() != null && holiday.getDate().isAfter(currentDate))
+			.filter(holiday -> holiday.getWorkLocations()
+				.stream()
+				.allMatch(location -> location.getWorkLocationId().equals(workLocationId)))
+			.toList();
+
+		if (!holidaysToDelete.isEmpty()) {
+			holidayDao.deleteAll(holidaysToDelete);
```
ThinuwanW, line 253: 🤖 **Claude** · 🔴 **Important**

This hard-deletes holidays and bypasses every guard the canonical holiday-delete paths enforce in `HolidayServiceImpl`. `deleteSelectedHolidays` (a) rejects deletion when PENDING/APPROVED leave requests exist on the holiday date (`PEOPLE_ERROR_HOLIDAY_CANNOT_BE_DELETED_LEAVES_EXIST`) and (b) performs a *soft* delete (`holiday.setActive(false)`); `deleteAllHolidays` also runs the `hasLeaveRequestsForDate` check. None of that happens here. Concretely: when a holiday is declared, `saveBulkHolidays` → `updateLeaveRequestDueHoliday` CANCELS/REVOKES overlapping leave requests and deducts entitlement days. If the work location is later deleted, this code removes the holiday row outright, leaving affected employees with neither the leave request nor the holiday, and with entitlement already deducted — with no audit trail because the row is gone rather than deactivated. Fix: expose a method on `HolidayService` (e.g. `deleteFutureHolidaysByWorkLocation(Long workLocationId)`) that reuses the existing leave-conflict validation and the same soft-delete semantics as `deleteSelectedHolidays`, and call it from here through the `HolidayService` interface. That also removes the cross-module business logic (holiday domain rules) from the common `WorkLocationServiceImpl`.

backend/src/main/java/com/skapp/community/common/service/impl/WorkLocationServiceImpl.java
```diff
@@ -231,6 +238,22 @@ else if (requestDto.getEmployeeIds() != null && !requestDto.getEmployeeIds().isE
 		}
 	}
 
+	private void deleteFutureHolidaysSpecificToWorkLocation(Long workLocationId) {
+		List<Holiday> locationHolidays = holidayDao.findAllByIsActiveTrueAndWorkLocationsWorkLocationId(workLocationId);
+		LocalDate currentDate = DateTimeUtils.getCurrentUtcDate();
+
+		List<Holiday> holidaysToDelete = locationHolidays.stream()
+			.filter(holiday -> holiday.getDate() != null && holiday.getDate().isAfter(currentDate))
```
ThinuwanW, line 246: 🤖 **Claude** · 🔴 **Important**

Holidays excluded by these filters are silently converted into organization-wide holidays rather than left untouched. The join table `shr_com_work_location_ppl_holiday` declares `FOREIGN KEY (work_location_id) ... ON DELETE CASCADE`, so when `workLocationDao.delete(workLocation)` runs on line 119, the DB drops the join rows for every holiday that still referenced this location. Any holiday scoped ONLY to this location but skipped here — past/today-dated holidays (`isAfter(currentDate)` excludes them) and soft-deleted ones (the query filters `IsActiveTrue`) — ends up with an EMPTY `workLocations` set. Throughout the codebase an empty `workLocations` set means "applies to all work locations": `HolidayDao.findAllByIsActiveTrueAndWorkLocationsIsEmpty` (used by `LeaveServiceImpl.getHolidaysForEmployee`), the `criteriaBuilder.isEmpty(...)` OR-branch in `HolidayRepositoryImpl.findAllActiveHolidaysByWorkLocationId`/`buildPredicates`, and `HolidayServiceImpl.normalizeWorkLocations` (renders them as `HOLIDAY_ALL_WORK_LOCATIONS`). So a historical Colombo-only holiday retroactively becomes a company-wide holiday, changing past leave/attendance calculations and the holiday list UI for every employee. Handle the excluded holidays explicitly — e.g. also process non-future and inactive location-specific holidays (deactivate rather than delete the historical ones so history is preserved), or keep an explicit marker so they are not reinterpreted as global.

backend/src/main/java/com/skapp/community/common/service/impl/WorkLocationServiceImpl.java
```diff
@@ -231,6 +238,22 @@ else if (requestDto.getEmployeeIds() != null && !requestDto.getEmployeeIds().isE
 		}
 	}
 
+	private void deleteFutureHolidaysSpecificToWorkLocation(Long workLocationId) {
+		List<Holiday> locationHolidays = holidayDao.findAllByIsActiveTrueAndWorkLocationsWorkLocationId(workLocationId);
```
ThinuwanW, line 242: 🤖 **Claude** · 🟡 **Suggestion**

Query efficiency: two problems. (1) `findAllByIsActiveTrueAndWorkLocationsWorkLocationId` is unbounded — it loads every active holiday ever recorded for the location (all years) just to keep the future ones; push the date predicate into the query (e.g. add `findAllByIsActiveTrueAndDateAfterAndWorkLocationsWorkLocationId(currentDate, workLocationId)`). (2) `Holiday.workLocations` is `FetchType.LAZY` and this derived query does not fetch it, so the `allMatch` on line 247 triggers one extra SELECT against the join table per holiday — a textbook N+1. The codebase already has the fetch-join pattern for this (`HolidayRepositoryImpl.findAllActiveHolidaysByDateWithWorkLocations` uses `root.fetch(Holiday_.workLocations, JoinType.LEFT)`); add a `HolidayRepository` method that fetch-joins `workLocations` (or that returns only holidays whose work-location count is 1) instead of filtering in memory.

backend/src/main/java/com/skapp/community/common/service/impl/WorkLocationServiceImpl.java
```diff
@@ -231,6 +238,22 @@ else if (requestDto.getEmployeeIds() != null && !requestDto.getEmployeeIds().isE
 		}
 	}
 
+	private void deleteFutureHolidaysSpecificToWorkLocation(Long workLocationId) {
```
ThinuwanW, line 241: 🤖 **Claude** · 🟡 **Suggestion**

No logging on a destructive, irreversible operation. Per the project convention (`log.info("methodName: execution started")`, and see `EpWorkLocationServiceImpl.clearTimeRecordLocationsForWorkLocation` which logs its deletions), add entry logging plus a record of what was removed — e.g. `log.info("deleteFutureHolidaysSpecificToWorkLocation: deleting {} future holidays for work location {}", holidaysToDelete.size(), workLocationId)` with the holiday IDs. Without this there is no way to reconstruct which holidays disappeared, especially given the hard delete.

backend/src/main/java/com/skapp/community/common/service/impl/WorkLocationServiceImpl.java
```diff
@@ -6,9 +6,12 @@
 import com.skapp.community.common.payload.response.WorkLocationDetailResponseDto;
 import com.skapp.community.common.payload.response.WorkLocationEmployeeResponseDto;
 import com.skapp.community.common.payload.response.WorkLocationNameAvailabilityResponseDto;
+import com.skapp.community.common.util.DateTimeUtils;
```
ThinuwanW, line 9: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Notification asymmetry: holiday *creation* notifies affected employees (`HolidayServiceImpl.saveBulkHolidays` calls `peopleEmailService.sendNewHolidayDeclarationEmail` and `peopleNotificationService.sendNewHolidayDeclarationNotification`), but this new path silently cancels already-announced upcoming holidays with no email or in-app notification. Employees who were told about an upcoming holiday will simply find it gone from their calendar. Consider sending a holiday-cancellation notification (or at minimum confirm with product that silent removal is intended, and note it in the code).

backend/src/main/java/com/skapp/community/common/service/impl/WorkLocationServiceImpl.java
```diff
@@ -108,6 +114,7 @@ public ResponseEntityDto deleteWorkLocation(Long id) {
 		WorkLocation workLocation = workLocationDao.findById(id)
 			.orElseThrow(() -> new ModuleException(CommonMessageConstant.COMMON_ERROR_WORK_LOCATION_NOT_FOUND));
 
+		deleteFutureHolidaysSpecificToWorkLocation(id);
 		clearWorkLocationFromEmployees(id);
```
ThinuwanW, line 118: 🤖 **Claude** · 🟡 **Suggestion**

Pre-existing gap that this PR's new work now depends on: `clearWorkLocationFromEmployees` only nulls the work location for employees in `ACTIVE`/`PENDING` status, but `AccountStatus` also has `TERMINATED`, `DEACTIVATED`, and `DELETED`, and `FK_employee_work_location` (in `people-ddl-script-v1-alter-table-employee-add-work-location-id.sql`) is declared with no `ON DELETE` clause, i.e. RESTRICT. So deleting a work location that any non-active employee still references makes `workLocationDao.delete(workLocation)` on line 119 throw a raw `DataIntegrityViolationException` (500 to the client, not a `ModuleException`/`ErrorResponse`), and the holiday deletion added above is rolled back with it — the fix silently does nothing in exactly that case. Either clear the work location for all account statuses, or pre-check and throw a `ModuleException` with a proper message constant.

backend/src/main/java/com/skapp/community/common/service/impl/WorkLocationServiceImpl.java
ThinuwanW, line 9: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

No test coverage. There is currently no unit or integration test for `WorkLocationServiceImpl`/`WorkLocationController` anywhere under `src/test`, and this PR adds destructive behavior with non-obvious boundaries. Please add tests covering: a future holiday scoped only to the deleted location is removed; a future holiday scoped to the deleted location AND another location is kept (and remains valid for the other location); a past/today-dated holiday is not removed; and that the overall `deleteWorkLocation` response is unchanged.

backend/src/main/java/com/skapp/community/common/service/impl/WorkLocationServiceImpl.java
```diff
@@ -40,6 +44,8 @@ public class WorkLocationServiceImpl implements WorkLocationService {
 
 	private final EmployeeDao employeeDao;
 
+	private final HolidayDao holidayDao;
```
ThinuwanW, line 47: 🤖 **Claude** · 🟡 **Suggestion**

Adding `holidayDao` to this `@RequiredArgsConstructor` class changes the generated constructor signature, which the enterprise submodule depends on: `EpWorkLocationServiceImpl` calls `super(workLocationDao, employeeDao, holidayDao, messageUtil)`. The matching submodule change is not part of this diff (it shows as a modified `backend/src/main/java/com/skapp/enterprise` submodule pointer). Make sure the enterprise submodule PR is merged and the submodule pointer bumped in lockstep, otherwise the enterprise build fails to compile. Also note `EpWorkLocationServiceImpl.deleteWorkLocation` calls `super.deleteWorkLocation(id)` last, so the new holiday cleanup runs after the geofence/time-record cleanup — worth a quick confirmation that this ordering is intended.

backend/src/main/java/com/skapp/community/common/service/impl/WorkLocationServiceImpl.java
```diff
@@ -231,6 +238,22 @@ else if (requestDto.getEmployeeIds() != null && !requestDto.getEmployeeIds().isE
 		}
 	}
 
+	private void deleteFutureHolidaysSpecificToWorkLocation(Long workLocationId) {
+		List<Holiday> locationHolidays = holidayDao.findAllByIsActiveTrueAndWorkLocationsWorkLocationId(workLocationId);
+		LocalDate currentDate = DateTimeUtils.getCurrentUtcDate();
+
+		List<Holiday> holidaysToDelete = locationHolidays.stream()
+			.filter(holiday -> holiday.getDate() != null && holiday.getDate().isAfter(currentDate))
+			.filter(holiday -> holiday.getWorkLocations()
```
ThinuwanW, line 247: 🤖 **Claude** · 🟣 **Nit**

Dead/defensive code in the new helper. (a) `holiday.getDate() != null` can never be false — `Holiday.date` is mapped `@Column(name = "date", nullable = false)` and existing code (`HolidayServiceImpl.deleteAllHolidays`, `canDeleteHoliday`) dereferences it without a null check. (b) `allMatch` returns `true` for an empty stream; it is only safe here because the derived query guarantees at least one matching location — expressing the intent as `holiday.getWorkLocations().size() == 1` (or a count-based query predicate) states "specific to exactly this location" directly and stays correct if the query ever changes. (c) the `if (!holidaysToDelete.isEmpty())` guard on line 252 is redundant since `deleteAll` on an empty collection is a no-op (though keep the branch if you add the deletion log message suggested above).

backend/src/main/java/com/skapp/community/common/service/impl/WorkLocationServiceImpl.java
```diff
@@ -108,6 +114,7 @@ public ResponseEntityDto deleteWorkLocation(Long id) {
 		WorkLocation workLocation = workLocationDao.findById(id)
 			.orElseThrow(() -> new ModuleException(CommonMessageConstant.COMMON_ERROR_WORK_LOCATION_NOT_FOUND));
 
+		deleteFutureHolidaysSpecificToWorkLocation(id);
```
ThinuwanW, line 117: 🤖 **Claude** · 🔴 **Important**

Privilege escalation: this call makes holiday deletion reachable through an endpoint with weaker authorization than the holiday endpoints themselves. `WorkLocationController.deleteWorkLocation` is annotated `@PreAuthorize("hasAnyRole('ROLE_SUPER_ADMIN','ROLE_ATTENDANCE_ADMIN','ROLE_PEOPLE_ADMIN')")`, while both `HolidayController.deleteSelectedHolidays` and `deleteAllHolidays` require `hasAnyRole('ROLE_SUPER_ADMIN','ROLE_PEOPLE_ADMIN')`. I confirmed in `AuthorityServiceImpl.addModuleRoleAuthorities`/`addRoleHierarchy` that the hierarchy is flat and per-module — `ATTENDANCE_ADMIN` grants only `ATTENDANCE_ADMIN`/`ATTENDANCE_MANAGER`/`ATTENDANCE_EMPLOYEE`, never `PEOPLE_ADMIN`. So a user holding only the attendance-admin role can now permanently destroy holiday rows by deleting a work location, an operation they are explicitly denied on `/v1/holiday/**`. Fix by one of: (a) narrow the delete endpoint to `hasAnyRole('ROLE_SUPER_ADMIN','ROLE_PEOPLE_ADMIN')`; (b) keep the endpoint roles but perform the holiday removal via `HolidayService` behind its own `@PreAuthorize`/authorization check so attendance admins get a 403 rather than a silent cascade; or (c) if attendance admins are genuinely meant to be able to do this, make it deliberate and documented rather than an implicit side effect of a work-location delete.

backend/src/main/java/com/skapp/community/common/service/impl/WorkLocationServiceImpl.java
ThinuwanW, line 241: 🤖 **Claude** · 🔴 **Important**

Holiday-lifecycle business logic is now implemented in the common module's work-location service instead of the holiday module, and it duplicates an existing rule. The date predicate on the first filter is a verbatim copy of `HolidayServiceImpl.canDeleteHoliday(...)` (`holiday.getDate().isAfter(DateTimeUtils.getCurrentUtcDate())`), so the definition of 'a holiday that may still be deleted' now lives in two places and will drift the moment one is changed (e.g. if the org timezone ever replaces UTC there). It also means the deletion rules the holiday module owns — the leave-request guard, soft-delete via `setActive(false)`, and the declaration email/notification counterparts — cannot be reused here. Recommended fix: add `deleteFutureHolidaysByWorkLocation(Long workLocationId)` to the `HolidayService` interface, implement it in `HolidayServiceImpl` where `canDeleteHoliday`, the leave-request checks, `peopleEmailService` and `peopleNotificationService` already are, and inject `HolidayService` (not `HolidayDao`) here. There is no circular dependency risk: `HolidayServiceImpl` depends on `WorkLocationDao`, not on `WorkLocationService`. This also aligns with the interface-first / cross-module-through-the-service-interface convention.

backend/src/main/java/com/skapp/community/common/service/impl/WorkLocationServiceImpl.java
ThinuwanW, line 247: 🤖 **Claude** · 🟡 **Suggestion**

Null-guarding is inverted relative to the rest of the codebase, and the guard that is missing fails open. `Holiday.date` is mapped `@Column(name = "date", nullable = false)` yet it is null-checked on the preceding line, while `holiday.getWorkLocations()` is dereferenced with no guard even though `HolidayServiceImpl` defensively null-checks that exact collection in two places (`saveBulkHolidays`, where it does `holidayWorkLocations == null || holidayWorkLocations.isEmpty()`, and `countOverlappingHolidays`, `existingLocations == null || existingLocations.isEmpty()`). The direction matters here: `Stream.allMatch` returns `true` for an empty stream, so an empty (or, if the codebase's own assumption ever holds, null) location set makes the holiday look 'exclusive to this location' and eligible for permanent deletion. Today the derived query guarantees at least one matching location so it is unreachable, but the predicate is written to fail toward deleting rather than toward preserving. Suggest restructuring so the intent is explicit and safe, e.g. capture the set once, require it to be non-empty, and then require `size() == 1` with that single entry matching `workLocationId`.

backend/src/main/java/com/skapp/community/common/service/impl/WorkLocationServiceImpl.java
```diff
@@ -231,6 +238,22 @@ else if (requestDto.getEmployeeIds() != null && !requestDto.getEmployeeIds().isE
...
+		LocalDate currentDate = DateTimeUtils.getCurrentUtcDate();
+
+		List<Holiday> holidaysToDelete = locationHolidays.stream()
+			.filter(holiday -> holiday.getDate() != null && holiday.getDate().isAfter(currentDate))
+			.filter(holiday -> holiday.getWorkLocations()
+				.stream()
+				.allMatch(location -> location.getWorkLocationId().equals(workLocationId)))
+			.toList();
+
+		if (!holidaysToDelete.isEmpty()) {
```
ThinuwanW, line 252: 🤖 **Claude** · 🟣 **Nit**

Redundant guard (separate from the null-date check on line 246): `JpaRepository.deleteAll(Iterable)` iterates the argument and is already a no-op for an empty collection — it issues no SQL. The `if (!holidaysToDelete.isEmpty())` wrapper adds a branch with no behavioral effect. Drop it and call `holidayDao.deleteAll(holidaysToDelete)` unconditionally. (If the intent was to gate a log line on 'something was actually deleted', then keep a condition but make it do that — log the deleted count — rather than only guarding a call that already handles the empty case.)

backend/src/main/java/com/skapp/community/common/service/impl/WorkLocationServiceImpl.java
ThinuwanW, line 242: 🤖 **Claude** · 🟣 **Nit**

An existing repository method and query pattern already cover this and were not reused. `HolidayRepository.findAllActiveHolidaysByWorkLocationId(Long)` exists and is implemented in `HolidayRepositoryImpl` with the Criteria API, and its sibling `findAllActiveHolidaysByDateWithWorkLocations` shows the established pattern for this entity, including `root.fetch(Holiday_.workLocations, JoinType.LEFT)` with `distinct(true)` so the location set arrives eagerly. Instead this adds a use of the plain derived finder and then post-filters in Java, which re-implements in application code what the module's `*RepositoryImpl` is the designated place for. Suggest adding a Criteria method to `HolidayRepositoryImpl` (e.g. `findFutureHolidaysExclusiveToWorkLocation(Long workLocationId, LocalDate afterDate)`) that applies the `isActive`, `date > afterDate`, and location-exclusivity predicates in SQL with the `workLocations` fetch join, following the existing pattern — the helper here then just calls it and deletes the result.

backend/src/main/java/com/skapp/community/common/service/impl/WorkLocationServiceImpl.java
ThinuwanW, line 9: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

The API gives the client no way to anticipate or observe this destructive side effect, so an admin cannot make an informed decision. Before the call, `getWorkLocationById` populates `WorkLocationDetailResponseDto` with `employeeCount` and the employee list but exposes nothing about holidays, so the UI cannot warn 'deleting this location will permanently remove N upcoming holidays.' After the call, `deleteWorkLocation` returns the unchanged `COMMON_SUCCESS_WORK_LOCATION_DELETED` message with no count, so the caller cannot tell whether zero or fifty holidays were destroyed. Given the deletion is a hard delete and irreversible, suggest at minimum adding an affected-future-holiday count to `WorkLocationDetailResponseDto` (so the confirmation dialog can surface it), and returning the deleted count in the delete response via a new key in `CommonMessageConstant` plus the corresponding entry in the messages properties file, following the existing message-constant convention rather than a hardcoded string.

shakila-kamalasena: No need

shakila-kamalasena: No need

shakila-kamalasena: No need

backend/src/main/java/com/skapp/community/common/service/impl/WorkLocationServiceImpl.java
shakila-kamalasena: Added

backend/src/main/java/com/skapp/community/common/service/impl/WorkLocationServiceImpl.java
shakila-kamalasena: Not related to this PR

backend/src/main/java/com/skapp/community/common/service/impl/WorkLocationServiceImpl.java
shakila-kamalasena: Not related to this PR

PR #1721, opened 2026-08-07
feat: enhance contact search functionality with token-based name matc…

YasinduRC:
…hing

### Why

Contact names are now stored split into `first_name` / `last_name`. The existing search used a single `%keyword%` `LIKE` over each field plus a `first_name + ' ' + last_name` concat, which breaks once the name lives in two columns: `"smith john"` (reversed) and `"jo sm"` (partial across both) return nothing, and full-name matching relied on a concat that can't use an index.

### What changed

One file — `backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java`. A new helper `buildContactNameTokenPredicate(...)` replaces the contact-name portion of the predicate on both search surfaces: `GET /v1/crm/contact/metrics` (list filter) and `GET /v1/crm/contact/lookup` (global search).

**Algorithm** — trim + lowercase the keyword, split on `\s+`; each token must match `token%` or `% token%` against **both** `first_name` and `last_name`; all tokens `AND`-ed.

| Query | Matches `John` / `Paul Smith`? | Why |
|---|---|---|
| `john smith` | ✅ | tokens span both fields |
| `smith john` | ✅ | order-insensitive |
| `smith` | ✅ | inner token of `Paul Smith` |
| `jo sm` | ✅ | prefix per token |
| `john xyz` | ❌ | AND — every token must match |
| `ann` vs `Susanna` | ❌ | prefix, not substring |

### Design decisions

- **Prefix per token, not substring** — typeahead-standard, keeps results tight, index-friendly.
- **Order-insensitive + AND** — users don't recall which name is first, but every token they typed should count.
- **`% token%` inner-token pattern** — makes compound names work (`"smith"` → `Paul Smith`); matches only at a token start, so no mid-word false positives.
- **No Java-side case/accent folding** — `crm_contact` inherits the DB's `utf8mb4` collation, so `LIKE` is already case- and accent-insensitive, same as every other name search in the module.
- **Non-name branches untouched** — owner-name and company-name matching keep the original `%keyword%` pattern, per the task's out-of-scope note.

Follows existing repo idiom: same `cb.like(cb.lower(...), pattern, '\\')` form, `StringUtils.escapeLikePattern`, `Locale.ROOT`, `build*Predicate` helper naming.

### Related work

- Name-split backend and frontend changes — done, in separate PRs.
- Table migration — done. Data migration ships in its own PR.

### Not included (deliberate)

- **Result ranking** (exact → first-name → last-name prefix) — `/metrics` sorts by won-deal value, `/lookup` alphabetically; changing that is a product-sort decision, not part of matching.
- **Punctuation folding** — `"obrien"` does not match `O'Brien`; the collation doesn't fold apostrophes. Easy follow-up if wanted.

### Testing

- Verified end-to-end with all parts combined — contact search behaves correctly.
- `CrmContactControllerIntegrationTest` — 62/62 pass, no regression.
### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
+++ b/backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
@@ -18,6 +18,7 @@ import jakarta.persistence.EntityManager;
 import jakarta.persistence.TypedQuery;
 import jakarta.persistence.criteria.CriteriaBuilder;
 import jakarta.persistence.criteria.CriteriaQuery;
+import jakarta.persistence.criteria.Expression;
 import jakarta.persistence.criteria.Fetch;
 import jakarta.persistence.criteria.Join;
 import jakarta.persistence.criteria.JoinType;
@@ -84,7 +85,7 @@ public class CrmContactRepositoryImpl implements CrmContactRepository {
 		if (searchKeyword != null && !searchKeyword.isBlank()) {
 			String escaped = StringUtils.escapeLikePattern(searchKeyword.trim().toLowerCase(Locale.ROOT));
 			String likePattern = "%" + escaped + "%";
-			predicates.add(cb.or(cb.like(cb.lower(contact.get(CrmContact_.name)), likePattern, '\\'),
+			predicates.add(cb.or(buildContactNameTokenPredicate(cb, contact, searchKeyword),
 					cb.like(cb.lower(owner.get(Employee_.firstName)), likePattern, '\\'),
 					cb.like(cb.lower(owner.get(Employee_.lastName)), likePattern, '\\')));
 		}
@@ -95,6 +96,26 @@ public class CrmContactRepositoryImpl implements CrmContactRepository {
 		}
 
 		return predicates.toArray(new Predicate[0]);
+	}
+
+	private Predicate buildContactNameTokenPredicate(CriteriaBuilder cb, Root<CrmContact> contact,
+			String searchKeyword) {
+		Expression<String> firstName = cb.lower(contact.get(CrmContact_.firstName));
+		Expression<String> lastName = cb.lower(contact.get(CrmContact_.lastName));
+
+		String[] tokens = searchKeyword.trim().toLowerCase(Locale.ROOT).split("\\s+");
+
+		List<Predicate> tokenPredicates = new ArrayList<>();
+		for (String token : tokens) {
+			String escaped = StringUtils.escapeLikePattern(token);
+			String prefixPattern = escaped + "%";
+			String innerTokenPattern = "% " + escaped + "%";
+			tokenPredicates
+				.add(cb.or(cb.like(firstName, prefixPattern, '\\'), cb.like(firstName, innerTokenPattern, '\\'),
+						cb.like(lastName, prefixPattern, '\\'), cb.like(lastName, innerTokenPattern, '\\')));
+		}
+
+		return cb.and(tokenPredicates.toArray(new Predicate[0]));
 	}
 
 	private Long getContactTotalCount(CriteriaBuilder cb, CrmContactMetricRequestDto filterDto) {
@@ -150,7 +171,7 @@ public class CrmContactRepositoryImpl implements CrmContactRepository {
 		if (searchKeyword != null && !searchKeyword.isBlank()) {
 			String escaped = StringUtils.escapeLikePattern(searchKeyword.trim().toLowerCase(Locale.ROOT));
 			String likePattern = "%" + escaped + "%";
-			predicates.add(cb.or(cb.like(cb.lower(contact.get(CrmContact_.name)), likePattern, '\\'),
+			predicates.add(cb.or(buildContactNameTokenPredicate(cb, contact, searchKeyword),
 					cb.like(cb.lower(company.get(CrmCompany_.name)), likePattern, '\\')));
 		}
 
--- a/backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
+++ b/backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
@@ -20,6 +20,7 @@ import jakarta.persistence.criteria.CriteriaBuilder;
 import jakarta.persistence.criteria.CriteriaQuery;
 import jakarta.persistence.criteria.Expression;
 import jakarta.persistence.criteria.Fetch;
+import jakarta.persistence.criteria.From;
 import jakarta.persistence.criteria.Join;
 import jakarta.persistence.criteria.JoinType;
 import jakarta.persistence.criteria.Order;
@@ -56,7 +57,7 @@ public class CrmContactRepositoryImpl implements CrmContactRepository {
 				JoinType.LEFT);
 
 		query.where(buildPredicates(cb, contact, owner, company, filterDto));
-		query.orderBy(buildOrderBy(cb, contact, query));
+		query.orderBy(buildOrderBy(cb, contact, query, filterDto.getSearchKeyword()));
 
 		TypedQuery<CrmContact> typedQuery = entityManager.createQuery(query);
 		typedQuery.setFirstResult((int) pageable.getOffset());
@@ -65,7 +66,8 @@ public class CrmContactRepositoryImpl implements CrmContactRepository {
 		return new PageImpl<>(typedQuery.getResultList(), pageable, getContactTotalCount(cb, filterDto));
 	}
 
-	private List<Order> buildOrderBy(CriteriaBuilder cb, Root<CrmContact> contact, CriteriaQuery<CrmContact> query) {
+	private List<Order> buildOrderBy(CriteriaBuilder cb, Root<CrmContact> contact, CriteriaQuery<CrmContact> query,
+			String searchKeyword) {
 		Subquery<BigDecimal> dealValueSub = query.subquery(BigDecimal.class);
 		Root<CrmDeal> deal = dealValueSub.from(CrmDeal.class);
 		dealValueSub.select(cb.coalesce(cb.sum(deal.get(CrmDeal_.amount).cast(BigDecimal.class)), BigDecimal.ZERO))
@@ -73,7 +75,11 @@ public class CrmContactRepositoryImpl implements CrmContactRepository {
 					cb.equal(deal.get(CrmDeal_.stage).get(CrmDealStage_.stageType), CrmDealStageType.WON),
 					cb.isFalse(deal.get(CrmDeal_.isDeleted)));
 
-		return List.of(cb.desc(dealValueSub), cb.asc(contact.get(CrmContact_.id)));
+		List<Order> orders = new ArrayList<>(buildNameRelevanceOrders(cb, contact, searchKeyword));
+		orders.add(cb.desc(dealValueSub));
+		orders.add(cb.asc(contact.get(CrmContact_.id)));
+
+		return orders;
 	}
 
 	private Predicate[] buildPredicates(CriteriaBuilder cb, Root<CrmContact> contact, Join<CrmContact, Employee> owner,
@@ -85,7 +91,7 @@ public class CrmContactRepositoryImpl implements CrmContactRepository {
 		if (searchKeyword != null && !searchKeyword.isBlank()) {
 			String escaped = StringUtils.escapeLikePattern(searchKeyword.trim().toLowerCase(Locale.ROOT));
 			String likePattern = "%" + escaped + "%";
-			predicates.add(cb.or(buildContactNameTokenPredicate(cb, contact, searchKeyword),
+			predicates.add(cb.or(buildAllTokensMatchNamePredicate(cb, contact, searchKeyword),
 					cb.like(cb.lower(owner.get(Employee_.firstName)), likePattern, '\\'),
 					cb.like(cb.lower(owner.get(Employee_.lastName)), likePattern, '\\')));
 		}
@@ -98,15 +104,13 @@ public class CrmContactRepositoryImpl implements CrmContactRepository {
 		return predicates.toArray(new Predicate[0]);
 	}
 
-	private Predicate buildContactNameTokenPredicate(CriteriaBuilder cb, Root<CrmContact> contact,
+	private Predicate buildAllTokensMatchNamePredicate(CriteriaBuilder cb, From<?, CrmContact> contact,
 			String searchKeyword) {
 		Expression<String> firstName = cb.lower(contact.get(CrmContact_.firstName));
 		Expression<String> lastName = cb.lower(contact.get(CrmContact_.lastName));
 
-		String[] tokens = searchKeyword.trim().toLowerCase(Locale.ROOT).split("\\s+");
-
 		List<Predicate> tokenPredicates = new ArrayList<>();
-		for (String token : tokens) {
+		for (String token : searchKeyword.trim().toLowerCase(Locale.ROOT).split("\\s+")) {
 			String escaped = StringUtils.escapeLikePattern(token);
 			String prefixPattern = escaped + "%";
 			String innerTokenPattern = "% " + escaped + "%";
@@ -116,6 +120,30 @@ public class CrmContactRepositoryImpl implements CrmContactRepository {
 		}
 
 		return cb.and(tokenPredicates.toArray(new Predicate[0]));
+	}
+
+	private List<Order> buildNameRelevanceOrders(CriteriaBuilder cb, From<?, CrmContact> contact,
+			String searchKeyword) {
+		if (searchKeyword == null || searchKeyword.isBlank()) {
+			return List.of();
+		}
+
+		String normalizedKeyword = searchKeyword.trim().toLowerCase(Locale.ROOT);
+		String prefixPattern = StringUtils.escapeLikePattern(normalizedKeyword) + "%";
+
+		Expression<String> firstName = cb.lower(contact.get(CrmContact_.firstName));
+		Expression<String> lastName = cb.lower(contact.get(CrmContact_.lastName));
+		Expression<String> fullName = cb
+			.lower(cb.concat(cb.concat(contact.get(CrmContact_.firstName), " "), contact.get(CrmContact_.lastName)));
+
+		Expression<Integer> relevanceRank = cb.<Integer>selectCase()
+			.when(cb.equal(fullName, normalizedKeyword), 0)
+			.when(cb.equal(firstName, normalizedKeyword), 0)
+			.when(cb.like(firstName, prefixPattern, '\\'), 1)
+			.when(cb.like(lastName, prefixPattern, '\\'), 2)
+			.otherwise(3);
+
+		return List.of(cb.asc(relevanceRank));
 	}
 
 	private Long getContactTotalCount(CriteriaBuilder cb, CrmContactMetricRequestDto filterDto) {
@@ -153,7 +181,11 @@ public class CrmContactRepositoryImpl implements CrmContactRepository {
 		List<Predicate> predicates = buildLookupPredicates(cb, query, contact, company, filterDto);
 
 		query.where(predicates.toArray(new Predicate[0]));
-		query.orderBy(cb.asc(cb.lower(contact.get(CrmContact_.name))), cb.asc(contact.get(CrmContact_.id)));
+
+		List<Order> orders = new ArrayList<>(buildNameRelevanceOrders(cb, contact, filterDto.getSearchKeyword()));
+		orders.add(cb.asc(cb.lower(contact.get(CrmContact_.name))));
+		orders.add(cb.asc(contact.get(CrmContact_.id)));
+		query.orderBy(orders);
 
 		TypedQuery<CrmContact> typedQuery = entityManager.createQuery(query);
 		typedQuery.setFirstResult((int) pageable.getOffset());
@@ -171,7 +203,7 @@ public class CrmContactRepositoryImpl implements CrmContactRepository {
 		if (searchKeyword != null && !searchKeyword.isBlank()) {
 			String escaped = StringUtils.escapeLikePattern(searchKeyword.trim().toLowerCase(Locale.ROOT));
 			String likePattern = "%" + escaped + "%";
-			predicates.add(cb.or(buildContactNameTokenPredicate(cb, contact, searchKeyword),
+			predicates.add(cb.or(buildAllTokensMatchNamePredicate(cb, contact, searchKeyword),
 					cb.like(cb.lower(company.get(CrmCompany_.name)), likePattern, '\\')));
 		}
 
--- a/backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
+++ b/backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
@@ -20,7 +20,6 @@ import jakarta.persistence.criteria.CriteriaBuilder;
 import jakarta.persistence.criteria.CriteriaQuery;
 import jakarta.persistence.criteria.Expression;
 import jakarta.persistence.criteria.Fetch;
-import jakarta.persistence.criteria.From;
 import jakarta.persistence.criteria.Join;
 import jakarta.persistence.criteria.JoinType;
 import jakarta.persistence.criteria.Order;
@@ -104,13 +103,17 @@ public class CrmContactRepositoryImpl implements CrmContactRepository {
 		return predicates.toArray(new Predicate[0]);
 	}
 
-	private Predicate buildAllTokensMatchNamePredicate(CriteriaBuilder cb, From<?, CrmContact> contact,
+	private Predicate buildAllTokensMatchNamePredicate(CriteriaBuilder cb, Root<CrmContact> contact,
 			String searchKeyword) {
 		Expression<String> firstName = cb.lower(contact.get(CrmContact_.firstName));
 		Expression<String> lastName = cb.lower(contact.get(CrmContact_.lastName));
 
 		List<Predicate> tokenPredicates = new ArrayList<>();
 		for (String token : searchKeyword.trim().toLowerCase(Locale.ROOT).split("\\s+")) {
+			if (token.isEmpty()) {
+				continue;
+			}
+
 			String escaped = StringUtils.escapeLikePattern(token);
 			String prefixPattern = escaped + "%";
 			String innerTokenPattern = "% " + escaped + "%";
@@ -119,10 +122,10 @@ public class CrmContactRepositoryImpl implements CrmContactRepository {
 						cb.like(lastName, prefixPattern, '\\'), cb.like(lastName, innerTokenPattern, '\\')));
 		}
 
-		return cb.and(tokenPredicates.toArray(new Predicate[0]));
+		return tokenPredicates.isEmpty() ? cb.conjunction() : cb.and(tokenPredicates.toArray(new Predicate[0]));
 	}
 
-	private List<Order> buildNameRelevanceOrders(CriteriaBuilder cb, From<?, CrmContact> contact,
+	private List<Order> buildNameRelevanceOrders(CriteriaBuilder cb, Root<CrmContact> contact,
 			String searchKeyword) {
 		if (searchKeyword == null || searchKeyword.isBlank()) {
 			return List.of();
--- a/backend/src/test/java/com/skapp/community/crmplanner/controller/v1/CrmContactControllerIntegrationTest.java
+++ b/backend/src/test/java/com/skapp/community/crmplanner/controller/v1/CrmContactControllerIntegrationTest.java
@@ -1073,4 +1073,60 @@ class CrmContactControllerIntegrationTest {
 			.andExpect(status().isForbidden());
 	}
 
+	// --- token-based name search ---
+
+	@Test
+	@DisplayName("Lookup with reversed token order - Tokens match across first and last name")
+	void getContactsLookup_ReversedTokenOrder_MatchesContact() throws Exception {
+		savedNamedContact("John", "Smith", null, "john.smith@lookup.com");
+
+		performRequest(get(LOOKUP_PATH).param("searchKeyword", "smith john").accept(MediaType.APPLICATION_JSON))
+			.andDo(print())
+			.andExpect(status().isOk())
+			.andExpect(jsonPath(STATUS_PATH).value(STATUS_SUCCESSFUL))
+			.andExpect(jsonPath("['results'][0]['totalItems']").value(1))
+			.andExpect(jsonPath("['results'][0]['items'][0]['firstName']").value("John"));
+	}
+
+	@Test
+	@DisplayName("Lookup where one token matches nothing - All tokens must match")
+	void getContactsLookup_TokenWithNoMatch_ReturnsNoResults() throws Exception {
+		savedNamedContact("John", "Smith", null, "john.smith@lookup.com");
+
+		performRequest(get(LOOKUP_PATH).param("searchKeyword", "john xyz").accept(MediaType.APPLICATION_JSON))
+			.andDo(print())
+			.andExpect(status().isOk())
+			.andExpect(jsonPath(STATUS_PATH).value(STATUS_SUCCESSFUL))
+			.andExpect(jsonPath("['results'][0]['totalItems']").value(0));
+	}
+
+	@Test
+	@DisplayName("Lookup with a token prefix - Matches name token start, not a mid-token substring")
+	void getContactsLookup_PrefixNotSubstring_ExcludesMidTokenMatch() throws Exception {
+		savedNamedContact("Anna", "Bell", null, "anna.bell@lookup.com");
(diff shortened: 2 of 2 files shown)


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR replaces the `contact.name LIKE %kw%` predicate in both contact search paths with a token-based matcher over `CrmContact_.firstName`/`CrmContact_.lastName`. The change does not compile — the `CrmContact` entity maps only `name` and has no `firstName`/`lastName` fields, so the generated static metamodel `CrmContact_` has no such attributes. Even if the entity were extended, nothing in the application ever populates `first_name`/`last_name` (the create/edit DTOs and service only handle `name`, and there is no backfill migration), so contact-name search would silently return nothing for every existing and newly created contact; two existing integration tests already assert the old behaviour and would fail.

Found **16** new issue(s): 🔴 8 important, 🟡 5 suggestion(s), 🟣 3 nit(s)

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
```diff
@@ -97,6 +98,26 @@ private Predicate[] buildPredicates(CriteriaBuilder cb, Root<CrmContact> contact
 		return predicates.toArray(new Predicate[0]);
 	}
 
+	private Predicate buildContactNameTokenPredicate(CriteriaBuilder cb, Root<CrmContact> contact,
+			String searchKeyword) {
+		Expression<String> firstName = cb.lower(contact.get(CrmContact_.firstName));
```
ThinuwanW, line 109: 🤖 **Claude** · 🔴 **Important**

This will not compile. `CrmContact_.firstName` and `CrmContact_.lastName` do not exist. The `CrmContact` entity (crmplanner/model/CrmContact.java) maps only `id`, `name`, `email`, `contactNumber`, `lastContactAt`, `company`, `owner`, `isDeleted` — there are no first/last name fields anywhere in the CRM module (grep for `firstName` in crmplanner only hits `Employee_.firstName` and owner response DTOs). `CrmContact_` is generated at build time by `hibernate-processor` (backend/pom.xml lines 139, 425) from the entity, so both attribute references are unresolved symbols. Fix: add the mapped fields to `CrmContact` (the `crm_contact` table already declares nullable `first_name` / `last_name` columns in common-ddl-script-v1-create-tables.sql:625-626) before referencing them here, or revert to `CrmContact_.name`.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
```diff
@@ -84,7 +85,7 @@ private Predicate[] buildPredicates(CriteriaBuilder cb, Root<CrmContact> contact
 		if (searchKeyword != null && !searchKeyword.isBlank()) {
 			String escaped = StringUtils.escapeLikePattern(searchKeyword.trim().toLowerCase(Locale.ROOT));
 			String likePattern = "%" + escaped + "%";
-			predicates.add(cb.or(cb.like(cb.lower(contact.get(CrmContact_.name)), likePattern, '\\'),
+			predicates.add(cb.or(buildContactNameTokenPredicate(cb, contact, searchKeyword),
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Functional regression: contact search no longer matches the contact's `name` at all, in either `findContacts` (this line) or `findContactsForLookup` (line 174). Nothing in the application writes `first_name`/`last_name`: `CrmContactCreateRequestDto` and `CrmContactEditRequestDto` carry only `name`, no mapper/service sets them, and there is no Liquibase backfill for existing rows. Because `LIKE` against NULL evaluates to NULL (not true), every existing and newly created contact becomes unfindable by name — the lookup endpoint would only match on company name, and the metrics endpoint only on owner first/last name. Existing tests `getContactsLookup_KeywordMatchesContactNameOnly_ReturnsContact` and `getContactsLookup_ContactWithoutCompany_MatchedByName` (CrmContactControllerIntegrationTest.java:1028-1053) will fail. Fix by either (a) tokenising `CrmContact_.name` itself instead of first/last name, or (b) OR-ing the original `cb.like(cb.lower(contact.get(CrmContact_.name)), likePattern, '\\')` back in alongside the token predicate, and separately adding entity fields + create/update population + a data backfill changeset if first/last name is genuinely being introduced.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
```diff
@@ -97,6 +98,26 @@ private Predicate[] buildPredicates(CriteriaBuilder cb, Root<CrmContact> contact
...
+		Expression<String> firstName = cb.lower(contact.get(CrmContact_.firstName));
+		Expression<String> lastName = cb.lower(contact.get(CrmContact_.lastName));
+
+		String[] tokens = searchKeyword.trim().toLowerCase(Locale.ROOT).split("\\s+");
+
+		List<Predicate> tokenPredicates = new ArrayList<>();
+		for (String token : tokens) {
+			String escaped = StringUtils.escapeLikePattern(token);
+			String prefixPattern = escaped + "%";
+			String innerTokenPattern = "% " + escaped + "%";
```
ThinuwanW, line 116: 🤖 **Claude** · 🔴 **Important**

The match semantics are narrowed in two ways that will surprise users. (1) The old pattern `%kw%` matched a substring anywhere; the new patterns are `kw%` (prefix) and `% kw%` (after a literal space), so searching "ohn" no longer finds "John". (2) The word-boundary pattern only recognises a single space as a separator — hyphenated or apostrophised names ("Mary-Jane", "O'Brien") and comma/tab-separated stored values will not word-match. If prefix/word-start matching is the intent, document it and make the separator handling explicit; otherwise keep an additional `%token%` alternative so substring search still works. Note also that all tokens are AND-ed (line 118), so multi-word keywords are strictly narrower than before — a keyword spanning contact name and owner name (e.g. "John Smith" where John is the contact and Smith the owner) now matches nothing, since the owner branch on lines 89-90 still uses the whole unsplit keyword.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
```diff
@@ -18,6 +18,7 @@
 import jakarta.persistence.TypedQuery;
 import jakarta.persistence.criteria.CriteriaBuilder;
 import jakarta.persistence.criteria.CriteriaQuery;
+import jakarta.persistence.criteria.Expression;
```
ThinuwanW, line 21: 🤖 **Claude** · 🔴 **Important**

📁 **File-level observation**

No tests were added or updated for a behavioural search change. Per the review guidelines, repository/service behaviour changes need corresponding test updates. This PR changes matching semantics for two public endpoints (contact metrics list and contact lookup) and their count queries, and it breaks two existing assertions in CrmContactControllerIntegrationTest. Please add integration tests covering: multi-token keyword, prefix vs. inner-word match, a contact with NULL first/last name, and that `findContacts`/`getContactTotalCount` and `findContactsForLookup`/`getLookupTotalCount` stay consistent (totalItems must equal the filtered item count).

ThinuwanW, line 21: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Inconsistent with the existing CRM search convention. `CrmDealRepositoryImpl` (lines 117-120, 168-170) and `CrmContactOwnerRepositoryImpl` (lines 92-96) already implement full-name search via `cb.concat(firstName, " ", lastName)` + a single LIKE. This PR introduces a third, different matching strategy inside the same module, so the same keyword now behaves differently across the contact, deal, and owner search endpoints. Consider extracting one shared name-search helper (e.g. into a CRM search util or `StringUtils`) and using it from all three repositories so results are predictable.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
```diff
@@ -97,6 +98,26 @@ private Predicate[] buildPredicates(CriteriaBuilder cb, Root<CrmContact> contact
 		return predicates.toArray(new Predicate[0]);
 	}
 
+	private Predicate buildContactNameTokenPredicate(CriteriaBuilder cb, Root<CrmContact> contact,
+			String searchKeyword) {
+		Expression<String> firstName = cb.lower(contact.get(CrmContact_.firstName));
+		Expression<String> lastName = cb.lower(contact.get(CrmContact_.lastName));
+
+		String[] tokens = searchKeyword.trim().toLowerCase(Locale.ROOT).split("\\s+");
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Duplicated normalisation. Both call sites (lines 86 and 172) already compute `searchKeyword.trim().toLowerCase(Locale.ROOT)` before invoking this helper, and the helper redoes it. Pass the already-normalised keyword (or the token array) as the parameter to avoid the duplicate work and the risk of the two normalisations drifting apart. Also note the callers pass the raw `searchKeyword` while their own `likePattern` uses the normalised one — easy to get out of sync later.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
```diff
@@ -97,6 +98,26 @@ private Predicate[] buildPredicates(CriteriaBuilder cb, Root<CrmContact> contact
...
+		for (String token : tokens) {
+			String escaped = StringUtils.escapeLikePattern(token);
+			String prefixPattern = escaped + "%";
+			String innerTokenPattern = "% " + escaped + "%";
+			tokenPredicates
+				.add(cb.or(cb.like(firstName, prefixPattern, '\\'), cb.like(firstName, innerTokenPattern, '\\'),
+						cb.like(lastName, prefixPattern, '\\'), cb.like(lastName, innerTokenPattern, '\\')));
+		}
+
+		return cb.and(tokenPredicates.toArray(new Predicate[0]));
```
ThinuwanW, line 122: 🤖 **Claude** · 🟡 **Suggestion**

Performance: `% token%` has a leading wildcard so it can never use an index, and `first_name`/`last_name` are `text` columns (unindexed) in `crm_contact`. Each search now issues 4 LIKE predicates per token against a full scan, and `buildPredicates`/`buildLookupPredicates` are evaluated twice per request (once for the page, once in `getContactTotalCount`/`getLookupTotalCount`), so cost scales with token count on both queries. Consider bounding the number of tokens processed, or indexing a generated/normalised search column, if contact volume is expected to grow.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
```diff
@@ -97,6 +98,26 @@ private Predicate[] buildPredicates(CriteriaBuilder cb, Root<CrmContact> contact
...
 
+	private Predicate buildContactNameTokenPredicate(CriteriaBuilder cb, Root<CrmContact> contact,
+			String searchKeyword) {
+		Expression<String> firstName = cb.lower(contact.get(CrmContact_.firstName));
+		Expression<String> lastName = cb.lower(contact.get(CrmContact_.lastName));
+
+		String[] tokens = searchKeyword.trim().toLowerCase(Locale.ROOT).split("\\s+");
+
+		List<Predicate> tokenPredicates = new ArrayList<>();
+		for (String token : tokens) {
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Defensive gap on empty tokens. `isBlank()` in the callers uses `Character.isWhitespace`, while `trim()` strips all chars <= U+0020, so an input such as "\u0000" passes the guard but trims to an empty string; `"".split("\\s+")` yields a single empty token, producing `firstName LIKE '%'` — i.e. the keyword filter degenerates into match-everything-with-a-non-null-name. Skip empty tokens and return `cb.conjunction()`/omit the predicate when no usable tokens remain, rather than relying on `cb.and(new Predicate[0])`.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
ThinuwanW, line 109: 🤖 **Claude** · 🔴 **Important**

The `first_name` / `last_name` columns this search reads are never populated by any code, so even after the metamodel compile error is fixed the feature returns nothing. Evidence: `common-ddl-script-v1-create-tables.sql:625-626` declares `first_name text DEFAULT NULL` and `last_name text DEFAULT NULL`, but (a) `CrmContact.java` maps only `name`, `email`, `contactNumber`, `lastContactAt`, `company`, `owner`, `isDeleted` — no first/last name fields; (b) `CrmContactServiceImpl.createContact` (line 119) and `editContact` (line 151) set only `contact.setName(...)`; (c) no request/response DTO or `CrmMapper` mapping references them; (d) the frontend `frontend/src/community/crm/types/CommonTypes.ts` `CrmContact` type carries a single `name` (firstName/lastName there belong to `CrmOwner`). Because `col LIKE 'x%'` evaluates to NULL (not TRUE) when the column is NULL, every contact fails the predicate and contact search silently returns an empty set for 100% of existing and newly created data. Completing this needs the full set of paired changes, not just the repository: add `firstName`/`lastName` to `CrmContact`, accept and persist them in `CrmContactCreateRequestDto`/`CrmContactEditRequestDto` + the service, expose them in the response DTO/mapper, and add a Liquibase changeset that backfills the two columns from the existing `name` for all rows (plus a decision on whether `name` becomes derived or stays authoritative). If that scope is not intended for this PR, keep matching `CrmContact_.name` in addition to the token predicate so the change is purely additive.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
ThinuwanW, line 21: 🤖 **Claude** · 🔴 **Important**

📁 **File-level observation**

Three existing integration tests in `CrmContactControllerIntegrationTest` will fail with this change — this is separate from the 'no new tests' point, these are pre-existing green tests that regress to zero results: (1) `getContactMetrics_FilteredByKeyword_ReturnsMatchingContacts` (line ~858) searches `searchKeyword="Test Contact"` and asserts `totalItems == 1`; (2) `getContactsLookup_KeywordMatchesContactNameOnly_ReturnsContact` (line ~1034) searches `"john"` for contact `"John Doe"` and asserts `totalItems == 1` and `items[0].name == "John Doe"`; (3) `getContactsLookup_ContactWithoutCompany_MatchedByName` (line ~1047) searches `"solo"` for `"Solo Contact"` created with a null company. Case (3) is the most important signal: its `@DisplayName` ("Not dropped by the company join") shows it was written to guard a specific fix — a contact with no company could previously only be found via the contact-name branch of the OR, and removing that branch silently reintroduces the bug it protects against. Please run `CrmContactControllerIntegrationTest` before merging and extend these cases (reversed-token order, single-token, multi-token) rather than deleting or weakening them.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The number of SQL predicates is now driven by unbounded user input. `searchKeyword` in `CrmContactMetricRequestDto` and `CrmContactFilterDto` has no `@Size` bound, the controller methods (`CrmContactController.getContactMetrics` / `getContactsLookup`) bind the filter DTO without `@Valid`, and `split("\\s+")` produces one token per whitespace-separated run. Each token adds a 4-way OR of leading-wildcard LIKEs, and `buildPredicates` runs twice per request (data query + `getContactTotalCount`). A 4000-character keyword of alternating characters and spaces yields ~2000 tokens → ~8000 LIKE conditions per query, ~16000 per request, all full-scan on unindexed `text` columns. Before this change the predicate count was constant (3) regardless of keyword length, so this is newly introduced. Fix: add `@Size(max = 100)` to `searchKeyword` on both filter DTOs and `@Valid` on the controller parameters, and additionally cap the token list inside the helper (e.g. take the first 5 tokens after splitting) so the query shape is bounded even if a caller bypasses bean validation.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Contact-name search is now inconsistent across the CRM module because the sibling repositories were not updated. `CrmDealRepositoryImpl:116` and `:167` still match contacts with `cb.like(cb.lower(contactJoin.get(CrmContact_.name)), "%kw%")`, and `CrmTaskRepositoryImpl:228` does the same. After this PR the identical keyword behaves differently depending on the screen: on the contacts list, "doe john" matches contact "John Doe" (token AND, order-insensitive) but "joh doe" / a mid-word substring does not; on the deals and tasks lists the opposite holds. Users searching the same person across pages will get contradictory results. Either extend the same token predicate to the deal and task contact-name searches (which is why the helper should be reusable — see the `From<?, CrmContact>` note), or keep contacts on the existing `%kw%`-over-`name` convention and drop the divergence.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
ThinuwanW, line 21: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Sorting was not updated alongside matching, and the module's relevance-ranking convention is not applied. `findAllContactsForBoardInit` (line 140) and `findContactsForLookup` (line 156) still order by `cb.lower(contact.get(CrmContact_.name))` while matching now happens on first/last name — if `name` ever diverges from `firstName + " " + lastName` (nothing enforces consistency), the lookup dropdown orders by one value and filters by another. Separately, `CrmCompanyRepositoryImpl.buildOrderBy` (lines 134-152) establishes the CRM convention for keyword search: a `selectCase` relevance rank of exact match (0) → prefix match (1) → contains (2), asserted by `CrmCompanyControllerIntegrationTest` ("exact match first, then prefix match, then contains match"). The new contact search has no equivalent, and `findContacts` still orders solely by won-deal value, so an exact name match on a low-value contact is pushed below unrelated high-value partial matches — the opposite of what a user typing a full name expects. Consider adding a matching relevance rank as the first `Order` when a keyword is present.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
```diff
@@ -97,6 +98,26 @@ private Predicate[] buildPredicates(CriteriaBuilder cb, Root<CrmContact> contact
 		return predicates.toArray(new Predicate[0]);
 	}
 
+	private Predicate buildContactNameTokenPredicate(CriteriaBuilder cb, Root<CrmContact> contact,
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Signature and naming make this helper harder to reuse and harder to read. (1) It takes `Root<CrmContact>`, but the module's convention for a shared predicate helper is `From<?, CrmContact>` — see `CrmCompanyRepositoryImpl.buildPredicates(CriteriaBuilder, From<?, CrmCompany>, String)` at line 121, which is exactly what lets it serve both the entity query and the count query root. Widening the parameter to `From<?, CrmContact>` would also let `CrmDealRepositoryImpl` and `CrmTaskRepositoryImpl` reuse it through their `Join<CrmDeal, CrmContact>` / `Join<CrmTask, CrmContact>`, which is the cleanest way to resolve the cross-repository inconsistency. (2) The name `buildContactNameTokenPredicate` reads as "a predicate over name tokens" and hides the key semantic — every token must match (`cb.and`), not any. Rename to something like `buildAllTokensMatchNamePredicate`, or add a short comment stating the AND-across-tokens / OR-across-fields contract, so the next reader does not assume OR.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
ThinuwanW, line 116: 🤖 **Claude** · 🟣 **Nit**

`innerTokenPattern = "% " + escaped + "%"` treats a single literal ASCII space as the only word boundary, which is narrower than the token splitter that produced these tokens (`split("\\s+")` accepts tab, newline, form feed, CR and collapses runs). Concretely: for stored name "O'Brien" the token `brien` matches neither `brien%` nor `% brien%`; same for "Smith-Jones" + `jones`, "Doe, John" + `john`, and any stored value where the separator is a tab or a non-breaking space rather than U+0020. Since the token predicate is the only path to a contact-name match after this change, these are hard misses, not degraded ranking. If word-boundary matching is the goal, either normalise separators on the stored side (a generated/normalised search column) or accept the simpler `%token%` substring form; if the current form is intentional, at minimum add the common punctuation boundaries and document the limitation.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
```diff
@@ -150,7 +171,7 @@ private <T> List<Predicate> buildLookupPredicates(CriteriaBuilder cb, CriteriaQu
 		if (searchKeyword != null && !searchKeyword.isBlank()) {
 			String escaped = StringUtils.escapeLikePattern(searchKeyword.trim().toLowerCase(Locale.ROOT));
 			String likePattern = "%" + escaped + "%";
-			predicates.add(cb.or(cb.like(cb.lower(contact.get(CrmContact_.name)), likePattern, '\\'),
+			predicates.add(cb.or(buildContactNameTokenPredicate(cb, contact, searchKeyword),
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

API documentation drifts from behaviour. `CrmContactController.getContactsLookup` (line 52) documents this endpoint as "Supports searching by contact name or company name", which is published to Swagger. After this change the contact side matches first/last name tokens with prefix/word-start semantics, not the contact `name` field, and the two sides of the OR now use different matching rules (token-prefix for the contact, `%kw%` substring for the company). Update the `@Operation` description to describe the actual behaviour once the semantics are settled.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
YasinduRC: Added Tests

YasinduRC: The requirement is give to use this

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
YasinduRC: valid ,The backend and frontend changes for the contact name are approved but haven’t been merged yet. When merging, they should be merged in this order.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
YasinduRC: The `first_name` and `last_name` fields will be populated as part of the table migration, so they won't remain `NULL` once the migration is applied.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
YasinduRC: Prefix matching is intentional, it's what the requirement asks for (ann should match Anna, not Susanna)

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
YasinduRC: the name split is a separate PR that's already approved, and the columns get populated there

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
YasinduRC: Ran the full CrmContactControllerIntegrationTest with the seperate branch that contains  the name split backend 66/66 green, including the two you flagged.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
YasinduRC: As a practice we do not use DTO Validations and no controller validates its filter DTO

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
YasinduRC: Deal and task search will still work , the name-split PR already moves them to firstName/lastName, just with the old substring matching, not this PR's tokenizer.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
YasinduRC: The description says "Supports searching by contact name or company name" , still true. It describes what is searchable, not the matching rules, so nothing became inaccurate.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
YasinduRC: Added, buildNameRelevanceOrders follows the CrmCompanyRepositoryImpl.buildOrderBy selectCase pattern: exact full name (0) → first-name prefix (1) → last-name prefix (2) → everything else (3), applied as the first Order on both endpoints when a keyword is present.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
YasinduRC: Leaving as is. Nothing in this codebase indexes search columns

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
YasinduRC: This is actually the pattern we use everywhere,Keeping it consistent with that rather than introducing something new here.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
YasinduRC: Fixed ,  empty tokens are now skipped before building the LIKE patterns, and if nothing's left after that, the filter is dropped entirely instead of matching everything.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
YasinduRC: This one's intentional , the requirement only asks for the query to be matched as typed against the name

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
YasinduRC: Keeping Root<CrmContact> , nothing outside this file calls the helper, so there's no reuse case to widen for. Renamed to buildAllTokensMatchNamePredicate to make the AND explicit, as suggested.

shakila-kamalasena (approved):

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
```diff
@@ -97,6 +103,52 @@ private Predicate[] buildPredicates(CriteriaBuilder cb, Root<CrmContact> contact
 		return predicates.toArray(new Predicate[0]);
 	}
 
+	private Predicate buildAllTokensMatchNamePredicate(CriteriaBuilder cb, Root<CrmContact> contact,
+			String searchKeyword) {
+		Expression<String> firstName = cb.lower(contact.get(CrmContact_.firstName));
+		Expression<String> lastName = cb.lower(contact.get(CrmContact_.lastName));
+
+		List<Predicate> tokenPredicates = new ArrayList<>();
+		for (String token : searchKeyword.trim().toLowerCase(Locale.ROOT).split("\\s+")) {
```
akilarootcode, line 112: trim() remove

akilarootcode, line 112: Locale.ROOT remove

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
```diff
@@ -84,7 +90,7 @@ private Predicate[] buildPredicates(CriteriaBuilder cb, Root<CrmContact> contact
 		if (searchKeyword != null && !searchKeyword.isBlank()) {
 			String escaped = StringUtils.escapeLikePattern(searchKeyword.trim().toLowerCase(Locale.ROOT));
 			String likePattern = "%" + escaped + "%";
-			predicates.add(cb.or(cb.like(cb.lower(contact.get(CrmContact_.name)), likePattern, '\\'),
+			predicates.add(cb.or(buildAllTokensMatchNamePredicate(cb, contact, searchKeyword),
 					cb.like(cb.lower(owner.get(Employee_.firstName)), likePattern, '\\'),
```
akilarootcode, line 94: remove '\\'

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
```diff
@@ -97,6 +103,52 @@ private Predicate[] buildPredicates(CriteriaBuilder cb, Root<CrmContact> contact
...
+			String innerTokenPattern = "% " + escaped + "%";
+			tokenPredicates
+				.add(cb.or(cb.like(firstName, prefixPattern, '\\'), cb.like(firstName, innerTokenPattern, '\\'),
+						cb.like(lastName, prefixPattern, '\\'), cb.like(lastName, innerTokenPattern, '\\')));
+		}
+
+		return tokenPredicates.isEmpty() ? cb.conjunction() : cb.and(tokenPredicates.toArray(new Predicate[0]));
+	}
+
+	private List<Order> buildNameRelevanceOrders(CriteriaBuilder cb, Root<CrmContact> contact,
```
akilarootcode, line 128: check this impl again

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
```diff
@@ -64,15 +65,20 @@ public Page<CrmContact> findContacts(CrmContactMetricRequestDto filterDto, Pagea
...
+	private List<Order> buildOrderBy(CriteriaBuilder cb, Root<CrmContact> contact, CriteriaQuery<CrmContact> query,
+			String searchKeyword) {
 		Subquery<BigDecimal> dealValueSub = query.subquery(BigDecimal.class);
 		Root<CrmDeal> deal = dealValueSub.from(CrmDeal.class);
 		dealValueSub.select(cb.coalesce(cb.sum(deal.get(CrmDeal_.amount).cast(BigDecimal.class)), BigDecimal.ZERO))
 			.where(cb.equal(deal.get(CrmDeal_.contact), contact),
 					cb.equal(deal.get(CrmDeal_.stage).get(CrmDealStage_.stageType), CrmDealStageType.WON),
 					cb.isFalse(deal.get(CrmDeal_.isDeleted)));
 
-		return List.of(cb.desc(dealValueSub), cb.asc(contact.get(CrmContact_.id)));
```
akilarootcode, line 75: Ak -> data
Ak Sl - >

akilarootcode (changes requested):

PR #1722, opened 2026-08-07
fix/ drawer navigation visible for tablets

umeshaliyanage:
### Summary

Drawer navigation is not visible on tablet screen sizes.

### How to test

1.go to inspect and change screen sizes

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/frontend/src/community/common/components/organisms/AppBar/AppBar.tsx
+++ b/frontend/src/community/common/components/organisms/AppBar/AppBar.tsx
@@ -44,9 +44,8 @@ const AppBar = () => {
   const { user } = useAuth();
   const userInfoRef = useRef<HTMLDivElement | null>(null);
 
-  const { notifyData, setNotifyData, breadcrumbs } = useCommonStore(
-    (state) => state
-  );
+  const { notifyData, setNotifyData, breadcrumbs, isDrawerExpanded } =
+    useCommonStore((state) => state);
 
   const { data: employee } = useGetUserPersonalDetails();
 
@@ -175,7 +174,7 @@ const AppBar = () => {
               />
             )}
 
-            {isBelow1024 && (
+            {isBelow1024 && !isDrawerExpanded && (
               <button
                 className="flex cursor-pointer items-center justify-center w-10 h-10"
                 onClick={handleDrawer}
--- a/frontend/src/community/common/components/organisms/AppBar/AppBar.tsx
+++ b/frontend/src/community/common/components/organisms/AppBar/AppBar.tsx
@@ -37,15 +37,15 @@ const AppBar = () => {
 
   const queryMatches = useMediaQuery();
   const isBelow600 = queryMatches(MediaQueries.BELOW_600);
-  const isBelow1024 = queryMatches(MediaQueries.BELOW_1024);
-  const { handleDrawer } = useDrawer();
+  const { handleDrawer, isDrawerExpanded, isBelow1024 } = useDrawer();
   const translateAria = useTranslator("commonAria", "components", "appBar");
 
   const { user } = useAuth();
   const userInfoRef = useRef<HTMLDivElement | null>(null);
 
-  const { notifyData, setNotifyData, breadcrumbs, isDrawerExpanded } =
-    useCommonStore((state) => state);
+  const { notifyData, setNotifyData, breadcrumbs } = useCommonStore(
+    (state) => state
+  );
 
   const { data: employee } = useGetUserPersonalDetails();
 
--- a/frontend/src/community/common/components/organisms/AppBar/AppBar.tsx
+++ b/frontend/src/community/common/components/organisms/AppBar/AppBar.tsx
@@ -37,6 +37,7 @@ const AppBar = () => {
 
   const queryMatches = useMediaQuery();
   const isBelow600 = queryMatches(MediaQueries.BELOW_600);
+  const isBelow1024 = queryMatches(MediaQueries.BELOW_1024);
   const { handleDrawer } = useDrawer();
   const translateAria = useTranslator("commonAria", "components", "appBar");
 
@@ -174,7 +175,7 @@ const AppBar = () => {
               />
             )}
 
-            {isBelow600 && (
+            {isBelow1024 && (
               <button
                 className="flex cursor-pointer items-center justify-center w-10 h-10"
                 onClick={handleDrawer}
--- a/frontend/src/community/common/components/organisms/Drawer/styles.ts
+++ b/frontend/src/community/common/components/organisms/Drawer/styles.ts
@@ -10,20 +10,23 @@ export const DRAWER_ANIMATION_DURATION = "0.3s";
 export const DRAWER_TRANSFORM_DURATION = "0.05s";
 
 export const styles = ({ theme }: stylesProps) => ({
-  iconBtn: (isDrawerExpanded: boolean) => ({
-display: { xs: isDrawerExpanded ? "flex" : "none", lg: "flex" },
-    position: "absolute",
-    top: { xs: "2.5rem", lg: "2.65rem" },
-    right: { xs: "2.25rem", lg: isDrawerExpanded ? "1.5rem" : "1.05rem" },
-    height: "2.5rem",
-    width: "2.5rem",
-    zIndex: ZIndexEnums.MODAL,
-    visibility: {
-      xs: "visible",
-      lg: isDrawerExpanded ? "hidden" : "visible"
-    },
-    transition: `right ${DRAWER_ANIMATION_DURATION} ease, opacity ${DRAWER_ANIMATION_DURATION} ease, visibility ${DRAWER_ANIMATION_DURATION} ease, transform ${DRAWER_TRANSFORM_DURATION} ease`
-  }),
+  iconBtn: (isDrawerExpanded: boolean, isBelow1024: boolean) => {
+    const compactDisplay = isDrawerExpanded ? "flex" : "none";
+    const desktopRight = isDrawerExpanded ? "1.5rem" : "1.05rem";
+    const desktopVisibility = isDrawerExpanded ? "hidden" : "visible";
+
+    return {
+      display: isBelow1024 ? compactDisplay : "flex",
+      position: "absolute",
+      top: isBelow1024 ? "2.5rem" : "2.65rem",
+      right: isBelow1024 ? "2.25rem" : desktopRight,
+      height: "2.5rem",
+      width: "2.5rem",
+      zIndex: ZIndexEnums.MODAL,
+      visibility: isBelow1024 ? "visible" : desktopVisibility,
+      transition: `right ${DRAWER_ANIMATION_DURATION} ease, opacity ${DRAWER_ANIMATION_DURATION} ease, visibility ${DRAWER_ANIMATION_DURATION} ease, transform ${DRAWER_TRANSFORM_DURATION} ease`
+    };
+  },
   iconToggleBox: {
     position: "relative",
     display: "flex",
--- a/frontend/src/community/common/components/organisms/Drawer/styles.ts
+++ b/frontend/src/community/common/components/organisms/Drawer/styles.ts
@@ -11,7 +11,7 @@ export const DRAWER_TRANSFORM_DURATION = "0.05s";
 
 export const styles = ({ theme }: stylesProps) => ({
   iconBtn: (isDrawerExpanded: boolean) => ({
-    display: { xs: "flex", sm: "none", lg: "flex" },
+    display: "flex",
     position: "absolute",
     top: { xs: "2.5rem", lg: "2.65rem" },
     right: { xs: "2.25rem", lg: isDrawerExpanded ? "1.5rem" : "1.05rem" },
--- a/frontend/src/community/common/components/organisms/Drawer/styles.ts
+++ b/frontend/src/community/common/components/organisms/Drawer/styles.ts
@@ -11,7 +11,7 @@ export const DRAWER_TRANSFORM_DURATION = "0.05s";
 
 export const styles = ({ theme }: stylesProps) => ({
   iconBtn: (isDrawerExpanded: boolean) => ({
-    display: "flex",
+display: { xs: isDrawerExpanded ? "flex" : "none", lg: "flex" },
     position: "absolute",
     top: { xs: "2.5rem", lg: "2.65rem" },
     right: { xs: "2.25rem", lg: isDrawerExpanded ? "1.5rem" : "1.05rem" },
--- a/frontend/src/community/common/components/organisms/Drawer/styles.ts
+++ b/frontend/src/community/common/components/organisms/Drawer/styles.ts
@@ -11,19 +11,19 @@ export const DRAWER_TRANSFORM_DURATION = "0.05s";
 
 export const styles = ({ theme }: stylesProps) => ({
   iconBtn: (isDrawerExpanded: boolean, isBelow1024: boolean) => {
-    const compactDisplay = isDrawerExpanded ? "flex" : "none";
-    const desktopRight = isDrawerExpanded ? "1.5rem" : "1.05rem";
-    const desktopVisibility = isDrawerExpanded ? "hidden" : "visible";
+    const drawerToggleDisplay = isDrawerExpanded ? "flex" : "none";
+    const drawerToggleRight = isDrawerExpanded ? "1.5rem" : "1.05rem";
+    const drawerToggleVisibility = isDrawerExpanded ? "hidden" : "visible";
 
     return {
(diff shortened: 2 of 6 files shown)


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR fixes a real gap: between 600px and 1023px the drawer was unreachable (the AppBar hamburger only rendered below 600px, and the drawer's own toggle had `display: none` at the `sm` breakpoint), so switching the hamburger to `BELOW_1024` and making the toggle always `display: flex` is the right direction. However, the two changes use two different breakpoint sources — a hardcoded `max-width: 1024px` media query in AppBar vs. the theme's `lg` breakpoint (also 1024) in the drawer styles — which collide at exactly 1024px, and forcing `display: flex` unconditionally leaves an off-screen but keyboard-focusable toggle button inside the 0-width collapsed drawer on tablets. Both are worth tightening before merge; the rest are smaller cleanups around now-dead style rules and missing test coverage.

Found **12** new issue(s): 🔴 1 important, 🟡 5 suggestion(s), 🟣 6 nit(s)

frontend/src/community/common/components/organisms/Drawer/styles.ts
```diff
@@ -11,7 +11,7 @@ export const DRAWER_TRANSFORM_DURATION = "0.05s";
 
 export const styles = ({ theme }: stylesProps) => ({
   iconBtn: (isDrawerExpanded: boolean) => ({
-    display: { xs: "flex", sm: "none", lg: "flex" },
+    display: "flex",
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Making `display` unconditionally `"flex"` re-introduces the toggle IconButton into the DOM/tab order at the `sm`–`md` range (600–1023px), where it is not actually visible. In `StyledDrawer.tsx` the paper is `width: open ? "17.75rem" : "0rem"` for `breakpoints.up("sm")` with `overflowX: "hidden"` (and `overflowY: "visible"` computes to `auto` once the other axis is `hidden`, so the paper clips on both axes). The button is `position: absolute; right: 2.25rem; width: 2.5rem`, so inside a 0-width paper it is laid out entirely outside the left edge and clipped away — yet it is still a rendered, focusable `IconButton`. A keyboard user on a tablet with the drawer collapsed will tab onto an invisible control announced as "expand"/"collapse" (WCAG 2.4.7 focus visibility). The previous `sm: "none"` removed it from the tab order in that range. Suggest keying the value off the drawer state rather than hardcoding it, e.g. `display: { xs: isDrawerExpanded ? "flex" : "none", lg: "flex" }` — that keeps the tablet fix (a close button is available once the drawer is open) without leaving a focusable, clipped button when collapsed.

frontend/src/community/common/components/organisms/AppBar/AppBar.tsx
```diff
@@ -37,6 +37,7 @@ const AppBar = () => {
 
   const queryMatches = useMediaQuery();
   const isBelow600 = queryMatches(MediaQueries.BELOW_600);
+  const isBelow1024 = queryMatches(MediaQueries.BELOW_1024);
```
ThinuwanW, line 40: 🤖 **Claude** · 🟡 **Suggestion**

Breakpoint boundary collision at exactly 1024px. `MediaQueries.BELOW_1024` is `(max-width: 1024px)` while the theme sets `lg: 1024` (`frontend/src/community/common/theme/theme.ts:551`), so `theme.breakpoints.up("lg")` is `(min-width: 1024px)`. At a viewport of exactly 1024px BOTH match: the AppBar hamburger renders, and simultaneously `StyledDrawer` applies the desktop docked rail (`4.6rem`) plus the `lg` `iconBtn` offsets — giving two toggle controls. Worse, `useDrawer(isBelow1024)` in `Drawer.tsx:78` fires its effect and force-collapses the drawer, so a desktop user at exactly 1024px can never keep the drawer expanded across a re-mount. Fix by making the query exclusive of the breakpoint, e.g. add a `BELOW_1024 = "(max-width: 1023.98px)"`-style value, or better, derive it from the theme with `useMediaQuery(theme.breakpoints.down("lg"))` so the AppBar and the drawer styles share one source of truth instead of duplicating the 1024 magic number in two places.

frontend/src/community/common/components/organisms/Drawer/styles.ts
```diff
@@ -11,7 +11,7 @@ export const DRAWER_TRANSFORM_DURATION = "0.05s";
 
 export const styles = ({ theme }: stylesProps) => ({
   iconBtn: (isDrawerExpanded: boolean) => ({
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Now that `display` is a constant, the remaining responsive logic in this style object is dead and misleading. The `visibility: { xs: "visible", lg: isDrawerExpanded ? "hidden" : "visible" }` block is unconditionally overridden by the consumer — `Drawer.tsx:482` spreads it and then sets `visibility: "visible"` — so the "hide the toggle when expanded on desktop" behaviour never applies. Correspondingly, the `"&:hover .MuiIconButton-root": { opacity: 1, visibility: "visible" }` reveal rule in `StyledDrawer.tsx` is inert (nothing ever sets `opacity: 0` or `visibility: hidden`). Since this PR is already touching this rule, either delete the dead `visibility` key here plus the hover rule in `StyledDrawer.tsx`, or drop the `visibility: "visible"` override in `Drawer.tsx` if the hover-to-reveal behaviour is actually still wanted. Leaving all three in place makes the next person reason about behaviour that cannot happen.

frontend/src/community/common/components/organisms/AppBar/AppBar.tsx
ThinuwanW, line 40: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Tablet drawer behaviour is now user-reachable for the first time, so it is worth confirming the resulting layout is intended. `StyledDrawer` is `variant="permanent"` and its paper is `width: open ? "17.75rem" : "0rem"` from `sm` up, so on a 600–1023px tablet opening the drawer pushes/squeezes the main content by ~284px rather than overlaying it. Below 600px the paper is `100%`, i.e. a full-screen takeover, so tablets get a third behaviour that nothing was previously exercising. Also note that with `variant="permanent"` the `onClose={handleDrawer}` and `hideBackdrop={false}` props on `StyledDrawer` are no-ops — there is no backdrop and tapping outside will not dismiss the drawer, so the only way to close it on a tablet is the hamburger or the in-drawer toggle. If an overlay is the intent for tablets, consider switching to `variant="temporary"` (or extending the `xs` full-width paper rule up to `md`) below the `lg` breakpoint; otherwise this is fine, but please verify the squeezed content layout at 600–768px.

ThinuwanW, line 40: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

No test coverage accompanies this change — there are no `AppBar.test.tsx` or `Drawer.test.tsx` files in either component directory, so the responsive contract this PR establishes (hamburger visible below 1024px, hidden at/above it; drawer toggle reachable when the drawer is open on tablet) is unverified and easy to regress on the next breakpoint tweak. Consider adding a test that mocks `useMediaQuery` at representative widths and asserts on the hamburger via `screen.getByRole("button", { name: ... })`, wrapped in the project's `MockTheme` provider so the custom breakpoint values (`lg: 1024`) are in play.

ThinuwanW, line 40: 🤖 **Claude** · 🟡 **Suggestion**

`isBelow1024` is now derived independently in two components that are always mounted together (ContentWithDrawer.tsx:35-37 renders `<Drawer />` and `<AppBar />` as siblings), and the two callers pass different arguments to the same hook: AppBar calls `useDrawer()` with no argument while Drawer.tsx:78 calls `useDrawer(isBelow1024)`.

That asymmetry is load-bearing but undocumented. `useDrawer`'s optional `isBelow1024` parameter drives a `useEffect` that force-collapses the drawer (useDrawer.ts:13-17), so the parameter effectively means "this caller owns the auto-collapse". If someone later 'tidies up' AppBar by passing the `isBelow1024` it now conveniently has in scope, two mounted effects will both call `setIsDrawerExpanded(false)` against the same global store — harmless today only because the value is idempotent, but it makes the collapse behaviour depend on which component happens to pass the flag.

Since the hook is now the single owner of this breakpoint's behaviour, move the media query inside it: have `useDrawer()` call `queryMatches(MediaQueries.BELOW_1024)` itself, drop the optional parameter, and return `{ handleDrawer, isDrawerExpanded, isBelow1024 }`. Both AppBar and Drawer then call `useDrawer()` identically, the query is evaluated in one place, and the `isBelow1024` needed for finding #1's `aria-expanded` comes along for free.

ThinuwanW, line 40: 🤖 **Claude** · 🟡 **Suggestion**

This widens an existing SSR/hydration flash from phones to every tablet. The `useMediaQuery` wrapper (hooks/useMediaQuery.ts:10-20) forwards straight to MUI's `useMediaQuery` with no options, so it has no `ssrMatchMedia` and no `noSsr` — it returns `false` during SSR and on the first client render, then flips on the post-hydration effect.

Two visible consequences on the 600–1023px range this PR newly targets: (a) the hamburger is absent in the server HTML and pops in after hydration, and (b) `isDrawerExpanded` defaults to `true` (stores/slices/templateSlice.ts:7), so the drawer paints expanded at 17.75rem, pushing the whole content column right, and then snaps to `0rem` once Drawer's `useDrawer(isBelow1024)` effect fires. On a phone this was a narrow-window annoyance; on tablets it is a full-width layout jump on every page load.

Fix: give the wrapper an options passthrough and use `{ noSsr: true }` for these layout-critical queries so MUI skips the SSR pass and evaluates on the first client render, or supply `ssrMatchMedia` at the theme level. If you prefer to avoid touching the shared hook, gate the drawer's initial render on a mounted flag so the collapsed state is the first thing painted below `lg`.

frontend/src/community/common/components/organisms/AppBar/AppBar.tsx
```diff
@@ -174,7 +175,7 @@ const AppBar = () => {
               />
             )}
 
-            {isBelow600 && (
+            {isBelow1024 && (
               <button
```
ThinuwanW, line 179: 🤖 **Claude** · 🟣 **Nit**

This native `<button>` has no `type` attribute, so it defaults to `type="submit"`. It's benign while AppBar is never rendered inside a `<form>`, but AppBar is a global chrome component rendered on every authenticated page via ContentWithDrawer, and it now renders on the large majority of viewports rather than just phones — so the blast radius of a future page that wraps content in a form has grown. Add `type="button"`. (Same applies to the sibling `<div role="button">` elements above, which would be better as real buttons, but that's pre-existing and out of scope here.)

frontend/src/community/common/components/organisms/AppBar/AppBar.tsx
ThinuwanW, line 40: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

No `data-testid` on the hamburger, which blocks the test coverage the first pass asked for. `appBarTestId.appBar` (constants/testIds.ts:24-29) only exposes `profileAvatar` and `viewAccountBtn`, while the Drawer's equivalent toggle has `appDrawerTestId.buttons.drawerToggleBtn`. Add e.g. `menuToggleBtn: "app-bar-menu-toggle-btn"` to `appBarTestId.appBar` and put it on this button. (Per the review guidelines, prefer `screen.getByRole("button", { name: ... })` in the actual test once the aria-label from finding #1 is state-aware — but the test id keeps this consistent with how the sibling drawer toggle is already targeted.)

ThinuwanW, line 40: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

Worth checking the header layout at the low end of the newly-covered range. The breadcrumb is still gated on `isBelow600` (line 99), so between 601px and 1023px the PageHeader now renders breadcrumb + ClockWidget + notification bell + avatar + the new 40px hamburger, with `gap-4` between them, inside a right slot that is already `w-full`. That is one more fixed-width item than this range has ever had to fit. On a 601px viewport (or a split-screen/foldable) this is the most likely place for the header to wrap or squash the breadcrumb.

I can't verify how `PageHeader` from `@rootcodelabs/skapp-ui` distributes its left/right slots (node_modules isn't installed here), so flagging as a nit rather than asserting a break. If it does overflow, the natural fix is to gate the breadcrumb on `!isBelow1024` instead of `!isBelow600` so the two responsive decisions in this component share one boundary.

frontend/src/community/common/components/organisms/Drawer/styles.ts
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

While you're removing stale responsive logic from this file, two style keys here are entirely unused: `applyLeaveBtn` (line 190) and `link` (line 194). Grepping `classes.applyLeaveBtn` / `classes.link` across `frontend/src` returns no consumers — the Drawer footer now renders a `ButtonV2` with the `getHelpLink` test id (Drawer.tsx:469-476), so the apply-leave affordance these styled is gone. Note `applyLeaveBtn` carries the exact `display: { xs: "none", sm: "flex" }` pattern this PR is fixing on line 14, so leaving it behind invites someone to copy it. Delete both keys.

frontend/src/community/common/components/organisms/Drawer/styles.ts
umeshaliyanage: did not remove out of scope

frontend/src/community/common/components/organisms/AppBar/AppBar.tsx
umeshaliyanage: out of scope of the bug

umeshaliyanage: out of scope for the bug

umeshaliyanage: Out of scope for the bug fix

umeshaliyanage: out of scope for the bug

umeshaliyanage: changed the drawer logic so that we are using the breakpoint values by re using isBelow600: boolean;
  isBelow1024: boolean; and passing them as props in Drawer.tsx

frontend/src/community/common/components/organisms/AppBar/AppBar.tsx
umeshaliyanage: out of scope here

frontend/src/community/common/components/organisms/Drawer/styles.ts
umeshaliyanage: changed and moved into  const compactDisplay = isDrawerExpanded ? "flex" : "none";

frontend/src/community/common/components/organisms/AppBar/AppBar.tsx
umeshaliyanage: out of scope for the bug i was assigned to

umeshaliyanage: out of scope

frontend/src/community/common/components/organisms/AppBar/AppBar.tsx
umeshaliyanage: out of scope

frontend/src/community/common/components/organisms/Drawer/styles.ts
umeshaliyanage: removed mui breakpoints

frontend/src/community/common/components/organisms/AppBar/AppBar.tsx
umeshaliyanage: out of scope and already is following what is implemented thru skapp

umeshaliyanage: moved it to useDrawer

shakila-kamalasena (dismissed):

shakila-kamalasena (approved):

akilarootcode (approved):

PR #1723, opened 2026-08-07
fix: end previous position one day before new position's start date

SanithuGunasena07:
Bug: When adding a new position for an employee without manually ending the previous one, and marking the new position as current, the system auto-ends the previous position but sets its end date to the same date as the new position's start date, causing a one-day overlap between the two positions.

Root Cause: Frontend logic in CareerProgressDetailsSection.tsx (Employment tab) and CareerProgressionDetailsSection.tsx (Add New Resource wizard) assigns the new position's start-date timestamp directly to the previous position's endDate with no -1 day adjustment.

Fix: Previous position's endDate now computed as newStartDate - 1 day (via Luxon .minus({ days: 1 })) instead of newStartDate directly, in both CareerProgressDetailsSection.tsx and CareerProgressionDetailsSection.tsx.

### DIFF
--- a/frontend/src/community/people/components/organisms/AddNewResourceFlow/EmploymentDetailsSection/CareerProgressionDetailsSection.tsx
+++ b/frontend/src/community/people/components/organisms/AddNewResourceFlow/EmploymentDetailsSection/CareerProgressionDetailsSection.tsx
@@ -28,7 +28,8 @@ import {
 } from "~community/common/utils/commonUtil";
 import {
   convertDateToFormat,
-  getDateFromTimeStamp
+  getDateFromTimeStamp,
+  getPreviousDayFormatted
 } from "~community/common/utils/dateTimeUtils";
 import { useGetPreprocessedRoles } from "~community/people/api/PeopleApi";
 import { JobFamilyActionModalEnums } from "~community/people/enums/JobFamilyEnums";
@@ -343,10 +344,7 @@ const CareerProgressionDetailsSection = ({
           positions[index] = {
             ...position,
             currentPosition: false,
-            endDate: convertDateToFormat(
-              DateTime.fromMillis(newStartDate).minus({ days: 1 }).toJSDate(),
-              LONG_DATE_TIME_FORMAT
-            )
+            endDate: getPreviousDayFormatted(startDate)
           };
         }
       });
--- a/frontend/src/community/people/components/organisms/AddNewResourceFlow/EmploymentDetailsSection/CareerProgressionDetailsSection.tsx
+++ b/frontend/src/community/people/components/organisms/AddNewResourceFlow/EmploymentDetailsSection/CareerProgressionDetailsSection.tsx
@@ -344,7 +344,7 @@ const CareerProgressionDetailsSection = ({
             ...position,
             currentPosition: false,
             endDate: convertDateToFormat(
-              new Date(newStartDate),
+              DateTime.fromMillis(newStartDate).minus({ days: 1 }).toJSDate(),
               LONG_DATE_TIME_FORMAT
             )
           };
--- a/frontend/src/community/people/components/organisms/EmploymentFormSection/SubSections/CareerProgressDetailsSection.tsx
+++ b/frontend/src/community/people/components/organisms/EmploymentFormSection/SubSections/CareerProgressDetailsSection.tsx
@@ -12,7 +12,8 @@ import { useTranslator } from "~community/common/hooks/useTranslator";
 import { IconName } from "~community/common/types/IconTypes";
 import {
   convertDateToFormat,
-  getDateFromTimeStamp
+  getDateFromTimeStamp,
+  getPreviousDayFormatted
 } from "~community/common/utils/dateTimeUtils";
 import PeopleFormTable from "~community/people/components/molecules/PeopleFormTable/PeopleFormTable";
 import { JobFamilyActionModalEnums } from "~community/people/enums/JobFamilyEnums";
@@ -120,10 +121,7 @@ const CareerProgressDetailsSection = ({
           positions[index] = {
             ...position,
             isCurrentEmployment: false,
-            endDate: convertDateToFormat(
-              DateTime.fromMillis(newStartDate).minus({ days: 1 }).toJSDate(),
-              LONG_DATE_TIME_FORMAT
-            )
+            endDate: getPreviousDayFormatted(startDate ?? "")
           };
         }
       });
--- a/frontend/src/community/people/components/organisms/EmploymentFormSection/SubSections/CareerProgressDetailsSection.tsx
+++ b/frontend/src/community/people/components/organisms/EmploymentFormSection/SubSections/CareerProgressDetailsSection.tsx
@@ -121,7 +121,7 @@ const CareerProgressDetailsSection = ({
           positions[index] = {
             ...position,
             isCurrentEmployment: false,
-            endDate: getPreviousDayFormatted(startDate ?? "")
+            endDate: getPreviousDayFormatted(startDate)
           };
         }
       });
--- a/frontend/src/community/people/components/organisms/EmploymentFormSection/SubSections/CareerProgressDetailsSection.tsx
+++ b/frontend/src/community/people/components/organisms/EmploymentFormSection/SubSections/CareerProgressDetailsSection.tsx
@@ -121,7 +121,7 @@ const CareerProgressDetailsSection = ({
             ...position,
             isCurrentEmployment: false,
             endDate: convertDateToFormat(
-              new Date(newStartDate),
+              DateTime.fromMillis(newStartDate).minus({ days: 1 }).toJSDate(),
               LONG_DATE_TIME_FORMAT
             )
           };
--- a/frontend/src/community/common/utils/dateTimeUtils.ts
+++ b/frontend/src/community/common/utils/dateTimeUtils.ts
@@ -28,6 +28,12 @@ export const getDateFromTimeStamp = (timestamp: string): string => {
   return timestamp.split("T")[0];
 };
 
+export const getPreviousDayFormatted = (timestamp: string): string => {
+  return DateTime.fromISO(getDateFromTimeStamp(timestamp))
+    .minus({ days: 1 })
+    .toFormat(LONG_DATE_TIME_FORMAT);
+};
+
 export const getOrdinalIndicator = (day: number) => {
   if (day === 1 || day === 21 || day === 31) {
     return "st";


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR fixes the career-progression overlap by setting a superseded current position's end date to the day before the new position's start date, in both the AddNewResource flow and the EmploymentFormSection flow. The intent is correct, but the arithmetic is done on a UTC-midnight timestamp reinterpreted in the browser's local zone, so the result is off by an extra day for any user west of UTC. Both call sites are otherwise consistent with each other, no other copies of this logic exist in the codebase, and no other behavior changed.

Found **10** new issue(s): 🔴 3 important, 🟡 5 suggestion(s), 🟣 2 nit(s)

frontend/src/community/people/components/organisms/AddNewResourceFlow/EmploymentDetailsSection/CareerProgressionDetailsSection.tsx
```diff
@@ -344,7 +344,7 @@ const CareerProgressionDetailsSection = ({
             ...position,
             currentPosition: false,
             endDate: convertDateToFormat(
-              new Date(newStartDate),
+              DateTime.fromMillis(newStartDate).minus({ days: 1 }).toJSDate(),
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Timezone-dependent off-by-one: the day subtraction is applied to a UTC instant but rendered in local time. `newStartDate` is built as `new Date(getDateFromTimeStamp(startDate)).getTime()` — `getDateFromTimeStamp` returns a date-only string ("2024-05-10"), and per the ECMAScript spec a date-only ISO string is parsed as **UTC midnight**. `DateTime.fromMillis(...)` then reinterprets that instant in the browser's local zone, and `convertDateToFormat` (`DateTime.fromJSDate(date).toFormat(...)`) formats it in local time too.

Result for start date 2024-05-10:
- UTC / UTC+X (e.g. Asia/Colombo): local is 2024-05-10T05:30 → minus 1 day → endDate "2024-05-09T05:30:00.000" ✓ correct day (but a stray non-midnight time component).
- UTC-X (all of the Americas, e.g. America/New_York): local is 2024-05-09T19:00 → minus 1 day → endDate "2024-05-08T19:00:00.000" ✗ **two** days before the new start date.

This is a user-visible regression for negative-offset users: before this PR the same expression produced 2024-05-09 there (accidentally correct). It can also make the save fail outright — `checkOverlap` only rejects `newStartDate <= oldStartDate`, so an old current position starting 2024-05-09 followed by a new one starting 2024-05-10 is allowed, and the computed endDate 2024-05-08 is before that position's own startDate. `EmployeeValidationServiceImpl.validateEndDateRules` (backend/src/main/java/com/skapp/community/peopleplanner/service/impl/EmployeeValidationServiceImpl.java:404) throws PEOPLE_ERROR_VALIDATION_CAREER_END_DATE_BEFORE_START_DATE in that case.

Fix: do the arithmetic on the calendar date rather than on a UTC instant, which also drops the stray time component:
`endDate: DateTime.fromISO(getDateFromTimeStamp(startDate)).minus({ days: 1 }).toFormat(LONG_DATE_TIME_FORMAT)`
(equivalently `DateTime.fromMillis(newStartDate, { zone: "utc" }).minus({ days: 1 }).toFormat(LONG_DATE_TIME_FORMAT)` — note `.toJSDate()` must be dropped, since routing through `convertDateToFormat` re-localizes and reintroduces the shift).

The identical expression at frontend/src/community/people/components/organisms/EmploymentFormSection/SubSections/CareerProgressDetailsSection.tsx:124 has the same defect and needs the same change (with `startDate ?? ""`).

frontend/src/community/people/components/organisms/EmploymentFormSection/SubSections/CareerProgressDetailsSection.tsx
```diff
@@ -121,7 +121,7 @@ const CareerProgressDetailsSection = ({
             ...position,
             isCurrentEmployment: false,
             endDate: convertDateToFormat(
-              new Date(newStartDate),
+              DateTime.fromMillis(newStartDate).minus({ days: 1 }).toJSDate(),
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

This whole "close the previous current position" block (the `if (isCurrentEmployment) { positions?.forEach(...) }` loop plus the endDate computation) is byte-for-byte duplicated in AddNewResourceFlow/EmploymentDetailsSection/CareerProgressionDetailsSection.tsx:340-353, modulo the `currentPosition` vs `isCurrentEmployment` field name. This PR had to patch the same one-liner twice, which is exactly the failure mode duplication causes. `checkOverlap` was already extracted into `useCareerProgressionFormHandler`; consider extracting a `getPreviousPositionEndDate(startDate: string): string` helper into `~community/people/utils/careerProgressUtils` (which already hosts `calculateTenure`) and having both call sites use it, so the timezone fix lives in one place.

ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

No test coverage for this change. There is no `.test.tsx`/`.test.ts` for either CareerProgress(ion)DetailsSection or for `useCareerProgressionFormHandler`, and `careerProgressUtils.ts` is untested as well. Because the correct behavior here is a pure function of a date string, it is cheap to lock down: if the endDate computation is extracted per the duplication comment, add a unit test asserting that start date "2024-05-10" yields end date "2024-05-09", and run it under at least one negative-offset TZ (e.g. `TZ=America/New_York`) so the off-by-one described above can never regress silently.

ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Separate from the day-offset problem: the value produced here carries a non-zero time-of-day, so the stored `endDate` is not in the same shape as every other date this form writes. `newStartDate` is a UTC-midnight epoch (`new Date("yyyy-MM-dd").getTime()`), `DateTime.fromMillis` re-interprets it in the local zone, and `convertDateToFormat` then formats local wall-clock with `LONG_DATE_TIME_FORMAT` (which emits no offset). In UTC+05:30 you get `...T05:30:00.000`, whereas every user-entered endDate goes through `convertDateToFormat(new Date(newValue), ...)` and comes out `...T00:00:00.000`. Consequences: (a) `calculateTenure` in careerProgressUtils.ts diffs against a shifted instant; (b) re-opening that row via `handleEdit` feeds `DateTime.fromISO` a 05:30 value into the date picker; (c) `isFieldDifferentAndValid` in peopleEditFlowUtils/employmentDetailsChangesUtils.ts compares this against the server's `LocalDate` string (`yyyy-MM-dd`) and reports a spurious diff; (d) the backend `EmployeeEmploymentCareerProgressionDetailsDto.endDate` is a `LocalDate`, so the time part is silently truncated and the round-trip is lossy. Build the value from the ISO string and normalise instead of round-tripping through a millis timestamp and a JS Date: derive it as `DateTime.fromISO(startDate).minus({ days: 1 }).startOf("day").toFormat(LONG_DATE_TIME_FORMAT)` and drop the `convertDateToFormat`/`toJSDate` hop entirely. Apply the identical change in the AddNewResourceFlow copy.

frontend/src/community/people/components/organisms/AddNewResourceFlow/EmploymentDetailsSection/CareerProgressionDetailsSection.tsx
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Behavioural side effect not covered by the fix itself: the tenure shown in the table for the auto-closed position drops by a month at exact boundaries. `calculateTenure` (inlined here at lines 160-181, and in careerProgressUtils.ts for the other flow) does `end.diff(start, ["years","months"])` and `Math.floor`s the months, i.e. it treats endDate as exclusive. Previously endDate equalled the next position's start, so a position running 2024-01-15 to 2024-04-15 rendered "3m"; with the new endDate of 2024-04-14 the diff is 2 months + 30 days and it now renders "2m". The same off-by-one hits the year rollover ("1y 0m" becomes "0y 11m" -> "11m"). Since endDate is now genuinely the last day worked, tenure should be inclusive: compute it as `DateTime.fromISO(endDate).plus({ days: 1 }).diff(start, ["years","months"])` for non-current positions (leave the `DateTime.now()` branch as is). Fix it in careerProgressUtils.ts and delete the inlined duplicate in this file so both flows agree.

frontend/src/community/people/components/organisms/EmploymentFormSection/SubSections/CareerProgressDetailsSection.tsx
```diff
@@ -121,7 +121,7 @@ const CareerProgressDetailsSection = ({
             ...position,
             isCurrentEmployment: false,
```
ThinuwanW, line 123: 🤖 **Claude** · 🟡 **Suggestion**

This loop does not skip the row currently being edited, unlike `checkOverlap` in useCareerProgressionFormHandler.tsx which starts with `if (index === rowEdited) return;`. When the user edits the existing current position and moves its start date later while leaving "current employment" ticked, `position.isCurrentEmployment` is true and `startDate > position.startDate` holds for the edited row itself, so it is flipped to `isCurrentEmployment: false` with a derived endDate. Today that is silently undone by the `positions[rowEdited] = { ...positions[rowEdited], ...values }` spread a few lines below, purely because Formik's `values` happens to still carry the `endDate` key. That is an accidental dependency: if `values` ever stops carrying an explicit `endDate` (e.g. `handleEdit` skips the field, or the initial values shape changes), the payload ships `isCurrentEmployment: true` together with a non-null `endDate`, which the backend rejects outright in `EmployeeValidationServiceImpl.validateEndDateRules` (PEOPLE_ERROR_VALIDATION_CAREER_PROGRESSION_END_DATE_NOT_REQUIRED_FOR_CURRENT_EMPLOYMENT_TRUE). Add `if (index === rowEdited) return;` as the first statement in the forEach in both files so the two loops agree with `checkOverlap`.

frontend/src/community/people/components/organisms/AddNewResourceFlow/EmploymentDetailsSection/CareerProgressionDetailsSection.tsx
```diff
@@ -344,7 +344,7 @@ const CareerProgressionDetailsSection = ({
             ...position,
             currentPosition: false,
             endDate: convertDateToFormat(
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The derived endDate is written straight into the store for a row the user is not editing, and it never passes through Yup (`employeeCareerDetailsValidation` only validates the row in the form). There is no guard that the computed value is not before that position's own `startDate`. The backend does enforce it - `EmployeeValidationServiceImpl.validateEndDateRules` throws PEOPLE_ERROR_VALIDATION_CAREER_END_DATE_BEFORE_START_DATE - so the failure mode is a rejected save of the entire employee record, surfaced as a generic error with no field-level indication, and pointing at a row the user never touched. `checkOverlap` only guarantees `newStartDate > oldStartDate` on UTC-midnight timestamps, which leaves no margin once the local-zone shift is applied. Clamp defensively before writing: compute the candidate, and if it is earlier than `position.startDate`, either fall back to `position.startDate` or abort the submit with `setFieldError("startDate", translateText(["overlapError"]))` rather than persisting an invalid row.

frontend/src/community/people/components/organisms/AddNewResourceFlow/EmploymentDetailsSection/CareerProgressionDetailsSection.tsx
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

The two components are hand-duplicated and have already drifted apart, which is why this one-line fix had to be applied twice. The EmploymentFormSection copy guards the comparison (`startDate && position?.startDate && startDate > position?.startDate`) while this one does not (`startDate > position.startDate` at line 342); if a `startDate` is ever missing on a legacy or bulk-uploaded row, the comparison quietly yields false, the previous current position is never closed, and the save fails on the backend's `ONLY_ONE_CURRENT_EMPLOYMENT_ALLOWED` check. Beyond that, this file re-implements `calculateTenure` (lines 160-181), `checkOverlap` (273-310), `formatData`, `handleEdit` and `handleDelete` that were already extracted into `~community/people/hooks/useCareerProgressionFormHandler.tsx` and `~community/people/utils/careerProgressUtils.ts`. Extract the date arithmetic into a single named helper in careerProgressUtils.ts (e.g. `getPrecedingPositionEndDate(startDate: string): string`), call it from both onSubmit handlers, and align the null guards. That also gives the logic a plain-function seam that can be unit-tested without rendering either component.

ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

Pre-existing in this touched file (line 8), and mirrored in useCareerProgressionFormHandler.tsx line 2: `import { rejects } from "assert"` pulls Node's core `assert` module into a client bundle, and its only use, `setFieldValue(...).catch(rejects)` at line 420, is not a valid rejection handler - `assert.rejects` expects a promise-returning function and will itself reject when handed an Error, producing an unhandled rejection instead of handling one. Drop the import and either handle the error properly or use `void setFieldValue(...)` to match the pattern already used in `handleEdit` (lines 253-258).

ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

Pre-existing in this touched file (lines 418-423) and duplicated in useCareerProgressionFormHandler.tsx (lines 266-271): `latestRoleLabel` is declared as `useState<number>()`, so its initial value is `undefined`, never `null`. The guard `if (latestRoleLabel !== null)` is therefore always true, so on first mount the effect sets `jobFamily` to `undefined` and fires an unnecessary `getPreprocessedRoles.refetch()` on top of the query's own initial fetch. Change the check to `if (latestRoleLabel !== undefined)` (or `!= null`), and add `getPreprocessedRoles` / `setFieldValue` handling to the dependency array or wrap the refetch so the effect does not rely on an incomplete dep list.

frontend/src/community/people/components/organisms/EmploymentFormSection/SubSections/CareerProgressDetailsSection.tsx
SanithuGunasena07: out of scope for PR

SanithuGunasena07: not needed

frontend/src/community/people/components/organisms/AddNewResourceFlow/EmploymentDetailsSection/CareerProgressionDetailsSection.tsx
SanithuGunasena07: now doing the day subtraction on the calendar date directly (DateTime.fromISO(getDateFromTimeStamp(startDate)).minus({ days: 1 })) instead of going through a UTC millis instant

frontend/src/community/people/components/organisms/EmploymentFormSection/SubSections/CareerProgressDetailsSection.tsx
SanithuGunasena07: This is already resolved by the fix for the day-offset comment above switching to DateTime.fromISO(getDateFromTimeStamp(startDate)) means we're parsing a plain date string, which gives local midnight directly

frontend/src/community/people/components/organisms/AddNewResourceFlow/EmploymentDetailsSection/CareerProgressionDetailsSection.tsx
SanithuGunasena07: out of scope for PR

frontend/src/community/people/components/organisms/EmploymentFormSection/SubSections/CareerProgressDetailsSection.tsx
SanithuGunasena07: Out of scope for PR (Already existing issue not created by the change)

frontend/src/community/people/components/organisms/AddNewResourceFlow/EmploymentDetailsSection/CareerProgressionDetailsSection.tsx
SanithuGunasena07: With the timezone fix in place, checkOverlap guarantees newStartDate > oldStartDate, and since we now subtract exactly one calendar day, the computed endDate can never be before position.startDate (worst case it equals it). So this specific failure mode isn't reachable today.

frontend/src/community/people/components/organisms/AddNewResourceFlow/EmploymentDetailsSection/CareerProgressionDetailsSection.tsx
SanithuGunasena07: Out of scope for PR (Already existing issue not created by the change)

SanithuGunasena07: Out of scope for PR (Already existing issue not created by the change)

SanithuGunasena07: Out of scope for PR (Already existing issue not created by the change)

shakila-kamalasena (approved):

shakila-kamalasena (dismissed):

frontend/src/community/people/components/organisms/EmploymentFormSection/SubSections/CareerProgressDetailsSection.tsx
```diff
@@ -120,10 +121,7 @@ const CareerProgressDetailsSection = ({
           positions[index] = {
             ...position,
             isCurrentEmployment: false,
-            endDate: convertDateToFormat(
-              new Date(newStartDate),
-              LONG_DATE_TIME_FORMAT
-            )
+            endDate: getPreviousDayFormatted(startDate ?? "")
```
akilarootcode: remove  ?? ""

akilarootcode (changes requested):

frontend/src/community/people/components/organisms/EmploymentFormSection/SubSections/CareerProgressDetailsSection.tsx
SanithuGunasena07: removed fallback

akilarootcode (approved):

PR #1724, opened 2026-08-07
[Feat] Sign in on the central OAuth auth host

akilarootcode:
## Summary

Completes the central-auth-host work on the frontend. The MCP connector now sends users to `auth.skapp.dev` to sign in, and the sign-in page has to work there — on a host that is **not** a tenant.

The frontend derived the tenant from the first hostname label (`hostname.split(".")[0]`) and froze it at module load. On `auth.skapp.dev` that produced a tenant called `auth`: the page rendered "Sign in to auth" with no form, and every API call carried `X-Tenant-ID: auth`. This was the last blocker for the central issuer.

Three things change:

1. **The tenant is resolved per request**, not once at import.
2. **The sign-in page resolves the tenant for the OAuth journey** — silently from the shared refresh cookie when the user is already signed in on their tenant host, otherwise from the company domain they type.
3. **Sign-in completes in place** on the auth host, so the OAuth session lands on the same origin that issues the authorization code (bouncing to the tenant host would create the session on the wrong host and loop).

## Changes by repo

**`skapp` (super repo)**
- `src/community/common/utils/tenantUtil.ts` *(new)* — one home for tenant derivation: `getTenantId()` (selected tenant, else subdomain), `setSelectedTenantId()` / `clearSelectedTenantId()`, `getTenantIdFromCookie()`, `getBaseDomain()`, and `getSubDomain()` moved out of the interceptor.
- `src/community/common/utils/axiosInterceptor.ts`, `src/community/auth/utils/authInterceptor.ts` — call `getTenantId()` inside the request interceptor instead of reading a module-level constant.
- `src/community/common/constants/stringConstants.ts` — add `AUTH`.

**`skapp-ep-fe-src`**
- The four enterprise interceptors (`axiosInterceptor`, `guestAxiosInterceptor`, `pmAxiosInterceptor`, `eSignAxiosInterceptor`) resolve the tenant per request.
- `common/utils/commonUtil.ts` — `getTenantID()` delegates to the shared resolver.

**`skapp-ep-fe-pages`**
- `signin.tsx` — treats the auth host as a tenant-selection host; tries silent sign-in first; renders the credentials form in place once the tenant is known; every host-derived gate now keys off the resolved tenant.

## Flow

```mermaid
sequenceDiagram
    participant C as Claude
    participant M as mcp.skapp.dev
    participant A as auth.skapp.dev
    participant API as Core API

    C->>M: POST /mcp
    M-->>C: 401 + resource_metadata
    C->>M: GET /.well-known/oauth-protected-resource/mcp
    M-->>C: authorization_servers: [auth.skapp.dev]
    C->>A: /oauth2/authorize?resource=...
    A-->>C: 302 /signin?callback=...

    alt already signed in on the tenant host
        A->>API: POST /v1/auth/session/refresh-token (tenant from shared cookie)
        API-->>A: access token
    else cold
        A->>API: POST /v1/ep/organization/login-method (company domain)
        API-->>A: login method
        A->>API: POST /v1/auth/session/sign-in (X-Tenant-ID: resolved)
        API-->>A: access token
    end

    A->>API: POST /v1/auth/oauth/session-login (same origin, tenant pinned)
    A-->>C: ?code=...
    C->>API: POST /oauth2/token (no tenant header)
    API-->>C: access_token (aud=mcp resource, tenant claim)
    C->>M: tools/list + Bearer
    M-->>C: tools
```

## Why silent sign-in is safe

The `tenant` cookie is readable by scripts, so it is a *hint*, not proof. It only names which tenant-namespaced refresh cookie to try. The refresh token itself is an httpOnly cookie verified under a per-tenant derived HMAC key, so a forged `tenant` value simply fails signature verification — it cannot be used to reach another tenant.

## Exclusions

- No AWS or infrastructure changes here — `auth.skapp.dev` already resolves via the `*.skapp.dev` wildcard; it needs ALB listener rules to reach the core target group.
- No local hosts, URLs or `local` environment branches; the local mirror was built by pointing DNS/nginx at the dev host names.

## Ignored

- Submodule pointer/hash bumps — done at deployment, deliberately not committed.

## Testing

| Check | Result |
|---|---|
| `tsc --noEmit` (changed files) | clean — remaining errors are pre-existing and unrelated |
| `eslint` (changed files) | 0 errors |
| `prettier` | applied |
| `npm run build` | passes |
| `jest` | no new failures — failing-suite list identical with and without the change (pre-existing `@rootcodelabs/skapp-ui` resolution failures) |
| Playwright end-to-end | 2 passed |

The end-to-end run covers the whole journey against a local mirror of the dev topology (`auth` / `mcp` / `api` / tenant hosts behind nginx):

- MCP returns 401 with protected-resource metadata; discovery points at `auth.skapp.dev`
- the auth host asks for the company domain (previously: nothing to fill)
- credentials are entered **on `auth.skapp.dev`**, no bounce to the tenant host
- every tenant-scoped call carries the resolved tenant; `session-login` is pinned to it
- token exchange straight to core **without** a tenant header returns 200, `aud` bound to the MCP resource, `tenant` claim correct
- `initialize` + `tools/list` with that Bearer returns the Skapp tools
- an already-signed-in user reaches the authorization code **without re-entering credentials**

## Related PRs

- Core BE: rootcodelabs/skapp-ep-be-src#739
- AI BE: rootcodelabs/skapp-ai#80
- Migrations: rootcodelabs/skapp-migrations#200
- BE tests: rootcodelabs/skapp-ep-be-tests#33
- FE (super repo): SkappHQ/skapp#1724
- FE (enterprise src): rootcodelabs/skapp-ep-fe-src#773
- FE (enterprise pages): rootcodelabs/skapp-ep-fe-pages#348


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR centralizes tenant-ID resolution behind `getTenantId()` from `~enterprise/common/utils/tenantUtil` (with a community fallback stub) so that the `auth.<domain>` sign-in host can resolve the tenant from the `?tenantId=` query param or the `tenant` cookie instead of blindly using the subdomain. The change itself is correct and consistent with the enterprise submodule, where every other axios interceptor (enterprise `axiosInterceptor`, `pmAxiosInterceptor`, `guestAxiosInterceptor`, `eSignAxiosInterceptor`, `SignApi`) already uses `getTenantId()`. The main gap is that the old `tenantID` module-level export is left in place and still feeds one remaining `X-Tenant-ID` producer plus ~10 other tenant-derived call sites, leaving two competing sources of truth; the fallback stub is also not at export parity with the real module.

Found **12** new issue(s): 🔴 2 important, 🟡 4 suggestion(s), 🟣 1 nit(s)

frontend/src/community/common/utils/axiosInterceptor.ts
```diff
@@ -1,6 +1,7 @@
 import axios, { InternalAxiosRequestConfig } from "axios";
 
 import { getAccessToken } from "~community/auth/utils/authUtils";
+import { getTenantId } from "~enterprise/common/utils/tenantUtil";
```
ThinuwanW, line 4: 🤖 **Claude** · 💬 **Note**

📁 **File-level observation**

Two competing sources of truth for the tenant ID remain after this change. `tenantID` (module-level const, evaluated once at import time from `window.location.hostname`, still exported from this file) is consumed by ~12 call sites, and one of them still builds the tenant header itself: `src/enterprise/projects/api/projectsApi.ts:34` sets `"X-Tenant-ID": tenantID` on its own `axios.post` to the PM GraphQL API, bypassing the interceptors. That is now the only `X-Tenant-ID` producer in the codebase not using `getTenantId()` — enterprise `axiosInterceptor.ts:24`, `pmAxiosInterceptor.ts:15`, `guestAxiosInterceptor.ts:16`, `sign/api/eSignAxiosInterceptor.ts:16` and `sign/api/SignApi.ts:116` were all migrated. On the `auth.<domain>` host `tenantID` evaluates to the literal string `"auth"`, so that request would carry `X-Tenant-ID: auth`; it is also stale-by-construction because it never re-reads after `router.replace` updates the `tenantId` query param (which is precisely the case `getTenantId()` was added to handle). Fix: import `getTenantId` in `projectsApi.ts` and call it inside `fetchProjects`. Beyond that, please finish the migration and delete the `tenantID` export (or mark it `@deprecated`) — remaining consumers that derive a tenant/company-domain value from it are `BaseLayout.tsx:41` (`useGetGlobalLoginMethod`), `OrganizationSettings.tsx:43,90`, `ForgotPasswordChangeModal.tsx:55`, `useSessionData.ts:186`, `AddNewResourceModal.tsx:142`, `IntegrationsSettings.tsx:110`, `guest-otp.tsx:147,176`. Leaving both APIs exported invites new code to pick the wrong one. This would also let you drop the now-redundant local `getSubDomain` helper (lines 9-12), which duplicates `getSubdomain` in `~enterprise/common/utils/commonUtil.ts:44` minus its empty-input guard.

frontend/src/fallback/common/utils/tenantUtil.ts
```diff
@@ -0,0 +1 @@
+export const getTenantId = (): string => "";
```
ThinuwanW, line 1: 🤖 **Claude** · 💬 **Note**

The fallback stub is not at export parity with the enterprise module it shadows. `src/enterprise/common/utils/tenantUtil.ts` exports three symbols — `getTenantId`, `getTenantIdFromCookie`, and `TENANT_QUERY_PARAM` — but the stub only provides `getTenantId`. This is riskier than usual here because `next.config.mjs` sets both `typescript.ignoreBuildErrors: true` and `eslint.ignoreDuringBuilds: true`, so a future import of `TENANT_QUERY_PARAM`/`getTenantIdFromCookie` from shared community code would not fail the community build — webpack would emit only a warning and the binding would be `undefined` at runtime (e.g. `new URLSearchParams(...).get(undefined)`). Note also that `jest.config.js:10` maps `^~enterprise(.*)$` to `src/fallback$1` unconditionally, so every test resolves this stub regardless of mode. Please add the missing exports: `export const TENANT_QUERY_PARAM = "tenantId";` and `export const getTenantIdFromCookie = (): string => "";`.

frontend/src/community/common/utils/axiosInterceptor.ts
```diff
@@ -38,8 +39,9 @@ const requestInterceptorConfig = async (config: InternalAxiosRequestConfig) => {
   }
 
   const isEnterpriseMode = process.env.NEXT_PUBLIC_MODE === "enterprise";
-  if (isEnterpriseMode && tenantID) {
-    config.headers["X-Tenant-ID"] = tenantID;
+  const tenantId = getTenantId();
```
ThinuwanW, line 42: 🤖 **Claude** · 💬 **Note**

Two consistency nits on this block. (1) The enterprise-mode check is inlined as `process.env.NEXT_PUBLIC_MODE === "enterprise"` while the sibling file changed in this same PR (`authInterceptor.ts:26`) uses the `isEnterpriseMode()` helper from `~community/common/utils/commonUtil`, which compares against `appModes.ENTERPRISE` rather than a raw string literal. Since this module now transitively pulls in `commonUtil` anyway (via `tenantUtil`), switch to `isEnterpriseMode()` so both interceptors share one definition. (2) `getTenantId()` is invoked unconditionally before the mode check. In enterprise mode on the auth host it re-parses `window.location.search` and runs a regex over `document.cookie` on every single request; move it inside the branch — `if (isEnterpriseMode()) { const tenantId = getTenantId(); if (tenantId) { ... } }` — which also short-circuits the work in community mode. Same applies to `authInterceptor.ts:25`.

frontend/src/community/common/constants/stringConstants.ts
```diff
@@ -82,4 +82,5 @@ export const SUCCESS = "success";
 export const CANCEL = "cancel";
 
 export const APP = "app";
+export const AUTH = "auth";
```
ThinuwanW, line 85: 🤖 **Claude** · 💬 **Note**

`AUTH = "auth"` is a reserved *subdomain label* (consumed by `enterprise/common/utils/tenantUtil.ts:20` and `pages/enterprise/signin.tsx:79` to detect the auth host), but the bare name collides conceptually with two unrelated `AUTH` symbols already in the codebase: `ROUTES.AUTH` (`constants/routes.ts:2`) and `AUTH = "/auth"` in `constants/configs.ts:7`. Consider naming it `AUTH_SUBDOMAIN` (and, if you're willing to touch it, `APP_SUBDOMAIN` for the adjacent `APP`) so intent is readable at the import site. Worth noting it is also dead in community-only builds — every reference lives in enterprise code — which is acceptable given `APP` sets the same precedent, but a short comment marking these as host labels would help.

frontend/src/community/auth/utils/authInterceptor.ts
```diff
@@ -22,8 +22,9 @@ const authAxios = axios.create({
 //  request interceptor
 authAxios.interceptors.request.use(
   async (config: InternalAxiosRequestConfig) => {
-    if (isEnterpriseMode() && tenantID) {
-      config.headers["X-Tenant-ID"] = tenantID;
+    const tenantId = getTenantId();
```
ThinuwanW, line 25: 🤖 **Claude** · 💬 **Note**

No test coverage exists for either interceptor's tenant-header behaviour (a repo-wide search finds no `authInterceptor`/`axiosInterceptor` test beyond `redirectionHandler.test.ts`, which only imports `authFetch`). Given this PR changes how a security-relevant header is derived — and moves it from a statically-evaluated constant to a per-request call — a small unit test would be cheap insurance: mock `~enterprise/common/utils/tenantUtil` (remember `jest.config.js` maps `~enterprise` to `src/fallback`, so the stub's `""` is what you get unmocked), set `process.env.NEXT_PUBLIC_MODE`, run `authAxios.interceptors.request.handlers[0].fulfilled(config)` and assert `X-Tenant-ID` is present in enterprise mode with a resolved tenant, and absent both in community mode and when `getTenantId()` returns an empty string.

frontend/src/community/common/utils/axiosInterceptor.ts
ThinuwanW, line 4: 🤖 **Claude** · 🔴 **Important**

📁 **File-level observation**

Missed paired change: one `X-Tenant-ID` sender was not migrated. Grepping every producer of that header shows `src/enterprise/common/utils/pmAxiosInterceptor.ts`, `guestAxiosInterceptor.ts`, `enterprise/common/utils/axiosInterceptor.ts`, `sign/api/eSignAxiosInterceptor.ts` and `sign/api/SignApi.ts` all call `getTenantId()`, and this PR converts the two community interceptors — but `src/enterprise/projects/api/projectsApi.ts` (line 34) still does `"X-Tenant-ID": tenantID`, importing the module-level const from this file. `fetchProjects` posts to `NEXT_PUBLIC_PM_API_URL` with the user's bearer token, so on the `auth` host it will send `X-Tenant-ID: auth` (and on any host where the tenant comes from the query param/cookie it sends the wrong tenant), while the bearer token belongs to the real tenant. Fix: in `projectsApi.ts` replace the `tenantID` import with `import { getTenantId } from "~enterprise/common/utils/tenantUtil"` and call it inside `fetchProjects`, or better, route the call through the existing `pmAuthFetch` instance in `pmAxiosInterceptor.ts` which already does exactly this (accessToken + tenant header against the same base URL) instead of building a raw `axios.post`.

frontend/src/community/common/constants/stringConstants.ts
ThinuwanW, line 85: 🤖 **Claude** · 🔴 **Important**

Community-side consumers of the new `auth` host were not updated alongside this constant. (1) `src/enterprise/people/api/GlobalLoginMethodApi.ts` gates its query with `enabled: isEnterpriseMode && tenantName !== "app" && tenantName !== "localhost"` — `"auth"` is not excluded. `BaseLayout.tsx:41` calls it as `useGetGlobalLoginMethod(isEnterprise, tenantID as string)`, so on `auth.<host>` the query is enabled with `tenantName=auth` and fires `/v1/ep/organization/login-method?tenantName=auth` for a tenant that does not exist; `isGlobalLoginMethodLoading` also gates a full-screen loader, so this is on the render path. Add `tenantName !== AUTH` to the `enabled` guard. (2) `pages/index.tsx:46` `handleEnterpriseNavigation` only treats `tenantId === APP` as the tenant-selection host, so on `auth.<host>` a logged-in user is redirected to `ROUTES.DASHBOARD.BASE` with no resolvable tenant; it should be `tenantId === APP || tenantId === AUTH`, matching `pages/enterprise/signin.tsx:81` (`isTenantSelectionHost`). While there, both of those sites hardcode the `"app"`/`"localhost"` string literals instead of importing the `APP`/`LOCALHOST` constants that live in this same file — worth switching so the next host label added here cannot be missed again.

frontend/src/community/common/utils/axiosInterceptor.ts
```diff
@@ -38,8 +39,9 @@ const requestInterceptorConfig = async (config: InternalAxiosRequestConfig) => {
...
+  if (isEnterpriseMode && tenantId) {
```
ThinuwanW, line 43: 🤖 **Claude** · 🟡 **Suggestion**

This change makes a URL query parameter the highest-precedence source of the tenant header for every authenticated request. `getTenantId()` returns `getTenantIdFromQuery() || getTenantIdFromCookie()` on the `auth` host, i.e. `?tenantId=` wins over the tenant cookie, and this interceptor is shared by `authFetch`, `authFetchV2` and `authFetchSameOrigin` — so any page loaded on the auth host with a stale, shared or crafted `?tenantId=` in the URL sends that value on all of its API calls. The backend `TenantFilter` (backend/.../TenantFilter.java:83) throws `COMMON_ERROR_UNAUTHORIZED_ACCESS` whenever the header disagrees with the OAuth session tenant, so a bookmarked/forwarded link with a stale `tenantId` degrades into a blanket unauthorized failure for the whole page rather than a recoverable error, and it broadens the surface for cross-tenant probing. Consider scoping the query-param read to the sign-in/tenant-selection flow that actually needs it (`pages/enterprise/signin.tsx` already reads `query[TENANT_QUERY_PARAM]` directly) and having `getTenantId()` prefer the cookie, or at minimum reverse the precedence to `getTenantIdFromCookie() || getTenantIdFromQuery()`.

frontend/src/community/common/utils/axiosInterceptor.ts
ThinuwanW, line 42: 🤖 **Claude** · 🟡 **Suggestion**

`getTenantId()` is evaluated before the mode guard, so it runs on every single outbound request regardless of mode. Previously the value was a module-level const computed once; now each request performs a `document.cookie` regex match and a `new URLSearchParams(window.location.search)` parse (see `src/enterprise/common/utils/tenantUtil.ts:13-22`). In community mode the fallback stub makes this cheap, but in a build where the enterprise submodule is present and `NEXT_PUBLIC_MODE` is not `enterprise` it is pure waste. Short-circuit instead: `if (isEnterpriseMode) { const tenantId = getTenantId(); if (tenantId) config.headers["X-Tenant-ID"] = tenantId; }`. Same applies to `authInterceptor.ts:25-28`.

frontend/src/community/common/utils/axiosInterceptor.ts
ThinuwanW, line 4: 🤖 **Claude** · 🟡 **Suggestion**

This import gives the API layer a new transitive runtime dependency on `next/server`. `~enterprise/common/utils/tenantUtil` imports `getCookieValue` from `~community/common/utils/commonUtil`, which imports `{ NextRequest, NextResponse }` from `next/server` and uses `NextResponse.redirect` as a value (commonUtil.ts:517), plus `~community/people/utils/holidayUtils/commonUtils.tsx`. `axiosInterceptor.ts` is imported by essentially every API module in the app, so in enterprise builds that whole middleware-oriented module is now pulled into client chunks that previously only needed axios and the auth token helper. (`authInterceptor.ts` already imported `commonUtil`, so only this file regresses.) Suggest moving `getCookieValue` into a small standalone module (e.g. `~community/common/utils/cookieUtil.ts`) and having `tenantUtil` import from there, keeping `commonUtil` — and `next/server` — out of the interceptor's dependency graph.

frontend/src/fallback/common/utils/tenantUtil.ts
ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

Two structural gaps with this stub. (1) `jest.config.js` maps `"^~enterprise(.*)$"` to `<rootDir>/src/fallback$1` unconditionally — not conditionally on mode — so in every jest run `getTenantId()` resolves to this stub and returns `""`. That means the enterprise branch added in both interceptors is unreachable from tests by construction: any test written against them will exercise the falsy path and pass without asserting anything about the header. Tests for the tenant-header behaviour must therefore `jest.mock("~enterprise/common/utils/tenantUtil")` explicitly; worth a short comment here so the next person does not write a green-but-vacuous test. (2) Every other stub in this directory ships a colocated test (`awsS3ServiceFunctions.test.ts`, `firebase.test.ts`, `monitoring.test.ts`, `getEnterpriseDrawerRoutes.test.ts`) — add a `tenantUtil.test.ts` asserting the stub returns `""` to match the established convention, since `collectCoverage` is on and this file is included.

frontend/src/community/common/utils/axiosInterceptor.ts
```diff
@@ -38,8 +39,9 @@ const requestInterceptorConfig = async (config: InternalAxiosRequestConfig) => {
...
+    config.headers["X-Tenant-ID"] = tenantId;
```
ThinuwanW, line 44: 🤖 **Claude** · 🟣 **Nit**

Naming hazard and duplication inside this file. (1) The new local `tenantId` differs from the surviving module-level `tenantID` (line 14) only by the case of a single letter, in the same scope chain — writing `config.headers["X-Tenant-ID"] = tenantID` here would compile cleanly and silently use the stale import-time value. Rename the local to something unambiguous (e.g. `resolvedTenantId`) or, better, delete the module-level const once its consumers are migrated. (2) The private `getSubDomain` helper at lines 9-12 duplicates the exported `getSubdomain` in `src/enterprise/common/utils/commonUtil.ts:44` — the enterprise version differs only by the `D`/`d` casing, adds an empty-url guard and carries an explicit `string | string[]` return type, while this one has no return type and its `multipleValues` parameter is never passed by any caller. Since `tenantUtil` already routes through the enterprise helper, drop the local copy rather than leaving two subdomain parsers with subtly different null behaviour.

frontend/src/community/common/utils/axiosInterceptor.ts
akilarootcode: Agreed — `projectsApi.ts` is the one straggler still reading the module-level `tenantID`, so it is the only path that actually ships a wrong header; repointing it at `getTenantId()` and retiring the export.

frontend/src/fallback/common/utils/tenantUtil.ts
akilarootcode: Valid — bringing the stub to full export parity with `TENANT_QUERY_PARAM` and `getTenantIdFromCookie` so a future community import cannot silently resolve to `undefined`.

frontend/src/community/common/utils/axiosInterceptor.ts
akilarootcode: Fair on both counts — switching to `isEnterpriseMode()` and moving the `getTenantId()` call inside the branch so community mode does no per-request work.

frontend/src/community/common/constants/stringConstants.ts
akilarootcode: Reasonable, but `APP` sets the same precedent immediately above it, so I would rather rename both host labels in one pass than leave the pair half-renamed here.

frontend/src/community/auth/utils/authInterceptor.ts
akilarootcode: Fair — moving a security-relevant header from a static constant to a per-request call is exactly the kind of change worth pinning, and the `~enterprise` mock is the part worth documenting.

frontend/src/community/common/utils/axiosInterceptor.ts
akilarootcode: Same straggler as the file-level note, and routing `fetchProjects` through the existing `pmAuthFetch` is the better fix than hand-rolling the header a second time.

frontend/src/community/common/constants/stringConstants.ts
akilarootcode: Both are real — `auth` needs excluding from the `useGetGlobalLoginMethod` guard and `pages/index.tsx` needs to treat it as a tenant-selection host, and the string literals there should be the constants so the next host label cannot be missed.

frontend/src/community/common/utils/axiosInterceptor.ts
akilarootcode: Good point on precedence — the query param should be scoped to the sign-in flow that needs it rather than winning on every authenticated request from the auth host.

frontend/src/community/common/utils/axiosInterceptor.ts
akilarootcode: Agreed, same as the sibling note — the resolver belongs behind the mode check in both interceptors.

frontend/src/community/common/utils/axiosInterceptor.ts
akilarootcode: Valid — splitting `getCookieValue` into its own module keeps `next/server` out of the interceptor graph, which every API module pulls in.

frontend/src/fallback/common/utils/tenantUtil.ts
akilarootcode: Right, and worth writing down — jest maps `~enterprise` to the stub unconditionally, so a test written against the enterprise branch would pass without asserting anything.

frontend/src/community/common/utils/axiosInterceptor.ts
akilarootcode: Agreed — a one-letter difference between `tenantId` and `tenantID` in the same scope is a genuine trap, and the duplicate `getSubDomain` should go with the module-level const.

hasalarootcode (dismissed):

hasalarootcode (approved):

PR #1727, opened 2026-08-10
fix: restore pending filter when navigating back from timesheet full list view

SanithuGunasena07:
Bug: On the All Timesheets page, the "Request Waiting for Approval" table doesn't show the full/correct list of pending requests after navigating back from the "View Full List" page. Only a subset of pending requests appears, and a manual page reload is required to see the correct list again.

Root Cause: The pending-requests widget (ManagerTimesheetRequestTable on All Timesheets) and the "View Full List" page share the same timesheetRequestParams state in the manager timesheet filter slice. When the full-list page unmounts (i.e. the user navigates back), a cleanup effect in useTimesheetRequestFilterState.ts called resetTimesheetRequestParams(), which sets status: ""  the correct default for entering the full list (show all statuses), but wrong for the widget, which needs status: PENDING. As a result, on navigating back, the widget's query fetched the most recent requests of any status instead of pending-only, then filtered that already-limited result set for pending items client-side surfacing only whichever pending requests happened to be in that unfiltered batch.

Fix: Added a separate restoreTimesheetRequestParams action (in managerTimesheetFiltersSlice.ts) that resets status back to PENDING instead of "", and wired the unmount cleanup in useTimesheetRequestFilterState.ts to call it (via a new cleanupParams reference) instead of reusing resetTimesheetRequestParams, which remains unchanged for its original uses (entering the full list, and the filter popover's "Reset" button).

### DIFF
--- a/frontend/src/community/attendance/hooks/useTimesheetRequestFilterState.ts
+++ b/frontend/src/community/attendance/hooks/useTimesheetRequestFilterState.ts
@@ -18,7 +18,8 @@ export const useTimesheetRequestFilterState = (
     timesheetRequestSelectedDates,
     setTimesheetRequestsFilters,
     setTimesheetRequestSelectedDates,
-    resetTimesheetRequestParams
+    resetTimesheetRequestParams,
+    restoreTimesheetRequestParams
   } = useAttendanceStore();
 
   const filterValues = isManager
@@ -39,6 +40,9 @@ export const useTimesheetRequestFilterState = (
   const resetParams = isManager
     ? resetTimesheetRequestParams
     : resetEmployeeTimesheetRequestParams;
+  const cleanupParams = isManager
+    ? restoreTimesheetRequestParams
+    : resetEmployeeTimesheetRequestParams;
 
   const [appliedStartDate, appliedEndDate] = appliedDates;
   const filterCount =
@@ -47,10 +51,10 @@ export const useTimesheetRequestFilterState = (
   useEffect(() => {
     if (!shouldRegisterCleanup) return;
     return () => {
-      resetParams();
+      cleanupParams();
       setDates(["", ""]);
     };
-  }, [shouldRegisterCleanup, resetParams, setDates]);
+  }, [shouldRegisterCleanup, cleanupParams, setDates]);
 
   return {
     filterValues,
--- a/frontend/src/community/attendance/hooks/useTimesheetRequestFilterState.ts
+++ b/frontend/src/community/attendance/hooks/useTimesheetRequestFilterState.ts
@@ -50,11 +50,12 @@ export const useTimesheetRequestFilterState = (
 
   useEffect(() => {
     if (!shouldRegisterCleanup) return;
+    resetParams();
     return () => {
       cleanupParams();
       setDates(["", ""]);
     };
-  }, [shouldRegisterCleanup, cleanupParams, setDates]);
+  }, [shouldRegisterCleanup, resetParams, cleanupParams, setDates]);
 
   return {
     filterValues,
--- a/frontend/src/community/attendance/store/slices/managerTimesheetFiltersSlice.ts
+++ b/frontend/src/community/attendance/store/slices/managerTimesheetFiltersSlice.ts
@@ -78,7 +78,7 @@ const managerTimesheetFiltersSlice = (
   restoreTimesheetRequestParams: () => {
     set((state) => ({
       timesheetRequestsFilters: {
-        status: [TimeSheetRequestStates.PENDING] as string[]
+        status: [] as string[]
       },
       timesheetRequestParams: {
         ...state.timesheetRequestParams,
--- a/frontend/src/community/attendance/store/slices/managerTimesheetFiltersSlice.ts
+++ b/frontend/src/community/attendance/store/slices/managerTimesheetFiltersSlice.ts
@@ -14,7 +14,7 @@ const managerTimesheetFiltersSlice = (
   set: SetType<ManagerTimesheetFiltersSliceTypes>
 ): ManagerTimesheetFiltersSliceTypes => ({
   timesheetRequestsFilters: {
-    status: []
+    status: [TimeSheetRequestStates.PENDING]
   },
   timesheetRequestSelectedDates: [],
   timesheetRequestParams: {
@@ -78,7 +78,7 @@ const managerTimesheetFiltersSlice = (
   restoreTimesheetRequestParams: () => {
     set((state) => ({
       timesheetRequestsFilters: {
-        status: [] as string[]
+        status: [TimeSheetRequestStates.PENDING] as string[]
       },
       timesheetRequestParams: {
         ...state.timesheetRequestParams,
--- a/frontend/src/community/attendance/store/slices/managerTimesheetFiltersSlice.ts
+++ b/frontend/src/community/attendance/store/slices/managerTimesheetFiltersSlice.ts
@@ -75,6 +75,19 @@ const managerTimesheetFiltersSlice = (
     }));
   },
 
+  restoreTimesheetRequestParams: () => {
+    set((state) => ({
+      timesheetRequestsFilters: {
+        status: [] as string[]
+      },
+      timesheetRequestParams: {
+        ...state.timesheetRequestParams,
+        status: TimeSheetRequestStates.PENDING,
+        page: 1
+      }
+    }));
+  },
+
   setTimesheetRequestSelectedDates: (value: string[]) => {
     set((state) => ({
       ...state,
--- a/frontend/src/community/attendance/store/slices/managerTimesheetFiltersSlice.ts
+++ b/frontend/src/community/attendance/store/slices/managerTimesheetFiltersSlice.ts
@@ -76,9 +76,9 @@ const managerTimesheetFiltersSlice = (
   },
 
   restoreTimesheetRequestParams: () => {
-    set((state) => ({
+    set((state): Partial<ManagerTimesheetFiltersSliceTypes> => ({
       timesheetRequestsFilters: {
-        status: [] as string[]
+        status: []
       },
       timesheetRequestParams: {
         ...state.timesheetRequestParams,
--- a/frontend/src/community/attendance/types/attendanceStoreTypes.ts
+++ b/frontend/src/community/attendance/types/attendanceStoreTypes.ts
@@ -18,6 +18,7 @@ interface actionsTypes {
     selectedFilters: Record<string, string[]>
   ) => void;
   resetTimesheetRequestParams: () => void;
+  restoreTimesheetRequestParams: () => void;
   setTimesheetRequestSelectedDates: (value: string[]) => void;
   setTimesheetRequestPagination: (page: number) => void;
   setTimesheetAnalyticsSelectedDates: (value: string[]) => void;
--- a/frontend/src/community/attendance/types/attendanceSliceTypes.ts
+++ b/frontend/src/community/attendance/types/attendanceSliceTypes.ts
@@ -25,6 +25,7 @@ export interface ManagerTimesheetFiltersSliceTypes
     | "timesheetRequestsFilterValues"
     | "setTimesheetRequestsFilters"
     | "resetTimesheetRequestParams"
+    | "restoreTimesheetRequestParams"
     | "setTimesheetRequestSelectedDates"
     | "setTimesheetRequestPagination"
     | "timesheetAnalyticsParams"


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds a `restoreTimesheetRequestParams` store action so that leaving the manager timesheet-requests full-list view restores the default `PENDING` status filter instead of clearing it — which correctly fixes the empty-widget bug caused by the dashboard table client-side filtering to PENDING while the shared query param was `""`. The typing changes (`attendanceSliceTypes.ts`, `attendanceStoreTypes.ts`) are correct and complete, the Zustand action uses immutable `set()` updates consistent with the neighbouring actions, and the `useEffect` dependency array was updated correctly (the store action identities are stable, so no stale-closure or re-run risk). Remaining concerns are state-consistency between `timesheetRequestsFilters` and `timesheetRequestParams`, duplication of the default status literal, an extra refetch introduced on navigating away, and missing test coverage.

Found **15** new issue(s): 🔴 1 important, 🟡 9 suggestion(s), 🟣 5 nit(s)

frontend/src/community/attendance/store/slices/managerTimesheetFiltersSlice.ts
```diff
@@ -75,6 +75,19 @@ const managerTimesheetFiltersSlice = (
     }));
   },
 
+  restoreTimesheetRequestParams: () => {
+    set((state) => ({
+      timesheetRequestsFilters: {
```
ThinuwanW, line 80: 🤖 **Claude** · 🟡 **Suggestion**

`restoreTimesheetRequestParams` leaves `timesheetRequestsFilters.status` empty (`[]`) while setting `timesheetRequestParams.status` to `PENDING`. These two pieces of state now disagree: `useTimesheetRequestFilterState` derives `filterCount` from `timesheetRequestsFilters.status.length`, and `TimesheetRequestFilterBody` seeds its `selectedStatus` from the same array. So if the full-list page is entered without going through the 'View Full List' button (browser forward/history restore, or any future nav entry point), the table will show only PENDING rows while the filter badge reads 0 and the Pending chip appears unselected — and the filter's Reset button is disabled because `isEmpty` is true. Consider setting `timesheetRequestsFilters: { status: [TimeSheetRequestStates.PENDING] }` here so the UI reflects the effective filter (note the same mismatch exists in the slice's initial state at lines 16-26, so fixing both keeps them aligned).

frontend/src/community/attendance/store/slices/managerTimesheetFiltersSlice.ts
```diff
@@ -75,6 +75,19 @@ const managerTimesheetFiltersSlice = (
     }));
   },
 
+  restoreTimesheetRequestParams: () => {
```
ThinuwanW, line 78: 🤖 **Claude** · 🟡 **Suggestion**

`restoreTimesheetRequestParams` is a byte-for-byte copy of `resetTimesheetRequestParams` (lines 65-76) apart from the status literal, and it re-hardcodes `TimeSheetRequestStates.PENDING`, which is already the default at line 21. If the default status ever changes, the initial state and this action will silently diverge. Extract a module-level constant (e.g. `const DEFAULT_TIMESHEET_REQUEST_PARAMS = { status: TimeSheetRequestStates.PENDING, startDate: "", endDate: "", page: 1, size: 5 }`) used by the initial state and by both actions, or collapse the two into one private helper parameterised by the status value.

frontend/src/community/attendance/store/slices/managerTimesheetFiltersSlice.ts
```diff
@@ -75,6 +75,19 @@ const managerTimesheetFiltersSlice = (
...
+    set((state) => ({
+      timesheetRequestsFilters: {
+        status: [] as string[]
+      },
+      timesheetRequestParams: {
+        ...state.timesheetRequestParams,
```
ThinuwanW, line 84: 🤖 **Claude** · 🟡 **Suggestion**

The action is not self-contained: it spreads the existing `timesheetRequestParams`, so `startDate`/`endDate` (and `timesheetRequestSelectedDates`) survive the 'restore'. It only works today because the single caller in `useTimesheetRequestFilterState` immediately follows it with `setDates(["", ""])`. A future caller invoking the action alone would end up with cleared status chips but a still-applied date range. Clear `startDate`, `endDate` and `timesheetRequestSelectedDates` inside the action so the store guarantees the restored state on its own.

frontend/src/community/attendance/hooks/useTimesheetRequestFilterState.ts
```diff
@@ -18,7 +18,8 @@ export const useTimesheetRequestFilterState = (
     timesheetRequestSelectedDates,
     setTimesheetRequestsFilters,
     setTimesheetRequestSelectedDates,
-    resetTimesheetRequestParams
+    resetTimesheetRequestParams,
```
ThinuwanW, line 21: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Architectural: the dashboard widget (`hasFullList=false`) and the full-list page share one `timesheetRequestParams` object, which is why an unmount-cleanup hack is needed to keep the widget working. Two side effects of this design that the PR makes slightly worse: (1) `ManagerTimesheetRequestTable` still filters `requestData.items` to PENDING client-side (line 96-100) even though the server now returns only PENDING, so the widget can under-fill its 5-row page; (2) on navigating back from the full list, the destination widget renders once with the old status (`""` set by the 'View Full List' button) and fires a query, then the cleanup flips it to `PENDING` and fires a second query — previously the cleanup was a no-op in that common path. Consider giving the dashboard widget its own fixed `{ status: PENDING, page: 1, size: 5 }` params (e.g. an argument to `useGetManagerTimeSheetRequests`) instead of mutating shared filter state on unmount; that would let both the restore action and the client-side filter go away.

frontend/src/community/attendance/hooks/useTimesheetRequestFilterState.ts
```diff
@@ -39,6 +40,9 @@ export const useTimesheetRequestFilterState = (
   const resetParams = isManager
     ? resetTimesheetRequestParams
     : resetEmployeeTimesheetRequestParams;
+  const cleanupParams = isManager
+    ? restoreTimesheetRequestParams
```
ThinuwanW, line 44: 🤖 **Claude** · 🟣 **Nit**

`resetParams` and `cleanupParams` now differ only for the manager branch (`restoreTimesheetRequestParams` vs `resetTimesheetRequestParams`), while the employee branch resolves both to `resetEmployeeTimesheetRequestParams`. This is easy to misread as an accidental duplicate. Add a short comment explaining the intent — that unmounting the manager full-list view must restore the default PENDING filter used by the dashboard widget, whereas the filter popover's Reset button must clear it — or name it `unmountParams`/`restoreDefaultParams` to make the distinction from `resetParams` explicit at the call site.

frontend/src/community/attendance/types/attendanceStoreTypes.ts
```diff
@@ -18,6 +18,7 @@ interface actionsTypes {
     selectedFilters: Record<string, string[]>
   ) => void;
   resetTimesheetRequestParams: () => void;
+  restoreTimesheetRequestParams: () => void;
```
ThinuwanW, line 21: 🤖 **Claude** · 🟣 **Nit**

`restoreTimesheetRequestParams` is a vague name next to `resetTimesheetRequestParams` — 'restore' does not convey 'reset to the default PENDING filter', and the two are trivially confused at call sites. Something like `resetTimesheetRequestParamsToDefault` (or `applyDefaultTimesheetRequestParams`) would document the behaviour in the name; remember to rename in the slice, the `Pick<>` union in `attendanceSliceTypes.ts`, and the hook.

frontend/src/community/attendance/store/slices/managerTimesheetFiltersSlice.ts
ThinuwanW, line 78: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

No test accompanies the new store action, though the repo has precedent for slice unit tests (e.g. `src/community/people/store/slices/projectTeamSlice.test.ts`, `terminateEmployeeSlice.test.ts`). This is a regression-prone behaviour (the bug being fixed is exactly a wrong default status). Add a `managerTimesheetFiltersSlice.test.ts` asserting that after applying a status filter, `restoreTimesheetRequestParams()` yields `timesheetRequestParams.status === TimeSheetRequestStates.PENDING`, `page === 1`, and empty `timesheetRequestsFilters.status` — and ideally a test on `useTimesheetRequestFilterState` verifying the unmount cleanup runs only when `shouldRegisterCleanup` is true.

frontend/src/community/attendance/hooks/useTimesheetRequestFilterState.ts
```diff
@@ -47,10 +51,10 @@ export const useTimesheetRequestFilterState = (
   useEffect(() => {
     if (!shouldRegisterCleanup) return;
     return () => {
-      resetParams();
+      cleanupParams();
```
ThinuwanW, line 54: 🤖 **Claude** · 🔴 **Important**

Doing the restore in an unmount cleanup makes the destination page render once with the stale params, producing a wasted request and a visible flash. Sequence: the manager is on the full list with `timesheetRequestParams.status === ""` (that is what the "View full list" button sets). On navigation, React renders the new tree first, so the dashboard's `useGetManagerTimeSheetRequests` (attendanceManagerApi.ts:91) mounts and fires with `status=""`; only afterwards does this cleanup commit `status=PENDING` + `startDate/endDate=""`, which changes the query key and fires a second request. There is no `staleTime` on that query, so both requests really go out. During the first render the widget receives a page of 5 mixed-status rows and client-side filters them down (ManagerTimesheetRequestTable.tsx:96-100), so the user sees 1-2 rows before it settles at 5. Fix: make the full-list view own its own entry/exit params — set the desired params on mount of the full-list page (or give the dashboard query its own fixed PENDING param object rather than sharing `timesheetRequestParams`) instead of relying on the previous page's teardown to fix up state the next page has already read.

frontend/src/community/attendance/store/slices/managerTimesheetFiltersSlice.ts
ThinuwanW, line 84: 🤖 **Claude** · 🟡 **Suggestion**

`restoreTimesheetRequestParams` spreads the previous params and only overrides `status` and `page`, so `size` is never restored. Both sibling implementations do restore it: `resetEmployeeTimesheetRequestParams` explicitly writes `size: 4` (employeeTimesheetFilterSlice.ts:55) and `resetLeaveRequestParams` writes `size: "4"` (leaveRequestFilterSlice.ts:81). Today the manager widget and the full list both happen to use `size: 5`, so nothing breaks — but the whole point of this action is that the two views share one params object, so the first time someone adds a page-size selector or a larger page size to the full list, that size leaks into the dashboard widget and the widget silently renders a different number of rows. Extract the initial `timesheetRequestParams` literal (lines 20-26) into an exported `DEFAULT_TIMESHEET_REQUEST_PARAMS` constant and have this action return `{ ...DEFAULT_TIMESHEET_REQUEST_PARAMS, status: TimeSheetRequestStates.PENDING }` so the default can never drift from the restore.

frontend/src/community/attendance/store/slices/managerTimesheetFiltersSlice.ts
```diff
@@ -75,6 +75,19 @@ const managerTimesheetFiltersSlice = (
     }));
   },
 
+  restoreTimesheetRequestParams: () => {
+    set((state) => ({
+      timesheetRequestsFilters: {
+        status: [] as string[]
```
ThinuwanW, line 81: 🤖 **Claude** · 🟡 **Suggestion**

`status: [] as string[]` uses a type assertion to work around inference, which the type-safety guideline explicitly calls out ("using type assertions (`as`) to silence errors instead of fixing the underlying type"). The assertion is only needed because the `set` callback's return type is being inferred from the object literal. Fix it at the source: annotate the returned object, e.g. `set((state): Partial<ManagerTimesheetFiltersSliceTypes> => ({ timesheetRequestsFilters: { status: [] }, ... }))`, and drop the `as`. Same applies to the pre-existing copy on line 68 — worth cleaning both up together rather than propagating the pattern into new code.

frontend/src/community/attendance/store/slices/managerTimesheetFiltersSlice.ts
ThinuwanW, line 78: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

i18n: `timesheetRequestsFilterValues.status` (lines 27-34) hardcodes the user-facing labels "Pending", "Approved", "Denied", "Cancelled". These are rendered verbatim as chip labels by `SelectableItemList` in TimesheetRequestFilterBody.tsx:83-86, so they are untranslated UI strings. This is directly in scope for this PR — it is the same status-filter state the new `restoreTimesheetRequestParams` resets. Store only the enum values in the slice and resolve labels in the component via `useTranslator("attendanceModule", "timesheet")` with keys like `timesheet.statusFilter.pending`. Note the employee slice (employeeTimesheetFilterSlice.ts:18-25) has the identical hardcoded list plus raw string literals instead of `TimeSheetRequestStates` — fix both in one pass.

frontend/src/community/attendance/hooks/useTimesheetRequestFilterState.ts
ThinuwanW, line 21: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

The entry half of the lifecycle this PR fixes still lives outside the hook: `ManagerTimesheetRequestTable.tsx:82,384` reaches into the store directly for `resetTimesheetRequestParams()` before `router.push(ROUTES.TIMESHEET.TIMESHEET_REQUESTS)`, while this hook now owns the exit half (`cleanupParams`). Having the paired enter/leave transitions in two different files with two different access paths is exactly how the original bug arose, and it means the full list only shows all statuses when reached via that one button — sidebar/deep-link/refresh entry lands on `status=PENDING` with `filterCount` 0. Expose the entry transition from this hook too (e.g. return a `prepareFullListParams` alongside `resetParams`) and have the table call that instead of the raw store action, so both halves stay in sync in one place.

frontend/src/community/attendance/hooks/useTimesheetRequestFilterState.ts
```diff
@@ -18,7 +18,8 @@ export const useTimesheetRequestFilterState = (
     timesheetRequestSelectedDates,
     setTimesheetRequestsFilters,
     setTimesheetRequestSelectedDates,
-    resetTimesheetRequestParams
+    resetTimesheetRequestParams,
+    restoreTimesheetRequestParams
```
ThinuwanW, line 22: 🤖 **Claude** · 🟣 **Nit**

`useAttendanceStore()` is called with no selector, so every consumer of this hook re-subscribes to the entire attendance store and re-renders on any unrelated change (modal open/close, clock-in type, time availability, attendance params). This PR adds a 12th destructured member, and the hook is consumed by two large tables plus the filter popover. The same module already demonstrates the correct pattern — `attendanceManagerApi.ts:92-94` uses `useAttendanceStore((state) => state.timesheetRequestParams)`. Prefer per-field selectors (or `useShallow` over the subset this hook needs) so the tables only re-render when timesheet filter state actually changes.

frontend/src/community/attendance/types/attendanceStoreTypes.ts
ThinuwanW, line 21: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

Dead/duplicate declaration in this changed file: `setTimesheetAnalyticsTeamName: (teamName: string) => void;` is declared on line 27 inside `actionsTypes` and again on line 93 inside `AttendanceStore`, which already extends `actionsTypes`. The second declaration is redundant and, being interleaved with state fields, obscures the actions/state separation this file otherwise keeps. Remove line 93 while you are already editing this interface.

frontend/src/community/attendance/hooks/useTimesheetRequestFilterState.ts
ThinuwanW, line 21: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

The hook has no explicit return type — the shape of the returned object (`filterValues`, `appliedStatus`, `appliedDates`, `setFilters`, `setDates`, `resetParams`, `filterCount`) is entirely inferred, and `shouldRegisterCleanup` relies on inference from its default value. The type-safety checklist requires explicit types for parameters and return values. Declare an exported `UseTimesheetRequestFilterStateReturn` interface and annotate both the parameter (`shouldRegisterCleanup: boolean = true`) and the return, so that the newly added `cleanupParams` branch and any future additions can't silently change the public shape of the hook.

frontend/src/community/attendance/store/slices/managerTimesheetFiltersSlice.ts
SanithuGunasena07: Fixed by setting timesheetRequestsFilters.status to [TimeSheetRequestStates.PENDING] in restoreTimesheetRequestParams

frontend/src/community/attendance/store/slices/managerTimesheetFiltersSlice.ts
SanithuGunasena07: Not needed (Change uses the same patterns and conventions used in the code file)

frontend/src/community/attendance/store/slices/managerTimesheetFiltersSlice.ts
SanithuGunasena07: Out of scope for PR (Was already there not a change made in this PR)

frontend/src/community/attendance/hooks/useTimesheetRequestFilterState.ts
SanithuGunasena07: Out of scope for PR (Its a full achitectural change for the pre existing architecture)

frontend/src/community/attendance/hooks/useTimesheetRequestFilterState.ts
SanithuGunasena07: not required

frontend/src/community/attendance/store/slices/managerTimesheetFiltersSlice.ts
SanithuGunasena07: not required

frontend/src/community/attendance/types/attendanceStoreTypes.ts
SanithuGunasena07: keeping it as restoreTimesheetRequestParams as it is similler to the naming cnvention of resetTimesheetRequestParams

frontend/src/community/attendance/hooks/useTimesheetRequestFilterState.ts
SanithuGunasena07: Out of scope as  fix touches attendanceManagerApi.ts and ManagerTImesheet.tsx, which are outside this PR's diff

frontend/src/community/attendance/store/slices/managerTimesheetFiltersSlice.ts
SanithuGunasena07: not required (Used the same convention used in the code already)

frontend/src/community/attendance/store/slices/managerTimesheetFiltersSlice.ts
SanithuGunasena07: not required (Used the same convention used in the code already)

frontend/src/community/attendance/store/slices/managerTimesheetFiltersSlice.ts
SanithuGunasena07: Out of scope for PR (Was already there not a change made in this PR)

frontend/src/community/attendance/hooks/useTimesheetRequestFilterState.ts
SanithuGunasena07: Out of scope

frontend/src/community/attendance/hooks/useTimesheetRequestFilterState.ts
SanithuGunasena07: Not a change done in this PR

frontend/src/community/attendance/types/attendanceStoreTypes.ts
SanithuGunasena07: Not a change done in this PR

frontend/src/community/attendance/hooks/useTimesheetRequestFilterState.ts
SanithuGunasena07: Not a change done in this PR

frontend/src/community/attendance/hooks/useTimesheetRequestFilterState.ts
```diff
@@ -39,6 +40,9 @@ export const useTimesheetRequestFilterState = (
   const resetParams = isManager
```
shakila-kamalasena, line 40: Remove the unused resetParams

shakila-kamalasena (changes requested):

frontend/src/community/attendance/hooks/useTimesheetRequestFilterState.ts
SanithuGunasena07: resetParams isn't called within this file — it's just computed and passed through in the return statement so TimesheetRequestFilterBody.tsx:55 can call it (the filter popover's Reset button). Not removing it, since removing it would break that button.

shakila-kamalasena (dismissed):

frontend/src/community/attendance/store/slices/managerTimesheetFiltersSlice.ts
akilarootcode: remove as

akilarootcode (changes requested):

frontend/src/community/attendance/store/slices/managerTimesheetFiltersSlice.ts
SanithuGunasena07: removed

akilarootcode (approved):

shakila-kamalasena (approved):

PR #1728, opened 2026-08-10
refactor:  CRM types for Company, Contact, Deal, and Task entities

YasinduRC:
## CRM-713 — Define new CRM entity types

**This PR adds types only.** It introduces one new file, `frontend/src/community/crm/types/CrmCommonTypes.ts`, which is the target type layer for the CRM store redesign. Nothing imports it yet, no existing file is touched, and there is no behaviour change.

### Why this is a standalone PR

The migration is sequenced **types → store + backend → UI**. This PR locks the shape so the backend contracts and the new store slices can be built against something fixed. Reviewing it in isolation is intentional.

### What's in it

- **Entities** — `CompanyEntity`, `ContactEntity`, `DealEntity`, `DealStageEntity`, `TaskEntity`
- **Metrics** — `CompanyMetrics`, `ContactMetrics`
- **Lookups** — `CompanyLookup`, `ContactLookup`, `DealLookup`, `Owner`, `TaskType`
- **Board** — `BoardColumn`, `DealMoveResult`
- **Filters** — request types per endpoint

### Conventions applied (please review against these)

- **Optional-first**, no `| null` unions — absence is `undefined`
- **One base type per entity** serving entity + form + create + edit, matching the people module — no separate `*FormTypes` / `*CreatePayload` / `*EditPayload`
- **Relations as ids**, not nested objects (`companyId`, not `company: {...}`)
- **Board stores ids only** — `BoardColumn.dealIds` points into the deal map; stage metadata lives separately
- **Monetary values as `string`** throughout
- `undefined` = not loaded, `[]` = loaded and empty

### What to review

Field coverage per entity, naming, relationship directions, and whether the conventions above are applied consistently.

### What is *not* in scope for this review

- **Duplication with `types/CommonTypes.ts`** — expected and intentional. The existing types stay in place until the store and UI migration PRs land, and are deleted there. Please don't raise overlap as a finding.
- **Nothing imports this file** — by design; consumers arrive in the store PR.
- **Mismatches with current API responses** — see below.

### Backend

**Backend responses will be aligned in a separate PR.** This file describes the target shape, not what the API returns today. Known deltas already identified and tracked for that PR:

- `CrmContactListItemDto.closedDealValue` is `BigDecimal` (serialises as a JSON number); every other monetary field is already `String`
- Contact `name` splits into `firstName` / `lastName`
- No `GET /company/{id}` exists yet
- Contact metric fields are split across the list and detail DTOs

### Verification

- `npx tsc --noEmit` — zero errors originating from this file (pre-existing repo errors unchanged)
- `npx prettier --check` — clean
- No runtime impact; the file has no consumers

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds a single new 230-line types file, `frontend/src/community/crm/types/CrmCommonTypes.ts`, containing normalized-entity shapes (entities + id references), lookups, and filter request types for the CRM module. I read the full file, the sibling `crm/types/CommonTypes.ts`, `BoardTypes.ts`, `StoreTypes.ts`/`SliceTypes.ts`, `crm/enums/common.ts`, the CRM API layer, and cross-checked every field against the backend `crmplanner` request/response DTOs and JPA models. The file currently has zero consumers and substantially duplicates the existing `CommonTypes.ts` in the same directory, and several field names/types do not match the backend contract (contact `firstName`/`lastName` vs backend `name`, `closedDealValue` as string vs BigDecimal, and `?`-optional used where the backend serializes explicit `null`).

Found **21** new issue(s): 🔴 11 important, 🟡 6 suggestion(s), 🟣 4 nit(s)

frontend/src/community/crm/types/CrmCommonTypes.ts
```diff
@@ -0,0 +1,230 @@
...
+
+export interface CompanyDomainSearchResponseType {
+  companies: CompanyEntity[];
+}
+
+// Contact
+
+export interface ContactEntity {
+  id?: number;
+  firstName?: string;
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`ContactEntity.firstName`/`lastName` (and `ContactLookup.firstName`/`lastName` at lines 86-87) do not match the backend contract. The CRM contact model (`backend/.../crmplanner/model/CrmContact.java`) and every contact DTO — `CrmContactResponseDto`, `CrmContactListItemDto`, `CrmContactLookupResponseDto`, `CrmContactDetailResponseDto` — expose a single `name: String`. There is no `firstName`/`lastName` on a CRM contact; those fields only exist on the *owner* (`CrmContactOwnerResponseDto`/`CrmOwnerResponseDto`, i.e. an Employee). The existing frontend types (`CommonTypes.ts` `CrmContact`, `CrmContactType`, `CrmContactLookup`) all correctly use `name`. As written, any component consuming these will render `undefined` for the contact name. Fix: replace `firstName`/`lastName` with `name: string | null` on `ContactEntity` and `name: string` on `ContactLookup`, or confirm a backend split is landing in the same release.

frontend/src/community/crm/v2/types/CrmCommonTypes.ts
```diff
@@ -0,0 +1,230 @@
+import { SortOrderTypes } from "~community/common/types/CommonTypes";
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

📁 **File-level observation**

This file duplicates a large portion of the existing `frontend/src/community/crm/types/CommonTypes.ts` (which is already the CRM module's common types file) and currently has zero consumers anywhere in `frontend/` — I grepped for `CrmCommonTypes`, `CompanyEntity`, `ContactEntity`, `DealEntity`, `TaskEntity`, `DealStageEntity`, `DealMoveResult`, `BoardColumn` and the only hits are the declarations themselves. Nearly every type here has a pre-existing counterpart: `Owner`→`CrmOwner`, `CompanyLookup`→`CompanyLookup`, `ContactLookup`→`CrmContactLookup`, `DealLookup`→`CrmDealLookup`, `DealStageEntity`→`CrmDealStageType`, `DealStageReorderItem`→`CrmDealStageReorderItem`, `TaskType`→`CrmTaskCategory`, `CompanyDomainSearchResponseType`→`CrmCompanyDomainSearchResponseType`, `DealFilterRequest`→`CrmDealFilterParams`, `TaskRelatedFilterRequest`→`RelatedTasksParams`, `DealsByStagesRequest`→`CrmBoardDealsGroupedRequest` (BoardTypes.ts), `DealMoveResult`→`CrmBoardReorderResultType`. Landing 230 lines of unused, near-duplicate types leaves the module with two competing sources of truth and no compiler pressure to keep them in sync. Also, the filename `CrmCommonTypes.ts` sitting next to `CommonTypes.ts` inside `crm/types/` is itself ambiguous. Recommend either (a) landing this together with the migration/consumers that use it, or (b) adding these normalized shapes to the existing `CommonTypes.ts` (or a clearly-named `NormalizedEntityTypes.ts`) and deleting the redundant re-declarations.

frontend/src/community/crm/types/CrmCommonTypes.ts
```diff
@@ -0,0 +1,230 @@
...
+  id?: number;
+  openTasksCount?: number;
+  overdue?: number;
+  openValue?: string;
+  accountValue?: string;
+  openDeals?: number;
+  closedDeals?: number;
+}
+
+export interface CompanyLookup {
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`CompanyLookup` is declared here with the exact same name and shape as `CompanyLookup` in `frontend/src/community/crm/types/CommonTypes.ts:102`, in the same directory. Two identically-named exported interfaces in sibling modules is an import hazard: auto-import will pick arbitrarily, and the two definitions can silently drift. `ContactLookup` (line 84) references this local copy while existing code references the `CommonTypes.ts` one. Fix: import `CompanyLookup` from `./CommonTypes` instead of redeclaring it, or delete one of the two.

frontend/src/community/crm/types/CrmCommonTypes.ts
```diff
@@ -0,0 +1,230 @@
...
+  lastContactAt?: string;
+  lastModifiedDate?: string;
+  companyId?: number;
+  ownerId?: number;
+  totalRevenue?: string;
+  pipelineRevenue?: string;
+  activeDealsCount?: number;
+  openTasksCount?: number;
+  overdueTasksCount?: number;
+  closedDealValue?: string;
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`closedDealValue?: string` is the wrong primitive type (same issue in `ContactMetrics` at line 80). The backend declares `private BigDecimal closedDealValue;` in `CrmContactListItemDto`, which Jackson serializes as a JSON **number**, not a string. The existing frontend type `CrmContact.closedDealValue` in `CommonTypes.ts:419` correctly types it as `number | null`. Note the contrast with `totalRevenue`/`pipelineRevenue`, which really are `String` on the backend (`CrmContactDetailResponseDto`) — so those two are correct as `string` here. Fix: change `closedDealValue` to `number | null` in both `ContactEntity` and `ContactMetrics`.

frontend/src/community/crm/types/CrmCommonTypes.ts
```diff
@@ -0,0 +1,230 @@
...
+  CrmDealStageColorsEnum,
+  CrmDealStageEnum,
+  CrmIndustryEnum,
+  CrmPriorityEnum
+} from "../enums/common";
+
+// Company
+
+export interface CompanyEntity {
+  id?: number;
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Nullability is modelled with `?` (i.e. `T | undefined`) throughout, but the backend sends explicit JSON `null` for these fields. There is no global `spring.jackson.default-property-inclusion: non_null` / `@JsonInclude(NON_NULL)` on the CRM DTOs — I grepped `backend/src/main` and the only `NON_NULL` usage is an unrelated BankID request DTO — so nullable columns (`website`, `address`, `contactNumber`, `description`, `amount`, `closingAt`, `dueAt`, `notes`, `lastContactAt`, `previousDealId`/`nextDealId`, etc.) arrive as `null`, not as absent keys. Every existing CRM type in the sibling `CommonTypes.ts`/`BoardTypes.ts` correctly uses `| null` for exactly these fields (e.g. `CrmCompany.website: string | null`, `CrmBoardDealNeighboursType.previousDealId: number | null`). Typing them as `?: T` tells TypeScript the value can never be `null`, which defeats strict-null checking at the exact places it matters — `value.trim()`, `value.length`, `Object.keys` merges, and MUI controlled inputs that warn on `value={null}`. Fix: use `field?: T | null` (or `field: T | null`) for all backend-nullable fields across `CompanyEntity`, `ContactEntity`, `DealEntity`, `TaskEntity`, `DealStageEntity`, `Owner`, `ContactLookup`, and `DealMoveResult`.

frontend/src/community/crm/types/CrmCommonTypes.ts
```diff
@@ -0,0 +1,230 @@
...
+}
+
+export interface ContactLookup {
+  id: number;
+  firstName: string;
+  lastName?: string;
+  company?: CompanyLookup;
+}
+
+export interface Owner {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Naming deviates from the module convention. `Owner`, `TaskType`, `BoardColumn`, `CompanyLookup`, `ContactLookup`, `DealLookup`, `DealStageEntity`, `DealStageReorderItem` are unprefixed, while the rest of the CRM module consistently prefixes shared exported types with `Crm` (`CrmOwner`, `CrmTaskCategory`, `CrmDealLookup`, `CrmDealStageType`, `CrmDealStageReorderItem`, `CrmBoardStageDealsType`, …). `Owner` and `TaskType` in particular are very generic names for module-wide exports and will collide/confuse in files that also import HR or common-module types. Fix: prefix these with `Crm` to match the surrounding code.

frontend/src/community/crm/types/CrmCommonTypes.ts
```diff
@@ -0,0 +1,230 @@
...
+  openValue?: string;
+  accountValue?: string;
+  openDeals?: number;
+  closedDeals?: number;
+  contactIds?: number[];
+  dealIds?: number[];
+  taskIds?: number[];
+}
+
+export interface CompanyMetrics {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`CompanyMetrics` (lines 31-39) restates the exact metric fields already present on `CompanyEntity` (lines 20-25), and `ContactMetrics` (lines 73-82) restates those on `ContactEntity` (lines 62-68). Two copies of the same field list will drift the first time a metric is added or its type is corrected (see the `closedDealValue` issue above, which now has to be fixed in two places). Fix: derive one from the other, e.g. `export interface CompanyMetrics extends Pick<CompanyEntity, "id" | "openTasksCount" | "overdue" | "openValue" | "accountValue" | "openDeals" | "closedDeals"> {}`, or define the metrics interface first and have the entity `extends` it.

frontend/src/community/crm/types/CrmCommonTypes.ts
```diff
@@ -0,0 +1,230 @@
...
+  CrmDealSortEnum,
+  CrmDealStageColorsEnum,
+  CrmDealStageEnum,
+  CrmIndustryEnum,
+  CrmPriorityEnum
+} from "../enums/common";
+
+// Company
+
+export interface CompanyEntity {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`id` is optional on every `*Entity` type (`CompanyEntity:14`, `ContactEntity:53`, `DealEntity:101`, `DealStageEntity:141`, `TaskEntity:157`). These shapes clearly exist to back a normalized, id-keyed store (they carry `contactIds`/`dealIds`/`taskIds`/`stageId`/`ownerId` instead of nested objects), and the id is the one field that is always present on a server-returned entity and is required to key it. Making it optional forces every consumer into `entity.id!` non-null assertions or defensive `if (!entity.id) return` branches. Fix: make `id: number` required on all five entity interfaces, and use a separate `Partial<XEntity>` (or a dedicated `XUpdatePayload`) for the patch/update case.

frontend/src/community/crm/types/CrmCommonTypes.ts
```diff
@@ -0,0 +1,230 @@
...
+  id: number;
+  name: string;
+}
+
+// Board
+
+export interface BoardColumn {
+  dealIds: number[];
+  totalCount: number;
+  page: number;
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`BoardColumn` drops pagination fields the board API actually returns and renames one. The backend `CrmDealsByStageResponseDto` returns `stageId`, `totalCount`, `currentPage`, `totalPages`, `pageSize`, `hasNextPage`, and the existing `CrmBoardStageDealsType` in `BoardTypes.ts:43` mirrors all of them. Here the field is `page` (not `currentPage`) and `totalPages`/`pageSize`/`stageId` are absent. If this is intended as purely client-side column state derived from the response, it's fine — but the field name should still track the API (`currentPage`) to avoid a silent mapping bug, and dropping `totalPages` means the existing `getNextPageParam`-style logic (`currentPage + 1 >= totalPages`) can't be expressed against this shape. Please confirm this is deliberate.

frontend/src/community/crm/types/CrmCommonTypes.ts
```diff
@@ -0,0 +1,230 @@
...
+
+// Deal
+
+export interface DealEntity {
+  id?: number;
+  name?: string;
+  description?: string;
+  priority?: CrmPriorityEnum;
+  orderIndex?: string;
+  amount?: string;
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`DealEntity` omits `currencyCode`, which the existing `CrmDealType` in `CommonTypes.ts:178` carries alongside `amount`. If deal amounts are ever rendered with a currency symbol, a normalized `DealEntity` with only `amount` won't be able to reproduce that without an extra lookup. Worth confirming whether the omission is intentional (e.g. currency now comes from a global org setting) before this shape replaces `CrmDealType`.

frontend/src/community/crm/v2/types/CrmCommonTypes.ts
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Dead code: this entire 230-line file is unreferenced. Grepping the whole `frontend/` tree (community + enterprise) for `CrmCommonTypes` and for each exported symbol (`CompanyEntity`, `ContactEntity`, `DealEntity`, `TaskEntity`, `DealStageEntity`, `BoardColumn`, `DealMoveResult`, `CompanyMetrics`, `ContactMetrics`) returns hits only inside this file itself. Consequence: `tsc` never checks these declarations against a real API response or payload, so every contract mismatch below is invisible to CI and will only surface when a future PR wires them up. Fix: either migrate at least one consumer in this PR (e.g. point `BoardApi.ts` / `crmDealApi.ts` / `ContactApi.ts` at the new types and delete the superseded declarations in `CommonTypes.ts`/`BoardTypes.ts`), or hold this file back and land it together with the consumer change. Shipping unreferenced parallel type definitions leaves the module with two competing sources of truth and no compiler enforcing either.

frontend/src/community/crm/types/CrmCommonTypes.ts
```diff
@@ -0,0 +1,230 @@
...
+
+export interface ContactEntity {
+  id?: number;
+  firstName?: string;
+  lastName?: string;
+  email?: string;
+  contactNumber?: string;
+  lastContactAt?: string;
+  lastModifiedDate?: string;
+  companyId?: number;
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The `*Entity` types conflate request-payload shape with response shape, so they describe neither correctly. Every CRM response DTO returns nested objects, never flattened FK scalars: `CrmContactResponseDto`/`CrmContactListItemDto`/`CrmContactDetailResponseDto` return `company: {id,name}` and `owner: {employeeId, firstName, lastName, authPic}` (no `companyId`/`ownerId`); `CrmTaskResponseDto`/`CrmTaskDetailResponseDto`/`CrmTaskViewResponseDto` return `owner`, `contact`, `deal`, `typeName` (no `ownerId`/`contactId`/`dealId`/`typeId`); `CrmDealResponseDto` returns `stage`, `owner`, `companyName`, `contactName` (no `stageId`/`ownerId`). The scalar id fields used here exist only on the create/edit request DTOs. Two fields match nothing at all: `TaskEntity.companyId` (line 166) appears on no task DTO in either direction — only on `CrmTaskFilterDto`, which is a query filter — and `DealEntity.closingAt` (line 107) exists only on `CrmDealCreateRequestDto` and the JPA model, never on a response. Also `TaskEntity` has no `typeName`, which every task response carries. Fix: split each concept into a response interface mirroring the DTO (nested `owner`/`contact`/`company`/`stage` objects, `typeName`) and a separate create/update payload interface holding the scalar ids, the way `CommonTypes.ts` already does with `CrmTaskDetailType` vs `CrmTaskCreatePayload`.

frontend/src/community/crm/types/CrmCommonTypes.ts
```diff
@@ -0,0 +1,230 @@
...
+// Board
+
+export interface BoardColumn {
+  dealIds: number[];
+  totalCount: number;
+  page: number;
+  hasNextPage: boolean;
+}
+
+export interface DealMoveResult {
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`DealMoveResult` matches neither side of the board move contract. `PATCH /v1/crm/board/deal-move-between-stages` accepts `CrmDealUpdateStageRequestDto { dealId, newStageId, previousDealId, nextDealId }` and returns a full `CrmDealResponseDto` (see `CrmDealServiceImpl.updateDealStage`). Three concrete problems: (1) line 132 names the field `stageId`, not `newStageId` — Jackson ignores the unknown property and `newStageId` arrives null, which the service rejects with `CRM_ERROR_DEAL_STAGE_ID_REQUIRED`, so the drag silently rolls back; (2) line 133 adds a client-supplied `orderIndex`, but the server computes it itself via `generateOrderIndex(...)` from the neighbour ids and ignores anything sent; (3) lines 134-135 type the neighbours as `number | undefined`, while `computeMoveNeighbors`/`computeReorderWithinStage` in `utils/kanbanUtil.ts` return `number | null` — the values actually flowing through `useKanbanDrag`. Fix: drop this type and reuse the correct `CrmBoardMoveBetweenStagesPayload` / `CrmBoardDealNeighboursType` from `types/BoardTypes.ts`, which already model the endpoint accurately.

frontend/src/community/crm/types/CrmCommonTypes.ts
```diff
@@ -0,0 +1,230 @@
...
+  website?: string;
+  address?: string;
+  contactNumber?: string;
+  openTasksCount?: number;
+  overdue?: number;
+  openValue?: string;
+  accountValue?: string;
+  openDeals?: number;
+  closedDeals?: number;
+  contactIds?: number[];
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`contactIds`, `dealIds` and `taskIds` (CompanyEntity lines 26-28, ContactEntity lines 69-70, DealEntity line 113) do not exist in any backend payload. A grep of the entire `crmplanner` package finds these identifiers only as local variables in repository/service implementations — no request or response DTO exposes them. What the backend actually returns for children is nested full objects: `CrmContactDetailResponseDto` carries `deals: CrmDealDetailResponseDto[]` and `tasks: CrmTaskDetailResponseDto[]`, and `CrmCompanyMetricsResponseDto` carries no child collections at all. Because these are optional, any consumer writing `company.contactIds?.length` or `deal.taskIds?.map(...)` compiles cleanly and silently reads `undefined`/0 forever — the worst kind of failure since there is no error to trace. Fix: delete the three id-array fields and model the real nested arrays (`contacts`, `deals`, `tasks`) on the detail types only, matching `CrmCompany`/`CrmContact` in `CommonTypes.ts`.

frontend/src/community/crm/types/CrmCommonTypes.ts
```diff
@@ -0,0 +1,230 @@
...
+
+export interface DealFilterRequest {
+  sortOrder?: SortOrderTypes;
+  sortKey?: CrmDealSortEnum;
+  searchKeyword?: string;
+  stageId?: number;
+  priority?: CrmPriorityEnum;
+  companyId?: number;
+  contactId?: number;
+  page: number;
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`DealFilterRequest.page` being required will break the existing deal infinite query the moment this type is adopted. `useGetDealsInfinite` in `api/crmDealApi.ts:43-48` builds its request as `params: { page: pageParam, ...params }` — the spread comes *after* `page`, so any `page` present on the filter object overrides the cursor. Today that is safe only because the existing `CrmDealFilterParams` (`CommonTypes.ts:305`) deliberately omits `page`; swapping in `DealFilterRequest` forces every caller to supply one, which then wins over `pageParam` and makes every scroll re-fetch the same page. Fix: make `page` optional here (the backend `CrmDealFilterDto` defaults it to 0 anyway), or keep the filter and pagination concerns in separate types as the current code does. Separately, hardening `useGetDealsInfinite` to `{ ...params, page: pageParam }` would remove the trap regardless.

frontend/src/community/crm/types/CrmCommonTypes.ts
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The edit-payload fields lose the tri-state semantics the backend depends on. `CrmContactEditRequestDto.companyId` is `JsonNullable<Long>` and `CrmTaskEditRequestDto.contactId`/`dealId`/`notes` are `JsonNullable<...>` initialised to `JsonNullable.undefined()`: an absent key means "leave unchanged", an explicit JSON `null` means "clear the association". `ContactEntity.companyId?: number` (line 60), `TaskEntity.contactId?: number` (165), `dealId?: number` (167) and `notes?: string` (162) can only express "absent", so "unassign this contact's company" or "detach this task from its deal" becomes untypable — a real product capability the UI already has. This is the request-side counterpart of the response-side nullability issue, and it is a capability loss rather than a modelling preference. The existing types get it right (`EditContactPayload.companyId?: number | null` at CommonTypes.ts:159, `CrmTaskUpdatePayload.contactId?: number | null` at CommonTypes.ts:383). Fix: use `?: number | null` on every field the backend declares as `JsonNullable`.

frontend/src/community/crm/types/CrmCommonTypes.ts
```diff
@@ -0,0 +1,230 @@
...
+  size: number;
+}
+
+export interface DealsByStagesRequest {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`DealsByStagesRequest` duplicates `CrmBoardDealsGroupedRequest` in `types/BoardTypes.ts:65` field for field (`stageIds`, `searchKeyword?`, `page`, `limit`). This is a second, independent collision from the `CommonTypes.ts` overlap already noted — and the existing type is the one actually in use: it is the parameter type of `fetchDealsGroupedByStages`, `useGetDealsGroupedByStages` and `useFetchMoreStageDeals` in `api/BoardApi.ts`. Two identical declarations of the same board request in two type files in the same directory guarantees they drift. Fix: delete `DealsByStagesRequest` and import `CrmBoardDealsGroupedRequest`, or if this file is meant to replace `BoardTypes.ts`, remove the old one in the same PR rather than leaving both.

frontend/src/community/crm/types/CrmCommonTypes.ts
```diff
@@ -0,0 +1,230 @@
...
+  openDeals?: number;
+  closedDeals?: number;
+}
+
+export interface CompanyLookup {
+  id: number;
+  name: string;
+}
+
+export interface CompanyDomainSearchResponseType {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`CompanyDomainSearchResponseType` over-widens the domain-search result. `GET /v1/crm/company/search-by-domain` returns `CrmCompanyDomainSearchResponseDto { companies: List<CrmCompanyResponseDto> }`, and `CrmCompanyResponseDto` has exactly six fields: id, name, industry, website, address, contactNumber. Typing the array as `CompanyEntity[]` advertises `openTasksCount`, `overdue`, `openValue`, `accountValue`, `openDeals`, `closedDeals` and the id arrays, none of which the endpoint ever sends — so a consumer can format `company.accountValue` in the domain-search dropdown, compile fine, and render blank. The existing `CrmCompanyDomainSearchResponseType` (CommonTypes.ts:134) correctly narrows this to `CrmCompanyType[]`. Fix: point `companies` at a narrow company-summary interface holding only the six fields the endpoint returns.

frontend/src/community/crm/types/CrmCommonTypes.ts
```diff
@@ -0,0 +1,230 @@
...
+}
+
+export interface TaskFilterRequest {
+  searchKeyword?: string;
+  contactId?: number;
+  dealId?: number;
+  companyId?: number;
+}
+
+export interface TaskCompletedFilterRequest {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`TaskCompletedFilterRequest` re-lists all four fields of `TaskFilterRequest` instead of extending it. The backend models the exact same relationship as inheritance — `CrmTaskCompletedFilterDto extends CrmTaskFilterDto` and adds only `page`/`size` — so mirroring that here keeps the two in lockstep; as written, adding a field to `TaskFilterRequest` (line 209) silently leaves the completed-task filter behind. There is precedent for the pattern in this module (`EditCompanyPayload extends CrmCompanyCreatePayload`, `CrmDealStageUpdatePayload extends Partial<CrmDealStageCreatePayload>`). Fix: `export interface TaskCompletedFilterRequest extends TaskFilterRequest { page: number; size: number }`, and consider deriving `TaskRelatedFilterRequest` from a `Pick<TaskFilterRequest, "contactId" | "dealId">` for the same reason.

frontend/src/community/crm/types/CrmCommonTypes.ts
```diff
@@ -0,0 +1,230 @@
...
+
+export interface TaskType {
+  id: number;
+  name: string;
+  orderIndex: number;
+}
+
+// Filters
+
+export interface CompanyDomainSearchFilterRequest {
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

The Filters section is incomplete relative to the endpoints the module actually calls. It covers domain search, contact lookup, deal list, deals-by-stage and the three task filters, but omits `CrmCompanyFilterDto` (`GET /company/lookup`), `CrmCompanyMetricRequestDto` (`GET /company/metrics`), `CrmContactMetricRequestDto` (`GET /contact/metrics`) and `CrmContactOwnerFilterDto` (`GET /contact/owners`). Those four are currently typed with ad-hoc local interfaces in the API layer (`ContactMetricsSearchParams` at `api/ContactApi.ts:32`, `CompanyMetricSearchParams` in `api/CompanyApi.ts`) or with untyped inline object literals (`useGetCrmContacts`, `useGetOwnerLookup`, `useGetCompanyLookup`). If this file is intended as the CRM contract source of truth, adding them lets the API layer drop those one-off shapes; if it is not, the partial coverage should be called out so the next contributor knows which half is authoritative.

frontend/src/community/crm/types/CrmCommonTypes.ts
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`Owner` drops `email`, which the owner lookup endpoint returns. `GET /v1/crm/contact/owners` and the contact list/response payloads serve `CrmContactOwnerResponseDto { employeeId, firstName, lastName, email, authPic }`, while the deal/task/detail payloads use the narrower `CrmOwnerResponseDto` (no email). Modelling both with a single `Owner` that omits `email` means the owner picker can never show or filter on an owner's email without an `as` cast. The pre-existing `CrmOwner` has the same gap, so this is an opportunity to fix it rather than re-copy it. Fix: either add `email?: string` to `Owner`, or declare two types (e.g. a lookup owner with email and a compact owner without) matching the two backend DTOs.

frontend/src/community/crm/v2/types/CrmCommonTypes.ts
YasinduRC: old types will be  deleted in the migration PRs

YasinduRC: These will used when we migrate these changes to component level

frontend/src/community/crm/types/CrmCommonTypes.ts
YasinduRC: We went with optional everywhere, same as people module. Nulls get cleaned up when the response is normalized, so they never reach the store.

frontend/src/community/crm/types/CrmCommonTypes.ts
YasinduRC: The nested deals/tasks on the contact get-by-id are being removed — those will come from their own endpoints filtered by contactId, same way the company side panel already works. The store keeps just the id arrays, which is why they're defined here

frontend/src/community/crm/types/CrmCommonTypes.ts
YasinduRC: the old CompanyLookup goes away in the migration PR, so only one will be left.

frontend/src/community/crm/types/CrmCommonTypes.ts
YasinduRC: We will add a metrics endpoint and this is the return type of it,

frontend/src/community/crm/types/CrmCommonTypes.ts
YasinduRC: everything's optional, so a partly filled company is still a valid CompanyEntity. Missing field just means not loaded yet.

frontend/src/community/crm/types/CrmCommonTypes.ts
YasinduRC: Known , the name split is coming from the backend in a separate PR. This file is the target shape, not what the API returns today.

frontend/src/community/crm/types/CrmCommonTypes.ts
YasinduRC: Backend is changing this one  it's the last BigDecimal left, everything else is already String. Separate PR.

frontend/src/community/crm/types/CrmCommonTypes.ts
YasinduRC: The small nested bits (company, owner) still come nested and we flatten them to ids when normalizing into the store. The bigger nested lists on contact get-by-id are going away — deals and tasks move to their own endpoints, so the contact just holds id arrays.

frontend/src/community/crm/types/CrmCommonTypes.ts
YasinduRC: Dropped it on purpose  it isn't used anywhere

frontend/src/community/crm/types/CrmCommonTypes.ts
YasinduRC: BoardTypes.ts gets replaced by this file in the migration, not kept alongside.

frontend/src/community/crm/types/CrmCommonTypes.ts
YasinduRC: Should have the discussion with the team

frontend/src/community/crm/types/CrmCommonTypes.ts
YasinduRC: Added crm prefix

YasinduRC: Added email to owner

frontend/src/community/crm/types/CrmCommonTypes.ts
YasinduRC: prefer keeping these flat since it is easier to read when each request type shows exactly what the endpoint takes. Will have to discuss with the team

frontend/src/community/crm/types/CrmCommonTypes.ts
YasinduRC: Made all pagination fields optional, the cursor stays with useInfiniteQuery and never sits in the filter.

frontend/src/community/crm/types/CrmCommonTypes.ts
YasinduRC: Added the missing ones

frontend/src/community/crm/types/CrmCommonTypes.ts
YasinduRC: Fixed , dropped CrmDealMoveResult and added CrmDealReorderWithinStageRequest and CrmDealMoveBetweenStagesRequest, one per endpoint

frontend/src/community/crm/types/CrmCommonTypes.ts
YasinduRC: Renamed page → currentPage to match the API

akilarootcode (approved):

akilarootcode (approved):

thusala (approved):

frontend/src/community/crm/types/CrmCommonTypes.ts
YasinduRC: We're keeping the types fully optional though, so instead of adding | null we'll build the edit payload from the Formik diff, compare initialValues against values, omit anything untouched, and send null for fields the user actually cleared.

PR #1729, opened 2026-08-10
Add Manual Time Entry restriction Toggle and disable manual time entry

erandi-rootcode:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds an organization-level "Restrict Manual Time Entries & Edits" setting (community config plumbing + enterprise-overridable enforcement hook + community UI gating). The config plumbing, DTO/enum extension and test updates are consistent with the existing `GEO_FENCING_ENABLED`/`FINGERPRINT_ATTENDANCE_ENABLED` patterns, but the enforcement layer has real gaps: the new error message constant is silently discarded by `GlobalExceptionHandler`, two other write endpoints that accept client-supplied timestamps are left ungated, `getEmployeeRole()` is dereferenced without the codebase's usual null check, and the admin-facing copy promises approval-free saves that the code does not implement. Frontend gating works but is permissive while the config query is loading and leaves the restricted table row keyboard-focusable with an interactive hover state.

Found **21** new issue(s): 🔴 6 important, 🟡 9 suggestion(s), 🟣 6 nit(s)

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeServiceImpl.java
```diff
@@ -2181,6 +2186,39 @@ protected boolean isGeoFencingEnabled() {
...
+			return;
+		}
+
+		EmployeeRole employeeRole = currentUser.getEmployee().getEmployeeRole();
+		boolean isAuthorized = Boolean.TRUE.equals(employeeRole.getIsSuperAdmin())
+				|| Role.ATTENDANCE_ADMIN.equals(employeeRole.getAttendanceRole())
+				|| Role.ATTENDANCE_MANAGER.equals(employeeRole.getAttendanceRole());
+
+		if (!isAuthorized) {
+			throw new AccessDeniedException(
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The new `TIME_ERROR_MANUAL_ENTRY_RESTRICTED` constant and its `time-messages.properties` entry are effectively dead code. `GlobalExceptionHandler.handleAccessDeniedException` (GlobalExceptionHandler.java:55-63) ignores `e.getMessage()` entirely and always responds with `COMMON_ERROR_ACCESS_DENIED` ("You don't have permission to access this resource") plus `messageKey = COMMON_ERROR_ACCESS_DENIED`. So the user never sees why the action was blocked, and the frontend cannot key off the message key to render a specific toast. This also deviates from the established convention for role-based denials in the service layer — see `TimeAnalyticsServiceImpl.java:229`, which throws `new ModuleException(TimeMessageConstant.TIME_ERROR_MANAGER_OR_ABOVE_PERMISSIONS_REQUIRED)`. Fix: throw `new ModuleException(TimeMessageConstant.TIME_ERROR_MANUAL_ENTRY_RESTRICTED)` and drop the now-unneeded `messageUtil.getMessage(...)` call and the `AccessDeniedException` import. If 403 (rather than ModuleException's 400) is required, add a message-key-preserving exception type + handler instead of reusing Spring Security's `AccessDeniedException`.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeServiceImpl.java
```diff
@@ -603,6 +606,8 @@ public ResponseEntityDto addManualEntryRequest(ManualEntryRequestDto timeRequest
 		User currentUser = userService.getCurrentUser();
 		log.info("addManualEntryRequest: execution started");
 
+		validateManualEntryRestriction(currentUser);
```
ThinuwanW, line 609: 🤖 **Claude** · 🔴 **Important**

Enforcement is incomplete: `validateManualEntryRestriction` is only wired into `addManualEntryRequest` (POST /v1/time/manual-entry) and `editTimeRequest` (PATCH /v1/time/request), but two other `ROLE_ATTENDANCE_EMPLOYEE` endpoints let an employee write arbitrary time data. (1) `addTimeRecord` (POST /v1/time/record, TimeServiceImpl.java:523) takes `addTimeRecordDto.getTime()` straight from the client and only rejects future dates; with `recordActionType = START` a restricted employee can create a brand-new `TimeRecord` for any past date with no supervisor approval at all (`recordClockInAndClockOut`, line 1973), which is exactly what the setting is meant to prevent. (2) `updateCurrentUserIncompleteTimeRecords` (PATCH /v1/time/incomplete-clockouts/{id}, line 674) accepts an arbitrary `requestDto.getClockOutTime()`. Since clock in/out must stay allowed, the fix is not a blanket block: when `isManualEntryRestrictionEnabled()` is true and the caller is not authorized, constrain `addTimeRecord` to `timeRecordDate.equals(currentDate)` (and reject timestamps materially in the past), and apply the same restriction to the incomplete-clock-out path.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeServiceImpl.java
```diff
@@ -2181,6 +2186,39 @@ protected boolean isGeoFencingEnabled() {
...
+	 * entry restriction is enabled. Only attendance managers, attendance admins and super
+	 * admins may add or change time records once the setting is on.
+	 * @param currentUser the user submitting the manual entry or edit request.
+	 */
+	private void validateManualEntryRestriction(User currentUser) {
+		if (!isManualEntryRestrictionEnabled()) {
+			return;
+		}
+
+		EmployeeRole employeeRole = currentUser.getEmployee().getEmployeeRole();
```
ThinuwanW, line 2198: 🤖 **Claude** · 🔴 **Important**

`currentUser.getEmployee().getEmployeeRole()` is dereferenced without a null check. `Employee.employeeRole` is an optional `@OneToOne(mappedBy = "employee")` association (Employee.java:133-135), and the codebase consistently guards it — see `CommonModuleUtils.isGuestEmployee` (line 356), `RolesServiceImpl.java:346`, `EmployeeValidationServiceImpl.java:219`, `PeopleServiceImpl.java:382`. For an employee row without an `employee_role` record this throws NPE and, via the catch-all `@ExceptionHandler(Exception.class)`, surfaces as a 500 instead of a clean denial. Guard it, e.g. `EmployeeRole employeeRole = currentUser.getEmployee() != null ? currentUser.getEmployee().getEmployeeRole() : null; boolean isAuthorized = employeeRole != null && (...);` so a missing role falls through to "not authorized" rather than a server error.

frontend/src/community/common/assets/languages/english/attendanceModule.json
```diff
@@ -22,6 +22,31 @@
     "configureWorkLocationsLink": "Configure work locations",
     "proTier": "Pro",
     "isClockInClockOutOnly": "Clock in / clock out only (no breaks)",
+    "manualEntryRestrictionTitle": "Restrict Manual Time Entries & Edits",
+    "manualEntryRestrictionDescription": "Control who is allowed to add or change time records in your organization.",
+    "isManualEntryRestrictionEnabled": "Restrict manual time entries & edits",
+    "manualEntryRestrictionTooltip": "When enabled, only Attendance Managers, Attendance Admins, and Super Admins can add or edit time entries. Their changes save immediately without approval.",
```
ThinuwanW, line 29: 🤖 **Claude** · 🔴 **Important**

The tooltip and the modal `enableDescription` (line 36) both promise "their changes save immediately without approval" for Attendance Managers/Admins/Super Admins, but nothing in this PR implements that. `addManualEntryRequest` still builds a PENDING `TimeRequest` and only auto-approves when the separate `AUTO_APPROVAL_FOR_CHANGES` config is on (TimeServiceImpl.java:625-629), and `EpTimeServiceImpl` does not override that path. Worse, `addManualEntryRequest` throws `PEOPLE_ERROR_NO_MANAGERS_FOUND` up front when the caller has no assigned supervisor (TimeServiceImpl.java:611) — so with the restriction on, an Attendance Admin or Super Admin who has no supervisor cannot create any time entry, leaving nobody able to fix employees' timesheets. Either implement the bypass (auto-approve, and skip the manager-exists precondition, when the restriction is on and the caller is authorized) or reword these strings to match actual behaviour.

frontend/src/community/attendance/components/organisms/AttendanceConfiguration/AttendanceConfiguration.tsx
```diff
@@ -175,6 +176,11 @@ const AttendanceConfiguration = (): JSX.Element => {
           onSwitchChange={handleSwitchChange}
         />
 
+        <ManualEntryRestrictionSettings
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`ManualEntryRestrictionSettings` is dropped into this form but saves independently (its own `useUpdateAttendanceConfiguration` with `{...initialConfig, isManualEntryRestrictionEnabled}`), unlike `GeoFencingSettings`/`FingerprintSettings` which route through `onSwitchChange` and the shared Save button. Two consequences: (a) its mutation invalidates `getAttendanceConfiguration`, the refetch updates `configData`, and the effect at line 59-64 overwrites both `config` and `initialConfig` — silently discarding any unsaved edits the admin has made to the other switches on the page; (b) the shared Save button posts the whole `config`, including `isManualEntryRestrictionEnabled: true`, so on a tenant that dropped below Core the enterprise tier guard (`EpAttendanceConfigServiceImpl.validateManualEntryRestrictionTier`) throws and every unrelated save on this page fails with the generic error toast. Consider making this control use `onSwitchChange` like its siblings (confirm modal gating the local state change), or have the parent effect only seed state when `initialConfig === null`, and omit the flag from the payload when the tier does not support it.

frontend/src/community/attendance/hooks/useManualEntryRestriction.ts
```diff
@@ -0,0 +1,27 @@
...
+const useManualEntryRestriction = () => {
+  const { data: attendanceConfig } = useGetAttendanceConfiguration();
+  const { isSuperAdmin, isAttendanceAdmin, isAttendanceManager } =
+    useSessionData();
+
+  const canManageTimeEntries = Boolean(
+    isSuperAdmin || isAttendanceAdmin || isAttendanceManager
+  );
+
+  const isManualEntryRestricted =
```
ThinuwanW, line 26: 🤖 **Claude** · 🟡 **Suggestion**

The hook fails open while the config query is in flight: `attendanceConfig` is `undefined` on first render, so `isManualEntryRestricted` is `false` and the "Add Time Entry" button plus clickable rows render, then disappear once the response lands — a restricted employee can click through in that window (the request is rejected server-side, but with the generic "access denied" toast). Return the query's `isLoading`/`isPending` alongside the flag so callers can hold off rendering the action, or treat unknown as restricted until the config resolves.

frontend/src/community/attendance/components/molecules/TimesheetDailyRecordTableRow/TimesheetDailyRecordTableRow.tsx
```diff
@@ -196,17 +198,25 @@ const TimesheetDailyRecordTableRow: FC<Props> = ({ record, headerLength }) => {
...
       justifyContent="space-between"
       alignItems="center"
-      sx={classes.stackContainerStyle}
-      onClick={() => mutate()}
+      sx={mergeSx([
+        classes.stackContainerStyle,
+        isManualEntryRestricted ? { cursor: "default" } : undefined
+      ])}
+      onClick={handleRowActivate}
       tabIndex={getTabIndex(isFreeTier)}
```
ThinuwanW, line 216: 🤖 **Claude** · 🟡 **Suggestion**

When restricted, the row is made a no-op but still advertises itself as interactive: `tabIndex` stays `getTabIndex(isFreeTier)` (0 for paid tiers) so keyboard users can focus it and press Enter with no feedback, and `stackContainerStyle` keeps its `&:hover` background change (styles.ts:36-41) — only `cursor` is overridden. Add `aria-disabled={isManualEntryRestricted}`, pass `tabIndex={isManualEntryRestricted ? -1 : getTabIndex(isFreeTier)}`, and include `"&:hover": { background: "inherit" }` in the restricted `sx` override. A short translated tooltip explaining why editing is unavailable would also avoid a silent dead click (this module already uses `translateAria` for exactly this kind of affordance).

frontend/src/community/attendance/components/molecules/TimesheetDailyRecordTableRow/TimesheetDailyRecordTableRow.tsx
```diff
@@ -49,6 +50,7 @@ interface Props {
 
 const TimesheetDailyRecordTableRow: FC<Props> = ({ record, headerLength }) => {
   const { isFreeTier } = useSessionData();
+  const { isManualEntryRestricted } = useManualEntryRestriction();
```
ThinuwanW, line 53: 🤖 **Claude** · 🟡 **Suggestion**

`useManualEntryRestriction` (and therefore `useGetAttendanceConfiguration`) is instantiated once per rendered row. React Query dedupes the in-flight request, but `useGetAttendanceConfiguration` sets no `staleTime` (AttendanceAdminApi.ts:13-24), so every mount/refocus of the table triggers a refetch of an org-level setting that changes almost never, and each row adds a query subscription. Prefer resolving the flag once in `TimesheetDailyRecordTable` (or `EmployeeTimesheet`) and passing it down one level, and/or add a `staleTime` to `useGetAttendanceConfiguration`.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeServiceImpl.java
```diff
@@ -30,6 +30,7 @@
 import com.skapp.community.peopleplanner.mapper.PeopleMapper;
 import com.skapp.community.peopleplanner.model.Employee;
 import com.skapp.community.peopleplanner.model.EmployeeManager;
+import com.skapp.community.peopleplanner.model.EmployeeRole;
```
ThinuwanW, line 33: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

No test covers the new authorization logic. `AttendanceConfigServiceImplUnitTest` only exercises the config read/write plumbing; `validateManualEntryRestriction` has zero coverage of its three authorized branches (super admin / attendance admin / attendance manager) or the denial branch — and since the community `isManualEntryRestrictionEnabled()` returns `false`, the branch is unreachable in community tests unless you stub the hook (a Mockito `spy` with `doReturn(true).when(spy).isManualEntryRestrictionEnabled()`, or a small test subclass, mirroring how `isGeoFencingEnabled` could be handled). Similarly on the frontend: no `useManualEntryRestriction` test, and `TimesheetDailyRecordTableRow.test.tsx` was not updated to assert the row is inert when restricted.

frontend/src/fallback/configurations/components/organisms/ManualEntryRestrictionSettings/ManualEntryRestrictionSettings.tsx
```diff
@@ -0,0 +1,12 @@
+import { AttendanceConfigurationType } from "~community/attendance/types/attendanceTypes";
+
+interface Props {
+  config: AttendanceConfigurationType | null;
+  initialConfig: AttendanceConfigurationType | null;
+}
+
+const ManualEntryRestrictionSettings = ({ config, initialConfig }: Props) => {
```
ThinuwanW, line 8: 🤖 **Claude** · 🟣 **Nit**

Both props are destructured but unused, which trips `unused-imports/no-unused-vars` (warn). The sibling fallback `FingerprintSettings` avoids this cleanly with `const FingerprintSettings: FC<Props> = () => <></>;`. Also add the explicit return type the enterprise counterpart uses (`(): JSX.Element`) — TS strict mode plus the project convention of annotating component return types (see the enterprise `ManualEntryRestrictionSettings` and `AttendanceConfiguration`).

frontend/pages/community/timesheet/my-timesheet.tsx
```diff
@@ -24,7 +26,11 @@ const MyTimeSheet: NextPage = () => {
       ]}
       pageHead={translateText(["timesheet.myTimesheet.pageHead"])}
       title={translateText(["timesheet.myTimesheet.title"])}
-      primaryButtonText={translateText(["timesheet.manualTimeEntryButtonTxt"])}
+      primaryButtonText={
```
ThinuwanW, line 29: 🤖 **Claude** · 🟣 **Nit**

Hiding the button by passing `""` relies on `ContentLayout`'s `{primaryButtonText && (...)}` falsiness check (ContentLayout.tsx:362) — it works, but it reads as "render a button with no label". `undefined` states the intent directly and matches the optional prop type (`primaryButtonText?: string | boolean`).

backend/src/main/java/com/skapp/community/timeplanner/service/impl/AttendanceConfigServiceImpl.java
```diff
@@ -104,17 +109,19 @@ public ResponseEntityDto getAllAttendanceConfigs() {
...
+				isGeoFencingEnabled, isClockInClockOutOnly, null, isManualEntryRestrictionEnabled);
 
 		log.info("getAllAttendanceConfigs: execution ended");
 
 		return new ResponseEntityDto(false, attendanceConfigRequestDto);
 	}
 
 	private static AttendanceConfigRequestDto getAttendanceConfigRequestDto(List<AttendanceConfig> attendanceConfigs) {
-		AttendanceConfigRequestDto dto = new AttendanceConfigRequestDto(false, false, false, false, false, false,
+		AttendanceConfigRequestDto dto = new AttendanceConfigRequestDto(false, false, false, false, false, false, false,
```
ThinuwanW, line 130: 🤖 **Claude** · 🟣 **Nit**

`AttendanceConfigRequestDto` is now built from eight positional `Boolean` arguments here and at line 115, plus three test sites. Every new flag forces an edit at all five call sites, and a mis-ordered argument compiles silently because all parameters share the same type (the diff already had to reflow these calls). Consider adding `@Builder`/`@NoArgsConstructor` to the DTO and constructing via named setters or the builder, so the field-to-value mapping is explicit.

frontend/src/community/attendance/hooks/useManualEntryRestriction.ts
ThinuwanW, line 26: 🤖 **Claude** · 🔴 **Important**

The hook enforces the restriction purely from the stored config value, but the backend only enforces it when the tenant is on CORE/PRO: `EpTimeServiceImpl.isManualEntryRestrictionEnabled()` returns false for tiers below Core, and the community `TimeServiceImpl.isManualEntryRestrictionEnabled()` always returns false. `getAllAttendanceConfigs` returns the raw DB value with no tier/edition filtering. So for a tenant that enabled the setting on Core and then downgraded (or a community deployment whose DB carries the flag), the server happily accepts manual entries/edits while the UI hides the Add Time Entry button and makes every timesheet row inert. Worse, there is no way to recover from the UI: the community fallback `ManualEntryRestrictionSettings` renders nothing, and the enterprise component disables the switch when `!isAtLeastCoreTier`, so the flag cannot be turned back off. Mirror the server guard in the hook, e.g. `const { isAtLeastCoreTier } = useTier();` and `const isManualEntryRestricted = isAtLeastCoreTier && Boolean(attendanceConfig?.isManualEntryRestrictionEnabled) && !canManageTimeEntries;` — the `src/fallback` `useTier` returns `isAtLeastCoreTier: false`, which exactly matches the community backend behaviour.

frontend/src/community/attendance/components/organisms/AttendanceConfiguration/AttendanceConfiguration.tsx
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`handleSaveBtnClick` sends the whole `config` object, and `config` is seeded from the fetched configuration, which now includes `isManualEntryRestrictionEnabled`. `EpAttendanceConfigServiceImpl.updateAttendanceConfig` runs `validateManualEntryRestrictionTier(...)` and throws `EP_TIME_ERROR_MANUAL_ENTRY_RESTRICTION_NOT_AVAILABLE_FOR_CURRENT_TIER` whenever the payload carries `true` and the tenant is below Core/Pro. Concrete failure: a tenant enables the restriction on Core, then downgrades to Free; the attendance admin flips an unrelated clock-in switch and presses Save — the PATCH still carries `isManualEntryRestrictionEnabled: true`, the tier check throws, and every attendance configuration save from then on fails with the generic "Failed to update the attendance configuration" toast. The admin also cannot clear the flag through the dedicated toggle because that switch is `disabled={!isAtLeastCoreTier}`. Since `ManualEntryRestrictionSettings` owns this field and saves it itself, exclude it from the main form payload, e.g. build the payload in `handleSaveBtnClick` by destructuring the field out (`const { isManualEntryRestrictionEnabled: _ignored, ...payload } = config; updateConfig(payload);`).

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeServiceImpl.java
```diff
@@ -2181,6 +2186,39 @@ protected boolean isGeoFencingEnabled() {
...
+		return false;
+	}
+
+	/**
+	 * Rejects manual time entry creation and edits made by employees while the manual
+	 * entry restriction is enabled. Only attendance managers, attendance admins and super
+	 * admins may add or change time records once the setting is on.
+	 * @param currentUser the user submitting the manual entry or edit request.
+	 */
+	private void validateManualEntryRestriction(User currentUser) {
```
ThinuwanW, line 2193: 🤖 **Claude** · 🟡 **Suggestion**

`validateManualEntryRestriction` is `private` while every other extension point around it (`isManualEntryRestrictionEnabled`, `isGeoFencingEnabled`, `populateEnterpriseChipFields`, `createTimeRecordChipDto`) is `protected`. The subclass that actually turns this feature on, `EpTimeServiceImpl`, therefore cannot invoke the guard from its own write paths (e.g. `addTimeRecordWithLocation`) or from any future enterprise override of `editTimeRequest`/`addManualEntryRequest`. Make it `protected` so the enterprise module can reuse the single implementation instead of re-deriving the role check.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeServiceImpl.java
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The denial path throws without logging anything. The only trace left is `GlobalExceptionHandler.handleAccessDeniedException`, which logs the generic `COMMON_ERROR_ACCESS_DENIED` key and never mentions which employee was blocked or that the manual-entry restriction was the cause — making support triage of "I can't add a time entry" reports guesswork. Add a `log.warn("validateManualEntryRestriction: manual entry blocked for employee {}", currentUser.getEmployee().getEmployeeId());` immediately before the throw, consistent with the module's logging conventions.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeServiceImpl.java
```diff
@@ -2181,6 +2186,39 @@ protected boolean isGeoFencingEnabled() {
...
+		if (!isManualEntryRestrictionEnabled()) {
+			return;
+		}
+
+		EmployeeRole employeeRole = currentUser.getEmployee().getEmployeeRole();
+		boolean isAuthorized = Boolean.TRUE.equals(employeeRole.getIsSuperAdmin())
```
ThinuwanW, line 2199: 🤖 **Claude** · 🟡 **Suggestion**

The "super admin or attendance admin/manager" predicate is now written a third time. `TimeAnalyticsServiceImpl` already has `isUserSuperAdminOrAttendanceAdmin(User)` (super-admin OR ATTENDANCE_ADMIN) and an inline admin-or-manager check in `getIndividualWorkUtilization`. Three subtly different copies of the same authorization rule will drift as roles change. Extract a single helper (e.g. `PeopleUtil.hasAttendanceManagerOrAbovePermissions(EmployeeRole)`) and call it from all three sites.

frontend/src/community/attendance/hooks/useManualEntryRestriction.ts
```diff
@@ -0,0 +1,27 @@
...
+import useSessionData from "~community/common/hooks/useSessionData";
+
+/**
+ * Resolves whether the signed in user is blocked from adding or editing time entries
+ * by the organization level "Restrict Manual Time Entries & Edits" setting. Attendance
+ * Managers, Attendance Admins and Super Admins keep their access while it is enabled.
+ *
+ * The setting is enforced server side as well, so this only drives what is rendered.
+ */
+const useManualEntryRestriction = () => {
```
ThinuwanW, line 15: 🤖 **Claude** · 🟡 **Suggestion**

This new hook is the single source of truth for whether the manual-entry UI is gated, yet it ships without a test. Its behaviour has several branches that are easy to regress: config flag true/false/undefined, and each of `isSuperAdmin` / `isAttendanceAdmin` / `isAttendanceManager` granting access. Add `useManualEntryRestriction.test.ts` mocking `useGetAttendanceConfiguration` and `useSessionData` and asserting each combination (including the in-flight `undefined` config case and, once fixed, the sub-Core tier case).

backend/src/test/java/com/skapp/community/timeplanner/service/impl/AttendanceConfigServiceImplUnitTest.java
```diff
@@ -109,14 +111,46 @@ void getAllAttendanceConfigs_adminReturnsFullDto() {
...
+
+		AttendanceConfigRequestDto dto = (AttendanceConfigRequestDto) attendanceConfigService.getAllAttendanceConfigs()
+			.getResults()
+			.getFirst();
+
+		assertTrue(dto.getIsManualEntryRestrictionEnabled());
+	}
+
 	@Test
 	void getAllAttendanceConfigs_nonAdminReturnsGeoFencingOnly() {
```
ThinuwanW, line 154: 🤖 **Claude** · 🟣 **Nit**

`getAllAttendanceConfigs_nonAdminReturnsGeoFencingOnly` is now a misnomer — after this change the non-admin DTO also carries `isManualEntryRestrictionEnabled`. Rename it (e.g. `getAllAttendanceConfigs_nonAdminReturnsEmployeeVisibleConfigsOnly`) and add `assertFalse(dto.getIsManualEntryRestrictionEnabled())` so the "config row absent ⇒ false, not null" contract that `useManualEntryRestriction` depends on is actually pinned. Also note `getAllAttendanceConfigs_adminReturnsFullDto` only asserts `assertNotNull`, so the new `MANUAL_ENTRY_RESTRICTION_ENABLED` switch branch in `getAttendanceConfigRequestDto` has no coverage on the admin path.

frontend/src/community/common/assets/languages/english/attendanceModule.json
```diff
@@ -22,6 +22,31 @@
...
+    "manualEntryRestrictionCoreTier": "Core",
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Several of the new keys duplicate strings that already exist in this same `attendanceConfiguration` namespace: `manualEntryRestrictionCoreTier: "Core"` sits next to the generic `proTier: "Pro"` (line 23) used by the geofencing badge; `manualEntryRestrictionErrorToastTitle` is character-for-character `updateErrorMessageTitle` (line 17); and `manualEntryRestrictionModal.cancelButton` duplicates `cancelButtonText` (line 14). Prefer adding a generic `coreTier` sibling to `proTier` and reusing the existing generic title/button keys, so future tier-gated settings don't each grow their own copy.

frontend/src/community/attendance/components/molecules/TimesheetDailyRecordTableRow/TimesheetDailyRecordTableRow.tsx
```diff
@@ -196,17 +198,25 @@ const TimesheetDailyRecordTableRow: FC<Props> = ({ record, headerLength }) => {
...
   return (
     <Stack
       direction="row"
       justifyContent="space-between"
       alignItems="center"
-      sx={classes.stackContainerStyle}
-      onClick={() => mutate()}
+      sx={mergeSx([
+        classes.stackContainerStyle,
+        isManualEntryRestricted ? { cursor: "default" } : undefined
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Overriding only `cursor` leaves the rest of the clickable affordance intact: `styles.ts` `stackContainerStyle` defines a `&:hover` rule that changes the row background and its first child's background (styles.ts lines 36-41). A restricted employee still sees the row light up under the pointer even though nothing happens on click. Extend the restricted branch to neutralise the hover state as well, e.g. `isManualEntryRestricted ? { cursor: "default", "&:hover": { background: theme.palette.common.white, "& > :first-child": { background: theme.palette.common.white } } } : undefined`.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeServiceImpl.java
erandi-rootcode: Fixed. Used ModuleException with the relevant error message

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeServiceImpl.java
erandi-rootcode: Not Valid. The requirement is Disable and Hide Manual time Entry.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeServiceImpl.java
erandi-rootcode: Not Valid, as per the design the EmployeeRole cannot be null for attendance module.

frontend/src/community/common/assets/languages/english/attendanceModule.json
erandi-rootcode: This is to be implemented via another ticket which will be taken after this. https://rootcode.skapp.com/pm/projects/SKAPP/items/2962

PR #1731, opened 2026-08-10
feat(crm): v2 company & contact response types

anusham-rc:
## CRM v2 — Company & Contact (response-type refactor)

First PR in a stacked series introducing `/v2/crm` endpoints that return the new
composed/embedded response shapes. **No v1 API is changed** — v2 is a parallel layer.

### Company
- `GET /v2/crm/company/metrics` → `CrmCompanyMetricsResponseDtoV2` = `{ company, metrics }` (nested).
- New metric types `CrmCompanyMetrics` (reused by the upcoming `GET /company/{id}/metrics`) and `CrmCompanyMetricsSummary` (list query projection).
- Repo reuses existing predicate/order helpers; service assembles entity + metrics (mirrors `getContactMetrics`).

### Contact
- `GET /v2/crm/contact/metrics` → `{ contact, metrics }` (4 list metrics).
- `GET /v2/crm/contact/{id}` → base `CrmContactResponseDtoV2` (full embedded `company` + `owner`).
- `POST` / `PATCH /{id}` → base contact.
- Writes reuse v1 persistence + enterprise tier-limit enforcement via additively-extracted `persistNewContact` / `applyContactEdit` (existing v1 `createContact`/`editContact` behavior unchanged).

### Notes
- Board, lookups, `/exists`, `/owners` stay on v1 (shapes unchanged).
- `GET /company/{id}/metrics` and `GET /contact/{id}/metrics` are intentionally deferred.


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds a V2 CRM surface (companies + contacts) alongside V1: two new controllers, two service interfaces/impls, a V2 MapStruct mapper, V2 response DTOs, two new Criteria-API repository methods, and integration tests. The refactor of `CrmContactServiceImpl` (extracting `persistNewContact`/`applyContactEdit`) preserves all validations, logging, ordering and error paths, roles match the V1 equivalents, and the metric-summary constructor argument ordering is correct. The main concerns are a 500-on-missing-`size` paging defect on the new `GET /v2/crm/company` endpoint, a large amount of copy-paste duplication in the new repository method and the V2 contact service, and several V1/V2 asymmetries (pageable construction, null defaults, logging names).

Found **21** new issue(s): 🔴 6 important, 🟡 10 suggestion(s), 🟣 5 nit(s)

backend/src/main/java/com/skapp/community/crmplanner/controller/v2/CrmCompanyControllerV2.java
```diff
@@ -0,0 +1,36 @@
...
+public class CrmCompanyControllerV2 {
+
+	private final CrmCompanyServiceV2 companyService;
+
+	@Operation(summary = "Get companies",
+			description = "Returns a paginated list of companies, each with its base details and metrics.")
+	@GetMapping
+	@PreAuthorize("hasAnyRole('ROLE_CRM_SALES_REPRESENTATIVE')")
+	public ResponseEntity<ResponseEntityDto> getCompanyMetrics(CrmCompanyMetricRequestDto requestDto) {
+		Pageable pageable = PageRequest.of(requestDto.getPage(), requestDto.getSize());
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`CrmCompanyMetricRequestDto` declares `private int page;` and `private int size;` with no initializers, so a request to `GET /v2/crm/company` that omits `size` binds `size = 0` and `PageRequest.of(0, 0)` throws `IllegalArgumentException("Page size must not be less than one")`. That is not handled by any specific `@ExceptionHandler` in `GlobalExceptionHandler`, so it falls through to the catch-all `Exception` handler and returns 500 to the client. The sibling V2 endpoint behaves differently: `CrmContactMetricRequestDto` defaults to `page = 0` / `size = 10`, so `GET /v2/crm/contact` works with no params (the new contact test relies on exactly that). Fix by giving `CrmCompanyMetricRequestDto` the same defaults (`private int page = 0; private int size = 10;`), or by clamping/validating in the service. Note the existing `/v1/crm/company/metrics` endpoint has the identical flaw, so a fix on the DTO repairs both. Also worth adding a test that hits `/v2/crm/company` with no query params — the current tests always pass both `page` and `size`, which is why this is not caught.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmCompanyRepositoryImpl.java
```diff
@@ -118,6 +119,90 @@ public Page<CrmCompanyMetricsResponseDto> getCompanyMetrics(Pageable pageable, S
...
+
+		countQuery.select(cb.countDistinct(countRoot.get(CrmCompany_.id)))
+			.where(buildPredicates(cb, countRoot, searchKeyword));
+		Long total = entityManager.createQuery(countQuery).getSingleResult();
+
+		return new PageImpl<>(content, pageable, total);
+	}
+
+	@Override
+	public List<CrmCompanyMetricsSummary> findCompanyMetricsSummaries(List<Long> companyIds) {
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`findCompanyMetricsSummaries` is a near-verbatim copy of lines 55–96 of `getCompanyMetrics`: the six correlated subqueries (open tasks, overdue tasks, open value, account value, closed count, open count) are duplicated character-for-character, and `searchCompanies` additionally duplicates the offset/max-results + `countDistinct` block from the same method. That is ~90 duplicated lines in one class. Any future change to the metric definition (e.g. adjusting what counts as "overdue", or adding a `stage.isDeleted` filter) now has to be made in two places and will silently diverge between the V1 and V2 endpoints. Extract private helpers that take `(CriteriaBuilder cb, AbstractQuery<?> query, From<?, CrmCompany> company, List<Long> closedStageIds)` and return each `Subquery`, then have both `getCompanyMetrics` and `findCompanyMetricsSummaries` call them; likewise extract the paging+count block shared by `getCompanyMetrics` and `searchCompanies`.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmContactServiceImplV2.java
```diff
@@ -0,0 +1,143 @@
...
+
+	private final CrmDealDao crmDealDao;
+
+	private final CrmTaskDao crmTaskDao;
+
+	private final CrmMapperV2 crmMapperV2;
+
+	@Override
+	@Transactional(readOnly = true)
+	public ResponseEntityDto getContactMetrics(CrmContactMetricRequestDto filterDto) {
```
ThinuwanW, line 37: 🤖 **Claude** · 🔴 **Important**

`getContactMetrics` here is a line-for-line duplicate of `CrmContactServiceImpl.getContactMetrics` (same `PageRequest.of`, same `findContacts`, same empty-id early return, same two `Collectors.toMap` summary lookups, same enrich loop) — the only difference is the item DTO shape. `enrichWithMetrics` and `buildPageDto` are duplicated the same way (and `buildPageDto` is duplicated a third time in `CrmCompanyServiceImplV2`). Extract the shared query+summary-fetch step into a reusable method (e.g. on `CrmContactService`, returning the page plus the two summary maps) and let each version only do its own DTO assembly; extract `buildPageDto` into a small shared helper (e.g. in `CrmUtil`) that takes `List<?> items` and a `Page<?>`. As written, a fix to the metrics query has to be applied twice.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmCompanyServiceImplV2.java
```diff
@@ -0,0 +1,83 @@
...
+		return new ResponseEntityDto(false, buildPageDto(items, companyPage));
+	}
+
+	private CrmCompanyListItemDtoV2 enrichWithMetrics(CrmCompany company,
+			Map<Long, CrmCompanyMetricsSummary> summaryMap) {
+		CrmCompanyResponseDto companyDto = crmMapper.crmCompanyToCrmCompanyResponseDto(company);
+
+		CrmCompanyMetricsSummary summary = summaryMap.get(company.getId());
+		CrmCompanyMetrics metrics = new CrmCompanyMetrics(summary != null ? summary.getOpenTasksCount() : 0L,
+				summary != null ? summary.getOverdue() : 0L, summary != null ? summary.getOpenValue() : null,
```

Inconsistent null defaults in `enrichWithMetrics`: the counts fall back to `0L` but `openValue` and `accountValue` fall back to `null`, so a client would see `"openDeals": 0` next to `"openValue": null`. The contact equivalent in `CrmContactServiceImplV2.enrichWithMetrics` uses `BigDecimal.ZERO.toPlainString()` (i.e. `"0"`) for the analogous field, so the two V2 metric payloads disagree on how "no data" is represented. Use `"0"` for both monetary fields. Separately, the fallbacks are effectively unreachable: `searchCompanies` already filters `isDeleted = false` and `findCompanyMetricsSummaries` selects every id in `companyIds` with the same filter inside the same read-only transaction, so `summaryMap.get(...)` never returns null — consider replacing the six repeated ternaries with a single `summaryMap.getOrDefault(company.getId(), EMPTY_SUMMARY)` (or a MapStruct mapping) for readability.

backend/src/main/java/com/skapp/community/crmplanner/controller/v2/CrmCompanyControllerV2.java
```diff
@@ -0,0 +1,36 @@
...
+@Tag(name = "CRM Companies Controller V2", description = "Operations related to CRM Companies")
+public class CrmCompanyControllerV2 {
+
+	private final CrmCompanyServiceV2 companyService;
+
+	@Operation(summary = "Get companies",
+			description = "Returns a paginated list of companies, each with its base details and metrics.")
+	@GetMapping
+	@PreAuthorize("hasAnyRole('ROLE_CRM_SALES_REPRESENTATIVE')")
+	public ResponseEntity<ResponseEntityDto> getCompanyMetrics(CrmCompanyMetricRequestDto requestDto) {
```
ThinuwanW, line 28: 🤖 **Claude** · 🟡 **Suggestion**

Asymmetric layering between the two new V2 endpoints: this controller builds the `Pageable` itself and the service signature is `getCompanyMetrics(String searchKeyword, Pageable pageable)`, while `CrmContactControllerV2.getContactMetrics` passes the whole `CrmContactMetricRequestDto` straight through and `CrmContactServiceImplV2` builds the `PageRequest`. Per the review guidelines controllers should only delegate. Since this is a brand-new V2 surface (not constrained by an existing V1 signature), align it with the contact version: change `CrmCompanyServiceV2` to `ResponseEntityDto getCompanyMetrics(CrmCompanyMetricRequestDto requestDto)` and move the `PageRequest.of(...)` into `CrmCompanyServiceImplV2`. This also puts the paging-default fix in one place.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmContactServiceImpl.java
```diff
@@ -96,6 +96,13 @@ public ResponseEntityDto checkContactEmailExists(String email) {
...
 	@Transactional
 	public ResponseEntityDto createContact(CrmContactCreateRequestDto requestDto) {
+		CrmContact savedContact = persistNewContact(requestDto);
+		return new ResponseEntityDto(false, crmMapper.crmContactToCrmContactResponseDto(savedContact));
+	}
+
+	@Override
+	@Transactional
+	public CrmContact persistNewContact(CrmContactCreateRequestDto requestDto) {
 		log.info("createContact: execution started");
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The extracted methods keep the old log labels: `persistNewContact` logs `"createContact: execution started"`/`"...ended"` and `applyContactEdit` (line 152/204) logs `"editContact: ..."`. The convention in this codebase is `log.info("<methodName>: execution started")`, and because `CrmContactServiceImplV2.createContact` also logs `"createContact: execution started"` before delegating, a single V2 create now emits `createContact: execution started` twice with nothing in the logs distinguishing the V2 wrapper from the V1 core — which makes tracing a failed create harder. Rename the log messages to `persistNewContact:` / `applyContactEdit:` to match their method names. Also note the thin `createContact`/`editContact` wrappers (lines 96–101, 142–147) now have no entry logging at all.

backend/src/main/java/com/skapp/community/crmplanner/type/CrmCompanyMetrics.java
```diff
@@ -0,0 +1,22 @@
+package com.skapp.community.crmplanner.type;
+
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+
+@Getter
+@AllArgsConstructor
```
ThinuwanW, line 7: 🤖 **Claude** · 🟡 **Suggestion**

Package placement and naming: `CrmCompanyMetrics` and `CrmContactMetrics` are serialized directly into the HTTP response (nested under `items[].metrics`), so they are response payloads and belong in `payload/response/v2/` next to `CrmCompanyListItemDtoV2`, named `*Dto` per the naming convention. The `type/` package is documented for enums and internal projections — `CrmCompanyMetricsSummary` legitimately belongs there (it mirrors the existing `CrmDealSummary`/`CrmTaskSummary` projections), but the two response-facing classes do not. Additionally, `CrmCompanyMetrics` is a field-for-field duplicate of `CrmCompanyMetricsSummary` minus `companyId`; consider dropping one of them (e.g. serialize the summary directly, or generate the DTO with a MapStruct mapping) rather than hand-copying six fields.

backend/src/main/java/com/skapp/community/crmplanner/service/CrmContactService.java
```diff
@@ -13,8 +14,12 @@ public interface CrmContactService {
 
 	ResponseEntityDto createContact(CrmContactCreateRequestDto requestDto);
 
+	CrmContact persistNewContact(CrmContactCreateRequestDto requestDto);
```
ThinuwanW, line 17: 🤖 **Claude** · 🟡 **Suggestion**

`persistNewContact` and `applyContactEdit` widen the public service contract to return the JPA entity `CrmContact`. Because `CrmContact.company` and `CrmContact.owner` are `FetchType.LAZY` and `applyContactEdit` loads the contact via `findByIdAndIsDeletedFalse` (no fetch join), the returned entity carries uninitialized proxies — it is only safe to map because `CrmContactServiceImplV2.editContact` happens to be `@Transactional`. Any future caller that forgets that annotation gets a `LazyInitializationException` at mapping time rather than a compile error. Consider narrowing the contract (e.g. return the id, or have the V1 service expose a `CrmContactResponseDtoV2`-producing method) or at minimum javadoc these two methods as "must be called inside an active transaction; associations are lazy".

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmCompanyControllerV2IntegrationTest.java
```diff
@@ -0,0 +1,172 @@
...
+		deal.setOwner(employeeDao.getReferenceById(1L));
+		deal.setPriority(CrmDealPriority.MEDIUM);
+		deal.setOrderIndex(orderIndex);
+		deal.setAmount(amount);
+		crmDealDao.save(deal);
+	}
+
+	@Test
+	@DisplayName("Get company metrics - Returns nested company and metrics with seeded values")
+	void getCompanyMetrics_HappyPath_ReturnsNestedCompanyAndMetrics() throws Exception {
```
ThinuwanW, line 155: 🤖 **Claude** · 🟡 **Suggestion**

Test coverage gaps on the company V2 endpoint. The happy-path test seeds only deals, so `metrics.openTasksCount` and `metrics.overdue` are never asserted — the two task subqueries in the new `findCompanyMetricsSummaries` are entirely unexercised, including the `dueAt < now` overdue branch. There is also no test for the empty-page branch in `CrmCompanyServiceImplV2.getCompanyMetrics` (the `companyIds.isEmpty()` early return), and no test that calls the endpoint without `page`/`size`, which is exactly the request shape that currently 500s. Please add: (1) an open task and an overdue task on the seeded company with assertions on both counters, (2) a search for a keyword matching nothing asserting `totalItems == 0` and an empty `items` array, (3) a no-query-param request once the paging default is fixed.

backend/src/main/java/com/skapp/community/crmplanner/controller/v2/CrmCompanyControllerV2.java
```diff
@@ -0,0 +1,36 @@
...
+@RequestMapping("/v2/crm/company")
+@Tag(name = "CRM Companies Controller V2", description = "Operations related to CRM Companies")
+public class CrmCompanyControllerV2 {
+
+	private final CrmCompanyServiceV2 companyService;
+
+	@Operation(summary = "Get companies",
+			description = "Returns a paginated list of companies, each with its base details and metrics.")
+	@GetMapping
+	@PreAuthorize("hasAnyRole('ROLE_CRM_SALES_REPRESENTATIVE')")
```
ThinuwanW, line 27: 🤖 **Claude** · 🟣 **Nit**

`hasAnyRole('ROLE_CRM_SALES_REPRESENTATIVE')` with a single role — use `hasRole(...)`, which is what the sibling `CrmContactControllerV2` uses on all four of its endpoints. The effective authorization is identical, so this is purely a consistency point (V1 is itself mixed on this), but the two new V2 controllers added in the same PR should at least agree. Relatedly, the method/operation name `getCompanyMetrics` for what is now the plain company collection endpoint (`GET /v2/crm/company`, described as "Returns a paginated list of companies") reads oddly; `getCompanies` would match the `get*` CRUD convention and the endpoint's actual semantics.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmContactControllerV2IntegrationTest.java
```diff
@@ -0,0 +1,248 @@
...
+	@Test
+	@DisplayName("Get contact metrics - Returns page with nested contact and metrics")
+	void getContactMetrics_WithContacts_ReturnsNestedContactAndMetrics() throws Exception {
+		Long companyId = savedCompany("Metrics V2 Corp").getId();
+		Long contactId = savedContact(companyId, "metrics.contact.v2@example.com").getId();
+
+		performGetMetricsRequest().andDo(print())
+			.andExpect(status().isOk())
+			.andExpect(jsonPath(STATUS_PATH).value(STATUS_SUCCESSFUL))
+			.andExpect(jsonPath("['results'][0]['totalItems']").value(1))
```
ThinuwanW, line 290: 🤖 **Claude** · 🟣 **Nit**

`getContactMetrics_WithContacts_ReturnsNestedContactAndMetrics` calls the endpoint with no `searchKeyword` and asserts `totalItems == 1`, i.e. it assumes the whole `crm_contact` table is empty apart from the one row it seeds. That holds today only because `data.sql` seeds no CRM contacts and `@Transactional` rolls each test back; it will break the moment any CRM contact is added to the seed data. Pass a unique `searchKeyword` (as `CrmCompanyControllerV2IntegrationTest` does with `MetricsCoV2Unique`) so the assertion is self-isolating. Also missing from this class: a not-found case for `GET /v2/crm/contact/{id}` (the `ModuleException` branch in `CrmContactServiceImplV2.getContactById` is untested) and a forbidden case for the edit/get-by-id endpoints.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmCompanyRepositoryImpl.java
```diff
@@ -118,6 +119,90 @@ public Page<CrmCompanyMetricsResponseDto> getCompanyMetrics(Pageable pageable, S
...
 	}
 
+	@Override
+	public Page<CrmCompany> searchCompanies(Pageable pageable, String searchKeyword) {
+		CriteriaBuilder cb = entityManager.getCriteriaBuilder();
+		CriteriaQuery<CrmCompany> query = cb.createQuery(CrmCompany.class);
+		Root<CrmCompany> company = query.from(CrmCompany.class);
+
+		query.where(buildPredicates(cb, company, searchKeyword));
+		query.orderBy(buildOrderBy(cb, company, searchKeyword));
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Unstable pagination: `searchCompanies` orders only by relevance rank and `lower(name)` (via `buildOrderBy`), with no unique tiebreaker. Companies that share a name — which is allowed, `crm_company.name` has no unique constraint and `existsByNameIgnoreCaseAndIsDeletedFalse` is only an advisory pre-check — have an undefined relative order, so the same company can appear on two consecutive pages while another is never returned. The sibling `CrmContactRepositoryImpl.findContacts` already handles this correctly (`List.of(cb.desc(dealValueSub), cb.asc(contact.get(CrmContact_.id)))`). Fix: append `cb.asc(company.get(CrmCompany_.id))` as the last order in `buildOrderBy` (this also fixes the pre-existing `getCompanyMetrics` path since the helper is shared). Separately, note that `searchCompanies` silently discards `pageable.getSort()` — either honour it or document that the ordering is fixed, so a future caller passing a `Sort` isn't surprised.

backend/src/main/java/com/skapp/community/crmplanner/type/CrmCompanyMetricsSummary.java
```diff
@@ -0,0 +1,24 @@
+package com.skapp.community.crmplanner.type;
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

📁 **File-level observation**

`CrmCompanyMetricsSummary` and `CrmCompanyMetrics` are the same value type twice: both declare `openTasksCount`, `overdue`, `openValue`, `accountValue`, `openDeals`, `closedDeals` in the same order, differing only by the summary's leading `companyId`. This forces the hand-written six-argument copy in `CrmCompanyServiceImplV2.enrichWithMetrics`, which is exactly where the inconsistent null defaults arise, and any future metric added to one class silently drifts from the other. Fix: keep one type — either have the projection carry a nested `CrmCompanyMetrics` (`cb.construct` supports nested construct expressions), or have the repository return `Map<Long, CrmCompanyMetrics>` keyed by company id and drop `CrmCompanyMetricsSummary` entirely. Then `enrichWithMetrics` becomes `summaryMap.getOrDefault(company.getId(), EMPTY_METRICS)` with a single shared default instance.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmCompanyServiceImplV2.java
```diff
@@ -0,0 +1,83 @@
...
+				summary != null ? summary.getAccountValue() : null, summary != null ? summary.getOpenDeals() : 0L,
+				summary != null ? summary.getClosedDeals() : 0L);
+
+		return new CrmCompanyListItemDtoV2(companyDto, metrics);
+	}
+
+	private PageDto buildPageDto(List<CrmCompanyListItemDtoV2> items, Page<CrmCompany> companyPage) {
```

Both new V2 services hand-roll page assembly (`buildPageDto` here and the byte-identical copy at `CrmContactServiceImplV2.java:134`) when the codebase already has a shared `PageTransformer` bean (`com.skapp.community.common.util.transformer.PageTransformer`) that does exactly this. The sibling V2 service `PeopleServiceImplV2` uses it (`PageDto pageDto = pageTransformer.transform(page); pageDto.setItems(items);`), as do `LeaveServiceImpl`, `TimeServiceImpl` and `EpCrmDealServiceImpl`. Fix: inject `PageTransformer` in both V2 services and delete both `buildPageDto` methods. One caveat worth checking before swapping: `PageTransformer.transform` returns a zeroed `PageDto` whenever `page.isEmpty()`, so for an out-of-range page request it reports `totalItems: 0` where the current code reports the real total — if that difference matters, fix `PageTransformer` rather than re-introducing the local copies.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmCompanyServiceImplV2.java
```diff
@@ -0,0 +1,83 @@
...
+@RequiredArgsConstructor
+public class CrmCompanyServiceImplV2 implements CrmCompanyServiceV2 {
+
+	private final CrmCompanyDao crmCompanyDao;
+
+	private final CrmMapper crmMapper;
+
+	@Override
+	@Transactional(readOnly = true)
+	public ResponseEntityDto getCompanyMetrics(String searchKeyword, Pageable pageable) {
```

No upper bound on the requested page size. `size` comes straight from the client into `PageRequest.of(...)` with no clamp, and this endpoint is unusually expensive per row: `findCompanyMetricsSummaries` runs six correlated subqueries per company and receives every id of the page in a single `IN (...)` list. A request with `size=100000` turns into one very large IN clause plus 600k correlated subquery evaluations — an easy accidental (or deliberate) way for an authenticated sales rep to stall the DB. Fix: clamp in the service, e.g. `int size = Math.min(pageable.getPageSize(), MAX_PAGE_SIZE)` with `MAX_PAGE_SIZE` as a constant (or an `application.yml` property), and rebuild the `Pageable` before querying. Applies equally to `CrmContactServiceImplV2.getContactMetrics`.

backend/src/main/java/com/skapp/community/crmplanner/payload/response/v2/CrmContactResponseDtoV2.java
```diff
@@ -0,0 +1,30 @@
+package com.skapp.community.crmplanner.payload.response.v2;
```
ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

`CrmContactResponseDtoV2` declares the same eight fields in the same order as the existing `CrmContactResponseDto` (id, name, email, contactNumber, lastContactAt, lastModifiedDate, company, owner); the only real difference is that `company` is `CrmCompanyResponseDto` instead of `CrmCompanyLookupResponseDto` and `owner` is `CrmOwnerResponseDto` instead of `CrmContactOwnerResponseDto`. That two-type difference also forced an entire new mapper interface, `CrmMapperV2`, that exists to declare a single method. Fix: either add `crmContactToCrmContactResponseDtoV2` to the existing `CrmMapper` (MapStruct resolves the nested company/owner mappings from the same interface, so `uses` isn't needed) and drop `CrmMapperV2`, or make the V1 DTO generic over the nested types. At minimum, if `CrmMapperV2` stays, it should be justified by more than one method — otherwise the V2 mapper surface will keep growing as a near-clone of `CrmMapper`.

backend/src/main/java/com/skapp/community/crmplanner/type/CrmCompanyMetrics.java
```diff
@@ -0,0 +1,22 @@
...
+public class CrmCompanyMetrics {
+
+	private final Long openTasksCount;
+
+	private final Long overdue;
+
```
ThinuwanW, line 13: 🤖 **Claude** · 🟡 **Suggestion**

The two metric DTOs introduced in this same PR name the same concepts inconsistently, which will surface directly in the public JSON. `CrmCompanyMetrics` exposes `overdue` and `openDeals`/`closedDeals`, while `CrmContactMetrics` exposes `overdueTasksCount` and `closedDealCount` for the equivalent values — so a client reads `metrics.overdue` on the company endpoint but `metrics.overdueTasksCount` on the contact endpoint, and `metrics.closedDeals` vs `metrics.closedDealCount`. `overdue` is also a bare adjective for what is a count. The `overdue` name is inherited from V1's `CrmCompanyMetricsResponseDto`, but a new V2 contract is the right place to correct it. Fix: rename to `overdueTasksCount`, `openDealsCount`, `closedDealsCount` in `CrmCompanyMetrics` and align `CrmContactMetrics` to the same `*Count` suffix convention.

backend/src/main/java/com/skapp/community/crmplanner/type/CrmContactMetrics.java
```diff
@@ -0,0 +1,18 @@
+package com.skapp.community.crmplanner.type;
+
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+
+@Getter
+@AllArgsConstructor
+public class CrmContactMetrics {
+
+	private final String closedDealValue;
```
ThinuwanW, line 10: 🤖 **Claude** · 🟣 **Nit**

`closedDealValue` is typed `String` here, whereas the V1 equivalent `CrmContactListItemDto.closedDealValue` is a `BigDecimal`. The same logical field therefore serialises as a JSON string (`"0"`, asserted as such in the new integration test) on `/v2/crm/contact` and as a JSON number on `/v1/crm/contact/metrics`. If the string form is deliberate (avoiding float precision loss in JS clients) that's defensible and matches `CrmCompanyMetrics.openValue`, but it is an undocumented representation change between v1 and v2 that will break any client that reuses its v1 parsing. Please either state the rationale in the `@Operation` description / API docs, or keep `BigDecimal` and let Jackson's `WRITE_BIGDECIMAL_AS_PLAIN` handle the formatting.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmContactControllerV2IntegrationTest.java
```diff
@@ -0,0 +1,248 @@
+package com.skapp.community.crmplanner.controller.v2;
```
ThinuwanW, line 1: 🤖 **Claude** · 🔴 **Important**

📁 **File-level observation**

Coverage gaps in the contact V2 test class, distinct from the assertion-fragility issue: (1) the single metrics test seeds a contact with no deals and no tasks, so every assertion lands on the `deals == null` / `tasks == null` fallback branch of `CrmContactServiceImplV2.enrichWithMetrics` — the real aggregation path (`deals.getTotalClosedValue().toPlainString()`, `getClosedDealCount()`, `getOpenTaskCount()`, `getOverdueTaskCount()`) is never executed, so a wrong summary-map lookup or a swapped constructor argument would pass; (2) the `contactIds.isEmpty()` early-return branch (which builds the page from `List.of()`) is never hit; (3) `getContactById` is only tested for the happy path — the `contact == null` → `ModuleException(CRM_ERROR_CONTACT_NOT_FOUND)` path has no test, unlike `editContact_NotFound_ReturnsBadRequest`; (4) `@PreAuthorize` is only asserted on `createContact` and `getContactMetrics`, not on `editContact` or `getContactById`. Fix: seed a WON deal and an overdue task and assert non-zero metric values, add a search-keyword test that matches nothing, add a not-found test for `getContactById`, and add forbidden-role tests for the two remaining endpoints.

backend/src/main/java/com/skapp/community/crmplanner/service/CrmContactService.java
ThinuwanW, line 17: 🤖 **Claude** · 🟣 **Nit**

`persistNewContact` and `applyContactEdit` don't follow the service-method verb convention used everywhere else in this module (`get*`, `create*`, `update*`, `delete*` — see `createContact`, `editContact`, `deleteContact`, `getContactById` in this same interface). Since these are now public interface methods, not private helpers, they should read consistently: e.g. `createContactEntity` / `updateContactEntity`, which also makes the entity-returning nature obvious at the call site in `CrmContactServiceImplV2`.

backend/src/main/java/com/skapp/community/crmplanner/controller/v2/CrmContactControllerV2.java
```diff
@@ -0,0 +1,64 @@
...
+@RequestMapping("/v2/crm/contact")
+@Tag(name = "CRM Contacts Controller V2", description = "Operations related to CRM Contacts")
+public class CrmContactControllerV2 {
+
+	private final CrmContactServiceV2 contactService;
+
+	@Operation(summary = "Create CRM contact", description = "Creates a CRM contact and assigns an owner.")
+	@PreAuthorize("hasRole('ROLE_CRM_SALES_REPRESENTATIVE')")
+	@PostMapping
+	public ResponseEntity<ResponseEntityDto> createContact(@RequestBody CrmContactCreateRequestDto requestDto) {
```
ThinuwanW, line 33: 🤖 **Claude** · 🟣 **Nit**

`@RequestBody CrmContactCreateRequestDto` (and the `CrmContactEditRequestDto` on `editContact`) has no `@Valid`, and the DTOs themselves carry no bean-validation annotations — `CrmContactCreateRequestDto` is five bare fields with no `@NotBlank`/`@Email`/`@NotNull`. All validation happens deep inside `persistNewContact` via `CrmValidations`, so a malformed payload is only rejected after `userService.getCurrentUser()` and the email-exists query have run. This mirrors V1 so it isn't a regression, but a new controller is a reasonable place to start converging: add `@NotBlank`/`@Email`/`@NotNull` to the request DTOs and `@Valid` to the parameters, so the payload is rejected at the binding layer before any service work.

backend/src/main/java/com/skapp/community/crmplanner/controller/v2/CrmCompanyControllerV2.java
anusham-rc: Fixed. `CrmCompanyMetricRequestDto` now defaults `page=0`/`size=10` (matching the contact DTO), so a no-param request no longer hits `PageRequest.of(0,0)`. This also repairs the shared v1 `/v1/crm/company/metrics` path. Added a no-param integration test.

backend/src/main/java/com/skapp/community/crmplanner/controller/v2/CrmCompanyControllerV2.java
anusham-rc: Keeping `hasAnyRole(...)` to match the V1 `getCompanyMetrics` endpoint; with a single role it is functionally identical. Method name stays `getCompanyMetrics` for symmetry with `getContactMetrics`.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmContactControllerV2IntegrationTest.java
anusham-rc: Hardened the metrics tests to filter by a unique `searchKeyword` so they are self-isolating, and added a `getContactById` not-found test plus forbidden-role tests for `editContact` and `getContactById`.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmCompanyServiceImplV2.java
anusham-rc: `buildPageDto` was removed from both V2 services. Kept explicit page assembly rather than `PageTransformer` on purpose: as you noted, `PageTransformer` zeroes `totalItems` on an empty/out-of-range page, which would report the wrong total here.

backend/src/main/java/com/skapp/community/crmplanner/type/CrmCompanyMetrics.java
anusham-rc: The duplication is resolved (`CrmCompanyMetricsSummary` was deleted). Leaving the two small metric value types in `type/` for now to avoid a broad rename/move; can revisit in a follow-up.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmCompanyRepositoryImpl.java
anusham-rc: Not adding an `id` tiebreaker. Active company names are kept unique by the app-level `existsByNameIgnoreCaseAndIsDeletedFalse` guard on create, so the name ordering has no ties in practice. Ordering here is intentionally fixed (relevance then name).

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmCompanyServiceImplV2.java
anusham-rc: Valid, but out of scope for this PR: there is no page-size clamp anywhere in the codebase (V1 included). Better handled as a dedicated hardening ticket so it is applied uniformly.

backend/src/main/java/com/skapp/community/crmplanner/type/CrmCompanyMetricsSummary.java
anusham-rc: Resolved exactly as suggested: dropped `CrmCompanyMetricsSummary` and used a nested `cb.construct` that carries `CrmCompanyMetrics` directly.

backend/src/main/java/com/skapp/community/crmplanner/type/CrmContactMetrics.java
anusham-rc: Intentional. The string form matches the company monetary fields (`openValue`/`accountValue`) and the frontend already consumes it as a string (`formatValue(row.closedDealValue?.toString())`).

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmContactServiceImplV2.java
anusham-rc: Resolved. V2 now runs one criteria query (`getContactMetricsV2`) and no longer shares the `findContacts` + summary-map + enrich flow with V1.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmContactServiceImpl.java
anusham-rc: Fixed. `persistNewContact` / `applyContactEdit` now log their own method names, so a V2 create no longer emits `createContact: execution started` twice.

backend/src/main/java/com/skapp/community/crmplanner/controller/v2/CrmContactControllerV2.java
anusham-rc: Mirrors V1 (no bean-validation on these DTOs). Adding `@Valid` plus constraints is a broader convergence change, better scoped to its own PR.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmCompanyControllerV2IntegrationTest.java
anusham-rc: Addressed. Added tests for an open task and an overdue task with assertions on both counters, plus an empty-search case (`totalItems == 0`, empty items) and a no-query-param request.

backend/src/main/java/com/skapp/community/crmplanner/controller/v2/CrmCompanyControllerV2.java
anusham-rc: Fixed. `CrmCompanyServiceV2.getCompanyMetrics` now takes `CrmCompanyMetricRequestDto` and builds the `PageRequest` in the service; the controller only delegates, matching `CrmContactControllerV2`.

backend/src/main/java/com/skapp/community/crmplanner/type/CrmCompanyMetrics.java
anusham-rc: Keeping the current field names. They are the existing frontend contract: `CompanyTable`/`companyUtil` read `overdue`, `openDeals`, `closedDeals`, and `ContactTable` reads `overdueTasksCount`, `closedDealCount`. Renaming would break the FE.

backend/src/main/java/com/skapp/community/crmplanner/service/CrmContactService.java
anusham-rc: Names are descriptive and clear at the call site; renaming ripples through the interface, impl, and V2 callers for marginal gain. Leaving as-is.

backend/src/main/java/com/skapp/community/crmplanner/payload/response/v2/CrmContactResponseDtoV2.java
anusham-rc: Leaving as-is. `CrmMapperV2` cleanly isolates the V2 mapping; folding it into `CrmMapper` and generifying the V1 DTO is churn for little gain on a new surface.

backend/src/main/java/com/skapp/community/crmplanner/service/CrmContactService.java
anusham-rc: This mirrors the existing V1 pattern (`editContact` maps the same lazily-loaded entity), and both V2 wrappers are `@Transactional`, so mapping is safe. Narrowing the contract is a larger change better scoped to a separate PR.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmCompanyRepositoryImpl.java
anusham-rc: Resolved by the single-query refactor. `findCompanyMetricsSummaries` and `searchCompanies` were removed; a single criteria query (`getCompanyMetricsV2`) now projects straight into the nested `CrmCompanyListItemDtoV2`, so the duplicated subquery block is gone.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmCompanyServiceImplV2.java
anusham-rc: Resolved. `enrichWithMetrics` was removed; metric values now come directly from the query projection, so the inconsistent fallbacks are gone.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmContactControllerV2IntegrationTest.java
anusham-rc: Addressed. Added a WON-deal + open/overdue-task test asserting non-zero metrics, a company-less-contact test (null company), a `getContactById` not-found test, and forbidden-role tests for edit and get-by-id. Note the `enrichWithMetrics` fallback branch referenced here no longer exists after the single-query refactor.

akilarootcode (dismissed):

PR #1732, opened 2026-08-10
feat(crm): v2 deal & task response types

anusham-rc:
## CRM v2 — Deal & Task (response-type refactor)

Stacked on #1731. Adds the `/v2/crm/deal` and `/v2/crm/task` endpoints returning the
fully-embedded response shapes. **No v1 API is changed.**

> Base is `feat/crm-response-types-refactor` (PR #1731). Review #1731 first; this PR
> shows only the deal + task diff and will retarget to `develop` once #1731 merges.

### Deal
- `GET /v2/crm/deal`, `GET /{id}`, `POST`, `PATCH /{id}` → `CrmDealResponseDtoV2`
  with fully embedded `stage`, `owner`, `company`, and `contact` (the contact carries
  its own `company` + `owner`) — self-contained, no client-side lookup needed.
- Reads preserve the sales-rep owner restriction; writes delegate to v1
  `persistNewDeal` / `applyDealEdit` so enterprise deal-limit + stage-access still run.

### Task
- `GET /v2/crm/task` (open), `GET /completed`, `GET /related`, `GET /{id}`, `POST`, `PATCH /{id}`
  → `CrmTaskResponseDtoV2` with fully embedded `type`, `owner`, `company`, `contact`, and `deal`.
- Open-list keeps the v1 `{ tasks: [...] }` envelope via `CrmGetTasksResponseDtoV2`; other reads paginated.
- Reads preserve owner restriction; writes delegate to v1 `persistNewTask` / `applyTaskEdit`
  so the enterprise task-limit still runs.

### Shared
- `CrmMapperV2` composes the nested v2 shapes (deal/task reuse the v2 contact mapping).
- Board deal endpoints stay on v1 (intentionally unchanged).
- v1 service changes are additive extractions only; existing endpoints behave identically.


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds a V2 API surface for CRM deals and tasks, extracting entity-returning helpers (`persistNewDeal`/`applyDealEdit`/`persistNewTask`/`applyTaskEdit`) from the V1 service impls and adding V2 controllers, services, DTOs and mapper methods. The write-path refactor is behavior-preserving (enterprise `validateDealCreationLimit`/`validateTaskCreationLimit` hooks and all `CrmValidations` calls remain in the extracted methods, and the `@Primary` enterprise subclasses are unaffected), and authorization/owner-scoping in the new V2 read methods matches V1. The main concerns are on the read path: the new nested V2 response DTOs traverse associations that the existing fetch graphs do not load (N+1), and V2 `getDeals` silently drops the soft-deleted-company masking that V1 `getDeals` applies.

Found **18** new issue(s): 🔴 4 important, 🟡 10 suggestion(s), 🟣 4 nit(s)

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmTaskServiceImplV2.java
```diff
@@ -0,0 +1,151 @@
...
+	@Transactional(readOnly = true)
+	public ResponseEntityDto getTasks(CrmTaskFilterDto filterDto) {
+		log.info("getTasks: execution started");
+
+		User currentUser = userService.getCurrentUser();
+		Long ownerId = CrmUtil.isCrmSalesRepresentative(currentUser) ? currentUser.getEmployee().getEmployeeId() : null;
+
+		List<CrmTaskResponseDtoV2> tasks = crmTaskDao.findTasks(ownerId, filterDto)
+			.stream()
+			.map(crmMapperV2::crmTaskToCrmTaskResponseDtoV2)
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

N+1 queries: the new V2 response DTOs traverse associations that the existing repository fetch graphs do not load.

Tasks — `CrmTaskRepositoryImpl.applyFetchGraph()` (used by `findTasks`, `findCompletedTasks`, `findRelatedTasks`) only fetches `type`, `owner`, `contact`, `deal`. `CrmTaskResponseDtoV2` additionally serializes `company`, `contact.company`, `contact.owner`, and a full `CrmDealResponseDtoV2` for `deal` (which pulls `deal.stage`, `deal.owner`, `deal.company`, `deal.contact`, and in turn `deal.contact.company` / `deal.contact.owner`). All of these are `FetchType.LAZY`, so mapping each row triggers up to ~7-8 extra SELECTs. V1's `CrmTaskResponseDto` only pulled `contact.company` (via `CrmContactLookupResponseDto`), so this is a large regression — and `getTasks` (line 54) is unpaginated, so the row count is unbounded. `findByIdWithAssociations` has the same gap (it fetches `deal.stage`/`deal.owner` but not `company`, `contact.company`, `contact.owner`, `deal.company`, `deal.contact`).

Deals — same problem in `CrmDealServiceImplV2.getDeals` (line 58): `CrmDealRepositoryImpl.findDeals` fetches `stage`/`company`/`contact`/`owner`, but `CrmDealResponseDtoV2.contact` is a `CrmContactResponseDtoV2` carrying `company` and `owner`, both lazy on `CrmContact` → 2 extra SELECTs per deal row. `CrmDealRepositoryImpl.findByIdWithAssociations` only fetches `stage` and `owner`, so `getDealById` lazily loads 4 more.

Fix: add dedicated V2 fetch graphs (or `@EntityGraph`) covering exactly what the V2 DTOs serialize — for tasks: `company`, `contact.company`, `contact.owner`, `deal.stage`, `deal.owner`, `deal.company`, `deal.contact.company`, `deal.contact.owner`; for deals: `contact.company`, `contact.owner` (plus `company`/`contact` on the by-id query). Since these are all `ManyToOne`, adding them as LEFT fetch joins will not break the existing offset pagination.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmDealServiceImplV2.java
```diff
@@ -0,0 +1,109 @@
...
+
+		User currentUser = userService.getCurrentUser();
+		Long ownerId = CrmUtil.isCrmSalesRepresentative(currentUser) ? currentUser.getEmployee().getEmployeeId() : null;
+
+		Page<CrmDeal> dealsPage = crmDealDao.findDeals(filterDto, ownerId,
+				PageRequest.of(filterDto.getPage(), filterDto.getSize()));
+
+		List<CrmDealResponseDtoV2> deals = dealsPage.getContent()
+			.stream()
+			.map(crmMapperV2::crmDealToCrmDealResponseDtoV2)
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Behavioral regression vs V1: soft-deleted companies are no longer masked. `CrmDealServiceImpl.getDeals` maps each row through `CrmUtil.toDealResponseDto`, which calls `CrmUtil.hasDeletedCompany(deal)` and nulls `companyName` when the linked company has `isDeleted = true`. `CrmCompanyServiceImpl.deleteCompany` only flips `isDeleted` — it does not clear `crm_deal.company_id` or `crm_contact.company_id` — so deals created before the deletion keep pointing at it. This V2 method maps the entity straight through `crmMapperV2::crmDealToCrmDealResponseDtoV2`, so the full `CrmCompanyResponseDto` (name, industry, website, address, contactNumber) of a soft-deleted company is now returned, both in `company` and in the newly-exposed `contact.company`. The same applies to `getDealById` (line 84) and to `CrmTaskServiceImplV2`'s `contact.company` / `deal.company`.

Fix: add a V2 counterpart to `CrmUtil.toDealResponseDto` (e.g. `toDealResponseDtoV2(crmMapperV2, deal)`) that nulls `company` when `CrmUtil.hasDeletedCompany(deal)` and nulls `contact.company` when `CrmUtil.hasDeletedCompany(deal.getContact())`, and route all V2 deal/task mapping through it.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmDealServiceImplV2.java
```diff
@@ -0,0 +1,109 @@
...
+
+	private final CrmMapperV2 crmMapperV2;
+
+	private final UserService userService;
+
+	private final PageTransformer pageTransformer;
+
+	@Override
+	@Transactional(readOnly = true)
+	public ResponseEntityDto getDeals(CrmDealFilterDto filterDto) {
```
ThinuwanW, line 42: 🤖 **Claude** · 🟡 **Suggestion**

Read-path logic is duplicated rather than shared. `getDeals` here is a line-for-line copy of `CrmDealServiceImpl.getDeals` (lines 164-180) — same `userService.getCurrentUser()`, same `CrmUtil.isCrmSalesRepresentative` owner scoping, same `crmDealDao.findDeals(...)` call, same `PageTransformer` usage — differing only in the mapper. The same holds for all four read methods in `CrmTaskServiceImplV2` versus `CrmTaskServiceImpl`. This is exactly the drift risk the write path avoided by extracting `persistNewDeal`/`applyDealEdit`: a future change to owner scoping or filtering in V1 will silently not apply to V2 (and vice versa).

Fix: mirror the write-path approach — add entity-returning helpers to `CrmDealService`/`CrmTaskService` (e.g. `Page<CrmDeal> findDealsForCurrentUser(CrmDealFilterDto)`, `CrmDeal loadDealForView(Long id)`, `List<CrmTask> loadTasksForCurrentUser(CrmTaskFilterDto)`) and have both V1 and V2 impls call them, so the query + authorization logic lives in one place.

backend/src/main/java/com/skapp/community/crmplanner/controller/v2/CrmDealControllerV2.java
```diff
@@ -0,0 +1,64 @@
...
+@RequestMapping("/v2/crm/deal")
+@Tag(name = "CRM Deal Controller V2", description = "Operations related to CRM deals")
+public class CrmDealControllerV2 {
+
+	private final CrmDealServiceV2 crmDealService;
+
+	@Operation(summary = "Create a new deal",
+			description = "This endpoint creates a new CRM deal with the provided details.")
+	@PostMapping
+	@PreAuthorize("hasAnyRole('ROLE_CRM_SALES_REPRESENTATIVE')")
```
ThinuwanW, line 33: 🤖 **Claude** · 🟡 **Suggestion**

Inconsistent authorization expressions within the same new controller: `createDeal` (line 33), `getDeals` (line 41) and `editDeal` (line 49) use `hasAnyRole('ROLE_CRM_SALES_REPRESENTATIVE')` while `getDealById` (line 58) uses `hasRole('ROLE_CRM_SALES_REPRESENTATIVE')`. With a single role argument these are equivalent (and the `CRM_ADMIN > CRM_SALES_MANAGER > CRM_SALES_REPRESENTATIVE` hierarchy in `AuthorityServiceImpl` makes the effective access correct either way), but the inconsistency is copied from V1 and is worth cleaning up in the new file. `CrmTaskControllerV2` and `CrmContactControllerV2` both use `hasRole` uniformly — use `hasRole(...)` on all four methods here to match.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
```diff
@@ -98,6 +98,13 @@ public ResponseEntityDto checkDealNameExists(String name) {
...
 	@Transactional
 	public ResponseEntityDto createDeal(CrmDealCreateRequestDto requestDto) {
+		CrmDeal savedDeal = persistNewDeal(requestDto);
+		return new ResponseEntityDto(false, crmMapper.crmDealToCrmDealResponseDto(savedDeal));
+	}
+
+	@Override
+	@Transactional
+	public CrmDeal persistNewDeal(CrmDealCreateRequestDto requestDto) {
 		log.info("createDeal: creating deal with name={}", requestDto.getName());
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Log prefixes no longer match their enclosing method after the extraction. `persistNewDeal` logs `"createDeal: creating deal with name={}"` (line 108) and `"createDeal: deal created with id={}"` (line 154); `applyDealEdit` logs `"editDeal: execution started/ended"` (lines 433, 506); `CrmTaskServiceImpl.persistNewTask` logs `"createTask: ..."` (lines 162, 206) and `applyTaskEdit` logs `"editTask: ..."` (lines 224, 301). Combined with the V2 wrappers, which log their own `"createDeal: execution started"` / `"editTask: execution started"`, a single V2 request now emits two interleaved `createDeal:`-prefixed start lines from two different classes, which makes log tracing ambiguous. Rename the prefixes in the extracted methods to `persistNewDeal:`, `applyDealEdit:`, `persistNewTask:`, `applyTaskEdit:` per the `log.info("methodName: execution started")` convention.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmDealServiceImplV2.java
```diff
@@ -0,0 +1,109 @@
...
+		log.info("getDeals: execution started");
+
+		User currentUser = userService.getCurrentUser();
+		Long ownerId = CrmUtil.isCrmSalesRepresentative(currentUser) ? currentUser.getEmployee().getEmployeeId() : null;
+
+		Page<CrmDeal> dealsPage = crmDealDao.findDeals(filterDto, ownerId,
+				PageRequest.of(filterDto.getPage(), filterDto.getSize()));
```
ThinuwanW, line 49: 🤖 **Claude** · 🟡 **Suggestion**

`PageRequest.of(filterDto.getPage(), filterDto.getSize())` is called on unvalidated client input. `CrmDealFilterDto.page`/`size` (and `CrmTaskCompletedFilterDto`, `CrmTaskRelatedFilterDto`) are plain `int` query params with no bounds, so `?page=-1` or `?size=0` throws `IllegalArgumentException`, which is not a `ModuleException` and falls through to the generic `@ExceptionHandler(Exception.class)` in `GlobalExceptionHandler` → 500 instead of a 400 error response. `?size=1000000` produces an unbounded query, which is especially costly given the nested associations the V2 DTOs pull in. This exists in V1 too, but this PR newly exposes it on four V2 endpoints (`CrmTaskServiceImplV2` lines 71 and 91 have the same issue). Add `@Min(0)`/`@Min(1)`+`@Max(...)` on the filter DTOs with `@Valid` on the controller params, or clamp the values before building the `PageRequest`.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmDealControllerV2IntegrationTest.java
```diff
@@ -0,0 +1,249 @@
+package com.skapp.community.crmplanner.controller.v2;
```
ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

`PATCH /v2/crm/deal/{id}` (`CrmDealControllerV2.editDeal`) has no integration test — the test class only exercises POST, GET by id and GET list (no `patch` import at all). The task V2 test covers `editTask`, so the deal side is an unexplained gap. Add at least: a happy-path edit asserting the V2 response shape (embedded `stage`/`owner`/`company`/`contact.company`), and a sales-rep-editing-another-owner's-deal case asserting `CRM_ERROR_DEAL_EDIT_DENIED`. A forbidden-role test for the deal endpoints (the task test has `createTask_WithoutCrmRole_ReturnsForbidden`) would also be worth adding.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmTaskServiceImplV2.java
```diff
@@ -0,0 +1,151 @@
...
+	public ResponseEntityDto editTask(Long id, CrmTaskEditRequestDto requestDto) {
+		log.info("editTask: execution started");
+
+		CrmTask updatedTask = crmTaskService.applyTaskEdit(id, requestDto);
+
+		log.info("editTask: execution ended");
+		return new ResponseEntityDto(false, crmMapperV2.crmTaskToCrmTaskResponseDtoV2(updatedTask));
+	}
+
+	private PageDto buildPageDto(Page<CrmTask> taskPage) {
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`buildPageDto` hand-rolls what `PageTransformer` already does (`setCurrentPage`/`setTotalItems`/`setTotalPages` from a `Page`), and `CrmDealServiceImplV2` in the very same PR uses `pageTransformer.transform(...)` for the identical job. Inject `PageTransformer` here and keep only the item-mapping in the helper, so the two new V2 services are consistent.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmDealServiceImplV2.java
```diff
@@ -0,0 +1,109 @@
...
+
+		List<CrmDealResponseDtoV2> deals = dealsPage.getContent()
+			.stream()
+			.map(crmMapperV2::crmDealToCrmDealResponseDtoV2)
+			.toList();
+
+		PageDto pageDto = pageTransformer.transform(dealsPage);
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Pagination metadata is silently wrong for any empty page. `PageTransformer.transform` short-circuits on `page == null || page.isEmpty()` and returns a default `PageDto` (currentPage=0, totalItems=0L, totalPages=0) without ever reading the `Page`. So `GET /v2/crm/deal?page=3&size=10` against 15 matching deals returns `items: [], totalItems: 0, totalPages: 0, currentPage: 0` — a client paging forward cannot tell 'past the end' from 'no data at all', and cannot render a correct pager. This also makes the two new V2 list endpoints in this same PR disagree: `CrmTaskServiceImplV2.buildPageDto` sets `currentPage`/`totalItems`/`totalPages` from the `Page` unconditionally, so `/v2/crm/task/completed` reports the real totals for an out-of-range page while `/v2/crm/deal` reports zeros. Note this cuts against consolidating on `PageTransformer`: fix by populating the three fields directly from `dealsPage` here (as the task service does), or fix `PageTransformer` to only skip `setItems` rather than returning early. Pick one and use it in both new V2 services so the metadata contract is identical.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmTaskServiceImpl.java
```diff
@@ -152,6 +152,13 @@ public ResponseEntityDto getRelatedTasks(CrmTaskRelatedFilterDto filterDto) {
 	@Override
 	@Transactional
 	public ResponseEntityDto createTask(CrmTaskCreateRequestDto requestDto) {
+		CrmTask savedTask = persistNewTask(requestDto);
+		return new ResponseEntityDto(false, crmMapper.crmTaskToCrmTaskResponseDto(savedTask));
+	}
+
+	@Override
+	@Transactional
+	public CrmTask persistNewTask(CrmTaskCreateRequestDto requestDto) {
```
ThinuwanW, line 161: 🤖 **Claude** · 🔴 **Important**

`persistNewTask` resolves the task type with `crmTaskTypeDao.getReferenceById(requestDto.getTypeId())` after only `CrmValidations.validateTaskTypeId`, which is a null-check. A request with a `typeId` that does not exist therefore passes validation and fails later — either as a `crm_task.FK_crm_task_crm_task_type_type_id` violation when the IDENTITY insert is forced (`savedTask.getId()` on the next line), or as `EntityNotFoundException` when the proxy is dereferenced. Either way the client gets a 500 instead of the established `ModuleException` -> 400 `ErrorResponse`. The new V2 endpoint makes this more reachable: `CrmTaskResponseDtoV2.type` is a full `CrmTaskTypeResponseDto`, so response mapping now always dereferences that proxy. `applyTaskEdit` (line 241) already does this correctly with `crmTaskTypeDao.findById(...).orElseThrow(() -> new ModuleException(CrmMessageConstant.CRM_ERROR_TASK_TYPE_NOT_FOUND))` — create and edit should not disagree. Change `persistNewTask` to use the same `findById(...).orElseThrow(...)`.

backend/src/main/java/com/skapp/community/crmplanner/service/CrmDealService.java
```diff
@@ -12,6 +13,8 @@ public interface CrmDealService {
 
 	ResponseEntityDto createDeal(CrmDealCreateRequestDto requestDto);
 
+	CrmDeal persistNewDeal(CrmDealCreateRequestDto requestDto);
```
ThinuwanW, line 16: 🤖 **Claude** · 🟡 **Suggestion**

The four new interface methods (`persistNewDeal`/`applyDealEdit` here, `persistNewTask`/`applyTaskEdit` in `CrmTaskService`) return managed JPA entities whose associations are all `FetchType.LAZY`, and none of the fetch paths the V2 mappers walk (deal.contact.company, deal.contact.owner, task.company, task.deal.stage, ...) are initialized before returning. This only works because each V2 caller happens to declare `@Transactional` itself; the interface carries no indication of that requirement. Any future caller that invokes these outside a transaction gets `LazyInitializationException` at mapping time rather than a compile or startup error. At minimum add Javadoc stating 'must be called within an active transaction; returns a managed entity with lazy associations', or have the V2 services re-read via the existing `findByIdWithAssociations` before mapping so the contract does not depend on caller-side annotations.

backend/src/main/java/com/skapp/community/crmplanner/controller/v2/CrmDealControllerV2.java
```diff
@@ -0,0 +1,64 @@
...
+	public ResponseEntity<ResponseEntityDto> createDeal(@RequestBody CrmDealCreateRequestDto crmDealCreateRequestDto) {
```
ThinuwanW, line 34: 🤖 **Claude** · 🟡 **Suggestion**

None of the four new write handlers annotate their `@RequestBody` with `@Valid` — `createDeal` (line 34) and `editDeal` (line 51) here, `createTask` and `editTask` in `CrmTaskControllerV2`. Bean validation is therefore never triggered on the new endpoints. Compounding this, `CrmDealCreateRequestDto`, `CrmDealEditRequestDto`, `CrmTaskCreateRequestDto` and `CrmTaskEditRequestDto` carry no constraint annotations at all (no `@NotBlank` on `name`, no `@NotNull` on `stageId`/`contactId`/`typeId`), so all validation lives imperatively in `CrmValidations`. Since V2 is a new contract, this is the cheap moment to add `@Valid` plus the corresponding field constraints (message keys from `CrmMessageConstant`) rather than carrying the imperative-only pattern forward.

backend/src/main/java/com/skapp/community/crmplanner/controller/v2/CrmDealControllerV2.java
```diff
@@ -0,0 +1,64 @@
+package com.skapp.community.crmplanner.controller.v2;
```
ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

The V2 surface is a partial mirror of V1, which forces clients to straddle both versions. `CrmDealControllerV2` omits `DELETE /{id}` (V1 has it, gated on the stricter `ROLE_CRM_SALES_MANAGER`) and `GET /exists` (the deal-name uniqueness pre-check clients call before create — without it a duplicate name only surfaces as a `CRM_ERROR_DEAL_EXISTS` failure from POST). `CrmTaskControllerV2` likewise omits `DELETE /{id}`. Either add the missing endpoints, or state in the `@Tag` description which operations intentionally remain on V1 so consumers know the V2 base path is not self-sufficient.

backend/src/main/java/com/skapp/community/crmplanner/payload/response/v2/CrmTaskResponseDtoV2.java
```diff
@@ -0,0 +1,38 @@
...
+
+	private CrmTaskTypeResponseDto type;
+
+	private CrmOwnerResponseDto owner;
+
+	private CrmCompanyResponseDto company;
+
+	private CrmContactResponseDtoV2 contact;
+
+	private CrmDealResponseDtoV2 deal;
```
ThinuwanW, line 42: 🤖 **Claude** · 🟡 **Suggestion**

Embedding the full `CrmDealResponseDtoV2` in every task makes the response graph much heavier than V1's and structurally redundant. Each task now carries `company` plus `contact` (with its own nested company and owner) plus `deal` (with stage, owner, company and again contact-with-company-and-owner), so the same contact and company are serialized two or three times per task. V1 used lookup-sized `CrmDealLookupResponseDto`/`CrmContactLookupResponseDto` for exactly this reason. This is worst on `GET /v2/crm/task`, which is unbounded — `CrmTaskFilterDto` has no page/size and `findTasks` applies no limit, so the payload scales with the rep's entire open-task list. Consider a lookup-shaped nested deal (id/name/stage) and paginating `/v2/crm/task` like the other two task endpoints.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmTaskControllerV2IntegrationTest.java
```diff
@@ -0,0 +1,341 @@
+package com.skapp.community.crmplanner.controller.v2;
```
ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Test gaps remaining after the PATCH-deal gap already noted. (1) No test covers the ownership guard on `GET /v2/crm/task/{id}` — `CrmTaskServiceImplV2.getTaskById` throws `CRM_ERROR_TASK_VIEW_DENIED` via `isOwnerRestrictedForRepresentative`, and that branch is unexercised (the deal test class does cover its equivalent, so the pattern is available to copy). (2) No test covers `GET /v2/crm/task/related` with neither `contactId` nor `dealId`, i.e. the `CrmValidations.validateRelatedTaskContextFilter` 400 path — worth having because `buildRelatedTaskPredicates` would otherwise emit an empty `cb.or(...)`. (3) `CrmDealControllerV2IntegrationTest` has no without-CRM-role 403 test, while this class does (`createTask_WithoutCrmRole_ReturnsForbidden`); the new deal endpoints' `@PreAuthorize` expressions are consequently never asserted.

backend/src/main/java/com/skapp/community/crmplanner/controller/v2/CrmDealControllerV2.java
```diff
@@ -0,0 +1,64 @@
...
+	@PreAuthorize("hasAnyRole('ROLE_CRM_SALES_REPRESENTATIVE')")
+	public ResponseEntity<ResponseEntityDto> createDeal(@RequestBody CrmDealCreateRequestDto crmDealCreateRequestDto) {
+		ResponseEntityDto response = crmDealService.createDeal(crmDealCreateRequestDto);
+		return new ResponseEntity<>(response, HttpStatus.CREATED);
+	}
+
+	@Operation(summary = "Get deals", description = "Returns a paginated list of CRM deals with optional filtering.")
+	@GetMapping
+	@PreAuthorize("hasAnyRole('ROLE_CRM_SALES_REPRESENTATIVE')")
+	public ResponseEntity<ResponseEntityDto> getDeals(CrmDealFilterDto crmDealFilterDto) {
```
ThinuwanW, line 42: 🤖 **Claude** · 🟣 **Nit**

`getDeals` binds `CrmDealFilterDto`, which declares `sortOrder` (default `Sort.Direction.ASC`) and `sortKey` (default `CrmDealSort.STAGE_ORDER`), but `CrmDealRepositoryImpl.findDeals` never reads either field — ordering is hard-coded to `stage.orderIndex, deal.orderIndex, deal.id`. These are dead parameters, and publishing them on a brand-new V2 contract (they appear in the generated OpenAPI schema) advertises sorting that silently does nothing. Either wire them into `findDeals` or drop them from the DTO before V2 consumers start sending them.

backend/src/main/java/com/skapp/community/crmplanner/payload/response/v2/CrmTaskResponseDtoV2.java
```diff
@@ -0,0 +1,38 @@
...
+
+}
```
ThinuwanW, line 44: 🤖 **Claude** · 🟣 **Nit**

`CrmTaskResponseDtoV2` has no `lastModifiedDate`, yet `crmTaskDao.findCompletedTasks` orders results by exactly that column (`Auditable_.lastModifiedDate` desc). A client consuming `/v2/crm/task/completed` receives rows in an order it cannot explain or display, and has no 'completed at' timestamp. The sibling `CrmContactResponseDtoV2` does expose `lastModifiedDate`, so the V2 DTO family is inconsistent here. Consider adding it (the field comes free from `Auditable`).

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmTaskControllerV2IntegrationTest.java
```diff
@@ -0,0 +1,341 @@
...
+		MvcResult result = performPatchRequest(task.getId(), dto).andDo(print())
+			.andExpect(status().isOk())
+			.andExpect(jsonPath(STATUS_PATH).value(STATUS_SUCCESSFUL))
+			.andExpect(jsonPath(RESULTS_0_PATH + "['name']").value("Updated Task V2"))
+			.andReturn();
+
+		Long updatedId = ((Number) JsonPath.read(result.getResponse().getContentAsString(), "$.results[0].id"))
+			.longValue();
+		CrmTask updated = crmTaskDao.findById(updatedId).orElseThrow();
+		org.junit.jupiter.api.Assertions.assertEquals("Updated Task V2", updated.getName());
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Two small style slips in the new test class. `editTask_UpdateName_ReturnsOk` calls `org.junit.jupiter.api.Assertions.assertEquals` fully qualified inline instead of statically importing `assertEquals` like the rest of the suite's matchers. Separately, `performGetRelatedRequest(Long contactId)` (line 153) shadows the instance field `contactId` set in `@BeforeEach`, so the helper reads correctly only by coincidence — rename the parameter (e.g. `targetContactId`) or drop it and use the field.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmTaskControllerV2IntegrationTest.java
anusham-rc: Added a `getTaskById` view-denied test (sales rep viewing another owners task) and a `getRelatedTasks` without contactId/dealId 400 test. The deal without-CRM-role 403 is covered in the deal test class.

backend/src/main/java/com/skapp/community/crmplanner/payload/response/v2/CrmTaskResponseDtoV2.java
anusham-rc: Fixed. Added `lastModifiedDate` to `CrmTaskResponseDtoV2` (mapped from `Auditable`).

backend/src/main/java/com/skapp/community/crmplanner/controller/v2/CrmDealControllerV2.java
anusham-rc: Noted. The omitted endpoints (`DELETE`, deal `GET /exists`) are intentional for this PR; consumers still use the V1 ones. Documenting the intended V2 surface is a reasonable follow-up rather than adding endpoints here.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmDealServiceImplV2.java
anusham-rc: Valid consistency gap, but extracting shared read+authorization helpers is a larger refactor; deferring to keep this PR scoped to the review fixes. The write path already shares via `persistNew*`/`apply*Edit`.

backend/src/main/java/com/skapp/community/crmplanner/controller/v2/CrmDealControllerV2.java
anusham-rc: The frontend does send `sortKey`/`sortOrder` (DealsSection), but `findDeals` has always ignored them (fixed ordering) -- pre-existing V1 behavior. Wiring real sorting is a feature, not part of this response-types PR; leaving as-is for now.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmTaskControllerV2IntegrationTest.java
anusham-rc: Fixed. Statically imported `assertEquals`, and renamed the `performGetRelatedRequest` parameter to `targetContactId` so it no longer shadows the instance field.

backend/src/main/java/com/skapp/community/crmplanner/controller/v2/CrmDealControllerV2.java
anusham-rc: Fixed. Made `CrmDealControllerV2` internally consistent using `hasAnyRole(...)` on all four methods (matching the V1 deal controller). `hasRole` vs `hasAnyRole` with a single role is equivalent.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmDealServiceImplV2.java
anusham-rc: Fixed. `getDeals` no longer uses `PageTransformer` (which zeroes totals on an empty/out-of-range page); it populates the three fields from `dealsPage` directly, matching `CrmTaskServiceImplV2`. Both v2 list endpoints now report consistent metadata.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmTaskServiceImplV2.java
anusham-rc: Resolved by making both consistent: `getDeals` now populates `currentPage`/`totalItems`/`totalPages` from the `Page` directly (same as the task service) instead of `PageTransformer` -- see the empty-page comment for why.

backend/src/main/java/com/skapp/community/crmplanner/controller/v2/CrmDealControllerV2.java
anusham-rc: Mirrors V1 (validation lives in `CrmValidations`, DTOs carry no constraints). Adding `@Valid` plus field constraints is a broader convergence change, better scoped to its own PR.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmDealServiceImplV2.java
anusham-rc: Fixed. Added `CrmUtil.toDealResponseDtoV2` / `toTaskResponseDtoV2` that null the company (and nested `contact.company` / `deal.company`) when the source company is soft-deleted, and routed all v2 deal/task mapping through them. Added a `getDealById` masking test.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
anusham-rc: Fixed. Renamed the log prefixes in the extracted methods to `persistNewDeal:` / `applyDealEdit:` / `persistNewTask:` / `applyTaskEdit:`.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmTaskServiceImplV2.java
anusham-rc: Acknowledged as a real perf regression. Deferring the fetch-graph work to a follow-up so this PR stays focused on correctness -- it touches shared repo methods (`findDeals`, `findByIdWithAssociations`, `applyFetchGraph`) used by V1 too, so it warrants its own change.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmDealServiceImplV2.java
anusham-rc: The filter DTOs already default `page=0`/`size=10`, so the no-param case is fine. Rejecting explicit out-of-range values (`page=-1`, `size=0`) with 400 and clamping a max size is a codebase-wide concern (V1 shares it), better handled as a dedicated hardening ticket.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmDealControllerV2IntegrationTest.java
anusham-rc: Added. `editDeal` happy-path (asserting the embedded stage/owner/company/contact.company), a sales-rep-editing-another-owners-deal edit-denied case, a without-CRM-role 403, and a soft-deleted-company masking test.

backend/src/main/java/com/skapp/community/crmplanner/service/CrmDealService.java
anusham-rc: This mirrors the existing V1 pattern and both v2 wrappers are `@Transactional`, so mapping is safe. Narrowing the contract (return id, or re-read via `findByIdWithAssociations`) is a larger change; deferring.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmTaskServiceImpl.java
anusham-rc: Fixed. `persistNewTask` now resolves the type via `crmTaskTypeDao.findById(...).orElseThrow(() -> new ModuleException(CRM_ERROR_TASK_TYPE_NOT_FOUND))`, matching `applyTaskEdit`, so an unknown typeId returns 400 instead of 500. Updated the V1 test that asserted the old 500.

backend/src/main/java/com/skapp/community/crmplanner/payload/response/v2/CrmTaskResponseDtoV2.java
anusham-rc: Fair point -- the embedded full deal duplicates contact/company. Deferring the slimming for now; there is no frontend consumer yet, so we can revisit the nested shape (lookup-sized deal) alongside the fetch-graph work.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmDealServiceImplV2.java
```diff
@@ -0,0 +1,109 @@
...
+
+	@Override
+	@Transactional(readOnly = true)
+	public ResponseEntityDto getDeals(CrmDealFilterDto filterDto) {
+		log.info("getDeals: execution started");
+
+		User currentUser = userService.getCurrentUser();
+		Long ownerId = CrmUtil.isCrmSalesRepresentative(currentUser) ? currentUser.getEmployee().getEmployeeId() : null;
+
+		Page<CrmDeal> dealsPage = crmDealDao.findDeals(filterDto, ownerId,
```
akilarootcode: do pagination impl also from criteria query

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmTaskServiceImplV2.java
```diff
@@ -0,0 +1,151 @@
...
+	public ResponseEntityDto editTask(Long id, CrmTaskEditRequestDto requestDto) {
+		log.info("editTask: execution started");
+
+		CrmTask updatedTask = crmTaskService.applyTaskEdit(id, requestDto);
+
+		log.info("editTask: execution ended");
+		return new ResponseEntityDto(false, CrmUtil.toTaskResponseDtoV2(crmMapperV2, updatedTask));
+	}
+
+	private PageDto buildPageDto(Page<CrmTask> taskPage) {
```
akilarootcode: remove this and handle pagination inside criteria query

akilarootcode (changes requested):

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmDealServiceImplV2.java
anusham-rc: Done. `getDeals` now uses a new `findDealsV2` that projects `Page<CrmDealResponseDtoV2>` directly from the criteria query, no more entity fetch and service-side mapping.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmTaskServiceImplV2.java
anusham-rc: Done. Removed `buildPageDto`. `getTasks`/`getCompletedTasks`/`getRelatedTasks` now use new `findTasksV2`/`findCompletedTasksV2`/`findRelatedTasksV2` that project `Page<CrmTaskResponseDtoV2>` inside the criteria query.

akilarootcode (approved):

PR #1737, opened 2026-08-11
fix: People configuration tab issue

SanithuGunasena07:
Merging the two people configs together


ThinuwanW (commented):
## 🤖 Claude Code Review

The PR merges the enterprise Google Workspace sync settings into the community `PeopleConfigurations` component and drops the connection-gated People tab, plus widens the birthday-notification endpoints to `ROLE_SUPER_ADMIN` (which correctly matches the convention used across the other people-planner controllers). The backend change is fine, but the frontend consolidation introduces three production-breaking issues: a missing `src/fallback` stub for the enterprise-only `usePeopleConfigurationsForm` hook (community build/Jest resolve `~enterprise/*` to `src/fallback/*`), shared Cancel/Save labels moved to the `peopleEnterprise` namespace that isn't loaded in community mode, and the Google connection-status query losing its super-admin gate even though the backend endpoint is `ROLE_SUPER_ADMIN`-only. Several smaller UI/behaviour regressions around the relocated button row, toast collisions, and missing error states are also flagged.

Found **22** new issue(s): 🔴 7 important, 🟡 9 suggestion(s), 🟣 6 nit(s)

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
```diff
@@ -13,16 +18,42 @@ import {
   BirthdayNotificationConfigPatchType,
   BirthdayNotificationConfigType
 } from "~community/people/types/PeopleConfigTypes";
+import { useGetEnvironment } from "~enterprise/common/hooks/useGetEnvironment";
+import { usePeopleConfigurationsForm } from "~enterprise/configurations/hooks/usePeopleConfigurationsForm";
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`usePeopleConfigurationsForm` is imported from `~enterprise/configurations/hooks/...`, but there is no fallback stub. `tsconfig.json` maps `~enterprise/*` to `["./src/enterprise/*", "./src/fallback/*"]` and `jest.config.js` maps `^~enterprise(.*)$` to `<rootDir>/src/fallback$1` unconditionally, so in a community build (submodule absent) and in every Jest run this resolves to `src/fallback/configurations/hooks/usePeopleConfigurationsForm` — which does not exist (`src/fallback/configurations/` only contains `components/` and `utils/`). Result: module-not-found at build/test time. Note the sibling import on line 23 works precisely because `src/fallback/people/api/GoogleWorkspaceSyncApi.ts` exists, and every other enterprise hook consumed from `src/community` (useGetEnvironment, useS3Download, useTier, useGoogleAnalyticsEvent, useCalendarIntegrations, useFCMToken) has a stub. Fix: add `src/fallback/configurations/hooks/usePeopleConfigurationsForm.ts` returning a typed no-op, e.g. a `useFormik` instance with `{ isAutoSyncEnabled: undefined, isSyncNotificationsEnabled: undefined }` and a no-op `onSubmit`, plus `isUpdating: false`, so `formik.dirty`, `formik.values`, `formik.initialValues`, `resetForm` and `handleSubmit` all exist.

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
```diff
@@ -158,31 +205,101 @@ const PeopleConfigurations: FC = () => {
...
+
+      <Box sx={googleWorkspaceStyles.buttonsContainer}>
+        <ButtonV2
+          variant="tertiary"
+          disabled={!isAnyChanged || isAnySubmitting}
+          onClick={handleCancelAll}
+          icon={<Icon name={IconName.CLOSE_ICON} />}
+          iconPosition="end"
+        >
+          {translateWorkspaceText(["peopleConfiguration", "cancelButtonText"])}
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The shared Cancel/Save button labels were moved from `configurations.people.buttons.cancel|save` to `peopleEnterprise.peopleConfiguration.cancelButtonText|saveButtonText`, but the `peopleEnterprise` namespace only exists in the enterprise bundle (`src/enterprise/common/assets/languages/english/peopleEnterprise.json`). `i18n.ts` loads `enterpriseEnglish` only when `NEXT_PUBLIC_MODE === "enterprise"`, and `useTranslator` returns the raw key on a miss (`t(key) || ""`). In community mode these two buttons will literally render `peopleEnterprise.peopleConfiguration.cancelButtonText` / `...saveButtonText`. Since this button row is now shared by both modes, keep the community keys: restore `const translateButton = useTranslator("configurations", "people", "buttons")` and use `translateButton(["cancel"])` / `translateButton(["save"])` (both keys still exist in `configurations.json`). Reserve `translateWorkspaceText` for the enterprise-only section.

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
```diff
@@ -13,16 +18,42 @@ import {
...
     "birthdayNotificationSection"
   );
-  const translateButton = useTranslator("configurations", "people", "buttons");
+  const translateWorkspaceText = useTranslator("peopleEnterprise");
   const { setToastMessage } = useToast();
 
+  const isEnterprise = useGetEnvironment() === appModes.ENTERPRISE;
+
+  const { data: connectionStatus, isLoading: isConnectionStatusLoading } =
+    useGetGoogleConnectionStatus(isEnterprise);
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The connection-status query lost its super-admin gate. The removed page code called `useGetGoogleConnectionStatus(isEnterprise && isSuperAdmin)`; here it is only `useGetGoogleConnectionStatus(isEnterprise)`. The backing endpoint `/v1/ep/integrations/google/status` is annotated `@PreAuthorize("hasRole('ROLE_SUPER_ADMIN')")` (EpGoogleWorkspaceOAuthController), while the People configuration tab is visible to `AdminTypes.PEOPLE_ADMIN` as well (configurationTabsUtil `requiredRoles: [SUPER_ADMIN, PEOPLE_ADMIN]`). So an enterprise people-admin opening this tab fires a request that 403s (with React Query's default retries), and because there is no error branch the UI falls through to the "Connect Google Workspace from the People Directory" copy — misleading, since that user could never manage it anyway. Fix: pull `isSuperAdmin` from `useSessionData()` and gate both the query (`isEnterprise && isSuperAdmin`) and the rendering of the whole Google Workspace `Box` on it. The `useUpdateSyncSettings` PATCH is likewise super-admin-only.

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
```diff
@@ -158,31 +205,101 @@ const PeopleConfigurations: FC = () => {
...
+                    )
+                  }
+                />
+              </Box>
+            )}
+          </Box>
         </>
       )}
+
+      <Box sx={googleWorkspaceStyles.buttonsContainer}>
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The button row was moved out of the `{!isError && config && (...)}` guard, so it now renders unconditionally. During load the Tailwind skeleton block (which already contains two placeholder button shapes, `h-10 w-28` and `h-10 w-36`) renders at the same time as this real row — the user sees a duplicated pair of buttons on every page load. It also renders on the error path, where there is nothing to save or cancel. Fix: render this row only when there is something actionable, e.g. `{((!isError && config) || (isEnterprise && isGoogleWorkspaceConnected)) && (...)}`, or move it back inside a guard that excludes the loading/error states.

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
```diff
@@ -82,6 +113,22 @@ const PeopleConfigurations: FC = () => {
...
+  const handleCancelAll = () => {
+    if (isFormChanged) handleCancel();
+    if (isWorkspaceChanged) {
+      workspaceFormik.resetForm({ values: workspaceFormik.initialValues });
+    }
+  };
+
+  const handleSaveAll = () => {
+    if (isFormChanged) handleSave();
+    if (isWorkspaceChanged) workspaceFormik.handleSubmit();
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`handleSaveAll` fires two independent mutations (birthday PATCH and the workspace formik submit) without coordinating their results. Both success/error handlers call `setToastMessage`, and `ToastProvider` holds a single `toastMessage` state slot — the second result overwrites the first. Concretely: user edits both sections, hits Save, the birthday PATCH fails while the workspace PATCH succeeds, and the success toast replaces the error toast, so the user believes both saved when the birthday config did not. Fix: coordinate the two calls (e.g. `mutateAsync`/`Promise.allSettled` and emit one combined toast), or suppress the per-mutation success toast when a combined save is in flight and only surface a single aggregated success/failure message.

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
```diff
@@ -158,31 +205,101 @@ const PeopleConfigurations: FC = () => {
...
+                "peopleConfiguration",
+                "googleWorkspaceDescription"
+              ])}
+            </Typography>
+
+            {isConnectionStatusLoading ? (
+              <Box sx={{ display: "flex", justifyContent: "center", py: 4 }}>
+                <Spinner />
+              </Box>
+            ) : !isGoogleWorkspaceConnected ? (
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`useGetGoogleConnectionStatus` exposes only `data` and `isLoading` here — there is no `isError` handling. Any failure (403, network, 5xx) leaves `connectionStatus` undefined, so the component silently falls into the `!isGoogleWorkspaceConnected` branch and instructs the admin to connect from the People Directory even when the integration is actually connected. Destructure `isError` and render a distinct error message (with the existing toast/error copy conventions) instead of conflating failure with "not connected".

frontend/pages/community/configurations/index.tsx
```diff
@@ -35,49 +28,18 @@ const Configurations: NextPage = () => {
...
 
   const handleTabChange = (id: string) => {
     setActiveTab(id);
-    const basePath = router.asPath.split("?")[0];
-    router.replace(
-      { pathname: router.pathname, query: { ...router.query, tab: id } },
-      `${basePath}?tab=${id}`,
-      { shallow: true, scroll: false }
-    );
+    replaceTabQueryParam(router.asPath, id);
```
ThinuwanW, line 42: 🤖 **Claude** · 🟡 **Suggestion**

`replaceTabQueryParam(router.asPath, id)` writes `?tab=<id>` to the URL, but this page never reads it back — `activeTab` is initialised from `visibleTabs[0]?.id` only. The settings page that this util was borrowed from pairs the write with a read (`useEffect` on `router.isReady` that sets `activeTab` from `router.query.tab` when it matches a visible tab). Without the symmetric read, deep links and refreshes always land on the first tab, so the query param is write-only. Add the same `router.isReady` effect here.

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
```diff
@@ -158,31 +205,101 @@ const PeopleConfigurations: FC = () => {
...
-              icon={<SaveIcon />}
-              iconPosition="end"
-              disabled={!isFormChanged || isPending}
-              onClick={handleSave}
+      {isEnterprise && (
+        <>
+          <hr className="w-full border-t border-secondary-accent" />
+
+          <Box>
+            <Typography variant="h4" sx={googleWorkspaceStyles.sectionTitle}>
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Styling and heading semantics are inconsistent with the rest of this component. The existing birthday section is pure Tailwind (`<h2 className="subtitle2">`, `<p className="body1">`, skeleton via `animate-pulse`), while the new section introduces MUI `Box`/`Typography` with `sx` objects, an MUI `Spinner` for loading, and a `variant="h4"` title — jumping from `h2` to `h4` skips a heading level (a11y) and gives the two sibling sections visually/semantically different headings. Either convert the new section to the same Tailwind markup used above, or state explicitly why MUI is needed here; at minimum use `variant="h2"`/`component="h2"` so the heading order is correct, and keep one loading treatment for both sections.

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
```diff
@@ -13,16 +18,42 @@ import {
   BirthdayNotificationConfigPatchType,
   BirthdayNotificationConfigType
 } from "~community/people/types/PeopleConfigTypes";
+import { useGetEnvironment } from "~enterprise/common/hooks/useGetEnvironment";
+import { usePeopleConfigurationsForm } from "~enterprise/configurations/hooks/usePeopleConfigurationsForm";
+import { useGetGoogleConnectionStatus } from "~enterprise/people/api/GoogleWorkspaceSyncApi";
+
+const googleWorkspaceStyles = {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`googleWorkspaceStyles` also holds `buttonsContainer`, which styles the shared (community + enterprise) button row — the name is misleading now that the row is not Google-Workspace-specific. The module convention is a colocated `styles.ts` (see `InvoiceConfigurations/styles.ts`, `signConfigurations/styles.ts`) rather than a module-level literal. Also, the spinner wrapper still uses an inline `sx={{ display: "flex", justifyContent: "center", py: 4 }}` instead of going through the same object, so the extraction is only half done. Move these into `./styles.ts` and rename the shared entry (e.g. `actionButtonsContainer`).

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
```diff
@@ -158,31 +205,101 @@ const PeopleConfigurations: FC = () => {
               </>
             )}
           </div>
+        </>
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

After the button row was moved out, this fragment (`<>...</>`) wraps a single `<div className="flex flex-col gap-6">`. Drop the fragment and return the `div` directly from the `{!isError && config && ...}` branch.

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
```diff
@@ -158,31 +205,101 @@ const PeopleConfigurations: FC = () => {
...
+          icon={<Icon name={IconName.CLOSE_ICON} />}
+          iconPosition="end"
+        >
+          {translateWorkspaceText(["peopleConfiguration", "cancelButtonText"])}
+        </ButtonV2>
+        <ButtonV2
+          disabled={!isAnyChanged || isAnySubmitting}
+          isLoading={isAnySubmitting}
+          onClick={handleSaveAll}
+          icon={<Icon name={IconName.RIGHT_ARROW_ICON} />}
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Two incidental changes to the button presentation: the Save icon changed from `SaveIcon` (skapp-ui) to `IconName.RIGHT_ARROW_ICON`, and `size="md"` was dropped from both buttons. The arrow may be deliberate (the enterprise `InvoiceConfigurations` uses `IconName.FORWARD_ARROW` for its save action), but if so consider matching that exact icon for consistency; and confirm that omitting `size` yields the same `md` sizing the previous buttons had, otherwise the button row will render at a different size than before.

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
```diff
@@ -1,10 +1,15 @@
-import { ButtonV2, CloseIcon, SaveIcon } from "@rootcodelabs/skapp-ui";
+import { Box, Typography } from "@mui/material";
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

There is no `PeopleConfigurations.test.tsx` for this component even though this PR gives it substantially more behaviour: an enterprise-only branch, a merged save/cancel path over two independent forms, and three distinct render states for the connection status. Add a test file (with the `MockTheme` wrapper and `@testing-library/react` role-based queries) covering: community mode hides the Google Workspace section, save/cancel disabled state derived from `isAnyChanged`, and that `handleSaveAll` only submits the form(s) that are actually dirty. Note this test will not run until the `src/fallback` stub from the first finding exists, since Jest always resolves `~enterprise` to `src/fallback`.

frontend/pages/community/configurations/index.tsx
ThinuwanW, line 42: 🤖 **Claude** · 🔴 **Important**

Behavioral regression in the tab-URL refactor. The removed code used `router.replace({pathname, query: {...router.query, tab: id}}, basePath+'?tab='+id, {shallow: true, scroll: false})`, which kept Next's router state in sync. `replaceTabQueryParam` (commonUtil.ts:611-616) instead calls `globalThis.history.replaceState(null, "", ...)` directly. Two concrete consequences: (1) `router.query.tab` and `router.asPath` are never updated, so any later `router.push`/`router.replace` that spreads `router.query` silently drops the tab, and any `useEffect` keyed on `router.query.tab` never fires — note the sibling settings page (pages/community/settings/index.tsx:66-72) *does* read `router.query.tab`, so this util is already fragile there; (2) passing `null` as the state object overwrites Next's own history entry (`{url, as, options, key}`) that its `popstate` handler relies on, so browser back/forward from this page can land on a state Next cannot reconcile. Fix: either restore `router.replace(..., {shallow: true, scroll: false})` here, or change `replaceTabQueryParam` to preserve the existing state (`globalThis.history.replaceState(globalThis.history.state, "", url)`).

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
```diff
@@ -158,31 +205,101 @@ const PeopleConfigurations: FC = () => {
...
-              {translateButton(["cancel"])}
-            </ButtonV2>
-            <ButtonV2
-              variant="primary"
-              size="md"
-              icon={<SaveIcon />}
-              iconPosition="end"
-              disabled={!isFormChanged || isPending}
-              onClick={handleSave}
+      {isEnterprise && (
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The whole Google Workspace block is gated only on `isEnterprise`, not on role. But the People tab's `requiredRoles` is `[AdminTypes.SUPER_ADMIN, AdminTypes.PEOPLE_ADMIN]` (community/configurations/utils/configurationTabsUtil.tsx:64), while every backing endpoint on `EpGoogleWorkspaceOAuthController` — `/status`, `/sync-settings`, `/disconnect` — is annotated `@PreAuthorize("hasRole('ROLE_SUPER_ADMIN')")`. So a People Admin who is not a Super Admin now renders this section, the status GET returns 403, `connectionStatus` stays undefined, and the UI falsely tells them "Connect Google Workspace from the People Directory to manage sync settings" even when the workspace *is* connected — and if they ever did reach the toggles the PATCH would 403. Gate the entire `{isEnterprise && (...)}` block (including the `<hr>`) on super-admin, e.g. `const { isSuperAdmin } = useSessionData()` and render on `isEnterprise && isSuperAdmin`, so the section matches the backend's authorization.

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
```diff
@@ -13,16 +18,42 @@ import {
...
-  const translateButton = useTranslator("configurations", "people", "buttons");
+  const translateWorkspaceText = useTranslator("peopleEnterprise");
   const { setToastMessage } = useToast();
 
+  const isEnterprise = useGetEnvironment() === appModes.ENTERPRISE;
+
+  const { data: connectionStatus, isLoading: isConnectionStatusLoading } =
+    useGetGoogleConnectionStatus(isEnterprise);
+  const isGoogleWorkspaceConnected = connectionStatus?.isConnected;
+  const { formik: workspaceFormik, isUpdating: isWorkspaceUpdating } =
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Missing paired change: `usePeopleConfigurationsForm(peopleConfigurationFormData, onDisconnected?)` takes a second `onDisconnected` callback, and this is now its only caller — it passes just the first argument. That makes the hook's `if (!data.isAutoSyncEnabled) { onDisconnected?.(); }` branch (usePeopleConfigurationsForm.ts:31-33) permanently dead, and the previous behaviour (page-level `handlePeopleDisconnected` moving the user off the People tab when auto-sync was switched off) is silently gone. Since the tab is no longer hidden, that behaviour is presumably intentional to drop — so delete the `onDisconnected` parameter and its call site from the enterprise hook rather than leaving an unreachable branch. Grep confirms no other caller (`onDisconnected` appears only inside that hook).

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
```diff
@@ -158,31 +205,101 @@ const PeopleConfigurations: FC = () => {
               </>
             )}
           </div>
+        </>
+      )}
```
ThinuwanW, line 186: 🤖 **Claude** · 🟡 **Suggestion**

No error UI for the birthday config. Both branches above are guarded by `!isError`, so when `useGetBirthdayNotificationConfig()` fails the entire birthday section silently disappears — no message, no retry — leaving only the heading/description and (now) a bare Cancel/Save row. Add an `{isError && (...)}` branch rendering a translated error message with a retry (`refetch`) affordance, using new keys under `configurations.people.birthdayNotificationSection` (the `toastMessages.errorTitle/errorDescription` keys already exist and could be reused for the copy).

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
```diff
@@ -158,31 +205,101 @@ const PeopleConfigurations: FC = () => {
...
+                {translateWorkspaceText([
+                  "peopleConfiguration",
+                  "notConnectedDescription"
+                ])}
+              </Typography>
+            ) : (
+              <Box sx={googleWorkspaceStyles.container}>
+                <SwitchRow
+                  labelId="google-workspace-auto-sync-enabled"
+                  label={translateWorkspaceText([
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

i18n key pulled from the wrong section. The auto-sync toggle label reads `peopleEnterprise.googleWorkspaceImport.autoSyncNewLabel` ("Automatically sync new members") — a key that belongs to the import wizard — while the switch right below it correctly uses `peopleEnterprise.peopleConfiguration.syncNotificationsLabel`. Editing the import-wizard copy will now silently change this configuration screen. Add a dedicated `peopleConfiguration.autoSyncLabel` in enterprise/common/assets/languages/english/peopleEnterprise.json (alongside `syncNotificationsLabel` at line 142) and reference that instead.

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Duplicated markup. This Cancel/Save row — `<Box sx={buttonsContainer}>` + tertiary `ButtonV2` with `<Icon name={IconName.CLOSE_ICON} />` `iconPosition="end"` + primary `ButtonV2` with an arrow icon — is a near-verbatim copy of enterprise/configurations/components/organisms/InvoiceConfigurations/InvoiceConfigurations.tsx:194-212 (which also defines an identical `buttonsContainer` style), and signConfigurations does the same. This PR makes it the third copy. Extract a shared molecule (e.g. `ConfigurationFormActions` under community/configurations/components/molecules) taking `onCancel`, `onSave`, `isDisabled`, `isLoading`, and label props, and reuse it in all three places — that also removes the need for a locally-defined `buttonsContainer`.

backend/src/main/java/com/skapp/community/peopleplanner/controller/v1/PeopleConfigController.java
```diff
@@ -26,7 +26,7 @@ public class PeopleConfigController {
 	@Operation(summary = "Get birthday notification configuration",
 			description = "Returns the organization-wide configuration controlling birthday notifications.")
 	@GetMapping(value = "/birthday-notifications")
-	@PreAuthorize("hasAnyRole('ROLE_PEOPLE_ADMIN')")
+	@PreAuthorize("hasAnyRole('ROLE_SUPER_ADMIN','ROLE_PEOPLE_ADMIN')")
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The widened role list is untested. In PeopleConfigControllerIntegrationTest, `adminToken` is employee 1, who per src/test/resources/data.sql:63 is *both* `is_super_admin = true` and `people_role = 'PEOPLE_ADMIN'`, so the happy-path tests passed before this change and prove nothing about the new `ROLE_SUPER_ADMIN` grant. The only negative test (`getBirthdayNotificationConfigs_WithoutPeopleAdminRole_ReturnsForbidden`) downgrades employee 2's `people_role` only — `ROLE_SUPER_ADMIN` comes from the separate `is_super_admin` flag (AuthorityServiceImpl.addSuperAdminAuthority), which for employee 2 is already `false`, so it still passes by accident. Add: (a) a test that a super admin whose `people_role` is `PEOPLE_EMPLOYEE` now gets 200 on both GET and PATCH, and (b) a forbidden test for the PATCH, which currently has no authorization test at all. Also rename the existing test/display name — "non people-admin" is now imprecise; it must also be a non-super-admin.

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
```diff
@@ -158,31 +205,101 @@ const PeopleConfigurations: FC = () => {
...
-            </ButtonV2>
-          </div>
+              {translateWorkspaceText([
+                "peopleConfiguration",
+                "googleWorkspaceDescription"
+              ])}
+            </Typography>
+
+            {isConnectionStatusLoading ? (
+              <Box sx={{ display: "flex", justifyContent: "center", py: 4 }}>
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Two small inconsistencies in this loading branch. (1) Accessibility: the sibling birthday skeleton carries `role="status"`, `aria-busy`, `aria-live="polite"` and a translated `aria-label`, but this Spinner wrapper has none, so screen-reader users get no announcement while the connection status loads — add `role="status"` / `aria-live="polite"` and a translated `aria-label`. (2) This is the only inline `sx` object literal in the file; every other style is hoisted into the style map at the top. Move it there (e.g. `spinnerWrapper`), and ideally move the whole map into a colocated `styles.ts`, which is the established pattern for configuration organisms (TimeConfigurations/styles.ts, RestrictedUserRolesModal/styles.ts, InvoiceConfigurations/styles.ts).

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
```diff
@@ -82,6 +113,22 @@ const PeopleConfigurations: FC = () => {
     updateConfig(config);
   };
 
+  const isWorkspaceChanged = isEnterprise && workspaceFormik.dirty;
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`isWorkspaceChanged` ignores whether the workspace is still connected. The toggles only render when `isGoogleWorkspaceConnected` is true, but formik state outlives them: if the connection-status query refetches to `isConnected: false` while the user has pending toggle changes (another admin disconnects, or the `useDisconnectGoogleWorkspace` invalidation fires from another tab), the switches unmount while `workspaceFormik.dirty` stays true — so the Save button remains enabled with no visible changed control and `handleSaveAll` PATCHes sync settings for a disconnected workspace. Consider `const isWorkspaceChanged = isEnterprise && !!isGoogleWorkspaceConnected && workspaceFormik.dirty;` and resetting the form when `isGoogleWorkspaceConnected` flips to false.

backend/src/main/java/com/skapp/community/peopleplanner/controller/v1/PeopleConfigController.java
```diff
@@ -35,7 +35,7 @@ public ResponseEntity<ResponseEntityDto> getBirthdayNotificationConfigs() {
 	@Operation(summary = "Update birthday notification configuration",
 			description = "Partially updates the organization-wide configuration controlling birthday notifications.")
 	@PatchMapping(value = "/birthday-notifications")
-	@PreAuthorize("hasAnyRole('ROLE_PEOPLE_ADMIN')")
+	@PreAuthorize("hasAnyRole('ROLE_SUPER_ADMIN','ROLE_PEOPLE_ADMIN')")
 	public ResponseEntity<ResponseEntityDto> updateBirthdayNotificationConfigs(
 			@RequestBody BirthdayNotificationConfigRequestDto requestDto) {
```
ThinuwanW, line 40: 🤖 **Claude** · 🟣 **Nit**

While touching the authorization on this endpoint: `@RequestBody BirthdayNotificationConfigRequestDto requestDto` has no `@Valid`, and the DTO itself (BirthdayNotificationConfigRequestDto) declares three all-nullable `Boolean` fields with no bean-validation annotations. The PATCH therefore accepts an empty `{}` body and returns 200 after a no-op write. Add `@Valid` for consistency with the rest of the codebase (e.g. AuthController.employeeResetPassword) and consider rejecting a payload where all three fields are null with a `ModuleException` backed by a `PeopleMessageConstant` key, so callers get a 400 rather than a silent success.

backend/src/main/java/com/skapp/community/peopleplanner/controller/v1/PeopleConfigController.java
```diff
@@ -35,7 +35,7 @@ public ResponseEntity<ResponseEntityDto> getBirthdayNotificationConfigs() {
 	@Operation(summary = "Update birthday notification configuration",
 			description = "Partially updates the organization-wide configuration controlling birthday notifications.")
 	@PatchMapping(value = "/birthday-notifications")
-	@PreAuthorize("hasAnyRole('ROLE_PEOPLE_ADMIN')")
+	@PreAuthorize("hasAnyRole('ROLE_SUPER_ADMIN','ROLE_PEOPLE_ADMIN')")
```
akilarootcode: remove super admin

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
SanithuGunasena07: fixed as part of previous comment

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
SanithuGunasena07: fixed as part of previous comment

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
SanithuGunasena07: fixed as part of previous comment

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
SanithuGunasena07: fixed as part of previous comment

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
SanithuGunasena07: not required

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
SanithuGunasena07: not required

frontend/pages/community/configurations/index.tsx
SanithuGunasena07: not required

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
SanithuGunasena07: Fixed the heading-skip only: added component="h2" to keep the h4 visual size but render correct semantic level, since the two sections are now siblings. Left the MUI-vs-Tailwind question alone  pre-existing choice, not a defect.

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
SanithuGunasena07: folded the spinner's inline sx into the local styles object.

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
SanithuGunasena07: droped

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
SanithuGunasena07: arrow is dilabarate

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
SanithuGunasena07: not requred

frontend/pages/community/configurations/index.tsx
SanithuGunasena07: pre exisitng

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
SanithuGunasena07: fixed as part of previous comment

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
SanithuGunasena07: fixed as part of previous comment

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
SanithuGunasena07: not required - pre existing

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
SanithuGunasena07: not required - pre existing

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
SanithuGunasena07: not required - pre existing

backend/src/main/java/com/skapp/community/peopleplanner/controller/v1/PeopleConfigController.java
SanithuGunasena07: not rquired

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
SanithuGunasena07: not required

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
SanithuGunasena07: not required

backend/src/main/java/com/skapp/community/peopleplanner/controller/v1/PeopleConfigController.java
SanithuGunasena07: not required - pre existing

frontend/src/community/common/components/molecules/Form/Form.tsx
```diff
@@ -1,13 +1,25 @@
-import { JSX, type ReactNode } from "react";
+import { type FormEvent, JSX, type ReactNode } from "react";
```
thusala: revert this file

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
```diff
@@ -124,66 +179,73 @@ const PeopleConfigurations: FC = () => {
...
+          <GoogleWorkspaceSyncSettings
+            actionsRef={workspaceActionsRef}
+            onStateChange={setWorkspaceState}
+          />
         </>
       )}
-    </div>
+
+      {isBirthdaySectionLoaded && (
+        <Box sx={buttonsContainerStyles}>
```
thusala: directly call the styling in here and remove buttonsContainerStyles  variable

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
```diff
@@ -13,16 +19,42 @@ import {
...
   const { setToastMessage } = useToast();
 
+  const isEnterprise = useGetEnvironment() === appModes.ENTERPRISE;
+  const { isSuperAdmin } = useSessionData();
+  const canManageGoogleWorkspace = isEnterprise && !!isSuperAdmin;
+
+  const workspaceActionsRef = useRef<GoogleWorkspaceSyncSettingsActions>({
+    save: () => {},
+    reset: () => {}
+  });
```
thusala: remove these refs

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
```diff
@@ -13,16 +19,42 @@ import {
...
+  const [workspaceState, setWorkspaceState] =
+    useState<GoogleWorkspaceSyncSettingsState>({
```
thusala: use this is formik or existing behaviour like other tabs

thusala (changes requested):

frontend/src/community/common/components/molecules/Form/Form.tsx
SanithuGunasena07: reverted

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
SanithuGunasena07: done

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
SanithuGunasena07: used formik

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
SanithuGunasena07: The ref is needed because Google Workspace's save/reset logic has to live inside its own enterprise-only component (for the fallback-stub/licensing separation this codebase relies on), but the Save button that triggers it lives in the shared parent. Since the parent can't call the child's internal formik directly

thusala: but u can simplify use and states for this? we might have to use global states for this, in order to handle correctly. refer the pattern we discussed earlier

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
thusala: when why do u need this use State? move into initalschema?

thusala (changes requested):

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
SanithuGunasena07: Removed the ref entirely, replaced it with global state. Extended useConfigurationStore with isPeopleWorkspaceSaveTriggered/isPeopleWorkspaceResetTriggered flags + setters (same shape as usePeopleStore's isUnsavedModalSaveButtonClicked pattern already used elsewhere in this codebase). The Save/Cancel buttons set the flags; GoogleWorkspaceSyncSettings watches them via useEffect, runs its own save/reset, then clears the flag. No useRef anywhere in these files now.

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
SanithuGunasena07: Removed useState. Moved isDirty/isSubmitting into useConfigurationStore (isPeopleWorkspaceDirty/isPeopleWorkspaceSubmitting) alongside the save/reset trigger flags from the ref fix. GoogleWorkspaceSyncSettings now writes directly into the store instead of calling an onStateChange prop, and takes no props at all the parent reads the same store fields directly. Both directions of parent and child communication now go through the same global state.

backend/src/main/java/com/skapp/community/peopleplanner/controller/v1/PeopleConfigController.java
shakila-kamalasena: Remove SUPER_ADMIN

backend/src/main/java/com/skapp/community/peopleplanner/controller/v1/PeopleConfigController.java
shakila-kamalasena: Remove SUPER_ADMIN

frontend/src/community/configurations/stores/configurationStore.ts
```diff
@@ -24,6 +28,26 @@ export const useConfigurationStore = create<
           ...state,
           moduleType: moduleType
         })),
+      setIsPeopleWorkspaceSaveTriggered: (isTriggered: boolean) =>
```
shakila-kamalasena: Remove this from community

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
```diff
@@ -56,45 +65,59 @@ const PeopleConfigurations: FC = () => {
...
   };
 
-  const handleCancel = () => setConfig(initialConfig);
-
-  const handleSave = () => {
-    if (!config || isPending) return;
-    updateConfig(config);
+  const handleSaveAll = async () => {
+    if (birthdayFormik.dirty) {
+      try {
```
shakila-kamalasena: remove try-catch

shakila-kamalasena (changes requested):

backend/src/main/java/com/skapp/community/peopleplanner/controller/v1/PeopleConfigController.java
SanithuGunasena07: removed

backend/src/main/java/com/skapp/community/peopleplanner/controller/v1/PeopleConfigController.java
SanithuGunasena07: removed

frontend/src/community/configurations/stores/configurationStore.ts
SanithuGunasena07: moved the workspace save/reset/dirty/submitting state out of the shared configurationStore.ts into a new dedicated useGoogleWorkspaceSyncSettingsStore under src/enterprise/people/store/, with a no-op fallback under src/fallback/people/store/ (same pattern already used for useUserLimitStore) so the community build still compiles without shipping any real Google Workspace state. PeopleConfigurations.tsx and GoogleWorkspaceSyncSettings.tsx now pull from that store instead.

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
SanithuGunasena07: removed

shakila-kamalasena (approved):

shakila-kamalasena (dismissed):

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
```diff
@@ -123,65 +142,67 @@ const PeopleConfigurations: FC = () => {
...
-                  tooltip={translateText(["organizationToggleTooltip"])}
-                  checked={config.isOrganizationWide}
-                  onChange={(checked) =>
-                    handleToggleChange("isOrganizationWide", checked)
-                  }
-                />
-              </>
-            )}
-          </div>
+      {!isError && data && (
```
akilarootcode: use proper name for data

akilarootcode (changes requested):

backend/src/main/java/com/skapp/community/peopleplanner/controller/v1/PeopleConfigController.java
SanithuGunasena07: already removed

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
SanithuGunasena07: Renamed data to birthdayNotificationConfig

akilarootcode (dismissed):

akilarootcode (approved):

PR #1738, opened 2026-08-11
fix: adjust dropdown width in DealNameStageSection and DealProperties…

YasinduRC:
…Sidebar for better responsiveness

### Problem
The stage dropdown's clickable/"inside" region was wider than the field itself. `Dropdown`
applies its `width` prop as an inline style on the trigger only, while its click-away check
tests the full-width root container. With `width="55%"`, the leftover 45% became an invisible
band that swallowed clicks at the field's row but counted as "outside" one row lower — so
clicking empty space next to the field behaved inconsistently.

### Fix
Let the container own the width and have the dropdown fill it (`width="100%"`) — the pattern
used by the other 15 `Dropdown` call sites in the codebase. Root, trigger and menu are now one
rectangle, so inside/outside matches what the user sees. Width set to the design's 219px.

- `DealPropertiesSidebar.tsx` — deal side panel stage field
- `DealNameStageSection.tsx` — add-deal side panel stage field

Also moved the loading skeleton inside the sized wrapper so the field no longer jumps width
when stages finish loading, and added `max-w-full` so the fixed 219px clamps instead of
overflowing below the `md` breakpoint.

### Testing
- Open the stage dropdown in both panels, click the empty space to the right of the field →
  closes cleanly (previously inconsistent).
- Selecting a stage, Escape, and Tab-out still work.
- 219px renders as specced at ≥768px; clamps to the column below that without horizontal scroll.

### Notes
The root cause is in `skapp-ui`'s `Dropdown` (`width` sizes the trigger, click-away tests the
root), so any consumer passing a sub-100% width hits this. Fixed at the call sites here since
those were the only two in the codebase; the upstream fix is tracked separately.

Two things to decide before you post it:

1. The store-slice changes are on this same branch (crmCompaniesSlice, crmContactsSlice, crmDealsSlice, crmLookupSlice, crmTasksSlice, crmUiSlice are all modified, plus the refactor/crm-715-Store-Slice-Normalizing commits). If this PR is meant to be just the dropdown fix, it needs its own branch off main — otherwise the description above understates what reviewers will see.
2. The 219px claim — I took it from your message, not from a design file. If the ticket/Figma link is handy, worth pasting it in place of "the design's 219px".
### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/frontend/src/community/crm/components/organisms/AddDealSidePanel/DealNameStageSection.tsx
+++ b/frontend/src/community/crm/components/organisms/AddDealSidePanel/DealNameStageSection.tsx
@@ -86,9 +86,9 @@ const DealNameStageSection: FC<DealNameStageSectionProps> = ({
         />
       </div>
       <div className="w-1/3 pt-6.5">
-        <div className="w-[219px] max-w-full">
+        <div className="w-full max-w-[13.688rem]">
           {isStagesLoading ? (
-            <MultipleSkeletons numOfSkeletons={1} height={38} />
+            <MultipleSkeletons numOfSkeletons={1} height={48} />
           ) : (
             <Dropdown
               options={stageOptions}
--- a/frontend/src/community/crm/components/organisms/AddDealSidePanel/DealNameStageSection.tsx
+++ b/frontend/src/community/crm/components/organisms/AddDealSidePanel/DealNameStageSection.tsx
@@ -86,22 +86,24 @@ const DealNameStageSection: FC<DealNameStageSectionProps> = ({
         />
       </div>
       <div className="w-1/3 pt-6.5">
-        {isStagesLoading ? (
-          <MultipleSkeletons numOfSkeletons={1} height={38} />
-        ) : (
-          <Dropdown
-            options={stageOptions}
-            value={formik.values.stageId}
-            onChange={(v) => formik.setFieldValue("stageId", v)}
-            variant={stageDropdownVariant}
-            className="rounded-lg"
-            width="55%"
-            placeholder={translateText(["placeholders", "stage"])}
-            required
-            errorMessage={stageErrorMessage}
-            ariaLabel={translateText(["ariaLabels", "stage"])}
-          />
-        )}
+        <div className="w-[219px] max-w-full">
+          {isStagesLoading ? (
+            <MultipleSkeletons numOfSkeletons={1} height={38} />
+          ) : (
+            <Dropdown
+              options={stageOptions}
+              value={formik.values.stageId}
+              onChange={(v) => formik.setFieldValue("stageId", v)}
+              variant={stageDropdownVariant}
+              className="rounded-lg"
+              width="100%"
+              placeholder={translateText(["placeholders", "stage"])}
+              required
+              errorMessage={stageErrorMessage}
+              ariaLabel={translateText(["ariaLabels", "stage"])}
+            />
+          )}
+        </div>
       </div>
     </div>
   );
--- a/frontend/src/community/crm/components/organisms/DealSidePanel/DealPropertiesSidebar.tsx
+++ b/frontend/src/community/crm/components/organisms/DealSidePanel/DealPropertiesSidebar.tsx
@@ -112,20 +112,22 @@ const DealPropertiesSidebar: FC<DealPropertiesSidebarProps> = ({
 
   return (
     <div className="w-1/3 flex flex-col gap-4 shrink-0">
-      {isStagesLoading ? (
-        <SkeletonShape className="h-9 w-full" />
-      ) : (
-        <Dropdown
-          options={stageOptions}
-          value={selectedStageId}
-          onChange={handleStageChange}
-          variant="primary"
-          className="rounded-lg"
-          width="55%"
-          placeholder={translateText(["placeholders", "stage"])}
-          ariaLabel={translateText(["ariaLabels", "stage"])}
-        />
-      )}
+      <div className="w-[219px] max-w-full">
+        {isStagesLoading ? (
+          <SkeletonShape className="h-9 w-full" />
+        ) : (
+          <Dropdown
+            options={stageOptions}
+            value={selectedStageId}
+            onChange={handleStageChange}
+            variant="primary"
+            className="rounded-lg"
+            width="100%"
+            placeholder={translateText(["placeholders", "stage"])}
+            ariaLabel={translateText(["ariaLabels", "stage"])}
+          />
+        )}
+      </div>
 
       <div className="border border-secondary-accent rounded-lg p-3 flex flex-col gap-2 w-full">
         <PropertyRow label={translateText(["contact"])} required>
--- a/frontend/src/community/crm/components/organisms/DealSidePanel/DealPropertiesSidebar.tsx
+++ b/frontend/src/community/crm/components/organisms/DealSidePanel/DealPropertiesSidebar.tsx
@@ -26,7 +26,6 @@ import {
 import { validateDealAmount } from "~community/crm/utils/dealValidations";
 
 interface DealPropertiesSidebarProps {
-  isOpen?: boolean;
   onStageChange: (stageId: number) => void;
   onAmountChange: (amount: string) => void;
   onPriorityChange: (priority: CrmPriorityEnum) => void;
@@ -35,7 +34,6 @@ interface DealPropertiesSidebarProps {
 }
 
 const DealPropertiesSidebar: FC<DealPropertiesSidebarProps> = ({
-  isOpen,
   onStageChange,
   onAmountChange,
   onPriorityChange,
@@ -112,9 +110,9 @@ const DealPropertiesSidebar: FC<DealPropertiesSidebarProps> = ({
 
   return (
     <div className="w-1/3 flex flex-col gap-4 shrink-0">
-      <div className="w-[219px] max-w-full">
+      <div className="w-full max-w-[13.688rem]">
         {isStagesLoading ? (
-          <SkeletonShape className="h-9 w-full" />
+          <SkeletonShape className="h-12 w-full" />
         ) : (
           <Dropdown
             options={stageOptions}
--- a/frontend/src/community/crm/components/organisms/DealSidePanel/DealSidePanel.tsx
+++ b/frontend/src/community/crm/components/organisms/DealSidePanel/DealSidePanel.tsx
@@ -210,7 +210,6 @@ const DealSidePanel: FC = () => {
                 </div>
               </div>
               <DealPropertiesSidebar
-                isOpen={isOpen}
                 onStageChange={(stageId) => updateDeal({ stageId })}
                 onAmountChange={(amount) => updateDeal({ amount })}
                 onPriorityChange={(priority) => updateDeal({ priority })}
--- a/frontend/src/community/crm/components/organisms/DealSidePanel/DealSidePanelSkeleton.tsx
+++ b/frontend/src/community/crm/components/organisms/DealSidePanel/DealSidePanelSkeleton.tsx
@@ -44,7 +44,7 @@ const DealSidePanelSkeleton: FC = () => (
       </div>
 
       <div className="w-1/3 flex flex-col gap-4 shrink-0">
-        <div className="border border-secondary-accent rounded-lg px-3 py-2 flex items-center justify-between">
+        <div className="w-full max-w-[13.688rem] border border-secondary-accent rounded-lg px-3 py-2 flex items-center justify-between">
           <div className="flex items-center gap-2">
             <SkeletonShape circle className="h-2 w-2 shrink-0" />
             <SkeletonShape className="h-2.5 w-20" />


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR is a purely presentational change: both stage dropdowns (add-deal panel and deal-detail sidebar) are now wrapped in a fixed 219px, `max-w-full` container with the `Dropdown` set to `width="100%"` instead of `55%`. I read both full files plus the parent panels (`AddDealSidePanel`, `DealSidePanel`), `useGetMappedDealStages`, `MultipleSkeletons`, `PriorityDropdown`, and grepped for other stage-dropdown call sites (only these two exist) and for existing `w-[Npx]`/`menuWidth` usage. No functional, type-safety, React Query, i18n, or security concerns — behaviour, query keys, handlers and error states are untouched. Remaining comments are about the duplicated magic width, loading-placeholder consistency, and dropdown-menu width alignment.

Found **9** new issue(s): 🟡 6 suggestion(s), 🟣 3 nit(s)

frontend/src/community/crm/components/organisms/AddDealSidePanel/DealNameStageSection.tsx
```diff
@@ -86,22 +86,24 @@ const DealNameStageSection: FC<DealNameStageSectionProps> = ({
...
-            variant={stageDropdownVariant}
-            className="rounded-lg"
-            width="55%"
-            placeholder={translateText(["placeholders", "stage"])}
-            required
-            errorMessage={stageErrorMessage}
-            ariaLabel={translateText(["ariaLabels", "stage"])}
-          />
-        )}
+        <div className="w-[219px] max-w-full">
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The magic width `219px` is hardcoded in two places (here and `DealPropertiesSidebar.tsx:115`) with no name or comment explaining where it comes from. It is not a Tailwind/theme token (grep shows `w-[219px]` exists nowhere else, and `styles/global.css` has no matching width/spacing token), so if the design spec changes, the two dropdowns will silently drift apart — which is exactly the inconsistency this PR is fixing. Both blocks are also near-identical markup: `useGetMappedDealStages()` → `stageOptions` memo → loading skeleton or `Dropdown` with the same `className`, `placeholder`, `ariaLabel` and now the same wrapper width. Suggest extracting a single `DealStageDropdown` molecule under `~community/crm/components/molecules/` that owns the hook, the `stageOptions` mapping, the skeleton and the width, and taking `value`/`onChange`/`variant`/`errorMessage`/`required` as props. At minimum, define the width once (e.g. a `STAGE_DROPDOWN_WIDTH` entry in `~community/crm/constants/dealConstants.ts` or a `--width-stage-dropdown` var in `global.css`) and reference it from both files.

frontend/src/community/crm/components/organisms/DealSidePanel/DealPropertiesSidebar.tsx
```diff
@@ -112,20 +112,22 @@ const DealPropertiesSidebar: FC<DealPropertiesSidebarProps> = ({
...
-          value={selectedStageId}
-          onChange={handleStageChange}
-          variant="primary"
-          className="rounded-lg"
-          width="55%"
-          placeholder={translateText(["placeholders", "stage"])}
-          ariaLabel={translateText(["ariaLabels", "stage"])}
-        />
-      )}
+      <div className="w-[219px] max-w-full">
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`w-[219px] max-w-full` swaps a fluid width (`55%` of the sidebar column) for a fixed pixel width that only ever shrinks. Inside this `w-1/3 shrink-0` column the effective width becomes `min(219px, column width)`, so on wider viewports the control stops scaling with the panel while every sibling in this column is `w-full` — verify against the design spec that a fixed cap (not a percentage) is intended at large breakpoints. Also note `w-full max-w-[219px]` expresses the same computed result more idiomatically and is easier to reason about than combining a fixed `w-` with `max-w-full`.

frontend/src/community/crm/components/organisms/AddDealSidePanel/DealNameStageSection.tsx
```diff
@@ -86,22 +86,24 @@ const DealNameStageSection: FC<DealNameStageSectionProps> = ({
...
+          {isStagesLoading ? (
+            <MultipleSkeletons numOfSkeletons={1} height={38} />
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Now that both stage dropdowns are pinned to the same width, their loading placeholders are the only thing left inconsistent: this one renders `MultipleSkeletons` with `height={38}` while `DealPropertiesSidebar.tsx:117` renders `SkeletonShape className="h-9 w-full"` (36px). Since the PR goal is visual alignment of these two controls, consider aligning the placeholder height too, and confirming it matches the rendered `Dropdown` height so there is no layout shift when `isStagesLoading` flips. (`MultipleSkeletons` defaults `width` to `100%`, so the new wrapper correctly constrains it to the dropdown width — that part is fine.)

frontend/src/community/crm/components/organisms/DealSidePanel/DealPropertiesSidebar.tsx
```diff
@@ -112,20 +112,22 @@ const DealPropertiesSidebar: FC<DealPropertiesSidebarProps> = ({
...
+        {isStagesLoading ? (
+          <SkeletonShape className="h-9 w-full" />
+        ) : (
+          <Dropdown
+            options={stageOptions}
+            value={selectedStageId}
+            onChange={handleStageChange}
+            variant="primary"
+            className="rounded-lg"
+            width="100%"
```
ThinuwanW, line 123: 🤖 **Claude** · 🟣 **Nit**

Neither stage `Dropdown` sets `menuWidth`, so the popup width falls back to the skapp-ui default — unlike `PriorityDropdown` in this same sidebar, which pairs `width="100%"` with `menuWidth="match"`. Because the trigger width changed in this PR, please visually verify the open menu still aligns with the trigger and that long stage names in the `StageLabel` options are not newly truncated at 219px; if the menu no longer lines up, add `menuWidth="match"` (or `"content"`) to both stage dropdowns for consistency with the sibling dropdown.

frontend/src/community/crm/components/organisms/AddDealSidePanel/DealNameStageSection.tsx
```diff
@@ -86,22 +86,24 @@ const DealNameStageSection: FC<DealNameStageSectionProps> = ({
         />
       </div>
       <div className="w-1/3 pt-6.5">
```
ThinuwanW, line 88: 🤖 **Claude** · 🟡 **Suggestion**

Flex minimum-size regression: the column that now hosts the fixed-width wrapper is a flex item with the default `min-width: auto`, and it has no `min-w-0` (unlike the sibling properties column in AddDealSidePanel.tsx:226, which is `w-1/3 min-w-0`). Previously the dropdown used a percentage width, so the item's min-content contribution was just the trigger text and the row shrank gracefully. A child with a fixed `219px` width raises that contribution to 219px, so the column can no longer shrink below it — and `max-w-full` does not save you here, because a percentage max-width does not constrain a flex item's min-content contribution during intrinsic sizing. At panel widths where the 1/3 column would otherwise be narrower than 219px, the deal-name input gets squeezed or the row overflows the side panel horizontally. Fix: either add `min-w-0` to this `w-1/3 pt-6.5` wrapper, or express the intent as `w-full max-w-[<constant>]` instead of `w-[<constant>] max-w-full` so the box genuinely shrinks.

frontend/src/community/crm/components/organisms/DealSidePanel/DealPropertiesSidebar.tsx
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Missing paired change: `DealSidePanelSkeleton.tsx` mirrors this sidebar layout (its right column uses the same `w-1/3 flex flex-col gap-4 shrink-0` classes), and its stage placeholder — the bordered box at DealSidePanelSkeleton.tsx:47 — has no width constraint, so it still stretches to the full width of the 1/3 column. That skeleton is what renders while `selectedDeal` is undefined in DealSidePanel.tsx:183. With the real dropdown now pinned to 219px, the user sees the stage control visibly snap narrower the moment the deal loads. Wrap the skeleton's stage placeholder in the same constrained container (ideally via the shared constant/component suggested elsewhere) so the two states line up.

ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Dead code: the `isOpen` prop is declared in `DealPropertiesSidebarProps` and destructured in the component signature, but it is never read anywhere in the body. DealSidePanel.tsx:213 dutifully passes it. Since this PR is already restructuring the top of this component's render tree, drop the prop from the interface, the destructuring, and the call site — or wire it up if it was meant to gate something (e.g. resetting the contact search term when the panel closes).

frontend/src/community/crm/components/organisms/DealSidePanel/DealPropertiesSidebar.tsx
```diff
@@ -112,20 +112,22 @@ const DealPropertiesSidebar: FC<DealPropertiesSidebarProps> = ({
...
-          onChange={handleStageChange}
-          variant="primary"
-          className="rounded-lg"
-          width="55%"
-          placeholder={translateText(["placeholders", "stage"])}
-          ariaLabel={translateText(["ariaLabels", "stage"])}
-        />
-      )}
+      <div className="w-[219px] max-w-full">
+        {isStagesLoading ? (
```
ThinuwanW, line 114: 🤖 **Claude** · 🟡 **Suggestion**

No error state for stage loading. `useGetMappedDealStages` returns `isError`, and DealNameStageSection uses it to surface a `validations.stageLoadError` message and switch the dropdown to the `primary-error` variant. Here only `isLoading` is destructured, so when the stages query fails the skeleton disappears and the user is left with an interactive-looking dropdown that has zero options and no explanation — and because the current stage value has no matching option, the trigger falls back to the placeholder, making it look as though the deal has no stage. Destructure `isError` and render the same error message/variant treatment used in the Add Deal panel.

frontend/src/community/crm/components/organisms/AddDealSidePanel/DealNameStageSection.tsx
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Code duplication: after this change the two stage dropdowns are structurally identical — same `useGetMappedDealStages` call, same `useMemo` that maps stages to `{id, value, label: <StageLabel/>}` with `String()` coercion, same width wrapper, same `rounded-lg` + `width="100%"` + placeholder/ariaLabel props. The codebase already has the right precedent for this: `PriorityDropdown` is a molecule that encapsulates exactly this pattern for priority. Extract a `StageDropdown` molecule under `~community/crm/components/molecules/` that owns the hook call, the options mapping, the loading placeholder, the error state, and the width, and takes `value`, `onChange`, plus optional `required`/`errorMessage`/`variant`. That removes the duplicated width literal, the duplicated options `useMemo`, and the divergent loading/error handling in one move, and gives future width tweaks a single place to live. (Note the two call sites use different translator namespaces — `addDealSidePanel` vs `sidePanel` — so pass the translated placeholder/aria strings in as props, or standardise the keys.)

frontend/src/community/crm/components/organisms/DealSidePanel/DealPropertiesSidebar.tsx
YasinduRC: No change needed. `menuWidth` already defaults to `"match"`, so setting it explicitly would be redundant. Also, the trigger is slightly wider now (~219px vs ~202px), so stage names have a bit more room.

frontend/src/community/crm/components/organisms/AddDealSidePanel/DealNameStageSection.tsx
YasinduRC: fixed. Switched to w-full max-w-[13.688rem] so the box actually shrinks instead of flooring the column at 219px.

frontend/src/community/crm/components/organisms/AddDealSidePanel/DealNameStageSection.tsx
YasinduRC: Keeping it inline per our convention, but converted to rem 13.688rem. Not pulling it into a constant for now.

YasinduRC: this'll get folded into a StageDropdown molecule later, no need to duplicate that work here.

frontend/src/community/crm/components/organisms/AddDealSidePanel/DealNameStageSection.tsx
YasinduRC: Fixed, and it was actually worse than 2px off  the real Dropdown trigger is 48px, both skeletons were under that. Set both to 48/h-12 now.

frontend/src/community/crm/components/organisms/DealSidePanel/DealPropertiesSidebar.tsx
YasinduRC: Fixed, same as the other one  w-full max-w-[13.688rem].

YasinduRC: fixed , added the same width cap to the skeleton so it doesn't snap narrower when the deal loads.

YasinduRC: It was dead, removed it

frontend/src/community/crm/components/organisms/DealSidePanel/DealPropertiesSidebar.tsx
YasinduRC: it's missing codebase-wide, not just here, and fixing it needs a new translation key too. Leaving it out of this PR.

shakila-kamalasena (approved):

akilarootcode (approved):

PR #1739, opened 2026-08-11
Fix/2516: Add Full Profile -"Are you sure?" modal does not appear when the user is on any Tab other than Personal

umeshaliyanage:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR extends `useStepper` so that add-flow navigation also syncs the shared `currentStep`/`nextStep` people-store state via a new `goToStep` helper, with tests updated accordingly. The refactor is behaviourally clean for `activeStep` (identical clamping logic preserved) and the tests cover the new assertions, but `goToStep` now writes to state that the *edit* flow owns, and nothing resets it when the add flow unmounts — which can leave the profile edit/account pages opening on the wrong tab. Secondary concerns are the unmapped entitlements step (index 4), a `Record<number, T>` lookup that TypeScript types as non-optional, and the step-index→section mapping being a third hardcoded copy of the step order.

Found **14** new issue(s): 🔴 2 important, 🟡 7 suggestion(s), 🟣 5 nit(s)

frontend/src/community/people/hooks/useStepper.tsx
```diff
@@ -32,15 +41,25 @@ const useStepper = () => {
     return steps;
   }, [isLeaveModuleEnabled, translateText]);
 
+  const goToStep = (step: number) => {
+    setActiveStep(step);
+
+    const section = stepSections[step];
+    if (section) {
+      setCurrentStep(section);
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`goToStep` now writes `currentStep`/`nextStep` — state that belongs to the *edit* flow — from the *add* flow, and nothing resets it when the add flow unmounts. `DirectoryAddSectionWrapper`/`PeopleDirectoryAdd` have no cleanup effect (unlike the legacy `AddNewResourceFlow`, which only calls `resetEmployeeData()`, not `resetPeopleSlice()`). So if a user reaches step 2/3 of Add People and then navigates away, `currentStep` is left at `EditPeopleFormTypes.employment`/`permission`. `PeopleFormSections.getEditFlowSection()` and `PeopleAccountSection` both switch on `currentStep`, and `DirectoryEditSectionWrapper`'s effect only does `setCurrentStep(nextStep)` — which is also stale, since `goToStep` sets both to the same value. Entry points that reset (`PeopleTable.handleRowClick`, `ProfileMenu`) are safe, but ones that don't are not: `pages/enterprise/sign/contacts.tsx` pushes `ROUTES.PEOPLE.EDIT(userId)` / `ROUTES.PEOPLE.ACCOUNT` with no reset, and `pages/enterprise/sign/{inbox,sent}/envelope/[id].tsx` push `ROUTES.PEOPLE.EDIT(fromUser)`. Result: the profile opens on the Employment or System Permissions tab instead of Personal. Fix: add a cleanup effect in the add-flow wrapper that resets `currentStep`/`nextStep` to `EditPeopleFormTypes.personal` on unmount (or call `resetPeopleSlice()`), or reset them defensively on mount of the edit/account wrappers.

frontend/src/community/people/hooks/useStepper.tsx
```diff
@@ -4,9 +4,18 @@ import useSessionData from "~community/common/hooks/useSessionData";
 import { useTranslator } from "~community/common/hooks/useTranslator";
 
 import { usePeopleStore } from "../store/store";
+import { EditPeopleFormTypes } from "../types/PeopleEditTypes";
+
+const stepSections: Record<number, EditPeopleFormTypes> = {
+  0: EditPeopleFormTypes.personal,
+  1: EditPeopleFormTypes.emergency,
+  2: EditPeopleFormTypes.employment,
+  3: EditPeopleFormTypes.permission
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`stepSections` has no entry for step index 4 (Entitlements), so while the Entitlements form is on screen `currentStep` silently remains `EditPeopleFormTypes.permission`. That matters because `DirectoryAddSectionWrapper` renders `RouteChangeAreYouSureModal`, which uses `useFormChangeDetector` — a hook that switches on `currentStep`. On the entitlements step the unsaved-changes/route-guard therefore keeps diffing `systemPermissions` instead of reflecting the visible step, and entitlement edits are never guarded at all. The new test encodes this as intended behaviour, so if it is deliberate please add a short comment explaining why index 4 is excluded; otherwise add an `entitlements` member to `EditPeopleFormTypes` (and a matching branch in `useFormChangeDetector`) so the mapping is total.

frontend/src/community/people/hooks/useStepper.tsx
```diff
@@ -4,9 +4,18 @@ import useSessionData from "~community/common/hooks/useSessionData";
 import { useTranslator } from "~community/common/hooks/useTranslator";
 
 import { usePeopleStore } from "../store/store";
+import { EditPeopleFormTypes } from "../types/PeopleEditTypes";
+
+const stepSections: Record<number, EditPeopleFormTypes> = {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`Record<number, EditPeopleFormTypes>` is a type lie for a partial map: `noUncheckedIndexedAccess` is not enabled in `frontend/tsconfig.json`, so `stepSections[step]` is typed as a non-optional `EditPeopleFormTypes` even though it is `undefined` for step 4. TypeScript will therefore see the `if (section)` guard as always-true, and a future `switch (section)` over this value would not be narrowed correctly. Declare it as `Partial<Record<number, EditPeopleFormTypes>>` (or `Record<number, EditPeopleFormTypes | undefined>`) so the runtime `undefined` case is visible in the type.

frontend/src/community/people/hooks/useStepper.tsx
```diff
@@ -4,9 +4,18 @@ import useSessionData from "~community/common/hooks/useSessionData";
 import { useTranslator } from "~community/common/hooks/useTranslator";
 
 import { usePeopleStore } from "../store/store";
+import { EditPeopleFormTypes } from "../types/PeopleEditTypes";
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

The step index → section mapping is now a third hardcoded copy of the same ordering: the `steps` label array in this hook (lines 29-35), `PeopleFormSections.getAddFlowSection()`'s `switch (activeStep)` (cases 0-4), and now `stepSections`. They can drift independently — in particular, the `isLeaveModuleEnabled` filter only happens to be safe because Entitlements is last; if a step were ever inserted or the filter applied to a non-last step, `stepSections` would map indices to the wrong sections with no compile-time error. Consider defining one ordered array of `{ section, labelKey }` in a shared constants/types file and deriving the labels, the index→section map, and the add-flow section switch from it.

frontend/src/community/people/hooks/useStepper.test.tsx
```diff
@@ -84,19 +92,34 @@ describe("useStepper", () => {
...
+  it("should not change the section on the entitlements step", () => {
+    (usePeopleStore as jest.Mock).mockReturnValue(mockStore(3));
+
+    const { result } = renderHook(() => useStepper());
+
+    act(() => {
+      result.current.handleNext();
     });
 
+    expect(mockSetActiveStep).toHaveBeenCalledWith(4);
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Two coverage gaps in the new tests. (1) There is no test for `handleBack` from the entitlements step — `mockStore(4)` + `handleBack()` should assert `setActiveStep(3)` and `setCurrentStep/setNextStep(EditPeopleFormTypes.permission)`; this is the path that restores the section after the unmapped step, and it is currently unverified. (2) There is no test with `isLeaveModuleEnabled: false`, where `steps.length` is 4 and `handleNext` from step 3 must clamp to 3 and keep `currentStep` at `permission` — the boundary the `stepSections` map depends on. Also, while you are here, `should not go beyond the last step` asserts only `not.toHaveBeenCalledWith(5)`; `expect(mockSetActiveStep).toHaveBeenCalledWith(4)` would actually pin the clamping behaviour.

frontend/src/community/people/hooks/useStepper.tsx
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

Pre-existing, but now behaviourally relevant: `AddNewResourceFlow.tsx:82` calls `useStepper(steps)` even though the hook takes no parameters — the argument is silently ignored (and is an arity mismatch under `strict`). That legacy `/people/directory/add-new-resource` page renders purely off `activeStep`, so before this PR it never touched `currentStep`/`nextStep`; after this change its Next/Back buttons mutate that shared state too, and it only calls `resetEmployeeData()` on unmount. Worth either dropping the stray argument and confirming the legacy page is still wanted, or deleting it now that `/people/directory/add` (`PeopleDirectoryAdd`) is the live add flow.

frontend/src/community/people/hooks/useStepper.tsx
```diff
@@ -32,15 +41,25 @@ const useStepper = () => {
...
+      setCurrentStep(section);
+      setNextStep(section);
+    }
+  };
+
   const handleNext = () => {
     const nextStep =
       activeStep < steps.length - 1 ? activeStep + 1 : activeStep;
-    setActiveStep(nextStep);
+    goToStep(nextStep);
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Last-step detection is now split between this hook and `AddSectionButtonWrapper`, and the two disagree when the leave module is off. `handleNext` clamps at `steps.length - 1`, and `steps` drops the entitlements entry when `isLeaveModuleEnabled` is falsy (`useSessionData` derives it from `user?.roles?.includes(EmployeeTypes.LEAVE_EMPLOYEE)`, so it is `false`/`undefined` for any admin without the leave-employee role). In that configuration the maximum index is 3. But `AddSectionButtonWrapper` renders the Save button only when `activeStep === 4` — and that Save button is the only thing wired to the `useAddEmployee` mutation. So on `/people/directory/add` with leave disabled, the System Permissions step (index 3) shows a 'Next' button whose handler calls `handleNext()`, which clamps back to 3: the user is stuck in a loop and the employee is never POSTed. The legacy `AddNewResourceFlow` avoids this because its `SystemPermissionForm` receives its own `onSave` prop. Fix by making this hook the single source of truth for the boundary — return something like `isLastStep: activeStep === steps.length - 1` (or `lastStepIndex`) — and change `AddSectionButtonWrapper` to use it instead of the literal `4`. Please also add an add-flow test with `isLeaveModuleEnabled: false` covering the final step, since no current test exercises navigation in that configuration.

frontend/src/community/people/hooks/useStepper.tsx
```diff
@@ -32,15 +41,25 @@ const useStepper = () => {
     return steps;
   }, [isLeaveModuleEnabled, translateText]);
 
+  const goToStep = (step: number) => {
+    setActiveStep(step);
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`goToStep` runs unconditionally even when the target step equals the current one. `handleNext` on the last step and `handleBack` on the first both compute `step === activeStep` and still fire `setActiveStep`, `setCurrentStep` and `setNextStep`. Each of those is a `set(() => ({...}))` in `peopleSlice`, which produces a fresh state object and notifies every subscriber — and every `useStepper` consumer subscribes to the entire store via `usePeopleStore((state) => state)`, so a no-op click re-renders `PeopleDirectoryAdd`, `DirectoryAddSectionWrapper`, `AddSectionButtonWrapper` and the active form section. Worse, the no-op now also writes `nextStep`, which is the edit flow's 'requested tab' signal read by `DirectoryEditSectionWrapper`'s sync effect and `DirectorySteppers`. Add an early return at the top of `goToStep`: `if (step === activeStep) return;`.

frontend/src/community/people/hooks/useStepper.tsx
```diff
@@ -32,15 +41,25 @@ const useStepper = () => {
...
+    const section = stepSections[step];
+    if (section) {
+      setCurrentStep(section);
+      setNextStep(section);
+    }
+  };
+
   const handleNext = () => {
     const nextStep =
       activeStep < steps.length - 1 ? activeStep + 1 : activeStep;
```
ThinuwanW, line 45: 🤖 **Claude** · 🟡 **Suggestion**

Naming collision introduced by this change: the local `const nextStep` in `handleNext` (and `prevStep` in `handleBack`) is a numeric step *index*, while the store value this hook now writes via `setNextStep` is an `EditPeopleFormTypes` form *section*. Two different `nextStep` meanings now live in the same scope, and `setNextStep(nextStep)` is a plausible typo that only fails to compile because the types happen to stay disjoint. Rename the locals to `nextIndex` / `prevIndex` (or `targetStep`) to keep index-space and section-space clearly separated.

frontend/src/community/people/hooks/useStepper.tsx
```diff
@@ -32,15 +41,25 @@ const useStepper = () => {
     return steps;
   }, [isLeaveModuleEnabled, translateText]);
 
+  const goToStep = (step: number) => {
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`goToStep`, `handleNext` and `handleBack` are re-created on every render of every consumer, and `AddNewResourceFlow` passes `handleNext`/`handleBack` straight down as `onNext`/`onBack` props to the step forms. The guidelines call out missing `useCallback` on callbacks passed to child components; `goToStep` is new code and now closes over three store setters, so this is a good moment to wrap all three in `useCallback` with `[activeStep, steps.length, setActiveStep, setCurrentStep, setNextStep]` (the setters are stable Zustand references).

frontend/src/community/people/hooks/useStepper.test.tsx
```diff
@@ -84,19 +92,34 @@ describe("useStepper", () => {
     expect(mockSetActiveStep).not.toHaveBeenCalledWith(5);
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Both boundary tests assert only negatives and cannot fail. 'should not go beyond the last step' asserts `expect(mockSetActiveStep).not.toHaveBeenCalledWith(5)` and 'should not go below the first step' asserts `not.toHaveBeenCalledWith(-1)` — both pass even if `handleNext`/`handleBack` did nothing at all, or if `goToStep` threw away the call entirely. Since this PR changes exactly what happens at the clamp (three store writes instead of one), tighten these: assert `expect(mockSetActiveStep).toHaveBeenCalledWith(4)` / `toHaveBeenCalledWith(0)` and pin the intended `setCurrentStep`/`setNextStep` behaviour on a clamped no-op (currently `handleBack` at step 0 writes `EditPeopleFormTypes.personal`, which nothing asserts).

frontend/src/community/people/hooks/useStepper.test.tsx
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The new `stepSections` map has four entries but only one of them is ever asserted. Every navigation test lands on index 1, so all three assertions check `EditPeopleFormTypes.emergency`; indices 0, 2 (`employment`) and 3 (`permission`) are never verified. Swapping `employment` and `permission` in the map — the exact kind of mistake a hardcoded index→enum table invites — would ship with a green suite. Add cases driving `handleNext` from `mockStore(1)` and `mockStore(2)` asserting `employment` and `permission`, or table-drive the existing test over `[[0,1,emergency],[1,2,employment],[2,3,permission]]`.

frontend/src/community/people/hooks/useStepper.test.tsx
```diff
@@ -84,19 +92,34 @@ describe("useStepper", () => {
...
   });
 
-  it("should handle previous step correctly", () => {
-    (usePeopleStore as jest.Mock).mockReturnValue({
-      activeStep: 2,
-      setActiveStep: mockSetActiveStep
+  it("should not change the section on the entitlements step", () => {
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Test name does not match what it does: 'should not change the section on the entitlements step' sets `activeStep` to 3, which is System Permissions — it actually tests the transition *into* entitlements. Rename to something like 'should not change the section when moving to the entitlements step'. It is also worth a short comment explaining *why* index 4 has no mapping, because as written the `not.toHaveBeenCalled()` assertions lock in the missing entry as intended behaviour with no rationale for the next reader.

frontend/src/community/people/hooks/useStepper.test.tsx
```diff
@@ -18,14 +19,20 @@ jest.mock("~community/common/hooks/useTranslator", () => ({
 
 describe("useStepper", () => {
   const mockSetActiveStep = jest.fn();
+  const mockSetCurrentStep = jest.fn();
+  const mockSetNextStep = jest.fn();
   const mockTranslateText = jest.fn((keys) => keys[0]);
 
+  const mockStore = (activeStep: number) => ({
```
ThinuwanW, line 41: 🤖 **Claude** · 🟣 **Nit**

`mockStore` has no return type annotation and returns only four of the many `peopleSlice` fields. If `useStepper` later destructures another store value (the hook already uses the whole-state selector `usePeopleStore((state) => state)`), the tests will silently hand back `undefined` instead of failing loudly. Annotate the return as `Partial<PeopleSliceTypes>` — or better, `Pick<PeopleSliceTypes, "activeStep" | "setActiveStep" | "setCurrentStep" | "setNextStep">` — so the mock is checked against the real slice contract.

frontend/src/community/people/hooks/useStepper.tsx
umeshaliyanage: added a use effect to reset the currentStep/nextStep in wrapper level

frontend/src/community/people/hooks/useStepper.test.tsx
umeshaliyanage: out of scope for the bug

frontend/src/community/people/hooks/useStepper.tsx
umeshaliyanage: exposed isLastStep as mentioned

frontend/src/community/people/hooks/useStepper.tsx
umeshaliyanage: not needed

frontend/src/community/people/hooks/useStepper.test.tsx
umeshaliyanage: already has "should not go beyond the system permissions step when leave module is disabled" tests

frontend/src/community/people/hooks/useStepper.tsx
umeshaliyanage: Declared it as Partial<Record<number, EditPeopleFormTypes>>

frontend/src/community/people/hooks/useStepper.tsx
umeshaliyanage: Confirmed the mechanism, RouteChangeAreYouSureModal → useFormChangeDetector switches on currentStep, so on the Entitlements step it keeps diffing systemPermissions.

frontend/src/community/people/hooks/useStepper.test.tsx
umeshaliyanage: Renamed to should not change the section when moving to the entitlements step. To note that entitlements has no EditPeopleFormTypes member, so stepSections has no entry for index 4 and the section stays on system permissions. Whether that gap should exist is the open question in the stepSections thread.

frontend/src/community/people/hooks/useStepper.tsx
umeshaliyanage: Added an early return at the top of goToStep: if (step === activeStep) return;.

frontend/src/community/people/hooks/useStepper.tsx
umeshaliyanage: renamed to to index

frontend/src/community/people/hooks/useStepper.tsx
umeshaliyanage: removed taking parameters from the hook

umeshaliyanage: not needed

frontend/src/community/people/utils/objectComparisonUtils.ts
```diff
@@ -0,0 +1,43 @@
+const getDefinedKeys = (value: Record<string, unknown>): string[] =>
```
hasalarootcode: These utils should be removed, and we need to handle it through the useFormChangeDetector

hasalarootcode (changes requested):

frontend/src/community/people/hooks/useStepper.test.tsx
umeshaliyanage: not needed

hasalarootcode (dismissed):

frontend/src/community/people/hooks/useStepper.test.tsx
umeshaliyanage: outdated

frontend/src/community/people/components/molecules/AddSectionButtonWrapper/AddSectionButtonWrapper.tsx
```diff
@@ -73,9 +74,13 @@ const AddSectionButtonWrapper = ({
...
 
-  const { employee, profilePic, thumbnail, setCommonDetails } = usePeopleStore(
-    (state) => state
-  );
+  const {
+    employee,
+    profilePic,
+    thumbnail,
+    setCommonDetails,
+    resetPeopleSlice
```
akilarootcode, line 82: use query selector

frontend/src/community/people/components/organisms/AccountSectionWrapper/AccountSectionWrapper.tsx
```diff
@@ -31,9 +32,15 @@ const AccountSectionWrapper = ({ employeeId }: Props) => {
...
     setEmployee,
-    setProjectTeamNames
+    setProjectTeamNames,
+    setNextStep
   } = usePeopleStore((state) => state);
 
+  useEffect(() => {
+    setCurrentStep(EditPeopleFormTypes.personal);
+    setNextStep(EditPeopleFormTypes.personal);
+  }, [setCurrentStep, setNextStep]);
```
akilarootcode: remove setters from setCurrentStep, setNextStep

frontend/src/community/people/components/organisms/DirectoryAddSectionWrapper/DirectoryAddSectionWrapper.tsx
```diff
@@ -12,6 +15,15 @@ const DirectoryAddSectionWrapper = () => {
...
   const { activeStep, steps } = useStepper();
 
+  const { setCurrentStep, setNextStep } = usePeopleStore((state) => state);
+
+  useEffect(() => {
+    return () => {
+      setCurrentStep(EditPeopleFormTypes.personal);
+      setNextStep(EditPeopleFormTypes.personal);
+    };
+  }, [setCurrentStep, setNextStep]);
```
akilarootcode: remove setCurrentStep, setNextStep

frontend/src/community/people/components/organisms/AddNewResourceFlow/AddNewResourceFlow.tsx
```diff
@@ -79,7 +79,7 @@ const AddNewResourceFlow = () => {
     return steps;
   }, [isLeaveModuleEnabled, translateText]);
 
-  const { activeStep, handleNext, handleBack } = useStepper(steps);
```
akilarootcode, line 82: revert this

frontend/src/community/people/components/organisms/DirectoryEditSectionWrapper/DirectoryEditSectionWrapper.tsx
```diff
@@ -40,9 +41,15 @@ const DirectoryEditSectionWrapper = ({ employeeId }: Props) => {
...
     setIsCancelModalConfirmButtonClicked,
-    setIsCancelChangesModalOpen
+    setIsCancelChangesModalOpen,
+    setNextStep
   } = usePeopleStore((state) => state);
 
+  useEffect(() => {
+    setCurrentStep(EditPeopleFormTypes.personal);
+    setNextStep(EditPeopleFormTypes.personal);
+  }, [setCurrentStep, setNextStep]);
```
akilarootcode: remove [setCurrentStep, setNextStep

frontend/src/community/people/hooks/useFormChangeDetector.test.tsx
```diff
@@ -86,4 +86,19 @@ describe("useFormChangeDetector", () => {
     expect(result.current.hasChanged).toBe(false);
     expect(result.current.apiPayload).toEqual({});
   });
+
+  it("should detect personal details entered during the add flow", () => {
```
akilarootcode, line 90: run the test and check.

akilarootcode (changes requested):

frontend/src/community/people/components/molecules/AddSectionButtonWrapper/AddSectionButtonWrapper.tsx
umeshaliyanage: used query selector

frontend/src/community/people/components/organisms/AddNewResourceFlow/AddNewResourceFlow.tsx
umeshaliyanage: reverted

frontend/src/community/people/components/organisms/DirectoryAddSectionWrapper/DirectoryAddSectionWrapper.tsx
umeshaliyanage: removed

frontend/src/community/people/components/organisms/DirectoryEditSectionWrapper/DirectoryEditSectionWrapper.tsx
umeshaliyanage: removed

frontend/src/community/people/hooks/useFormChangeDetector.test.tsx
umeshaliyanage: ran passed the test

frontend/src/community/people/components/molecules/AddSectionButtonWrapper/AddSectionButtonWrapper.tsx
```diff
@@ -73,9 +74,11 @@ const AddSectionButtonWrapper = ({
 
   const { mutate: createCustomSkills } = useCreateCustomSkills();
 
-  const { employee, profilePic, thumbnail, setCommonDetails } = usePeopleStore(
-    (state) => state
-  );
+  const employee = usePeopleStore((state) => state.employee);
```
akilarootcode: const {
  employee,
  profilePic,
  thumbnail,
  setCommonDetails,
  resetPeopleSlice
} = usePeopleStore((state) => ({
  employee: state.employee,
  profilePic: state.profilePic,
  thumbnail: state.thumbnail,
  setCommonDetails: state.setCommonDetails,
  resetPeopleSlice: state.resetPeopleSlice
}));

akilarootcode (changes requested):

frontend/src/community/people/components/organisms/AccountSectionWrapper/AccountSectionWrapper.tsx
umeshaliyanage: removed

frontend/src/community/people/components/molecules/AddSectionButtonWrapper/AddSectionButtonWrapper.tsx
umeshaliyanage: combined the query selector as in the comment

akilarootcode (approved):

PR #1740, opened 2026-08-11
feat(crm): add GET /v1/crm/company/{id} company detail endpoint

anusham-rc:
## What

Adds `GET /v1/crm/company/{id}` — returns the base details of a single company (`CrmCompanyResponseDto`).

- **Resolution:** `findByIdAndIsDeletedFalse(...).orElseThrow(CRM_ERROR_COMPANY_NOT_FOUND)` → missing **or soft-deleted** company → clean **400**, not 500.
- **Auth:** `@PreAuthorize("hasAnyRole('ROLE_CRM_SALES_REPRESENTATIVE')")`, matching the v1 company controller.

Placed on the **v1** controller: a brand-new endpoint with no existing consumer, so there is nothing to keep compatible on v1 and no later v1→v2 migration; keeping it on v1 avoids fragmenting the company surface across versions.

## Tests

`CrmCompanyControllerIntegrationTest` (v1) — happy-path, not-found → 400, soft-deleted → 400, without-CRM-role → 403. Company v1 + v2 suites green (24 + 5).

## Notes


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds a straightforward `GET /v2/crm/company/{id}` endpoint. The layering (controller → interface → impl → Dao), `@Transactional(readOnly = true)`, entry/exit logging, `ModuleException` with an existing message constant, soft-delete filtering via `findByIdAndIsDeletedFalse`, MapStruct mapping to the existing `CrmCompanyResponseDto` (the same DTO the V2 list endpoint nests, so the response shape is consistent), and the `ROLE_CRM_SALES_REPRESENTATIVE` check (satisfied by CRM_ADMIN/CRM_SALES_MANAGER via the hierarchy in `AuthorityServiceImpl`) are all correct and consistent with the module. I verified no enterprise subclass of `CrmCompanyServiceImplV2` exists, so the added constructor arg from `@RequiredArgsConstructor` breaks nothing, and no other mapping collides with `/v2/crm/company/**`. No correctness defects found; the remaining items are robustness and test-consistency polish.

Found **8** new issue(s): 🟡 4 suggestion(s), 🟣 4 nit(s)

backend/src/main/java/com/skapp/community/crmplanner/controller/v2/CrmCompanyControllerV2.java
```diff
@@ -30,4 +31,12 @@ public ResponseEntity<ResponseEntityDto> getCompanyMetrics(CrmCompanyMetricReque
 		return new ResponseEntity<>(responseDto, HttpStatus.OK);
 	}
 
+	@Operation(summary = "Get company by ID", description = "Returns the base details of a single company.")
+	@GetMapping("/{id}")
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

A non-numeric path segment now yields a 500 instead of a client error. `GET /v2/crm/company/abc` (or `GET /v2/crm/company/metrics` — a plausible client mistake, since the V1 equivalent is `/v1/crm/company/metrics`) fails Long conversion with `MethodArgumentTypeMismatchException`. `GlobalExceptionHandler` does not extend `ResponseEntityExceptionHandler` and has no handler for it, so it falls through to the catch-all `@ExceptionHandler(Exception.class)` → HTTP 500 plus an `ERROR`-level log line for what is really bad client input. Fix by constraining the mapping to digits, e.g. `@GetMapping("/{id:\\d+}")` (unmatched paths then return 404), or by adding a `MethodArgumentTypeMismatchException` handler to `GlobalExceptionHandler` that returns 400. This pattern pre-dates the PR (V1 `deleteCompany`/`editCompany` have the same exposure), so a global handler would be the higher-value fix.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmCompanyControllerV2IntegrationTest.java
```diff
@@ -231,4 +232,57 @@ void getCompanyMetrics_WithoutCrmRole_ReturnsForbidden() throws Exception {
...
 	}
 
+	// --- getCompanyById ---
+
+	@Test
+	@DisplayName("Get company by ID - Returns base company details")
+	void getCompanyById_HappyPath_ReturnsCompany() throws Exception {
+		CrmCompany company = savedCompany("DetailCoV2Unique");
+
+		performRequest(get(METRICS_PATH + "/" + company.getId()).accept(MediaType.APPLICATION_JSON), authToken)
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Test-path naming and duplication. `METRICS_PATH` (line 56) is now also used as the base for the by-ID URL, which reads misleadingly, and the same request-building block `performRequest(get(METRICS_PATH + "/" + company.getId()).accept(MediaType.APPLICATION_JSON), token)` is repeated across three of the four new tests with manual string concatenation. The sibling `CrmContactControllerV2IntegrationTest` already establishes the convention: `BASE_PATH = "/v2/crm/contact"`, `BY_ID_PATH = BASE_PATH + "/{id}"`, `METRICS_PATH = BASE_PATH`, plus a `performGetByIdRequest(id)` helper using the URI template form `get(BY_ID_PATH, id)`. Mirror that here: rename the constant to `BASE_PATH`, add `BY_ID_PATH`, and extract a `performGetByIdRequest(Long id, String token)` helper.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmCompanyControllerV2IntegrationTest.java
```diff
@@ -231,4 +232,57 @@ void getCompanyMetrics_WithoutCrmRole_ReturnsForbidden() throws Exception {
...
+			.andExpect(jsonPath("['results'][0]['address']").value("123 Metrics St"))
+			.andExpect(jsonPath("['results'][0]['contactNumber']").value("94771234567"));
+	}
+
+	@Test
+	@DisplayName("Get company by ID that does not exist - Returns Bad Request")
+	void getCompanyById_NotFound_ReturnsBadRequest() throws Exception {
+		performRequest(get(METRICS_PATH + "/999999").accept(MediaType.APPLICATION_JSON), authToken).andDo(print())
+			.andExpect(status().isBadRequest())
+			.andExpect(jsonPath(STATUS_PATH).value(STATUS_UNSUCCESSFUL));
```

The not-found and soft-deleted tests only assert HTTP 400 and `status == unsuccessful`, so any unrelated `ModuleException` (or a future validation error on the path) would keep them green — including the soft-delete test, whose whole point is proving `findByIdAndIsDeletedFalse` filtered the row. Add a message assertion as `CrmCompanyControllerIntegrationTest` does, e.g. inject `MessageUtil` and assert `jsonPath("['results'][0]['message']").value(messageUtil.getMessage(CrmMessageConstant.CRM_ERROR_COMPANY_NOT_FOUND))`.

backend/src/main/java/com/skapp/community/crmplanner/controller/v2/CrmCompanyControllerV2.java
```diff
@@ -30,4 +31,12 @@ public ResponseEntity<ResponseEntityDto> getCompanyMetrics(CrmCompanyMetricReque
...
+	@PreAuthorize("hasAnyRole('ROLE_CRM_SALES_REPRESENTATIVE')")
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`hasAnyRole(...)` is used with a single role. Functionally identical to `hasRole(...)`, but every other V2 CRM controller (`CrmContactControllerV2`, `CrmDealControllerV2`, `CrmTaskControllerV2`) and `/v1/crm/company/lookup` use `hasRole('ROLE_CRM_SALES_REPRESENTATIVE')` for single-role checks. Prefer `hasRole` for consistency (the neighbouring `getCompanyMetrics` has the same inconsistency, so fixing both here would be reasonable).

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmCompanyServiceImplV2.java
```diff
@@ -40,4 +47,18 @@ public ResponseEntityDto getCompanyMetrics(CrmCompanyMetricRequestDto requestDto
 		return new ResponseEntityDto(false, pageDto);
 	}
 
+	@Override
+	@Transactional(readOnly = true)
+	public ResponseEntityDto getCompanyById(Long id) {
+		log.info("getCompanyById: execution started");
+
+		CrmCompany company = crmCompanyDao.findByIdAndIsDeletedFalse(id)
+			.orElseThrow(() -> new ModuleException(CrmMessageConstant.CRM_ERROR_COMPANY_NOT_FOUND));
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

A missing/soft-deleted company returns HTTP 400 rather than 404, because `ModuleException` maps to `BAD_REQUEST` in `GlobalExceptionHandler`. This matches the CRM module's existing convention (`CrmCompanyServiceImpl.deleteCompany`/`editCompany`, `CrmContactServiceImplV2.getContactById`), so it is internally consistent — but the codebase already has `EntityNotFoundException` mapped to 404 and uses it widely in the people/leave/time/enterprise modules for exactly this case. For a brand-new GET-by-ID endpoint, 404 is the more accurate signal for clients that need to distinguish "no such record" from "bad request". Your call whether to diverge here or align the CRM module as a whole later.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmCompanyServiceImplV2.java
```diff
@@ -21,6 +26,8 @@ public class CrmCompanyServiceImplV2 implements CrmCompanyServiceV2 {
 
 	private final CrmCompanyDao crmCompanyDao;
 
+	private final CrmMapper crmMapper;
```

This is the only V2 CRM service impl that injects the V1 `CrmMapper`; `CrmContactServiceImplV2`, `CrmDealServiceImplV2`, and `CrmTaskServiceImplV2` all inject `CrmMapperV2`, which is already declared as `@Mapper(componentModel = "spring", uses = CrmMapper.class)` precisely so V2 has a single mapping entry point that can delegate to V1. The consequence is that the new V2 detail contract is now welded to `CrmCompanyResponseDto`, which is simultaneously the response body for three V1 endpoints (`POST /v1/crm/company`, `PATCH /v1/crm/company/{id}`, `GET /v1/crm/company/search-by-domain`) and is additionally built positionally via `cb.construct(CrmCompanyResponseDto.class, id, name, industry, website, address, contactNumber)` in `CrmCompanyRepositoryImpl.getCompanyMetricsV2`. So adding a field for any one consumer silently changes the V2 detail payload and, because of `@AllArgsConstructor`, also breaks that positional `cb.construct` call at query-build time. Fix: add the company mapping method to `CrmMapperV2` (it will delegate to `CrmMapper` through `uses`) and inject `CrmMapperV2` here instead, so V2 can diverge from V1 later without touching V1 callers.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmCompanyControllerV2IntegrationTest.java
```diff
@@ -231,4 +232,57 @@ void getCompanyMetrics_WithoutCrmRole_ReturnsForbidden() throws Exception {
...
+			.andDo(print())
+			.andExpect(status().isOk())
+			.andExpect(jsonPath(STATUS_PATH).value(STATUS_SUCCESSFUL))
+			.andExpect(jsonPath("['results'][0]['id']").value(company.getId()))
+			.andExpect(jsonPath("['results'][0]['name']").value("DetailCoV2Unique"))
+			.andExpect(
+					jsonPath("['results'][0]['industry']").value(CrmIndustry.TECHNOLOGY_INFORMATION_AND_MEDIA.name()))
+			.andExpect(jsonPath("['results'][0]['website']").value("https://metrics-v2.com"))
+			.andExpect(jsonPath("['results'][0]['address']").value("123 Metrics St"))
+			.andExpect(jsonPath("['results'][0]['contactNumber']").value("94771234567"));
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The new by-ID tests assert only that the six fields that are present have the right values; nothing asserts what must be absent, so the tests do not lock the response shape. The sibling `CrmContactControllerV2IntegrationTest.getContactById_HappyPath_ReturnsBaseContactWithoutMetrics` deliberately adds `jsonPath(RESULTS_0_PATH + "['metrics']").doesNotExist()` and `..."['totalRevenue']").doesNotExist()` for exactly this reason — it proves the detail endpoint returns base details only and not the metrics list-item shape. Without an equivalent assertion, if someone later switches `getCompanyById` to return `CrmCompanyListItemDtoV2` (which nests the company under a `company` key), every assertion here would break loudly — but the reverse drift (extra fields such as metrics being added to the detail payload) would pass silently. Add `.andExpect(jsonPath("['results'][0]['metrics']").doesNotExist())` to `getCompanyById_HappyPath_ReturnsCompany` to pin the documented "base details of a single company" contract.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmCompanyControllerV2IntegrationTest.java
```diff
@@ -231,4 +232,57 @@ void getCompanyMetrics_WithoutCrmRole_ReturnsForbidden() throws Exception {
...
+	@Test
+	@DisplayName("Get company by ID - Returns base company details")
+	void getCompanyById_HappyPath_ReturnsCompany() throws Exception {
+		CrmCompany company = savedCompany("DetailCoV2Unique");
+
+		performRequest(get(METRICS_PATH + "/" + company.getId()).accept(MediaType.APPLICATION_JSON), authToken)
+			.andDo(print())
+			.andExpect(status().isOk())
+			.andExpect(jsonPath(STATUS_PATH).value(STATUS_SUCCESSFUL))
+			.andExpect(jsonPath("['results'][0]['id']").value(company.getId()))
```

The new assertions hand-write the JSON path prefix `"['results'][0]"` six times instead of using the shared `TestConstants.RESULTS_0_PATH` constant, which already exists alongside the `STATUS_PATH`/`STATUS_UNSUCCESSFUL` constants this PR imports. The sibling `CrmContactControllerV2IntegrationTest` uses `RESULTS_0_PATH + "['id']"` throughout its by-ID tests, and so do the V1 CRM integration tests. Add `import static com.skapp.support.TestConstants.RESULTS_0_PATH;` and switch the six new expressions to `RESULTS_0_PATH + "['name']"` form. (The pre-existing metrics tests in this file use raw strings for the deeper `items[0]` paths, so this only concerns the newly added by-ID assertions.)

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmCompanyControllerV2IntegrationTest.java
```diff
@@ -231,4 +232,57 @@ void getCompanyMetrics_WithoutCrmRole_ReturnsForbidden() throws Exception {
 			.andExpect(status().isForbidden());
 	}
 
+	// --- getCompanyById ---
```
akilarootcode: remove controller tests and if needed implement integrations test.

akilarootcode (changes requested):

backend/src/main/java/com/skapp/community/crmplanner/controller/v2/CrmCompanyControllerV2.java
anusham-rc: Valid, but this is pre-existing and module-wide — the v1 `deleteCompany`/`editCompany` routes have the same `/{id}` exposure. Better addressed globally in `GlobalExceptionHandler` (a `MethodArgumentTypeMismatchException` handler) as a separate change than by constraining just this one route, so leaving it out of scope here.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmCompanyControllerV2IntegrationTest.java
anusham-rc: The endpoint moved to the v1 controller; the v1 test already uses `BASE_PATH` (not `METRICS_PATH`) and follows the existing suite's style, so this is resolved.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmCompanyControllerV2IntegrationTest.java
anusham-rc: added `CRM_ERROR_COMPANY_NOT_FOUND` message assertions to both the not-found and soft-deleted tests, so the soft-delete test now actually proves `findByIdAndIsDeletedFalse` filtered the row.

backend/src/main/java/com/skapp/community/crmplanner/controller/v2/CrmCompanyControllerV2.java
anusham-rc: On the v1 company controller the single-role checks use `hasAnyRole(...)` throughout (delete/edit/metrics/lookup etc.), so this matches its neighbours. Keeping `hasAnyRole` for local consistency.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmCompanyServiceImplV2.java
anusham-rc: Intentional — matches the CRM module convention (`ModuleException` -> 400). Aligning the module to 404 would be a separate, module-wide change.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmCompanyServiceImplV2.java
anusham-rc: The endpoint now lives on the v1 controller and uses the v1 `CrmMapper` like the rest of `CrmCompanyServiceImpl`, so the V2-mapper coupling no longer applies.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmCompanyControllerV2IntegrationTest.java
anusham-rc: `CrmCompanyResponseDto` has no metrics variant (unlike the contact DTO), so there is no metrics field that could leak into this payload — nothing to pin here.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmCompanyControllerV2IntegrationTest.java
anusham-rc: The v1 tests already use `RESULTS_0_PATH` throughout, so this is resolved.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmCompanyControllerV2IntegrationTest.java
anusham-rc: These are full `@SpringBootTest` integration tests running against a real database, not controller-only tests. Each by-id test inserts a real company row via `crmCompanyDao.save(...)`, then calls `GET /v1/crm/company/{id}`, which goes through controller → service → repository → DB and back — so the assertions verify data that actually round-tripped through the database. The soft-deleted case inserts a company, marks it deleted, and confirms the DB query (`findByIdAndIsDeletedFalse`) filters it out (400).

This also matches how GET endpoints are tested across the codebase (e.g. `getJobFamilyById`, `getContactById`) — assert the response from a DB-backed request. Direct DAO read-backs are used only to verify writes, so there's nothing controller-level to remove here.

akilarootcode (dismissed):

akilarootcode (approved):

PR #1741, opened 2026-08-11
feat: add CRM slices for companies, contacts, deals, tasks, lookup, a…

YasinduRC:
…nd UI state management

## CRM-715 — Define the new CRM store

**Definitions only.** Adds `frontend/src/community/crm/v2/store/` — one composed `CrmStore` type and six slice interfaces. State shapes and action signatures, no implementations. Nothing imports it and no existing file is touched.

Depends on #1728 (CRM entity types).

### The architecture

- **Normalized** — each entity stored once in `Record<id, Entity>`; views hold ordered id arrays, relations are FK ids plus id arrays on the parent.
- **Deals and the board share one slice and one `CrmDealEntity`** — `dealIds` is the table order, `board` is `Record<stageId, CrmBoardColumn>` holding ids. Removes the current two-shape/two-slice split.
- **UI state separated** into `crmUiSlice`, mirroring the existing store field-for-field so call sites don't change.
- **Lookups resolve by id** — `owners`, `stages`, `taskTypes` as maps instead of arrays scanned with `.find()`.

### Conventions

- **`set*` writes data + order, `upsert*` writes data only** — so an edit can never reorder or duplicate a row.
- **`currentPage` decides replace vs extend** (omitted/`0` replaces, `>0` extends). The caller passes `res.currentPage` through, so it can't pick wrong.
- **Relation setters take no page** — panel lists have no search, so first-load vs load-more is inferable.
- **Lookup order is derived** from `orderIndex`, not stored.

### Contracts the implementation must honour

1. `upsert*` merges field by field, never replaces — responses are partial views.
2. Selectors validate membership against the FK, so moves and deletes self-heal.
3. One funnel handles insertion on FK change, in the same `set()`.
4. Counts come from `/metrics/{id}`, never derived — a loaded array is one page.
5. `append` dedupes; normalize never defaults an id array to `[]`.

### Not in scope

Action implementations, the existing `crm/store/` (untouched, still in use), and where filter state lives — all follow in the migration PR.

### Verification

`tsc --noEmit` clean for `v2/store`, prettier clean, no consumers so no runtime impact.

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds a new `v2/store` layer for the CRM module consisting of six Zustand slice *interfaces* and a combined `CrmStore` intersection type. The blocking problem is that every slice imports entity types from `~community/crm/types/CrmTypes`, a module that does not exist anywhere in the repo (verified by `find` and grep — none of `CrmCompanyEntity`, `CrmContactEntity`, `CrmDealEntity`, `CrmTaskEntity`, `CrmBoardColumn`, `CrmOwnerEntity`, `CrmStageEntity`, `CrmTaskTypeEntity` are declared), so all seven files fail `tsc --noEmit`. Beyond that, there is no `create()`/`devtools()` store instance or slice implementation, nothing in the codebase imports these files, and several design details (Record-based state losing stage ordering, unusable `currentPage` params, cross-slice relation setters, index-based `moveDeal` vs. the neighbour-based reorder API) will cause problems once implementations land.

Found **20** new issue(s): 🔴 7 important, 🟡 8 suggestion(s), 🟣 5 nit(s)

frontend/src/community/crm/v2/store/slices/crmCompaniesSlice.ts
```diff
@@ -0,0 +1,20 @@
+import {
+  CrmCompanyEntity,
+  CrmContactEntity,
+  CrmDealEntity,
+  CrmTaskEntity
+} from "~community/crm/types/CrmTypes";
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Unresolvable import: `~community/crm/types/CrmTypes` does not exist. `~community/*` maps to `./src/community/*` (tsconfig.json paths), and there is no `src/community/crm/types/CrmTypes.ts` — the types dir only contains TaskTabTypes, ModalTypes, BoardTypes, SidePanelTypes, SliceTypes, StoreTypes and CommonTypes, and there is no barrel `index.ts`. A repo-wide search shows none of the eight referenced entity types (`CrmCompanyEntity`, `CrmContactEntity`, `CrmDealEntity`, `CrmTaskEntity`, `CrmBoardColumn`, `CrmOwnerEntity`, `CrmStageEntity`, `CrmTaskTypeEntity`) is declared anywhere in the frontend. This is the same broken import in all five slice files (crmCompaniesSlice, crmContactsSlice, crmDealsSlice, crmLookupSlice, crmTasksSlice), so `tsc --noEmit` fails with TS2307 for each and the build is red. Fix: land `src/community/crm/types/CrmTypes.ts` (or the equivalent v2 types file) with these entity definitions in the same PR, or point the imports at the existing `~community/crm/types/CommonTypes` / `BoardTypes` types (`CrmCompany`, `CrmContact`, `CrmDealResponseType`, `CrmTaskDetailType`, `CrmBoardStageDealsType`).

frontend/src/community/crm/v2/store/store.ts
```diff
@@ -0,0 +1,13 @@
+import { CrmCompaniesSlice } from "./slices/crmCompaniesSlice";
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

📁 **File-level observation**

This file only declares a type — there is no `create<CrmStore>()(devtools(...))` store instance, and none of the six slices ship a creator function/implementation. Combined with the fact that no file outside `v2/` imports anything from `crm/v2` (verified by grep), the whole directory is unreachable dead code: the interfaces are never structurally checked against an implementation, so nothing guarantees they are even satisfiable, and reviewers cannot verify behaviour. Please either include the slice creators plus the store instance in this PR, or state in the PR description that this is a deliberate types-only scaffold step. When adding the store, follow the established pattern (see `src/enterprise/crm/store/crmStore.ts`): `create<CrmStore>()(devtools((set, get) => ({...}), { name: "crmStoreV2", enabled: process.env.NODE_ENV !== "production" }))` — note that `src/community/crm/store/store.ts` declares a `["zustand/persist", CrmStore]` mutator it never applies, so do not copy that file verbatim.

frontend/src/community/crm/v2/store/slices/crmCompaniesSlice.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export interface CrmCompaniesSlice {
+  companies: Record<number, CrmCompanyEntity>;
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`Record<number, CrmCompanyEntity>` is unsound for a normalized-by-id map. `noUncheckedIndexedAccess` is not enabled in frontend/tsconfig.json, so `companies[selectedCompanyId]` is typed as `CrmCompanyEntity` while being `undefined` at runtime for any id not yet loaded — e.g. deep-linking to a company side panel before the list query resolves gives `companies[id].name` and a TypeError, with no compile error to warn you. The v1 store modelled this correctly via `getCompanyById: (id: number) => CrmCompany | undefined` (types/StoreTypes.ts:20), and v2 drops those selectors, so this is a safety regression. Same applies to `contacts`, `deals`, `tasks`, `board`, `owners`, `stages` and `taskTypes` across the other slices. Fix: declare these as `Record<number, CrmCompanyEntity | undefined>` (or `Partial<Record<...>>`), or keep explicit `getXById(id): X | undefined` selectors on each slice so callers are forced to handle the miss.

frontend/src/community/crm/v2/store/slices/crmDealsSlice.ts
```diff
@@ -0,0 +1,27 @@
+import {
+  CrmBoardColumn,
+  CrmDealEntity,
+  CrmTaskEntity
+} from "~community/crm/types/CrmTypes";
+
+export interface CrmDealsSlice {
+  deals: Record<number, CrmDealEntity>;
+  dealIds: number[];
+  board: Record<number, CrmBoardColumn>;
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`board: Record<number, CrmBoardColumn>` loses stage ordering. v1 deliberately stores board state as an ordered array (`boardStageDeals: CrmBoardStageDealsType[]`, types/StoreTypes.ts:70) which preserves the stage order returned by the API. With a numeric-keyed object, iteration order is ascending-by-integer-key, so kanban lanes will silently render sorted by `stageId` rather than by the configured stage display order — the bug only shows up once stages are reordered or a new stage is inserted. The same gap exists in `crmLookupSlice.stages: Record<number, CrmStageEntity>` (line 9), which has no companion ids array even though every other slice here pairs its Record with an ordering array (`companyIds`, `contactIds`, `dealIds`, `taskIds`). Fix: add `boardStageIds: number[]` (and `stageIds: number[]` in the lookup slice) maintained alongside the Records, and drive rendering off those arrays.

frontend/src/community/crm/v2/store/slices/crmCompaniesSlice.ts
```diff
@@ -0,0 +1,20 @@
...
+  companyIds: number[];
+
+  setCompanies: (companies: CrmCompanyEntity[], currentPage?: number) => void;
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`setCompanies(companies, currentPage?)` takes a `currentPage` that no field in the slice can hold — there is no `currentPage`, `totalPages`, `pageSize`, `totalCount` or `hasNextPage` state here (same for `setContacts`, `setDeals`, `setTasks`). The param therefore only exists to implicitly switch behaviour ("page 1 replaces `companyIds`, page > 1 appends"), which is invisible at the call site and easy to get wrong. Contrast `setBoardColumn` in crmDealsSlice, which passes pagination explicitly — though it also drops `totalPages` and `pageSize` that v1's `CrmBoardStageDealsType` (types/BoardTypes.ts:43-51) carries. Fix: either add explicit pagination state per collection and pass the full metadata, or split the ambiguity into two intent-revealing actions (`setCompanies(companies)` to replace, `appendCompanies(companies)` to page in) and drop the optional param.

frontend/src/community/crm/v2/store/slices/crmCompaniesSlice.ts
```diff
@@ -0,0 +1,20 @@
...
+  upsertCompanies: (companies: CrmCompanyEntity[]) => void;
+  upsertCompany: (company: CrmCompanyEntity) => void;
+  removeCompany: (companyId: number) => void;
+
+  setCompanyContacts: (companyId: number, contacts: CrmContactEntity[]) => void;
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Cross-slice relation setters are declared on the wrong slice: `setCompanyContacts`/`setCompanyDeals`/`setCompanyTasks` live on `CrmCompaniesSlice` but accept `CrmContactEntity[]`/`CrmDealEntity[]`/`CrmTaskEntity[]`, i.e. entities owned by the contacts/deals/tasks slices (same for `setContactDeals`/`setContactTasks` in crmContactsSlice and `setDealTasks` in crmDealsSlice). With this repo's slice pattern the creator is typed `SetType<TSlice>` (see store/slices/crmBoardSlice.ts:15), so an implementation typed against `SetType<CrmCompaniesSlice>` cannot write into `contacts`/`deals`/`tasks` — you will end up either widening to `SetType<CrmStore>` or reaching for an `as` cast to silence it. Decide now: type these creators against the combined `CrmStore` (as store/slices/crmSidePanelSlice.ts:5 already does), or keep only the relation ids on the parent entity (`contactIds: number[]` on `CrmCompanyEntity`) and upsert the entities through their owning slice's `upsertContacts`/`upsertDeals`/`upsertTasks`.

frontend/src/community/crm/v2/store/store.ts
```diff
@@ -0,0 +1,13 @@
...
+import { CrmContactsSlice } from "./slices/crmContactsSlice";
+import { CrmDealsSlice } from "./slices/crmDealsSlice";
+import { CrmLookupSlice } from "./slices/crmLookupSlice";
+import { CrmTasksSlice } from "./slices/crmTasksSlice";
+import { CrmUiSlice } from "./slices/crmUiSlice";
+
+export type CrmStore = CrmCompaniesSlice &
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Naming collision: `CrmStore` is already exported from `~community/crm/types/StoreTypes` (v1) and `~enterprise/crm/types/CrmStoreTypes`. This makes a third, structurally unrelated `CrmStore`, so during the migration any file touching both layers gets ambiguous imports and needs aliasing. Rename to something version-explicit (`CrmStoreV2`) and name the eventual hook `useCrmStoreV2`, so `useCrmStore` keeps pointing unambiguously at the existing store.

frontend/src/community/crm/v2/store/slices/crmUiSlice.ts
```diff
@@ -0,0 +1,38 @@
+import { CrmModalTypes } from "~community/crm/types/ModalTypes";
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

`CrmUiSlice` re-declares essentially all of the v1 UI state (`isCompanyModalOpen`, `companyModalType`, `isContactModalOpen`, `contactModalType`, `isTaskModalOpen`, `taskModalType`, `selected*Id`, `isCrmSidePanelOpen`, `crmSidePanelType`, `preselectedStageId` — compare types/StoreTypes.ts:54-71). Once a v2 store instance exists, two Zustand stores will hold the same modal/side-panel/selection state while ~20 existing consumers (pages/community/crm/*.tsx and the `*SidePanel`/`*ModalController` organisms) still read v1, so the two can diverge — e.g. a panel opened through v2 while v1 still reports `isCrmSidePanelOpen: false`. Please migrate the consumers and delete the v1 slices in the same change (or an immediate follow-up), and consider adding a `resetCrmUi()` action so stale `selected*Id`/modal state is cleared on route change — v1 lacks one and `closeCrmSidePanel` leaves `selected*Id` set.

frontend/src/community/crm/v2/store/slices/crmDealsSlice.ts
```diff
@@ -0,0 +1,27 @@
...
+  removeDeal: (dealId: number) => void;
+
+  setBoardColumn: (
+    stageId: number,
+    deals: CrmDealEntity[],
+    totalCount: number,
+    currentPage: number,
+    hasNextPage: boolean
+  ) => void;
+  moveDeal: (dealId: number, toStageId: number, toIndex: number) => void;
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`moveDeal(dealId, toStageId, toIndex)` is index-based, but the backend reorder contract is neighbour-based: `CrmBoardReorderWithinStagePayload` / `CrmBoardMoveBetweenStagesPayload` require `previousDealId`/`nextDealId` (+`newStageId`), computed by `computeMoveNeighbors`/`kanbanUtil.ts:92-96,274-279` and consumed in `useKanbanDrag.ts:124-160`. An index-only signature forces the caller to recompute neighbours from the column array after the store has already mutated it, which is exactly the ordering hazard v1 avoids by computing neighbours before the optimistic update. Consider having `moveDeal` return the resulting `{ previousDealId, nextDealId }` (like v1's `CrmBoardReorderResultType`) so the mutation payload can be derived from a single source of truth.

frontend/src/community/crm/v2/store/slices/crmLookupSlice.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export interface CrmLookupSlice {
+  owners: Record<number, CrmOwnerEntity>;
+  stages: Record<number, CrmStageEntity>;
+  taskTypes: Record<number, CrmTaskTypeEntity>;
+
+  upsertOwners: (owners: CrmOwnerEntity[]) => void;
+
+  upsertStages: (stages: CrmStageEntity[]) => void;
+  removeStage: (stageId: number) => void;
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Asymmetric lookup API: `removeStage` exists but there is no `removeOwner`/`removeTaskType`, so those maps can only grow — a deactivated owner or deleted task type stays in the store for the session. Either add the matching removers or, since owners/stages/taskTypes are pure server-derived lookup data, consider leaving them in the React Query cache with an appropriate `staleTime` instead of mirroring them into Zustand (per the guideline on not duplicating server cache in the store); note the deal-stage lookup is already owned by `src/community/configurations/stores/slices/dealStageSlice.ts`, which this slice would partially duplicate. Also worth confirming the placement convention: `v2/` is the only such directory under `src/`, and v1 keeps slice/store types in `types/SliceTypes.ts` and `types/StoreTypes.ts` rather than inline in `store/slices/*` — a short note in the PR description on the intended v2 layout and migration path would help future readers.

frontend/src/community/crm/v2/store/slices/crmDealsSlice.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export interface CrmDealsSlice {
+  deals: Record<number, CrmDealEntity>;
+  dealIds: number[];
+  board: Record<number, CrmBoardColumn>;
+
+  setDeals: (deals: CrmDealEntity[], currentPage?: number) => void;
+  upsertDeals: (deals: CrmDealEntity[]) => void;
+  upsertDeal: (deal: CrmDealEntity) => void;
+  removeDeal: (dealId: number) => void;
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The deals slice keeps `board` as separate state from `deals`/`dealIds`, but the contract has no action to keep the board in sync on create/update/delete. v1 deliberately exposes `addDealToStage`, `updateDealInStage` and `removeDealFromStage` alongside `removeDeal`/`updateDeal`, and there are live consumers that must call both: `DeleteDealModal.tsx:46-47` calls `removeDeal(id)` AND `removeDealFromStage(id)`; `AddDealSidePanel.tsx:82` calls `addDealToStage(...)`; `DealSidePanel.tsx:96` calls `updateDealInStage(...)`. With only `upsertDeal`/`removeDeal` declared here, either the board silently keeps ghost cards after a delete (and stale cards after a stage change), or the cascade is an undocumented implicit behaviour of `removeDeal`/`upsertDeal` that consumers cannot see from the type. The same gap exists across slices: `CrmLookupSlice.removeStage(stageId)` has no way to drop the orphaned `board[stageId]` column, and `removeCompany`/`removeContact` leave dangling `companyId`/`contactId` references on cached deals/tasks. Fix: either declare explicit board actions (`addDealToBoard`, `removeDealFromBoard`, `moveDealToStage`) or document on the interface (JSDoc) that `upsertDeal`/`removeDeal`/`removeStage` cascade into `board`, and make sure the implementation actually does.

frontend/src/community/crm/v2/store/slices/crmDealsSlice.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  setBoardColumn: (
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`setBoardColumn` narrows the server contract and is call-site unsafe. The backend response type `CrmBoardStageDealsResponseType` (types/BoardTypes.ts:33-41) returns `stageId, deals, totalCount, currentPage, totalPages, pageSize, hasNextPage`, and v1's `CrmBoardStageDealsType` stores all seven. This signature accepts only five, dropping `totalPages` and `pageSize` — `totalPages` is the field the CRM infinite queries actually use to compute the next page (`CompanyApi.ts:55`, `crmDealApi.ts:52`, `TaskApi.ts:43`), so this is real behavioural drift from v1. Separately, five positional parameters ending in `(…, totalCount: number, currentPage: number, hasNextPage: boolean)` makes transposing the two adjacent numbers a silent, type-checked bug. Fix: take a single object argument typed against the existing board response shape, e.g. `setBoardColumn: (column: CrmBoardColumn) => void` where `CrmBoardColumn` carries the full pagination metadata.

frontend/src/community/crm/v2/store/slices/crmCompaniesSlice.ts
```diff
@@ -0,0 +1,20 @@
+import {
+  CrmCompanyEntity,
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The new `*Entity` types are a second, parallel domain model for entities that already exist in `~community/crm/types/CommonTypes.ts`: `CrmCompanyEntity` vs `CrmCompany` (line 31), `CrmContactEntity` vs `CrmContact` (line 410), `CrmDealEntity` vs `CrmDealResponseType` (line 185), `CrmTaskEntity` vs `CrmTaskDetailType` (line 251), `CrmOwnerEntity` vs `CrmOwner` (line 14), `CrmStageEntity` vs `CrmDealStageType` (line 205), `CrmTaskTypeEntity` vs `CrmTaskCategory` (line 266). This is inconsistent even within the PR — `crmUiSlice.ts` reuses the v1 `CrmModalTypes`/`CrmSidePanelTypes` enums from `~community/crm/types/`, so the module half-reuses and half-forks its own types. Two entity models over one backend contract means every API mapper needs a translation layer and enum/field drift becomes invisible to the compiler. Fix: reuse the existing `CommonTypes`/`BoardTypes` interfaces (aliasing or `Pick`ing them if the store only needs a subset) rather than declaring a new entity set.

frontend/src/community/crm/v2/store/store.ts
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

This contract stores React-Query-owned server state in Zustand and extends it further than v1 did: normalized entity maps for companies/contacts/deals/tasks, lookup maps for owners/stages/taskTypes, plus per-column pagination metadata (`totalCount`, `currentPage`, `hasNextPage`) and `currentPage` on every list setter. Server cache duplicated into Zustand is explicitly called out as an anti-pattern in the review guidelines, and the CRM module already has infinite queries that own exactly this data (`CompanyApi.ts`, `ContactApi.ts`, `crmDealApi.ts`, `TaskApi.ts`) with `getNextPageParam` driving pagination. Keeping a second copy means every mutation now needs both `invalidateQueries` and a store write, and the two will drift. Fix: keep only genuine client state in this store (the `CrmUiSlice` fields, board drag/optimistic-ordering state) and read entities/lookups/pagination from React Query via `select`, or at minimum document why the board specifically needs a mirrored cache for optimistic drag-and-drop.

frontend/src/community/crm/v2/store/store.ts
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Structural and naming deviations from the module conventions. (1) `crm/v2/` is the only `v2` directory anywhere under `frontend/src` — the documented module layout is `module/{api,components,store,utils,types}` with no version segment, and `v2` here reads like an API version (the codebase uses `authFetchV2`/`v1` vs `v2` for endpoints), which it is not. (2) The slice interfaces are colocated in `store/slices/*.ts`, but this module's convention puts store types in `types/SliceTypes.ts` and `types/StoreTypes.ts`. (3) The files are pluralized (`crmCompaniesSlice.ts`, `crmContactsSlice.ts`, `crmDealsSlice.ts`) while the existing ones one directory up are singular (`crmCompanySlice.ts`, `crmContactSlice.ts`, `crmDealSlice.ts`), so the two sets are near-indistinguishable in an editor's file switcher. Fix: pick a naming/location scheme that matches the module (either extend `types/SliceTypes.ts` and `store/slices/` in place, or state the v2 migration convention in the module README so reviewers know it is deliberate).

frontend/src/community/crm/v2/store/slices/crmCompaniesSlice.ts
```diff
@@ -0,0 +1,20 @@
...
+  CrmDealEntity,
+  CrmTaskEntity
+} from "~community/crm/types/CrmTypes";
+
+export interface CrmCompaniesSlice {
+  companies: Record<number, CrmCompanyEntity>;
+  companyIds: number[];
+
+  setCompanies: (companies: CrmCompanyEntity[], currentPage?: number) => void;
+  upsertCompanies: (companies: CrmCompanyEntity[]) => void;
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The `setX` vs `upsertX` split is undocumented and silently inverts v1's behaviour. In v1, `setCompanies` and `setContacts` do NOT replace — they merge (`set({ companies: mergeWithExisting(get().companies, companies) })`, crmCompanySlice.ts:22-23 and crmContactSlice.ts:25-26), which is what keeps accumulated infinite-scroll pages alive. Here the naming implies `setCompanies` replaces and `upsertCompanies` merges, so a straight port of the existing call sites would drop every previously loaded page on each fetch. Fix: add JSDoc on each setter stating replace-vs-merge semantics and what happens to `companyIds` (full set vs current-page ordering), and audit the v1 call sites when they are migrated. Same applies to `setContacts`, `setDeals`, `setTasks`.

frontend/src/community/crm/v2/store/slices/crmContactsSlice.ts
```diff
@@ -0,0 +1,18 @@
...
+  contacts: Record<number, CrmContactEntity>;
+  contactIds: number[];
+
+  setContacts: (contacts: CrmContactEntity[], currentPage?: number) => void;
+  upsertContacts: (contacts: CrmContactEntity[]) => void;
+  upsertContact: (contact: CrmContactEntity) => void;
+  removeContact: (contactId: number) => void;
+
+  setContactDeals: (contactId: number, deals: CrmDealEntity[]) => void;
+  setContactTasks: (contactId: number, tasks: CrmTaskEntity[]) => void;
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Missing paired capability: v1 exposes `updateContactTaskCompletion(contactId, taskId, isCompleted)` (StoreTypes.ts:32-36, crmContactSlice.ts:29-41) and it has a live consumer at `components/molecules/TaskRow/TaskRow.tsx:46` for toggling a task's completion from within the contact side panel. There is no v2 equivalent — `setContactTasks` replaces the whole list and `CrmTasksSlice.upsertTask` updates the task map, but with contact→task relations held separately there is no declared action that flips completion in the contact-scoped view. Fix: add an explicit `setTaskCompletion(taskId, isCompleted)` on `CrmTasksSlice` and make the contact-scoped view derive from the task map, or carry `updateContactTaskCompletion` forward, so `TaskRow` has a migration target.

frontend/src/community/crm/v2/store/store.ts
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Nothing in `src` or `pages` imports anything under `crm/v2` (grep for `crm/v2` returns no hits), and the existing `useCrmStore` in `crm/store/store.ts` remains the only wired store. Shipping a second, unreferenced store contract for the same module to main leaves two competing definitions of `CrmStore` with no stated migration path, and unreferenced type-only files will not be caught by any test or build failure if they drift from the backend. Fix: either land this together with its first consumer (store instance + one migrated hook/component) or add a short README/ADR under `crm/v2/` recording the migration plan and which v1 pieces it replaces, so the next reader knows which contract is authoritative.

frontend/src/community/crm/v2/store/slices/crmLookupSlice.ts
```diff
@@ -0,0 +1,18 @@
+import {
+  CrmOwnerEntity,
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

The `*Entity` suffix is introduced by this PR and appears nowhere else in `frontend/src` — the established suffixes are `*Type`, `*ResponseType`, `*Payload`, or bare domain names (`CrmOwner`, `CrmCompany`). If the parallel type model is kept despite the duplication noted above, align the suffix with the existing convention so the CRM module does not carry two naming schemes side by side.

frontend/src/community/crm/v2/store/slices/crmUiSlice.ts
```diff
@@ -0,0 +1,38 @@
...
+  contactModalType: CrmModalTypes;
+  isTaskModalOpen: boolean;
+  taskModalType: CrmModalTypes;
+
+  selectedCompanyId: number | null;
+  selectedContactId: number | null;
+  selectedDealId: number | null;
+  selectedTaskId: number | null;
+
+  isCrmSidePanelOpen: boolean;
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

The UI slice permits illegal states that a fresh design could rule out at the type level. `isCrmSidePanelOpen: boolean` plus `crmSidePanelType: CrmSidePanelTypes | null` allows `(true, null)` and `(false, SOME_TYPE)`; `openCrmSidePanel`/`closeCrmSidePanel` are the only intended writers, so the pair could be a single `crmSidePanel: CrmSidePanelTypes | null` with openness derived. Likewise `companyModalType: CrmModalTypes` accepts any member of the enum, so `companyModalType = CrmModalTypes.ADD_TASK_MODAL` type-checks; narrowing each to its own subset (e.g. `Extract<CrmModalTypes, ...>` or per-entity enums) would make the wrong modal type a compile error. This mirrors v1, so it is only worth doing if v2 is meant to be an improvement over it.

frontend/src/community/crm/v2/store/store.ts
YasinduRC: Fixed, real store instance now exists (useCrmStoreV2), both slices are implemented and wired in.

frontend/src/community/crm/v2/store/slices/crmCompaniesSlice.ts
YasinduRC: Fixed, the entity-types branch has landed, imports are repointed to the real v2/types/CrmCommonTypes and v2/types/CrmTypes, and the module compiles clean.

frontend/src/community/crm/v2/store/slices/crmLookupSlice.ts
YasinduRC: Intentional , Entity marks what's actually in the store, separate from Response/Request. That's kind of the whole point of this model. It was merged with v2 types

frontend/src/community/crm/v2/store/slices/crmUiSlice.ts
YasinduRC: Fair, but intentional mirroring v1 exactly is what lets existing components move over without change

frontend/src/community/crm/v2/store/store.ts
YasinduRC: v2/ and the colocated types are both intentional.The plural/singular filename thing doesn't apply anymore though we collapsed all five files into one (crmDataSlice.ts).

frontend/src/community/crm/v2/store/slices/crmContactsSlice.ts
YasinduRC: Still covered, just a different call now. We can do it via the setter

frontend/src/community/crm/v2/store/slices/crmCompaniesSlice.ts
YasinduRC: The id-array part (append vs replace) is intentional — the setter doesn't decide that anymore, the caller does.

frontend/src/community/crm/v2/store/store.ts
YasinduRC: valid, nothing in the codebase imports v2/store yet it will be wired with new implementations

frontend/src/community/crm/v2/store/slices/crmCompaniesSlice.ts
YasinduRC: These replace the old types, they don't sit alongside them

frontend/src/community/crm/v2/store/store.ts
YasinduRC: Used CrmStoreV2

frontend/src/community/crm/v2/store/slices/crmLookupSlice.ts
YasinduRC: Removed the remove stage

frontend/src/community/crm/v2/store/slices/crmDealsSlice.ts
YasinduRC: Removed the moveDeal

frontend/src/community/crm/v2/store/slices/crmUiSlice.ts
YasinduRC: Valid re-declared since v1 will replaced by v2

frontend/src/community/crm/v2/store/slices/crmCompaniesSlice.ts
YasinduRC: Removed all the cross slice setters

frontend/src/community/crm/v2/store/slices/crmCompaniesSlice.ts
YasinduRC: Those will not handled in the store itself, those will handled outside the store

frontend/src/community/crm/v2/store/slices/crmDealsSlice.ts
YasinduRC: No need since all the deals in a single place only ids are stored in deals both board and table render from the data of it, those arrays will update with the utils

frontend/src/community/crm/v2/store/store.ts
YasinduRC: Zustand is the source of truth for CRM data, and React Query is only used for specific invalidation needs

frontend/src/community/crm/v2/store/slices/crmDealsSlice.ts
YasinduRC: CrmBoardColumn is intentionally a client-side model, not a mirror of the API response. We only store the fields the board needs, so totalPages and pageSize are intentionally omitted. I agree that setBoardColumn can take a CrmBoardColumn object to make the call site safer.

frontend/src/community/crm/v2/store/slices/crmDealsSlice.ts
YasinduRC: Added stageIds: number[];

frontend/src/community/crm/v2/store/slices/crmCompaniesSlice.ts
YasinduRC: We only access companies from the loaded list and don't rely on direct companies[id] lookups, so a missing-ID selector isn't needed here.

frontend/src/community/crm/v2/types/SliceTypes.ts
```diff
@@ -0,0 +1,61 @@
+import { CrmStore } from "./StoreTypes";
```
thusala: why do we need Pick? do not follow existing store pattern. check the benefits and find the proper purpose for this.

thusala (changes requested):

frontend/src/community/crm/v2/types/SliceTypes.ts
YasinduRC: I checked the actual purpose of Pick, and we don't need it here, so I removed it.

In our case, Pick was only narrowing what set could update; it wasn't validating that the slice actually implemented those fields. It also meant declaring the same fields in multiple places.

What it is now
 - same folder structure, mechanism swapped: - SliceTypes.ts declares CrmDataSliceTypes / CrmUiSliceTypes directly 

- StoreTypes.ts: export type CrmStore = CrmDataSliceTypes & CrmUiSliceTypes 

- slices are typed StateCreator<CrmStore, [["zustand/devtools", never]], [], CrmDataSliceTypes>

[REF](https://chatgpt.com/share/6a7d6f76-69f4-83ee-8232-31a43c1fed7f)

thusala (approved):

akilarootcode (approved):

PR #1742, opened 2026-08-11
feat(crm): add GET /v1/crm/company/{id}/metrics endpoint

anusham-rc:
## What

Adds `GET /v1/crm/company/{id}/metrics` — the aggregated task/deal metrics for a single company (`CrmCompanyMetrics`).

- **Resolution:** a dedicated criteria query filters on `id` + `isDeleted = false`; missing **or soft-deleted** → **400** (`CRM_ERROR_COMPANY_NOT_FOUND`); existing-but-empty → zeroed metrics.
- **Auth:** `@PreAuthorize("hasAnyRole('ROLE_CRM_SALES_REPRESENTATIVE')")`.

Placed on the **v1** controller (brand-new endpoint, no consumer). The new `getCompanyMetricsById` query is self-contained; the list query is untouched.

## Tests

`CrmCompanyControllerIntegrationTest` (v1) — aggregated-metrics (seeded deals + open/overdue tasks), zero-metrics, not-found → 400, soft-deleted → 400, without-CRM-role → 403. Company v1 + v2 suites green (25 + 5).

## Notes


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds a `GET /v2/crm/company/{id}/metrics` endpoint backed by a new Criteria API repository method. The layering, `@PreAuthorize`, `@Transactional(readOnly = true)`, entry/exit logging, `ModuleException` + message constant, and integration test coverage all follow the module's established conventions, and the `cb.construct` argument order correctly matches `CrmCompanyMetrics`. The main concerns are that `getCompanyMetricsById` is a ~50-line verbatim copy of the metrics subquery block that now exists three times in the same file, plus an inherited inconsistency in how "open" vs "closed" deals are derived, and some smaller convention/robustness nits.

Found **15** new issue(s): 🔴 3 important, 🟡 7 suggestion(s), 🟣 5 nit(s)

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmCompanyRepositoryImpl.java
```diff
@@ -198,6 +199,65 @@ public Page<CrmCompanyListItemDtoV2> getCompanyMetricsV2(Pageable pageable, Stri
 		return new PageImpl<>(content, pageable, total);
 	}
 
+	@Override
```
ThinuwanW, line 202: 🤖 **Claude** · 🔴 **Important**

The entire six-subquery metrics block (open tasks, overdue tasks, open value, account value, open count, closed count) is now duplicated a third time in this file — `getCompanyMetrics` (lines 58-105), `getCompanyMetricsV2` (lines 133-182) and now `getCompanyMetricsById` (lines 210-255) are character-for-character identical except for the final `cb.construct(...)` and the `where` clause. That is ~50 lines copied, and it means any future change to the metric definition (e.g. the overdue rule or the closed-stage rule) has to be applied in three places or the endpoints silently disagree. Review guideline checklist item 9 calls this out explicitly.

Suggested fix: extract a private helper that builds and returns the metrics selection, e.g. `private Expression<CrmCompanyMetrics> buildMetricsSelection(CriteriaBuilder cb, AbstractQuery<?> query, From<?, CrmCompany> company, List<Long> closedStageIds)`, which creates the six subqueries via `query.subquery(...)` and returns `cb.construct(CrmCompanyMetrics.class, ...)`. All three methods can then call it. Note one trap when doing this: `CrmCompanyMetrics` declares `openDeals` before `closedDeals`, while `CrmCompanyMetricsResponseDto` declares `closedDeals` before `openDeals`, so the v1 call site must keep passing the two count subqueries in the opposite order — worth adding a short comment at that call site so a later refactor does not silently swap them.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmCompanyRepositoryImpl.java
```diff
@@ -198,6 +199,65 @@ public Page<CrmCompanyListItemDtoV2> getCompanyMetricsV2(Pageable pageable, Stri
...
+					cb.isFalse(subOverdueTask.get(CrmTask_.isCompleted)),
+					cb.isNotNull(subOverdueTask.get(CrmTask_.dueAt)),
+					cb.lessThan(subOverdueTask.get(CrmTask_.dueAt), cb.localDateTime()));
+
+		Subquery<BigDecimal> openValueSubquery = query.subquery(BigDecimal.class);
+		Root<CrmDeal> openDeal = openValueSubquery.from(CrmDeal.class);
+		openValueSubquery
+			.select(cb.coalesce(cb.sum(openDeal.get(CrmDeal_.amount).cast(BigDecimal.class)), BigDecimal.ZERO))
+			.where(cb.equal(openDeal.get(CrmDeal_.company), company), cb.isFalse(openDeal.get(CrmDeal_.isDeleted)),
+					cb.not(openDeal.get(CrmDeal_.stage).get(CrmDealStage_.id).in(closedStageIds)));
```
ThinuwanW, line 230: 🤖 **Claude** · 🟡 **Suggestion**

The new query (copied from the existing ones) derives "open" and "closed" using two different and non-complementary rules: `openValue`/`openDeals` use `NOT (stage.id IN closedStageIds)` where `closedStageIds` comes from `getClosedStageIds()` (which filters `stage.isDeleted = false`), while `accountValue`/`closedDeals` use `stage.stageType = WON` with no `isDeleted` filter on the stage. Two consequences:

1. If a WON stage is soft-deleted, its id is absent from `closedStageIds`, so deals in that stage are counted as OPEN (`openDeals`, `openValue`) *and* as closed (`closedDeals`, `accountValue`) at the same time — the same deal appears in both buckets.
2. If no non-deleted WON/LOST stage exists at all, `closedStageIds` is empty and `NOT (id IN ())` evaluates to true for every row, so every deal — including WON ones — is reported as open.

Since `CrmDeal.stage` is `nullable = false`, you can avoid both by expressing openness directly against the joined stage instead of a pre-fetched id list, e.g. `cb.not(deal.get(CrmDeal_.stage).get(CrmDealStage_.stageType).in(CrmConstants.TERMINAL_STAGES))`. That also removes the extra `getClosedStageIds()` round trip this single-company lookup currently performs. This is pre-existing behaviour, but it is being replicated into new code, so it is worth fixing in the shared helper suggested above rather than propagating it.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmCompanyServiceImplV2.java
```diff
@@ -40,4 +43,16 @@ public ResponseEntityDto getCompanyMetrics(CrmCompanyMetricRequestDto requestDto
...
+	@Override
+	@Transactional(readOnly = true)
+	public ResponseEntityDto getCompanyMetricsById(Long id) {
+		log.info("getCompanyMetricsById: execution started");
+
+		CrmCompanyMetrics metrics = crmCompanyDao.getCompanyMetricsById(id)
+			.orElseThrow(() -> new ModuleException(CrmMessageConstant.CRM_ERROR_COMPANY_NOT_FOUND));
+
+		log.info("getCompanyMetricsById: execution ended");
+		return new ResponseEntityDto(false, metrics);
```

`CrmCompanyMetrics` is returned directly as the top-level response payload of a public API endpoint, but it lives in `crmplanner/type/` and is not named `*ResponseDto`. Per the review guidelines, response payloads belong in `payload/response/` (here `payload/response/v2/`) and follow the `*ResponseDto` naming convention; `type/` is for enums and internal types. Until now `CrmCompanyMetrics` was only ever a nested field of `CrmCompanyListItemDtoV2`, so its placement was invisible to API consumers — this change promotes it to a contract-level DTO.

Also worth considering: the response carries no company identifier, so a client that fires several of these concurrently cannot correlate a response to its request. Recommend introducing e.g. `CrmCompanyMetricsResponseDtoV2` in `payload/response/v2/` that includes the company id alongside the metrics, or at minimum moving/aliasing `CrmCompanyMetrics` into the payload package.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmCompanyControllerV2IntegrationTest.java
```diff
@@ -231,4 +232,91 @@ void getCompanyMetrics_WithoutCrmRole_ReturnsForbidden() throws Exception {
 			.andExpect(status().isForbidden());
 	}
 
+	// --- getCompanyMetricsById ---
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Coverage gaps in the new test block, given the query semantics being introduced:

1. No test with a LOST-stage deal. `openValue`/`openDeals` exclude LOST via `closedStageIds` while `accountValue`/`closedDeals` only count WON, so a LOST deal should appear in neither bucket. That is the least obvious part of the query and is currently untested for this endpoint. Adding a LOST deal to `getCompanyMetricsById_HappyPath_ReturnsMetrics` would also have surfaced the soft-deleted-stage / empty-`closedStageIds` issue noted on the repository.
2. No test that a completed (`isCompleted = true`) or soft-deleted task is excluded from `openTasksCount`/`overdue`, even though the query filters on both.
3. Per review guideline checklist item 10 (symmetric endpoints): this endpoint and `GET /v2/crm/company` compute the same six metrics from duplicated code. A test asserting that the by-id response matches the `metrics` block returned by the list endpoint for the same seeded company would catch the two implementations drifting apart — which is the concrete risk created by the code duplication.

backend/src/main/java/com/skapp/community/crmplanner/controller/v2/CrmCompanyControllerV2.java
```diff
@@ -30,4 +31,13 @@ public ResponseEntity<ResponseEntityDto> getCompanyMetrics(CrmCompanyMetricReque
 		return new ResponseEntity<>(responseDto, HttpStatus.OK);
 	}
 
+	@Operation(summary = "Get company metrics by ID",
+			description = "Returns the aggregated task and deal metrics for a single company.")
+	@GetMapping("/{id}/metrics")
+	@PreAuthorize("hasAnyRole('ROLE_CRM_SALES_REPRESENTATIVE')")
+	public ResponseEntity<ResponseEntityDto> getCompanyMetricsById(@PathVariable Long id) {
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`@PathVariable Long id` with a non-numeric segment (e.g. `GET /v2/crm/company/abc/metrics`) raises `MethodArgumentTypeMismatchException`. `GlobalExceptionHandler` has no handler for it, so it falls through to the catch-all `@ExceptionHandler(Exception.class)` (GlobalExceptionHandler.java:178) and the client gets a 500 with `COMMON_ERROR_MODULE_EXCEPTION` instead of a 400 — review guideline checklist item 1. This is a codebase-wide pattern rather than something this PR invented (every `@PathVariable Long` in the CRM v1 controllers has the same exposure), so the right fix is probably a single `@ExceptionHandler(MethodArgumentTypeMismatchException.class)` in `GlobalExceptionHandler` returning `HttpStatus.BAD_REQUEST`, rather than anything local to this endpoint. Flagging so it is a conscious decision.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmCompanyServiceImplV2.java
```diff
@@ -40,4 +43,16 @@ public ResponseEntityDto getCompanyMetrics(CrmCompanyMetricRequestDto requestDto
 		return new ResponseEntityDto(false, pageDto);
 	}
 
+	@Override
+	@Transactional(readOnly = true)
+	public ResponseEntityDto getCompanyMetricsById(Long id) {
+		log.info("getCompanyMetricsById: execution started");
+
+		CrmCompanyMetrics metrics = crmCompanyDao.getCompanyMetricsById(id)
+			.orElseThrow(() -> new ModuleException(CrmMessageConstant.CRM_ERROR_COMPANY_NOT_FOUND));
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

A GET-by-id for a company that does not exist (or is soft-deleted) returns HTTP 400 via `ModuleException`, whereas `GlobalExceptionHandler` already maps `EntityNotFoundException` to 404 (GlobalExceptionHandler.java:119). 404 is the more accurate status for this read path and would let clients distinguish "bad id format / bad request" from "no such company". That said, `CrmCompanyServiceImpl` (lines 158, 175) and `CrmValidations` (line 44) all use `ModuleException` + `CRM_ERROR_COMPANY_NOT_FOUND` for the same condition, so the current choice is at least internally consistent — leaving this to your judgement rather than asking for a one-off deviation.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmCompanyControllerV2IntegrationTest.java
```diff
@@ -231,4 +232,91 @@ void getCompanyMetrics_WithoutCrmRole_ReturnsForbidden() throws Exception {
...
+		performRequest(get(METRICS_PATH + "/" + company.getId() + "/metrics").accept(MediaType.APPLICATION_JSON),
+				authToken)
+			.andDo(print())
+			.andExpect(status().isOk())
+			.andExpect(jsonPath(STATUS_PATH).value(STATUS_SUCCESSFUL))
+			.andExpect(jsonPath("['results'][0]['openDeals']").value(0))
+			.andExpect(jsonPath("['results'][0]['closedDeals']").value(0))
+			.andExpect(jsonPath("['results'][0]['openTasksCount']").value(0))
+			.andExpect(jsonPath("['results'][0]['overdue']").value(0))
+			.andExpect(jsonPath("['results'][0]['openValue']").value("0"))
```

`openValue`/`accountValue` are asserted here as exact strings (`.value("0")`), but those values are produced by a database-side `CAST(COALESCE(SUM(...), 0) AS <char>)`. Tests run on H2 (`src/test/resources/application.yml`) while production is MySQL, and the two render decimal-to-string differently (`"0"` vs `"0.00"`, and MySQL's `CAST(... AS DECIMAL)` without an explicit precision defaults to scale 0). So this assertion can pass on H2 while the production response format differs. Every other value assertion in this file — including the happy-path test you added just above at lines 262-265 — uses `new BigDecimal(value)` with `isEqualByComparingTo`, which is immune to formatting. Recommend doing the same here for consistency and robustness.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmCompanyServiceImplV2.java
```diff
@@ -40,4 +43,16 @@ public ResponseEntityDto getCompanyMetrics(CrmCompanyMetricRequestDto requestDto
 		return new ResponseEntityDto(false, pageDto);
 	}
 
+	@Override
+	@Transactional(readOnly = true)
+	public ResponseEntityDto getCompanyMetricsById(Long id) {
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Missing sales-representative owner scoping — this endpoint leaks pipeline data a rep is otherwise blocked from seeing.

Everywhere else in the CRM module, a user whose crm_role is exactly CRM_SALES_REPRESENTATIVE is restricted to records they own: `CrmDealServiceImplV2.getDealById` and `CrmTaskServiceImplV2.getTaskById` throw `CRM_ERROR_DEAL_VIEW_DENIED` / `CRM_ERROR_TASK_VIEW_DENIED` via `CrmValidations.isOwnerRestrictedForRepresentative`, and `getDeals` / `getRelatedTasks` pass `ownerId = CrmUtil.isCrmSalesRepresentative(currentUser) ? currentUser.getEmployee().getEmployeeId() : null` down to the repository.

`getCompanyMetricsById` does none of this — `CrmCompanyServiceImplV2` does not even inject `UserService`. The `@PreAuthorize("hasAnyRole('ROLE_CRM_SALES_REPRESENTATIVE')")` on the controller admits reps, and they receive `openValue`, `accountValue`, `openDeals`, `closedDeals`, `openTasksCount` and `overdue` summed across every deal and task on the company, including ones owned by other reps that they cannot list or open individually. Aggregate revenue totals are exactly the kind of data the per-record checks exist to withhold.

Fix: inject `UserService`, resolve `Long ownerId = CrmUtil.isCrmSalesRepresentative(currentUser) ? currentUser.getEmployee().getEmployeeId() : null`, pass it into `getCompanyMetricsById`, and add `ownerId == null ? cb.conjunction() : cb.equal(subRoot.get(owner).get(employeeId), ownerId)` to each of the six subqueries — mirroring how `CrmDealDao.findDeals` and `CrmTaskDao.findRelatedTasks` already do it.

Note this same gap exists on the pre-existing list endpoint (`getCompanyMetricsV2`); if the intent is genuinely that company-level rollups are visible to all CRM users, that decision should be stated explicitly rather than left implicit, because it directly contradicts the per-record rules.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmCompanyControllerV2IntegrationTest.java
```diff
@@ -231,4 +232,91 @@ void getCompanyMetrics_WithoutCrmRole_ReturnsForbidden() throws Exception {
 			.andExpect(status().isForbidden());
 	}
 
+	// --- getCompanyMetricsById ---
+
+	@Test
+	@DisplayName("Get company metrics by ID - Returns aggregated deal and task metrics")
+	void getCompanyMetricsById_HappyPath_ReturnsMetrics() throws Exception {
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The new tests cannot detect a broken subquery correlation — the single most likely failure mode for this query.

`backend/src/test/resources/data.sql` contains no `crm_company`, `crm_deal`, `crm_task`, `crm_contact` or `crm_deal_stage` rows (verified by grep), and each test is `@Transactional` and rolled back. So in `getCompanyMetricsById_HappyPath_ReturnsMetrics` the company it creates is the *only* company with deals and tasks in the database. If the correlation predicates (`cb.equal(subTask.get(CrmTask_.company), company)`, `cb.equal(openDeal.get(CrmDeal_.company), company)`, etc.) were dropped or wired to the wrong root — the classic copy-paste hazard in a block replicated three times — every metric would silently become a global aggregate and all six assertions would still pass.

Fix: in the happy-path test, create a second company with its own contact, deals (in both the open and won stages) and tasks, then assert the requested company's metrics are unchanged. That single addition turns the test into a real guard on the correlation and costs almost nothing given the existing `savedCompany`/`savedDeal`/`savedTask` helpers.

backend/src/main/java/com/skapp/community/crmplanner/controller/v2/CrmCompanyControllerV2.java
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Response shape is asymmetric with the list endpoint and the resource it hangs off does not exist.

`GET /v2/crm/company` returns items shaped `{company: {...}, metrics: {...}}` (`CrmCompanyListItemDtoV2`). The new `GET /v2/crm/company/{id}/metrics` returns a bare `CrmCompanyMetrics` with no company identity at all — not even the id — so a client holding the response has no way to confirm which company it describes, and cannot reuse whatever type it already generated for the list items. There is also no `GET /v2/crm/company/{id}`, so `/{id}/metrics` is a sub-resource of a resource that has no representation in v2.

Fix: return the same `CrmCompanyListItemDtoV2` shape (company + metrics) from the by-id endpoint, or at minimum add the company id to the payload. Reusing `CrmCompanyListItemDtoV2` also lets the repository reuse the existing projection instead of adding a third copy of the metrics block.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmCompanyServiceImplV2.java
```diff
@@ -40,4 +43,16 @@ public ResponseEntityDto getCompanyMetrics(CrmCompanyMetricRequestDto requestDto
...
+		log.info("getCompanyMetricsById: execution started");
+
+		CrmCompanyMetrics metrics = crmCompanyDao.getCompanyMetricsById(id)
```

Skips `CrmValidations.validateCompanyId(id)`, which every other company-by-id service method calls.

`CrmCompanyServiceImpl.deleteCompany` (line 155) and `editCompany` (line 172) both call `CrmValidations.validateCompanyId(id)` before the repository lookup. The new method goes straight to `crmCompanyDao.getCompanyMetricsById(id)`. Today the outcome happens to be the same (a null/absent id can't reach a `@PathVariable Long`, and a missing row already yields `CRM_ERROR_COMPANY_NOT_FOUND`), so this is a consistency point rather than a live defect — but the established convention is that the id guard is applied at the service boundary, since the service interface is also reachable from other callers where a null id is possible.

Fix: add `CrmValidations.validateCompanyId(id);` as the first statement after the entry log.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmCompanyControllerV2IntegrationTest.java
```diff
@@ -231,4 +232,91 @@ void getCompanyMetrics_WithoutCrmRole_ReturnsForbidden() throws Exception {
...
+			.andExpect(jsonPath("['results'][0]['accountValue']").value("0"));
+	}
+
+	@Test
+	@DisplayName("Get company metrics by ID that does not exist - Returns Bad Request")
+	void getCompanyMetricsById_NotFound_ReturnsBadRequest() throws Exception {
+		performRequest(get(METRICS_PATH + "/999999/metrics").accept(MediaType.APPLICATION_JSON), authToken)
+			.andDo(print())
+			.andExpect(status().isBadRequest())
+			.andExpect(jsonPath(STATUS_PATH).value(STATUS_UNSUCCESSFUL));
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The two 400 tests assert only the status flag, not which error was returned.

`getCompanyMetricsById_NotFound_ReturnsBadRequest` and `getCompanyMetricsById_SoftDeleted_ReturnsBadRequest` check `status().isBadRequest()` and `STATUS_UNSUCCESSFUL`, so they would pass for *any* `ModuleException` — a wrong message constant, or an unrelated validation failure thrown before the lookup, is indistinguishable from the intended not-found. The v1 counterparts do this properly: `CrmCompanyControllerIntegrationTest:283` and `:479` assert `messageUtil.getMessage(CrmMessageConstant.CRM_ERROR_COMPANY_NOT_FOUND)`.

Fix: inject `MessageUtil` into the test class and add `.andExpect(jsonPath("$.results[0].message").value(messageUtil.getMessage(CrmMessageConstant.CRM_ERROR_COMPANY_NOT_FOUND)))` (matching the exact JSON path the v1 tests use) to both tests.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmCompanyControllerV2IntegrationTest.java
```diff
@@ -231,4 +232,91 @@ void getCompanyMetrics_WithoutCrmRole_ReturnsForbidden() throws Exception {
...
+		performRequest(get(METRICS_PATH + "/" + company.getId() + "/metrics").accept(MediaType.APPLICATION_JSON),
+				authToken)
+			.andDo(print())
+			.andExpect(status().isBadRequest())
+			.andExpect(jsonPath(STATUS_PATH).value(STATUS_UNSUCCESSFUL));
+	}
+
+	@Test
+	@DisplayName("Get company metrics by ID without CRM role - Returns Forbidden")
+	void getCompanyMetricsById_WithoutCrmRole_ReturnsForbidden() throws Exception {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

No test exercises a user whose CRM role is actually `CRM_SALES_REPRESENTATIVE`.

In `data.sql`, employee 1 (user1@gmail.com — the `authToken` user) has `crm_role = 'CRM_ADMIN'` and employee 2 (user2@gmail.com) has `crm_role = null`. Per `AuthorityServiceImpl.addRoleHierarchy`, CRM_ADMIN is granted ROLE_CRM_SALES_MANAGER and ROLE_CRM_SALES_REPRESENTATIVE, so the happy-path tests run as an admin and the forbidden test runs as a user with no CRM role at all. The role the `@PreAuthorize` actually names is never tested in isolation — which is also why the owner-scoping gap flagged above is invisible to this suite.

Fix: add a seeded employee with `crm_role = 'CRM_SALES_REPRESENTATIVE'` (or reuse one if another CRM test already establishes one) and assert the expected behaviour for that user — currently 200 with full company-wide totals, which is the behaviour worth pinning down explicitly once the scoping question is settled.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmCompanyRepositoryImpl.java
```diff
@@ -198,6 +199,65 @@ public Page<CrmCompanyListItemDtoV2> getCompanyMetricsV2(Pageable pageable, Stri
 		return new PageImpl<>(content, pageable, total);
 	}
 
+	@Override
+	public Optional<CrmCompanyMetrics> getCompanyMetricsById(Long companyId) {
+		List<Long> closedStageIds = getClosedStageIds();
```
ThinuwanW, line 204: 🤖 **Claude** · 🟣 **Nit**

`closedStageIds` can legitimately be empty, and `cb.not(x.in(emptyList))` is dialect-dependent.

`getClosedStageIds()` returns stage ids whose `stageType` is in `TERMINAL_STAGES` (WON, LOST) and which are not soft-deleted. If an installation has soft-deleted both terminal stages, or during early setup before stages are seeded, this list is empty. `openValueSubquery` (line 230) and `openCountSubquery` (line 251) then build `cb.not(stage.id.in(emptyList))`. Hibernate 6/7 normally renders an empty `IN` as `1=0`, making `NOT(1=0)` true and the behaviour correct — but some renderings emit `IN (null)`, in which case `NOT(... IN (null))` evaluates to NULL, every deal is filtered out, and the endpoint silently reports `openValue = 0` / `openDeals = 0` with no error. `getCompanyMetricsById_NoActivity_ReturnsZeroMetrics` does hit the empty-list path (it creates no stages) but has zero deals, so it cannot distinguish the two behaviours.

Flagging as a nit because the current Hibernate version most likely renders this correctly. If you want it deterministic, guard explicitly: use `closedStageIds.isEmpty() ? cb.conjunction() : cb.not(...in(closedStageIds))`. This applies to all three copies of the block, so it is best fixed when the duplication is extracted into a shared helper.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmCompanyRepositoryImpl.java
```diff
@@ -198,6 +199,65 @@ public Page<CrmCompanyListItemDtoV2> getCompanyMetricsV2(Pageable pageable, Stri
...
+
+		CriteriaBuilder cb = entityManager.getCriteriaBuilder();
+		CriteriaQuery<CrmCompanyMetrics> query = cb.createQuery(CrmCompanyMetrics.class);
+		Root<CrmCompany> company = query.from(CrmCompany.class);
+
+		Subquery<Long> taskSubquery = query.subquery(Long.class);
+		Root<CrmTask> subTask = taskSubquery.from(CrmTask.class);
+		taskSubquery.select(cb.count(subTask.get(CrmTask_.id)))
+			.where(cb.equal(subTask.get(CrmTask_.company), company), cb.isFalse(subTask.get(CrmTask_.isDeleted)),
```
ThinuwanW, line 213: 🤖 **Claude** · 🟣 **Nit**

Task metrics under-count tasks that are attached to a deal but not to a contact.

`openTasksCount` and `overdue` match on `CrmTask_.company` only. In `CrmTaskServiceImpl.persistNewTask` (lines 189-203), `task.setCompany(...)` is populated *only* from `contact.getCompany()` when `contactId` is supplied; when a task is created against a `dealId` alone, `company_id` stays null even though the deal has a `company_id`. `crm_task.company_id` is nullable in the DDL, so those rows exist. The result is that a company-scoped task metric silently omits deal-only tasks.

This is inherited from the existing list query rather than introduced here, but the new endpoint is specifically named as *the* per-company metrics resource, so the discrepancy becomes more visible. Fix: either widen the task subqueries to also match `subTask.deal.company = company`, or backfill `task.setCompany(deal.getCompany())` in `persistNewTask`/`applyTaskEdit` when only a deal is supplied — the latter is preferable since it makes every consumer of `crm_task.company_id` consistent.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmCompanyRepositoryImpl.java
```diff
@@ -198,6 +199,65 @@ public Page<CrmCompanyListItemDtoV2> getCompanyMetricsV2(Pageable pageable, Stri
...
+		Subquery<Long> overdueSubquery = query.subquery(Long.class);
+		Root<CrmTask> subOverdueTask = overdueSubquery.from(CrmTask.class);
+		overdueSubquery.select(cb.count(subOverdueTask.get(CrmTask_.id)))
+			.where(cb.equal(subOverdueTask.get(CrmTask_.company), company),
+					cb.isFalse(subOverdueTask.get(CrmTask_.isDeleted)),
+					cb.isFalse(subOverdueTask.get(CrmTask_.isCompleted)),
+					cb.isNotNull(subOverdueTask.get(CrmTask_.dueAt)),
+					cb.lessThan(subOverdueTask.get(CrmTask_.dueAt), cb.localDateTime()));
+
+		Subquery<BigDecimal> openValueSubquery = query.subquery(BigDecimal.class);
```
akilarootcode, line 225: remove bigdecimal

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmCompanyControllerV2IntegrationTest.java
```diff
@@ -231,4 +232,91 @@ void getCompanyMetrics_WithoutCrmRole_ReturnsForbidden() throws Exception {
...
+		assertThat(new BigDecimal(openValue)).as("open value sums non-closed deals").isEqualByComparingTo("200");
+		assertThat(new BigDecimal(accountValue)).as("account value sums WON deals").isEqualByComparingTo("400");
+	}
+
+	@Test
+	@DisplayName("Get company metrics by ID with no deals or tasks - Returns zero metrics")
+	void getCompanyMetricsById_NoActivity_ReturnsZeroMetrics() throws Exception {
+		CrmCompany company = savedCompany("EmptyMetricsCoV2");
+
+		performRequest(get(METRICS_PATH + "/" + company.getId() + "/metrics").accept(MediaType.APPLICATION_JSON),
```
akilarootcode: check tests again

akilarootcode (changes requested):

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmCompanyServiceImplV2.java
anusham-rc: Intentional — these are company-wide metrics, same as the existing list-metrics endpoints. will handle separately if needed

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmCompanyRepositoryImpl.java
anusham-rc: Acknowledged. We're intentionally keeping the three metric queries (`getCompanyMetrics`, `getCompanyMetricsV2`, `getCompanyMetricsById`) self-contained rather than extracting a shared builder — each is an independent criteria query, so we'll keep them independent by design.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmCompanyRepositoryImpl.java
anusham-rc: The double-count described here requires a soft-deleted WON/LOST stage (whose id would drop out of `getClosedStageIds()` while its deals still match `stageType = WON`). That state can't occur — terminal (WON/LOST) stages cannot be deleted in the product — so `stage.id NOT IN closedStageIds` and `stageType = WON` never disagree in practice. Leaving the logic as-is.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmCompanyServiceImplV2.java
anusham-rc: This follows the existing convention — the sibling metrics type `CrmContactMetrics` also lives in `crmplanner/type/` and is returned directly. Keeping it consistent with the established layout rather than introducing a separate placement for this one type.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmCompanyControllerV2IntegrationTest.java
anusham-rc: The LOST-stage semantics are already covered:
- `getCompanyMetrics_ClassifiesWonAndLostAsClosed` — WON and LOST both excluded from open; only WON counted as closed.
- `getCompanyMetrics_LostOnlyCompany_ReportsZeroOpenAndZeroClosed` — a LOST-only company reports zero open and zero closed value/count.

These exercise the same subquery predicates the by-id endpoint uses.

backend/src/main/java/com/skapp/community/crmplanner/controller/v2/CrmCompanyControllerV2.java
anusham-rc: Out of scope for this PR — a non-numeric `@PathVariable Long` raising `MethodArgumentTypeMismatchException` is handled globally in `GlobalExceptionHandler` and applies to every `Long` path variable across the app, not just this endpoint. Best addressed centrally rather than per-endpoint.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmCompanyServiceImplV2.java
anusham-rc: This matches the existing convention — `ModuleException` maps to 400 across the CRM endpoints (e.g. the other company-by-id methods). Changing just this endpoint to 404 would make the API inconsistent; a 400→404 alignment would be a codebase-wide change.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmCompanyControllerV2IntegrationTest.java
anusham-rc: Fixed. The no-activity test now reads `openValue`/`accountValue` and asserts with `new BigDecimal(...).isEqualByComparingTo("0")`, matching the happy-path style. Value comparison is now independent of DB `CAST` string formatting.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmCompanyControllerV2IntegrationTest.java
anusham-rc: Fixed. The happy-path test now seeds a second company (`OtherMetricsCo`) with its own open/won deals (999/888) and tasks. If any subquery lost its `deal/task.company = company` correlation, the target company's aggregates would pick up the other company's rows and the assertions would fail.

backend/src/main/java/com/skapp/community/crmplanner/controller/v2/CrmCompanyControllerV2.java
anusham-rc: Acknowledged. The by-id endpoint returns the metrics object directly, which is the natural shape for a single-resource read; the list endpoint wraps items because it paginates. We're keeping the current shape.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmCompanyServiceImplV2.java
anusham-rc: The outcome is equivalent: the criteria query filters `isDeleted = false` and a missing/soft-deleted company yields an empty result, which `orElseThrow` maps to `CRM_ERROR_COMPANY_NOT_FOUND` (the same error `validateCompanyId` raises). Can add the explicit call for consistency if preferred, but there's no behavioral gap.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmCompanyControllerV2IntegrationTest.java
anusham-rc: Fixed. `getCompanyMetricsById_NotFound_...` and `..._SoftDeleted_...` now also assert the response message equals `messageUtil.getMessage(CrmMessageConstant.CRM_ERROR_COMPANY_NOT_FOUND)`, so a wrong-but-still-400 error would be caught.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmCompanyControllerV2IntegrationTest.java
anusham-rc: Tied to the owner-scoping decision. Since the endpoint currently returns a company-wide rollup for every CRM role by design, a rep-specific assertion would just duplicate the existing coverage. We'll add role-scoped tests together with the scoping decision.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmCompanyRepositoryImpl.java
anusham-rc: In practice `closedStageIds` is never empty: WON/LOST stages always exist and cannot be deleted, so `getClosedStageIds()` always returns at least the terminal stages. The empty-list branch isn't reachable, so keeping the current predicate.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmCompanyRepositoryImpl.java
anusham-rc: Pre-existing behavior — `openTasksCount`/`overdue` match on `CrmTask.company` only, the same as the other company metrics methods. Aligning task attribution (company vs deal vs contact) is a broader change that should be applied consistently across all metrics, so it's out of scope here.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmCompanyRepositoryImpl.java
anusham-rc: The output fields (`openValue`, `accountValue`) are already `String` — the response type `CrmCompanyMetrics` exposes no `BigDecimal`, matching `CrmContactMetrics.closedDealValue: String`.

The only `BigDecimal` is internal: `CrmDeal.amount` is stored as a `String` column, so to `SUM()` it we cast `String → BigDecimal` inside the subquery, then cast the total back to `String` for the response. This is the same pattern already used in `CrmContactRepositoryImpl` (`closedValueSub`), so removing it here would diverge from the established approach and break the numeric sum. Output values are unchanged.

akilarootcode (dismissed):

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmCompanyControllerV2IntegrationTest.java
akilarootcode: checked

akilarootcode (dismissed):

akilarootcode (approved):

PR #1743, opened 2026-08-11
fix: trim leading whitespace from search input in CompanyTable

dulsara-manakal:
### Summary
- Prevent leading spaces from being typed into the company search field, by trimming on input change (trimStart()), consistent with how other search inputs handle it (e.g., Teams, Directory).

## Scope

This PR is scoped strictly to that one fix. It does not touch backend search logic, ranking, or any other CRM search field, out of scope for this change.

## Note on backend trimming

No backend `.trim()` was added. Incoming request strings are already trimmed globally via JacksonConfig, so an additional trim in the repository layer would be redundant, the only gap was on the frontend, at the point where the input value is captured.

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

Single-line fix that strips leading whitespace from the CompanyTable search input before it enters state. The change is correct, matches the established codebase convention (12+ call sites use `event.target.value.trimStart()`, e.g. EmployeeData.tsx:92, CustomerTable.tsx:124), and safely preserves the controlled-input contract since React restores the DOM value after the handler runs. Remaining observations are non-blocking: trailing whitespace is still forwarded into the React Query key and API param, and the two sibling search inputs in the same CRM module still have the unfixed behavior.

Found **11** new issue(s): 🔴 1 important, 🟡 5 suggestion(s), 🟣 5 nit(s)

frontend/src/community/crm/components/organisms/CompanyTable/CompanyTable.tsx
```diff
@@ -57,8 +57,7 @@ export const CompanyTable: FC = () => {
   }, [fetchedCompanies]);
 
   const handleSearchChange = (event: ChangeEvent<HTMLInputElement>) => {
-    const value = event.target.value;
-    setSearchTerm(value);
+    setSearchTerm(event.target.value.trimStart());
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`trimStart()` fixes leading whitespace, but trailing whitespace is still forwarded to the server. `debouncedSearch` is passed straight into `useGetCompanyMetrics` (line 41), which builds the key `companyQueryKeys.GET_COMPANY_DATA_BY_SEARCH(searchKeyword, limit)` -> `["get-company-data", searchKeyword, limit]`. So "acme" and "acme " produce two distinct cache entries and two separate network requests for what the user perceives as the same search, and the backend `searchKeyword` LIKE match may return nothing for the trailing-space variant. Recommended fix: keep `trimStart()` for the displayed value (so the user can still type multi-word searches) and trim the value actually used for fetching, i.e. pass `debouncedSearch.trim()` to `useGetCompanyMetrics` on line 41. That also makes the existing `debouncedSearch.trim() === ""` check on line 36 consistent with what was queried.

ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

The same bug still exists in two near-identical siblings in this module, so the CRM search inputs now behave inconsistently: `ContactTable.tsx:222` (`onChange={(event) => setSearchTerm(event.target.value)}`) and `TaskTabContent.tsx:38` (`setSearchTerm(event.target.value)`). Both feed a `useDebounce` value into an infinite query whose key includes the raw search keyword, exactly like CompanyTable did. Consider applying the same `trimStart()` in this PR so a user typing a leading space gets the same behavior on every CRM table. Given this sanitizer is now repeated at 13 sites across the codebase, a tiny shared helper (e.g. `sanitizeSearchInput(value)` in `~community/common/utils`) used by all three CRM handlers would remove the duplication and keep future tables consistent.

ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

No test locks in this behavior — there is no `CompanyTable.test.tsx`, and the CRM module currently has only util-level tests (`utils/__test__/companyUtil.test.ts`, etc.), so nothing prevents a future refactor from reintroducing the untrimmed value. Cheapest option consistent with the module's existing test layout: extract the sanitizer into a CRM util and add a unit test for it (empty string on whitespace-only input, leading whitespace removed, inner/trailing spaces preserved). Alternatively add a component test following the pattern in `src/community/attendance/components/...` (RTL + `MockTheme`, `screen.getByRole("searchbox")`, `fireEvent.change`).

ThinuwanW: 🤖 **Claude** · 🔴 **Important**

📁 **File-level observation**

No error state is rendered, and combined with the Zustand mirror this shows stale rows as search results. `useGetCompanyMetrics` is destructured for `data`/`isLoading` only — `isError` is ignored. The table renders `companies ?? []` from the Zustand store, which is only ever written when `fetchedCompanies` is truthy. Failure scenario: user has searched 'acme' (store holds Acme rows), then types 'zzz'. React Query creates a new cache entry, `data` is undefined, and if that request fails, `isLoading` flips back to false while `data` stays undefined — so the effect never fires, the store still holds the Acme rows, and the table renders them under the search term 'zzz'. The user sees wrong results with no indication of failure. Fix: destructure `isError` and render an error view before falling through to the table, following the established pattern in `TaskTabContent.tsx` (which checks `isOpenTasksError`/`isCompletedTasksError` and renders `EmptyDataView` with `table.errorState.title`/`description`). Note `companies.table.errorState` does not exist in `crmModule.json` yet (the tasks table has one at line 781) — add the key alongside the code change. Also consider clearing/ignoring the store while a new query key is pending so stale rows can never be attributed to a new term.

ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

The empty-state logic at lines 35-38 duplicates an existing shared helper. `getEmptyStateType` in `~community/crm/utils/crmUtil.ts:128` is character-for-character the same ternary (`searchTerm.trim() === "" ? NO_DATA : NO_SEARCH_RESULTS`) and is already consumed by `TaskTabContent.tsx:74`. Replace the inline block with `const emptyStateType = getEmptyStateType(debouncedSearch);` and drop the now-unused `EmptyStateTypeEnum` import. This is worth doing in this PR specifically because that helper is the natural single place for search-term normalization — three CRM tables currently each re-derive it inline, which is how the inconsistency between them arose.

ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Every keystroke rebuilds the full table render tree. `searchTerm` state lives in this component, so each character typed re-renders it; the `columns` array (lines 63-148, five entries each with a fresh `render` closure), `loadMore`, `handleSearchChange`, and the inline `onRowClick` arrow are all recreated with new identities on every render and passed straight into `Table`. With a 500ms debounce the query is throttled, but the table's row rendering is not — it re-renders on every character. Fix: wrap `columns` in `useMemo` with `[translateText]` as the dependency, and wrap `loadMore`, `handleSearchChange` and the row-click handler in `useCallback`. The guidelines call this out explicitly (missing `useMemo`/`useCallback` on callbacks passed to child components).

ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Null-safety contradiction between the component and the query's `getNextPageParam`. `fetchCompanyMetrics` (CompanyApi.ts:34-42) returns `response?.data?.results?.[0]` but is typed as a non-optional `Promise<CrmCompanyMetricsResponseType>`, so `undefined` is reachable at runtime when the response envelope has no results. Line 52 here defensively writes `page?.items ?? []`, which shows the author expects a nullish page — but `getNextPageParam` at CompanyApi.ts:55 dereferences `lastPage.currentPage` and `lastPage.totalPages` with no guard, so the same condition would throw a TypeError inside React Query and surface as a query error (which, per the finding above, renders as a misleading empty state). Pick one contract: either widen the return type to `CrmCompanyMetricsResponseType | undefined` and add `if (!lastPage) return undefined;` at the top of `getNextPageParam`, or guarantee non-null in the fetcher and remove the redundant optional chain on line 52. The same unguarded pattern exists in `crmDealApi.ts:52`, `ContactApi.ts:77` and `TaskApi.ts:42`/`161`, so fixing it consistently is preferable.

frontend/src/community/crm/components/organisms/CompanyTable/CompanyTable.tsx
```diff
@@ -57,8 +57,7 @@ export const CompanyTable: FC = () => {
   }, [fetchedCompanies]);
```
ThinuwanW, line 55: 🤖 **Claude** · 🟣 **Nit**

`setCompanies` is missing from this effect's dependency array. Zustand setters are referentially stable so this is not a live bug, but it violates the exhaustive-deps rule and will be flagged if lint rules tighten. Since this PR is already touching the search flow that feeds this effect, add `setCompanies` to the array.

frontend/src/community/crm/components/organisms/CompanyTable/CompanyTable.tsx
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Sanitizing inside a controlled input's onChange moves the caret to the end of the field. Reproduction: the field contains 'Acme', the user clicks to place the caret at position 0 and types a space. The DOM value becomes ' Acme', `trimStart()` produces 'Acme', which equals the current state, so React bails out of re-rendering — but its controlled-input restore step then assigns `node.value` back to 'Acme', and assigning `value` on an input resets the selection to the end. The caret jumps away from where the user was editing. Low impact for a search box, but if you want to avoid it, sanitize at the consumption boundary instead of the input boundary — keep `setSearchTerm(event.target.value)` and normalize when building the query argument, e.g. pass a trimmed value into `useGetCompanyMetrics`. That also resolves the trailing-whitespace gap in one place rather than two.

ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

`columns as TableColumn<any>[]` (line 185) trips two guideline rules at once — an `any` type and an `as` assertion used to silence a type error rather than fix it — under a strict-mode tsconfig. The cast is needed only because `Table` is not generic over the row type at this call site. If `@rootcodelabs/skapp-ui`'s `Table` accepts a row type parameter, use `<Table<CrmCompany> columns={columns} ...>` and delete the cast; if it genuinely does not, add a short comment explaining the cast is a library-typing limitation so it is not mistaken for an oversight. The identical cast appears at `ContactTable.tsx:237`, so fix both or neither.

ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

Accessibility inconsistency with the sibling table, separate from the whitespace issue. CompanyTable's search input passes `aria-label={translateText(["table", "searchAriaLabel"])}` (line 164), backed by `companies.table.searchAriaLabel` in `crmModule.json:296`. `ContactTable.tsx:211-224` has no `aria-label` at all, and `contacts.table` (crmModule.json:175-193) has no `searchAriaLabel` key to supply one — so that search box is announced only by its placeholder. Since this PR is already normalizing behaviour across these two inputs, add the `contacts.table.searchAriaLabel` translation key (e.g. 'Search contacts by name') and wire the `aria-label` prop on ContactTable's InputField to match. The guidelines list missing accessibility-label translations as a flaggable i18n gap.

dulsara-manakal: Out of scope. not related to this bug fix.

dulsara-manakal: No need to test for this since this is not even a util or a hook.

dulsara-manakal: Trailing space is acceptable according to requirements.

dulsara-manakal: No need for this error state we have an empty state in case data fail and also according to the requirements, this actualy not valid to show that kind of error state.

dulsara-manakal: Out of scope.

dulsara-manakal: Keep it consistent with other places, and keep this as it is. Also, this is out of scope of this bug fix.

frontend/src/community/crm/components/organisms/CompanyTable/CompanyTable.tsx
dulsara-manakal: setters doesn't want to be in the dependency array, it doesn't make any impact. also not violate our ESLint config.

frontend/src/community/crm/components/organisms/CompanyTable/CompanyTable.tsx
dulsara-manakal: This actually handles the consuming component. This is out of scope anyway.

dulsara-manakal: This is related to the company search, not even a touched file.

dulsara-manakal: That's the existing behaviour and also the requirement.

dulsara-manakal: he any cast is only there because Table isn't generic over the row type, TypeScript can't verify columns matches without it.

hasalarootcode (dismissed):

frontend/src/community/crm/components/organisms/CompanyTable/CompanyTable.tsx
```diff
@@ -30,10 +30,11 @@ export const CompanyTable: FC = () => {
 
   const [searchTerm, setSearchTerm] = useState("");
   const debouncedSearch = useDebounce(searchTerm, COMPANY_NAME_DEBOUNCE_DELAY);
```
akilarootcode: trim before going to debounce

akilarootcode (changes requested):

frontend/src/community/crm/components/organisms/CompanyTable/CompanyTable.tsx
```diff
@@ -29,8 +29,11 @@ export const CompanyTable: FC = () => {
   const translateText = useTranslator("crmModule", "companies");
 
   const [searchTerm, setSearchTerm] = useState("");
-  const debouncedSearch = useDebounce(searchTerm, COMPANY_NAME_DEBOUNCE_DELAY);
-  const isEmptyFilterState = debouncedSearch.trim() !== "";
+  const debouncedSearch = useDebounce(
+    searchTerm.trim(),
+    COMPANY_NAME_DEBOUNCE_DELAY
+  );
+  const isEmptyFilterState = debouncedSearch !== "";
```
akilarootcode: proper variable name

akilarootcode (changes requested):

akilarootcode (dismissed):

frontend/src/community/crm/components/organisms/CompanyTable/CompanyTable.tsx
dulsara-manakal: Fixed

frontend/src/community/crm/components/organisms/CompanyTable/CompanyTable.tsx
dulsara-manakal: Changed the name

akilarootcode (approved):

PR #1744, opened 2026-08-11
feat(crm): add GET /v1/crm/contact/{id}/metrics endpoint

anusham-rc:
## What

Adds `GET /v1/crm/contact/{id}/metrics` — the aggregated deal/task metrics for a single contact.

- **Response:** `CrmContactMetrics` (closedDealValue, closedDealCount, openTasksCount, overdueTasksCount).
- **Resolution:** a dedicated criteria query filters on `id` + `isDeleted = false`; missing **or soft-deleted** → **400** (`CRM_ERROR_CONTACT_NOT_FOUND`); existing-but-empty → zeroed metrics.
- **Auth:** `@PreAuthorize("hasRole('ROLE_CRM_SALES_REPRESENTATIVE')")`, matching the v1 contact controller.

Placed on the **v1** controller (not v2): it is a brand-new endpoint with no existing consumer, so there is nothing to keep compatible on v1 and no later v1→v2 migration — keeping it on v1 avoids fragmenting the contact surface across versions. The new `getContactMetricsById` repository query is self-contained (the list query is untouched).

## Tests

`CrmContactControllerIntegrationTest` (v1) — aggregated-metrics (seeded WON deals + open/overdue tasks), zero-metrics, not-found → 400, without-CRM-role → 403. Contact v1 + v2 suites green (62 + 12).

## Notes


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds a `GET /v2/crm/contact/{id}/metrics` endpoint (controller → service → custom repository) plus four integration tests. The layering, `@PreAuthorize` role (consistent with siblings, and `ROLE_CRM_ADMIN`/`ROLE_CRM_SALES_MANAGER` inherit it via `AuthorityServiceImpl.addRoleHierarchy`), `@Transactional(readOnly = true)`, entry/exit logging, `ModuleException` with an existing message constant, and soft-delete filtering are all correct, and I found no functional bug — the new query is a faithful copy of the metrics computation already used by `getContactMetricsV2`. The main concerns are the ~35 lines of verbatim duplicated Criteria subquery code inside a single class (which will silently drift), a JVM-timezone/inlined-literal overdue check that is also inconsistent with the company metrics implementation, and test hygiene/coverage gaps (no soft-deleted-contact case, no error-message assertion, hand-concatenated URLs bypassing existing helpers).

Found **12** new issue(s): 🔴 2 important, 🟡 3 suggestion(s), 🟣 7 nit(s)

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
```diff
@@ -166,6 +167,48 @@ private Predicate[] buildPredicates(CriteriaBuilder cb, Root<CrmContact> contact
 		return predicates.toArray(new Predicate[0]);
 	}
 
+	@Override
```
ThinuwanW, line 137: 🤖 **Claude** · 🔴 **Important**

The four metrics subqueries (closed deal value, closed deal count, open tasks, overdue tasks) in `getContactMetricsById` are a byte-for-byte copy of lines 85-124 in `getContactMetricsV2` in this same class — same aliases, same predicates, same `cb.construct(CrmContactMetrics.class, ...)` argument order. This is exactly the duplication the guidelines call out (2+ methods with near-identical bodies → extract a shared helper), and here the risk is concrete: the list endpoint and the by-id endpoint must return identical numbers for the same contact, but any future change to one block (e.g. fixing the overdue-task definition, excluding deals whose stage is soft-deleted, or switching the WON check to stage IDs) will only be applied to one of them and the two endpoints will silently disagree. Extract a private helper alongside the existing `buildPredicates`/`buildOrderBy` helpers, e.g. `private Selection<CrmContactMetrics> buildMetricsSelection(CriteriaBuilder cb, CriteriaQuery<?> query, Root<CrmContact> contact)` that creates the four correlated subqueries and returns `cb.construct(CrmContactMetrics.class, closedValueSub.cast(String.class), closedCountSub, openTaskSub, overdueTaskSub)`, then call it from both methods. Note `buildOrderBy` already takes `CriteriaQuery<?>` and creates a subquery, so the same signature pattern works.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
```diff
@@ -166,6 +167,48 @@ private Predicate[] buildPredicates(CriteriaBuilder cb, Root<CrmContact> contact
...
+			.where(cb.equal(openTask.get(CrmTask_.contact), contact), cb.isFalse(openTask.get(CrmTask_.isCompleted)),
+					cb.isFalse(openTask.get(CrmTask_.isDeleted)));
+
+		Subquery<Long> overdueTaskSub = query.subquery(Long.class);
+		Root<CrmTask> overdueTask = overdueTaskSub.from(CrmTask.class);
+		overdueTaskSub.select(cb.count(overdueTask.get(CrmTask_.id)))
+			.where(cb.equal(overdueTask.get(CrmTask_.contact), contact),
+					cb.isFalse(overdueTask.get(CrmTask_.isCompleted)), cb.isFalse(overdueTask.get(CrmTask_.isDeleted)),
+					cb.isNotNull(overdueTask.get(CrmTask_.dueAt)),
+					cb.lessThan(overdueTask.get(CrmTask_.dueAt), cb.literal(LocalDate.now().atStartOfDay())));
```
ThinuwanW, line 178: 🤖 **Claude** · 🟣 **Nit**

The overdue-task cutoff `cb.lessThan(dueAt, cb.literal(LocalDate.now().atStartOfDay()))` has three drawbacks that this PR propagates into a second call site. (1) `LocalDate.now()` resolves in the JVM default timezone, so the boundary shifts with the app server's zone rather than the DB's. (2) `cb.literal(...)` inlines the timestamp into the SQL text instead of binding a parameter, so every distinct day/instant produces a new statement and defeats prepared-statement/plan caching. (3) It is inconsistent with the equivalent company metrics query, which uses `cb.localDateTime()` (CrmCompanyRepositoryImpl.java:70 and :145) — so a task due at 09:00 today is counted as overdue in company metrics at 15:00 but never in contact metrics. Consider switching contact metrics to `cb.localDateTime()` (or to a bound parameter) so contact and company overdue counts agree; if the start-of-day semantics are intentional for contacts, a short comment stating that would prevent it from being 'fixed' later. Behaviour is unchanged relative to the existing list endpoint, so this is not a regression — just worth resolving while both call sites are being touched.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
ThinuwanW, line 137: 🤖 **Claude** · 🟣 **Nit**

The new public `getContactMetricsById` is inserted between the private helpers `buildPredicates` (ends line 168) and `getContactTotalCount` (line 212), breaking the class's existing grouping where public `@Override` methods precede their private helpers. Move it up next to `getContactMetricsV2` (after line 135) so the two metrics queries sit together — which also makes the duplication noted above obvious to future readers.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmContactControllerV2IntegrationTest.java
```diff
@@ -391,4 +392,75 @@ void editContact_WithoutCrmRole_ReturnsForbidden() throws Exception {
...
+
+		CrmDealStage openStage = savedStage("V2 Open Stage", CrmDealStageType.OPEN, 1);
+		CrmDealStage wonStage = savedStage("V2 Won Stage", CrmDealStageType.WON, 2);
+		savedDeal(contact, openStage, "150", "a0");
+		savedDeal(contact, wonStage, "400", "a1");
+		savedDeal(contact, wonStage, "600", "a2");
+		savedTask(contact, LocalDateTime.now().plusDays(3));
+		savedTask(contact, LocalDateTime.now().minusDays(2));
+
+		String content = performRequest(
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

The three new tests bypass the conventions already established in this test class: (a) the URL is hand-concatenated as `BASE_PATH + "/" + contact.getId() + "/metrics"` in three places instead of adding a `private static final String METRICS_BY_ID_PATH = BY_ID_PATH + "/metrics";` constant and using `get(METRICS_BY_ID_PATH, id)` with a URI variable (as `performGetByIdRequest`/`performPatchRequest` do); (b) `authToken` is passed explicitly to `performRequest(request, token)` even though the single-argument `performRequest(request)` overload at line 102 already defaults to it — only the forbidden-role test needs the two-arg form; (c) JSON paths are written as raw `"['results'][0]['closedDealCount']"` even though `RESULTS_0_PATH` is imported and used elsewhere in this file (e.g. line 218). Adding a small `performGetMetricsByIdRequest(Long id)` helper and using `RESULTS_0_PATH + "['closedDealCount']"` would keep the new block consistent with the rest of the class.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmContactControllerV2IntegrationTest.java
```diff
@@ -391,4 +392,75 @@ void editContact_WithoutCrmRole_ReturnsForbidden() throws Exception {
...
+			.andExpect(jsonPath(STATUS_PATH).value(STATUS_SUCCESSFUL))
+			.andExpect(jsonPath("['results'][0]['closedDealValue']").value("0"))
+			.andExpect(jsonPath("['results'][0]['closedDealCount']").value(0))
+			.andExpect(jsonPath("['results'][0]['openTasksCount']").value(0))
+			.andExpect(jsonPath("['results'][0]['overdueTasksCount']").value(0));
+	}
+
+	@Test
+	@DisplayName("Get contact metrics by ID that does not exist - Returns Bad Request")
+	void getContactMetricsById_NotFound_ReturnsBadRequest() throws Exception {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Coverage gaps in the new tests. (1) `getContactMetricsById_NotFound_ReturnsBadRequest` only asserts status 400 and `status == unsuccessful`; it does not assert the message is the contact-not-found message, so the test would still pass if the 400 came from an unrelated cause (e.g. a path-variable conversion failure). The v1 tests assert `jsonPath(MESSAGE_PATH).value(messageUtil.getMessage(CrmMessageConstant.CRM_ERROR_CONTACT_NOT_FOUND))` (CrmContactControllerIntegrationTest.java:508) — do the same here (this class would need to inject `MessageUtil`). (2) The repository query filters `cb.isFalse(contact.get(CrmContact_.isDeleted))`, but no test covers a soft-deleted contact returning 400 — that predicate is currently unverified. (3) Nothing verifies the subquery correlation, i.e. that deals/tasks belonging to a *different* contact are excluded; seeding a second contact with its own WON deal and asserting the first contact's totals are unaffected would catch a mis-correlated subquery, which is the most likely way this query could break during a future refactor.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmContactControllerV2IntegrationTest.java
```diff
@@ -391,4 +392,75 @@ void editContact_WithoutCrmRole_ReturnsForbidden() throws Exception {
 			.accept(MediaType.APPLICATION_JSON), noRoleToken).andDo(print()).andExpect(status().isForbidden());
 	}
 
+	// --- getContactMetricsById ---
+
+	@Test
+	@DisplayName("Get contact metrics by ID - Returns aggregated deal and task metrics")
+	void getContactMetricsById_HappyPath_ReturnsMetrics() throws Exception {
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`getContactMetricsById_HappyPath_ReturnsMetrics` duplicates the fixture setup and assertion values of `getContactMetrics_WithDealsAndTasks_ReturnsAggregatedMetrics` (lines 303-330): same two stages, same three deal amounts (150/400/600), same two tasks, same expected 2/2/1 and 1000. Extract the seeding into a helper such as `private void seedDealsAndTasks(CrmContact contact)` (and reuse the expected-total constants) so the two tests cannot drift apart while supposedly asserting the same aggregation.

backend/src/main/java/com/skapp/community/crmplanner/controller/v2/CrmContactControllerV2.java
```diff
@@ -61,4 +61,13 @@ public ResponseEntity<ResponseEntityDto> getContactMetrics(CrmContactMetricReque
 		return new ResponseEntity<>(response, HttpStatus.OK);
 	}
 
+	@Operation(summary = "Get contact metrics by ID",
+			description = "Returns the aggregated deal and task metrics for a single contact.")
+	@PreAuthorize("hasRole('ROLE_CRM_SALES_REPRESENTATIVE')")
+	@GetMapping("/{id}/metrics")
+	public ResponseEntity<ResponseEntityDto> getContactMetricsById(@PathVariable Long id) {
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Response-shape asymmetry between the two endpoints that expose the same metrics: `GET /v2/crm/contact` returns `CrmContactListItemDtoV2`, so a client reads `results[0].items[0].metrics.closedDealValue`, whereas this new endpoint returns the bare `CrmContactMetrics`, so the same four fields appear flattened at `results[0].closedDealValue`. A client cannot reuse one deserialization path for both. Consider wrapping the payload so metrics stay under a `metrics` key (or returning the same `CrmContactListItemDtoV2` shape) for consistency. Related: `CrmContactMetrics` lives in `type/` and is not named `*ResponseDto` — acceptable while it was only a nested field (and `CrmCompanyMetrics` sets the same precedent), but now that it is a top-level response body it would be more consistent with the guidelines to move/rename it under `payload/response/v2` as a `*ResponseDto`. Both points are judgement calls, not defects.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
```diff
@@ -166,6 +167,48 @@ private Predicate[] buildPredicates(CriteriaBuilder cb, Root<CrmContact> contact
...
+		Root<CrmDeal> countDeal = closedCountSub.from(CrmDeal.class);
+		closedCountSub.select(cb.count(countDeal.get(CrmDeal_.id)))
+			.where(cb.equal(countDeal.get(CrmDeal_.contact), contact),
+					cb.equal(countDeal.get(CrmDeal_.stage).get(CrmDealStage_.stageType), CrmDealStageType.WON),
+					cb.isFalse(countDeal.get(CrmDeal_.isDeleted)));
+
+		Subquery<Long> openTaskSub = query.subquery(Long.class);
+		Root<CrmTask> openTask = openTaskSub.from(CrmTask.class);
+		openTaskSub.select(cb.count(openTask.get(CrmTask_.id)))
+			.where(cb.equal(openTask.get(CrmTask_.contact), contact), cb.isFalse(openTask.get(CrmTask_.isCompleted)),
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Task attribution in the new query diverges from the existing single-contact task metrics query, so the same contact will report different open/overdue task counts depending on which endpoint the client calls.

The new `openTaskSub`/`overdueTaskSub` correlate only on the direct link (`CrmTask.contact == contact`). The pre-existing `CrmTaskRepositoryImpl.findTaskMetricsByContactId(Long)` — which backs `openTasksCount`/`overdueTasksCount` on `GET /v1/crm/contact/{id}` (see `CrmContactServiceImpl.getContactById`) — counts a task if it is linked to the contact directly OR through one of the contact's deals (`task.deal.contact.id = contactId`). `CrmTaskRepositoryImpl.findByContactIdWithAssociations` uses the same direct-OR-via-deal rule to build the task list shown on the contact detail screen, and `CrmContactServiceImpl.deleteContact` soft-deletes `findByDeal_Contact_IdAndIsDeletedFalse` tasks as part of deleting a contact — so "tasks belonging to a contact" is established elsewhere in this module as including deal-linked tasks.

Concretely: a contact with one deal that has 3 open tasks attached to the deal (and none attached to the contact directly) reports `openTasksCount: 3` from `GET /v1/crm/contact/{id}` and `openTasksCount: 0` from the new `GET /v2/crm/contact/{id}/metrics`, while the contact's task list on screen shows 3.

Fix: pick one rule and state it. Either (a) reuse the existing single-contact query — call `crmTaskDao.findTaskMetricsByContactId(id)` for the task half of the metrics instead of hand-rolling two more subqueries, or (b) add the deal-linked branch here, i.e. wrap the contact predicate in an OR over `openTask.get(CrmTask_.contact)` and `openTask.get(CrmTask_.deal).get(CrmDeal_.contact)` (LEFT joins). Either way, add an integration test that attaches a task to a deal but not to the contact, so the chosen semantics are locked in — none of the four new tests exercise a deal-linked task.

backend/src/main/java/com/skapp/community/crmplanner/controller/v2/CrmContactControllerV2.java
```diff
@@ -61,4 +61,13 @@ public ResponseEntity<ResponseEntityDto> getContactMetrics(CrmContactMetricReque
 		return new ResponseEntity<>(response, HttpStatus.OK);
 	}
 
+	@Operation(summary = "Get contact metrics by ID",
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The new v2 metrics endpoint exposes a strict subset of the per-contact metrics v1 already returns, and renames one of them, so there is no v2 way to obtain the rest.

`GET /v1/crm/contact/{id}` returns `totalRevenue`, `pipelineRevenue`, `activeDealsCount`, `openTasksCount`, `overdueTasksCount` (computed by the existing `CrmDealRepositoryImpl.findDealMetricsByContactId` + `CrmTaskRepositoryImpl.findTaskMetricsByContactId`). The new endpoint returns `closedDealValue`, `closedDealCount`, `openTasksCount`, `overdueTasksCount`. `closedDealValue` is the same figure v1 calls `totalRevenue` (sum of WON deal amounts) under a different name, and `pipelineRevenue`/`activeDealsCount` have no v2 equivalent anywhere — v2's `getContactById` deliberately returns the base contact with no metrics. Note also that `findDealMetricsByContactId` is an existing single-contact deal-metrics query that this PR does not reuse even though it computes the WON sum already.

Separately, `closedDealValue` is serialized as a JSON string here, whereas v1's `CrmContactListItemDto.closedDealValue` is a `BigDecimal` (JSON number) — same field name, different wire type across versions.

Fix: if this endpoint is meant to supersede the v1 detail metrics, extend it to the full set (reusing `findDealMetricsByContactId` for the deal half) and align the field name with `totalRevenue`, or explicitly document in the `@Operation` description that v2 intentionally narrows the contract and which v1 fields are dropped.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmContactServiceImplV2.java
```diff
@@ -50,6 +51,18 @@ public ResponseEntityDto getContactMetrics(CrmContactMetricRequestDto filterDto)
...
+	@Override
+	@Transactional(readOnly = true)
+	public ResponseEntityDto getContactMetricsById(Long id) {
+		log.info("getContactMetricsById: execution started");
+
+		CrmContactMetrics metrics = crmContactDao.getContactMetricsById(id)
+			.orElseThrow(() -> new ModuleException(CrmMessageConstant.CRM_ERROR_CONTACT_NOT_FOUND));
+
+		log.info("getContactMetricsById: execution ended");
+		return new ResponseEntityDto(false, metrics);
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

This PR promotes `CrmContactMetrics` from an internal Criteria projection to the top-level response body of a public endpoint, but the class is not a DTO by naming or package.

`CrmContactMetrics` lives in `crmplanner/type/` — the package the module uses for enums and internal projection carriers (`CrmDealSummary`, `CrmTaskSummary`, `CrmContactDealMetrics`, `CrmTaskFilterParams`) — and is not named `*Dto`/`*ResponseDto`. Until now it only ever appeared nested inside `CrmContactListItemDtoV2`; no other v2 controller returns a `type/` class as the payload. The guidelines require controller responses to be DTOs and classes to follow `*Dto`/`*ResponseDto` naming with placement in `payload/response`.

It is also write-only for Jackson: `@Getter @AllArgsConstructor` with four `final` fields, no no-arg constructor and no `@Setter`, so any client or contract test generated from this schema cannot round-trip it, and it carries no `@Schema` documentation for the now-public fields.

Fix: move/rename to `payload/response/v2` as e.g. `CrmContactMetricsResponseDtoV2` (updating the single nested usage in `CrmContactListItemDtoV2` and the two `cb.construct` call sites) rather than adding a parallel class, so the wire contract lives in one DTO.

backend/src/main/java/com/skapp/community/crmplanner/controller/v2/CrmContactControllerV2.java
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

A non-numeric `{id}` on this new public path returns 500, not 400.

`@PathVariable Long id` binding failure raises `MethodArgumentTypeMismatchException`, and `GlobalExceptionHandler` (`common/exception/GlobalExceptionHandler.java`) is a plain `@ControllerAdvice` that does not extend `ResponseEntityExceptionHandler` and has no handler for it, so it falls through to `@ExceptionHandler(Exception.class)` at line 178 → `HttpStatus.INTERNAL_SERVER_ERROR` with `COMMON_ERROR_MODULE_EXCEPTION`. So `GET /v2/crm/contact/abc/metrics` yields a 500.

This matters more than usual for this PR because v1's list endpoint was `GET /v1/crm/contact/metrics`; a client migrating to v2 that keeps calling `.../contact/metrics` will bind `"metrics"` to the existing `@GetMapping("/{id}")` and get a 500 rather than a 4xx.

Fix: add `@ExceptionHandler(MethodArgumentTypeMismatchException.class)` to `GlobalExceptionHandler` returning `HttpStatus.BAD_REQUEST` in the standard `ErrorResponse` shape. Pre-existing for `/{id}`, but this PR adds a second path with the same exposure.

backend/src/main/java/com/skapp/community/crmplanner/repository/CrmContactRepository.java
```diff
@@ -4,17 +4,21 @@
...
 import java.util.List;
+import java.util.Optional;
 
 public interface CrmContactRepository {
 
 	Page<CrmContact> findContacts(CrmContactMetricRequestDto filterDto, Pageable pageable);
 
 	Page<CrmContactListItemDtoV2> getContactMetricsV2(CrmContactMetricRequestDto filterDto, Pageable pageable);
 
+	Optional<CrmContactMetrics> getContactMetricsById(Long contactId);
```
ThinuwanW, line 20: 🤖 **Claude** · 🟣 **Nit**

Repository-fragment method name diverges from its two closest analogues and reads ambiguously.

Every other read on this fragment uses the `find*` prefix (`findContacts`, `findContactsForLookup`, `findAllContactsForBoardInit`, `findByIdWithAssociations`), and the two existing per-contact metric queries this one sits alongside are named `findDealMetricsByContactId` and `findTaskMetricsByContactId` (on `CrmDealRepository`/`CrmTaskRepository`). `get*` is the service/controller verb in this codebase; the only `get*` on a fragment is the older `getContactMetricsV2`. `getContactMetricsById` also parses ambiguously — it returns metrics selected by contact id, not a metrics entity by its own id.

Fix: rename the fragment method (and the impl) to `findMetricsByContactId` or `findContactMetricsByContactId`, keeping the service-layer method as `getContactMetricsById`. This also removes any risk of Spring Data attempting derived-query parsing on the name should the fragment implementation ever be dropped.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
anusham-rc: Intentional — we're keeping the metric queries self-contained rather than extracting a shared builder. Each is an independent criteria query, so they stay independent by design.

backend/src/main/java/com/skapp/community/crmplanner/controller/v2/CrmContactControllerV2.java
anusham-rc: Keeping the current shape. The by-id endpoint returns the metrics object directly, which is the natural shape for a single-resource read; the list endpoint wraps items because it paginates. `CrmContactMetrics` living in `type/` follows the existing precedent (`CrmCompanyMetrics` is placed and returned the same way).

backend/src/main/java/com/skapp/community/crmplanner/controller/v2/CrmContactControllerV2.java
anusham-rc: Intentional — v2 narrows the contract to the metrics this endpoint needs, and `closedDealValue` is serialized as a String to match `CrmContactMetrics` (consistent with the company metrics endpoint). We're not aligning the names/types with the v1 detail metrics here.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmContactServiceImplV2.java
anusham-rc: `CrmContactMetrics` follows the existing module convention — it sits in `type/` alongside the other projection carriers, same as `CrmCompanyMetrics`, which is also returned directly. Not introducing a parallel `*ResponseDto` for this single type.

backend/src/main/java/com/skapp/community/crmplanner/controller/v2/CrmContactControllerV2.java
anusham-rc: Out of scope for this PR. A non-numeric `@PathVariable Long` raises `MethodArgumentTypeMismatchException`, which is handled globally in `GlobalExceptionHandler` and applies to every `Long` path variable across the app, not just this endpoint. Better addressed centrally than by constraining this one route.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
anusham-rc: Fixed. `getContactMetricsById` now counts open/overdue tasks linked to the contact directly or through one of the contact's deals — the same rule as `findTaskMetricsByContactId`, which backs `GET /v1/crm/contact/{id}` — so the two endpoints report identical task counts. Kept it a single query. Added a test for deal-linked tasks and one asserting parity between the two endpoints.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmContactControllerV2IntegrationTest.java
anusham-rc: Added all three:
- Not-found and soft-deleted tests now assert the message is `CRM_ERROR_CONTACT_NOT_FOUND`, not just the 400 status.
- A soft-deleted contact test confirms `isDeleted` filtering returns 400.
- A correlation test seeds a second contact with its own WON deal and tasks and asserts the target contact's metrics are unaffected (closed value/count and task counts exclude the other contact).

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
anusham-rc: Keeping the existing cutoff (`LocalDate.now().atStartOfDay()`). It matches `findTaskMetricsByContactId`, which backs the contact detail endpoint, so the two contact endpoints report overdue counts consistently. Aligning contact and company metrics onto a single cutoff would be a separate, module-wide change.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
anusham-rc: Moved `getContactMetricsById` up next to `getContactMetricsV2` so the two contact-metrics queries sit together and the public method is no longer between private helpers. Pure reorder, no logic change.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmContactControllerV2IntegrationTest.java
anusham-rc: Done. Added a `METRICS_BY_ID_PATH` constant and a `performGetMetricsByIdRequest(id)` helper using the URI-template form, switched the new by-id assertions to `RESULTS_0_PATH`, and dropped the explicit token where the single-arg `performRequest` already defaults to it (only the forbidden-role test keeps the two-arg form).

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmContactControllerV2IntegrationTest.java
anusham-rc: The `getContactMetrics_WithDealsAndTasks` test this refers to lived in the v2 controller test; the endpoint moved to v1 and that list test isn't in `CrmContactControllerIntegrationTest`, so the fixture now appears only in the by-id happy-path test - no duplication to extract.

backend/src/main/java/com/skapp/community/crmplanner/repository/CrmContactRepository.java
anusham-rc: Keeping `getContactMetricsById` for consistency with its sibling `getContactMetricsV2` on the same fragment, which already uses the `get*` name. Renaming just this one would make the two metrics methods inconsistent with each other.

akilarootcode (dismissed):

akilarootcode (approved):

PR #1745, opened 2026-08-11
fix: add error toast messages for deal update failures in kanban

dulsara-manakal:
### Summary

**Fix: Show error toast when CRM Kanban deal drag fails**

### What
When dragging a deal card to a new stage/position in the CRM Deals Kanban view fails due to a backend error, the card now shows an error toast ("Failed to update the deal — Oops! Something went wrong. Please try again.") instead of silently reverting with no feedback.

### Why
The drag `onError`/rollback handler reset the card's position but never notified the user, unlike every other mutation-error handler in the app (e.g. delete deal, leave policy toggle), which all show a toast via `useToast()`. This was just a missing call — same pattern as everywhere else, added here.

### Scope
This PR touches **only** the two files needed for this exact fix — nothing else was changed, refactored, or cleaned up:

- `skapp/frontend/src/community/crm/hooks/useKanbanDrag.ts` — added the missing `setToastMessage` call inside the drag rollback handler.
- `skapp/frontend/src/community/common/assets/languages/english/crmModule.json` — added the two new toast i18n strings (`kanban.toastMessages.moveErrorTitle` / `moveErrorDescription`).

No other components, hooks, API files, or unrelated code paths were touched.

### Testing
- `tsc --noEmit` and `eslint` pass clean on both changed files.
- Dev server compiles with no errors.
- Full manual repro (forcing a real backend 500 mid-drag) wasn't run in this environment since it needs a live backend + auth + seeded deals — will do a quick manual check in dev before merge.

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR correctly adds the missing user-facing error feedback for kanban drag failures — the i18n key nesting (`crmModule.deals.kanban.toastMessages.*`) matches the JSON structure, `ToastType.ERROR` and the `setToastMessage` shape are consistent with the rest of the CRM module, and the toast is rendered by the global `ContentWithDrawer`/`ContentWithoutDrawer` layout so it will actually display on the deals page. English is the only locale bundle, so no translations are missing. Remaining comments concern the shared rollback snapshot and interaction with the global axios offline toast, both of which the new toast makes user-visible.

Found **12** new issue(s): 🔴 2 important, 🟡 5 suggestion(s), 🟣 5 nit(s)

frontend/src/community/crm/hooks/useKanbanDrag.ts
```diff
@@ -53,6 +59,13 @@ export const useKanbanDrag = (): UseKanbanDragReturn => {
     if (dragStartSnapshotRef.current) {
```
ThinuwanW, line 59: 🤖 **Claude** · 🟡 **Suggestion**

`rollback` reads `dragStartSnapshotRef.current` at error time, but that ref is overwritten by every new `handleDragStart`. With two overlapping drags (drag A's PATCH still in flight when drag B starts), a failure of A restores the snapshot taken at B's drag start — which still contains A's failed optimistic move and discards B's move. The board then shows a state that matches neither the server nor the user's actions, and now the new toast tells them something failed while the wrong card moved back. Fix by binding the snapshot to the mutation rather than to a shared ref: turn rollback into a factory (`const rollbackTo = (snapshot: CrmBoardStageDealsType[]) => () => { setBoardStageDeals(snapshot); showErrorToast(); }`) and pass it per call via React Query's per-mutate options — `reorderDealWithinStage(payload, { onError: rollbackTo(snapshot) })` / `moveDealToStage(payload, { onError: rollbackTo(snapshot) })` — since `snapshot` is already a local const in `handleDragEnd`. Alternatively use `onMutate` to return the snapshot as mutation context and read it from `onError`'s third argument.

frontend/src/community/crm/hooks/useKanbanDrag.ts
```diff
@@ -53,6 +59,13 @@ export const useKanbanDrag = (): UseKanbanDragReturn => {
...
       setBoardStageDeals(dragStartSnapshotRef.current);
     }
+    setToastMessage({
```
ThinuwanW, line 62: 🤖 **Claude** · 🟡 **Suggestion**

Toast state is a single global value, so this call can clobber a more informative message. When the request fails because the connection dropped mid-flight, the response interceptor in `TanStackProvider.tsx` already sets an error toast ("Oops! Something went wrong." / "No internet connection. Please check your network and try again.") before rejecting; React Query's `onError` then runs `rollback` and overwrites it with the generic "Failed to update the deal" text, so the user loses the actionable network detail. Consider accepting the `AxiosError` argument that `useReorderDealWithinStage`/`useMoveDealBetweenStages` already type on `onError` (`const rollback = (error: AxiosError): void => {...}`) and skipping the toast when `!navigator.onLine` or when there is no `error.response`, while still performing the state rollback.

frontend/src/community/common/assets/languages/english/crmModule.json
```diff
@@ -487,7 +487,11 @@
       "switchDealViewAriaLabel": "Switch deal view"
     },
     "kanban": {
-      "addDealBtn": "Add deal"
+      "addDealBtn": "Add deal",
+      "toastMessages": {
+        "moveErrorTitle": "Failed to update the deal",
+        "moveErrorDescription": "Oops! Something went wrong. Please try again."
```

Copy is inconsistent with the sibling CRM error toasts. `deals.deleteDealModal.toastMessages.errorDescription` uses "Something went wrong. Please try again." and `addDeal...errorDescription` uses "Please try again or contact support" — neither has the "Oops!" prefix. Suggest dropping "Oops! " so the description reads "Something went wrong. Please try again." for consistency across the module.

frontend/src/community/crm/hooks/useKanbanDrag.ts
```diff
@@ -53,6 +59,13 @@ export const useKanbanDrag = (): UseKanbanDragReturn => {
     if (dragStartSnapshotRef.current) {
       setBoardStageDeals(dragStartSnapshotRef.current);
```
ThinuwanW, line 60: 🤖 **Claude** · 🔴 **Important**

The rollback restores the entire board snapshot taken at drag start, so it also discards every unrelated change made to `boardStageDeals` while the PATCH was in flight. All of these write to the same store slice: `addDealToStage` (deal created from the side panel), `removeDealFromStage` (DeleteDealModal), `updateDealInStage` (inline edit / DealSidePanel save), `appendBoardStageDeals` (load-more in any lane), and `replaceBoardStageDeals` (refreshStages). Failure scenario: user drags a deal, and while the request is pending deletes another deal in a different lane (which succeeds server-side) — the drag request then fails, the snapshot is restored, and the deleted deal reappears on the board with a stale `totalCount`, now accompanied by a toast that only mentions the move. Because this PR surfaces the failure to the user, the resulting inconsistent board is far more likely to be noticed and reported. Fix: stop restoring the global snapshot; scope recovery to the two affected stages, ideally by reusing the existing `useRefreshStageDeals().refreshStages([sourceStageId, targetStageId])` hook (already the established recovery path in DealSidePanel.handleSuccess) instead of writing a whole-map snapshot back. That requires passing the source/target stage ids into the error handler (e.g. build the handler per-mutate, or store them alongside the snapshot in the ref).

frontend/src/community/crm/hooks/useKanbanDrag.ts
```diff
@@ -53,6 +59,13 @@ export const useKanbanDrag = (): UseKanbanDragReturn => {
...
+    setToastMessage({
+      open: true,
+      toastType: ToastType.ERROR,
+      title: translateText(["toastMessages", "moveErrorTitle"]),
+      description: translateText(["toastMessages", "moveErrorDescription"]),
+      isIcon: true
+    });
   };
 
   const { mutate: reorderDealWithinStage } =
```
ThinuwanW, line 70: 🤖 **Claude** · 🔴 **Important**

Not every failed move reaches this `onError`. The response interceptor in ToastProvider's sibling `TanStackProvider` returns `undefined` (bare `return;`) for `error?.response?.status === 401` instead of rejecting, so `authFetch.patch` resolves, `mutationFn` resolves, and React Query treats the mutation as a success. Failure scenario: a 401 that is not one of the token-expired/invalid messageKeys (e.g. permission revoked for the deal) — the deal stays visually in the new stage, no rollback runs, and no toast is shown, so the user believes the move was persisted. Since this PR's goal is guaranteed user feedback on move failure, either make the interceptor reject for 401s it does not handle via signOut/refresh, or verify the response in an `onSuccess` before treating the move as persisted.

frontend/src/community/crm/hooks/useKanbanDrag.ts
```diff
@@ -36,6 +39,9 @@ interface UseKanbanDragReturn {
 }
 
 export const useKanbanDrag = (): UseKanbanDragReturn => {
+  const translateText = useTranslator("crmModule", "deals", "kanban");
+  const { setToastMessage } = useToast();
```
ThinuwanW, line 43: 🤖 **Claude** · 🟡 **Suggestion**

Calling `useToast()` here subscribes `useKanbanDrag` — and therefore `DealsKanbanBoard` — to ToastContext. ToastProvider recreates its context value on every toast state change, so from now on the whole kanban tree (every DealStageLane and every DealCard, none of which are wrapped in `memo`) re-renders whenever any toast anywhere in the app opens or auto-hides. Previously the board was insulated because the only consumer above it was ContentWithDrawer, whose `children` element reference is unchanged. On a board with several lanes of deals this is a noticeable regression, and it also fires during drag interactions. Fix options: expose the setter through its own stable context (or a `useRef`-held setter) so consumers that only dispatch toasts don't re-render on toast state, and/or wrap DealStageLane/DealCard in `memo`.

frontend/src/community/crm/hooks/useKanbanDrag.ts
```diff
@@ -53,6 +59,13 @@ export const useKanbanDrag = (): UseKanbanDragReturn => {
...
       setBoardStageDeals(dragStartSnapshotRef.current);
     }
+    setToastMessage({
+      open: true,
+      toastType: ToastType.ERROR,
+      title: translateText(["toastMessages", "moveErrorTitle"]),
+      description: translateText(["toastMessages", "moveErrorDescription"]),
+      isIcon: true
+    });
   };
```
ThinuwanW, line 68: 🤖 **Claude** · 🟡 **Suggestion**

Neither mutation invalidates the React Query cache, so client and server state can diverge silently in both directions. `useGetDealsGroupedByStages` is registered with no `staleTime`, so its data is always stale and any window refocus refetch re-runs the `useEffect` in useBoardData and overwrites `boardStageDeals` from the cached/refetched response. After this rollback the cache is never refreshed, so the board can keep showing client-side state that contradicts the server (including the case where the backend partially applied the reorder before erroring). Add reconciliation in `onSettled` — `queryClient.invalidateQueries({ queryKey: crmBoardQueryKeys.GET_DEALS_GROUPED_BY_STAGES(...) })`, or the narrower `refreshStages([sourceStageId, targetStageId])` — so both the success and failure paths end with server truth rather than optimistic/reverted local state.

frontend/src/community/common/assets/languages/english/crmModule.json
```diff
@@ -487,7 +487,11 @@
       "switchDealViewAriaLabel": "Switch deal view"
     },
     "kanban": {
-      "addDealBtn": "Add deal"
+      "addDealBtn": "Add deal",
+      "toastMessages": {
+        "moveErrorTitle": "Failed to update the deal",
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

This new string pair duplicates an existing one for the same user-facing failure. `deals.sidePanel.toastMessages.editErrorTitle` / `editErrorDescription` already reads "Failed to update deal" / "Something went wrong. Please try again" and is shown when a deal update fails from the side panel — including a stage change, which is exactly what a kanban drag does. The new keys say "Failed to update the deal" and "Oops! Something went wrong. Please try again.", so the same operation gets two slightly different messages depending on where the user performed it. Either reuse the existing keys from the hook, or make the wording byte-identical to the sidePanel pair.

frontend/src/community/common/assets/languages/english/crmModule.json
```diff
@@ -487,7 +487,11 @@
       "switchDealViewAriaLabel": "Switch deal view"
     },
     "kanban": {
-      "addDealBtn": "Add deal"
+      "addDealBtn": "Add deal",
+      "toastMessages": {
```

Key naming: `moveErrorTitle` / `moveErrorDescription` implies a stage change, but the same `rollback` is wired as `onError` for both `useMoveDealBetweenStages` and `useReorderDealWithinStage`, so half the time this toast fires for a pure within-stage reorder where nothing moved between stages. Rename to `updateErrorTitle`/`updateErrorDescription` (matches the sidePanel `editError*` intent) or to the plain `errorTitle`/`errorDescription` used by the sibling `deleteDealModal` and `inlineAddDeal` toastMessages blocks. If distinct copy is wanted for reorder vs move, add two pairs and pass a distinct handler to each mutation.

frontend/src/community/crm/hooks/useKanbanDrag.ts
ThinuwanW, line 62: 🤖 **Claude** · 🟣 **Nit**

React Query invokes the observer's `onError` even if the board has already unmounted, so if the user drags a deal and immediately navigates away, this global `setToastMessage` surfaces "Failed to update the deal" on an unrelated page while the rollback writes to a store no longer rendered (and which will be re-hydrated from the query cache on the next board visit anyway). Consider gating both the restore and the toast on a mounted ref, or scoping the message so it's meaningful out of context.

frontend/src/community/crm/hooks/useKanbanDrag.ts
```diff
@@ -53,6 +59,13 @@ export const useKanbanDrag = (): UseKanbanDragReturn => {
     if (dragStartSnapshotRef.current) {
       setBoardStageDeals(dragStartSnapshotRef.current);
     }
+    setToastMessage({
+      open: true,
+      toastType: ToastType.ERROR,
+      title: translateText(["toastMessages", "moveErrorTitle"]),
+      description: translateText(["toastMessages", "moveErrorDescription"]),
+      isIcon: true
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`isIcon: true` is redundant — ToastMessage already defaults `isIcon` to `true`, and no other toast call site in the CRM module sets it (AddCompanyModalContent, AddTaskModalContent, DeleteDealModal, DealSidePanel all omit it). Drop it to stay consistent with the module's toast payload shape.

frontend/src/community/crm/hooks/useKanbanDrag.ts
```diff
@@ -5,6 +5,9 @@ import type {
 } from "@dnd-kit/core";
 import { useRef, useState } from "react";
 
+import { ToastType } from "~community/common/enums/ComponentEnums";
```
ThinuwanW, line 8: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

No test covers the new failure path. The module already tests the pure helpers (`kanbanUtil.test.ts`), but the rollback + toast behaviour — arguably the most failure-prone part of the drag flow — has none, and this PR makes the hook harder to test by adding two provider dependencies. Consider a `useKanbanDrag` test with `renderHook`, a mocked `useToast`, and a rejecting mutation, asserting both that the board state is restored and that an error toast is dispatched.

frontend/src/community/common/assets/languages/english/crmModule.json
dulsara-manakal: Renamed

frontend/src/community/common/assets/languages/english/crmModule.json
dulsara-manakal: Thats the requierment.

frontend/src/community/crm/hooks/useKanbanDrag.ts
dulsara-manakal: This is the behavior so  no need to change and this condition is rare.

frontend/src/community/crm/hooks/useKanbanDrag.ts
dulsara-manakal: This is out of scope and the context is different than describing in this  comment.

frontend/src/community/crm/hooks/useKanbanDrag.ts
dulsara-manakal: This mean not signed in so it actuly should handle the way it handle currently and this is correct.

frontend/src/community/crm/hooks/useKanbanDrag.ts
dulsara-manakal: No need test

frontend/src/community/crm/hooks/useKanbanDrag.ts
dulsara-manakal: ToastProvider  canot be change

frontend/src/community/common/assets/languages/english/crmModule.json
dulsara-manakal: Its different from what we newly added.

frontend/src/community/crm/hooks/useKanbanDrag.ts
dulsara-manakal: Removed

frontend/src/community/crm/hooks/useKanbanDrag.ts
dulsara-manakal: This is normal behavior and the error message should display anyway.

frontend/src/community/crm/hooks/useKanbanDrag.ts
dulsara-manakal: we are not handle it by invalidating.  we handle it through store.

frontend/src/community/crm/hooks/useKanbanDrag.ts
dulsara-manakal: According to requierment no need to add   show message specificaly for each scenario.

thusala (dismissed):

frontend/src/community/common/assets/languages/english/crmModule.json
dulsara-manakal: Make it consistent with other places. added "Oops!"

frontend/src/community/common/assets/languages/english/crmModule.json
dulsara-manakal: Make it consistent with other places. added "Oops!"

thusala (approved):

akilarootcode (approved):

akilarootcode (approved):

PR #1748, opened 2026-08-12
[Fix] Restore the signup page broken by the tenantID migration

akilarootcode:
## Summary

PR #773 deleted `getTenantID` and narrowed `getSubdomain` to a single argument, but left two helpers in the enterprise `commonUtil` still calling the old API. Because `next.config.mjs` sets `ignoreBuildErrors`, those broken references shipped as a green build and threw `ReferenceError: getTenantID is not defined` on the first render of the sign-up page — taking sign-up down on **every** host, not only tenant subdomains.

This moves both helpers into `tenantUtil` and rebuilds them on the existing `getHostSubdomain()`, preserving the pre-#773 behaviour exactly, and clears one further dangling import of the same shape.

- Added `isNotOnAppSubdomain` and `getAppSubdomainUrl` stubs to `frontend/src/fallback/common/utils/tenantUtil.ts`, so the community fallback keeps mirroring the enterprise module's export surface as the two helpers move into it.
## Feature scope (all repos)

This fix spans multiple repos on branch `fix/signup-failure`:

| Repo | What changed |
|------|--------------|
| `SkappHQ/skapp` (super) | Fallback stubs so the community `tenantUtil` keeps mirroring the enterprise module's exports |
| `rootcodelabs/skapp-ep-fe-src` | Moved the two broken helpers into `tenantUtil`; removed an unreachable branch with the same dangling-import defect |
| `rootcodelabs/skapp-ep-fe-pages` | Repointed the `signup.tsx` imports at `tenantUtil` |

## Architecture / flow

```mermaid
flowchart TD
    A["User opens /signup"] --> B{"isNotOnAppSubdomain()"}
    B -- "on a tenant host" --> C["getAppSubdomainUrl(/signup + query)"]
    C --> D["window.location.replace to the app host"]
    D --> E["Sign-up form renders"]
    B -- "already on the app host" --> E

    subgraph Broken ["Before — broken on every host"]
        X["isNotOnAppSubdomain() in commonUtil"] --> Y["calls getTenantID(), deleted by #773"]
        Y --> Z["ReferenceError caught by ErrorBoundary<br/>sign-up page replaced by an error screen"]
    end
```

The helpers move module, not behaviour: `getHostSubdomain()` returns the first
host label, exactly what the deleted `getTenantID()` returned, so the resolved
redirect target is byte-for-byte what it was before #773.

- **Host-derivation hardening.** `getAppSubdomainUrl` strips the first label
  unconditionally, so a two-label host yields `app.<tld>` and `localhost:3000`
  yields `app.`, and the caller navigates without a trusted-origin check. These
  are pre-existing weaknesses carried over unchanged; fixing them here would have
  made this PR a behaviour change rather than a crash fix. Raised separately.
- **`SameSite=Strict` → `Lax` cookie change in `rootcodelabs/skapp-pm`** — an
  unrelated local edit, kept out and raised on its own.
- **Pre-existing lint findings in the files touched** — the unused `sendEvent` in
  `signup.tsx` and the unused `React` / `tempData` in `InvoicePopupController`.
  All three predate this branch (baselined against HEAD) and are left for a
  dedicated clean-up.

- **Submodule hash / pointer bumps** are intentionally excluded. Gitlink updates
  are done at **deployment time**, not in feature PRs.
- Untracked JVM crash logs (`hs_err_pid*.log`, `replay_pid*.log`) left in the
  workspace by an unrelated local run.

## Testing

| Check | Ran? | Result |
|-------|------|--------|
| Unit tests | n/a | No jest spec covers the sign-up flow |
| Formatter (BE only) | n/a | FE-only change |
| tsc + lint (BE only) | n/a | FE-only change — see note below |
| e2e (Playwright) | yes | **pass** — 3/3 (`auth` + `smoke`) |

`tsc` was diffed anyway, because this bug is precisely what a green build hides:
`next.config.mjs` sets `ignoreBuildErrors`, so the broken references compiled and
shipped. Baselined by `file(line,col): TScode` before and after — **665 → 661
errors, no new ones**. The four cleared are exactly this defect:

```
commonUtil.ts(52,20)             TS2552  Cannot find name 'getTenantID'
commonUtil.ts(60,21)             TS2352  bad cast on getSubdomain
commonUtil.ts(60,40)             TS2554  getSubdomain expects 1 argument, got 2
InvoicePopupController.tsx(7,10) TS2305  no exported member 'tempShouldUseCustomDashboard'
```

The e2e guard was verified non-vacuous: reverted to HEAD, `signup-page.spec.ts`
fails (times out waiting for the form, page sitting in the ErrorBoundary);
restored, it passes. Prettier clean; eslint reports only the pre-existing
findings listed under Exclusions.

## Related PRs

Same branch (`fix/signup-failure`), same fix, split across the repos it touches:

- SkappHQ/skapp#1748 — community fallback stubs for the relocated helpers
- rootcodelabs/skapp-ep-fe-src#778 — the crash fix; helpers moved into `tenantUtil`, plus the unreachable invoice branch
- rootcodelabs/skapp-ep-fe-pages#349 — `signup.tsx` import repoint

Merge order does not matter: each repo compiles independently, since the pages
change only repoints an import that both the old and new module resolve.

Raised alongside these but a **separate concern on its own branch**, not part of
this fix:

- rootcodelabs/skapp-pm#1217 — access-token cookie `SameSite` alignment (`fix/pm-access-token-samesite`)

### DIFF
--- a/frontend/src/fallback/common/utils/tenantUtil.ts
+++ b/frontend/src/fallback/common/utils/tenantUtil.ts
@@ -10,6 +10,10 @@ export const isTenantSelectionHost = (): boolean => false;
 
 export const getTenantId = (): string => "";
 
+export const isNotOnAppSubdomain = (): boolean => false;
+
+export const getAppSubdomainUrl = (): string => "";
+
 export const getTenantHostUrl = (): string => "";
 
 export const getTenantQueryPath = (): string => "";


hasalarootcode (approved):

ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds two stub exports (`isNotOnAppSubdomain`, `getAppSubdomainUrl`) to the community fallback for `~enterprise/common/utils/tenantUtil`, restoring name parity with the enterprise implementation so community-only checkouts and Jest (which always maps `~enterprise` → `src/fallback`) can resolve the imports added in `pages/enterprise/signup.tsx`. The export set now matches the enterprise module exactly and the stub ordering mirrors it, and `isNotOnAppSubdomain() === false` is the semantically correct no-op for community (no tenant subdomains, so no redirect). The only concerns are signature/return-value parity with the real implementation: `getAppSubdomainUrl` drops the `path: string` parameter its enterprise counterpart takes and returns `""` rather than the path.

Found **10** new issue(s): 🔴 2 important, 🟡 4 suggestion(s), 🟣 4 nit(s)

frontend/src/fallback/common/utils/tenantUtil.ts
```diff
@@ -10,6 +10,10 @@ export const isTenantSelectionHost = (): boolean => false;
 
 export const getTenantId = (): string => "";
 
+export const isNotOnAppSubdomain = (): boolean => false;
+
+export const getAppSubdomainUrl = (): string => "";
```
ThinuwanW, line 15: 🤖 **Claude** · 🟡 **Suggestion**

Signature parity: the enterprise implementation is `getAppSubdomainUrl(path: string): string`, but this stub is typed `(): string`. The only caller (`pages/enterprise/signup.tsx`) invokes it with one argument, so wherever TypeScript resolves `~enterprise` to `src/fallback` — Jest's `moduleNameMapper` maps `^~enterprise(.*)$` to `src/fallback` unconditionally, and tsconfig falls through to `./src/fallback/*` when the `src/enterprise` submodule is not checked out — this becomes `TS2554: Expected 0 arguments, but got 1`. It is currently masked only because `next.config.mjs` sets `typescript.ignoreBuildErrors: true` and because `pages/enterprise/` is a submodule that is absent from community checkouts, so it silently rots. Every other stub in `src/fallback` keeps the real parameter list (e.g. `commonUtil.shouldUseDefaultTheme(asPath: string)`, `setDeviceTokenApi.setDeviceToken(fcmToken: string)`, `awsS3ServiceFunctions.getS3FoldersByUrl(filePath: string)`); this file's pre-existing stubs (`normalizeTenantId`, `readTenantIdFromQuery`, `getTenantHostUrl`, `getTenantQueryPath`) are the outliers. Suggest declaring the stub as accepting an underscore-prefixed unused param so it mirrors the real signature and stays lint-clean: `export const getAppSubdomainUrl = (_path: string): string => "";` — and consider doing the same for the four pre-existing stubs while you are in this file.

ThinuwanW, line 15: 🤖 **Claude** · 🟣 **Nit**

Return-value parity: the enterprise implementation returns the untouched `path` on its non-browser branch (`if (typeof window === "undefined") return path;`), whereas this stub returns `""`. The sole consumer does `window.location.replace(getAppSubdomainUrl(...))`, and `window.location.replace("")` re-navigates to the current URL — i.e. a reload loop rather than a no-op. Today that path is unreachable because the paired `isNotOnAppSubdomain()` stub is hardcoded `false`, so this is defence-in-depth only, but returning the passed path (`(path: string): string => path;`) makes the stub inherently safe if the two stubs ever drift or a future community caller uses `getAppSubdomainUrl` without the guard.

frontend/src/fallback/common/utils/tenantUtil.ts
```diff
@@ -10,6 +10,10 @@ export const isTenantSelectionHost = (): boolean => false;
 
 export const getTenantId = (): string => "";
 
+export const isNotOnAppSubdomain = (): boolean => false;
```
ThinuwanW, line 13: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

No `tenantUtil.test.ts` exists alongside this fallback, even though sibling fallbacks in the same directory are covered (`awsS3ServiceFunctions.test.ts`, `firebase.test.ts`, `monitoring.test.ts`, `getEnterpriseDrawerRoutes.test.ts`) and those tests do exactly the useful thing here — assert the stub's return value while calling it with the real argument list, which is what would have caught the arity mismatch above. Pre-existing gap rather than something this PR introduced, but adding a small `tenantUtil.test.ts` asserting `isNotOnAppSubdomain() === false` and the `getAppSubdomainUrl` contract would lock in the community no-op behaviour.

frontend/src/fallback/common/utils/tenantUtil.ts
ThinuwanW, line 15: 🤖 **Claude** · 🔴 **Important**

The enterprise counterpart this stub shadows (`src/enterprise/common/utils/tenantUtil.ts`, `getAppSubdomainUrl`) derives the root host with `host.split(".").slice(1).join(".")`, which is only correct for hosts of the form `<sub>.<root>`. Two concrete breakages: (1) On `localhost:3000`, `getSubdomain("localhost")` returns `"localhost"` (it is just `url.split(".")[0]`, with no localhost special-case — the `LOCALHOST` constant in `src/community/common/constants/stringConstants.ts` is declared but never used anywhere), so `isNotOnAppSubdomain()` is `true`, the redirect fires, and `rootHost` computes to the empty string, producing `http://app./signup` — enterprise-mode `next dev` on /signup is unusable. (2) On a two-label apex host such as `skapp.com`, `getSubdomain` returns `"skapp"`, so `isNotOnAppSubdomain()` is again `true` and the user is navigated to `https://app.com/signup` — an unrelated third-party registrable domain. Adding this always-`""` community stub means neither case can ever surface in a community build or a Jest run, so the defect is invisible outside enterprise production/dev. Fix in the enterprise util: treat a host with fewer than three labels (and `localhost`) as having no subdomain so `isNotOnAppSubdomain()` returns `false`, and have `getAppSubdomainUrl` return `path` unchanged whenever a root host cannot be derived, rather than emitting `app.` + empty string.

frontend/src/fallback/common/utils/tenantUtil.ts
ThinuwanW, line 13: 🤖 **Claude** · 🔴 **Important**

The sole consumer of this helper, `pages/enterprise/signup.tsx:43`, uses it as a lazy state initialiser: `useState(isNotOnAppSubdomain)`. That runs during render. On the server `getHostSubdomain()` short-circuits to `""` because `window` is undefined, so the pre-rendered HTML is built with `isLoading === false` and contains the full sign-up form; on the client's first (hydration) render on any non-`app` subdomain it returns `true` and the component returns `<FullScreenLoader />` instead. That is a React 18 hydration mismatch — the server markup is thrown away and a client-side re-render is forced, with a console error in dev. In community mode this stub returns `false` on both sides so the mismatch never reproduces during local/community testing. Fix: initialise `useState(false)` and call `setIsLoading(true)` inside the existing `useEffect` alongside the `window.location.replace(...)` call, so the subdomain check only ever runs client-side after mount. While there, note that `isLoading` is never reset if the navigation is blocked, leaving the page stuck on the full-screen loader.

ThinuwanW, line 13: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Signature drift in this file is systemic, not limited to the newly added `getAppSubdomainUrl`. Every stub here is zero-argument while the enterprise implementations take parameters: `normalizeTenantId(value: string)`, `readTenantIdFromQuery(value: string | string[] | undefined)`, `getTenantHostUrl(tenantId: string, path: string)` and `getTenantQueryPath(tenantId: string)`. This file is also the outlier within its own directory — the sibling fallback `src/fallback/common/utils/awsS3ServiceFunctions.ts` faithfully mirrors parameter lists (`uploadFileToS3ByUrl(file, fileCategory, setFileUploadProgress?)`). Nothing catches the drift: `next.config.mjs` sets `typescript.ignoreBuildErrors: true`, and community builds never compile the enterprise callers that pass the arguments. Rather than patching only the one function, add a compile-time contract — declare the module shape once (e.g. `export type TenantUtilModule = { getAppSubdomainUrl: (path: string) => string; ... }` in a community-side types file) and have both `src/enterprise/common/utils/tenantUtil.ts` and this fallback assert against it with `satisfies`, so the next helper added to the enterprise module fails the build until its stub matches.

ThinuwanW, line 13: 🤖 **Claude** · 🟡 **Suggestion**

These two stubs are in a different reachability class from their neighbours, and nothing in the file says so. `isAuthHost`/`isTenantSelectionHost` are imported by `pages/index.tsx` and `getTenantId` by numerous `src/community/**` files, so those stubs genuinely execute in community production builds. `isNotOnAppSubdomain`/`getAppSubdomainUrl` have exactly one consumer — `pages/enterprise/signup.tsx` — which lives in the enterprise submodule and is never compiled in community mode (`next.config.mjs` rewrites `/signup` to `/community/signup` there). They therefore exist solely to satisfy `jest.config.js`'s unconditional `"^~enterprise(.*)$" -> "<rootDir>/src/fallback$1"` mapping. Two consequences worth acting on: add a short comment marking them as test-only shims so a future maintainer does not mistake a load-bearing stub for a dead one; and be aware that because that mapping applies even in enterprise checkouts, any test of the sign-up page gets `isNotOnAppSubdomain() === false` and can never cover the new redirect branch — such a test must `jest.mock("~enterprise/common/utils/tenantUtil")` explicitly.

ThinuwanW, line 13: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

`pages/enterprise/signup.tsx`, modified by this change to import the new helpers, contains two violations of the strict-mode/null-safety checklist in code adjacent to the change. Its `onSuccess` handler is typed `(response: any)` — an explicit `any` in a strict-mode codebase — and it then does `response?.data?.results[0]?.authUrl`: the optional chain stops at `data`, so if the API returns a body without `results` (or with `results: null`), indexing `[0]` throws a TypeError inside the mutation success path, where there is no `onError` to surface a toast and `isLoading` has already been set to `true`, leaving the user on a permanent loader. Type the response against the existing auth-URL response type and make the whole access optional: `response?.data?.results?.[0]?.authUrl`, with a guard that resets `isLoading` and shows the existing error toast when no `authUrl` is present.

frontend/src/fallback/common/utils/tenantUtil.ts
ThinuwanW, line 15: 🤖 **Claude** · 🟣 **Nit**

The enterprise module now has two functions that both answer "what is the root host?" using different and incompatible algorithms: `getAppSubdomainUrl` blindly strips the first label (`host.split(".").slice(1).join(".")`) while `getTenantHostUrl` strips only known selection subdomains via `new RegExp('^(' + TENANT_SELECTION_SUBDOMAINS.join("|") + ')\\.')`. They will disagree on any host whose first label is neither `app` nor `auth`, and every future change has to be made twice — including whatever localhost/short-host guard the fix for the host-parsing defect introduces. Extract a single `getRootHost()` helper used by both, and mirror it here as one stub instead of letting the two derivations drift.

frontend/src/fallback/common/utils/tenantUtil.ts
ThinuwanW, line 13: 🤖 **Claude** · 🟣 **Nit**

Naming: `isNotOnAppSubdomain` is a negated predicate, which diverges from the positive-predicate convention its immediate neighbours follow (`isAuthHost`, `isTenantSelectionHost`) and forces double negatives at any call site that needs the affirmative case. Prefer `isOnAppSubdomain` and invert at the two call sites in `pages/enterprise/signup.tsx`. This has to be done in the same change as the enterprise export so the fallback and the real module stay in sync.

shakila-kamalasena (approved):

frontend/src/fallback/common/utils/tenantUtil.ts
akilarootcode: no need

akilarootcode: no need

frontend/src/fallback/common/utils/tenantUtil.ts
akilarootcode: no need

frontend/src/fallback/common/utils/tenantUtil.ts
akilarootcode: no need

frontend/src/fallback/common/utils/tenantUtil.ts
akilarootcode: no need

akilarootcode: no need

akilarootcode: no need

akilarootcode: no need

frontend/src/fallback/common/utils/tenantUtil.ts
akilarootcode: no need

frontend/src/fallback/common/utils/tenantUtil.ts
akilarootcode: no need

PR #1749, opened 2026-08-12
feat(crm): v2 deal & task response types

anusham-rc:
## CRM v2 — Deal & Task (response-type refactor)

Adds the `/v2/crm/deal` and `/v2/crm/task` endpoints returning the fully-embedded response shapes. **No v1 API is changed.**

> Re-targets the work from #1732 to `develop`. #1732 was merged into `feat/crm-response-types-refactor` by mistake, so these commits never reached `develop`; #1731 (the response-types refactor) is already on `develop`, so the net diff here is only the deal + task v2 work.

### Deal
- `GET /v2/crm/deal`, `GET /{id}`, `POST`, `PATCH /{id}` → `CrmDealResponseDtoV2` with fully embedded `stage`, `owner`, `company`, and `contact` (the contact carries its own `company` + `owner`) — self-contained, no client-side lookup needed.
- Reads preserve the sales-rep owner restriction; writes delegate to v1 `persistNewDeal` / `applyDealEdit` so enterprise deal-limit + stage-access still run.

### Task
- `GET /v2/crm/task` (open), `GET /completed`, `GET /related`, `GET /{id}`, `POST`, `PATCH /{id}` → `CrmTaskResponseDtoV2` with fully embedded `type`, `owner`, `company`, `contact`, and `deal`.
- Open-list keeps the v1 `{ tasks: [...] }` envelope via `CrmTaskListResponseDtoV2`; other reads paginated.
- Reads preserve owner restriction; writes delegate to v1 `persistNewTask` / `applyTaskEdit` so the enterprise task-limit still runs.

### Shared
- `CrmMapperV2` composes the nested v2 shapes (deal/task reuse the v2 contact mapping).
- Board deal endpoints stay on v1 (intentionally unchanged).
- v1 service changes are additive extractions only; existing endpoints behave identically.


akilarootcode (approved):

ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds a V2 REST surface for CRM deals and tasks (controllers, service interfaces/impls, Criteria-API projection queries, V2 response DTOs) and refactors the V1 create/edit service methods into reusable `persistNew*`/`apply*Edit` entity-returning helpers. The refactor is behaviour-preserving — all validations, enterprise limit hooks (`validateDealCreationLimit`/`validateTaskCreationLimit`), owner-restriction checks and ordering are retained, and the `getReferenceById` → `findById` change for task types is a genuine 500→400 fix with an updated test. Security annotations mirror V1 exactly. The main concern is that the new nested `cb.construct` projections emit fully-null nested objects instead of `null`, which makes the list endpoints structurally inconsistent with the mapper-based detail/create/edit endpoints in the same PR.

Found **13** new issue(s): 🔴 3 important, 🟡 6 suggestion(s), 🟣 4 nit(s)

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmTaskRepositoryImpl.java
```diff
@@ -204,6 +217,141 @@ public CrmContactTaskMetrics findTaskMetricsByContactId(Long contactId) {
...
+		Join<CrmDeal, CrmDealStage> dealStage = deal.join(CrmDeal_.stage, JoinType.LEFT);
+		Join<CrmDeal, Employee> dealOwner = deal.join(CrmDeal_.owner, JoinType.LEFT);
+		Join<CrmDeal, CrmCompany> dealCompany = deal.join(CrmDeal_.company, JoinType.LEFT);
+		dealCompany.on(cb.isFalse(dealCompany.get(CrmCompany_.isDeleted)));
+		Join<CrmDeal, CrmContact> dealContact = deal.join(CrmDeal_.contact, JoinType.LEFT);
+		Join<CrmContact, CrmCompany> dealContactCompany = dealContact.join(CrmContact_.company, JoinType.LEFT);
+		dealContactCompany.on(cb.isFalse(dealContactCompany.get(CrmCompany_.isDeleted)));
+		Join<CrmContact, Employee> dealContactOwner = dealContact.join(CrmContact_.owner, JoinType.LEFT);
+
+		return cb.construct(CrmTaskResponseDtoV2.class, task.get(CrmTask_.id), task.get(CrmTask_.name),
```
ThinuwanW, line 313: 🤖 **Claude** · 🔴 **Important**

Nested `cb.construct(...)` never yields `null` for an unmatched LEFT JOIN — Hibernate always invokes the constructor, so every optional association is materialised as a DTO whose fields are all null rather than as `null`. `CrmTask.contact`, `CrmTask.company` and `CrmTask.deal` are all nullable `@ManyToOne` (and `CrmDeal.company` likewise), so `GET /v2/crm/task` will return, for a task with no linked deal, `"deal": {"id":null,"name":null,"stage":{...all null},"owner":{...},"company":{...},"contact":{...}}` instead of `"deal": null`. The same happens in `CrmDealRepositoryImpl.findDealsV2` (line 120) for `company`/`contact`. This also defeats the soft-delete masking intent: `company.on(cb.isFalse(isDeleted))` makes the join miss, but the client still receives a `company` object rather than `null`.

The result is that the same resource is serialised two different ways inside this PR: `GET /v2/crm/task/{id}`, `POST /v2/crm/task` and `PATCH /v2/crm/task/{id}` go through `CrmMapperV2` + `CrmUtil.toTaskResponseDtoV2`, which set the fields to real `null`; the list endpoints do not. A client doing `if (task.deal)` or `if (deal.company)` will behave differently per endpoint. It also inflates list payloads noticeably (each dealless task carries ~6 empty nested objects).

Fix: after building the page/list content, null out nested DTOs whose identifier came back null — e.g. in `CrmTaskServiceImplV2`/`CrmDealServiceImplV2`, or in a small private post-processing helper in the repository impl: `if (dto.getDeal() != null && dto.getDeal().getId() == null) dto.setDeal(null);` and the same for `company`, `contact`, `contact.company`, `deal.company`, `deal.contact`, `deal.contact.company`. Alternatively project the association ids as discriminators and assemble the DTOs in Java.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmDealControllerV2IntegrationTest.java
```diff
@@ -0,0 +1,352 @@
...
+		savedDeal("List Deleted Co Deal V2", stage, company, contact, 1L);
+
+		company.setIsDeleted(true);
+		crmCompanyDao.save(company);
+
+		performGetDealsRequest(company.getId()).andDo(print())
+			.andExpect(status().isOk())
+			.andExpect(jsonPath(RESULTS_0_PATH + "['items'].length()").value(1))
+			.andExpect(jsonPath(RESULTS_0_PATH + "['items'][0]['name']").value("List Deleted Co Deal V2"))
+			.andExpect(jsonPath(RESULTS_0_PATH + "['items'][0]['company']['id']").doesNotExist())
```
ThinuwanW, line 264: 🤖 **Claude** · 🟡 **Suggestion**

The masking assertions use `jsonPath("...['company']['id']").doesNotExist()`. Spring's `doesNotExist()` passes when the JSONPath resolves to `null`, so this assertion is satisfied both when `company` is genuinely absent/null AND when `company` is an object whose `id` field is null — which is exactly what the projection query actually produces (see the `CrmTaskRepositoryImpl` finding). The test therefore cannot distinguish correct masking from the current behaviour. Assert on the container instead: `jsonPath(RESULTS_0_PATH + "['items'][0]['company']").doesNotExist()` (or `.value(nullValue())`). Same applies to `getDealById_SoftDeletedCompany_MasksCompany` and to the task V2 test — and please add a case asserting `deal` is null for a task with no linked deal, which is currently uncovered.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmDealRepositoryImpl.java
```diff
@@ -96,6 +102,61 @@ public Page<CrmDeal> findDeals(CrmDealFilterDto filterDto, Long ownerId, Pageabl
...
+						contact.get(CrmContact_.lastContactAt), contact.get(Auditable_.lastModifiedDate),
+						cb.construct(CrmCompanyResponseDto.class, contactCompany.get(CrmCompany_.id),
+								contactCompany.get(CrmCompany_.name), contactCompany.get(CrmCompany_.industry),
+								contactCompany.get(CrmCompany_.website), contactCompany.get(CrmCompany_.address),
+								contactCompany.get(CrmCompany_.contactNumber)),
+						cb.construct(CrmOwnerResponseDto.class, contactOwner.get(Employee_.employeeId),
+								contactOwner.get(Employee_.firstName), contactOwner.get(Employee_.lastName),
+								contactOwner.get(Employee_.authPic)))));
+
+		query.where(buildPredicates(cb, deal, filterDto, ownerId).toArray(new Predicate[0]));
```
ThinuwanW, line 142: 🤖 **Claude** · 🟡 **Suggestion**

Redundant duplicate joins. `findDealsV2` already creates LEFT joins for `company`, `contact` and `owner` to build the projection, and then `buildPredicates(cb, deal, ...)` unconditionally creates a *second* LEFT join to `CrmDeal_.company` (line 109 of this file) plus second joins to `contact`/`owner` whenever a search keyword is supplied — all on the same root. The emitted SQL therefore contains several duplicate joins. Results are unaffected because every association is `@ManyToOne` (no row multiplication), but it is avoidable query cost on a paginated endpoint. The same pattern occurs in `CrmTaskRepositoryImpl` where `buildTaskProjectionSelection` and `buildTaskPredicates`/`buildRelatedTaskPredicates` each create their own `contact`/`deal`/`company` joins. Consider overloading the predicate builders to accept the already-created joins (as `CrmContactRepositoryImpl.buildPredicates(cb, contact, owner, company, filterDto)` already does).

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmTaskRepositoryImpl.java
```diff
@@ -204,6 +217,141 @@ public CrmContactTaskMetrics findTaskMetricsByContactId(Long contactId) {
 		return entityManager.createQuery(query).getSingleResult();
 	}
 
+	@Override
+	public List<CrmTaskResponseDtoV2> findTasksV2(Long ownerId, CrmTaskFilterDto filterDto) {
```
ThinuwanW, line 221: 🤖 **Claude** · 🟡 **Suggestion**

`findTasksV2` (backing `GET /v2/crm/task`) is unbounded — no `setMaxResults`, and `CrmTaskServiceImplV2.getTasks` returns every matching row in a `CrmTaskListResponseDtoV2`. V1 shared this limitation, but V1's query fetched 4 associations whereas this one joins 13 tables and constructs 7 nested DTOs per row, so the cost per unbounded row is substantially higher. For a CRM admin (ownerId resolves to `null`, i.e. no owner filter) this scans all open non-deleted tasks in the tenant. Consider adding pagination or at least a hard cap, consistent with the other V2 list endpoints.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmDealServiceImplV2.java
```diff
@@ -0,0 +1,102 @@
...
+	public ResponseEntityDto getDeals(CrmDealFilterDto filterDto) {
+		log.info("getDeals: execution started");
+
+		User currentUser = userService.getCurrentUser();
+		Long ownerId = CrmUtil.isCrmSalesRepresentative(currentUser) ? currentUser.getEmployee().getEmployeeId() : null;
+
+		Page<CrmDealResponseDtoV2> dealsPage = crmDealDao.findDealsV2(filterDto, ownerId,
+				PageRequest.of(filterDto.getPage(), filterDto.getSize()));
+
+		PageDto pageDto = new PageDto();
```
ThinuwanW, line 51: 🤖 **Claude** · 🟡 **Suggestion**

`getDeals` inlines the four-line `PageDto` assembly that `CrmTaskServiceImplV2` has already extracted into a private `toPageDto(...)` helper, and both duplicate what `PageTransformer.transform(page)` (already injected and used by `CrmDealServiceImpl.getDeals`) does. Since `PageDto` only carries `items`/`currentPage`/`totalItems`/`totalPages` the output is equivalent, but this is the third copy of the same block. Inject `PageTransformer` here (or move a shared `toPageDto` helper into a common util) so the page envelope is built in exactly one place.

backend/src/main/java/com/skapp/community/crmplanner/util/CrmUtil.java
```diff
@@ -82,4 +86,33 @@ public CrmDealByStageItemResponseDto toDealByStageItemDto(CrmMapper crmMapper, C
...
+
+	public CrmTaskResponseDtoV2 toTaskResponseDtoV2(CrmMapperV2 crmMapperV2, CrmTask task) {
+		CrmTaskResponseDtoV2 dto = crmMapperV2.crmTaskToCrmTaskResponseDtoV2(task);
+		if (isCompanyDeleted(task.getCompany())) {
+			dto.setCompany(null);
+		}
+		if (dto.getContact() != null && hasDeletedCompany(task.getContact())) {
+			dto.getContact().setCompany(null);
+		}
+		if (task.getDeal() != null && dto.getDeal() != null) {
```
ThinuwanW, line 103: 🤖 **Claude** · 🟣 **Nit**

`if (task.getDeal() != null && dto.getDeal() != null)` — the first condition is dead. `dto` comes from `crmMapperV2.crmTaskToCrmTaskResponseDtoV2(task)`, so `dto.getDeal()` can only be non-null when `task.getDeal()` is non-null; the `dto.getDeal() != null` check alone is sufficient and is the one that actually guards the dereference inside `maskDeletedCompaniesOnDeal`. Note the adjacent contact branch on the line above already uses only the `dto.getContact() != null` form, so dropping the redundant half also makes the two branches consistent.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmTaskServiceImpl.java
```diff
@@ -284,8 +299,8 @@ public ResponseEntityDto editTask(Long id, CrmTaskEditRequestDto requestDto) {
 
 		CrmTask updatedTask = crmTaskDao.save(task);
 
-		log.info("editTask: execution ended successfully");
-		return new ResponseEntityDto(false, crmMapper.crmTaskToCrmTaskResponseDto(updatedTask));
+		log.info("applyTaskEdit: execution ended successfully");
```
ThinuwanW, line 302: 🤖 **Claude** · 🔴 **Important**

`CrmTaskResponseDtoV2` newly exposes `lastModifiedDate` (v1's `CrmTaskResponseDto` had no such field), but the value returned by `PATCH /v2/crm/task/{id}` is the pre-edit timestamp. `applyTaskEdit` ends with `crmTaskDao.save(task)`, which for an already-managed entity is a `merge` — it does not flush. `@LastModifiedDate` is populated by `AuditingEntityListener.touchForUpdate()` on the JPA `@PreUpdate` callback, which only fires when Hibernate flushes at commit, i.e. *after* `CrmTaskServiceImplV2.editTask` has already built the response DTO. Net effect: the client gets a stale `lastModifiedDate` even though `GET /v2/crm/task/completed` (which orders by `lastModifiedDate DESC`) immediately reorders that task to the top — the list and the edit response disagree. Same applies to `applyDealEdit`/`applyContactEdit` where the field is surfaced. Fix: use `crmTaskDao.saveAndFlush(task)` (and `crmDealDao.saveAndFlush`) so the `@PreUpdate` callback runs before the entity is mapped, or have the V2 service flush the entity manager before calling `CrmUtil.toTaskResponseDtoV2`.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmDealServiceImplV2.java
```diff
@@ -0,0 +1,102 @@
...
+	@Override
+	@Transactional(readOnly = true)
+	public ResponseEntityDto getDeals(CrmDealFilterDto filterDto) {
+		log.info("getDeals: execution started");
+
+		User currentUser = userService.getCurrentUser();
+		Long ownerId = CrmUtil.isCrmSalesRepresentative(currentUser) ? currentUser.getEmployee().getEmployeeId() : null;
+
+		Page<CrmDealResponseDtoV2> dealsPage = crmDealDao.findDealsV2(filterDto, ownerId,
+				PageRequest.of(filterDto.getPage(), filterDto.getSize()));
```
ThinuwanW, line 49: 🤖 **Claude** · 🔴 **Important**

`PageRequest.of(filterDto.getPage(), filterDto.getSize())` is called on raw, unvalidated request parameters. `page` and `size` are plain `int`s on `CrmDealFilterDto`/`CrmTaskCompletedFilterDto`/`CrmTaskRelatedFilterDto` with no bounds. `?page=-1` or `?size=0` binds successfully (no `BindException`) and then `PageRequest.of` throws `IllegalArgumentException`; `GlobalExceptionHandler` has no `IllegalArgumentException` handler, so it falls through to the generic `@ExceptionHandler(Exception.class)` and the caller gets a 500 instead of a 400. There is also no upper bound, so `?size=1000000` makes the new projection query (which for tasks joins 13 tables) materialise a million wide rows. This affects all four new V2 list endpoints (`GET /v2/crm/deal`, `/v2/crm/task/completed`, `/v2/crm/task/related`). Fix: clamp/validate in one place — e.g. a shared helper that rejects `page < 0` and clamps `size` to `[1, CrmConstants.MAX_PAGE_SIZE]` and throws `ModuleException` with a message constant, or add `@Min`/`@Max` on the filter DTO fields so the failure surfaces as a `BindException` (which the handler already maps to 400). Mirror it in `CrmTaskServiceImplV2.getCompletedTasks`/`getRelatedTasks`.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmDealServiceImplV2.java
```diff
@@ -0,0 +1,102 @@
...
+		log.info("getDeals: execution ended with {} result(s)", dealsPage.getNumberOfElements());
+		return new ResponseEntityDto(false, pageDto);
+	}
+
+	@Override
+	@Transactional(readOnly = true)
+	public ResponseEntityDto getDealById(Long id) {
+		log.info("getDealById: execution started");
+
+		CrmDeal deal = crmDealDao.findByIdWithAssociations(id);
```
ThinuwanW, line 66: 🤖 **Claude** · 🟡 **Suggestion**

`crmDealDao.findByIdWithAssociations(id)` was written for the v1 DTO: it only does `fetch(stage, INNER)` and `fetch(owner, INNER)`. The new `CrmDealResponseDtoV2` additionally reads `company`, `contact`, `contact.company` and `contact.owner`, all of which are `FetchType.LAZY` — so `GET /v2/crm/deal/{id}` now fires ~4 extra SELECTs per request that v1 never did, plus `CrmUtil.toDealResponseDtoV2` initialises `deal.getCompany()` again for the `isDeleted` check. `CrmTaskServiceImplV2.getTaskById` is worse: `CrmTaskRepositoryImpl.findByIdWithAssociations` fetches `type`, `owner`, `contact`, `deal`, `deal.stage`, `deal.owner`, while `CrmTaskResponseDtoV2` also needs `task.company`, `contact.company`, `contact.owner`, `deal.company`, `deal.contact`, `deal.contact.company` and `deal.contact.owner` (~7 extra SELECTs). Fix: extend the two `findByIdWithAssociations` fetch graphs to cover the associations the V2 DTOs actually read, or reuse the new `buildTaskProjectionSelection` / `findDealsV2` projection for the by-id case so detail and list go through one code path.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmTaskControllerV2IntegrationTest.java
```diff
@@ -0,0 +1,396 @@
+package com.skapp.community.crmplanner.controller.v2;
```
ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

The task-V2 test class has no coverage for the two riskiest pieces of new logic in this PR. (1) Soft-deleted-company masking: `buildTaskProjectionSelection` adds `company.on(isFalse(isDeleted))`, `contactCompany.on(...)`, `dealCompany.on(...)` and `dealContactCompany.on(...)` — four distinct masking paths, none exercised. `CrmUtil.toTaskResponseDtoV2` and its `maskDeletedCompaniesOnDeal` branch (used by `getTaskById`/`createTask`/`editTask`) are likewise untested. The deal test class at least has `getDeals_SoftDeletedCompany_MasksCompany`; the task side has nothing equivalent. (2) Sales-rep owner scoping: no V2 list test asserts that a `CRM_SALES_REPRESENTATIVE` sees only their own tasks/deals, even though `findTasksV2`/`findCompletedTasksV2`/`findRelatedTasksV2`/`findDealsV2` are all new query implementations of that filter. Also missing: an `editTask` denied-for-other-owner test (the deal class has `editDeal_SalesRepEditingOthersDeal_ReturnsBadRequest`) and a V2 counterpart to the `createTask_NonExistentTaskType_ReturnsBadRequest` case this PR just changed in v1. Please add a soft-deleted-company test per V2 task endpoint and an owner-scoping test per V2 list endpoint.

backend/src/main/java/com/skapp/community/crmplanner/util/CrmUtil.java
```diff
@@ -82,4 +86,33 @@ public CrmDealByStageItemResponseDto toDealByStageItemDto(CrmMapper crmMapper, C
...
+			maskDeletedCompaniesOnDeal(dto.getDeal(), task.getDeal());
+		}
+		return dto;
+	}
+
+	private void maskDeletedCompaniesOnDeal(CrmDealResponseDtoV2 dto, CrmDeal deal) {
+		if (hasDeletedCompany(deal)) {
+			dto.setCompany(null);
+		}
+		if (dto.getContact() != null && deal.getContact() != null && hasDeletedCompany(deal.getContact())) {
```
ThinuwanW, line 113: 🤖 **Claude** · 🟣 **Nit**

`deal.getContact() != null` in `maskDeletedCompaniesOnDeal` can never be false: `CrmDeal.contact` is `@ManyToOne @JoinColumn(name = "contact_id", nullable = false)`, so any persisted deal reaching this helper always has a contact (v1 code such as `CrmDealServiceImpl.editDeal` dereferences `deal.getContact().getId()` unguarded on the same basis). Drop the middle condition and keep `if (dto.getContact() != null && hasDeletedCompany(deal.getContact()))`. Separately, consider adding a `hasDeletedCompany(CrmTask)` overload next to the existing `hasDeletedCompany(CrmContact)`/`hasDeletedCompany(CrmDeal)` pair so `toTaskResponseDtoV2` does not have to reach for the private `isCompanyDeleted(task.getCompany())` directly — that asymmetry is what makes this helper read inconsistently.

backend/src/main/java/com/skapp/community/crmplanner/controller/v2/CrmDealControllerV2.java
```diff
@@ -0,0 +1,64 @@
+package com.skapp.community.crmplanner.controller.v2;
```
ThinuwanW, line 1: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

Two gratuitous inconsistencies between the two new sibling controllers added in the same PR. (1) `CrmDealControllerV2` uses `@PreAuthorize("hasAnyRole('ROLE_CRM_SALES_REPRESENTATIVE')")` with a single role on all four endpoints, while `CrmTaskControllerV2` uses `hasRole(...)`. They are functionally identical for one role, but `hasAnyRole` with a single argument reads as if more roles were intended; use `hasRole` consistently. (2) `CrmTaskControllerV2.editTask` declares `@PatchMapping(value = "/{id}", produces = MediaType.APPLICATION_JSON_VALUE)` while `CrmDealControllerV2.editDeal` declares `@PatchMapping("/{id}")` — same operation shape, different mapping. Pick one and apply it to both.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/CrmDealServiceV2.java
```diff
@@ -0,0 +1,18 @@
+package com.skapp.community.crmplanner.service.v2;
```
ThinuwanW, line 1: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

The V2 surface is partial in a way that is not documented: `CrmDealServiceV2` omits `deleteDeal` and `checkDealNameExists`, and `CrmTaskServiceV2` omits `deleteTask`. A client on `/v2/crm/deal` and `/v2/crm/task` therefore has to fall back to `/v1/...` for deletes and for the deal name-exists check, and those v1 responses are v1-shaped. That is fine if intentional, but it should be stated — either add the missing operations as thin delegations to `crmDealService.deleteDeal(id)` / `crmTaskService.deleteTask(id)` (they return only a message, so no DTO work is needed), or note the v1 fallback in the `@Tag` description on the V2 controllers so API consumers are not left guessing.

PR #1750, opened 2026-08-12
[Feat] Serve auth host frontend under the /auth path prefix

akilarootcode:
## Summary

The AWS ALB routes the OAuth issuer host `auth.skapp.dev` by path: `/auth*` goes to the core frontend, everything else (`/*`) to the core backend. This serves the OAuth 2.1 browser leg from **dedicated pages under `/auth`**, so the shared tenant sign-in page no longer carries any auth-host or OAuth branching.

- `next.config.mjs`: `assetPrefix` set to `/auth` in enterprise mode, plus three ordinary page rewrites in the existing array (`/auth/signin`, `/auth/oauth-consent`, `/auth/redirect`). **No `beforeFiles`, no `images` config, no new top-level constants** — `rewrites()` still returns a plain array, exactly the existing pattern used by the other 109 routes.
- **Removed both duplicate root `/oauth-consent` rewrites.** Consent only exists inside the OAuth flow, so it now lives solely at `/auth/oauth-consent`.
- `routes.ts`: `OAUTH_SIGNIN`, `OAUTH_CONSENT`, `OAUTH_REDIRECT` are literal `/auth/*` paths.

### Why the pages are rewrites, not real `/auth/*` files

`frontend/pages/enterprise` is a git submodule mounted at `/enterprise/*`, and a submodule cannot be mounted twice. Real `pages/auth/*` files would therefore have to sit in this (community) repo while importing `~enterprise` components — which breaks community builds when the submodule is empty. The rewrite is what lets a submodule-owned page answer at a super-repo URL. Every one of the other 114 routes works this way.

### Why `assetPrefix` is the one unavoidable addition

Everything else was stripped out and re-tested. `assetPrefix` cannot be:

Without it, the browser is told to fetch `/_next/static/*` **unprefixed**, which the ALB default rule sends to the backend. Measured: page HTML returns `200`, then **11 chunk requests 404** and the page sits on its loading spinner forever — the domain field never renders. `assetPrefix` is the only mechanism that changes what the browser asks for.

With it, `_next/static/*` and `public/` files resolve **natively — no rewrite needed** (verified: correct MIME types, byte-identical payloads). The logo uses a static import, so it is served as `/auth/_next/static/media/logo.<hash>.png` and the image optimizer is not involved at all.

The three page rewrites are ordinary `afterFiles` entries, identical in shape to `/signin → /enterprise/signin`.

## Feature scope (all repos)

This feature spans multiple repos on branch `feat/oauth-fe-route-changes`:

| Repo | What changed |
|------|--------------|
| `SkappHQ/skapp` (super) | `assetPrefix` + three page rewrites in the existing array, `/auth/*` route constants, duplicate consent rewrites removed |
| `rootcodelabs/skapp-ep-fe-src` | OAuth callback/method query-param constants, `buildSsoReturnUrl` |
| `rootcodelabs/skapp-ep-fe-pages` | **New** `auth/signin`, `auth/redirect`; consent **moved** to `auth/oauth-consent`; OAuth branching removed from shared `signin`/`redirect` |
| `rootcodelabs/skapp-ep-be-src` | `OAUTH_SIGN_IN_PATH` / `OAUTH_CONSENT_PAGE` composed from an `/auth` prefix constant |

## Architecture / flow

```mermaid
flowchart LR
    C[Claude MCP client] -->|/oauth2/authorize| ALB{ALB rule<br/>auth.skapp.dev}
    ALB -->|/auth*| FE[core FE]
    ALB -->|/* default| BE[core BE]
    BE -->|302 /auth/signin?callback=| P1[enterprise/auth/signin]
    P1 -->|credentials| S[POST /v1/auth/oauth/session-login]
    P1 -->|SSO| P3[enterprise/auth/redirect]
    P3 --> S
    S --> BE
    BE -->|302 /auth/oauth-consent| P2[enterprise/auth/oauth-consent]
    P2 -->|approve| BE
    BE -->|302 code| C
```

- **The raw `fetch` in `AuthProvider.signIn`** stays a raw `fetch`. Pre-existing, and it targets a same-origin Next API route rather than the backend; routing it through `authFetchSameOrigin` would attach the auth/tenant interceptors and call `getAccessToken()` during sign-in, before credentials exist. Only the path changed.
- **`?? ""` in `getSafeOAuthCallback`** — pre-existing normalisation of `URLSearchParams.get()`, which returns `null`. The function's public contract still returns `undefined` for absent.
- **Optimising a fixed-size PNG logo through `next/image`** is wasteful, but pre-existing across 13 files. A static import in `SkappLogo` is the right fix and belongs in its own PR.

- **Submodule hash / pointer bumps** are intentionally excluded. Gitlink updates happen at **deployment time**, not in feature PRs.
- **ALB configuration.** The path pattern must be `/auth*` (or `/auth/*`) — ALB matches `/auth` literally, so `/auth/signin` would miss the rule.
- **CDN behaviours keyed on `/_next/*`** must become `/auth/_next/*`; `assetPrefix` changes asset URLs on every host.
- **`/api/clear-cookies` is no longer prefixed on the auth host**, so the pre-sign-in cookie clear silently no-ops there (`fetch` resolves on 404). Deliberate: that endpoint clears every cookie for the domain, which on `.skapp.dev` would drop other tenants' refresh tokens and the shared `tenant` cookie that session detection relies on. Sign-in verified working without it.
- **`/_next/data/*` is not prefixed by `assetPrefix`** — not an issue here, as all three auth-host pages are purely client-rendered (no `getServerSideProps`/`getStaticProps`), so Next issues no data requests for them. It would matter if a future auth-host page added server-side data fetching.

## Testing

| Check | Ran? | Result |
|-------|------|--------|
| Unit tests | n/a | No existing spec covers the changed modules; verified by live flow below |
| Formatter (BE only) | yes | `mvn checkstyle:check` pass |
| tsc + lint (BE only) | yes | `mvn compile` pass. FE `tsc` re-baselined against current `develop` after rebase — **655 before, 655 after, identical per file+code**. `next lint` clean, prettier clean |
| e2e (Playwright) | yes | Full OAuth flow driven in a real browser against `https://auth.skapp.dev` — pass |

Verified with nginx locally mirroring the ALB topology (`/auth*` → FE, `/*` → BE) on the real hostname, with the stored consent row deleted first so the consent screen was genuinely exercised:

- Backend `302` on unauthenticated `/oauth2/authorize` targets `/auth/signin?callback=…`
- Request trail served the **new** chunks: `pages/enterprise/auth/signin`, `pages/enterprise/auth/oauth-consent`
- `/auth/signin`, `/auth/oauth-consent`, `/auth/redirect` → `200`; `/auth/_next/static/*` and the static-import logo → `200` with correct MIME, **no rewrite covering them**; browser run shows **zero failed requests**
- Sign-in → `session-login` → consent → `claude.ai/api/mcp/auth_callback?code=…` issued a real authorization code; consent row and master authorization confirmed in MySQL
- Old root `/oauth-consent` now `404` on tenant hosts, as intended
- Tenant host regression-checked: `/signin` 200, `/dashboard` 307 (auth redirect), `/redirect` 200, `/api/clear-cookies` 200 — all unaffected
- **Counter-test**: with `assetPrefix` removed, page HTML still returned `200` but 11 chunks 404'd and the page hung on its spinner — which is why it is the one addition that stays

## Related PRs

Same branch `feat/oauth-fe-route-changes` in every repo; all four merge together — the auth host is broken with a partial merge.

- SkappHQ/skapp#1750 — super repo: rewrites, assetPrefix, route constants
- rootcodelabs/skapp-ep-fe-src#779 — callback constants, SSO return URL builder
- rootcodelabs/skapp-ep-fe-pages#350 — dedicated `auth/*` pages; OAuth branching removed from shared pages
- rootcodelabs/skapp-ep-be-src#747 — backend sign-in + consent page constants

### DIFF
--- a/frontend/next.config.mjs
+++ b/frontend/next.config.mjs
@@ -1,10 +1,24 @@
+const isEnterpriseMode = process.env.NEXT_PUBLIC_MODE === "enterprise";
+
 /** @type {import('next').NextConfig} */
 const nextConfig = {
   output: "standalone",
   reactStrictMode: false,
+  assetPrefix: isEnterpriseMode ? "/auth" : undefined,
   async rewrites() {
-    const isEnterpriseMode = process.env.NEXT_PUBLIC_MODE === "enterprise";
     return [
+      {
+        source: "/auth/signin",
+        destination: "/enterprise/auth/signin"
+      },
+      {
+        source: "/auth/oauth-consent",
+        destination: "/enterprise/auth/oauth-consent"
+      },
+      {
+        source: "/auth/redirect",
+        destination: "/enterprise/auth/redirect"
+      },
       {
         source: "/welcome",
         destination: "/community/welcome"
@@ -46,10 +60,6 @@ const nextConfig = {
         destination: isEnterpriseMode
           ? "/enterprise/signin"
           : "/community/signin"
-      },
-      {
-        source: "/oauth-consent",
-        destination: "/enterprise/oauth-consent"
       },
       {
         source: "/settings",
@@ -254,10 +264,6 @@ const nextConfig = {
       {
         source: "/redirect",
         destination: "/enterprise/redirect"
-      },
-      {
-        source: "/oauth-consent",
-        destination: "/enterprise/oauth-consent"
       },
       {
         source: "/verify/reset-password",
--- a/frontend/src/community/common/constants/routes.ts
+++ b/frontend/src/community/common/constants/routes.ts
@@ -200,3 +200,8 @@ export const nonSuperAdminRestrictedRoutes = [
 export const managerRestrictedRoutes = [ROUTES.PEOPLE.ADD];
 
 export const userRolesRestrictedRoutes = [ROUTES.CONFIGURATIONS.USER_ROLES];
+
+export const oAuthCallbackRoutes = [
+  ROUTES.AUTH.OAUTH_AUTHORIZE,
+  ROUTES.AUTH.OAUTH_CONSENT
+];
--- a/frontend/src/community/common/constants/routes.ts
+++ b/frontend/src/community/common/constants/routes.ts
@@ -18,7 +18,9 @@ const ROUTES = {
     FORGET_PASSWORD: "/forget-password",
     SYSTEM_UPDATE: "/system-update",
     OAUTH_AUTHORIZE: "/oauth2/authorize",
-    OAUTH_CONSENT: "/oauth-consent"
+    OAUTH_SIGNIN: "/auth/signin",
+    OAUTH_CONSENT: "/auth/oauth-consent",
+    OAUTH_REDIRECT: "/auth/redirect"
   },
   ORGANIZATION: {
     SETUP: "/setup-organization",


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR serves the enterprise auth-host frontend routes under a `/auth` path prefix by setting a build-wide `assetPrefix`/`images.path` and adding `beforeFiles` rewrites, plus routing the clear-cookies call through a new `getAuthHostPath()` helper (with a fallback stub for community). The routing/config changes are coherent in intent, but the asset story looks incomplete: `/auth/_next/static/*` has no rewrite even though `/auth/_next/image` needed one, and `assetPrefix` applies to every enterprise host rather than just the auth host. I also could not find a `getAuthHostPath` export in the enterprise `tenantUtil` implementation in this tree, and `typescript.ignoreBuildErrors: true` would let that slip through to a runtime failure in the sign-in path.

Found **16** new issue(s): 🔴 4 important, 🟡 9 suggestion(s), 🟣 3 nit(s)

frontend/next.config.mjs
```diff
@@ -453,6 +462,32 @@ const nextConfig = {
         destination: "/community/crm/tasks"
       }
     ];
+
+    return {
+      beforeFiles: [
+        {
+          source: `${AUTH_HOST_PATH_PREFIX}${IMAGE_OPTIMIZER_PATH}`,
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Only `/auth/_next/image` is rewritten back to `/_next/image`; there is no rewrite for `/auth/_next/static/:path*`. Because `assetPrefix` is applied at build time, every rendered page (on every enterprise host) will reference its JS/CSS chunks as `/auth/_next/static/chunks/*.js`. In production (`next start` / standalone) Next.js does not strip a path-style `assetPrefix` from incoming request URLs — that is precisely why you needed the explicit image-optimizer rewrite — so those chunk requests will 404 and pages will render blank/unhydrated. Note this can pass in `next dev` and fail only in prod. Fix: replace the single image entry with a catch-all in `beforeFiles`, e.g. source `/auth/_next/:path*` → destination `/_next/:path*` (this subsumes the image rewrite and also covers `_next/data` and webpack-hmr), or switch to Next's `basePath` which handles prefixing end-to-end. Conversely, if your reverse proxy already strips `/auth` before forwarding to Next, then the `/auth/_next/image`, `/auth/api/clear-cookies` and `/auth/signin|oauth-consent|redirect` rewrites are all dead config — one of the two assumptions has to be wrong, so please confirm against a real production build before merging.

frontend/src/community/auth/providers/AuthProvider.tsx
```diff
@@ -13,9 +13,11 @@ import {
   EnterpriseSignInParams,
   EnterpriseSignUpParams
 } from "~enterprise/auth/utils/authUtils";
+import { getAuthHostPath } from "~enterprise/common/utils/tenantUtil";
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`getAuthHostPath` is imported from `~enterprise/common/utils/tenantUtil`, but the enterprise implementation at `frontend/src/enterprise/common/utils/tenantUtil.ts` in this tree does not export it (only the fallback stub added in this PR does, and `frontend/src/enterprise/common/constants/stringConstants.ts` has no `/auth` prefix constant either). `tsconfig.json` resolves `~enterprise/*` to `./src/enterprise/*` BEFORE `./src/fallback/*`, so in enterprise builds this resolves to the real module and `getAuthHostPath` will be `undefined` → `TypeError: getAuthHostPath is not a function` on the very first line of `signIn`, breaking enterprise sign-in entirely. `next.config.mjs` has `typescript.ignoreBuildErrors: true`, so the missing export will not fail CI — there is no compile-time safety net here. Please make sure the `src/enterprise` submodule bump that adds `getAuthHostPath` (and the `/auth` prefix constant) is part of this PR; the same applies to the `pages/enterprise/{signin,redirect,oauth-consent}.tsx` usages.

frontend/next.config.mjs
```diff
@@ -1,10 +1,19 @@
...
+const IMAGE_OPTIMIZER_PATH = "/_next/image";
+const CLEAR_COOKIES_API_PATH = "/api/clear-cookies";
+
+const isEnterpriseMode = process.env.NEXT_PUBLIC_MODE === "enterprise";
+
 /** @type {import('next').NextConfig} */
 const nextConfig = {
   output: "standalone",
   reactStrictMode: false,
+  assetPrefix: isEnterpriseMode ? AUTH_HOST_PATH_PREFIX : undefined,
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`assetPrefix` and `images.path` are gated on `NEXT_PUBLIC_MODE === "enterprise"`, which is true for every enterprise deployment/host — not just the auth host. That means pages served on `app.*` and `{tenant}.*` will also load all their static assets and optimized images from `/auth/_next/*`, so the ingress for every enterprise host must forward `/auth/_next/*` to the frontend or the whole app breaks, not just the auth-host pages. It also affects local enterprise dev on `localhost`, where no `/auth` prefix is otherwise in play. If the prefix is really an auth-host-only deployment concern, gate it on a dedicated env var (e.g. `NEXT_PUBLIC_AUTH_HOST_PATH_PREFIX` set only for the auth-host deployment) instead of the global mode flag; if it is intentional for all hosts, please note that in a comment here so the ingress requirement is explicit.

frontend/next.config.mjs
```diff
@@ -453,6 +462,32 @@ const nextConfig = {
         destination: "/community/crm/tasks"
       }
     ];
+
+    return {
+      beforeFiles: [
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The `beforeFiles` rewrites are unconditional, unlike every other enterprise-specific rewrite in this file (which uses `isEnterpriseMode ? ... : ...`). In community builds `/auth/signin`, `/auth/oauth-consent` and `/auth/redirect` will now rewrite to `/enterprise/*` pages that do not exist (the `pages/enterprise` submodule is empty in community), producing confusing dead routes. Wrap the array: `beforeFiles: isEnterpriseMode ? [ ... ] : []` for consistency with `afterFiles` and with the `assetPrefix`/`images` gating above.

frontend/next.config.mjs
```diff
@@ -453,6 +462,32 @@ const nextConfig = {
...
+        {
+          source: `${AUTH_HOST_PATH_PREFIX}${IMAGE_OPTIMIZER_PATH}`,
+          destination: IMAGE_OPTIMIZER_PATH
+        },
+        {
+          source: `${AUTH_HOST_PATH_PREFIX}${CLEAR_COOKIES_API_PATH}`,
+          destination: CLEAR_COOKIES_API_PATH
+        },
+        {
+          source: `${AUTH_HOST_PATH_PREFIX}/signin`,
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The `/auth` prefix is only applied at three of the auth-host entry points. Other code paths that navigate to the sign-in page still use the unprefixed route and were not updated: `frontend/pages/index.tsx:49` (`isAuthHost()` branch does `router.replace(ROUTES.AUTH.SIGNIN)` → `/signin`), and `frontend/src/community/auth/utils/authUtils.ts:345` in `signOut()` (`${ROUTES.AUTH.SIGNIN}?callback=...`), which is invoked from `authInterceptor.ts:46` on invalid/expired refresh tokens. A 401 or logout while the user is on `/auth/oauth-consent` will therefore bounce them to `/auth`-less `/signin` on the auth host, which under this new scheme is no longer the frontend route. Route these through `getAuthHostPath(ROUTES.AUTH.SIGNIN)` the same way `pages/enterprise/redirect.tsx` and `oauth-consent.tsx` do.

frontend/src/community/auth/providers/AuthProvider.tsx
```diff
@@ -85,7 +87,7 @@ export const AuthProvider: React.FC<AuthProviderProps> = ({ children }) => {
     async (params: EnterpriseSignInParams): Promise<AuthResponseType> => {
       try {
         // Clear all existing cookies before signing in
-        await fetch("/api/clear-cookies", {
+        await fetch(getAuthHostPath(CLEAR_COOKIES_API_PATH), {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The clear-cookies response is never inspected. A 404 (exactly what happens if the `/auth/api/clear-cookies` rewrite is missing or the prefix is wrong on a given host) does not reject a `fetch`, so sign-in silently proceeds with stale cookies from a previous tenant/session — the failure mode this call exists to prevent, and one this PR makes host-dependent. Capture the response and at minimum log/handle a non-OK status, e.g. `const res = await fetch(getAuthHostPath(CLEAR_COOKIES_API_PATH), { method: "POST" }); if (!res.ok) { console.error("Failed to clear cookies before sign-in", res.status); }` (or throw, if proceeding with stale cookies is unacceptable).

frontend/src/community/common/constants/stringConstants.ts
```diff
@@ -84,3 +84,5 @@ export const CANCEL = "cancel";
 export const APP = "app";
 export const AUTH = "auth";
 export const LOCALHOST = "localhost";
+
+export const CLEAR_COOKIES_API_PATH = "/api/clear-cookies";
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`/api/clear-cookies` is now defined twice — here and as `CLEAR_COOKIES_API_PATH` in `next.config.mjs:3` — and the `/auth` prefix literal now exists in at least three places: `next.config.mjs:1`, the (missing) enterprise `getAuthHostPath` implementation, and `EpCommonConstants.AUTH_HOST_PATH_PREFIX` in the backend. Since `next.config.mjs` cannot import TS, full deduplication is not possible, but exporting `AUTH_HOST_PATH_PREFIX = "/auth"` from this file (next to the existing `APP`/`AUTH` constants) would at least give the TS side a single source of truth, and a short comment in `next.config.mjs` pointing at it would flag the drift risk.

frontend/next.config.mjs
```diff
@@ -453,6 +462,32 @@ const nextConfig = {
         destination: "/community/crm/tasks"
       }
     ];
+
+    return {
+      beforeFiles: [
+        {
+          source: `${AUTH_HOST_PATH_PREFIX}${IMAGE_OPTIMIZER_PATH}`,
+          destination: IMAGE_OPTIMIZER_PATH
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`public/` assets are covered neither by `assetPrefix` (Next only prefixes `_next/*`) nor by the new rewrites. `pages/_document.tsx` uses relative favicon hrefs (`favicon/favicon-16x16.png`), which resolve against the current path — on `/auth/signin` the browser requests `/auth/favicon/favicon-16x16.png` and gets a 404. Low impact (favicon only today), but any future raw `` or CSS `url(/...)` on an auth-host page will hit the same gap. Either make those hrefs root-absolute (`/favicon/...`) or add a passthrough rewrite for the public paths the auth host needs.

frontend/next.config.mjs
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The `beforeFiles` block only maps three pages under the prefix (`/auth/signin`, `/auth/oauth-consent`, `/auth/redirect`), but the auth-host sign-in journey reaches several more Next pages. There is no rewrite for the bare `/auth` entry point (a user landing on `https://auth.<domain>/auth` gets a 404, and `pages/index.tsx` only redirects when it is reached at `/`), nor for `/auth/forget-password`, `/auth/reset-password`, `/auth/verify/reset-password`, `/auth/verify/email`, `/auth/verify/success`, `/auth/signup` or `/auth/module-selection` — all of which exist as unprefixed `afterFiles` entries and are linked from the sign-in page (e.g. `SignUpPromptLink`, the forgot-password link, and `ROUTES.AUTH.FORGET_PASSWORD`/`RESET_PASSWORD`). Under a gateway that forwards only `/auth/*` to the frontend these all 404. Fix: build the prefixed list from a single source of truth rather than hand-enumerating three entries — e.g. define `const AUTH_HOST_ROUTES = ["/signin", "/oauth-consent", "/redirect", "/forget-password", "/reset-password", "/verify/reset-password", "/verify/email", "/verify/success", "/signup", "/module-selection"]` and `.map()` it into both the prefixed `beforeFiles` entries and the existing unprefixed `afterFiles` entries, plus add a `${AUTH_HOST_PATH_PREFIX}` -> `/enterprise/signin` entry for the bare prefix. Whatever set you choose must match the set of paths the client code prefixes via `getAuthHostPath`, otherwise the two halves drift.

frontend/next.config.mjs
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The new `/auth/*` URL namespace is invisible to `middleware.ts`: its `config.matcher` (middleware.ts:381-423) lists only unprefixed paths, and middleware runs on the incoming request path before `beforeFiles` rewrites are applied. Today the three mapped pages are unauthenticated so there is no live hole, but this makes the prefix a silent security footgun — adding any protected page (e.g. `/auth/module-selection`, `/auth/settings/...`) to this `beforeFiles` list would bypass every role check in `allowedRoutes`, the `nonSuperAdminRestrictedRoutes`/`managerRestrictedRoutes` checks, and the `isPasswordChangedForTheFirstTime === "false"` forced-reset redirect. It also means the unconditional `/auth/signin` rewrite gives every host a second URL for the sign-in page that middleware never sees. Fix: add `"/auth/:path*"` to the middleware matcher and normalise the prefix at the top of `middleware()` (e.g. `const currentPath = request.nextUrl.pathname.replace(/^\/auth(?=\/|$)/, "") || "/"`) so all downstream `currentPath`/`request.nextUrl.pathname` comparisons keep working for both prefixed and unprefixed URLs.

frontend/src/fallback/common/utils/tenantUtil.ts
```diff
@@ -8,6 +8,8 @@ export const isAuthHost = (): boolean => false;
 
 export const isTenantSelectionHost = (): boolean => false;
 
+export const getAuthHostPath = (path: string): string => path;
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`getAuthHostPath` has an ambiguous contract that nothing enforces. (a) The name says "auth host", but the fallback returns a bare same-origin path. The single caller in community code does `fetch(getAuthHostPath(CLEAR_COOKIES_API_PATH), { method: "POST" })` with no `credentials` and no `mode` — if the enterprise implementation ever returns an absolute cross-origin URL (e.g. `https://auth.<domain>/api/clear-cookies`) that call silently fails CORS and the `Set-Cookie` response is discarded. Either rename to something unambiguous (`withAuthHostPathPrefix` / `getAuthHostPathPrefix`) and add a JSDoc stating it returns a same-origin path only, or make the caller explicit with `credentials: "same-origin"`. (b) There is no shared type contract between this fallback module and `src/enterprise/common/utils/tenantUtil.ts`, and `next.config.mjs` sets both `typescript.ignoreBuildErrors: true` and `eslint.ignoreDuringBuilds: true`, so any signature or export drift between the two ships silently and surfaces only as a runtime `TypeError: ... is not a function`. Declare a shared `TenantUtils` interface in community code that both modules `satisfies`, and/or add a `tenantUtil.test.ts` asserting export parity — note the sibling utils in this directory (`awsS3ServiceFunctions`, `firebase`, `monitoring`, `getEnterpriseDrawerRoutes`) all have tests but `tenantUtil` does not.

frontend/src/community/auth/providers/AuthProvider.tsx
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

This is a raw `fetch` outside the established API layer. The codebase already exports a client for exactly this case: `authFetchSameOrigin` in `~community/common/utils/axiosInterceptor` (`baseURL: ""`, `withCredentials: true`, request interceptor attached), and it is the pattern used by `src/enterprise/common/api/authApi.ts:149,169` for same-origin POSTs. Switching to `await authFetchSameOrigin.post(getAuthHostPath(CLEAR_COOKIES_API_PATH))` keeps the convention, makes credential handling explicit rather than relying on the `fetch` default, and (because axios rejects on non-2xx) removes the need for manual status inspection. If the raw `fetch` is kept, at minimum add `credentials: "same-origin"`.

frontend/src/community/auth/providers/AuthProvider.tsx
```diff
@@ -13,9 +13,11 @@ import {
   EnterpriseSignInParams,
   EnterpriseSignUpParams
 } from "~enterprise/auth/utils/authUtils";
+import { getAuthHostPath } from "~enterprise/common/utils/tenantUtil";
 
 import FullScreenLoader from "../../common/components/molecules/FullScreenLoader/FullScreenLoader";
 import ROUTES from "../../common/constants/routes";
+import { CLEAR_COOKIES_API_PATH } from "../../common/constants/stringConstants";
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Deep relative import across modules. The review guidelines require the path aliases (`~community/*`) and this PR uses one on the line just above (line 16, `~enterprise/common/utils/tenantUtil`) while adding a relative one here — inconsistent within the same diff. Change to `import { CLEAR_COOKIES_API_PATH } from "~community/common/constants/stringConstants";`. The pre-existing relative imports on lines 18-29 are worth normalising in the same pass since the file is already being touched.

frontend/src/community/common/constants/stringConstants.ts
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Wrong home for this constant (separate from the value being duplicated in `next.config.mjs`). `stringConstants.ts` holds UI literals, enums and length limits; an API route path belongs with the other endpoint constants — `~community/common/api/utils/ApiEndpoints.ts` (or a small `internalApiEndpoints` object there, since this is a Next API route rather than a backend endpoint). Putting it there also keeps it discoverable next to `authenticationEndpoints`/`organizationCreateEndpoints` so the next person adding a same-origin API call finds it instead of hardcoding the string again.

frontend/next.config.mjs
```diff
@@ -453,6 +462,32 @@ const nextConfig = {
...
+        {
+          source: `${AUTH_HOST_PATH_PREFIX}${CLEAR_COOKIES_API_PATH}`,
+          destination: CLEAR_COOKIES_API_PATH
+        },
+        {
+          source: `${AUTH_HOST_PATH_PREFIX}/signin`,
+          destination: "/enterprise/signin"
+        },
+        {
+          source: `${AUTH_HOST_PATH_PREFIX}/oauth-consent`,
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The rewrite table has dead duplicate entries that this PR is now adding a third variant to. `/oauth-consent` is declared twice in `afterFiles` (lines 59-62 and 267-270) plus once here in `beforeFiles`; only the first `afterFiles` entry is ever reachable. `/verify/guest` is declared twice back to back (lines 279-282 and 283-286) and `/leave/analytics/:id` twice (lines 183-186 and 251-254). Since the PR is already restructuring the return value of `rewrites()`, delete the unreachable duplicates — otherwise a future change to one copy appears to take effect but does not.

frontend/next.config.mjs
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`AUTH_HOST_PATH_PREFIX` is reused for two conceptually different things: the route prefix the auth host is mounted under, and the mount point for static assets (`assetPrefix`, plus `images.path` on line 13). They happen to be the same string today, but they are set by different infrastructure concerns — if the CDN/asset origin ever moves independently of the auth-host route prefix, the coupling breaks silently in both directions. Consider two named constants (`AUTH_HOST_PATH_PREFIX` and `ASSET_PREFIX = AUTH_HOST_PATH_PREFIX`) with a short comment on why they currently coincide, so the intent is explicit at each use site.

frontend/next.config.mjs
akilarootcode: removed

frontend/src/community/auth/providers/AuthProvider.tsx
akilarootcode: removed

frontend/next.config.mjs
akilarootcode: remove both of them

frontend/next.config.mjs
akilarootcode: removed before files

frontend/next.config.mjs
akilarootcode: removed entry points and changed the approach

frontend/src/community/auth/providers/AuthProvider.tsx
akilarootcode: removed this as well

frontend/src/community/common/constants/stringConstants.ts
akilarootcode: removed this

frontend/next.config.mjs
akilarootcode: removed asset suffix

frontend/next.config.mjs
akilarootcode: removed before files

frontend/next.config.mjs
akilarootcode: removed the flow and added as route

frontend/src/fallback/common/utils/tenantUtil.ts
akilarootcode: remove these

frontend/src/community/auth/providers/AuthProvider.tsx
akilarootcode: removed and role back to existing impl

frontend/src/community/auth/providers/AuthProvider.tsx
akilarootcode: removed

frontend/src/community/common/constants/stringConstants.ts
akilarootcode: removed

frontend/next.config.mjs
akilarootcode: removed

frontend/next.config.mjs
akilarootcode: removed

hasalarootcode (dismissed):

hasalarootcode (approved):

PR #1751, opened 2026-08-12
fix: persist holiday before processing conflicting leave requests

SanithuGunasena07:
Bug: Creating a holiday returns a 500 error on the first attempt when a pending or approved leave request exists on the same date. Retrying the same submission succeeds.

Root Cause: In saveBulkHolidays, the new Holiday was added to a leave-request-conflict handler (updateLeaveRequestDueHoliday) before being persisted to the database, so its id was still null. The conflict handler's notification/email step calls holiday.getId().toString(), throwing a NullPointerException. As a side effect, the colliding leave request's status was already flipped to CANCELLED/REVOKED and saved before the crash, so a retry no longer finds a conflicting request and succeeds.

Fix: Persist each Holiday (holidayDao.save(holiday)) immediately after building it, before checking for conflicting leave requests, so it has a valid ID by the time the conflict-notification path runs

### DIFF
--- a/backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
@@ -58,6 +58,7 @@ import java.time.format.DateTimeParseException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.EnumSet;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -124,11 +125,13 @@ public class HolidayServiceImpl implements HolidayService {
 	}
 
 	@Override
+	@Transactional(propagation = Propagation.REQUIRED)
 	public ResponseEntityDto saveBulkHolidays(HolidayBulkRequestDto holidayBulkRequestDto) {
 		log.info("saveBulkHolidays: execution started");
 
 		List<HolidayDtoStatusResponseDto> holidayDtoStatusList = new ArrayList<>();
-		List<Holiday> savableHolidays = new ArrayList<>();
+		List<Holiday> savedHolidays = new ArrayList<>();
+		Map<LocalDate, List<Holiday>> systemHolidaysByDate = new HashMap<>();
 		AtomicInteger holidaysOnCurrentDate = new AtomicInteger();
 		AtomicInteger holidaysOnPastDates = new AtomicInteger();
 
@@ -141,7 +144,8 @@ public class HolidayServiceImpl implements HolidayService {
 			try {
 
 				LocalDate holidayDate = DateTimeUtils.parseUtcDate(holidayDto.getDate());
-				List<Holiday> systemHolidays = holidayDao.findAllByIsActiveTrueAndDate(holidayDate);
+				List<Holiday> systemHolidays = systemHolidaysByDate.computeIfAbsent(holidayDate,
+						holidayDao::findAllByIsActiveTrueAndDate);
 
 				validateHolidayDto(holidayDto, holidayDate, systemHolidays, holidaysOnCurrentDate, holidaysOnPastDates,
 						holidayBulkRequestDto.getYear(), validWorkLocationNames);
@@ -151,9 +155,7 @@ public class HolidayServiceImpl implements HolidayService {
 				Set<WorkLocation> workLocations = resolveWorkLocations(holidayDto.getWorkLocations(),
 						workLocationsByName);
 				holiday.setWorkLocations(workLocations);
-				Holiday savedHoliday = holidayDao.save(holiday);
-
-				savableHolidays.add(savedHoliday);
+				holidayDao.save(holiday);
 
 				LeaveRequestFilterDto leaveRequestFilterDto = new LeaveRequestFilterDto();
 				leaveRequestFilterDto.setEndDate(holidayDate);
@@ -161,8 +163,10 @@ public class HolidayServiceImpl implements HolidayService {
 				List<LeaveRequest> leaveRequests = leaveRequestDao
 					.findAllLeaveRequestsByDateRange(leaveRequestFilterDto);
 				if (systemHolidays.isEmpty() && !leaveRequests.isEmpty()) {
-					leaveRequests.forEach(leaveRequest -> updateLeaveRequestDueHoliday(savedHoliday, leaveRequest));
+					leaveRequests.forEach(leaveRequest -> updateLeaveRequestDueHoliday(holiday, leaveRequest));
 				}
+
+				savedHolidays.add(holiday);
 			}
 			catch (ModuleException | DateTimeParseException e) {
 				log.warn("saveBulkHolidays: Found a record with error: {}", e.getMessage());
@@ -171,21 +175,21 @@ public class HolidayServiceImpl implements HolidayService {
 			}
 		});
 
-		if (savableHolidays.size() == 1) {
+		if (savedHolidays.size() == 1) {
 
-			Set<WorkLocation> holidayWorkLocations = savableHolidays.getFirst().getWorkLocations();
+			Set<WorkLocation> holidayWorkLocations = savedHolidays.getFirst().getWorkLocations();
 			Set<WorkLocation> workLocationsFilter = (holidayWorkLocations == null || holidayWorkLocations.isEmpty())
 					? null : holidayWorkLocations;
 
 			List<Employee> employees = employeeDao.findAllActiveEmployeesExcludingRole(Role.PM_GUEST_EMPLOYEE,
 					workLocationsFilter);
 
-			peopleEmailService.sendNewHolidayDeclarationEmail(savableHolidays.getFirst(), employees);
-			peopleNotificationService.sendNewHolidayDeclarationNotification(savableHolidays.getFirst(), employees);
+			peopleEmailService.sendNewHolidayDeclarationEmail(savedHolidays.getFirst(), employees);
+			peopleNotificationService.sendNewHolidayDeclarationNotification(savedHolidays.getFirst(), employees);
 		}
 
 		HolidayBulkSaveResponseDto responseDto = getHolidayBulkUploadResponseSummaryText(
-				holidayBulkRequestDto.getHolidayDtoList().size(), savableHolidays.size(),
+				holidayBulkRequestDto.getHolidayDtoList().size(), savedHolidays.size(),
 				holidaysOnCurrentDate.get() > 0, holidaysOnPastDates.get() > 0, holidayDtoStatusList);
 
 		log.info("saveBulkHolidays: execution ended");
--- a/backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
@@ -151,8 +151,9 @@ public class HolidayServiceImpl implements HolidayService {
 				Set<WorkLocation> workLocations = resolveWorkLocations(holidayDto.getWorkLocations(),
 						workLocationsByName);
 				holiday.setWorkLocations(workLocations);
+				Holiday savedHoliday = holidayDao.save(holiday);
 
-				savableHolidays.add(holiday);
+				savableHolidays.add(savedHoliday);
 
 				LeaveRequestFilterDto leaveRequestFilterDto = new LeaveRequestFilterDto();
 				leaveRequestFilterDto.setEndDate(holidayDate);
@@ -160,7 +161,7 @@ public class HolidayServiceImpl implements HolidayService {
 				List<LeaveRequest> leaveRequests = leaveRequestDao
 					.findAllLeaveRequestsByDateRange(leaveRequestFilterDto);
 				if (systemHolidays.isEmpty() && !leaveRequests.isEmpty()) {
-					leaveRequests.forEach(leaveRequest -> updateLeaveRequestDueHoliday(holiday, leaveRequest));
+					leaveRequests.forEach(leaveRequest -> updateLeaveRequestDueHoliday(savedHoliday, leaveRequest));
 				}
 			}
 			catch (ModuleException | DateTimeParseException e) {
@@ -170,18 +171,17 @@ public class HolidayServiceImpl implements HolidayService {
 			}
 		});
 
-		List<Holiday> savedHolidays = holidayDao.saveAll(savableHolidays);
-		if (savedHolidays.size() == 1) {
+		if (savableHolidays.size() == 1) {
 
-			Set<WorkLocation> holidayWorkLocations = savedHolidays.getFirst().getWorkLocations();
+			Set<WorkLocation> holidayWorkLocations = savableHolidays.getFirst().getWorkLocations();
 			Set<WorkLocation> workLocationsFilter = (holidayWorkLocations == null || holidayWorkLocations.isEmpty())
 					? null : holidayWorkLocations;
 
 			List<Employee> employees = employeeDao.findAllActiveEmployeesExcludingRole(Role.PM_GUEST_EMPLOYEE,
 					workLocationsFilter);
 
-			peopleEmailService.sendNewHolidayDeclarationEmail(savedHolidays.getFirst(), employees);
-			peopleNotificationService.sendNewHolidayDeclarationNotification(savedHolidays.getFirst(), employees);
+			peopleEmailService.sendNewHolidayDeclarationEmail(savableHolidays.getFirst(), employees);
+			peopleNotificationService.sendNewHolidayDeclarationNotification(savableHolidays.getFirst(), employees);
 		}
 
 		HolidayBulkSaveResponseDto responseDto = getHolidayBulkUploadResponseSummaryText(
--- a/backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
@@ -186,8 +186,8 @@ public class HolidayServiceImpl implements HolidayService {
 		}
 
 		HolidayBulkSaveResponseDto responseDto = getHolidayBulkUploadResponseSummaryText(
-				holidayBulkRequestDto.getHolidayDtoList().size(), savedHolidays.size(),
-				holidaysOnCurrentDate.get() > 0, holidaysOnPastDates.get() > 0, holidayDtoStatusList);
+				holidayBulkRequestDto.getHolidayDtoList().size(), savedHolidays.size(), holidaysOnCurrentDate.get() > 0,
+				holidaysOnPastDates.get() > 0, holidayDtoStatusList);
 
 		log.info("saveBulkHolidays: execution ended");
 		return new ResponseEntityDto(false, responseDto);
--- a/backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
@@ -58,7 +58,6 @@ import java.time.format.DateTimeParseException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.EnumSet;
-import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -131,7 +130,6 @@ public class HolidayServiceImpl implements HolidayService {
 
 		List<HolidayDtoStatusResponseDto> holidayDtoStatusList = new ArrayList<>();
 		List<Holiday> savedHolidays = new ArrayList<>();
-		Map<LocalDate, List<Holiday>> systemHolidaysByDate = new HashMap<>();
 		AtomicInteger holidaysOnCurrentDate = new AtomicInteger();
 		AtomicInteger holidaysOnPastDates = new AtomicInteger();
 
@@ -144,8 +142,7 @@ public class HolidayServiceImpl implements HolidayService {
 			try {
 
 				LocalDate holidayDate = DateTimeUtils.parseUtcDate(holidayDto.getDate());
-				List<Holiday> systemHolidays = systemHolidaysByDate.computeIfAbsent(holidayDate,
-						holidayDao::findAllByIsActiveTrueAndDate);
+				List<Holiday> systemHolidays = holidayDao.findAllByIsActiveTrueAndDate(holidayDate);
 
 				validateHolidayDto(holidayDto, holidayDate, systemHolidays, holidaysOnCurrentDate, holidaysOnPastDates,
 						holidayBulkRequestDto.getYear(), validWorkLocationNames);


ThinuwanW (commented):
## 🤖 Claude Code Review

The fix correctly addresses the root cause: `PeopleNotificationServiceImpl` calls `holiday.getId().toString()` on the (previously transient) holiday, which NPE'd into a 500 whenever a bulk-uploaded holiday collided with an existing leave request. Saving the holiday before `updateLeaveRequestDueHoliday` is the right call. However, `saveBulkHolidays` has no `@Transactional`, so moving `save()` inside the loop turns an atomic `saveAll()` into N independently-committed transactions, and it also makes each iteration's duplicate/limit lookup see holidays inserted earlier in the same request — a silent change to both validation and leave-cancellation behavior that no test covers.

Found **15** new issue(s): 🔴 6 important, 🟡 5 suggestion(s), 🟣 4 nit(s)

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
```diff
@@ -151,16 +151,17 @@ public ResponseEntityDto saveBulkHolidays(HolidayBulkRequestDto holidayBulkReque
 				Set<WorkLocation> workLocations = resolveWorkLocations(holidayDto.getWorkLocations(),
 						workLocationsByName);
 				holiday.setWorkLocations(workLocations);
+				Holiday savedHoliday = holidayDao.save(holiday);
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`saveBulkHolidays` is not annotated `@Transactional`, so with the save moved into the loop each `holidayDao.save()` now commits in its own transaction (SimpleJpaRepository's own `@Transactional`). Previously the single `saveAll()` at the end made the batch effectively atomic: any unchecked exception thrown during the loop (DataIntegrityViolationException, a NullPointerException while walking `leaveRequest.getEmployee().getUser()`, a DB error in `leaveRequestDao.save`/`leaveEntitlementDao.save`, etc.) meant nothing was persisted, and the client could safely retry. Now: (a) holidays processed before the failure are already committed while later ones are not, so an admin retrying the upload re-inserts the already-created holidays (there is no unique constraint on holiday date+name, and the MAXIMUM_HOLIDAYS_PER_DAY=3 guard only stops the 4th); (b) the holiday is committed *before* `updateLeaveRequestDueHoliday` runs, so a failure inside that method leaves the holiday persisted with the conflicting leave requests neither cancelled/revoked nor their entitlements deducted — a permanently inconsistent state that the response does not report. Fix: annotate the method with `@Transactional(propagation = Propagation.REQUIRED)`, matching `deleteAllHolidays` and `deleteSelectedHolidays` in this same class. The NPE fix still works under a single transaction because `Holiday.id` uses `GenerationType.IDENTITY`, which forces Hibernate to issue the INSERT (and populate the id) at `save()` time. This also removes the N-transactions-per-request overhead that the per-row `save()` introduced versus the previous batched `saveAll()`.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
```diff
@@ -151,16 +151,17 @@ public ResponseEntityDto saveBulkHolidays(HolidayBulkRequestDto holidayBulkReque
...
 
-				savableHolidays.add(holiday);
+				savableHolidays.add(savedHoliday);
 
 				LeaveRequestFilterDto leaveRequestFilterDto = new LeaveRequestFilterDto();
 				leaveRequestFilterDto.setEndDate(holidayDate);
 				leaveRequestFilterDto.setStartDate(holidayDate);
 				List<LeaveRequest> leaveRequests = leaveRequestDao
 					.findAllLeaveRequestsByDateRange(leaveRequestFilterDto);
 				if (systemHolidays.isEmpty() && !leaveRequests.isEmpty()) {
```
ThinuwanW, line 162: 🤖 **Claude** · 🔴 **Important**

Unintended side effect of committing each holiday inside the loop: `holidayDao.findAllByIsActiveTrueAndDate(holidayDate)` (called at the top of the next iteration) now returns holidays inserted by *earlier items of the same bulk payload*, because `Holiday.isActive` defaults to `true` and the row is already committed. Two behaviour changes follow. (1) `countOverlappingHolidays` / the `MAXIMUM_HOLIDAYS_PER_DAY` check now counts intra-request duplicates, so payloads that previously succeeded can now fail with PEOPLE_ERROR_HOLIDAY_MAXIMUM_PER_DAY. (2) More seriously, this line's `systemHolidays.isEmpty()` gate is now false for the 2nd and subsequent holiday on the same date, so conflicting leave requests are silently NOT adjusted for those holidays — e.g. uploading a HALF_DAY_MORNING and a HALF_DAY_EVENING holiday for the same date in one request only processes leave requests for the first. Decide explicitly which semantics you want and encode it: if the gate is meant to mean "no holiday existed on this date before this upload started", capture that up front (e.g. pre-load the existing dates once before the loop, or track a `Set<LocalDate>` of dates already handled in this request) instead of relying on a per-iteration DB read that now sees this request's own writes.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
```diff
@@ -151,16 +151,17 @@ public ResponseEntityDto saveBulkHolidays(HolidayBulkRequestDto holidayBulkReque
 				Set<WorkLocation> workLocations = resolveWorkLocations(holidayDto.getWorkLocations(),
 						workLocationsByName);
 				holiday.setWorkLocations(workLocations);
+				Holiday savedHoliday = holidayDao.save(holiday);
 
-				savableHolidays.add(holiday);
+				savableHolidays.add(savedHoliday);
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`savableHolidays.add(savedHoliday)` happens before the leave-request block, so a record can be counted as both a success and a failure: if `updateLeaveRequestDueHoliday` throws a `ModuleException`, the catch adds the DTO to `holidayDtoStatusList` (ERROR) while the record still contributes to `savableHolidays.size()`, which is passed as `savedRecordsCount` into `getHolidayBulkUploadResponseSummaryText`. The resulting `BulkStatusSummary(savedRecordsCount, inputListCount - savedRecordsCount)` then under-reports `failedCount` while `bulkRecordErrorLogs` lists the record as failed. This existed before, but the PR makes the mismatch materially worse because the holiday row is now genuinely committed at that point. Move `savableHolidays.add(savedHoliday)` to the end of the `try` block, after the leave-request processing, so the success list only contains fully-processed records.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

No test covers the bug being fixed. `HolidayControllerIntegrationTest.saveBulkHolidays_ReturnsCreated` only uploads two holidays on distinct future dates with no conflicting leave requests, so it exercises neither the original failure (`holiday.getId().toString()` NPE in `PeopleNotificationServiceImpl.sendHolidaySingleDayPendingLeaveRequestCancellationEmployeeNotification`, which is not wrapped in a try/catch and therefore surfaced as a 500) nor the new save ordering. Add an integration test that seeds a PENDING (and one APPROVED) leave request on a future date, bulk-uploads a holiday on that same date, and asserts: 201 with successCount 1, the leave request transitioned to CANCELLED/REVOKED, and the created notification's resourceId equals the persisted holiday id. A second case with two holidays on the same date would also pin down the `systemHolidays` behaviour change noted above.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
```diff
@@ -170,18 +171,17 @@ public ResponseEntityDto saveBulkHolidays(HolidayBulkRequestDto holidayBulkReque
 			}
 		});
 
-		List<Holiday> savedHolidays = holidayDao.saveAll(savableHolidays);
-		if (savedHolidays.size() == 1) {
+		if (savableHolidays.size() == 1) {
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

The list is no longer a set of "savable" (pending) holidays — it now holds already-persisted entities. Rename `savableHolidays` to `savedHolidays` (the name freed up by removing the old local) so the variable matches reality and lines up with the `savedRecordsCount` parameter of `getHolidayBulkUploadResponseSummaryText`.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
```diff
@@ -170,18 +171,17 @@ public ResponseEntityDto saveBulkHolidays(HolidayBulkRequestDto holidayBulkReque
...
```
ThinuwanW, line 176: 🤖 **Claude** · 🟣 **Nit**

Pre-existing, but worth confirming while these lines are being touched: the new-holiday declaration email and notification only fire when the batch contains exactly one holiday (`size() == 1`). Any bulk upload of two or more holidays notifies nobody. If that is intentional (avoid mailing everyone N times for a year-long upload), a short comment explaining it would prevent this from being "fixed" incorrectly later; if not, iterate over the saved list (or send one digest email) instead.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
```diff
@@ -151,16 +151,17 @@ public ResponseEntityDto saveBulkHolidays(HolidayBulkRequestDto holidayBulkReque
...
+				savableHolidays.add(savedHoliday);
 
 				LeaveRequestFilterDto leaveRequestFilterDto = new LeaveRequestFilterDto();
 				leaveRequestFilterDto.setEndDate(holidayDate);
 				leaveRequestFilterDto.setStartDate(holidayDate);
 				List<LeaveRequest> leaveRequests = leaveRequestDao
 					.findAllLeaveRequestsByDateRange(leaveRequestFilterDto);
 				if (systemHolidays.isEmpty() && !leaveRequests.isEmpty()) {
-					leaveRequests.forEach(leaveRequest -> updateLeaveRequestDueHoliday(holiday, leaveRequest));
+					leaveRequests.forEach(leaveRequest -> updateLeaveRequestDueHoliday(savedHoliday, leaveRequest));
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

This PR activates `deductDaysUsed` for cancelled/revoked leaves for the first time, and that method over-deducts. Before the fix, `sendEmailsAndNotifications` (called on the line right before the entitlement block in `updateLeaveRequestDueHoliday`) always threw NPE on `holiday.getId()` whenever the status had been set to CANCELLED or REVOKED, so the `leaveRequestEntitlementDao.findAllByLeaveRequestOrderByLeaveEntitlement_ValidToAsc(...)` / `deductDaysUsed(...)` block below it was unreachable in exactly those cases. Now it runs. `deductDaysUsed` unconditionally subtracts 1.0: for a cancelled single-day HALF-day leave, `daysUsed` is 0.5, so it takes the else branch and writes `daysUsed = -0.5` and `totalDaysUsed = totalDaysUsed - 1`, leaving a negative days-used row and an under-counted entitlement balance that never self-corrects. Two fixes needed: (1) deduct the actual amount the leave was reduced by — the full `leaveRequestEntitlement.getDaysUsed()` when the request was fully CANCELLED/REVOKED, otherwise 1.0 for a FULL_DAY holiday and 0.5 for a half-day holiday — and clamp at 0; (2) replace the exact float equality `getDaysUsed() == 1` with a comparison that also covers the fully-consumed case (e.g. delete when the remaining amount would be <= 0). Please also add a test that declares a half-day holiday on a date with a matching half-day leave and asserts the resulting `days_used` / `total_days_used`.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
```diff
@@ -151,16 +151,17 @@ public ResponseEntityDto saveBulkHolidays(HolidayBulkRequestDto holidayBulkReque
...
 				holiday.setWorkLocations(workLocations);
+				Holiday savedHoliday = holidayDao.save(holiday);
 
-				savableHolidays.add(holiday);
+				savableHolidays.add(savedHoliday);
 
 				LeaveRequestFilterDto leaveRequestFilterDto = new LeaveRequestFilterDto();
 				leaveRequestFilterDto.setEndDate(holidayDate);
 				leaveRequestFilterDto.setStartDate(holidayDate);
 				List<LeaveRequest> leaveRequests = leaveRequestDao
```
ThinuwanW, line 160: 🤖 **Claude** · 🔴 **Important**

The leave-request lookup is not scoped by the holiday's work locations, and this PR makes that scoping bug actually take effect end-to-end. `LeaveRequestFilterDto` is populated with only start/end date, so `findAllLeaveRequestsByDateRange` returns every PENDING/APPROVED leave request organisation-wide, and each one is cancelled/revoked or has its duration reduced — even for employees at a work location the holiday does not apply to. This directly contradicts the block at lines 180-181, which deliberately restricts the declaration email/notification audience via `findAllActiveEmployeesExcludingRole(Role.PM_GUEST_EMPLOYEE, workLocationsFilter)`. So a holiday declared for one office notifies only that office but cancels leave for everyone. Fix: after fetching `leaveRequests`, filter to employees whose `getWorkLocation()` is contained in the resolved `workLocations` set (treating an empty set as "all locations", matching `resolveWorkLocations`), or better, add a work-location parameter to the repository query. Note `Employee.workLocation` is a nullable `@ManyToOne`, so the filter must be null-safe.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
```diff
@@ -151,16 +151,17 @@ public ResponseEntityDto saveBulkHolidays(HolidayBulkRequestDto holidayBulkReque
...
 					.findAllLeaveRequestsByDateRange(leaveRequestFilterDto);
 				if (systemHolidays.isEmpty() && !leaveRequests.isEmpty()) {
-					leaveRequests.forEach(leaveRequest -> updateLeaveRequestDueHoliday(holiday, leaveRequest));
+					leaveRequests.forEach(leaveRequest -> updateLeaveRequestDueHoliday(savedHoliday, leaveRequest));
 				}
 			}
 			catch (ModuleException | DateTimeParseException e) {
```
ThinuwanW, line 168: 🤖 **Claude** · 🔴 **Important**

The catch clause only handles `ModuleException | DateTimeParseException`, which was tolerable when persistence happened after the loop but is not any more. Any unchecked exception thrown from the leave-request cascade — e.g. NPE on `leaveRequest.getDurationDays() - 1` in `updateLeaveRequestDueHoliday` (`LeaveRequest.durationDays` is a nullable `Float` and `duration_days` has no NOT NULL constraint), or an NPE/DataAccessException from the email/notification chain — now escapes the lambda and aborts the whole `forEach`. Because the save moved above this block, the holiday for the failing record plus every earlier record is already written, so the client gets a 500 while the DB holds a partially applied batch and `holidayDtoStatusList` records nothing. Previously the same throw meant zero holidays were persisted, which was at least self-consistent. Fix: wrap the leave-request block (lines 158-165) in its own try/catch that catches `RuntimeException`, logs with `log.error("...", e)`, marks that record as `BulkRecordStatus.ERROR`, and undoes the just-saved holiday — or make the per-record unit `@Transactional(propagation = Propagation.REQUIRES_NEW)` so a failure rolls back only that holiday and the loop continues.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Now that the notification chain no longer NPEs, managers will start receiving wrong content from two pre-existing template mix-ups in code this PR activates. (1) In `sendEmailsAndNotifications`, the CANCELLED branch (a PENDING leave) sends the employee the "MultipleDayPending...Employee" message but sends the manager the "MultipleDayApproved...Manager" one, while the REVOKED branch (an APPROVED leave) does the mirror image — the two manager calls are swapped between the branches. (2) In `PeopleNotificationServiceImpl.sendHolidaySingleDayPendingLeaveRequestCancellationManagerNotification`, the template passed to `createNotification` is `PEOPLE_MODULE_HOLIDAY_MULTI_DAY_APPROVED_LEAVE_REQUEST_UPDATED_EMPLOYEE` — a multi-day, approved, employee-facing template for a single-day pending manager notification. Because the email call precedes the notification call and the notification always threw before this fix, none of these manager messages were ever delivered, so this PR is the point at which the mistakes become visible. Please pair each branch's employee/manager calls consistently and correct that template constant. Separately, `peopleNotificationService.sendHolidayMultipleDayApprovedLeaveRequestUpdatedManagerEmail` is a notification-service method whose name ends in `Email` — rename to `...ManagerNotification` to match every sibling method.

ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Missing-parentheses bug in `updateLeaveRequestDueHoliday`'s `isSingleDayHalfDayLeave` computation. It is written as `!isMultiple && state == HALFDAY_MORNING || state == HALFDAY_EVENING`; `&&` binds tighter than `||`, so it evaluates as `(!isMultiple && MORNING) || EVENING`. A MULTI-day leave whose `leaveState` is HALFDAY_EVENING therefore takes the single-day branch and gets fully CANCELLED/REVOKED instead of having 0.5 days deducted. This is reachable: `leaveState` comes straight from the request payload and nothing on the apply-leave path forces FULLDAY when start != end. Fix: wrap the state comparison in parentheses — `!isMultiple && (state == HALFDAY_MORNING || state == HALFDAY_EVENING)`. Given this PR makes the downstream entitlement deduction run too, the blast radius of getting this wrong just grew.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
```diff
@@ -151,16 +151,17 @@ public ResponseEntityDto saveBulkHolidays(HolidayBulkRequestDto holidayBulkReque
...
 
 				LeaveRequestFilterDto leaveRequestFilterDto = new LeaveRequestFilterDto();
 				leaveRequestFilterDto.setEndDate(holidayDate);
 				leaveRequestFilterDto.setStartDate(holidayDate);
 				List<LeaveRequest> leaveRequests = leaveRequestDao
 					.findAllLeaveRequestsByDateRange(leaveRequestFilterDto);
 				if (systemHolidays.isEmpty() && !leaveRequests.isEmpty()) {
-					leaveRequests.forEach(leaveRequest -> updateLeaveRequestDueHoliday(holiday, leaveRequest));
+					leaveRequests.forEach(leaveRequest -> updateLeaveRequestDueHoliday(savedHoliday, leaveRequest));
 				}
```
ThinuwanW, line 164: 🤖 **Claude** · 🟡 **Suggestion**

Behavioural gap now worth closing: for a multi-day leave and a FULL_DAY holiday, `updateLeaveRequestDueHoliday` only decrements `durationDays` and leaves the status as PENDING/APPROVED. `sendEmailsAndNotifications` then matches neither the CANCELLED nor the REVOKED branch, so the employee and manager are never told their leave was shortened by a day — even though `sendHolidayMultipleDayPendingLeaveRequestUpdated*` / `sendHolidayMultipleDayApprovedLeaveRequestUpdated*` exist for precisely this scenario and are currently unreachable for it. Suggest driving `sendEmailsAndNotifications` off what actually changed (cancelled vs. shortened) rather than off the post-mutation status, so the duration-reduction case dispatches the "...Updated" messages.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

This endpoint has other unguarded 500 paths of the same class as the one being fixed, because `HolidayBulkRequestDto` is under-validated. `holidayDtoList` has no annotation at all, so a body omitting it (or sending null) NPEs on `holidayBulkRequestDto.getHolidayDtoList().forEach(...)` at line 140 and returns a 500 rather than a 400. It also lacks `@Valid`, which means the `@NotNull` annotations on `HolidayRequestDto.date`/`name`/`holidayDuration` are never evaluated — all of that validation is silently dead and only the manual checks in `validateHolidayDto` run. And `@NotNull` on the primitive `private int year` can never fail. Fix in `HolidayBulkRequestDto`: annotate the list `@NotEmpty @Valid`, add a `@Size(max = ...)` upper bound since the batch size is currently unbounded (each element now costs a separate commit), and either drop `@NotNull` from `year` or change the field to `Integer`.

ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`savedHoliday` and `holiday` are the same object — `SimpleJpaRepository.save` returns the very instance it was given for a transient entity — so swapping the variable on lines 156 and 164 is a no-op. The only thing that actually fixes the NPE is that `save()` now runs before the leave-request block, because it is what populates the IDENTITY-generated `id` that `PeopleEmailServiceImpl`/`PeopleNotificationServiceImpl` dereference via `holiday.getId().toString()`. Introducing a second name for one object hides that and reads as though two instances exist. Suggest keeping a single `holiday` variable and adding a one-line comment on the save explaining that it must precede `updateLeaveRequestDueHoliday` because the notification services require a non-null holiday ID — otherwise a future cleanup can trivially move it back and reintroduce this exact 500.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
```diff
@@ -151,16 +151,17 @@ public ResponseEntityDto saveBulkHolidays(HolidayBulkRequestDto holidayBulkReque
...
+				Holiday savedHoliday = holidayDao.save(holiday);
 
-				savableHolidays.add(holiday);
+				savableHolidays.add(savedHoliday);
 
 				LeaveRequestFilterDto leaveRequestFilterDto = new LeaveRequestFilterDto();
 				leaveRequestFilterDto.setEndDate(holidayDate);
 				leaveRequestFilterDto.setStartDate(holidayDate);
 				List<LeaveRequest> leaveRequests = leaveRequestDao
 					.findAllLeaveRequestsByDateRange(leaveRequestFilterDto);
```
ThinuwanW, line 161: 🤖 **Claude** · 🟣 **Nit**

Dead condition in the helper this line feeds. `findAllLeaveRequestsByDateRange` hard-codes `status IN (APPROVED, PENDING)` in `LeaveRequestRepositoryImpl` (it delegates to `findLeaveRequestsByDateRange(dto, null)`, which always adds that predicate and ignores any status set on the filter DTO). So the guard `if (status == PENDING || status == APPROVED)` inside `updateLeaveRequestDueHoliday`'s full-day/single-day branch is always true and can be dropped. Worth noting the same root cause makes `leaveRequestFilterDto.setStatus(List.of(PENDING, APPROVED))` in `deleteSelectedHolidays` a no-op — either honour the DTO's status field in the repository or remove it from the DTO so callers stop believing it works.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
SanithuGunasena07: Fixed by adding @Transactional(propagation = Propagation.REQUIRED) to saveBulkHolidays, matching the other methods in this class

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
SanithuGunasena07: Fixed by caching systemHolidays per date (computeIfAbsent) so it's queried once per distinct date and reused for later items in the same payload restores "no holiday existed before this upload" semantics for both the max-per-day check and the leave-conflict gate, regardless of intra-request insert order.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
SanithuGunasena07: not required

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
SanithuGunasena07: moved savableHolidays.add(savedHoliday) to after the leave-request processing, so a record that throws during updateLeaveRequestDueHoliday is no longer counted in both the success and failure buckets.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
SanithuGunasena07: renamed to savedHolidays

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
SanithuGunasena07: not required - pre existing

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
SanithuGunasena07: Pre existing out of scope for PR

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
SanithuGunasena07: Out of scope for PR

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
SanithuGunasena07: not required out scope

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
SanithuGunasena07: Pre Existing - Out of scope for PR

SanithuGunasena07: Pre Existing - Out of scope for PR

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
SanithuGunasena07: Pre Existing - Out of scope for PR

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
SanithuGunasena07: Pre Existing - Out of scope for PR

SanithuGunasena07: renamed back to holiday

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/HolidayServiceImpl.java
SanithuGunasena07: Pre Existing - Out of scope for PR

shakila-kamalasena (approved):

akilarootcode (approved):

PR #1752, opened 2026-08-13
feat: implement CrmDataProvider and integrate it into MyApp component

YasinduRC:
### Summary

-## Summary

Adds `CrmDataProvider`, a root-mounted provider that bootstraps the v2 CRM store
(`useCrmStoreV2`) with lookup data — stages, owners, contacts, task types — the moment
a user navigates under `/crm`, mirroring the pattern PM's `ProjectDataProvider` uses
(route match → gated query → push into the global store) adapted to this app's
Pages Router (no per-record URL key exists here, so the trigger is "are we on a CRM
route" rather than a URL param).

## What's included

- **`CrmDataProvider`** (`v2/providers/CrmDataProvider.tsx`) — mounted once in `_app.tsx`
  inside `TanStackProvider`, gated by `router.pathname.startsWith(ROUTES.CRM.BASE)` so
  the underlying query only runs on CRM pages.
- **`CrmInitDataResponse`** (`v2/types/CrmTypes.ts`) — the v2-shaped bootstrap response
  (`stages`/`owners`/`contacts`/`taskTypes`, all as v2 entities). The current API call
  still hits the existing `useGetBoardInitData` endpoint; the response is cast to this
  target shape rather than mapped field-by-field from the old DTO, since the backend
  will move onto this contract directly.
- **Per-entity record/id-array utils** (`v2/store/utils/crmEntityUtils.ts`) —
  `toStagesRecord`/`toOwnersRecord`/`toContactsRecord`/`toTaskTypesRecord` plus
  `replace*Ids`/`append*Ids` per entity, called at the provider (call site), not
  inside the setters — setters stay plain assigns.

## Deliberately not in this PR

- **No loading/error state.** Checked PM's `ProjectDataProvider` for comparison — it
  explicitly threads `loading`/`error` from Apollo into the store (`setLoading`,
  `setError`) so consumers can render around both states. This PR's provider only
  checks `if (!data) return`, same as "not loaded yet" and "failed to load" being
  indistinguishable. Deferring since nothing currently reads from `useCrmStoreV2` to
  render — adding store-level loading/error now would be state nobody consumes yet.
- **Board-sync-on-stage-change, `moveDeal`, map-merge safety on the paginated
  setters** — all separately tracked from the store-redesign review, not part of this
  provider work.

## Test plan

- [ ] Navigate to `/crm/companies` (or any `/crm/*` route) and confirm
      `useCrmStoreV2.getState().stages/owners/contacts/taskTypes` populate via
      devtools.
- [ ] Navigate to a non-CRM route and confirm the query does not fire
      (`enabled: false` via `isOnCrmRoute`).
- [ ] `tsc --noEmit` and `eslint` clean on all touched files.
### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR mounts a new app-level `CrmDataProvider` in `_app.tsx` that mirrors CRM board init data into a v2 Zustand store. The `_app.tsx` provider-tree restructure itself is behavior-preserving for both the community and enterprise branches, but the new provider is not shippable as-is: it imports three modules (`~community/crm/v2/store/store`, `~community/crm/v2/store/utils/crmEntityUtils`, `~community/crm/v2/types/CrmTypes`) that do not exist anywhere in the repo, so `next build` will fail with module-resolution errors for every page (note `typescript.ignoreBuildErrors` hides type errors but not missing modules). Beyond that, the route guard is checked against `router.pathname`, which is the rewritten page path (`/community/crm/...`), so the query is never enabled; a type assertion hides that `taskTypes` is not part of the board init-data contract; and the loading state can get stuck true.

Found **23** new issue(s): 🔴 8 important, 🟡 10 suggestion(s), 🟣 5 nit(s)

frontend/src/community/crm/v2/providers/CrmDataProvider.tsx
```diff
@@ -0,0 +1,62 @@
+import { useRouter } from "next/router";
+import { FC, PropsWithChildren, useEffect } from "react";
+
+import ROUTES from "~community/common/constants/routes";
+import { useGetBoardInitData } from "~community/crm/api/BoardApi";
+import { useCrmStoreV2 } from "~community/crm/v2/store/store";
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Build-breaking: none of the three v2 modules imported here exist in the repository. `find`/`grep` over `frontend/` shows `src/community/crm/v2/` contains only this file — there is no `v2/store/store.ts`, no `v2/store/utils/crmEntityUtils.ts`, and no `v2/types/CrmTypes.ts` (nor `useCrmStoreV2`, `toStagesRecord`, `replaceStageIds`, `toOwnersRecord`, `toContactsRecord`, `toTaskTypesRecord`, `CrmInitDataResponse` anywhere else). Since `_app.tsx` imports this provider, webpack will fail with `Module not found` on every page build — `next.config.mjs` sets `typescript.ignoreBuildErrors: true`, which suppresses type errors but NOT unresolved module errors. Jest suites that render `_app` will also fail. Either land the v2 store/types/utils files in this PR or hold this PR until they exist. Note `~community/*` has no fallback path in `tsconfig.json` (only `~enterprise/*` does), so nothing rescues these imports.

frontend/src/community/crm/v2/providers/CrmDataProvider.tsx
```diff
@@ -0,0 +1,62 @@
...
+  toContactsRecord,
+  toOwnersRecord,
+  toStagesRecord,
+  toTaskTypesRecord
+} from "~community/crm/v2/store/utils/crmEntityUtils";
+import { CrmInitDataResponse } from "~community/crm/v2/types/CrmTypes";
+
+export const CrmDataProvider: FC<PropsWithChildren> = ({ children }) => {
+  const router = useRouter();
+  const isOnCrmRoute = router.pathname.startsWith(ROUTES.CRM.BASE);
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The CRM route guard will always evaluate to `false`, so the query is never enabled and the store is never populated. CRM pages live at `pages/community/crm/*` and are reached via the rewrites in `next.config.mjs` (`/crm/contacts` → `/community/crm/contacts`, `/crm` → `/community/crm/contacts`, etc.). On the client, `useRouter().pathname` is the resolved *page* route (`/community/crm/contacts`), not the requested URL, so `pathname.startsWith("/crm")` is false. Compare against `router.asPath` instead (this is the pattern already used in `src/community/auth/utils/authUtils.ts:37-41` for public URL prefixes), e.g. derive the guard from `asPath.split("?")[0]`. Note the existing `Drawer.tsx` works only because it uses `pathname.includes(...)`, which incidentally matches the rewritten destination — `startsWith` does not. Also note `middleware.ts:313` legitimately uses `startsWith` because `request.nextUrl.pathname` there is the pre-rewrite URL; that does not apply to the client router.

frontend/src/community/crm/v2/providers/CrmDataProvider.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  const { data, isLoading, error } = useGetBoardInitData(isOnCrmRoute) as {
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

This `as` assertion silences a real contract mismatch instead of fixing the type. `useGetBoardInitData` is declared as `UseQueryResult<CrmBoardInitDataResponse>` and `CrmBoardInitDataResponse` (src/community/crm/types/BoardTypes.ts:4-9) only has `stages`, `contacts`, `owners`, `crmRoles` — there is no `taskTypes`. The backend agrees: `CrmBoardInitDataResponseDto` sets only stages/contacts/crmRoles/owners (`CrmDealServiceImpl.getBoardInitData`), and `taskTypes` is returned by a separate endpoint (`CrmTaskTypeController` → `CrmGetTaskTypeResponseDto`). So `data.taskTypes` is `undefined` at runtime and `toTaskTypesRecord(undefined)` will either throw or silently store an empty record. Fix by removing the assertion (let the hook's generic type flow through) and either (a) extending the backend init-data DTO + `CrmBoardInitDataResponse` with `taskTypes`, or (b) fetching task types from the task-types endpoint. Per the guidelines, type assertions must not be used to paper over an incorrect type.

frontend/src/community/crm/v2/providers/CrmDataProvider.tsx
```diff
@@ -0,0 +1,62 @@
...
+      return;
+    }
+
+    if (error) {
+      setCrmDataError(error.message || "Failed to load CRM data");
+      setCrmDataLoading(false);
+      return;
+    }
+
+    if (data) {
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The effect's state machine has no terminal branch for "query settled but `data` is falsy", so `crmDataLoading` can stay `true` forever with no error set, leaving CRM consumers stuck on a skeleton. Two concrete paths: (1) `fetchBoardInitData` returns `response?.data?.results?.[0]`, which is `undefined` when `results` is empty; (2) the response interceptor in `TanStackProvider.tsx:75-77` returns `undefined` (resolves, does not reject) on HTTP 401, so the query function resolves with `undefined`. Add an `else` branch that clears loading and sets a generic error, e.g. after the `if (data)` block: `else { setCrmDataError(<i18n key>); setCrmDataLoading(false); }` — or invert to `if (!data) { ...clear loading... } else { ...map... }`.

frontend/pages/_app.tsx
```diff
@@ -100,38 +101,45 @@ function MyApp({
           <WebSocketProvider>
             <ToastProvider>
               <TanStackProvider>
+                <CrmDataProvider>
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`CrmDataProvider` is mounted outside the `ErrorBoundary` (and outside `ThemeProvider`/`I18nextProvider`) in both the community and enterprise branches (lines 104 and 125). React error boundaries catch errors thrown from child effects during commit, so if the sync effect throws — very plausible given `data.taskTypes` is `undefined` in the current API contract, see the finding on the type assertion — the whole app renders a blank screen with no `FallbackComponent` and no themed/translated error UI. Move `CrmDataProvider` inside the `ErrorBoundary` (it renders no UI of its own, so it can wrap just `BaseLayout`/`Component`), or at minimum inside `ThemeProvider`+`I18nextProvider` so it can use `t()` for its error messages.

frontend/src/community/crm/v2/providers/CrmDataProvider.tsx
```diff
@@ -0,0 +1,62 @@
...
+  useEffect(() => {
+
+    if (isLoading) {
+      setCrmDataLoading(true);
+      setCrmDataError(null);
+      return;
+    }
+
+    if (error) {
+      setCrmDataError(error.message || "Failed to load CRM data");
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

i18n violation and leaked technical error text. `"Failed to load CRM data"` is a hardcoded English user-facing string, and `error.message` on an Axios error yields raw text like "Request failed with status code 403", which will be shown verbatim wherever `crmDataError` is rendered. Store a translation key (or translate here via `useTranslator`, following the `module.section.key` convention, e.g. `crm.common.failedToLoadCrmData`) rather than the raw message, and don't surface `error.message` to end users. Consider also raising a toast via `useToast` so the failure is visible even if no consumer renders `crmDataError`.

frontend/src/community/crm/v2/providers/CrmDataProvider.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  const {
+    setStages,
+    setStageIds,
+    setOwners,
+    setContacts,
+    setTaskTypes,
+    setCrmDataLoading,
+    setCrmDataError
+  } = useCrmStoreV2((state) => state);
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`useCrmStoreV2((state) => state)` subscribes this provider to the entire store, so it re-renders on every unrelated CRM state change (side panel open/close, table filters, etc.). Select each setter individually (`useCrmStoreV2((s) => s.setStages)` …) or use `useShallow` from `zustand/react/shallow` with an object selector, which is the cheaper and more conventional pattern. Since the store slices are also new in the (missing) v2 store, please also confirm it is created with the `devtools()` wrapper and the `enabled: process.env.NODE_ENV !== "production"` option, matching `src/community/crm/store/store.ts`.

frontend/src/community/crm/v2/providers/CrmDataProvider.tsx
```diff
@@ -0,0 +1,62 @@
...
+    if (data) {
+      setStages(toStagesRecord(data.stages));
+      setStageIds(replaceStageIds(data.stages));
+      setOwners(toOwnersRecord(data.owners));
+      setContacts(toContactsRecord(data.contacts));
+      setTaskTypes(toTaskTypesRecord(data.taskTypes));
+      setCrmDataError(null);
+      setCrmDataLoading(false);
+    }
+  }, [data, isLoading, error, isOnCrmRoute]);
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Dependency array issues: `isOnCrmRoute` is listed but never read inside the effect (dead dependency), while the seven store setters are read but omitted (they are referentially stable in Zustand, but `react-hooks/exhaustive-deps` will flag them — pull them via individual selectors and list them, or add an explicit eslint-disable with a reason). The stray `isOnCrmRoute` dep also has a behavioural side effect: navigating away from CRM re-runs the effect and re-writes the store from the now-stale cached `data`, and re-applies a stale `error` that was raised on the previous CRM visit. Consider also resetting the CRM store when leaving the CRM section.

frontend/src/community/crm/v2/providers/CrmDataProvider.tsx
```diff
@@ -0,0 +1,62 @@
+import { useRouter } from "next/router";
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Architecture: this introduces a parallel `crm/v2/` tree (`v2/providers`, `v2/store`, `v2/types`) alongside the existing `crm/` module, which diverges from the documented module layout (`api/`, `components/`, `store/`, `utils/`, `types/` directly under the module) and mixes v1 (`crm/api/BoardApi`) with v2 state. There is also no barrel `index.tsx` export for the new provider directory, per the component-file convention. Separately, nothing in the codebase reads the v2 store yet, so this provider is effectively dead code that is now pulled into *every* page bundle via `_app.tsx` (all routes, not just CRM). Consider either landing it together with its consumers, or scoping it to the CRM subtree (e.g. wrap only CRM pages / a CRM layout) instead of the global app shell.

frontend/src/community/crm/v2/providers/CrmDataProvider.tsx
```diff
@@ -0,0 +1,62 @@
...
+import {
+  replaceStageIds,
+  toContactsRecord,
+  toOwnersRecord,
+  toStagesRecord,
+  toTaskTypesRecord
+} from "~community/crm/v2/store/utils/crmEntityUtils";
+import { CrmInitDataResponse } from "~community/crm/v2/types/CrmTypes";
+
+export const CrmDataProvider: FC<PropsWithChildren> = ({ children }) => {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

No test file accompanies this new component. Per the testing guideline, add `CrmDataProvider.test.tsx` covering: the route guard enabling/disabling the query (this would have caught the `router.pathname` vs rewritten-path bug), the loading → success mapping into the store, the error path, and the settled-but-no-data path. Use `@testing-library/react` with the `MockTheme` wrapper and a mocked `next/router`.

frontend/src/community/crm/v2/providers/CrmDataProvider.tsx
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`useGetBoardInitData` sets no `staleTime`/`gcTime`, so with the query now mounted for the whole app lifetime (rather than only while the board is mounted) the full board init payload will refetch on every window refocus/reconnect. Consider adding a `staleTime` to the query in `BoardApi.ts`. Also note this hook is already called with default `enabled: true` from `useBoardData`, `useLoadMoreStageDeals` and `useRefreshStageDeals`; those share the `crmBoardQueryKeys.BOARD_INIT_DATA` key so there is no duplicate request, but it does mean the query is effectively always enabled on CRM pages regardless of this provider's guard.

frontend/src/community/crm/v2/providers/CrmDataProvider.tsx
```diff
@@ -0,0 +1,62 @@
...
+      setStageIds(replaceStageIds(data.stages));
+      setOwners(toOwnersRecord(data.owners));
+      setContacts(toContactsRecord(data.contacts));
+      setTaskTypes(toTaskTypesRecord(data.taskTypes));
+      setCrmDataError(null);
+      setCrmDataLoading(false);
+    }
+  }, [data, isLoading, error, isOnCrmRoute]);
+
+  return children;
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Returning `children` bare from an `FC` typed component may not type-check depending on the resolved `@types/react` minor (`FunctionComponent`'s declared return type was `ReactElement | null` before the `ReactNode` widening). Every other provider in this repo (`AuthProvider`, `TanStackProvider`) returns JSX. Safer and consistent: `return <>{children}</>;`. Low confidence on whether it errors with the pinned `@types/react` ^18.3.4, and `next.config.mjs` ignores build type errors anyway — your call.

frontend/src/community/crm/v2/providers/CrmDataProvider.tsx
```diff
@@ -0,0 +1,62 @@
...
+      setCrmDataLoading(false);
+      return;
+    }
+
+    if (data) {
+      setStages(toStagesRecord(data.stages));
+      setStageIds(replaceStageIds(data.stages));
+      setOwners(toOwnersRecord(data.owners));
+      setContacts(toContactsRecord(data.contacts));
+      setTaskTypes(toTaskTypesRecord(data.taskTypes));
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Verified contract mismatch: `data.taskTypes` does not exist on this endpoint. The backend DTO for `/v1/crm/board/init-data` (`CrmBoardInitDataResponseDto`) declares exactly four fields — stages, contacts, crmRoles, owners — and the existing frontend type `CrmBoardInitDataResponse` in `~community/crm/types/BoardTypes` mirrors that. So `toTaskTypesRecord(data.taskTypes)` receives `undefined` on every successful load. If that mapper does anything other than a defensive early return (e.g. `.reduce`/`.forEach` on the argument) it throws inside the effect, and because the provider is mounted above the ErrorBoundary in `_app.tsx` the throw escapes to the root and white-screens the whole app on every page, not just CRM pages. Fix: either drop the taskTypes wiring from this provider and fetch task types from the endpoint that actually serves them, or add `taskTypes` to the backend init-data DTO/service first and ship both sides together. Do not paper over it with `data.taskTypes ?? []`, since that silently stores an empty record and hides the missing paired backend change.

frontend/src/community/crm/v2/providers/CrmDataProvider.tsx
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

📁 **File-level observation**

Nothing reads the store this provider writes, and no existing consumer was migrated. Grepping the frontend for `useCrmStoreV2` returns only this file. Meanwhile the three existing consumers of the same server data — `useBoardData`, `useLoadMoreStageDeals`, `useRefreshStageDeals` — still read stages/owners/contacts straight from the React Query cache via `useGetBoardInitData()`, and CRM pages still read UI state from the original `useCrmStore` slice store. As merged this is a third source of truth for one payload (React Query cache + `useCrmStore` + `useCrmStoreV2`), and it directly hits the review rule against duplicating React Query server cache into Zustand. It also means the new `crmDataLoading` / `crmDataError` state is written but never rendered anywhere, so the error path produces no user-visible feedback at all. Either land this together with the consumer migration (convert `useBoardData` and friends to read from the v2 store, and render `crmDataLoading` / `crmDataError` in the CRM page shell) or hold the provider out of `_app.tsx` until those consumers exist.

frontend/src/community/crm/v2/providers/CrmDataProvider.tsx
```diff
@@ -0,0 +1,62 @@
...
+    setStages,
+    setStageIds,
+    setOwners,
+    setContacts,
+    setTaskTypes,
+    setCrmDataLoading,
+    setCrmDataError
+  } = useCrmStoreV2((state) => state);
+
+  useEffect(() => {
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

No teardown: the store is never reset when the user leaves CRM or signs out. Because the provider is mounted at the app root it never unmounts for the whole session, and the effect only ever writes — there is no cleanup and no branch that clears state when `isOnCrmRoute` goes false. Consequences: (1) the full contacts and owners lists (names, and whatever PII the contact lookup carries) stay resident in the global store while the user browses unrelated modules; (2) `signOut()` (called from ProfileMenu, ContentLayout, and the TanStackProvider refresh-failure path) does a redirect but does not clear Zustand stores, so if a second user signs in without a hard reload the previous user's CRM entities are still in the store and can be rendered before the new fetch resolves; (3) a stale `crmDataError` string set on a CRM page survives forever once the query is disabled again. Fix: add a cleanup branch that resets stages/stageIds/owners/contacts/taskTypes plus the loading and error flags when the route guard is false, and return the same reset from the effect's cleanup function so unmount and sign-out clear it too.

frontend/src/community/crm/v2/providers/CrmDataProvider.tsx
```diff
@@ -0,0 +1,62 @@
+import { useRouter } from "next/router";
+import { FC, PropsWithChildren, useEffect } from "react";
+
+import ROUTES from "~community/common/constants/routes";
+import { useGetBoardInitData } from "~community/crm/api/BoardApi";
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The provider's loading and error state is coupled to a query key it does not own. `useGetBoardInitData` uses the shared key `crmBoardQueryKeys.BOARD_INIT_DATA`, which three existing hooks (`useBoardData`, `useLoadMoreStageDeals`, `useRefreshStageDeals`) also subscribe to. Two consequences worth designing for: a background refetch failure triggered by the deals board will flip the app-wide `crmDataError` even though the provider itself did not ask for anything; and conversely a disabled observer in React Query v5 still reads whatever is already in the cache, so once the board page populates the key, this provider will opportunistically write entities to the store on that page but not on the contacts/companies/tasks pages, making population intermittent and page-dependent rather than deterministic. If this provider is meant to own the lifecycle of that data, give it its own hook/key (or make it the only caller and have the other three hooks read from the store), so ownership of loading/error is unambiguous.

frontend/src/community/crm/v2/providers/CrmDataProvider.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  useEffect(() => {
+
+    if (isLoading) {
+      setCrmDataLoading(true);
+      setCrmDataError(null);
+      return;
+    }
+
+    if (error) {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The error branch is checked before the data branch and returns early, so a failed *refetch* after a successful initial load leaves the store in a contradictory state: the entity records still hold the previously loaded (now possibly stale) values while `crmDataError` is set, and consumers have no way to distinguish "nothing loaded, show error" from "stale data available, show a non-blocking error". Either gate the error branch on the absence of data (only surface a blocking error when there is nothing cached to show) or keep both signals and document that consumers must check data presence before treating the error as fatal.

frontend/src/community/crm/v2/providers/CrmDataProvider.tsx
```diff
@@ -0,0 +1,62 @@
...
+      setCrmDataError(error.message || "Failed to load CRM data");
+      setCrmDataLoading(false);
+      return;
+    }
+
+    if (data) {
+      setStages(toStagesRecord(data.stages));
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

No null/undefined guarding on the response fields before they are handed to the mappers. Even setting aside the missing taskTypes field, the fetcher in BoardApi returns `response?.data?.results?.[0]`, i.e. it indexes into an array that may be empty, and the Java DTO fields are plain collections that can serialize as null if the service leaves one unset. `toStagesRecord`, `toOwnersRecord` and `toContactsRecord` are therefore invoked with values that can be undefined at runtime, inside a provider mounted above the ErrorBoundary. Default each argument (empty array fallback) at the call site, or make the mappers themselves accept an optional argument and return an empty record — and add a matching runtime shape check so a malformed payload degrades to an error state rather than a root-level crash.

frontend/src/community/crm/v2/providers/CrmDataProvider.tsx
```diff
@@ -0,0 +1,62 @@
...
+      setStageIds(replaceStageIds(data.stages));
+      setOwners(toOwnersRecord(data.owners));
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`crmRoles` is dropped. The init-data payload carries a `crmRoles: string[]` field (present in both the backend DTO and the existing `CrmBoardInitDataResponse` type) that is the server's statement of what the current user may do in CRM, but the provider maps only stages/owners/contacts/taskTypes into the store. If this provider is intended to become the single app-wide entry point for CRM bootstrap data, the permission field is exactly the thing consumers will need and its omission will force a second fetch or a fallback to client-side role guessing later. Add a `setCrmRoles` to the v2 store and populate it here.

frontend/src/community/crm/v2/providers/CrmDataProvider.tsx
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The `enabled` predicate is path-only, with no role check, and uses a loose prefix match. Two separate points from the pathname-resolution issue already raised: (1) the endpoint is annotated `@PreAuthorize("hasAnyRole('ROLE_CRM_SALES_REPRESENTATIVE')")` and `middleware.ts` redirects users without `RepresentativeTypes.CRM_SALES_REPRESENTATIVE` away from CRM routes — once the route check is corrected, mirror that role condition into `enabled` (roles are available via the session data hook) so the app never fires a request it knows will 403; (2) `startsWith("/crm")` also matches any future sibling segment such as `/crm-reports` or `/crmadmin`, so prefer an exact match against the known CRM route constants or a segment-aware comparison.

frontend/pages/_app.tsx
```diff
@@ -100,38 +101,45 @@ function MyApp({
...
+                    </I18nextProvider>
+                  </ThemeProvider>
+                </CrmDataProvider>
+              </TanStackProvider>
+            </ToastProvider>
+          </WebSocketProvider>
+        ) : (
+          <ToastProvider>
+            <TanStackProvider>
+              <CrmDataProvider>
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

This PR adds the same provider insertion twice, once per branch of the community/enterprise ternary, deepening an already duplicated six-level provider tree (ToastProvider → TanStackProvider → CrmDataProvider → ThemeProvider → I18nextProvider → ErrorBoundary → BaseLayout, plus ReactQueryDevtools). The two branches now differ only by WebSocketProvider on the outside and AnnouncementProvider/AnnouncementWrapper on the inside, so every future provider change has to be made in two places and will silently drift if one is missed — which is exactly the class of bug the ErrorBoundary-placement issue represents. Extract a shared `AppProviders` component that takes the two branch-specific pieces as props (or renders them conditionally inside a single tree), so the ordering is defined once.

frontend/src/community/crm/v2/providers/CrmDataProvider.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+    if (error) {
+      setCrmDataError(error.message || "Failed to load CRM data");
+      setCrmDataLoading(false);
+      return;
+    }
+
+    if (data) {
+      setStages(toStagesRecord(data.stages));
+      setStageIds(replaceStageIds(data.stages));
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`stageIds` is derived state being persisted into the global store: it is fully computable from the `stages` record the previous line already stores, so the two can now disagree if either is ever updated independently. Prefer a store selector (or a `useMemo` in the consumer) that derives the ordered ids from `stages`. Separately, the helper name `replaceStageIds` reads as a mutating operation on existing state while it is being used as a pure projection of the response — name it for what it returns (e.g. `toStageIds`) to match the sibling `to*Record` mappers.

frontend/src/community/crm/v2/providers/CrmDataProvider.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { useGetBoardInitData } from "~community/crm/api/BoardApi";
+import { useCrmStoreV2 } from "~community/crm/v2/store/store";
+import {
+  replaceStageIds,
+  toContactsRecord,
+  toOwnersRecord,
+  toStagesRecord,
+  toTaskTypesRecord
+} from "~community/crm/v2/store/utils/crmEntityUtils";
+import { CrmInitDataResponse } from "~community/crm/v2/types/CrmTypes";
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`CrmInitDataResponse` in the new v2 types redeclares a contract that already exists as `CrmBoardInitDataResponse` in `~community/crm/types/BoardTypes`, which is the type the API hook itself is declared to return. Having two hand-maintained descriptions of one endpoint response is how the taskTypes divergence flagged above happened, and it will drift again the next time the backend DTO changes. Import and reuse (or extend) the existing type instead of redefining it under v2.

frontend/src/community/crm/v2/providers/CrmDataProvider.tsx
YasinduRC: Removed the setting  stageIds from the provider

frontend/src/community/crm/v2/providers/CrmDataProvider.tsx
YasinduRC: This is intentional since v1 board types will be removed

frontend/src/community/crm/v2/providers/CrmDataProvider.tsx
YasinduRC: CRM roles are not needed to store in the zustand store

frontend/src/community/crm/v2/providers/CrmDataProvider.tsx
YasinduRC: v2/store/store.ts is added

frontend/src/community/crm/v2/providers/CrmDataProvider.tsx
YasinduRC: This was removed

frontend/src/community/crm/v2/providers/CrmDataProvider.tsx
YasinduRC: "as" was  removed  and backend will be updated to send the tasksTypes

frontend/src/community/crm/v2/providers/CrmDataProvider.tsx
YasinduRC: Removed the hard coded error message and added the translated key

frontend/src/community/crm/v2/providers/CrmDataProvider.tsx
YasinduRC: Fixed selected each setter individually

frontend/src/community/crm/v2/providers/CrmDataProvider.tsx
YasinduRC: Intentional, existing store and types will be removed after the refactor

frontend/src/community/crm/v2/providers/CrmDataProvider.tsx
YasinduRC: Added a test file for the provider

frontend/src/community/crm/v2/providers/CrmDataProvider.tsx
YasinduRC: Fixed now it returns  <>{children}</> and updated props as well

frontend/src/community/crm/v2/providers/CrmDataProvider.tsx
YasinduRC: This will be fixed they will be added via a backend PR

frontend/src/community/crm/v2/providers/CrmDataProvider.tsx
YasinduRC: This was removed

frontend/src/community/crm/v2/providers/CrmDataProvider.tsx
YasinduRC: the old store will replace with the v2 one when the refactoring is done

frontend/src/community/crm/v2/providers/CrmDataProvider.tsx
YasinduRC: No staleTime convention exists anywhere in the codebase today, this matches how every other query already behaves, not a regression.

frontend/pages/_app.tsx
YasinduRC: fixed. It now sits inside ErrorBoundary, wrapping only BaseLayout, so it also has theme and translations available

frontend/src/community/crm/v2/providers/CrmDataProvider.tsx
YasinduRC: This was removed with the new approach

frontend/pages/_app.tsx
YasinduRC: This was removed with the new approach

frontend/src/community/crm/v2/providers/CrmDataProvider.tsx
YasinduRC: No need of them since fields are typed as arrays while the Java DTO can serialise them as null

frontend/src/community/crm/v2/providers/CrmDataProvider.tsx
YasinduRC: IsLoading is passed out now and it is handled on the pages itself

frontend/src/community/crm/v2/providers/CrmDataProvider.tsx
YasinduRC: This is intentional. The provider unmounts when leaving CRM, but we keep the store data so re-entry is instant.

On sign-out the store goes with it: signOut() navigates via window.location.href, which is a full page load, and all four call sites use the redirecting default. Since switching users has to go through sign-out, one user's CRM data can't still be in memory for the next one.

frontend/src/community/crm/v2/providers/CrmDataProvider.tsx
YasinduRC: After the refactor, v1 will be removed and replaced with v2, so we can give v2 its own hook and query key at that point without changing v1 behavior in this branch.

frontend/src/community/crm/v2/providers/CrmDataProvider.tsx
YasinduRC: Once the store is initialized with data, we stop checking the initialization error, so a later refetch failure won't override the already-loaded data with a blocking error.

frontend/src/community/crm/v2/hooks/useInitializeCrmData.test.ts
```diff
@@ -0,0 +1,161 @@
+import "@testing-library/jest-dom/extend-expect";
```
thusala: remove tests for this

thusala (changes requested):

frontend/src/community/crm/v2/hooks/useInitializeCrmData.test.ts
YasinduRC: Removed

thusala (dismissed):

hasalarootcode (dismissed):

frontend/pages/community/crm/companies.tsx
```diff
@@ -55,7 +59,11 @@ const Companies: NextPage = () => {
 
         <CompanyModalController />
         <TaskModalController />
-        <CompanyTable />
+        {isCrmInitialDataLoading ? (
```
akilarootcode: remove this new impl from v1

akilarootcode (changes requested):

frontend/pages/community/crm/companies.tsx
YasinduRC: Removed

frontend/src/community/crm/v2/hooks/useInitializeCrmData.ts
```diff
@@ -0,0 +1,67 @@
...
+    setContacts: state.setContacts,
+    setTaskTypes: state.setTaskTypes,
+    setIsCrmDataInitialized: state.setIsCrmDataInitialized
+  }));
+
+  const { data, isLoading, isError, isSuccess } =
+    useGetBoardInitData(!isCrmDataInitialized);
+
+  useEffect(() => {
+    if (isError) {
```
thusala, line 45: do not we need a return function in here? if error occured i believe we do not need to excute below methods could u verify?

thusala (changes requested):

hasalarootcode (approved):

frontend/src/community/crm/v2/hooks/useInitializeCrmData.ts
YasinduRC: No extra return needed here  isSuccess and isError can't both be true at the same time . So if (isCrmDataInitialized || !isSuccess) return; already skips the code below whenever there's an error.

thusala (approved):

akilarootcode (approved):

ThinuwanW (approved):

PR #1753, opened 2026-08-13
Fix/2438 work location delete future holidays

shakila-kamalasena:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR converts work-location deletion from a hard delete to a soft delete (`is_deleted`), filters all read paths through new `*AndIsDeletedFalse` DAO methods, drops the two `ON DELETE CASCADE` FKs that are now unreachable, and deactivates future holidays that are exclusively scoped to the deleted location. The DAO/query filtering is thorough and I found no call site that still reads work locations unfiltered. The main problems are: the H2 test fixture (`data.sql`) does not populate the new NOT NULL column and will break every integration test; the new holiday-deactivation path skips the leave-request guard that the normal holiday-delete path enforces; and join rows in `shr_com_work_location_ppl_holiday` are never cleaned up, so deleted locations keep surfacing in holiday responses and skew the new exclusivity check.

Found **17** new issue(s): 🔴 7 important, 🟡 6 suggestion(s), 🟣 4 nit(s)

backend/src/main/java/com/skapp/community/common/model/WorkLocation.java
```diff
@@ -28,4 +28,7 @@ public class WorkLocation extends Auditable<String> {
 	@Column(name = "address")
 	private String address;
 
+	@Column(name = "is_deleted", nullable = false)
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The new `is_deleted` column is mapped `nullable = false`, and the test profile uses H2 with `spring.jpa.hibernate.ddl-auto: create` (Liquibase is disabled in `src/test/resources/application.yml`). Hibernate will therefore generate `is_deleted BOOLEAN NOT NULL` with no DB-level default, but `src/test/resources/data.sql:81` still inserts `INSERT INTO com_work_location (id, name, address) VALUES ...` without the column. `defer-datasource-initialization: true` runs `data.sql` after schema creation and there is no `spring.sql.init.continue-on-error`, so the insert fails and the Spring context fails to start — breaking `HolidayControllerIntegrationTest` and every other `@SpringBootTest` that depends on `data.sql`. Fix: update `data.sql` to `INSERT INTO com_work_location (id, name, address, is_deleted) VALUES (1, 'Head Office', '123 Main Street', false), (2, 'Branch Office', '456 Second Avenue', false);` (and add a soft-deleted row while you are there, so the new filters get covered).

backend/src/main/java/com/skapp/community/common/service/impl/WorkLocationServiceImpl.java
```diff
@@ -231,6 +238,15 @@ else if (requestDto.getEmployeeIds() != null && !requestDto.getEmployeeIds().isE
 		}
 	}
 
+	private void deleteFutureHolidaysSpecificToWorkLocation(Long workLocationId) {
+		List<Holiday> holidaysToDelete = holidayDao.findFutureActiveHolidaysExclusiveToWorkLocation(workLocationId);
```
ThinuwanW, line 249: 🤖 **Claude** · 🔴 **Important**

`deleteFutureHolidaysSpecificToWorkLocation` deactivates holidays with none of the guards the normal holiday-deletion path applies. `HolidayServiceImpl.deleteSelectedHolidays` refuses to deactivate a holiday when PENDING/APPROVED leave requests exist on that date (`PEOPLE_ERROR_HOLIDAY_CANNOT_BE_DELETED_LEAVES_EXIST`), because declaring the holiday already CANCELLED/REVOKED those requests and deducted entitlement days via `updateLeaveRequestDueHoliday`/`deductDaysUsed`. Deleting a work location now silently removes such a holiday without that check and without restoring the affected leave requests or entitlement balances, so employees permanently lose leave days for a holiday that no longer exists, with no notification. Either reuse the same guard (block the work-location delete, or skip holidays that have leave requests on their date and report that back), or add explicit compensation that restores the leave requests — and document the chosen behaviour.

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/HolidayRepositoryImpl.java
```diff
@@ -116,6 +116,26 @@ public List<Holiday> findAllActiveHolidaysByDateAndWorkLocationId(LocalDate date
...
+		CriteriaQuery<Holiday> criteriaQuery = criteriaBuilder.createQuery(Holiday.class);
+		Root<Holiday> root = criteriaQuery.from(Holiday.class);
+		Join<Holiday, WorkLocation> workLocationJoin = root.join(Holiday_.workLocations, JoinType.INNER);
+
+		List<Predicate> predicates = new ArrayList<>();
+		predicates.add(criteriaBuilder.equal(root.get(Holiday_.isActive), true));
+		predicates.add(criteriaBuilder.greaterThan(root.get(Holiday_.date), DateTimeUtils.getCurrentUtcDate()));
+		predicates.add(criteriaBuilder.equal(workLocationJoin.get(WorkLocation_.workLocationId), workLocationId));
+
+		predicates.add(criteriaBuilder.equal(criteriaBuilder.size(root.get(Holiday_.workLocations)), 1));
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Because the work location is only soft-deleted, its rows in `shr_com_work_location_ppl_holiday` are never removed, and nothing in the read paths filters them out. Two concrete consequences: (1) `PeopleMapper.holidayToHolidayResponseDto` maps `holiday.workLocations` straight through, so `GET /v1/holidays` keeps showing the deleted location's name for every past/current holiday that was scoped to it and for every holiday shared with other locations — while `getAllWorkLocations()` no longer returns it, so it also cannot be filtered on; `HolidayServiceImpl.countOverlappingHolidays` likewise still matches on deleted location names. (2) `criteriaBuilder.equal(criteriaBuilder.size(root.get(Holiday_.workLocations)), 1)` counts join rows including rows pointing at already-soft-deleted locations, so a future holiday scoped to {deleted-WL-A, WL-B} is not deactivated when WL-B is deleted even though it then applies to zero live locations. Cleanest fix: when soft-deleting a work location, also remove it from every holiday's `workLocations` set (delete the join rows) after deactivating the ones that would be left empty; alternatively replace `size(...)` with a subquery that counts only non-deleted locations and filter deleted locations out of the holiday response mapping.

backend/src/main/java/com/skapp/community/common/model/WorkLocation.java
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

`name` is declared `@Column(name = "name", nullable = false, unique = true)`, which is now inconsistent with soft delete: a soft-deleted row keeps occupying its name. Production MySQL happens to be safe because `people-ddl-script-v1-create-table-com-work-location.sql` never created a unique index, but the H2 test schema (`ddl-auto: create`) does enforce it, so the obvious post-change flow — delete "Head Office", then create a new "Head Office" (which `existsByNameIgnoreCaseAndIsDeletedFalse` now allows) — succeeds in production and throws a `DataIntegrityViolationException` (500, not a clean `ModuleException`) under the test schema. Decide the intent explicitly: either drop `unique = true` from the entity to match the real schema, or keep uniqueness and scope it to live rows (composite unique on `(name, is_deleted)` will not work for repeated deletes; a generated/nulled discriminator column or a name-mangling-on-delete strategy is needed).

backend/src/main/java/com/skapp/community/common/controller/v1/WorkLocationController.java
```diff
@@ -91,7 +91,8 @@ public ResponseEntity<ResponseEntityDto> updateWorkLocation(@PathVariable Long i
 	}
 
 	@Operation(summary = "Delete a work location",
-			description = "Deletes a work location along with its employee assignments and geo-fence.")
+			description = "Deletes a work location along with its employee assignments and geo-fence. "
+					+ "Future holidays scoped exclusively to this work location are also permanently removed.")
```
ThinuwanW, line 94: 🤖 **Claude** · 🟡 **Suggestion**

The Swagger description says future holidays are "permanently removed", but the implementation sets `holiday.setActive(false)` — the rows are retained and only deactivated, same as `deleteSelectedHolidays`. The work location itself is no longer removed either, yet the description still says "Deletes a work location along with its employee assignments and geo-fence." Reword to reflect reality, e.g. "Deactivates a work location and unassigns its employees; future holidays scoped exclusively to this location are deactivated." (Note the geo-fence *is* still hard-deleted by the enterprise override, so keep that distinction accurate.)

backend/src/main/java/com/skapp/community/common/service/impl/WorkLocationServiceImpl.java
```diff
@@ -231,6 +238,15 @@ else if (requestDto.getEmployeeIds() != null && !requestDto.getEmployeeIds().isE
 		}
 	}
 
+	private void deleteFutureHolidaysSpecificToWorkLocation(Long workLocationId) {
```
ThinuwanW, line 248: 🤖 **Claude** · 🟡 **Suggestion**

Naming and observability: `deleteFutureHolidaysSpecificToWorkLocation` / `holidaysToDelete` do not delete anything — they deactivate. Rename to `deactivateFutureHolidaysExclusiveToWorkLocation` / `holidaysToDeactivate` so the method name matches the repository method (`findFutureActiveHolidaysExclusiveToWorkLocation`) and the actual effect. Also, this is a destructive side effect of a delete request with no audit trail: add a log line such as `log.info("deactivateFutureHolidaysExclusiveToWorkLocation: deactivated {} holidays for work location {}", holidays.size(), workLocationId)` so operators can reconstruct why a holiday disappeared.

backend/src/main/java/com/skapp/community/common/service/impl/WorkLocationServiceImpl.java
```diff
@@ -8,7 +8,9 @@
 import com.skapp.community.common.payload.response.WorkLocationNameAvailabilityResponseDto;
 import com.skapp.community.common.util.MessageUtil;
 import com.skapp.community.peopleplanner.model.Employee;
+import com.skapp.community.peopleplanner.model.Holiday;
```
ThinuwanW, line 11: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

There is no test coverage for any of this. `src/test/java/com/skapp/community/common/service/impl/` contains only the encryption tests — no `WorkLocationServiceImplTest` exists, and no test was added for the new `findFutureActiveHolidaysExclusiveToWorkLocation` Criteria query or for the soft-delete filtering. At minimum add: a unit test for `deleteWorkLocation` asserting `isDeleted` is set, employees are unassigned and only exclusive future holidays are deactivated; and an integration test covering the exclusivity query (holiday on one location vs. shared across two, past vs. future date), since the `cb.size(...) == 1` predicate is the kind of Criteria construct that is easy to get subtly wrong and impossible to verify with mocks.

backend/src/main/java/com/skapp/community/common/repository/WorkLocationDao.java
```diff
@@ -4,13 +4,20 @@
...
 import java.util.Optional;
 
 @Repository
 public interface WorkLocationDao extends JpaRepository<WorkLocation, Long>, WorkLocationRepository {
 
-	boolean existsByNameIgnoreCase(String name);
+	boolean existsByNameIgnoreCaseAndIsDeletedFalse(String name);
 
-	Optional<WorkLocation> findByNameIgnoreCase(String name);
+	Optional<WorkLocation> findByNameIgnoreCaseAndIsDeletedFalse(String name);
```
ThinuwanW, line 15: 🤖 **Claude** · 🟡 **Suggestion**

The soft-delete predicate is now duplicated in five derived query names here plus two hand-written Criteria predicates in `WorkLocationRepositoryImpl`. Any future query that forgets the suffix silently leaks deleted locations — the exact failure mode this PR had to sweep the whole codebase to avoid. Consider centralising it with Hibernate's `@SQLRestriction("is_deleted = false")` on the `WorkLocation` entity (Spring Boot 4 / Hibernate 6 supports it), which would let `findById`, `existsById`, `findAll` and the Criteria queries keep working unchanged and would make this class much smaller. If you keep the explicit approach, at least add a short comment on the entity stating the invariant so future contributors know the suffix is mandatory.

backend/src/main/java/com/skapp/community/common/repository/impl/WorkLocationRepositoryImpl.java
```diff
@@ -66,6 +67,8 @@ private List<Predicate> buildPredicates(CriteriaBuilder cb, Root<WorkLocation> w
 			WorkLocationFilterDto workLocationFilterDto) {
 		List<Predicate> predicates = new ArrayList<>();
 
+		predicates.add(cb.isFalse(workLocation.get(WorkLocation_.isDeleted)));
```
ThinuwanW, line 68: 🤖 **Claude** · 🟣 **Nit**

Now that `buildPredicates` unconditionally adds the `isDeleted` predicate, it can never return an empty list, so the `if (!predicates.isEmpty())` guards in `findWorkLocations` (line 36) and `getTotalCount` (line 87) are dead conditions. Drop the guards and call `query.where(predicates.toArray(new Predicate[0]))` directly.

backend/src/main/java/com/skapp/community/common/service/impl/WorkLocationServiceImpl.java
```diff
@@ -231,6 +238,15 @@ else if (requestDto.getEmployeeIds() != null && !requestDto.getEmployeeIds().isE
 		}
 	}
 
+	private void deleteFutureHolidaysSpecificToWorkLocation(Long workLocationId) {
+		List<Holiday> holidaysToDelete = holidayDao.findFutureActiveHolidaysExclusiveToWorkLocation(workLocationId);
+
+		if (!holidaysToDelete.isEmpty()) {
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

The `if (!holidaysToDelete.isEmpty())` guard is redundant — `forEach` and `saveAll` are both no-ops on an empty list. The `holidayDao.saveAll(...)` call is also redundant: the method runs inside the caller's `@Transactional`, so the `Holiday` entities are managed and the `setActive(false)` change is flushed at commit anyway. Keeping `saveAll` is harmless and arguably clearer, but the `isEmpty` check can go.

backend/src/main/resources/community/db/changelog/ddl-script/people-ddl-script-v1-alter-table-com-work-location-geofence-drop-work-location-cascade.sql
```diff
@@ -0,0 +1,10 @@
+-- liquibase formatted sql
+
+-- changeset shakila:people-ddl-script-v1-alter-table-com-work-location-geofence-drop-work-location-cascade
+ALTER TABLE `com_work_location_geofence`
+    DROP FOREIGN KEY `FK_com_work_location_geofence_com_work_location_id`;
```
ThinuwanW, line 5: 🤖 **Claude** · 🟣 **Nit**

Both new cascade-drop changesets run `ALTER TABLE ... DROP FOREIGN KEY <name>` unconditionally, which fails hard if the constraint is absent or was created under a different name (the parent tables are created with `CREATE TABLE IF NOT EXISTS`, so a pre-existing table from an older install could differ). Consider adding a Liquibase `preConditions` block with `foreignKeyConstraintExists` and `onFail: MARK_RAN` so the migration is idempotent across environments rather than aborting startup.

backend/src/main/java/com/skapp/community/common/service/impl/WorkLocationServiceImpl.java
```diff
@@ -105,11 +109,14 @@ public ResponseEntityDto updateWorkLocation(Long id, WorkLocationRequestDto work
 	public ResponseEntityDto deleteWorkLocation(Long id) {
 		log.info("deleteWorkLocation: execution started");
 
-		WorkLocation workLocation = workLocationDao.findById(id)
+		WorkLocation workLocation = workLocationDao.findByWorkLocationIdAndIsDeletedFalse(id)
 			.orElseThrow(() -> new ModuleException(CommonMessageConstant.COMMON_ERROR_WORK_LOCATION_NOT_FOUND));
 
+		deleteFutureHolidaysSpecificToWorkLocation(id);
 		clearWorkLocationFromEmployees(id);
```
ThinuwanW, line 123: 🤖 **Claude** · 🔴 **Important**

`clearWorkLocationFromEmployees` only clears employees whose account status is ACTIVE or PENDING, but the delete semantics just changed underneath it. Previously `workLocationDao.delete(workLocation)` hit `FK_employee_work_location` (declared with no ON DELETE action in people-ddl-script-v1-alter-table-employee-add-work-location-id.sql), so any TERMINATED / DEACTIVATED / DELETED employee still pointing at the row made the delete blow up with a DataIntegrityViolationException — loud, but it guaranteed no orphan references ever existed. With the soft delete, that guard is gone: those employees keep `work_location_id` pointing at a row that is now invisible to every read path, and `PeopleReadServiceImpl` (around line 337) still emits that `workLocationId` in the employee response, so the UI shows/sends an ID that no longer resolves. Fix: drop the `accountStatusIn` filter in `clearWorkLocationFromEmployees` (or add a separate sweep for the remaining statuses) so the delete detaches the location from every employee regardless of status. See also the related update-path regression in PeopleServiceImpl.

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
```diff
@@ -441,10 +441,11 @@ private Employee createEmployeeEntity(Employee employee, CreateEmployeeRequestDt
 		if (requestDto != null && requestDto.getEmployment() != null
 				&& requestDto.getEmployment().getEmploymentDetails() != null
 				&& requestDto.getEmployment().getEmploymentDetails().getWorkLocationId() != null) {
-			employee.setWorkLocation(
-					workLocationDao.findById(requestDto.getEmployment().getEmploymentDetails().getWorkLocationId())
-						.orElseThrow(() -> new EntityNotFoundException(
-								PeopleMessageConstant.PEOPLE_ERROR_VALIDATION_WORK_LOCATION_NOT_FOUND)));
+			employee.setWorkLocation(workLocationDao
```
ThinuwanW, line 444: 🤖 **Claude** · 🔴 **Important**

Behavioral drift on the update path. `createEmployeeEntity` is not create-only — `updateEmployee` calls it too (line 355), and so does the guest-conversion flow (line 315). Switching `findById` to `findByWorkLocationIdAndIsDeletedFalse` means any employee that still references a soft-deleted work location (see the stale-reference issue in WorkLocationServiceImpl — non ACTIVE/PENDING employees are never cleared) can no longer be updated at all: the client reads the profile, gets `workLocationId` back from `PeopleReadServiceImpl`, echoes it in the PATCH, and now gets `EntityNotFoundException(PEOPLE_ERROR_VALIDATION_WORK_LOCATION_NOT_FOUND)` for a field they never touched. `Validations.validateWorkLocation` (Validations.java:401, reached via EmployeeValidationServiceImpl:87) fails the same request even earlier with a ValidationException. Fix: either guarantee no employee can reference a deleted location (clear it for all account statuses on delete), or treat the employee's currently-assigned work location as acceptable — e.g. resolve with `findById` and only reject when the id differs from `employee.getWorkLocation()` and the located row is deleted.

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/HolidayRepositoryImpl.java
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The exclusivity predicate `size(workLocations) = 1` counts rows in `shr_com_work_location_ppl_holiday`, and since work locations are only soft-deleted those rows are never removed — so the count includes locations that are already deleted. Concrete failure: holiday H is assigned to locations A and B. Delete A → size is 2, H is skipped (correct, B is still live). Delete B → size is still 2 (the A row is stale), so H is skipped again and stays active even though every location it belongs to is gone. H then applies to nobody (`buildWorkLocationPredicate` only treats an *empty* collection as "all locations"), yet it keeps showing up in the holiday list with two non-existent locations. Fix: make the count ignore deleted locations — replace `cb.size(...)` with a correlated subquery counting the joined `WorkLocation` rows where `isDeleted` is false, or remove the join-table rows for the location as part of the soft delete so `size()` stays truthful.

backend/src/main/java/com/skapp/community/common/service/impl/WorkLocationServiceImpl.java
```diff
@@ -40,6 +42,8 @@ public class WorkLocationServiceImpl implements WorkLocationService {
 
 	private final EmployeeDao employeeDao;
 
+	private final HolidayDao holidayDao;
```
ThinuwanW, line 50: 🤖 **Claude** · 🔴 **Important**

Adding the `holidayDao` field changes the `@RequiredArgsConstructor`-generated constructor signature of `WorkLocationServiceImpl`, and that class is extended by `EpWorkLocationServiceImpl` in the `backend/src/main/java/com/skapp/enterprise` git submodule, which calls `super(workLocationDao, employeeDao, holidayDao, messageUtil)` explicitly. The submodule's own sources also now call the new `findByWorkLocationIdAndIsDeletedFalse` / `findByNameIgnoreCaseAndIsDeletedFalse` methods. This diff contains no submodule pointer bump, so merging it alone leaves the enterprise build unable to compile. Make sure the paired skapp-ep-be-src (and skapp-ep-be-tests) PRs are merged and the submodule pointers are updated in this PR, and call it out in the description so it is not merged independently.

backend/src/main/java/com/skapp/community/common/service/impl/WorkLocationServiceImpl.java
```diff
@@ -8,7 +8,9 @@
 import com.skapp.community.common.payload.response.WorkLocationNameAvailabilityResponseDto;
 import com.skapp.community.common.util.MessageUtil;
 import com.skapp.community.peopleplanner.model.Employee;
+import com.skapp.community.peopleplanner.model.Holiday;
 import com.skapp.community.peopleplanner.repository.EmployeeDao;
+import com.skapp.community.peopleplanner.repository.HolidayDao;
```
ThinuwanW, line 13: 🤖 **Claude** · 🟡 **Suggestion**

The `common` module service now reaches directly into the `peopleplanner` module's repository layer (`HolidayDao` + the `Holiday` entity) to mutate holiday state. `common` is the base module that `peopleplanner` depends on, so this inverts the dependency direction, and per the review guidelines cross-module work should go through a service interface rather than another module's DAO (the existing `EmployeeDao` injection is the same smell, not a justification to add more). Prefer adding a method such as `HolidayService.deactivateFutureHolidaysForWorkLocation(Long workLocationId)` and injecting the interface — that also gives the holiday module a place to apply its own deletion rules (leave-request checks, `canDeleteHoliday`) instead of duplicating/omitting them here.

backend/src/main/java/com/skapp/community/common/controller/v1/WorkLocationController.java
```diff
@@ -91,7 +91,8 @@ public ResponseEntity<ResponseEntityDto> updateWorkLocation(@PathVariable Long i
 	}
 
 	@Operation(summary = "Delete a work location",
-			description = "Deletes a work location along with its employee assignments and geo-fence.")
+			description = "Deletes a work location along with its employee assignments and geo-fence. "
```
ThinuwanW, line 93: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

`jakarta.validation.Valid` is imported at the top of this controller but never applied — neither `createWorkLocation` nor `updateWorkLocation` annotates its `@RequestBody WorkLocationRequestDto` with `@Valid`, so bean validation on the DTO never runs and the name-length/blank checks are only enforced ad hoc in `checkWorkLocationNameExists`. Pre-existing, but worth fixing while this file is open: add `@Valid` to both `@RequestBody` parameters (and ensure `WorkLocationRequestDto.name` carries `@NotBlank` plus a `@Size` matching `CommonConstants.WORK_LOCATION_NAME_MAX_LENGTH`), otherwise remove the unused import.

shakila-kamalasena: We do not use them as a practice. Remove the unused import as well

backend/src/main/java/com/skapp/community/common/controller/v1/WorkLocationController.java
shakila-kamalasena: The current one is correct

backend/src/main/java/com/skapp/community/common/model/WorkLocation.java
shakila-kamalasena: Fixed

shakila-kamalasena: Fixed

backend/src/main/java/com/skapp/community/common/repository/impl/WorkLocationRepositoryImpl.java
shakila-kamalasena: Dropped the guard

backend/src/main/java/com/skapp/community/common/repository/WorkLocationDao.java
shakila-kamalasena: We do not use the @SQLRestriction

backend/src/main/java/com/skapp/community/common/service/impl/WorkLocationServiceImpl.java
shakila-kamalasena: No need tests

backend/src/main/java/com/skapp/community/common/service/impl/WorkLocationServiceImpl.java
shakila-kamalasena: Not needed

backend/src/main/java/com/skapp/community/common/service/impl/WorkLocationServiceImpl.java
shakila-kamalasena: Added it to submodules

backend/src/main/java/com/skapp/community/common/service/impl/WorkLocationServiceImpl.java
shakila-kamalasena: Not related to this PR

backend/src/main/java/com/skapp/community/common/service/impl/WorkLocationServiceImpl.java
shakila-kamalasena: No need

backend/src/main/java/com/skapp/community/common/service/impl/WorkLocationServiceImpl.java
shakila-kamalasena: Fixed

backend/src/main/java/com/skapp/community/common/service/impl/WorkLocationServiceImpl.java
shakila-kamalasena: Removed

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/HolidayRepositoryImpl.java
shakila-kamalasena: This could be misread as an "all work locations" holiday rather than deactivated

shakila-kamalasena: Fixed

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
shakila-kamalasena: Not related to this PR

backend/src/main/resources/community/db/changelog/ddl-script/people-ddl-script-v1-alter-table-com-work-location-geofence-drop-work-location-cascade.sql
shakila-kamalasena: Not an issue

erandi-rootcode (dismissed):

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/HolidayRepositoryImpl.java
```diff
@@ -116,6 +117,32 @@ public List<Holiday> findAllActiveHolidaysByDateAndWorkLocationId(LocalDate date
...
+		CriteriaQuery<Holiday> criteriaQuery = criteriaBuilder.createQuery(Holiday.class);
+		Root<Holiday> root = criteriaQuery.from(Holiday.class);
+		Join<Holiday, WorkLocation> workLocationJoin = root.join(Holiday_.workLocations, JoinType.INNER);
+
+		List<Predicate> predicates = new ArrayList<>();
+		predicates.add(criteriaBuilder.equal(root.get(Holiday_.isActive), true));
+		predicates.add(criteriaBuilder.greaterThan(root.get(Holiday_.date), DateTimeUtils.getCurrentUtcDate()));
+		predicates.add(criteriaBuilder.equal(workLocationJoin.get(WorkLocation_.workLocationId), workLocationId));
+
+		Subquery<Long> liveLocationCount = criteriaQuery.subquery(Long.class);
```
akilarootcode: check why we need live live location

backend/src/main/resources/community/db/changelog/ddl-script/people-ddl-script-v1-alter-table-com-work-location-add-is-deleted.sql
```diff
@@ -0,0 +1,7 @@
+-- liquibase formatted sql
+
+-- changeset shakila:people-ddl-script-v1-alter-table-com-work-location-add-is-deleted
+ALTER TABLE `com_work_location`
+    ADD COLUMN `is_deleted` boolean NOT NULL DEFAULT FALSE;
```
akilarootcode: remove  DEFAULT FALSE

backend/src/main/resources/community/db/changelog/ddl-script/people-ddl-script-v1-alter-table-com-work-location-add-is-deleted.sql
```diff
@@ -0,0 +1,7 @@
+-- liquibase formatted sql
+
+-- changeset shakila:people-ddl-script-v1-alter-table-com-work-location-add-is-deleted
+ALTER TABLE `com_work_location`
```
akilarootcode, line 4: remove default null

akilarootcode (changes requested):

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/HolidayRepositoryImpl.java
shakila-kamalasena: That gets the count of the non deleted work locations. Renamed that as well.

backend/src/main/resources/community/db/changelog/ddl-script/people-ddl-script-v1-alter-table-com-work-location-add-is-deleted.sql
shakila-kamalasena: Done

backend/src/main/resources/community/db/changelog/ddl-script/people-ddl-script-v1-alter-table-com-work-location-add-is-deleted.sql
shakila-kamalasena: Done

akilarootcode (dismissed):

akilarootcode (approved):

PR #1754, opened 2026-08-13
feat(crm): return task types in board init-data endpoint

anusham-rc:
## What

`GET /v1/crm/board/init-data` now returns `taskTypes` alongside the existing `stages`, `contacts`, `crmRoles`, and `owners`, so the kanban board can load task-type options in its initial payload instead of making a separate request.

## Changes

- **`CrmBoardInitDataResponseDto`** — added `List<CrmTaskTypeResponseDto> taskTypes`.
- **`CrmDealServiceImpl.getBoardInitData()`** — injected `CrmTaskTypeDao` and populated `taskTypes`, reusing the existing `crmTaskTypesToCrmTaskTypeResponseDtos` mapper and `findAllByOrderByOrderIndexAscIdAsc()` ordering (identical to the standalone task-types endpoint).
- **`CrmBoardControllerIntegrationTest`** — added `getBoardInitData_ReturnsTaskTypes` (the endpoint previously had no test coverage).

## Testing

`CrmBoardControllerIntegrationTest` — 16 tests, 0 failures. Full `mvn compile` clean.

### DIFF
--- a/backend/src/main/java/com/skapp/community/crmplanner/payload/response/board/CrmBoardInitDataResponseDto.java
+++ b/backend/src/main/java/com/skapp/community/crmplanner/payload/response/board/CrmBoardInitDataResponseDto.java
@@ -1,5 +1,6 @@
 package com.skapp.community.crmplanner.payload.response.board;
 
+import com.skapp.community.crmplanner.payload.response.CrmTaskTypeResponseDto;
 import lombok.Getter;
 import lombok.Setter;
 
@@ -16,5 +17,7 @@ public class CrmBoardInitDataResponseDto {
 	private List<String> crmRoles;
 
 	private List<CrmBoardOwnerResponseDto> owners;
+
+	private List<CrmTaskTypeResponseDto> taskTypes;
 
 }
--- a/backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
@@ -24,6 +24,7 @@ import com.skapp.community.crmplanner.payload.request.CrmDealReorderRequestDto;
 import com.skapp.community.crmplanner.payload.request.board.CrmDealsByStagesRequestDto;
 import com.skapp.community.crmplanner.payload.response.CrmExistsResponseDto;
 import com.skapp.community.crmplanner.payload.response.CrmDealResponseDto;
+import com.skapp.community.crmplanner.payload.response.CrmTaskTypeResponseDto;
 import com.skapp.community.crmplanner.payload.response.board.CrmBoardContactResponseDto;
 import com.skapp.community.crmplanner.payload.response.board.CrmBoardInitDataResponseDto;
 import com.skapp.community.crmplanner.payload.response.board.CrmBoardOwnerResponseDto;
@@ -36,6 +37,7 @@ import com.skapp.community.crmplanner.repository.CrmContactOwnerRepository;
 import com.skapp.community.crmplanner.repository.CrmDealDao;
 import com.skapp.community.crmplanner.repository.CrmDealStageDao;
 import com.skapp.community.crmplanner.repository.CrmTaskDao;
+import com.skapp.community.crmplanner.repository.CrmTaskTypeDao;
 import com.skapp.community.crmplanner.service.CrmDealService;
 import com.skapp.community.crmplanner.service.CrmOwnerResolverService;
 import com.skapp.community.crmplanner.util.CrmUtil;
@@ -78,6 +80,8 @@ public class CrmDealServiceImpl implements CrmDealService {
 	private final CrmOwnerResolverService crmOwnerResolver;
 
 	private final CrmTaskDao crmTaskDao;
+
+	private final CrmTaskTypeDao crmTaskTypeDao;
 
 	private final MessageUtil messageUtil;
 
@@ -264,11 +268,15 @@ public class CrmDealServiceImpl implements CrmDealService {
 					o.getAuthPic()))
 			.toList();
 
+		List<CrmTaskTypeResponseDto> taskTypes = crmMapper
+			.crmTaskTypesToCrmTaskTypeResponseDtos(crmTaskTypeDao.findAllByOrderByOrderIndexAscIdAsc());
+
 		CrmBoardInitDataResponseDto responseDto = new CrmBoardInitDataResponseDto();
 		responseDto.setStages(stages);
 		responseDto.setContacts(contacts);
 		responseDto.setCrmRoles(CrmConstants.ASSIGNABLE_CRM_ROLES.stream().map(Enum::name).sorted().toList());
 		responseDto.setOwners(owners);
+		responseDto.setTaskTypes(taskTypes);
 
 		log.info("getBoardInitData: execution ended");
 		return new ResponseEntityDto(false, responseDto);
--- a/backend/src/test/java/com/skapp/community/crmplanner/controller/v1/CrmBoardControllerIntegrationTest.java
+++ b/backend/src/test/java/com/skapp/community/crmplanner/controller/v1/CrmBoardControllerIntegrationTest.java
@@ -106,6 +106,8 @@ class CrmBoardControllerIntegrationTest {
 
 	private CrmTaskType taskType;
 
+	private CrmTaskType emailTaskType;
+
 	@BeforeEach
 	void setup() {
 		adminToken = jwtService.generateAccessToken(userDetailsService.loadUserByUsername("user1@gmail.com"), 1L);
@@ -134,9 +136,14 @@ class CrmBoardControllerIntegrationTest {
 		contact.setOwner(employeeDao.getReferenceById(1L));
 		crmContactDao.save(contact);
 
+		emailTaskType = new CrmTaskType();
+		emailTaskType.setName("Email");
+		emailTaskType.setOrderIndex(1);
+		emailTaskType = crmTaskTypeDao.save(emailTaskType);
+
 		taskType = new CrmTaskType();
 		taskType.setName("Call");
-		taskType.setOrderIndex(1);
+		taskType.setOrderIndex(2);
 		taskType = crmTaskTypeDao.save(taskType);
 	}
 
@@ -456,11 +463,29 @@ class CrmBoardControllerIntegrationTest {
 			.andDo(print())
 			.andExpect(status().isOk())
 			.andExpect(jsonPath(STATUS_PATH).value(STATUS_SUCCESSFUL))
+			.andExpect(jsonPath(RESULTS_0_PATH + "['stages']").isArray())
+			.andExpect(jsonPath(RESULTS_0_PATH + "['stages']").isNotEmpty())
+			.andExpect(jsonPath(RESULTS_0_PATH + "['contacts']").isArray())
+			.andExpect(jsonPath(RESULTS_0_PATH + "['contacts']").isNotEmpty())
+			.andExpect(jsonPath(RESULTS_0_PATH + "['owners']").isArray())
+			.andExpect(jsonPath(RESULTS_0_PATH + "['crmRoles']").isArray())
 			.andExpect(jsonPath(RESULTS_0_PATH + "['taskTypes']").isArray())
-			.andExpect(jsonPath(RESULTS_0_PATH + "['taskTypes'][?(@.name=='Call')]").exists())
-			.andExpect(jsonPath(RESULTS_0_PATH + "['taskTypes'][0]['id']").value(taskType.getId()))
-			.andExpect(jsonPath(RESULTS_0_PATH + "['taskTypes'][0]['name']").value("Call"))
-			.andExpect(jsonPath(RESULTS_0_PATH + "['taskTypes'][0]['orderIndex']").value(1));
+			.andExpect(jsonPath(RESULTS_0_PATH + "['taskTypes'][0]['id']").value(emailTaskType.getId()))
+			.andExpect(jsonPath(RESULTS_0_PATH + "['taskTypes'][0]['name']").value("Email"))
+			.andExpect(jsonPath(RESULTS_0_PATH + "['taskTypes'][0]['orderIndex']").value(1))
+			.andExpect(jsonPath(RESULTS_0_PATH + "['taskTypes'][1]['id']").value(taskType.getId()))
+			.andExpect(jsonPath(RESULTS_0_PATH + "['taskTypes'][1]['name']").value("Call"))
+			.andExpect(jsonPath(RESULTS_0_PATH + "['taskTypes'][1]['orderIndex']").value(2));
+	}
+
+	@Test
+	@DisplayName("Board init data - without CRM role returns Forbidden")
+	void getBoardInitData_WithoutCrmRole_ReturnsForbidden() throws Exception {
+		String noCrmRoleToken = jwtService.generateAccessToken(userDetailsService.loadUserByUsername("user4@gmail.com"),
+				1L);
+
+		mvc.perform(get("/v1/crm/board/init-data").accept(MediaType.APPLICATION_JSON)
+			.with(SecurityTestUtils.bearerToken(noCrmRoleToken))).andDo(print()).andExpect(status().isForbidden());
 	}
 
 	private CrmTask createTask(CrmDeal deal) {
--- a/backend/src/test/java/com/skapp/community/crmplanner/controller/v1/CrmBoardControllerIntegrationTest.java
+++ b/backend/src/test/java/com/skapp/community/crmplanner/controller/v1/CrmBoardControllerIntegrationTest.java
@@ -48,6 +48,7 @@ import static com.skapp.support.TestConstants.STATUS_UNSUCCESSFUL;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
 import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.patch;
 import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.post;
 import static org.springframework.test.web.servlet.result.MockMvcResultHandlers.print;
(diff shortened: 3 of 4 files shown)


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds `taskTypes` to the CRM board init-data response by injecting `CrmTaskTypeDao` into `CrmDealServiceImpl` and mapping the same query the existing `/v1/crm/task/type` endpoint uses. The change is functionally correct and correctly placed inside the existing `@Transactional(readOnly = true)` method, but adding a `final` field to this `@RequiredArgsConstructor` class changes the generated constructor signature and breaks the enterprise submodule subclass `EpCrmDealServiceImpl`, which calls `super(...)` with the old 11-argument list. Secondary issues: the task-type fetch/map logic is now duplicated verbatim from `CrmTaskTypeServiceImpl` (bypassing `CrmTaskTypeService`), the new integration test does not actually verify the ordering its name claims, and no consumer (frontend type) was updated so the new field is currently unused payload.

Found **8** new issue(s): 🔴 1 important, 🟡 5 suggestion(s), 🟣 2 nit(s)

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
```diff
@@ -79,6 +81,8 @@ public class CrmDealServiceImpl implements CrmDealService {
 
 	private final CrmTaskDao crmTaskDao;
 
+	private final CrmTaskTypeDao crmTaskTypeDao;
```
ThinuwanW, line 84: 🤖 **Claude** · 🔴 **Important**

Breaking change for the enterprise build. `CrmDealServiceImpl` is annotated `@RequiredArgsConstructor`, so adding this `final` field between `crmTaskDao` and `messageUtil` changes the generated constructor from 11 to 12 parameters, with `crmTaskTypeDao` inserted at position 11. The enterprise submodule subclass `backend/src/main/java/com/skapp/enterprise/crmplanner/service/impl/EpCrmDealServiceImpl.java` (submodule `backend-src`, path declared in `.gitmodules`) calls `super(crmDealDao, crmDealStageDao, crmCompanyDao, crmContactDao, crmContactOwnerRepository, crmMapper, pageTransformer, userService, crmOwnerResolver, crmTaskDao, messageUtil)` — 11 args — which will no longer compile. This PR only touches community files, so the enterprise build will fail. Fix by landing a matching enterprise submodule change (add `CrmTaskTypeDao` to `EpCrmDealServiceImpl`'s constructor and pass it to `super(...)`) and bumping the submodule pointer in the same PR. Alternatively, avoid mutating the generated constructor signature — e.g. obtain the task types through the existing `CrmTaskTypeService` bean (see the duplication comment below) if that injection can be added in a way the subclass already forwards, or add an explicit constructor here that keeps the old arity working. Note the same coupling exists for `EpCrmTaskServiceImpl`/`CrmTaskServiceImpl`, so this is a known constraint in this codebase, not a one-off.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
```diff
@@ -264,11 +268,15 @@ public ResponseEntityDto getBoardInitData() {
 					o.getAuthPic()))
 			.toList();
 
+		List<CrmTaskTypeResponseDto> taskTypes = crmMapper
```
ThinuwanW, line 271: 🤖 **Claude** · 🟡 **Suggestion**

Duplicated logic and a layering shortcut. These two lines are byte-for-byte identical to `CrmTaskTypeServiceImpl.getTaskTypes()` (lines 30-31): same DAO method, same mapper method. Per the review guidelines (`2+ methods with near-identical bodies -> extract shared helper`, and prefer going through the service interface rather than reaching into another feature's DAO), `CrmDealServiceImpl` should not inject `CrmTaskTypeDao` at all. Suggested fix: add `List<CrmTaskTypeResponseDto> getTaskTypeDtos();` to `CrmTaskTypeService`, implement it in `CrmTaskTypeServiceImpl` (and have the existing `getTaskTypes()` wrap it into `CrmGetTaskTypeResponseDto`), then inject `CrmTaskTypeService` here and call `crmTaskTypeService.getTaskTypeDtos()`. That keeps the ordering/mapping definition in exactly one place, so future changes (e.g. filtering soft-deleted or tier-restricted task types) can't silently diverge between `/v1/crm/task/type` and `/v1/crm/board/init-data`.

backend/src/test/java/com/skapp/community/crmplanner/controller/v1/CrmBoardControllerIntegrationTest.java
```diff
@@ -447,6 +448,21 @@ private ResultActions performPostDealsByStagesRequest(CrmDealsByStagesRequestDto
 			.with(SecurityTestUtils.bearerToken(token)));
 	}
 
+	@Test
+	@DisplayName("Board init data - returns task types ordered by orderIndex")
```
ThinuwanW, line 459: 🤖 **Claude** · 🟡 **Suggestion**

The test does not verify what its name and `@DisplayName` claim ("ordered by orderIndex"). Only a single `CrmTaskType` ("Call", orderIndex 1) is created in `setup()`, so any ordering — or none at all — would pass. The assertions on `taskTypes[0]` are also positionally brittle: they only hold because `crm_task_type` has no seed rows today (there is no DML changeset and `backend/src/test/resources/data.sql` inserts none). If default task types are ever seeded via Liquibase, or an entity with a null `order_index` is added (MySQL sorts NULLs first, and `CrmTaskType.orderIndex` is nullable), index 0 stops being the row created here and this test fails for reasons unrelated to the code under test. Fix both at once: create a second task type in `setup()` with a lower `orderIndex` (mirroring `CrmTaskTypeControllerIntegrationTest.setup()`, which uses Email=1/Call=2) and assert the relative order of the two known names, or assert against `[?(@.name=='Call')]` projections rather than fixed indices.

backend/src/main/java/com/skapp/community/crmplanner/payload/response/board/CrmBoardInitDataResponseDto.java
```diff
@@ -17,4 +18,6 @@ public class CrmBoardInitDataResponseDto {
 
 	private List<CrmBoardOwnerResponseDto> owners;
 
+	private List<CrmTaskTypeResponseDto> taskTypes;
```
ThinuwanW, line 21: 🤖 **Claude** · 🟡 **Suggestion**

No consumer was updated, so this field is currently unused payload. The frontend response type `frontend/src/community/crm/types/BoardTypes.ts` -> `CrmBoardInitDataResponse` still declares only `stages`, `contacts`, `owners`, `crmRoles`, and the board still fetches task types separately via `useGetTaskTypes()` (`frontend/src/community/crm/api/TaskApi.ts:230`, consumed by `useGetTaskTypeOptions`). If the intent is to remove the extra round-trip on board load, add `taskTypes: CrmTaskTypeEntity[]` to `CrmBoardInitDataResponse` and switch the board's task-type consumers to read it from the init-data query; otherwise every board load now pays for an extra `crm_task_type` query whose result nothing reads.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
```diff
@@ -264,11 +268,15 @@ public ResponseEntityDto getBoardInitData() {
...
+			.crmTaskTypesToCrmTaskTypeResponseDtos(crmTaskTypeDao.findAllByOrderByOrderIndexAscIdAsc());
+
 		CrmBoardInitDataResponseDto responseDto = new CrmBoardInitDataResponseDto();
 		responseDto.setStages(stages);
 		responseDto.setContacts(contacts);
 		responseDto.setCrmRoles(CrmConstants.ASSIGNABLE_CRM_ROLES.stream().map(Enum::name).sorted().toList());
 		responseDto.setOwners(owners);
+		responseDto.setTaskTypes(taskTypes);
```
ThinuwanW, line 279: 🤖 **Claude** · 🟣 **Nit**

The API documentation for this endpoint is now stale. `CrmBoardController.getBoardInitData()` (`backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmBoardController.java:30`) documents the response as "stages, contacts, CRM roles, and owners" — please add task types to that `@Operation(description = ...)` so the generated Swagger contract matches the payload.

backend/src/test/java/com/skapp/community/crmplanner/controller/v1/CrmBoardControllerIntegrationTest.java
```diff
@@ -447,6 +448,21 @@ private ResultActions performPostDealsByStagesRequest(CrmDealsByStagesRequestDto
 			.with(SecurityTestUtils.bearerToken(token)));
 	}
 
+	@Test
```
ThinuwanW, line 458: 🤖 **Claude** · 🟡 **Suggestion**

Coverage gaps for the first test ever written against `/v1/crm/board/init-data`. Two things are missing: (1) no negative-authorization test — the sibling suite `CrmTaskTypeControllerIntegrationTest` has `getTaskTypes_WithoutCrmRole_ReturnsForbidden`, which pins the `@PreAuthorize("hasAnyRole('ROLE_CRM_SALES_REPRESENTATIVE')")` contract; without an equivalent here, a future change to that annotation on `getBoardInitData()` (or an accidental removal) goes undetected. (2) the test asserts only `taskTypes` and ignores the rest of the documented response contract, so it will not catch a regression that drops or renames `stages`, `contacts`, `crmRoles`, or `owners`. Suggested fix: add a second test using a token for a user with no CRM role expecting `status().isForbidden()` (follow the pattern in `CrmTaskTypeControllerIntegrationTest`), and extend this test with `isArray()`/non-empty assertions for `stages`, `contacts`, `crmRoles`, and `owners`.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
ThinuwanW, line 271: 🤖 **Claude** · 🟡 **Suggestion**

Inconsistent data hygiene for the newly exposed collection. Every other collection in this response is filtered: stages come from `findAllByIsDeletedFalseOrderByOrderIndexAsc()` and additionally pass through the overridable `filterVisibleDealStages(...)` hook; contacts come from `findAllContactsForBoardInit()`. The new task-type fetch is an unfiltered `findAll`-style query, because `crm_task_type` is the only CRM table in `common-ddl-script-v1-create-tables.sql` with no `is_deleted` column, and `CrmTaskType` is the only CRM entity that does not extend `Auditable<String>` (contrary to the entity convention — no createdBy/createdDate/lastModified* on a row now surfaced in the board's primary payload). Today the table is only populated by `CrmConfigServiceImpl.setDefaultCrmConfig()`, so nothing breaks, but the moment task types become user-manageable/deactivatable, the board payload will keep serving removed types with no code change here to notice it. Suggested fix: add a Liquibase changeset giving `crm_task_type` the standard audit columns plus `is_deleted`, make `CrmTaskType extend Auditable<String>`, and switch to an `...AndIsDeletedFalse...` derived query (in the shared accessor if the duplication flagged in the first pass is consolidated). If that is deliberately out of scope, note it on the entity so the unfiltered read is an explicit decision rather than an oversight.

backend/src/test/java/com/skapp/community/crmplanner/controller/v1/CrmBoardControllerIntegrationTest.java
```diff
@@ -447,6 +448,21 @@ private ResultActions performPostDealsByStagesRequest(CrmDealsByStagesRequestDto
...
+	void getBoardInitData_ReturnsTaskTypes() throws Exception {
+		mvc.perform(get("/v1/crm/board/init-data").accept(MediaType.APPLICATION_JSON)
+			.with(SecurityTestUtils.bearerToken(repToken)))
+			.andDo(print())
+			.andExpect(status().isOk())
+			.andExpect(jsonPath(STATUS_PATH).value(STATUS_SUCCESSFUL))
+			.andExpect(jsonPath(RESULTS_0_PATH + "['taskTypes']").isArray())
+			.andExpect(jsonPath(RESULTS_0_PATH + "['taskTypes'][?(@.name=='Call')]").exists())
+			.andExpect(jsonPath(RESULTS_0_PATH + "['taskTypes'][0]['id']").value(taskType.getId()))
+			.andExpect(jsonPath(RESULTS_0_PATH + "['taskTypes'][0]['name']").value("Call"))
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Test fixture values do not match the data real clients receive. Task types are seeded by `CrmConfigServiceImpl.setDefaultCrmConfig()` via `DefaultCrmTaskTypeTemplate.getDefaultTaskTypes()`, which persists names from `CrmTaskTypeName.getDisplayName()` — i.e. uppercase `CALL`, `EMAIL`, `MEETING`, `OTHER`. This test asserts `name == "Call"`, a casing that never occurs in a real environment, and the consumer side depends on that casing (`frontend/src/community/crm/hooks/useGetTaskTypeOptions.tsx` builds i18n keys with `category.name.toLowerCase()`). The assertion therefore green-lights a payload shape that differs from production. Suggested fix: build the fixture from the same source of truth — persist `DefaultCrmTaskTypeTemplate.getDefaultTaskTypes()` in `setup()` (or at least use `CrmTaskTypeName.CALL.getDisplayName()`) and assert against that value instead of the literal `"Call"`.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
anusham-rc: Fixed — updated the `@Operation(description = ...)` on `CrmBoardController.getBoardInitData()` to include task types alongside stages, contacts, CRM roles, and owners.

backend/src/test/java/com/skapp/community/crmplanner/controller/v1/CrmBoardControllerIntegrationTest.java
anusham-rc: Fixed — `setup()` now creates two task types (Email/orderIndex 1 and Call/orderIndex 2) and the test asserts their relative order (Email at index 0, Call at index 1), so the ordering is now genuinely exercised instead of passing vacuously on a single row.

backend/src/test/java/com/skapp/community/crmplanner/controller/v1/CrmBoardControllerIntegrationTest.java
anusham-rc: Fixed — added `getBoardInitData_WithoutCrmRole_ReturnsForbidden` (using user4, who has no CRM role) to pin the `@PreAuthorize` contract, and extended the existing test to assert `stages`, `contacts`, `owners`, and `crmRoles` are present (stages/contacts non-empty) so a regression dropping any of them is caught.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
anusham-rc: Leaving this as-is for now, intentionally. The arity concern flagged above has been resolved directly (updating `EpCrmDealServiceImpl`'s constructor), so it's decoupled from this duplication point. The duplicated fetch/map here mirrors `CrmTaskTypeServiceImpl.getTaskTypes()`, but the same query already backs the existing `/v1/crm/task/type` endpoint, so consolidating into a shared `CrmTaskTypeService` method is a low-risk cleanup we're choosing not to bundle into this PR. Noting it as a deliberate decision rather than an oversight.

backend/src/main/java/com/skapp/community/crmplanner/payload/response/board/CrmBoardInitDataResponseDto.java
anusham-rc: Intentional for this PR. The goal here is to land the backend payload first so the board can consume task types from init-data; the frontend wiring (adding `taskTypes` to `CrmBoardInitDataResponse` and switching the board off the separate `useGetTaskTypes()` round-trip) is a follow-up frontend change tracked separately. It's short-lived unused payload, not a permanent extra query.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
anusham-rc: Correct that `crm_task_type` has no `is_deleted` and `CrmTaskType` doesn't extend `Auditable`. That's a pre-existing entity/schema decision spanning the whole task-type feature — the existing `/v1/crm/task/type` endpoint already returns the same unfiltered set via the identical query, so this PR isn't introducing new divergence. Adding audit columns + soft-delete is a Liquibase migration + entity change that's out of scope here; flagging it as an explicit, deliberate decision and leaving it to a dedicated follow-up rather than bolting a schema change onto this PR.

backend/src/test/java/com/skapp/community/crmplanner/controller/v1/CrmBoardControllerIntegrationTest.java
anusham-rc: The test builds its own fixtures (`Email`/`Call`) and asserts against exactly those, so it's self-consistent and green for the right reasons within its own data — it doesn't rely on the default-template seeding at all. Mirroring the production `getDisplayName()` casing (CALL/EMAIL) would only couple the board test to `DefaultCrmTaskTypeTemplate` without adding coverage, so leaving the fixtures as-is.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
anusham-rc: Fixed. `EpCrmDealServiceImpl` in the enterprise submodule now accepts `CrmTaskTypeDao` and forwards it to `super(...)`, matching the new 12-argument community constructor (submodule commit `c79f3ab0`, pointer bumped in this PR). Verified with a clean `mvn clean compile` — BUILD SUCCESS.

akilarootcode (approved):

PR #1755, opened 2026-08-13
Fix/2979 2988 2989 bug fixes

shakila-kamalasena:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR widens Documents-tab visibility in DirectorySteppers, narrows a `useMemo` dependency in IdentificationDetailsSection, and adds a null-safe `isPeopleAdmin(EmployeeRole)` helper to `CommonModuleUtils` (consumed by the enterprise `EnvelopeServiceImpl`). The backend helper is correct and consistent with the existing `isGuestEmployee` style. Two frontend changes need attention: the Documents tab lost its last module/mode gate, which surfaces an empty tab in community builds, and the memo dependency change makes `initialValues` permanently stale for a given employee, breaking Cancel/discard after a save.

Found **11** new issue(s): 🔴 4 important, 🟡 4 suggestion(s), 🟣 3 nit(s)

frontend/src/community/people/components/molecules/DirectorySteppers/DirectorySteppers.tsx
```diff
@@ -97,8 +99,8 @@ const DirectorySteppers = ({
     user?.roles?.includes(EmployeeTypes.ATTENDANCE_EMPLOYEE)
       ? [translateText(["editAllInfo", "timesheet"])]
       : []),
-    ...((isEditView && (isSuperAdmin || (isPeopleAdmin && isESignSender))) ||
-    (isAccountView && user?.roles?.includes(EmployeeTypes.ESIGN_EMPLOYEE))
+    ...(isAccountView ||
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The account-view branch dropped its only availability gate. Previously `isAccountView && user?.roles?.includes(EmployeeTypes.ESIGN_EMPLOYEE)` doubled as a module/mode check: community `AuthorityServiceImpl.addModuleRoleAuthorities` only emits PEOPLE/LEAVE/ATTENDANCE/OKR/CRM authorities (esign is added exclusively by the enterprise `EpAuthorityServiceImpl`, which additionally filters on `moduleConfig.isEsignModule()`), so `ROLE_ESIGN_EMPLOYEE` is never present in a community JWT nor in an enterprise tenant with eSign disabled. With the new `isAccountView || ...` every user now gets a "Documents" step, but in community builds `~enterprise/people/.../IndividualEmployeeDocumentView` resolves to the fallback stub at `src/fallback/people/components/molecules/IndividualEmployeeDocumentView/IndividualEmployeeDocumentView.tsx`, which renders `<></>` — i.e. a blank tab for every community user on /account. The same now applies to the edit-view branch: dropping `isESignSender` means people admins and self-viewers hit the same empty tab in community mode (previously only super admins did). Fix: keep an availability condition alongside the new role logic, e.g. `isEnterpriseMode()` from `~community/common/utils/commonUtil` (or `isEsignatureModuleEnabled` from `useSessionData` if eSign specifically is the gate) ANDed with `isAccountView || (isEditView && (isSuperAdmin || isPeopleAdmin || isOwnProfile))`.

frontend/src/community/people/components/organisms/EmploymentFormSection/SubSections/IdentificationDetailsSection.tsx
```diff
@@ -75,7 +75,7 @@ const IdentificationDetailsSection = forwardRef<FormMethods, Props>(
       () =>
         employee?.employment
           ?.identificationAndDiversityDetails as L3IdentificationAndDiversityDetailsType,
-      [employee]
+      [employee?.common?.employeeId]
```
ThinuwanW, line 78: 🤖 **Claude** · 🔴 **Important**

Keying the `initialValues` memo on `employee?.common?.employeeId` means it recomputes only when the *identity* of the employee changes, never when that employee's data is replaced. `useEditEmployee` has `onSettled: queryClient.invalidateQueries()`, which refetches EMPLOYEE_BY_ID and calls `setEmployee(fresh)` in DirectoryEditSectionWrapper/AccountSectionWrapper — same employeeId, so the memo (and therefore Formik's `initialValues` under `enableReinitialize`) stays pinned to the mount-time snapshot. Concrete failure: open Employment, change TIN A→B, Save (store + `initialEmployee` become B); change TIN to C and hit Cancel → `EmploymentDetailsForm.onCancel` calls `identificationDetailsRef.current.resetForm()`, which restores the stale snapshot A, while `setEmployee(initialEmployee)` puts B in the store — the field now displays a value that is neither the saved nor the current server value, and no further reinit will correct it. Suggested fix: depend on the pristine copy instead, `[initialEmployee?.employment?.identificationAndDiversityDetails]` (pull `initialEmployee` from `usePeopleStore`). `initialEmployee` is only ever written by `setEmployee` (peopleSlice.tsx:145), never by `setEmploymentDetails`, so this still avoids the per-keystroke reinitialization that the current change was clearly trying to eliminate, while resyncing on load, save and discard. Two related notes: (a) the omitted dependency will trip `react-hooks/exhaustive-deps` (configured as "warn") — add an eslint-disable with the rationale if you keep a narrowed key; (b) the sibling sections (EmploymentDetailsSection.tsx:114, SocialMediaDetailsSection.tsx:37, HealthAndOtherDetailsSection.tsx:41, GeneralDetailsSections.tsx:89) still use `[employee]`, so whichever pattern is correct should be applied consistently rather than leaving one section diverging.

frontend/src/community/people/components/molecules/DirectorySteppers/DirectorySteppers.tsx
```diff
@@ -32,7 +32,7 @@ const DirectorySteppers = ({
 
   const { user } = useAuth();
 
-  const { isSuperAdmin, isPeopleAdmin, isESignSender } = useSessionData();
+  const { isSuperAdmin, isPeopleAdmin, userId } = useSessionData();
```
ThinuwanW, line 35: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

`DirectorySteppers.test.tsx` was not updated for the new Documents-tab rules, and it cannot cover them as written: it mocks `next-auth/react`, but the component reads roles via `useAuth()` from `~community/auth/providers/AuthProvider` and the new flags via `useSessionData()`, neither of which is mocked. Add mocks for `~community/common/hooks/useSessionData` and the auth provider, then assert the new matrix — account view shows Documents; edit view shows it for super admin / people admin / own profile and hides it for a people manager viewing someone else's profile.

backend/src/main/java/com/skapp/community/common/util/CommonModuleUtils.java
```diff
@@ -358,4 +359,8 @@ public static boolean isGuestEmployee(Employee employee) {
 				&& Role.PM_GUEST_EMPLOYEE == employee.getEmployeeRole().getPmRole();
 	}
 
+	public static boolean isPeopleAdmin(EmployeeRole employeeRole) {
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

The helper itself is correct and null-safe. Two follow-ups worth considering: (1) the same check is open-coded in several places that would benefit from it, and two of them are not null-safe — `LeaveAnalyticsServiceImpl:1489` and `:1863` do `employeeRole.getPeopleRole().equals(Role.PEOPLE_ADMIN)`, which NPEs when `people_role` is null (it is a nullable column on `employee_role`); `RolesServiceImpl:350` and `:534` are the same comparison. Routing them through `CommonModuleUtils.isPeopleAdmin` would remove the duplication and the NPE. (2) The neighbouring `isGuestEmployee` takes an `Employee` and null-checks `getEmployeeRole()` internally, while this one takes an `EmployeeRole`, pushing that null-check onto callers — and the sole caller doesn't do it consistently: `EnvelopeServiceImpl:862-867` dereferences `role.getIsSuperAdmin()` unguarded right beside the null-safe `isPeopleAdmin(role)` call, so a null `employee_role` still NPEs there. An `isPeopleAdmin(Employee)` overload mirroring `isGuestEmployee` would make the null handling uniform.

frontend/src/community/people/components/molecules/DirectorySteppers/DirectorySteppers.tsx
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Making the Documents tab unconditional in account view turns a latent step-desync bug into an easily reachable one. `currentStep`/`nextStep` live in the shared `usePeopleStore` slice (peopleSlice.tsx:116-117) and are only reset by the store's full reset action — none of AccountSectionWrapper, IndividualSectionWrapper or DirectoryEditSectionWrapper resets them on unmount. Repro: any user opens /account, clicks the newly-always-visible Documents tab (currentStep = EditPeopleFormTypes.documents), then navigates to /people/individual/<id>. IndividualSectionWrapper's effect sees currentStep === nextStep so it leaves the value alone; DirectorySteppers renders the individual-view step list which has no Documents entry (so no tab is highlighted), and PeopleIndividualSection's switch has no `documents` case, so its `default: return;` renders an empty Box — a blank page with no highlighted tab and no obvious recovery. The same stale value on a directory edit page is worse in spirit: PeopleFormSections.tsx:106 still has a `case EditPeopleFormTypes.documents` branch, so IndividualEmployeeDocumentView is rendered for another employee even when this component deliberately hid the tab for that viewer (the backend does enforce document permissions, so it is a UI/state bug rather than a data leak, but the UI gate is bypassed). Fix: after computing `steps`, add an effect that resets to `EditPeopleFormTypes.personal` when the current step is not present in the visible `steps` list, or reset currentStep/nextStep on wrapper unmount.

frontend/src/community/people/components/molecules/DirectorySteppers/DirectorySteppers.tsx
```diff
@@ -55,6 +55,8 @@ const DirectorySteppers = ({
 
   const isEditView = !isIndividualView && !isAccountView;
 
+  const isOwnProfile = userId === employeeId;
```
ThinuwanW, line 59: 🤖 **Claude** · 🔴 **Important**

`isOwnProfile` is now the fourth open-coded copy of the same 'viewer is looking at their own profile' predicate: PeopleFormSections.tsx:36 (`!isPeopleAdmin && userId !== employeeId`), PeopleFormSections.tsx:44 (`isPeopleAdmin && !isSuperAdmin && userId === employeeId`), IndividualEmployeeDocumentView.tsx:37 (`selectedUser === userId`), and EditInfoCard.tsx:115. The EditInfoCard copy is broken: it compares `user?.userId` (typed `number | undefined`, authUtils.ts:67) against `employee?.common?.employeeId` (typed `string | undefined`, PeopleTypes.ts:66). TypeScript does not reject it because both unions contain `undefined`, but the two are never equal when both are set, so `hasTerminationAbility` stays true and a People Admin is offered the Terminate action on their own profile. Please extract a single `useIsOwnProfile(employeeId: number): boolean` hook (or a `isOwnProfile(userId, employeeId)` util) in ~community/people and use it in all four places, fixing EditInfoCard to compare `user?.userId !== Number(employee?.common?.employeeId)` — note SystemPermissionFormSection.tsx:85 already wraps the same field in `Number(...)`, confirming the string/number mismatch.

frontend/src/community/people/components/molecules/DirectorySteppers/DirectorySteppers.tsx
```diff
@@ -97,8 +99,8 @@ const DirectorySteppers = ({
     user?.roles?.includes(EmployeeTypes.ATTENDANCE_EMPLOYEE)
       ? [translateText(["editAllInfo", "timesheet"])]
       : []),
-    ...((isEditView && (isSuperAdmin || (isPeopleAdmin && isESignSender))) ||
-    (isAccountView && user?.roles?.includes(EmployeeTypes.ESIGN_EMPLOYEE))
+    ...(isAccountView ||
+    (isEditView && (isSuperAdmin || isPeopleAdmin || isOwnProfile))
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`isSuperAdmin || isPeopleAdmin` is very likely redundant. Super admins get `peopleRole = Role.PEOPLE_ADMIN` written to their EmployeeRole (RolesServiceImpl.java:517), and AuthorityServiceImpl.addModuleRoleAuthorities expands PEOPLE_ADMIN into ROLE_PEOPLE_ADMIN/ROLE_PEOPLE_MANAGER/ROLE_PEOPLE_EMPLOYEE, so a super admin's JWT already carries ROLE_PEOPLE_ADMIN and `useSessionData().isPeopleAdmin` is true for them. Flagging as a nit rather than a fix because legacy tenants could have a super admin whose peopleRole was never backfilled; if you keep the `isSuperAdmin ||` guard for that reason, a short comment saying so would stop the next reader from 'simplifying' it away.

frontend/src/community/people/components/organisms/EmploymentFormSection/SubSections/IdentificationDetailsSection.tsx
ThinuwanW, line 78: 🤖 **Claude** · 🟡 **Suggestion**

This change leaves the employee form's sub-sections inconsistent with one another. Four sibling sub-sections that follow the identical `useMemo` + `useFormik({ enableReinitialize: true })` pattern still key their `initialValues` on the whole store object: EmploymentDetailsSection.tsx:114, GeneralDetailsSections.tsx:89, HealthAndOtherDetailsSection.tsx:41, and SocialMediaDetailsSection.tsx:37 all use `[employee]`. After this PR, IdentificationDetailsSection is the lone outlier, so two tabs of the same form now have different reinitialization semantics for the same store updates — which will read as accidental to the next maintainer and makes bugs in this area tab-dependent. Whichever dependency turns out to be correct (see the first-pass note on this line), apply it uniformly across all five sub-sections in the same PR, or add a comment explaining why identification details specifically must not reinitialize.

frontend/src/community/people/components/organisms/EmploymentFormSection/SubSections/IdentificationDetailsSection.tsx
```diff
@@ -75,7 +75,7 @@ const IdentificationDetailsSection = forwardRef<FormMethods, Props>(
       () =>
         employee?.employment
           ?.identificationAndDiversityDetails as L3IdentificationAndDiversityDetailsType,
```
ThinuwanW, line 77: 🤖 **Claude** · 🟡 **Suggestion**

The `as L3IdentificationAndDiversityDetailsType` assertion masks the fact that this expression is `undefined` until the employee is loaded into the store, and narrowing the dependency array makes that window more consequential — the memo is now pinned to whatever the value was the first time `employee.common.employeeId` took its current value. Formik is therefore constructed with `initialValues === undefined`, which is why every render site below needs `values?.ssn ?? ""`-style guards. Per the type-safety checklist (no `as` to silence a nullable type), change the memo body to `employee?.employment?.identificationAndDiversityDetails ?? ({} as L3IdentificationAndDiversityDetailsType)` — or better, make the fields on `L3IdentificationAndDiversityDetailsType` optional and return `?? {}` with no assertion, so `values` is always a real object and the optional-chaining guards become provably unnecessary.

backend/src/main/java/com/skapp/community/common/util/CommonModuleUtils.java
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The helper's null guard is defeated by its only call site. In EnvelopeServiceImpl.getAllUserEnvelopesByUserId (line 862) the code does `EmployeeRole role = currentUser.getEmployee().getEmployeeRole();` and then, one line before calling this helper, `Boolean.TRUE.equals(role.getIsSuperAdmin())` — an unguarded dereference of the very reference `isPeopleAdmin` is careful to null-check. The codebase treats `getEmployeeRole()` as genuinely nullable (AuthorityServiceImpl.getAuthorities explicitly returns an empty authority list when it is null), so if that state is ever reachable here the request produces a raw NPE / 500 rather than an ErrorResponse, and the helper's null-safety gives false reassurance. Pick one: either guard once at the call site (`if (role == null) throw new ModuleException(CommonMessageConstant.COMMON_ERROR_UNAUTHORIZED_ACCESS);` before line 866) and drop the redundant check here, or keep the null check here and add a matching null-safe accessor for the super-admin flag. Note also that the sibling `getAllUserEnvelopes` null-checks `currentUser` while this method does not — that asymmetry is harmless today (`UserServiceImpl.getCurrentUser` throws rather than returning null) but is worth making consistent.

backend/src/main/java/com/skapp/community/common/util/CommonModuleUtils.java
```diff
@@ -358,4 +359,8 @@ public static boolean isGuestEmployee(Employee employee) {
...
+		return employeeRole != null && Role.PEOPLE_ADMIN == employeeRole.getPeopleRole();
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

The name over-promises: because super admins are provisioned with `peopleRole = Role.PEOPLE_ADMIN` (RolesServiceImpl.java:517), this returns true for super admins as well as actual people admins, and it deliberately ignores `employeeRole.getIsSuperAdmin()`. It happens to be harmless at the current call site because `isSuperAdmin` is evaluated first, but a future caller writing `if (isPeopleAdmin(role) && !somethingSuperAdminOnly)` will silently include super admins. Consider renaming to `hasPeopleAdminRole` (mirroring the frontend's explicit `isSuperOrPeopleAdmin` in commonUtil.ts:56) or adding a one-line Javadoc stating that super admins satisfy this predicate. Also note there is no CommonModuleUtilsTest anywhere under backend/src/test, so this new public predicate ships with no coverage.

frontend/src/community/people/components/molecules/DirectorySteppers/DirectorySteppers.tsx
shakila-kamalasena: Not an issue

backend/src/main/java/com/skapp/community/common/util/CommonModuleUtils.java
shakila-kamalasena: Not related to this PR

shakila-kamalasena: Not related to this PR

backend/src/main/java/com/skapp/community/common/util/CommonModuleUtils.java
shakila-kamalasena: Not related to this PR

frontend/src/community/people/components/molecules/DirectorySteppers/DirectorySteppers.tsx
shakila-kamalasena: Needed behaviour. Because Documents tab is available for everyone

shakila-kamalasena: Not related to this PR

frontend/src/community/people/components/organisms/EmploymentFormSection/SubSections/IdentificationDetailsSection.tsx
shakila-kamalasena: Not related to this PR

frontend/src/community/people/components/organisms/EmploymentFormSection/SubSections/IdentificationDetailsSection.tsx
shakila-kamalasena: Needed behaviour

shakila-kamalasena: Not related to this PR

frontend/src/community/people/components/molecules/DirectorySteppers/DirectorySteppers.tsx
shakila-kamalasena: Not needed new hook

frontend/src/community/people/components/molecules/DirectorySteppers/DirectorySteppers.tsx
shakila-kamalasena: Dropped that

thusala (dismissed):

backend/src/main/java/com/skapp/community/common/util/CommonModuleUtils.java
akilarootcode: remove this

akilarootcode (changes requested):

akilarootcode (dismissed):

backend/src/main/java/com/skapp/community/common/util/CommonModuleUtils.java
shakila-kamalasena: Removed

akilarootcode (dismissed):

ThinuwanW (dismissed):

akilarootcode (approved):

PR #1756, opened 2026-08-13
fix: align time entry table badges and spacing with design

Deshan1100:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/frontend/pages/community/timesheet/timesheet-requests.tsx
+++ b/frontend/pages/community/timesheet/timesheet-requests.tsx
@@ -39,6 +39,7 @@ const TimesheetRequestsPage: NextPage = () => {
       pageHead={translateText(["timesheet.timeEntryRequests.pageHead"])}
       title={translateText(["timesheet.timeEntryRequests.title"])}
       isDividerVisible={true}
+      dividerStyles={{ paddingBottom: "0rem" }}
       isBackButtonVisible={true}
       onBackClick={router.back}
     >
--- a/frontend/src/community/attendance/components/molecules/EmployeeTimesheetRequestTable/EmployeeTimesheetRequestTable.tsx
+++ b/frontend/src/community/attendance/components/molecules/EmployeeTimesheetRequestTable/EmployeeTimesheetRequestTable.tsx
@@ -225,9 +225,7 @@ const EmployeeTimesheetRequestTable: FC<Props> = ({
               backgroundColor="bg-tertiary-background"
               textColor="text-secondary-text"
             >
-              <span role="img" aria-hidden="true">
-                {requestTypeSelector(timesheetRequest?.status)}
-              </span>
+              <span>{requestTypeSelector(timesheetRequest?.status)}</span>
               {pascalCaseFormatter(timesheetRequest?.status)}
             </Badge>
             <Box sx={classes.kebabMenuBoxStyle}>
--- a/frontend/src/community/attendance/components/molecules/EmployeeTimesheetRequestTable/EmployeeTimesheetRequestTable.tsx
+++ b/frontend/src/community/attendance/components/molecules/EmployeeTimesheetRequestTable/EmployeeTimesheetRequestTable.tsx
@@ -1,5 +1,6 @@
 import { Box, Typography } from "@mui/material";
 import { type Theme, useTheme } from "@mui/material/styles";
+import { Badge } from "@rootcodelabs/skapp-ui";
 import { FC, JSX } from "react";
 
 import { useCancelTimeRequest } from "~community/attendance/api/AttendanceEmployeeApi";
@@ -150,76 +151,85 @@ const EmployeeTimesheetRequestTable: FC<Props> = ({
         ),
         from: (
           <Box sx={classes.outerBoxWrapper}>
-            <Box sx={classes.innerBoxWrapper}>
-              {timesheetRequest?.initialClockIn &&
-                timesheetRequest.requestType ===
-                  TimeSheetRequestTypes.EDIT_RECORD_REQUEST && (
-                  <Typography
-                    variant="body2"
-                    sx={classes.startTimeTextStyles(timesheetRequest)}
-                  >
-                    {timesheetRequest?.initialClockIn}
-                  </Typography>
-                )}
-              {timesheetRequest?.requestedStartTime &&
-                timesheetRequest?.requestedStartTime !==
-                  timesheetRequest?.initialClockIn && (
-                  <Typography variant="body2" sx={classes.errorTextStyles}>
-                    {timesheetRequest?.requestedStartTime}
-                  </Typography>
-                )}
-            </Box>
+            <Badge
+              backgroundColor="bg-tertiary-background"
+              textColor="text-secondary-text"
+            >
+              <Box sx={classes.timeBadgeContentStyles}>
+                {timesheetRequest?.initialClockIn &&
+                  timesheetRequest.requestType ===
+                    TimeSheetRequestTypes.EDIT_RECORD_REQUEST && (
+                    <Typography
+                      variant="body2"
+                      sx={classes.startTimeTextStyles(timesheetRequest)}
+                    >
+                      {timesheetRequest?.initialClockIn}
+                    </Typography>
+                  )}
+                {timesheetRequest?.requestedStartTime &&
+                  timesheetRequest?.requestedStartTime !==
+                    timesheetRequest?.initialClockIn && (
+                    <Typography variant="body2" sx={classes.errorTextStyles}>
+                      {timesheetRequest?.requestedStartTime}
+                    </Typography>
+                  )}
+              </Box>
+            </Badge>
           </Box>
         ),
         to: (
           <Box sx={classes.outerBoxWrapper}>
-            <Box sx={classes.innerBoxWrapper}>
-              {timesheetRequest?.initialClockOut &&
-                timesheetRequest.requestType ===
-                  TimeSheetRequestTypes.EDIT_RECORD_REQUEST && (
-                  <Typography
-                    variant="body2"
-                    sx={classes.endTimeTextStyles(timesheetRequest)}
-                  >
-                    {timesheetRequest?.initialClockOut}
-                  </Typography>
-                )}
-              {timesheetRequest?.requestedEndTime &&
-                timesheetRequest?.requestedEndTime !==
-                  timesheetRequest?.initialClockOut && (
-                  <Typography variant="body2" sx={classes.errorTextStyles}>
-                    {timesheetRequest?.requestedEndTime}
-                  </Typography>
-                )}
-            </Box>
+            <Badge
+              backgroundColor="bg-tertiary-background"
+              textColor="text-secondary-text"
+            >
+              <Box sx={classes.timeBadgeContentStyles}>
+                {timesheetRequest?.initialClockOut &&
+                  timesheetRequest.requestType ===
+                    TimeSheetRequestTypes.EDIT_RECORD_REQUEST && (
+                    <Typography
+                      variant="body2"
+                      sx={classes.endTimeTextStyles(timesheetRequest)}
+                    >
+                      {timesheetRequest?.initialClockOut}
+                    </Typography>
+                  )}
+                {timesheetRequest?.requestedEndTime &&
+                  timesheetRequest?.requestedEndTime !==
+                    timesheetRequest?.initialClockOut && (
+                    <Typography variant="body2" sx={classes.errorTextStyles}>
+                      {timesheetRequest?.requestedEndTime}
+                    </Typography>
+                  )}
+              </Box>
+            </Badge>
           </Box>
         ),
         workedHours: (
           <Box sx={classes.workHoursBoxStyle}>
-            <Typography
-              variant="body2"
-              sx={classes.workHoursTextStyle(timesheetRequest, totalHours)}
+            <Badge
+              backgroundColor={
+                timesheetRequest?.workHours >= totalHours
+                  ? "bg-tertiary-background"
+                  : "bg-semantic-red-background"
+              }
+              textColor="text-secondary-text"
             >
               {formatDuration(timesheetRequest?.workHours)}
-            </Typography>
+            </Badge>
           </Box>
         ),
         status: (
           <Box sx={classes.statusOuterBoxStyles}>
-            <div
-              style={{
-                display: "flex",
-                alignItems: "center",
-                justifyContent: "center",
-                backgroundColor: theme.palette.common.white,
-                borderRadius: "9.375rem",
-                padding: "0.5rem 1rem",
-                gap: "0.5rem"
-              }}
+            <Badge
+              backgroundColor="bg-tertiary-background"
+              textColor="text-secondary-text"
             >
-              <span>{pascalCaseFormatter(timesheetRequest?.status)}</span>
-              {requestTypeSelector(timesheetRequest?.status)}
-            </div>
+              <span role="img" aria-hidden="true">
+                {requestTypeSelector(timesheetRequest?.status)}
+              </span>
+              {pascalCaseFormatter(timesheetRequest?.status)}
+            </Badge>
             <Box sx={classes.kebabMenuBoxStyle}>
               {timesheetRequest?.status === TimeSheetRequestStates.PENDING && (
                 <KebabMenu
--- a/frontend/src/community/attendance/components/molecules/EmployeeTimesheetRequestTable/styles.ts
+++ b/frontend/src/community/attendance/components/molecules/EmployeeTimesheetRequestTable/styles.ts
@@ -89,16 +89,6 @@ const styles = (theme: Theme) => ({
     },
     position: "relative"
   },
-  iconChipStyles: {
-    color: theme.palette.text.secondary,
-    minWidth: "6.875rem",
-    justifyContent: "center",
-    pr: "0rem",
-    [theme.breakpoints.down("lg")]: {
-      minWidth: "0rem",
-      gap: "1rem"
-    }
-  },
   kebabMenuBoxStyle: {
     position: "absolute",
     right: "10%",
--- a/frontend/src/community/attendance/components/molecules/EmployeeTimesheetRequestTable/styles.ts
+++ b/frontend/src/community/attendance/components/molecules/EmployeeTimesheetRequestTable/styles.ts
@@ -25,16 +25,12 @@ const styles = (theme: Theme) => ({
     flex: 1,
     maxWidth: "auto"
   },
-  innerBoxWrapper: {
-    display: "flex",
+  timeBadgeContentStyles: {
+    display: "inline-flex",
     flexDirection: "row",
     alignItems: "center",
     justifyContent: "center",
     gap: "0.5rem",
-    bgcolor: "white",
-    paddingX: "1.25rem",
-    paddingY: "0.625rem",
-    borderRadius: "4rem",
     [theme.breakpoints.down("md")]: {
       flexDirection: "column"
     }
@@ -81,24 +77,6 @@ const styles = (theme: Theme) => ({
     flex: 1,
     maxWidth: "auto"
   },
-  workHoursTextStyle: (
-    timesheetRequest: TimeRequestDataType,
-    totalHours: number
-  ) => ({
-    letterSpacing: "0.03em",
-    color: theme.palette.text.secondary,
-    whiteSpace: "nowrap",
-    textAlign: "center",
-    overflow: "hidden",
-    textOverflow: "ellipsis",
-    backgroundColor:
-      timesheetRequest?.workHours >= totalHours
-        ? theme.palette.grey[200]
-        : theme.palette.error.main,
-    py: "0.25rem",
-    px: "1rem",
-    borderRadius: "4rem"
-  }),
   statusOuterBoxStyles: {
     display: "flex",
     flexDirection: "row",
--- a/frontend/src/community/attendance/components/molecules/ManagerTimesheetRequestTable/ManagerTimesheetRequestTable.tsx
+++ b/frontend/src/community/attendance/components/molecules/ManagerTimesheetRequestTable/ManagerTimesheetRequestTable.tsx
@@ -352,6 +352,7 @@ const ManagerTimesheetRequestTable: FC<Props> = ({
         headers={tableHeaders}
         rows={transformToTableRows()}
         isLoading={isRequestLoading}
+        minHeight="min-h-[390px]"
         emptyState={{
           title: translateText(["emptyRequestTitle"]),
           description: translateText(["emptyRequestDesEmployee"])
--- a/frontend/src/community/attendance/components/molecules/ManagerTimesheetRequestTable/ManagerTimesheetRequestTable.tsx
+++ b/frontend/src/community/attendance/components/molecules/ManagerTimesheetRequestTable/ManagerTimesheetRequestTable.tsx
@@ -1,6 +1,6 @@
 import { Box, IconButton, Typography } from "@mui/material";
 import { type Theme, useTheme } from "@mui/material/styles";
-import { ButtonV2 } from "@rootcodelabs/skapp-ui";
+import { Badge, ButtonV2 } from "@rootcodelabs/skapp-ui";
 import { useRouter } from "next/navigation";
 import { FC, JSX } from "react";
 
@@ -191,58 +191,72 @@ const ManagerTimesheetRequestTable: FC<Props> = ({
       ),
       from: (
         <Box sx={classes.outerBoxWrapper}>
-          <Box sx={classes.innerBoxWrapper}>
-            {timesheetRequest?.initialClockIn &&
-              timesheetRequest.requestType ===
-                TimeSheetRequestTypes.EDIT_RECORD_REQUEST && (
-                <Typography
-                  variant="body2"
-                  sx={classes.startTimeTextStyles(timesheetRequest)}
-                >
-                  {timesheetRequest?.initialClockIn}
-                </Typography>
-              )}
-            {timesheetRequest?.requestedStartTime &&
-              timesheetRequest?.requestedStartTime !==
-                timesheetRequest?.initialClockIn && (
-                <Typography variant="body2" sx={classes.errorTextStyles}>
-                  {timesheetRequest?.requestedStartTime}
-                </Typography>
-              )}
-          </Box>
+          <Badge
+            backgroundColor="bg-tertiary-background"
+            textColor="text-secondary-text"
+          >
+            <Box sx={classes.timeBadgeContentStyles}>
+              {timesheetRequest?.initialClockIn &&
+                timesheetRequest.requestType ===
+                  TimeSheetRequestTypes.EDIT_RECORD_REQUEST && (
+                  <Typography
+                    variant="body2"
+                    sx={classes.startTimeTextStyles(timesheetRequest)}
+                  >
+                    {timesheetRequest?.initialClockIn}
+                  </Typography>
+                )}
+              {timesheetRequest?.requestedStartTime &&
+                timesheetRequest?.requestedStartTime !==
+                  timesheetRequest?.initialClockIn && (
+                  <Typography variant="body2" sx={classes.errorTextStyles}>
+                    {timesheetRequest?.requestedStartTime}
+                  </Typography>
+                )}
+            </Box>
+          </Badge>
         </Box>
       ),
       to: (
         <Box sx={classes.outerBoxWrapper}>
-          <Box sx={classes.innerBoxWrapper}>
-            {timesheetRequest?.initialClockOut &&
-              timesheetRequest.requestType ===
-                TimeSheetRequestTypes.EDIT_RECORD_REQUEST && (
-                <Typography
-                  variant="body2"
-                  sx={classes.endTimeTextStyles(timesheetRequest)}
-                >
-                  {timesheetRequest?.initialClockOut}
-                </Typography>
-              )}
-            {timesheetRequest?.requestedEndTime &&
-              timesheetRequest?.requestedEndTime !==
-                timesheetRequest?.initialClockOut && (
-                <Typography variant="body2" sx={classes.errorTextStyles}>
-                  {timesheetRequest?.requestedEndTime}
-                </Typography>
-              )}
-          </Box>
+          <Badge
+            backgroundColor="bg-tertiary-background"
+            textColor="text-secondary-text"
+          >
+            <Box sx={classes.timeBadgeContentStyles}>
+              {timesheetRequest?.initialClockOut &&
+                timesheetRequest.requestType ===
+                  TimeSheetRequestTypes.EDIT_RECORD_REQUEST && (
+                  <Typography
+                    variant="body2"
+                    sx={classes.endTimeTextStyles(timesheetRequest)}
+                  >
+                    {timesheetRequest?.initialClockOut}
+                  </Typography>
+                )}
+              {timesheetRequest?.requestedEndTime &&
+                timesheetRequest?.requestedEndTime !==
+                  timesheetRequest?.initialClockOut && (
+                  <Typography variant="body2" sx={classes.errorTextStyles}>
+                    {timesheetRequest?.requestedEndTime}
+                  </Typography>
+                )}
+            </Box>
+          </Badge>
         </Box>
       ),
       workedHours: (
         <Box sx={classes.workHoursBoxStyle}>
-          <Typography
-            variant="body2"
-            sx={classes.workHoursTextStyle(timesheetRequest, totalHours)}
+          <Badge
+            backgroundColor={
+              timesheetRequest?.workHours >= totalHours
+                ? "bg-tertiary-background"
+                : "bg-semantic-red-background"
+            }
+            textColor="text-secondary-text"
           >
             {formatDuration(timesheetRequest?.workHours)}
-          </Typography>
+          </Badge>
         </Box>
       ),
       status:
@@ -298,29 +312,15 @@ const ManagerTimesheetRequestTable: FC<Props> = ({
           </>
         ) : (
           <>
-            <Box
-              sx={{
-                display: "flex",
-                flexDirection: "row",
-                justifyContent: "center",
-                gap: "0.5rem",
-                alignItems: "center",
-                backgroundColor: theme.palette.common.white,
-                borderRadius: "9.375rem",
-                margin: "0rem auto",
-                padding: "0.5rem 1rem"
-              }}
+            <Badge
+              backgroundColor="bg-tertiary-background"
+              textColor="text-secondary-text"
             >
-              <Box
-                sx={{
-                  width: "0.5rem",
-                  height: "0.5rem",
-                  borderRadius: "50%",
-                  backgroundColor: theme.palette.success.light
-                }}
-              />
+              <span role="img" aria-hidden="true">
+                {requestTypeSelector(timesheetRequest?.status)}
+              </span>
               {pascalCaseFormatter(timesheetRequest?.status)}
-            </Box>
+            </Badge>
             {timesheetRequest?.status === TimeSheetRequestStates.PENDING && (
               <Box sx={classes.kebabMenuBoxStyle}>
                 <KebabMenu
--- a/frontend/src/community/attendance/components/molecules/TimesheetRequestFilterBody/TimesheetRequestFilterBody.tsx
+++ b/frontend/src/community/attendance/components/molecules/TimesheetRequestFilterBody/TimesheetRequestFilterBody.tsx
@@ -71,13 +71,13 @@ const TimesheetRequestFilterBody: FC<Props> = ({
         "aria-label": translateText(["applyBtnTxt"])
       }}
     >
-      <div className="flex flex-col gap-2">
-        <p className="subtitle1">{translateText(["dateRangeLabel"])}</p>
-        <DateRangePicker
-          value={selectedDateRange}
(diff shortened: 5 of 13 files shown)


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR swaps hand-rolled pill markup in the timesheet request tables for the design-system `Badge` component, adds placeholder i18n keys for the date-range filter, and tweaks spacing in the shared `TableView`. The component-level changes are behaviour-preserving overall, but two shared-surface changes in `TableView` affect all 10 consuming tables, the Manager table's non-pending status cell loses its horizontal centering wrapper, and the badge markup is now duplicated verbatim across two already near-identical components. No correctness/security issues found; findings are mostly styling consistency, a11y, and duplication.

Found **22** new issue(s): 🔴 1 important, 🟡 12 suggestion(s), 🟣 9 nit(s)

frontend/src/community/attendance/components/molecules/ManagerTimesheetRequestTable/ManagerTimesheetRequestTable.tsx
```diff
@@ -298,29 +312,15 @@ const ManagerTimesheetRequestTable: FC<Props> = ({
...
-                flexDirection: "row",
-                justifyContent: "center",
-                gap: "0.5rem",
-                alignItems: "center",
-                backgroundColor: theme.palette.common.white,
-                borderRadius: "9.375rem",
-                margin: "0rem auto",
-                padding: "0.5rem 1rem"
-              }}
+            <Badge
```
ThinuwanW, line 315: 🤖 **Claude** · 🔴 **Important**

The non-pending status cell lost its centering wrapper. The removed `Box` carried `margin: "0rem auto"` plus `display: flex; justifyContent: center`, which is what horizontally centered the pill inside the status column (headers are declared with `align: "center"` at line 166). The `<Badge>` is now rendered as a bare fragment child with no layout wrapper, so approved/denied/cancelled rows will likely left-align while the pending branch (which uses `margin: "0rem 0.75rem 0rem auto"` / `"0rem auto 0rem 0rem"` on its IconButtons) stays centered — a visible misalignment in the very column this PR is meant to tidy. Note the Employee table kept its `classes.statusOuterBoxStyles` wrapper for exactly this purpose. Fix: wrap the `<Badge>` in `<Box sx={classes.statusOuterBoxStyles}>` (or add a `statusBadgeWrapper` style with `display: flex; justifyContent: center`) to match the Employee table.

frontend/src/community/common/components/organisms/TableView/TableView.tsx
```diff
@@ -96,7 +99,7 @@ const TableView: FC<TableViewProps> = ({
         loader={loader}
         emptyState={emptyState}
         onRowClick={onRowClick}
-        className={minHeight}
+        className={isEmptyStateVisible ? EMPTY_STATE_MIN_HEIGHT : minHeight}
```
ThinuwanW, line 102: 🤖 **Claude** · 🟡 **Suggestion**

`minHeight` is now silently ignored whenever the table is empty — `className={isEmptyStateVisible ? EMPTY_STATE_MIN_HEIGHT : minHeight}` discards a caller-supplied `minHeight` and hardcodes `min-h-80`. `TableView` is shared by 10 tables (leave requests, leave types, custom allocations, invoice/customer tables, both timesheet tables), so this is a global behaviour change made from a PR scoped to timesheet badge alignment. No consumer currently passes `minHeight`, so nothing breaks today, but the prop contract is now misleading. Suggest either exposing the empty-state height as an optional prop (e.g. `emptyStateMinHeight?: string` defaulting to `"min-h-80"`) or dropping the special case and just raising the `minHeight` default — whichever matches the design intent.

frontend/src/community/common/components/organisms/TableView/TableView.tsx
```diff
@@ -63,7 +66,7 @@ const TableView: FC<TableViewProps> = ({
 
   return (
     <div className={`flex w-full flex-col gap-3 ${className}`}>
-      {heading && <h2 className="h2 my-4">{heading}</h2>}
+      {heading && <h2 className="h2 mt-4 mb-1">{heading}</h2>}
```
ThinuwanW, line 69: 🤖 **Claude** · 🟡 **Suggestion**

The heading margin change (`my-4` → `mt-4 mb-1`) is another global change to a shared organism: it tightens the heading/table gap for every `TableView` that renders a `heading`, not just the timesheet tables. Please confirm this was verified against the other consumers (LeaveRequests, ManagerLeaveRequest, LeaveTypesTable, PolicyLeaveTypesTable, CustomLeaveAllocationsTable, AssignLeavePolicyForm, InvoiceTable, CustomerTable) — if the tightening is only wanted for the timesheet screens, prefer passing it through `className` from those two callers rather than changing the shared default.

frontend/src/community/attendance/components/molecules/EmployeeTimesheetRequestTable/EmployeeTimesheetRequestTable.tsx
```diff
@@ -150,76 +151,85 @@ const EmployeeTimesheetRequestTable: FC<Props> = ({
...
-                gap: "0.5rem"
-              }}
+            <Badge
+              backgroundColor="bg-tertiary-background"
+              textColor="text-secondary-text"
             >
-              <span>{pascalCaseFormatter(timesheetRequest?.status)}</span>
-              {requestTypeSelector(timesheetRequest?.status)}
-            </div>
+              <span role="img" aria-hidden="true">
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`<span role="img" aria-hidden="true">` is self-contradictory and will likely trip jsx-a11y: `role="img"` promises an image that needs an accessible name (`aria-label`/`aria-labelledby`), while `aria-hidden="true"` removes the element from the accessibility tree entirely. Since the status text right next to it already conveys the meaning, the icon is purely decorative — drop `role="img"` and keep just `<span aria-hidden="true">`. Same issue at ManagerTimesheetRequestTable.tsx:319. Also note `requestTypeSelector` returns `<></>` in its `default` case, so an unrecognised status yields an empty span — harmless once `role="img"` is removed.

frontend/src/community/attendance/components/molecules/EmployeeTimesheetRequestTable/EmployeeTimesheetRequestTable.tsx
```diff
@@ -150,76 +151,85 @@ const EmployeeTimesheetRequestTable: FC<Props> = ({
...
         workedHours: (
           <Box sx={classes.workHoursBoxStyle}>
-            <Typography
-              variant="body2"
-              sx={classes.workHoursTextStyle(timesheetRequest, totalHours)}
+            <Badge
+              backgroundColor={
+                timesheetRequest?.workHours >= totalHours
+                  ? "bg-tertiary-background"
+                  : "bg-semantic-red-background"
```
ThinuwanW, line 214: 🤖 **Claude** · 🟡 **Suggestion**

`bg-semantic-red-background` is paired with `textColor="text-secondary-text"`, which breaks the token pairing used everywhere else in the codebase — see ContactTable.tsx:169-170 and the crm/enterprise constants, which all pair `bg-semantic-red-background` with `text-semantic-red-text`. Contrast is technically fine (#52525b on #fee2e2), but the under-hours badge will read as a neutral-grey number on a pink chip instead of the semantic red the design system defines. Suggest making `textColor` conditional alongside `backgroundColor`: `text-semantic-red-text` when under hours, `text-secondary-text` otherwise. Same at ManagerTimesheetRequestTable.tsx:256.

frontend/src/community/attendance/components/molecules/ManagerTimesheetRequestTable/ManagerTimesheetRequestTable.tsx
```diff
@@ -1,6 +1,6 @@
 import { Box, IconButton, Typography } from "@mui/material";
 import { type Theme, useTheme } from "@mui/material/styles";
-import { ButtonV2 } from "@rootcodelabs/skapp-ui";
+import { Badge, ButtonV2 } from "@rootcodelabs/skapp-ui";
```
ThinuwanW, line 3: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

This PR increases an already-large duplication. `EmployeeTimesheetRequestTable/styles.ts` and `ManagerTimesheetRequestTable/styles.ts` are now byte-for-byte identical (verified with diff), and after this change the `from`, `to` and `workedHours` cell renderers plus `requestTypeSelector` are character-identical across the two components — every one of those blocks was edited twice in this PR. Per the review guideline on duplication (2+ components with near-identical markup → extract a shared component), consider extracting the shared pieces into `~community/attendance/components/molecules/TimesheetRequestTable/` — e.g. a `TimeRangeBadge`, a `WorkedHoursBadge`, a `TimesheetStatusBadge`, and a single shared `styles.ts` — so the next design tweak is a one-file change instead of a two-file copy-paste.

frontend/src/community/attendance/components/molecules/ManagerTimesheetRequestTable/ManagerTimesheetRequestTable.tsx
```diff
@@ -298,29 +312,15 @@ const ManagerTimesheetRequestTable: FC<Props> = ({
...
-                  borderRadius: "50%",
-                  backgroundColor: theme.palette.success.light
-                }}
-              />
+              <span role="img" aria-hidden="true">
+                {requestTypeSelector(timesheetRequest?.status)}
+              </span>
               {pascalCaseFormatter(timesheetRequest?.status)}
-            </Box>
+            </Badge>
```
ThinuwanW, line 323: 🤖 **Claude** · 🟡 **Suggestion**

Dead code in the block this PR just rewrote: this is the `else` branch of `timesheetRequest?.status === TimeSheetRequestStates.PENDING`, so the immediately following `{timesheetRequest?.status === TimeSheetRequestStates.PENDING && (<Box sx={classes.kebabMenuBoxStyle}><KebabMenu .../></Box>)}` (lines 324-339) can never render. It pulls in `getKebabMenuOptions`, `cancelMutate` and `useCancelTimeRequest` that are consequently unused in this component. Since you're already restructuring this branch, either delete the unreachable block (and the now-dead `onSuccess`/`onError`/`getKebabMenuOptions`/`useCancelTimeRequest` wiring), or move it into the PENDING branch if a manager is actually supposed to be able to cancel a pending request.

frontend/src/community/attendance/components/molecules/EmployeeTimesheetRequestTable/EmployeeTimesheetRequestTable.tsx
```diff
@@ -1,5 +1,6 @@
 import { Box, Typography } from "@mui/material";
 import { type Theme, useTheme } from "@mui/material/styles";
+import { Badge } from "@rootcodelabs/skapp-ui";
```
ThinuwanW, line 3: 🤖 **Claude** · 🟣 **Nit**

`Badge` is a new import from `@rootcodelabs/skapp-ui` and is not used anywhere else in the repo, whereas every other pill/chip with this exact `backgroundColor`/`textColor` prop shape uses `Label` (GuestRequestsTable.tsx:130-148, ContactTable.tsx:168-173, PeopleSyncOptionCard.tsx:35, GroupsCell.tsx:7). Please double-check that `Badge` is actually exported by skapp-ui 1.1.1-beta.10 with these props (node_modules isn't checked in, so I couldn't verify — a `tsc --noEmit` run will confirm), and if `Label` is the sanctioned pill primitive, use it here for consistency. Note there is also an unrelated `Badge` stub at `src/fallback/common/components/atoms/Badge/Badge.tsx` that returns an empty fragment — worth confirming no resolution ambiguity.

frontend/src/community/attendance/components/molecules/TimesheetRequestFilterBody/TimesheetRequestFilterBody.tsx
```diff
@@ -71,13 +71,13 @@ const TimesheetRequestFilterBody: FC<Props> = ({
...
     >
-      <div className="flex flex-col gap-2">
-        <p className="subtitle1">{translateText(["dateRangeLabel"])}</p>
-        <DateRangePicker
-          value={selectedDateRange}
-          onChange={(range) => setSelectedDateRange(clampToCurrentYear(range))}
-        />
-      </div>
+      <DateRangePicker
+        label={translateText(["dateRangeLabel"])}
```
ThinuwanW, line 75: 🤖 **Claude** · 🟣 **Nit**

`dateRangeLabel` is `"Date:"` (attendanceModule.json:140) — it carried the trailing colon because it used to be rendered as a free-standing `<p className="subtitle1">`. Now that it's passed as the design-system `label` prop on `DateRangePicker`, the component owns label formatting and may append its own punctuation/required marker, giving "Date::". Suggest changing the value to `"Date"` — but check whether the key is reused elsewhere before editing it.

frontend/src/community/attendance/components/molecules/EmployeeTimesheetRequestTable/styles.ts
```diff
@@ -25,16 +25,12 @@ const styles = (theme: Theme) => ({
     flex: 1,
     maxWidth: "auto"
   },
-  innerBoxWrapper: {
-    display: "flex",
+  timeBadgeContentStyles: {
```
ThinuwanW, line 28: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

While cleaning up this file (removing `innerBoxWrapper` and `workHoursTextStyle`), `iconChipStyles` (line 92) is also dead — it isn't referenced by `EmployeeTimesheetRequestTable.tsx` or anywhere else. Same in `ManagerTimesheetRequestTable/styles.ts`. Worth removing in the same pass.

frontend/src/community/attendance/components/molecules/EmployeeTimesheetRequestTable/EmployeeTimesheetRequestTable.tsx
```diff
@@ -150,76 +151,85 @@ const EmployeeTimesheetRequestTable: FC<Props> = ({
...
+              </Box>
+            </Badge>
           </Box>
         ),
         workedHours: (
           <Box sx={classes.workHoursBoxStyle}>
-            <Typography
-              variant="body2"
-              sx={classes.workHoursTextStyle(timesheetRequest, totalHours)}
+            <Badge
```
ThinuwanW, line 210: 🤖 **Claude** · 🟣 **Nit**

The badge colour here depends on `totalHours`, which defaults to a bare literal `8` in this component's props (line 46), while `ManagerTimesheetRequestTable` uses `DEFAULT_TOTAL_HOURS` from `~community/attendance/constants/constants` for the same default. Since the two tables must colour identically, import and use `DEFAULT_TOTAL_HOURS` here too so the threshold can't drift.

frontend/src/community/attendance/components/molecules/EmployeeTimesheetRequestTable/EmployeeTimesheetRequestTable.tsx
```diff
@@ -150,76 +151,85 @@ const EmployeeTimesheetRequestTable: FC<Props> = ({
...
-                  <Typography variant="body2" sx={classes.errorTextStyles}>
-                    {timesheetRequest?.requestedStartTime}
-                  </Typography>
-                )}
-            </Box>
+            <Badge
+              backgroundColor="bg-tertiary-background"
+              textColor="text-secondary-text"
+            >
+              <Box sx={classes.timeBadgeContentStyles}>
```
ThinuwanW, line 158: 🤖 **Claude** · 🟡 **Suggestion**

An MUI `Box` (renders a `div` by default) is now a direct child of the design-system `Badge` in all four from/to cells (Employee 158, 186; Manager 198, 226). Every other `Badge`/`Label` call site in this repo passes either plain text or a `<span>` (see PeopleSyncStatusLabel, GroupsCell, GuestRequestsTable, BillingSettings). Badge/pill roots in this library are almost certainly inline elements, and a block-level `div` inside a `span` is invalid DOM nesting: React logs a `validateDOMNesting` warning, the browser auto-closes the span so the pill background stops wrapping its content, and under Next.js SSR the server/client trees diverge into a hydration error. The fix is free and strictly safer either way: change to `<Box component="span" sx={classes.timeBadgeContentStyles}>`, or drop the Box entirely and put the flex/gap classes on a `<span className="inline-flex flex-col items-center justify-center gap-2 md:flex-row">`. Worth confirming the Badge root element in @rootcodelabs/skapp-ui@1.1.1-beta.10 before merging.

frontend/src/community/attendance/components/molecules/EmployeeTimesheetRequestTable/EmployeeTimesheetRequestTable.tsx
```diff
@@ -150,76 +151,85 @@ const EmployeeTimesheetRequestTable: FC<Props> = ({
...
+              backgroundColor={
+                timesheetRequest?.workHours >= totalHours
+                  ? "bg-tertiary-background"
+                  : "bg-semantic-red-background"
+              }
+              textColor="text-secondary-text"
             >
               {formatDuration(timesheetRequest?.workHours)}
-            </Typography>
+            </Badge>
```
ThinuwanW, line 219: 🤖 **Claude** · 🟡 **Suggestion**

The worked-hours cell lost its text styling in the rewrite. Previously `formatDuration(...)` was rendered inside `<Typography variant="body2" sx={classes.workHoursTextStyle(...)}>`, which supplied `letterSpacing: 0.03em`, `whiteSpace: "nowrap"`, `overflow: hidden` and `textOverflow: ellipsis`. Now the formatted duration is a bare string child of `Badge` (same at ManagerTimesheetRequestTable:259), and the deleted `workHoursTextStyle` was the only place those rules lived. Unless Badge itself applies `whitespace-nowrap`, a value like "8h 30m" can wrap onto two lines inside the pill on narrow viewports — and the font will no longer match the `body2` used by the neighbouring date/from/to cells. Follow the established pattern for Badge text content in this repo (BillingSettings, GuestRequestsTable) and wrap it: `<span className="body2 whitespace-nowrap">{formatDuration(timesheetRequest?.workHours)}</span>`.

frontend/src/community/attendance/components/molecules/EmployeeTimesheetRequestTable/EmployeeTimesheetRequestTable.tsx
```diff
@@ -150,76 +151,85 @@ const EmployeeTimesheetRequestTable: FC<Props> = ({
...
           </Box>
         ),
         workedHours: (
           <Box sx={classes.workHoursBoxStyle}>
-            <Typography
-              variant="body2"
-              sx={classes.workHoursTextStyle(timesheetRequest, totalHours)}
+            <Badge
+              backgroundColor={
```
ThinuwanW, line 211: 🤖 **Claude** · 🟡 **Suggestion**

The worked-hours shortfall is now signalled by background colour alone, with no text or accessible alternative — a screen-reader or colour-blind user sees only "7h 30m" with no indication it is under the expected daily hours. The old code had the same limitation, but this cell is being rewritten here so it is the right moment to fix it: add an accessible description to the Badge, e.g. `aria-label={timesheetRequest?.workHours >= totalHours ? formatDuration(timesheetRequest?.workHours) : translateText(["belowExpectedHoursAria"], { hours: formatDuration(timesheetRequest?.workHours), expected: totalHours })}` (plus the new key in `attendanceAria`), or add a small visible indicator icon next to the value. Same applies at ManagerTimesheetRequestTable:250-259.

frontend/src/community/attendance/components/molecules/EmployeeTimesheetRequestTable/styles.ts
```diff
@@ -25,16 +25,12 @@ const styles = (theme: Theme) => ({
...
+    display: "inline-flex",
```
ThinuwanW, line 29: 🤖 **Claude** · 🟡 **Suggestion**

`timeBadgeContentStyles` keeps the `theme.breakpoints.down("md")` → `flexDirection: "column"` rule, but the vertical padding and radius that used to make the pill grow with its content (`paddingY: "0.625rem"`, `paddingX: "1.25rem"`, `borderRadius: "4rem"`) were removed and are now owned by `Badge`. Below `md` the initial and requested times stack vertically inside a badge whose padding/height this style no longer controls, so the two lines will very likely overflow or be clipped by the pill. Either drop the responsive column rule and let the times stay on one row inside the badge, or render two separate stacked Badges on small screens. Same change is duplicated verbatim in ManagerTimesheetRequestTable/styles.ts:28-36 — verify the mobile rendering of both tables before merge.

frontend/src/community/attendance/components/molecules/TimesheetRequestFilterBody/TimesheetRequestFilterBody.tsx
```diff
@@ -71,13 +71,13 @@ const TimesheetRequestFilterBody: FC<Props> = ({
...
       }}
     >
-      <div className="flex flex-col gap-2">
-        <p className="subtitle1">{translateText(["dateRangeLabel"])}</p>
-        <DateRangePicker
-          value={selectedDateRange}
-          onChange={(range) => setSelectedDateRange(clampToCurrentYear(range))}
-        />
-      </div>
+      <DateRangePicker
```
ThinuwanW, line 74: 🤖 **Claude** · 🟡 **Suggestion**

Missing paired change. `ManagerLeaveRequestFilterBody.tsx:83-89` is the same filter-popover pattern (BasicFilterStructure → DateRangePicker → SelectableItemList) and still uses the old markup this PR is replacing: a wrapping `<div className="flex flex-col gap-2">` with a free-standing `<p className="subtitle1">{translateText(["dateTitle"])}</p>` and no `fromPlaceholder`/`toPlaceholder`. After this PR the two filter popovers render their date field differently — different label placement, and the leave one shows the library's default placeholders while the timesheet one shows "From"/"To". Update `ManagerLeaveRequestFilterBody` to the new `label`/`fromPlaceholder`/`toPlaceholder` props in the same PR so the design change lands consistently.

frontend/src/community/attendance/components/molecules/EmployeeTimesheetRequestTable/EmployeeTimesheetRequestTable.tsx
```diff
@@ -150,76 +151,85 @@ const EmployeeTimesheetRequestTable: FC<Props> = ({
...
-              style={{
-                display: "flex",
-                alignItems: "center",
-                justifyContent: "center",
-                backgroundColor: theme.palette.common.white,
-                borderRadius: "9.375rem",
-                padding: "0.5rem 1rem",
-                gap: "0.5rem"
-              }}
+            <Badge
```
ThinuwanW, line 224: 🤖 **Claude** · 🟡 **Suggestion**

Component-level duplication (distinct from the styles.ts duplication): after this PR the status pill block is byte-identical between the two tables — the whole `Badge` + `<span role="img">` + `pascalCaseFormatter(status)` markup here matches ManagerTimesheetRequestTable:315-323, and `requestTypeSelector` (Employee 108-121 / Manager 136-149), `onSuccess`, `onError`, `getKebabMenuOptions` and the `useCancelTimeRequest` wiring are already copy-pasted between them. Every fix raised on this PR (the background/text token pairing, the `role="img"` wrapper, the missing `size`) therefore has to be applied twice and can silently drift. Extract a `TimesheetRequestStatusBadge` atom taking `status: TimeSheetRequestStates` (it owns the icon map and the token pair) and a `useTimesheetRequestCancel` hook for the toast/kebab wiring, and consume both from the two tables.

frontend/pages/community/timesheet/timesheet-requests.tsx
```diff
@@ -39,6 +39,7 @@ const TimesheetRequestsPage: NextPage = () => {
       pageHead={translateText(["timesheet.timeEntryRequests.pageHead"])}
       title={translateText(["timesheet.timeEntryRequests.title"])}
       isDividerVisible={true}
+      dividerStyles={{ paddingBottom: "0rem" }}
```
ThinuwanW, line 42: 🤖 **Claude** · 🟣 **Nit**

`dividerStyles={{ paddingBottom: "0rem" }}` tunes a placeholder that renders nothing. In `ContentLayout.tsx:387-392` the `Divider` is commented out ("with new design, divider is removed but not finalized yet"), so `isDividerVisible` now only renders an empty `Stack` carrying `padding: "1rem 0rem 1rem 0rem"` from `classes.dividerWrapper`. This page's spacing therefore depends on a spacer that the comment says is pending removal — when that block is finalised the top gap here changes silently, and the intent ("remove the bottom gap") is not discoverable from the call site. Prefer expressing the spacing on the page content (or on `TableView`'s `className`), or resolve the ContentLayout placeholder. Note `my-timesheet.tsx:36` already uses `dividerStyles={{ my: "1rem" }}`, which is dead since `dividerWrapper` sets `padding`, not margin — the same trap.

frontend/src/community/common/components/organisms/TableView/TableView.tsx
```diff
@@ -8,6 +8,8 @@ import { FC, MouseEvent, useState } from "react";
 
 import { TableViewProps } from "./types";
 
+const EMPTY_STATE_MIN_HEIGHT = "min-h-80";
+
```
ThinuwanW, line 12: 🤖 **Claude** · 🟣 **Nit**

Dead API surface in the file this PR is editing: `TableViewProps.footer?: TableViewFooter` (and the `TableViewFooter` interface) is declared in `./types.ts:59-62, 83` but `footer` is never destructured from props nor rendered anywhere in `TableView`. A consumer that passes `footer={...}` gets silent no-op behaviour with no type error. Since you are already changing the render path here, either wire it up (render `footer.left`/`footer.right` below `TableV2`) or delete the prop and the interface.

frontend/src/community/attendance/components/molecules/EmployeeTimesheetRequestTable/EmployeeTimesheetRequestTable.tsx
```diff
@@ -150,76 +151,85 @@ const EmployeeTimesheetRequestTable: FC<Props> = ({
...
-              {timesheetRequest?.requestedStartTime &&
-                timesheetRequest?.requestedStartTime !==
-                  timesheetRequest?.initialClockIn && (
-                  <Typography variant="body2" sx={classes.errorTextStyles}>
-                    {timesheetRequest?.requestedStartTime}
-                  </Typography>
-                )}
-            </Box>
+            <Badge
+              backgroundColor="bg-tertiary-background"
```
ThinuwanW, line 155: 🤖 **Claude** · 🟣 **Nit**

None of the five new `Badge` usages pass `size`, while every pre-existing `Badge` call site in the repo pins it explicitly — `PeopleSyncStatusLabel.tsx:13`, `PeopleSyncBanner.tsx:47/56/66/74`, `GoogleWorkspaceSyncTabLabel.tsx`, and `GroupsCell.tsx:6` (`size: "sm"` in shared `badgeProps`). Relying on the library default means these pills can render at a different height from every other badge in the app, and a default change in a future @rootcodelabs/skapp-ui bump silently alters table row heights. If the larger default is the intended design, that is fine — otherwise add `size="sm"` for consistency (ideally once, inside the shared status-badge atom suggested separately).

frontend/src/community/common/assets/languages/english/attendanceModule.json
```diff
@@ -138,6 +138,8 @@
     "weekTabTxt": "This Week",
     "monthTabTxt": "This Month",
     "dateRangeLabel": "Date:",
+    "fromDatePlaceholder": "From",
```
ThinuwanW, line 141: 🤖 **Claude** · 🟣 **Nit**

`fromDatePlaceholder: "From"` and `toDatePlaceholder: "To"` duplicate values that already exist in this same `timesheet` namespace: `fromTimeLabel: "From"` and `toTimeLabel: "To"` (around lines 127-128). Four keys now hold two strings in one section, which means translators localise "From"/"To" twice and they can drift apart. These are also generic UI strings that the leave-module filter will need once it adopts the same `DateRangePicker` props — consider putting a single pair in `common.json` and referencing it from both filter bodies instead of adding module-local copies.

frontend/src/community/common/components/organisms/TableView/TableView.tsx
```diff
@@ -36,6 +38,7 @@ const TableView: FC<TableViewProps> = ({
   const isInfiniteScroll = infiniteScroll?.isEnabled;
   const isPaginated = !isInfiniteScroll && pagination;
   const popoverId = filter?.popoverId;
+  const isEmptyStateVisible = !isLoading && rows.length === 0;
```
ThinuwanW, line 41: 🤖 **Claude** · 🟣 **Nit**

New conditional logic in a shared organism with no test. `TableView` has no `.test.tsx` anywhere in the repo yet is now consumed by 11 tables (Employee/Manager timesheet requests, LeaveRequests, ManagerLeaveRequest, LeaveTypesTable, PolicyLeaveTypesTable, CustomLeaveAllocationsTable, AssignLeavePolicyForm, InvoiceTable, CustomerTable), so this `isEmptyStateVisible` branch changes every one of them with nothing guarding the behaviour. Add a test with `MockTheme` covering the three states: `isLoading` true → skeleton path, `rows: []` + `isLoading` false → empty-state height, and non-empty rows → the caller's `minHeight` honoured. That test would also lock in whichever resolution you pick for the `minHeight`-vs-empty-state precedence raised elsewhere on this PR.

frontend/pages/community/timesheet/timesheet-requests.tsx
Deshan1100: add this to remove divider down padding

frontend/src/community/attendance/components/molecules/EmployeeTimesheetRequestTable/EmployeeTimesheetRequestTable.tsx
Deshan1100: updated with badge

frontend/src/community/attendance/components/molecules/EmployeeTimesheetRequestTable/EmployeeTimesheetRequestTable.tsx
Deshan1100: need this

frontend/src/community/attendance/components/molecules/EmployeeTimesheetRequestTable/EmployeeTimesheetRequestTable.tsx
Deshan1100: No need

frontend/src/community/attendance/components/molecules/EmployeeTimesheetRequestTable/EmployeeTimesheetRequestTable.tsx
Deshan1100: kept as previous

frontend/src/community/attendance/components/molecules/EmployeeTimesheetRequestTable/EmployeeTimesheetRequestTable.tsx
Deshan1100: old code had the same limitation

frontend/src/community/attendance/components/molecules/EmployeeTimesheetRequestTable/EmployeeTimesheetRequestTable.tsx
Deshan1100: No need

frontend/src/community/attendance/components/molecules/EmployeeTimesheetRequestTable/EmployeeTimesheetRequestTable.tsx
Deshan1100: updated

frontend/src/community/attendance/components/molecules/EmployeeTimesheetRequestTable/styles.ts
Deshan1100: updated

frontend/src/community/attendance/components/molecules/EmployeeTimesheetRequestTable/EmployeeTimesheetRequestTable.tsx
Deshan1100: kept as previous

frontend/src/community/attendance/components/molecules/EmployeeTimesheetRequestTable/styles.ts
Deshan1100: removed

frontend/src/community/attendance/components/molecules/ManagerTimesheetRequestTable/ManagerTimesheetRequestTable.tsx
Deshan1100: updated

frontend/src/community/attendance/components/molecules/ManagerTimesheetRequestTable/ManagerTimesheetRequestTable.tsx
Deshan1100: need this change

frontend/src/community/attendance/components/molecules/EmployeeTimesheetRequestTable/EmployeeTimesheetRequestTable.tsx
Deshan1100: kept as previous components

frontend/src/community/attendance/components/molecules/ManagerTimesheetRequestTable/ManagerTimesheetRequestTable.tsx
Deshan1100: kept previous component as it is

frontend/src/community/attendance/components/molecules/TimesheetRequestFilterBody/TimesheetRequestFilterBody.tsx
Deshan1100: updated it date range picker component level

frontend/src/community/attendance/components/molecules/TimesheetRequestFilterBody/TimesheetRequestFilterBody.tsx
Deshan1100: updated in component level skapp ui

frontend/src/community/common/assets/languages/english/attendanceModule.json
Deshan1100: need this

frontend/src/community/common/components/organisms/TableView/TableView.tsx
Deshan1100: need to increase the min hight used in Empty View

frontend/src/community/common/components/organisms/TableView/TableView.tsx
Deshan1100: No need test

frontend/src/community/common/components/organisms/TableView/TableView.tsx
Deshan1100: need this change

frontend/src/community/common/components/organisms/TableView/TableView.tsx
Deshan1100: need to get minHeight here

shakila-kamalasena (dismissed):

shakila-kamalasena (approved):

frontend/src/community/attendance/components/molecules/EmployeeTimesheetRequestTable/EmployeeTimesheetRequestTable.tsx
```diff
@@ -150,76 +151,83 @@ const EmployeeTimesheetRequestTable: FC<Props> = ({
...
-                gap: "0.5rem"
-              }}
+            <Badge
+              backgroundColor="bg-tertiary-background"
+              textColor="text-secondary-text"
             >
-              <span>{pascalCaseFormatter(timesheetRequest?.status)}</span>
-              {requestTypeSelector(timesheetRequest?.status)}
-            </div>
+              <span>{requestTypeSelector(timesheetRequest?.status)}</span>
```
akilarootcode, line 228: replace with old pattern

akilarootcode (changes requested):

frontend/src/community/attendance/components/molecules/EmployeeTimesheetRequestTable/EmployeeTimesheetRequestTable.tsx
Deshan1100: need this change 
image
 now new badge looks like this

akilarootcode (approved):

PR #1757, opened 2026-08-14
Refactor/leave policy flow implementations

hasalarootcode:
…mponents

- Added new hooks for handling policy leave application submission and availability checks.
- Introduced components for date selection and request details in the Apply Policy Leave modal.
- Updated constants and enums to support new toast messages for leave actions.
- Refactored utility functions for better handling of leave request statuses and toasts.
- Improved state management for leave requests and team availability.
- Added new types and interfaces for better type safety and clarity in the leave management process.

### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This is a well-executed decomposition of the policy-leave apply/review modals: ~600 lines of inline logic moved into 6 focused hooks, 2 presentational sections, and pure utils, plus MUI→Tailwind markup conversion. I verified behavioral preservation line by line — effect dependencies, effect ordering, mutation payloads, toast key paths, popup-type resolution, and query enablement all match the originals, and every translation key referenced by the new constant maps exists in leaveModule.json. The Tailwind conversions are numerically exact (grey[400] #3F3F46 == zinc-700, gap-2/gap-5/mt-4/mt-5/mb-1.5 all match the old rem values), and `isIcon: true` in the unified toast helper is a no-op since the Toast component already defaults it to true. I found no correctness regressions; the findings below are consistency, state-placement, and test-coverage issues.

Found **18** new issue(s): 🔴 2 important, 🟡 7 suggestion(s), 🟣 9 nit(s)

frontend/src/community/leave/hooks/usePolicyLeaveAvailabilityCheck.ts
```diff
@@ -0,0 +1,108 @@
...
+    selectedDates,
+    selectedDuration,
+    availability,
+    setAvailability,
+    setFormError
+  } = usePolicyLeaveStore((state) => ({
+    selectedPolicyBalance: state.selectedPolicyBalance,
+    selectedDates: state.selectedDates,
+    selectedDuration: state.selectedDuration,
+    availability: state.availability,
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

After this refactor, the `availability` slice in `policyLeaveStore` has exactly one reader and one writer, and both are inside this hook (I grepped the whole `src` tree — `state.availability` / `setAvailability` appear only here and in the store definition itself). Keeping it in Zustand means every availability response broadcasts a store update to all seven subscribers of `usePolicyLeaveStore` in the apply-modal tree, re-rendering the calendar and details sections for data only this hook consumes. Since the extraction has isolated it, replace `availability` / `setAvailability` with a local `useState<PolicyLeaveAvailabilityType | null>(null)` in this hook and drop `availability` + `setAvailability` from `PolicyLeaveStore`, `emptyForm()`, and the store type. This also matches the guideline against holding data in global state that should be local component state.

frontend/src/community/leave/components/molecules/PolicyLeaveModals/ApplyPolicyLeaveModal/ApplyPolicyLeaveModal.tsx
```diff
@@ -95,376 +26,40 @@ const ApplyPolicyLeaveModal = () => {
...
-    setFormError,
-    setFormErrors,
-    setAvailability,
     setModalType
-  } = usePolicyLeaveStore();
-
-  const isCalendarSelectionInvalid = useLeaveStore(
-    (state) => state.isApplyLeaveModalBtnDisabled
-  );
+  } = usePolicyLeaveStore((state) => ({
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The PR introduces seven new `usePolicyLeaveStore((state) => ({ ... }))` subscriptions (here, PolicyLeaveDateSection, PolicyLeaveRequestDetailsSection, useApplyPolicyLeaveSubmit, usePolicyLeaveAvailabilityCheck, usePolicyLeaveCalendarData, usePolicyLeaveTeamAvailability). Each selector returns a fresh object literal, so with zustand 4.5 (no `equalityFn` passed) every one of them re-renders on *any* store write — e.g. each keystroke in the comment box (`setComment`) re-renders the calendar section and re-runs all four hooks. This isn't a regression (the old monolithic component called `usePolicyLeaveStore()` with no selector at all), but the decomposition only pays off with shallow comparison. Wrap each object selector in `useShallow` from `zustand/react/shallow` (available in 4.5, currently unused anywhere in the repo), e.g. `usePolicyLeaveStore(useShallow((state) => ({ ... })))`. Selectors that return a single value (`selectHasUnsavedChanges`, `state.setSelectedDuration`) can stay as-is.

frontend/src/community/leave/hooks/usePolicyLeaveReviewAction.ts
```diff
@@ -0,0 +1,68 @@
+import { useTranslator } from "~community/common/hooks/useTranslator";
```
ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

None of the newly extracted units have tests. The repo has 219 test files but only one under the policy-leave feature (`policyLeaveTypeUtils.test.ts`). The whole point of this extraction is that these are now trivially testable in isolation — in particular the pure functions added in this PR: `getPolicyPeriodYears` (fallback-year branch and the spansTwoYears boundary), `mapToBlockingLeaveRequests` (the `toLeaveStatus` null filter that silently drops DENIED/CANCELLED/REVOKED requests), `getIsStorageFull` (community-only + undefined availableSpace), `getPolicyLeaveResultStatus` (the four-way popupType resolution, including the REVOKED fallback), `formatOptionalDate` (empty/invalid date -> ""), and `handlePolicyLeaveReviewToast` / `REVIEW_TOAST_CONFIG` (which key namespace each of the 10 toast enums maps to — a silent typo here renders a raw key string to users). Add a `policyLeaveReviewUtils.test.ts` and `policyLeaveUtils.test.ts` covering these; the hooks themselves can follow with `renderHook` if desired.

frontend/src/community/leave/components/molecules/PolicyLeaveReviewModals/PolicyLeaveRequestSummary/PolicyLeaveRequestSummary.tsx
```diff
@@ -36,66 +34,58 @@ const PolicyLeaveRequestSummary: FC<Props> = ({
...
-          variant="body1"
-          sx={{
-            color: theme.palette.grey[400],
-            marginTop: "1rem"
-          }}
-          tabIndex={0}
-        >
-          {translateText(["myLeaveRequests", layout.descriptionKey])}
-        </Typography>
+        <p className="body1 text-zinc-700 mt-4">
```
ThinuwanW, line 53: 🤖 **Claude** · 🟣 **Nit**

`text-zinc-700` is introduced here (and at PolicyCancelLeaveModal.tsx:47) and these are the only two occurrences in the entire codebase. The established convention is the semantic token `text-secondary-text` (87 usages) with `text-tertiary-text` as the secondary option. The raw value is defensible — zinc-700 (#3f3f46) exactly reproduces the old `theme.palette.grey[400]`, whereas `--color-secondary-text` is #52525b, so switching would be a real (if small) visual change. Worth a deliberate call with design: either add a token for #3f3f46 to the `@theme` block in styles/global.css, or accept the shift to `text-secondary-text` so these two elements stay on the design system.

frontend/src/community/leave/components/molecules/PolicyLeaveModals/ApplyPolicyLeaveModal/PolicyLeaveDateSection.tsx
```diff
@@ -0,0 +1,91 @@
...
+      formErrors: state.formErrors,
+      setSelectedDates: state.setSelectedDates,
+      setSelectedMonth: state.setSelectedMonth
+    }));
+
+  const dateFieldRef = useRef<HTMLFieldSetElement>(null);
+
+  const dateError = formErrors?.selectedDates;
+
+  useEffect(() => {
```
ThinuwanW, line 52: 🤖 **Claude** · 🟣 **Nit**

The focus effect changed from keying on a boolean to keying on the message string. Previously `const hasDateError = Boolean(formErrors?.selectedDates)` with `useEffect(..., [hasDateError])` fired only on the false→true transition; now `useEffect(..., [dateError])` also fires whenever the error *text* changes while an error is already displayed. Concretely: dates selected produce "You only have N day(s) remaining on X", the user then switches duration to half-day and the availability check returns "You already have a leave request that overlaps these dates" — focus is yanked back to the fieldset from wherever the user was. Restore the previous semantics with `const hasDateError = Boolean(dateError)` and `useEffect(() => { if (hasDateError) dateFieldRef.current?.focus(); }, [hasDateError])`, keeping `dateError` for the className and the `error` prop.

frontend/src/community/leave/utils/policyLeave/policyLeaveReviewUtils.ts
```diff
@@ -1,11 +1,145 @@
...
+    description: translateText(config.descriptionKey),
+    isIcon: true
+  });
+};
+
+export const isApprovedPopupType = (popupType: PolicyLeavePopupType): boolean =>
+  popupType === PolicyLeaveRequestStatus.APPROVED ||
+  popupType === PolicyLeaveReviewModalEnums.APPROVED_STATUS;
+
+export const isDeniedPopupType = (popupType: PolicyLeavePopupType): boolean =>
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`isDeniedPopupType` is exported but has no consumer outside this module — its only call site is `getPolicyLeaveResultStatus` twelve lines below (I grepped `src` and `pages`). `isApprovedPopupType` is genuinely used externally by PolicyLeaveReviewResultModal, so the asymmetry looks accidental. Drop the `export` keyword on `isDeniedPopupType` to keep the module surface honest.

frontend/src/community/leave/hooks/usePolicyLeaveReviewAction.ts
```diff
@@ -0,0 +1,68 @@
...
+  );
+
+  const reviewRequest = (
+    leaveRequestId: number,
+    reviewerComment?: string
+  ): void => {
+    mutate({
+      leaveRequestId,
+      status,
+      ...(reviewerComment === undefined ? {} : { reviewerComment })
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

The conditional spread `...(reviewerComment === undefined ? {} : { reviewerComment })` is a no-op: `PolicyLeaveReviewPayload.reviewerComment` is already optional, and `reviewPolicyLeaveRequest` destructures then re-serializes it into the request body, so an omitted key and an explicit `undefined` both get dropped by JSON serialization. Simplify to `mutate({ leaveRequestId, status, reviewerComment })`.

frontend/src/community/leave/components/molecules/PolicyLeaveReviewModals/PolicyLeaveRequestSummary/PolicyLeaveRequestSummary.tsx
```diff
@@ -36,66 +34,58 @@ const PolicyLeaveRequestSummary: FC<Props> = ({
...
-  const theme: Theme = useTheme();
-  const translateText = useTranslator("leaveModule", "myRequests");
+  const translateText = useTranslator(
+    "leaveModule",
+    "myRequests",
+    "myLeaveRequests"
+  );
+  const translateMyRequestsText = useTranslator("leaveModule", "myRequests");
 
   const layout = SUMMARY_LAYOUTS[popupType];
```
ThinuwanW, line 44: 🤖 **Claude** · 🟣 **Nit**

`SUMMARY_LAYOUTS` is typed `Record<string, PolicyLeaveSummaryLayout>`, so `SUMMARY_LAYOUTS[popupType]` is typed as always-defined while it can be `undefined` at runtime for any popupType not in the map (PENDING, DECLINE, APPROVED_STATUS, DECLINE_STATUS, CANCEL_REQUEST_POPUP). Today the sole caller guards with `SUMMARY_POPUP_TYPES.includes(popupType)` so nothing breaks, but the type gives no protection if a second caller appears — `layout.descriptionKey` would throw. Since this PR is already reshaping these constants, tighten it to `Partial<Record<PolicyLeavePopupType, PolicyLeaveSummaryLayout>>` and add an `if (!layout) return null;` guard here; `SUMMARY_POPUP_TYPES` derived via `Object.keys` continues to work unchanged.

frontend/src/community/leave/hooks/useApplyPolicyLeaveSubmit.ts
```diff
@@ -0,0 +1,154 @@
...
+    setFormErrors
+  ]);
+
+  useEffect(() => {
+    if (hasPolicyLeaveFormErrors(formErrors)) {
+      validate();
+    }
+  }, [selectedDates, comment, attachments]);
+
+  const submitPolicyLeave = async (): Promise<void> => {
```
ThinuwanW, line 125: 🤖 **Claude** · 🔴 **Important**

Duplicate-submit window: `submitPolicyLeave` awaits `uploadPolicyLeaveAttachments(...)` before it ever calls `applyPolicyLeave`, but the only pending flag returned from this hook is `isApplyPending`, which comes from the `useApplyPolicyLeave` mutation. During the whole upload phase `isApplyPending` is still false, so in ApplyPolicyLeaveModal the Confirm button (`disabled={isSubmitDisabled || isApplyPending}`, `isLoading={isApplyPending}`) stays enabled and un-spinnered. A second click on a request with attachments re-runs validation (which passes), re-uploads every attachment to S3, and fires a second `applyPolicyLeave` mutation — the user gets two identical leave requests and duplicate stored files. The refactor is the right moment to close this: add `const [isSubmitting, setIsSubmitting] = useState(false)`, set it true at the top of `submitPolicyLeave`, clear it in a `finally`, and return `isApplyPending: isSubmitting || isApplyPending` so the existing disabled/loading wiring covers the upload phase too.

frontend/src/community/leave/utils/policyLeave/policyLeaveReviewUtils.ts
```diff
@@ -1,11 +1,145 @@
...
+    descriptionKey: myRequestKey("nudgeSuccessDescription")
+  },
+  [PolicyLeaveReviewToastEnums.NUDGE_ERROR]: {
+    toastType: ToastType.ERROR,
+    titleKey: myRequestKey("nudgeErrorTitle"),
+    descriptionKey: myRequestKey("nudgeErrorDescription")
+  }
+};
+
+export const handlePolicyLeaveReviewToast = ({
```
ThinuwanW, line 102: 🤖 **Claude** · 🔴 **Important**

`handlePolicyLeaveReviewToast` has an undocumented, unenforceable contract on its `translateText` argument. REVIEW_TOAST_CONFIG mixes two namespace roots — `managerKey()` emits `leaveRequests.leaveManagerEmployee.*` while `myRequestKey()` emits `myRequests.myLeaveRequests.*` — so the function only works if the caller passes a translator bound to exactly `useTranslator("leaveModule")` and nothing deeper. The parameter type is just `TranslateFn`, which every translator in the codebase satisfies, so passing the component's natural translator compiles fine and fails at runtime. It fails silently, too: `useTranslator` does `t(key) || ""`, and i18next returns the key string on a miss, so a wrong namespace renders a toast titled `leaveModule.myRequests.myLeaveRequests.leaveRequests.leaveManagerEmployee.approveLeaveSuccessTitle` in production. This already forced PolicyEmployeeRequestModal to add a third translator (`translateLeaveModuleText`) purely to satisfy it. Make the contract explicit — e.g. rename the param to `translateLeaveModuleText`, add a doc comment stating the required base namespace, or introduce a branded `LeaveModuleTranslateFn` type so a mis-scoped translator is a type error rather than a runtime string.

frontend/src/community/leave/components/molecules/PolicyLeaveModals/ApplyPolicyLeaveModal/PolicyLeaveRequestDetailsSection.tsx
```diff
@@ -0,0 +1,169 @@
...
+    comment: state.comment,
+    attachments: state.attachments,
+    formErrors: state.formErrors,
+    setSelectedDuration: state.setSelectedDuration,
+    setComment: state.setComment,
+    setAttachments: state.setAttachments,
+    setModalType: state.setModalType
+  }));
+
+  const { data: storageAvailabilityData } = useStorageAvailability();
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`useStorageAvailability()` is called with no `enabled` gate, so the storage-availability request fires on every open of the apply-leave modal — including for policies where `policyBalance.leaveType.isAttachment` is false (the attachment icon is never rendered, so the data can never be read) and in enterprise mode where `getIsStorageFull` short-circuits on `environment === appModes.COMMUNITY` and ignores the result entirely. The hook already accepts an enabled flag (`useStorageAvailability(isSessionDataAvailable: boolean = true)`), and this component now has both facts in scope. Change to `useStorageAvailability(policyBalance.leaveType.isAttachment)` (optionally also `&& environment === appModes.COMMUNITY`). Guideline: "Using useQuery without an enabled flag when the query depends on runtime state."

frontend/src/community/leave/hooks/usePolicyLeaveCalendarData.ts
```diff
@@ -0,0 +1,156 @@
...
+      getDurationSelectorDisabledOptions({
+        selectedDates,
+        duration: minDuration,
+        myLeaveRequests: blockingLeaveRequests,
+        allHolidays
+      }),
+    [selectedDates, minDuration, blockingLeaveRequests, allHolidays]
+  );
+
+  useEffect(() => {
```
ThinuwanW, line 140: 🤖 **Claude** · 🟡 **Suggestion**

`usePolicyLeaveCalendarData` and `usePolicyLeaveTeamAvailability` are named and typed as read-only data providers (`PolicyLeaveCalendarData`, `PolicyLeaveTeamAvailability` — both plain derived-value shapes), but each hides a store write in an effect: this one calls `setSelectedDuration(getDurationInitialValue(...))`, and the team hook calls `setSelectedTeam(myTeams[0])`. Nothing in either signature signals that merely calling the hook mutates global form state, so a second call site (a sibling section, a test render, a future summary panel) would silently reset the user's chosen duration or team. Either rename to reflect the effect (`usePolicyLeaveCalendarDataSync` / or split the reset effect into a dedicated `usePolicyLeaveDurationDefault` hook owned by the one component that should drive it), or add a doc comment stating the hook must be mounted exactly once per modal.

frontend/src/community/leave/components/molecules/PolicyLeaveReviewModals/PolicyEmployeeRequestModal/PolicyEmployeeRequestModal.tsx
```diff
@@ -27,15 +26,26 @@ import {
...
 }
 
 const PolicyEmployeeRequestModal: FC<Props> = ({ request, setPopupType }) => {
-  const translateText = useTranslator("leaveModule", "myRequests");
+  const translateText = useTranslator(
+    "leaveModule",
+    "myRequests",
+    "myLeaveRequests"
+  );
+  const translateLeaveModuleText = useTranslator("leaveModule");
```
ThinuwanW, line 46: 🤖 **Claude** · 🟡 **Suggestion**

The nudge action was left out of the abstraction this PR introduced. Approve, decline, revoke and cancel all now go through `usePolicyLeaveReviewAction` (mutation + toast + analytics + onSuccess in one place), but nudge still hand-wires `useNudgePolicyLeaveRequestManagers` with two inline callbacks, its own `useToast()`, and a third `useTranslator` instance created solely to feed `handlePolicyLeaveReviewToast`. The result is that this one component now carries three translators over overlapping prefixes of the same tree (`leaveModule`, `leaveModule.myRequests`, `leaveModule.myRequests.myLeaveRequests`) and is the only review-flow component still importing `useToast` directly. Extract a sibling `usePolicyLeaveNudgeAction({ successToast, errorToast, onSuccess })` mirroring `usePolicyLeaveReviewAction`; that removes `useToast`, `translateLeaveModuleText` and both callbacks from this file.

frontend/src/community/leave/components/molecules/PolicyLeaveReviewModals/PolicyEmployeeRequestModal/PolicyEmployeeRequestModal.tsx
```diff
@@ -1,22 +1,21 @@
...
-import { convertDateToFormat } from "~community/common/utils/dateTimeUtils";
 import {
   useCheckPolicyLeaveAlreadyNudged,
   useNudgePolicyLeaveRequestManagers
 } from "~community/leave/api/PolicyLeaveReviewApi";
 import PolicyLeaveAttachmentRow from "~community/leave/components/molecules/PolicyLeaveAttachmentRow/PolicyLeaveAttachmentRow";
 import StatusPopupColumn from "~community/leave/components/molecules/StatusPopupColumn/StatusPopupColumn";
 import StatusPopupRow from "~community/leave/components/molecules/StatusPopupRow/StatusPopupRow";
-import { PolicyLeaveReviewModalEnums } from "~community/leave/enums/PolicyLeaveReviewEnums";
+import {
```
ThinuwanW, line 15: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Remaining duplication that the PR's own `SUMMARY_LAYOUTS` mechanism was built to remove: this component's row block (type → status → duration → dateApplied → recipient → reason → PolicyLeaveAttachmentRow) is a near-verbatim copy of PolicyLeaveRequestSummary's, down to the `styles={{alignItems:'flex-start'}}` / `textStyles={{mt:'0.75rem'}}` recipient tweak and the `handleDurationDay(..., translateMyRequestsText)` call. This PR edited both copies in exactly the same way (namespace re-basing, `formatOptionalDate`), which is the tell — the next change to a row will have to be made twice and one copy will drift. Add a `[PolicyLeaveRequestStatus.PENDING]` entry to SUMMARY_LAYOUTS (showDateApplied: true, showReason: true, showAttachments: true, isRecipientTopAligned: true) and give PolicyLeaveRequestSummary an optional `actions?: ReactNode` slot for the nudge/cancel buttons; PolicyEmployeeRequestModal then shrinks to the mutation wiring plus the two buttons. Guideline: "2+ components with near-identical markup → extract shared component."

frontend/src/community/leave/constants/policyLeaveReviewConstants.ts
```diff
@@ -1,10 +1,32 @@
...
+  PolicyLeaveResultStatus,
+  IconName
+> = {
+  [PolicyLeaveRequestStatus.APPROVED]: IconName.APPROVED_STATUS_ICON,
+  [PolicyLeaveRequestStatus.DENIED]: IconName.DENIED_STATUS_ICON,
+  [PolicyLeaveRequestStatus.CANCELLED]: IconName.CANCELLED_STATUS_ICON,
+  [PolicyLeaveRequestStatus.REVOKED]: IconName.REVOKED_STATUS_ICON
+};
+
+export const POLICY_LEAVE_CHIP_STYLES = {
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`POLICY_LEAVE_CHIP_STYLES` and `POLICY_LEAVE_ATTACHMENT_CHIP_STYLES` are the only exports in this file without an explicit type annotation — `RESULT_STATUS_ICON_NAMES`, `STATUS_POPUP_TYPES`, `SUMMARY_LAYOUTS`, `REVIEW_RESULT_POPUP_TYPES` and both title-key maps are all explicitly typed. They are also MUI `sx` fragments (`backgroundColor: 'grey.100'`, `py`) living in a `constants/` module rather than a colocated `styles.ts`, and they still hardcode the `grey.100` palette shorthand while the rest of this PR is converting these surfaces to Tailwind. Suggest at minimum an explicit type (a small local `interface PolicyLeaveChipStyles { backgroundColor: string; py: string; px?: string }` keeps the object spreadable in PolicyLeaveAttachmentRow, unlike `SxProps<Theme>` which is a union and cannot be spread), and consider moving them next to the components or finishing the Tailwind migration for these chips.

frontend/src/community/leave/components/molecules/PolicyLeaveReviewModals/PolicyLeaveReviewResultModal/PolicyLeaveReviewResultModal.tsx
```diff
@@ -50,85 +41,38 @@ const PolicyLeaveReviewResultModal: FC<Props> = ({
...
-      leaveRequestId: request.leaveRequestId,
+  const { reviewRequest: revokeRequest, isPending } =
+    usePolicyLeaveReviewAction({
       status: PolicyLeaveRequestStatus.REVOKED,
-      reviewerComment: ""
+      successToast: PolicyLeaveReviewToastEnums.REVOKE_SUCCESS,
+      errorToast: PolicyLeaveReviewToastEnums.REVOKE_ERROR,
+      analyticsEvent: GoogleAnalyticsTypes.GA4_LEAVE_REQUEST_REVOKED,
+      onSuccess: () => {
+        setPopupType(PolicyLeaveReviewModalEnums.NONE);
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`setPopupType(PolicyLeaveReviewModalEnums.NONE)` here is dead: the very next line calls `closeModal()`, which the controller wires to `closePopup`, and `usePolicyLeaveStatusPopup.closePopup` already does `setPopupType(NONE)` before invoking `onClose`. Now that the reset lives inside the extracted hook, this line just re-states an invariant the hook owns and will go stale if `closePopup` changes. Drop it and leave `onSuccess: closeModal`.

frontend/src/community/leave/hooks/usePolicyLeaveCalendarData.ts
```diff
@@ -0,0 +1,156 @@
...
+      selectedYear: state.selectedYear,
+      selectedPolicyBalance: state.selectedPolicyBalance,
+      selectedDates: state.selectedDates
+    }));
+
+  const setSelectedDuration = usePolicyLeaveStore(
+    (state) => state.setSelectedDuration
+  );
+
+  const { startYear, endYear, spansTwoYears } = getPolicyPeriodYears(
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`getPolicyPeriodYears(selectedPolicyBalance, selectedYear)` runs on every render and performs two Luxon `convertYYYYMMDDToDateTime` parses, while every other derived value in this hook (`allHolidays`, `workingDays`, `blockingLeaveRequests`, `minDate`, `maxDate`, `disabledDurationOptions`) is wrapped in `useMemo`. Its outputs also feed four query keys, so it is the hottest derivation in the hook. Wrap it: `useMemo(() => getPolicyPeriodYears(selectedPolicyBalance, selectedYear), [selectedPolicyBalance, selectedYear])` — consistent with the rest of the file and cheap to do.

frontend/src/community/leave/components/molecules/PolicyLeaveModals/ApplyPolicyLeaveModal/ApplyPolicyLeaveModal.tsx
```diff
@@ -486,98 +81,22 @@ const ApplyPolicyLeaveModal = () => {
...
+          minDate={minDate}
+          maxDate={maxDate}
+        />
+        <PolicyLeaveRequestDetailsSection
+          policyBalance={selectedPolicyBalance}
+          myTeams={myTeams}
+          resourceAvailability={resourceAvailability}
+          workingDays={workingDays}
+          disabledDurationOptions={disabledDurationOptions}
+          isSummaryVisible={!isSubmitDisabled}
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`isSummaryVisible={!isSubmitDisabled}` bakes an unrelated coupling into the new component boundary: the leave summary is now defined as "whatever makes the submit button enabled", which currently also includes `isCalendarSelectionInvalid` from a completely different store (`useLeaveStore`). The name no longer explains the condition, and the first time `isSubmitDisabled` gains a submission-lifecycle term (e.g. the `isSubmitting` flag suggested for useApplyPolicyLeaveSubmit) the summary will start vanishing mid-submit. Derive it explicitly instead — `const isSummaryVisible = selectedDates.length > 0 && selectedDuration !== LeaveStates.NONE && !availabilityError;` — so the two concerns can diverge safely.

frontend/src/community/leave/hooks/usePolicyLeaveAvailabilityCheck.ts
hasalarootcode: Fixed. Moved to local state in the hook and dropped `availability` / `setAvailability` from the store.

frontend/src/community/leave/components/molecules/PolicyLeaveModals/ApplyPolicyLeaveModal/ApplyPolicyLeaveModal.tsx
hasalarootcode: Fixed. Wrapped every object selector in `useShallow` - the seven you listed plus the two review controllers.

frontend/src/community/leave/hooks/usePolicyLeaveReviewAction.ts
hasalarootcode: No tests in this PR yet. The pure utils are the cheap ones to cover - `getPolicyPeriodYears`, `mapToBlockingLeaveRequests`, `getIsStorageFull`, `getPolicyLeaveResultStatus`, `formatOptionalDate` and the toast config map. Add them here, or a follow-up?

frontend/src/community/leave/components/molecules/PolicyLeaveReviewModals/PolicyLeaveRequestSummary/PolicyLeaveRequestSummary.tsx
hasalarootcode: `text-zinc-700` is there to hold the exact `grey[400]` these had before the MUI conversion. Moving to `text-secondary-text` is a visible shift on both elements. Take the shift, or add a token for #3f3f46?

frontend/src/community/leave/components/molecules/PolicyLeaveModals/ApplyPolicyLeaveModal/PolicyLeaveDateSection.tsx
hasalarootcode: Fixed. Back on `hasDateError`; `dateError` still feeds the className and the `error` prop.

frontend/src/community/leave/utils/policyLeave/policyLeaveReviewUtils.ts
hasalarootcode: Removed.

frontend/src/community/leave/hooks/usePolicyLeaveReviewAction.ts
hasalarootcode: Fixed.

frontend/src/community/leave/components/molecules/PolicyLeaveReviewModals/PolicyLeaveRequestSummary/PolicyLeaveRequestSummary.tsx
hasalarootcode: Fixed. `Partial<Record<PolicyLeavePopupType, PolicyLeaveSummaryLayout>>` with a null guard here.

frontend/src/community/leave/hooks/useApplyPolicyLeaveSubmit.ts
hasalarootcode: Fixed. `isSubmitting` wraps the whole submit including the upload, and is folded into the returned `isApplyPending`.

frontend/src/community/leave/utils/policyLeave/policyLeaveReviewUtils.ts
hasalarootcode: Renamed the param to `translateLeaveModuleText` on the interface and at both call sites.

frontend/src/community/leave/components/molecules/PolicyLeaveModals/ApplyPolicyLeaveModal/PolicyLeaveRequestDetailsSection.tsx
hasalarootcode: Fixed - gated on `policyBalance.leaveType.isAttachment`. Left the community-mode check to `getIsStorageFull`.

frontend/src/community/leave/hooks/usePolicyLeaveCalendarData.ts
hasalarootcode: Both writes are there because the modal mounts each hook exactly once. Splitting the duration reset into its own hook owned by the details section looks better than renaming - want that in this PR?

frontend/src/community/leave/components/molecules/PolicyLeaveReviewModals/PolicyEmployeeRequestModal/PolicyEmployeeRequestModal.tsx
hasalarootcode: Nudge is a different endpoint and payload, so it would be a sibling hook rather than a reuse of `usePolicyLeaveReviewAction`. Add it here, or together with the PolicyEmployeeRequestModal restructure?

frontend/src/community/leave/components/molecules/PolicyLeaveReviewModals/PolicyEmployeeRequestModal/PolicyEmployeeRequestModal.tsx
hasalarootcode: This needs a PENDING entry in SUMMARY_LAYOUTS plus an `actions` slot on PolicyLeaveRequestSummary, and it reshapes the employee pending modal rather than just deduping. Here, or a follow-up?

frontend/src/community/leave/constants/policyLeaveReviewConstants.ts
hasalarootcode: Typed both as `PolicyLeaveChipStyles`. Left them in constants - `IconChip` and `ReadOnlyChip` still take `sx`, so there is nothing to move to Tailwind until those atoms change.

frontend/src/community/leave/components/molecules/PolicyLeaveReviewModals/PolicyLeaveReviewResultModal/PolicyLeaveReviewResultModal.tsx
hasalarootcode: Removed, along with the `setPopupType` prop - that line was its only use.

frontend/src/community/leave/hooks/usePolicyLeaveCalendarData.ts
hasalarootcode: Fixed.

frontend/src/community/leave/components/molecules/PolicyLeaveModals/ApplyPolicyLeaveModal/ApplyPolicyLeaveModal.tsx
hasalarootcode: Fixed. Kept `isCalendarSelectionInvalid` in the condition, since dropping it would show the summary on a selection the calendar has already rejected.

frontend/src/community/leave/components/organisms/PolicyLeaveReviewModalController/PolicyLeaveReviewModalController.tsx
```diff
@@ -20,93 +24,64 @@ const PolicyLeaveReviewModalController: FC = () => {
...
-      isManagerModalOpen: state.isManagerModalOpen,
-      selectedRequestId: state.selectedRequestId,
-      closeManagerModal: state.closeManagerModal
-    }));
-
-  const [popupType, setPopupType] = useState<PolicyLeavePopupType>(
-    PolicyLeaveReviewModalEnums.NONE
-  );
+    usePolicyLeaveStore(
+      useShallow((state) => ({
```
thusala, line 28: why useShallow?

frontend/src/community/leave/components/organisms/PolicyLeaveReviewModalController/PolicyLeaveReviewModalController.tsx
```diff
@@ -20,93 +24,64 @@ const PolicyLeaveReviewModalController: FC = () => {
...
     return null;
   };
 
   return (
     <SmallModal
       isOpen={isManagerModalOpen && !!popupType}
-      onClose={closeModal}
-      modalHeader={getModalTitle()}
+      onClose={closePopup}
+      modalHeader={titleKey ? translateText([titleKey]) : ""}
```
thusala, line 84: why fallback?

frontend/src/community/leave/constants/policyLeaveReviewConstants.ts
```diff
@@ -1,10 +1,33 @@
...
+
+export const POLICY_LEAVE_CHIP_STYLES: PolicyLeaveChipStyles = {
+  backgroundColor: "grey.100",
+  py: "0.75rem"
+};
+
+export const POLICY_LEAVE_ATTACHMENT_CHIP_STYLES: PolicyLeaveChipStyles = {
+  ...POLICY_LEAVE_CHIP_STYLES,
+  px: "0.75rem"
+};
```
thusala, line 29: shall keep this inside style.ts or related place?

frontend/src/community/leave/hooks/useApplyPolicyLeaveSubmit.ts
```diff
@@ -0,0 +1,166 @@
...
+    selectedPolicyBalance,
+    selectedDates,
+    selectedDuration,
+    comment,
+    attachments,
+    formErrors,
+    setFormErrors,
+    setModalType
+  } = usePolicyLeaveStore(
+    useShallow((state) => ({
```
thusala, line 58: remove useShallow

frontend/src/community/leave/hooks/useApplyPolicyLeaveSubmit.ts
```diff
@@ -0,0 +1,166 @@
...
+
+    return !hasPolicyLeaveFormErrors(errors);
+  }, [
+    selectedDates.length,
+    comment,
+    attachments,
+    selectedPolicyBalance,
+    availabilityError,
+    translateText,
+    setFormErrors
```
thusala, line 116: remove setter

frontend/src/community/leave/hooks/usePolicyLeaveAvailabilityCheck.ts
```diff
@@ -0,0 +1,111 @@
...
+  }, [
+    hasAvailabilityCheckFailed,
+    availability,
+    selectedPolicyBalance,
+    translateText
+  ]);
+
+  useEffect(() => {
+    setFormError("selectedDates", availabilityError);
+  }, [availabilityError, setFormError]);
```
thusala, line 106: remove setter

frontend/src/community/leave/hooks/usePolicyLeaveTeamAvailability.ts
```diff
@@ -0,0 +1,47 @@
...
+import { TeamNamesType } from "~community/people/types/TeamTypes";
+
+interface PolicyLeaveTeamAvailability {
+  myTeams: TeamNamesType[] | undefined;
+  resourceAvailability: ResourceAvailabilityPayload[] | undefined;
+}
+
+const usePolicyLeaveTeamAvailability = (): PolicyLeaveTeamAvailability => {
+  const { selectedMonth, selectedTeam, setSelectedTeam } = usePolicyLeaveStore(
+    useShallow((state) => ({
```
thusala, line 18: remove useShallow

frontend/src/community/leave/types/PolicyLeaveReviewTypes.ts
```diff
@@ -49,9 +55,15 @@ export interface PolicyLeaveReviewSliceType {
   resetReviewRequestFilters: () => void;
 }
 
+export interface PolicyLeaveChipStyles {
```
thusala, line 58: why do we need interface for styles?

thusala (changes requested):

PR #1758, opened 2026-08-14
Refactor/crm 713 v2 entity type defining refactor 

YasinduRC:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/frontend/src/community/crm/v2/types/CrmCommonTypes.ts
+++ b/frontend/src/community/crm/v2/types/CrmCommonTypes.ts
@@ -14,12 +14,7 @@ export interface CrmCompanyEntity {
   website?: string;
   address?: string;
   contactNumber?: string;
-  openTasksCount?: number;
-  overdue?: number;
-  openValue?: string;
-  accountValue?: string;
-  openDeals?: number;
-  closedDeals?: number;
+  metrics?: CrmCompanyMetrics;
   contactIds?: number[];
   dealIds?: number[];
   taskIds?: number[];
@@ -47,13 +42,7 @@ export interface CrmContactEntity {
   lastModifiedDate?: string;
   companyId?: number;
   ownerId?: number;
-  totalRevenue?: string;
-  pipelineRevenue?: string;
-  activeDealsCount?: number;
-  openTasksCount?: number;
-  overdueTasksCount?: number;
-  closedDealValue?: string;
-  closedDealCount?: number;
+  metrics?: CrmContactMetrics;
   dealIds?: number[];
   taskIds?: number[];
 }
--- a/frontend/src/community/crm/v2/types/CrmCommonTypes.ts
+++ b/frontend/src/community/crm/v2/types/CrmCommonTypes.ts
@@ -34,6 +34,7 @@ export interface CrmCompanyMetrics {
 
 export interface CrmContactEntity {
   id?: number;
+  name?: string;
   firstName?: string;
   lastName?: string;
   email?: string;
@@ -80,7 +81,7 @@ export interface CrmDealEntity {
   companyId?: number;
   contactId?: number;
   ownerId?: number;
-  taskCount?: number;
+  openTaskCount?: number;
   taskIds?: number[];
 }
 
--- a/frontend/src/community/crm/v2/types/CrmCommonTypes.ts
+++ b/frontend/src/community/crm/v2/types/CrmCommonTypes.ts
@@ -23,11 +23,11 @@ export interface CrmCompanyEntity {
 export interface CrmCompanyMetrics {
   id?: number;
   openTasksCount?: number;
-  overdue?: number;
+  overdueCount?: number;
   openValue?: string;
   accountValue?: string;
-  openDeals?: number;
-  closedDeals?: number;
+  openDealsCount?: number;
+  closedDealsCount?: number;
 }
 
 // Contact
--- a/frontend/src/community/crm/v2/types/CrmCommonTypes.ts
+++ b/frontend/src/community/crm/v2/types/CrmCommonTypes.ts
@@ -34,7 +34,6 @@ export interface CrmCompanyMetrics {
 
 export interface CrmContactEntity {
   id?: number;
-  name?: string;
   firstName?: string;
   lastName?: string;
   email?: string;
@@ -81,7 +80,7 @@ export interface CrmDealEntity {
   companyId?: number;
   contactId?: number;
   ownerId?: number;
-  openTaskCount?: number;
+  taskCount?: number;
   taskIds?: number[];
 }
 
--- a/frontend/src/community/crm/v2/types/CrmCommonTypes.ts
+++ b/frontend/src/community/crm/v2/types/CrmCommonTypes.ts
@@ -80,7 +80,7 @@ export interface CrmDealEntity {
   companyId?: number;
   contactId?: number;
   ownerId?: number;
-  openTasksCount?: number;
+  taskCount?: number;
   taskIds?: number[];
 }
 


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR is type-only groundwork in the unreferenced `frontend/src/community/crm/v2` module: it collapses the flat metric fields on `CrmCompanyEntity`/`CrmContactEntity` into the pre-existing `CrmCompanyMetrics`/`CrmContactMetrics` interfaces and renames `CrmDealEntity.openTasksCount` to `taskCount`. I verified the whole v2 folder (3 files) has no importers anywhere in `frontend/` (including the enterprise submodule), so nothing breaks at compile time, and `taskCount` correctly matches the backend wire name in `CrmDealByStageItemResponseDto`. Remaining concerns are contract-accuracy and consistency issues that will be expensive to unwind once the v2 store/API layer is built on top of these types.

Found **12** new issue(s): 🔴 2 important, 🟡 4 suggestion(s), 🟣 6 nit(s)

frontend/src/community/crm/v2/types/CrmCommonTypes.ts
```diff
@@ -47,13 +42,7 @@ export interface CrmContactEntity {
...
   companyId?: number;
   ownerId?: number;
-  totalRevenue?: string;
-  pipelineRevenue?: string;
-  activeDealsCount?: number;
-  openTasksCount?: number;
-  overdueTasksCount?: number;
-  closedDealValue?: string;
-  closedDealCount?: number;
+  metrics?: CrmContactMetrics;
```
ThinuwanW, line 45: 🤖 **Claude** · 🔴 **Important**

`CrmContactMetrics.closedDealValue` is typed `string`, but the backend returns it as a JSON number: `CrmContactListItemDto.closedDealValue` is a `BigDecimal` (backend/src/main/java/com/skapp/community/crmplanner/payload/response/CrmContactListItemDto.java:27). The existing v1 type agrees — `CrmContact.closedDealValue: number | null` in frontend/src/community/crm/types/CommonTypes.ts:419 — and v1 consumers treat it as a number (`row.closedDealValue?.toString()` in ContactTable.tsx:147). This PR promotes `CrmContactMetrics` to be the canonical shape for contact metrics by wiring it into `CrmContactEntity`, so the wrong type is now baked into the v2 contract. Fix: change to `closedDealValue?: number;`. Note the sibling money fields are genuinely different — `totalRevenue`/`pipelineRevenue` are `String` on `CrmContactDetailResponseDto`, so `string` is correct for those; only `closedDealValue` is mistyped.

frontend/src/community/crm/v2/types/CrmCommonTypes.ts
```diff
@@ -91,7 +80,7 @@ export interface CrmDealEntity {
   companyId?: number;
   contactId?: number;
   ownerId?: number;
-  openTasksCount?: number;
+  taskCount?: number;
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The commit message says this rename is "for consistency", but it actually makes the file less consistent semantically, and the new name is misleading. The backend's deal `taskCount` is NOT a total task count — `CrmTaskRepositoryImpl.countTasksByDealIds` (backend/.../repository/impl/CrmTaskRepositoryImpl.java:320-321) filters on `cb.isFalse(task.get(CrmTask_.isCompleted))`, so it counts only open tasks. Meanwhile `CrmCompanyMetrics.openTasksCount` and `CrmContactMetrics.openTasksCount` keep the explicit name for the exact same concept. Matching the wire name is the right call for deserialization, so keep `taskCount`, but add a short doc comment (e.g. `/** Open (non-completed) task count — backend excludes completed tasks. */`) so consumers don't render it as "N tasks" when it means "N open tasks".

frontend/src/community/crm/v2/types/CrmCommonTypes.ts
```diff
@@ -14,12 +14,7 @@ export interface CrmCompanyEntity {
   website?: string;
   address?: string;
   contactNumber?: string;
-  openTasksCount?: number;
-  overdue?: number;
-  openValue?: string;
-  accountValue?: string;
-  openDeals?: number;
-  closedDeals?: number;
+  metrics?: CrmCompanyMetrics;
```
ThinuwanW, line 17: 🤖 **Claude** · 🟡 **Suggestion**

Now that metrics are nested inside the parent entity, the `id` field on `CrmCompanyMetrics` (line 24) and `CrmContactMetrics` (line 51) is redundant and creates two sources of truth for the same key — the backend's `CrmCompanyMetricsResponseDto.id` *is* the company id, so `company.id` and `company.metrics.id` will always hold the same value. In a normalized entity model that's exactly the kind of duplication that drifts. Either drop `id` from both metrics interfaces, or — if metrics are meant to be fetched/invalidated independently — keep the `id` and store metrics in their own `Record<number, CrmCompanyMetrics>` map instead of nesting them in the entity. Pick one; the current shape is a hybrid of both.

frontend/src/community/crm/v2/types/CrmCommonTypes.ts
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Structural inconsistency introduced by this PR: company and contact counts are now grouped under a `metrics` object, but the deal's count stays as a flat `taskCount` field on the entity. If the intent is a uniform entity shape (so the store can refresh/invalidate metrics independently of core entity data), deals should get the same treatment — `metrics?: CrmDealMetrics` with `taskCount?: number` — otherwise any generic metrics handling in the upcoming v2 store will need a special case for deals. If deals are intentionally excluded because they only carry a single derived count, a one-line comment explaining that would prevent the next person from "fixing" it.

frontend/src/community/crm/v2/types/CrmCommonTypes.ts
ThinuwanW, line 17: 🤖 **Claude** · 🟣 **Nit**

`CrmCompanyMetrics.overdue` (line 27) is vague next to `CrmContactMetrics.overdueTasksCount` (line 56) — both mean overdue task count. It matches the backend field name (`CrmCompanyMetricsResponseDto.overdue`), so renaming it client-side would require a mapper, but since this PR is explicitly a naming/consistency cleanup it's worth deciding now: either rename to `overdueTasksCount` and map it at the API boundary, or leave it and add a comment noting it mirrors the backend field. Flagging as a nit since it's pre-existing, but this PR is the natural place to settle it.

ThinuwanW, line 17: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

Context worth recording for whoever wires this up: no backend endpoint returns metrics nested under a `metrics` key. Company metrics come from the flat `CrmCompanyMetricsResponseDto` (which also carries `name`/`industry`/`website`/`address`/`contactNumber`, i.e. it is the company list payload), and contact metrics are flat on `CrmContactListItemDto`/`CrmContactDetailResponseDto`. The nesting this PR introduces is therefore a purely client-side normalization that requires an explicit mapper, and the v2 module currently has only `types/` and `enums/` — no `api/`, `store/`, or `utils/` — so nothing enforces the shape at the boundary yet. When the v2 API layer lands, add the flat→nested mapping function in `v2/utils/` with a unit test, rather than casting the raw response to `CrmCompanyEntity`/`CrmContactEntity` (an `as` cast here would silently produce entities with `metrics === undefined`).

ThinuwanW, line 17: 🤖 **Claude** · 🔴 **Important**

Null-safety drift in the refactor: every field inside the new `CrmCompanyMetrics` and `CrmContactMetrics` interfaces is declared as optional-only (`?: number` / `?: string`), i.e. `T | undefined`. But the backend serializes these as explicit JSON `null`, not as absent keys — there is no global Jackson `NON_NULL` configuration in this repo (the only `@JsonInclude(NON_NULL)` is on one enterprise BankId DTO), and the wrapper types on the backend are nullable (`Long openTasksCount`, `Long overdue`, `String openValue`, `String accountValue`, `BigDecimal closedDealValue`, `Long closedDealCount`). The v1 types this module replaces model exactly these fields as `number | null` / `string | null` (see `frontend/src/community/crm/types/CommonTypes.ts` `CrmCompany` and `CrmContact`), and the existing v1 test fixtures literally set `closedDealValue: null, closedDealCount: null`. As written, TypeScript will accept a guard like `if (metrics.openValue !== undefined)` and then hand `null` to `formatMonetaryValue`/`String(...)`, which is exactly the class of bug strict mode is supposed to prevent. Fix: declare these as `openTasksCount?: number | null` (or `number | null` if the key is always present) to match the wire contract and the v1 types.

frontend/src/community/crm/v2/types/CrmCommonTypes.ts
ThinuwanW, line 45: 🤖 **Claude** · 🟡 **Suggestion**

`CrmContactMetrics` merges fields that never arrive together from any single endpoint. `CrmContactListItemDto` returns only `closedDealValue`, `closedDealCount`, `openTasksCount`, `overdueTasksCount`; `CrmContactDetailResponseDto` returns only `totalRevenue`, `pipelineRevenue`, `activeDealsCount`, `openTasksCount`, `overdueTasksCount`. So no response ever populates all seven, and a consumer reading `metrics.totalRevenue` off a contact loaded from the list endpoint gets `undefined` with no compile-time warning — the existing v1 usage confirms this split (SidePanelCompanyContacts/ContactTable read the list set; contactUtil's metric mapper reads the detail set). Grouping the fields into a dedicated object was the right moment to split them: consider `CrmContactListMetrics` and `CrmContactDetailMetrics` (sharing the two task counts), or a discriminated union, so the type says which fields are actually available in each context. For contrast, `CrmCompanyMetrics` maps cleanly 1:1 onto `CrmCompanyMetricsResponseDto` and does not have this problem.

frontend/src/community/crm/v2/types/CrmCommonTypes.ts
ThinuwanW, line 17: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Beyond the `overdue` vs `overdueTasksCount` mismatch, the two new metrics interfaces name four more identical concepts differently: closed-deal count is `closedDeals` on company but `closedDealCount` on contact; open/active deal count is `openDeals` vs `activeDealsCount`; open pipeline value is `openValue` vs `pipelineRevenue`; and won/closed value is `accountValue` vs `closedDealValue`/`totalRevenue`. Since the whole point of this change is to make the shapes symmetric, either align the field names (and have the API layer map from the backend wire names) or extract a shared `CrmMetricsBase` with `openTasksCount`/`overdueTasksCount` plus per-entity extensions, so the two objects can be rendered by one shared metrics component instead of two near-identical mappers (the v1 `companyUtil.mapCompanyToMetricItems` and `contactUtil.mapContactToMetricItems` are already duplicated this way).

frontend/src/community/crm/v2/types/CrmCommonTypes.ts
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`taskCount` is only ever produced by one endpoint. The board-by-stage response (`CrmDealByStageItemResponseDto`) is the sole backend DTO with a `taskCount` field, and it is always populated (`taskCountMap.getOrDefault(deal.getId(), 0L)` in `CrmDealServiceImpl`). The deal list and deal detail DTOs (`CrmDealResponseDto`, `CrmDealDetailResponseDto`) have no such field, so on any `CrmDealEntity` that came from `CrmDealListResponse` this will silently always be `undefined`. Two things worth deciding here: (a) declaring it optional is inconsistent with the existing board types in `frontend/src/community/crm/types/BoardTypes.ts`, which declare `taskCount: number` as required and let `kanbanUtil` default it — pick one convention; (b) add a short comment noting the field is board-only, otherwise the next person will assume it is available on list-loaded deals.

frontend/src/community/crm/v2/types/CrmCommonTypes.ts
ThinuwanW, line 45: 🤖 **Claude** · 🟣 **Nit**

While restructuring `CrmContactEntity`, note that its `firstName`/`lastName` fields do not match any CRM contact endpoint. Both `CrmContactListItemDto` and `CrmContactDetailResponseDto` return a single `name` string, and the v1 frontend types (`CrmContactType`, `CrmContact`) model it as `name: string`. `firstName`/`lastName` only exist on the owner DTOs (`CrmOwnerResponseDto`, `CrmContactOwnerResponseDto`), which is correctly reflected in `CrmOwnerEntity` right below. Unless a backend name split is planned, this entity should use `name?: string` — otherwise the normalizer will produce contacts with permanently undefined names. Flagging as a nit since I cannot rule out a planned backend change.

frontend/src/community/crm/v2/types/CrmCommonTypes.ts
ThinuwanW, line 17: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

Nothing in the repo imports this module, so this rename has zero call sites and neither `tsc` nor the test suite can validate it. `frontend/src/community/crm/v2/` contains only three files (`enums/common.ts`, `types/CrmCommonTypes.ts`, `types/CrmTypes.ts`), the only importer of `CrmCommonTypes` is its sibling `CrmTypes.ts`, and there is no corresponding backend surface either — every CRM controller is mapped under `/v1/crm` (`CrmCompanyController`, `CrmContactController`, `CrmDealController`, `CrmBoardController`, `CrmTaskController`), with no `/v2` equivalent. That means a field rename like this one carries no verification signal today. Consider landing the v2 type churn together with the `api/`, `store/`, and `utils/` layers that consume it (per the module structure convention), so renames are checked by the compiler rather than by review alone.

frontend/src/community/crm/v2/types/CrmCommonTypes.ts
YasinduRC: v2 api's closedDealValue is String, that matches the frontend

YasinduRC: Added name for now it will be removed when the migration happens

frontend/src/community/crm/v2/types/CrmCommonTypes.ts
YasinduRC: Backend Reponse will be updated as opeTaskCount

YasinduRC: This is intentional deals do not have metrics

frontend/src/community/crm/v2/types/CrmCommonTypes.ts
YasinduRC: Known, the old strucutures will be removed after the refactor

frontend/src/community/crm/v2/types/CrmCommonTypes.ts
YasinduRC: This will be fixed from the backend side

frontend/src/community/crm/v2/types/CrmCommonTypes.ts
YasinduRC: Known it will be used on that endpoint only

frontend/src/community/crm/v2/types/CrmCommonTypes.ts
YasinduRC: The new V2 endpoint returns overdueTasksCount

YasinduRC: The type types are intention for counts we use number for monitory values we use String

YasinduRC: This is intentional because metrics are fetched through a separate metrics endpoint and have their own lifecycle. We keep the id to identify the metrics resource independently, even though it is nested in the entity for convenient access.

YasinduRC: The backend will be refactored to return the data in the same nested shape as the frontend types, so a separate client-side mapper won't be needed.

YasinduRC: The fields are intentionally defined as optional in v2 to allow them to be omitted when the data is not available.

thusala (approved):

hasalarootcode (approved):

akilarootcode (approved):

PR #1759, opened 2026-08-14
Chore/2950 leave UI issues

shakila-kamalasena:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/frontend/src/community/attendance/components/molecules/AttendanceCard/AttendanceCard.tsx
+++ b/frontend/src/community/attendance/components/molecules/AttendanceCard/AttendanceCard.tsx
@@ -31,15 +31,35 @@ const AttendanceCard: FC<Props> = ({
   const { setClockInType } = useAttendanceStore((state) => state);
   const theme = useTheme();
 
+  const handleCardClick = () => {
+    type === ClockInOutGraphTypes.CLOCK_IN
+      ? setClockInType({})
+      : setClockInType({
+          "Clock-ins": [ClockInSummaryTypes.LATE_CLOCK_INS]
+        });
+
+    router.replace(ROUTES.DASHBOARD.ATTENDANCE.CLOCK_IN_SUMMARY);
+  };
+
   return (
     <>
       <Box
+        tabIndex={0}
+        role="button"
+        aria-label={iconAriaLabel}
         sx={{
           flex: 1,
           backgroundColor: "grey.50",
           p: 1.5,
           borderRadius: 1.5,
-          height: "100%"
+          height: "100%",
+          cursor: "pointer"
+        }}
+        onClick={handleCardClick}
+        onKeyDown={(e) => {
+          if (shouldActivateButton(e.key)) {
+            handleCardClick();
+          }
         }}
       >
         <>
@@ -58,32 +78,7 @@ const AttendanceCard: FC<Props> = ({
             >
               {title}
             </Typography>
-            <Box
-              tabIndex={0}
-              role="button"
-              aria-label={iconAriaLabel}
-              sx={{ cursor: "pointer" }}
-              onClick={() => {
-                type === ClockInOutGraphTypes.CLOCK_IN
-                  ? setClockInType({})
-                  : setClockInType({
-                      "Clock-ins": [ClockInSummaryTypes.LATE_CLOCK_INS]
-                    });
-
-                router.replace(ROUTES.DASHBOARD.ATTENDANCE.CLOCK_IN_SUMMARY);
-              }}
-              onKeyDown={(e) => {
-                if (shouldActivateButton(e.key)) {
-                  type === ClockInOutGraphTypes.CLOCK_IN
-                    ? setClockInType({})
-                    : setClockInType({
-                        "Clock-ins": [ClockInSummaryTypes.LATE_CLOCK_INS]
-                      });
-
-                  router.replace(ROUTES.DASHBOARD.ATTENDANCE.CLOCK_IN_SUMMARY);
-                }
-              }}
-            >
+            <Box sx={{ cursor: "pointer" }}>
               <Icon name={IconName.NEW_WINDOW_ICON} />
             </Box>
           </Box>
--- a/frontend/src/community/attendance/components/molecules/AttendanceCard/AttendanceCard.tsx
+++ b/frontend/src/community/attendance/components/molecules/AttendanceCard/AttendanceCard.tsx
@@ -4,6 +4,7 @@ import { FC } from "react";
 
 import {
   ClockInOutGraphTypes,
+  ClockInSummaryFilterTypes,
   ClockInSummaryTypes
 } from "~community/attendance/enums/dashboardEnums";
 import { useAttendanceStore } from "~community/attendance/store/attendanceStore";
@@ -32,11 +33,13 @@ const AttendanceCard: FC<Props> = ({
   const theme = useTheme();
 
   const handleCardClick = (): void => {
-    type === ClockInOutGraphTypes.CLOCK_IN
-      ? setClockInType({})
-      : setClockInType({
-          "Clock-ins": [ClockInSummaryTypes.LATE_CLOCK_INS]
-        });
+    type !== ClockInOutGraphTypes.CLOCK_IN
+      ? setClockInType({
+          [ClockInSummaryFilterTypes.CLOCK_INS]: [
+            ClockInSummaryTypes.LATE_CLOCK_INS
+          ]
+        })
+      : setClockInType({});
 
     router.replace(ROUTES.DASHBOARD.ATTENDANCE.CLOCK_IN_SUMMARY);
   };
--- a/frontend/src/community/attendance/components/molecules/AttendanceCard/AttendanceCard.tsx
+++ b/frontend/src/community/attendance/components/molecules/AttendanceCard/AttendanceCard.tsx
@@ -31,7 +31,7 @@ const AttendanceCard: FC<Props> = ({
   const { setClockInType } = useAttendanceStore((state) => state);
   const theme = useTheme();
 
-  const handleCardClick = () => {
+  const handleCardClick = (): void => {
     type === ClockInOutGraphTypes.CLOCK_IN
       ? setClockInType({})
       : setClockInType({
@@ -78,9 +78,7 @@ const AttendanceCard: FC<Props> = ({
             >
               {title}
             </Typography>
-            <Box sx={{ cursor: "pointer" }}>
-              <Icon name={IconName.NEW_WINDOW_ICON} />
-            </Box>
+            <Icon name={IconName.NEW_WINDOW_ICON} />
           </Box>
           <Stack direction="row" justifyContent="left">
             <div style={{ display: "flex", alignItems: "baseline" }}>
--- a/frontend/src/community/common/assets/languages/english/leaveModule.json
+++ b/frontend/src/community/common/assets/languages/english/leaveModule.json
@@ -91,7 +91,10 @@
       "awayMembers": "Away Members",
       "fullTeamAvailable": "Full team available",
       "fullTeamAway": "Full team away",
-      "onLeaveTitle": "On leave : {{date}}",
+      "awayToday": "Away Today",
+      "awayYesterday": "Away Yesterday",
+      "awayTomorrow": "Away Tomorrow",
+      "awayOnDate": "Away on {{date}}",
       "goBackBtn": "Go back"
     },
     "addAttachmentModal": {
--- a/frontend/src/community/common/components/molecules/AnalyticCard/AnalyticCard.tsx
+++ b/frontend/src/community/common/components/molecules/AnalyticCard/AnalyticCard.tsx
@@ -31,9 +31,10 @@ const AnalyticCard: FC<Props> = ({
   return (
     <>
       <Box
-        sx={classes.card}
+        sx={{ ...classes.card, cursor: isExpandable ? "pointer" : "default" }}
         tabIndex={accessibility?.tabIndex ?? -1}
         role={accessibility?.role ?? "group"}
+        onClick={isExpandable ? onExpand : undefined}
         onKeyDown={(e) => {
           if (shouldActivateButton(e.key)) {
             onExpand?.();
@@ -50,7 +51,7 @@ const AnalyticCard: FC<Props> = ({
               {title}
             </Typography>
             {isExpandable && (
-              <Box sx={{ cursor: "pointer" }} onClick={onExpand}>
+              <Box sx={{ cursor: "pointer" }}>
                 <Icon name={IconName.NEW_WINDOW_ICON} />
               </Box>
             )}
--- a/frontend/src/community/common/components/molecules/AnalyticCard/AnalyticCard.tsx
+++ b/frontend/src/community/common/components/molecules/AnalyticCard/AnalyticCard.tsx
@@ -50,11 +50,7 @@ const AnalyticCard: FC<Props> = ({
             >
               {title}
             </Typography>
-            {isExpandable && (
-              <Box sx={{ cursor: "pointer" }}>
-                <Icon name={IconName.NEW_WINDOW_ICON} />
-              </Box>
-            )}
+            {isExpandable && <Icon name={IconName.NEW_WINDOW_ICON} />}
           </Box>
           <Box>{children}</Box>
         </>
--- a/frontend/src/community/leave/components/molecules/AvailabilityCalendar/AvailabilityCalendarCard.tsx
+++ b/frontend/src/community/leave/components/molecules/AvailabilityCalendar/AvailabilityCalendarCard.tsx
@@ -55,10 +55,21 @@ const AvailabilityCalendarCard = ({
     setTodaysAvailability: state.setTodaysAvailability
   }));
 
+  const getOnLeaveModalTitle = () => {
+    const today = DateTime.now().toFormat(DATE_FORMAT);
+    const yesterday = DateTime.now().minus({ days: 1 }).toFormat(DATE_FORMAT);
+    const tomorrow = DateTime.now().plus({ days: 1 }).toFormat(DATE_FORMAT);
+
+    if (actualDate === today) return translateText(["awayToday"]);
+    if (actualDate === yesterday) return translateText(["awayYesterday"]);
+    if (actualDate === tomorrow) return translateText(["awayTomorrow"]);
+    return translateText(["awayOnDate"], { date: actualDate });
+  };
+
   const handleOnLeaveModalOpen = () => {
     setIsManagerModal(true);
     setIsOnLeaveModalOpen(true);
-    setOnLeaveModalTitle(translateText(["onLeaveTitle"], { date: actualDate }));
+    setOnLeaveModalTitle(getOnLeaveModalTitle());
     setTodaysAvailability(onLeaveEmployees);
   };
 
--- a/frontend/src/community/leave/components/molecules/AvailabilityCalendar/AvailabilityCalendarCard.tsx
+++ b/frontend/src/community/leave/components/molecules/AvailabilityCalendar/AvailabilityCalendarCard.tsx
@@ -55,12 +55,12 @@ const AvailabilityCalendarCard = ({
     setTodaysAvailability: state.setTodaysAvailability
   }));
 
-  const getOnLeaveModalTitle = () => {
-    const today = DateTime.now().toFormat(DATE_FORMAT);
-    const yesterday = DateTime.now().minus({ days: 1 }).toFormat(DATE_FORMAT);
-    const tomorrow = DateTime.now().plus({ days: 1 }).toFormat(DATE_FORMAT);
+  const getOnLeaveModalTitle = (): string => {
+    const now = DateTime.now();
+    const yesterday = now.minus({ days: 1 }).toFormat(DATE_FORMAT);
+    const tomorrow = now.plus({ days: 1 }).toFormat(DATE_FORMAT);
 
-    if (actualDate === today) return translateText(["awayToday"]);
+    if (isToday()) return translateText(["awayToday"]);
     if (actualDate === yesterday) return translateText(["awayYesterday"]);
     if (actualDate === tomorrow) return translateText(["awayTomorrow"]);
     return translateText(["awayOnDate"], { date: actualDate });
--- a/frontend/src/community/leave/components/molecules/AvailabilityCalendar/AvailabilityCalendarCard.tsx
+++ b/frontend/src/community/leave/components/molecules/AvailabilityCalendar/AvailabilityCalendarCard.tsx
@@ -7,6 +7,7 @@ import { JSX } from "react";
 import AvatarGroup from "~community/common/components/molecules/AvatarGroup/AvatarGroup";
 import { DATE_FORMAT } from "~community/common/constants/timeConstants";
 import { useTranslator } from "~community/common/hooks/useTranslator";
+import { getRelativeDates } from "~community/common/utils/dateTimeUtils";
 import { shouldActivateButton } from "~community/common/utils/keyboardUtils";
 import { useLeaveStore } from "~community/leave/store/store";
 import { LeaveRequest } from "~community/leave/types/ResourceAvailabilityTypes";
@@ -56,11 +57,9 @@ const AvailabilityCalendarCard = ({
   }));
 
   const getOnLeaveModalTitle = (): string => {
-    const now = DateTime.now();
-    const yesterday = now.minus({ days: 1 }).toFormat(DATE_FORMAT);
-    const tomorrow = now.plus({ days: 1 }).toFormat(DATE_FORMAT);
+    const { today, yesterday, tomorrow } = getRelativeDates();
 
-    if (isToday()) return translateText(["awayToday"]);
+    if (actualDate === today) return translateText(["awayToday"]);
     if (actualDate === yesterday) return translateText(["awayYesterday"]);
     if (actualDate === tomorrow) return translateText(["awayTomorrow"]);
     return translateText(["awayOnDate"], { date: actualDate });
--- a/frontend/src/community/leave/components/organisms/LeaveDashboard/LeaveDashboard.tsx
+++ b/frontend/src/community/leave/components/organisms/LeaveDashboard/LeaveDashboard.tsx
@@ -92,9 +92,7 @@ const LeaveDashboard = (): JSX.Element => {
   const todaysAwayEmployees: LeaveRequest[] =
     todaysResourceAvailability?.find(
       (record: ResourceAvailabilityRecord) => record.actualDate === today
-    )?.leaveRequests ??
-    todaysResourceAvailability?.[0]?.leaveRequests ??
-    [];
+    )?.leaveRequests ?? [];
 
   const handleTodaysAvailabilityModalOpen = (): void => {
     setTodaysAvailability(todaysAwayEmployees);
--- a/frontend/src/community/leave/components/organisms/LeaveDashboard/LeaveDashboard.tsx
+++ b/frontend/src/community/leave/components/organisms/LeaveDashboard/LeaveDashboard.tsx
@@ -16,7 +16,8 @@ import {
   useGetAbsenceRate,
   useGetAvailability,
   useGetLeaveTypeBreakdownChartData,
-  useGetPendingLeavesData
+  useGetPendingLeavesData,
+  useGetResourceAvailability
 } from "~community/leave/api/LeaveDashboard";
 import AvailabilityCalendarWidget from "~community/leave/components/molecules/AvailabilityCalendar/AvailabilityCalendarWidget";
 import AvailableChip from "~community/leave/components/molecules/LeaveDashboardChips/AvailableChip";
@@ -24,6 +25,10 @@ import AwayChip from "~community/leave/components/molecules/LeaveDashboardChips/
 import HolidayChip from "~community/leave/components/molecules/LeaveDashboardChips/HolidayChip";
 import LeaveTypeBreakdownChart from "~community/leave/components/molecules/LeaveUtilizationGraph/LeaveTypeBreakdownChart";
 import { useLeaveStore } from "~community/leave/store/store";
+import {
+  LeaveRequest,
+  ResourceAvailabilityRecord
+} from "~community/leave/types/ResourceAvailabilityTypes";
 import useGoogleAnalyticsEvent from "~enterprise/common/hooks/useGoogleAnalyticsEvent";
 import { GoogleAnalyticsTypes } from "~enterprise/common/types/GoogleAnalyticsTypes";
 
@@ -31,6 +36,11 @@ import styles from "./styles";
 
 const LeaveDashboard = (): JSX.Element => {
   const translateText = useTranslator("leaveModule", "dashboard");
+  const availabilityModalText = useTranslator(
+    "leaveModule",
+    "myRequests",
+    "teamAvailabilityModal"
+  );
   const [teamId, setTeamId] = useState<string | number>("");
   const [teamName, setTeamName] = useState<string>(translateText(["all"]));
   const [isFetchingEnabled, setIsFetchingEnabled] = useState<boolean>(false);
@@ -41,12 +51,25 @@ const LeaveDashboard = (): JSX.Element => {
   const {
     viewedPendingLeaveCount,
     setViewedPendingLeaveCount,
-    setPendingLeaveCount
+    setPendingLeaveCount,
+    setIsManagerModal,
+    setIsOnLeaveModalOpen,
+    setOnLeaveModalTitle,
+    setTodaysAvailability
   } = useLeaveStore((state) => state);
+
+  const today = DateTime.now().toFormat(DATE_FORMAT);
 
   const { data: todaysAvailability } = useGetAvailability(
     teamId,
-    DateTime.now().toFormat(DATE_FORMAT),
+    today,
+    isFetchingEnabled
+  );
+
+  const { data: todaysResourceAvailability } = useGetResourceAvailability(
+    teamId,
+    today,
+    today,
     isFetchingEnabled
   );
   const { data: pendingLeaves } = useGetPendingLeavesData(teamId);
@@ -65,6 +88,20 @@ const LeaveDashboard = (): JSX.Element => {
 
   const newPendingRequests =
     pendingLeaves?.[0]?.items?.length - viewedPendingLeaveCount;
+
+  const todaysAwayEmployees: LeaveRequest[] =
+    todaysResourceAvailability?.find(
+      (record: ResourceAvailabilityRecord) => record.actualDate === today
+    )?.leaveRequests ??
+    todaysResourceAvailability?.[0]?.leaveRequests ??
+    [];
+
+  const handleTodaysAvailabilityModalOpen = () => {
+    setTodaysAvailability(todaysAwayEmployees);
+    setOnLeaveModalTitle(availabilityModalText(["awayToday"]));
+    setIsOnLeaveModalOpen(true);
+    setIsManagerModal(true);
+  };
 
   useEffect(() => {
     if (user) setIsFetchingEnabled(true);
@@ -138,7 +175,15 @@ const LeaveDashboard = (): JSX.Element => {
       </Box>
       <Grid container spacing={1}>
         <Grid sx={{ width: { xs: "100%", md: "32.5%" } }}>
-          <AnalyticCard title={translateText(["todayAvailability"]) ?? ""}>
+          <AnalyticCard
+            title={translateText(["todayAvailability"]) ?? ""}
+            isExpandable={!!todaysAvailability}
+            onExpand={handleTodaysAvailabilityModalOpen}
+            accessibility={{
+              tabIndex: 0,
+              role: "button"
+            }}
+          >
             {todaysAvailability && resourceDetails()}
           </AnalyticCard>
         </Grid>
--- a/frontend/src/community/leave/components/organisms/LeaveDashboard/LeaveDashboard.tsx
+++ b/frontend/src/community/leave/components/organisms/LeaveDashboard/LeaveDashboard.tsx
@@ -96,7 +96,7 @@ const LeaveDashboard = (): JSX.Element => {
     todaysResourceAvailability?.[0]?.leaveRequests ??
     [];
 
-  const handleTodaysAvailabilityModalOpen = () => {
+  const handleTodaysAvailabilityModalOpen = (): void => {
     setTodaysAvailability(todaysAwayEmployees);
     setOnLeaveModalTitle(availabilityModalText(["awayToday"]));
     setIsOnLeaveModalOpen(true);
--- a/frontend/src/community/leave/components/organisms/LeaveManagerModalController/LeaveManagerModalController.tsx
+++ b/frontend/src/community/leave/components/organisms/LeaveManagerModalController/LeaveManagerModalController.tsx
@@ -139,7 +139,7 @@ const LeaveManagerModalController = () => {
   if (popupType === LeaveExtraPopupTypes.ON_LEAVE_MODAL) {
     return (
       <LargeModal
-        isOpen={isManagerModalOpen && !!popupType}
+        isOpen={isManagerModalOpen}
         id="on-leave-modal"
         onClose={handelManagerModal}
         modalHeader={modalTitle}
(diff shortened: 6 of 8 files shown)


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR makes dashboard analytic/attendance cards clickable as a whole and switches the ON_LEAVE_MODAL to `LargeModal`, plus adds relative-day modal titles. The refactors are mostly clean, but `AnalyticCard` now gates `onClick` on `isExpandable` while leaving `onKeyDown` ungated, so keyboard users can trigger expand actions on non-expandable cards (e.g. navigating to the pending-leaves page when there are zero pending requests). Additionally, `LeaveDashboard` gates the new Today's-Availability expander on a different query (`todaysAvailability`) than the one that supplies the modal rows (`todaysResourceAvailability`), so the modal can open empty, and the new `LargeModal` omits the width/backdrop props every sibling controller passes.

Found **24** new issue(s): 🔴 3 important, 🟡 11 suggestion(s), 🟣 10 nit(s)

frontend/src/community/common/components/molecules/AnalyticCard/AnalyticCard.tsx
```diff
@@ -31,9 +31,10 @@ const AnalyticCard: FC<Props> = ({
   return (
     <>
       <Box
-        sx={classes.card}
+        sx={{ ...classes.card, cursor: isExpandable ? "pointer" : "default" }}
         tabIndex={accessibility?.tabIndex ?? -1}
         role={accessibility?.role ?? "group"}
+        onClick={isExpandable ? onExpand : undefined}
         onKeyDown={(e) => {
```
ThinuwanW, line 38: 🤖 **Claude** · 🔴 **Important**

Mouse and keyboard activation are now inconsistent. `onClick` was changed to `isExpandable ? onExpand : undefined`, but `onKeyDown` still calls `onExpand?.()` unconditionally. Because `LeaveDashboard` passes `accessibility={{ tabIndex: 0, role: "button" }}` unconditionally while `isExpandable` is conditional, a non-expandable card is still tab-focusable and Enter/Space fires the action: with zero pending leaves the Pending Leaves card ignores a click but on Enter runs `setPendingLeaveCount(...)`/`setViewedPendingLeaveCount(...)` and `router.replace(ROUTES.LEAVE.LEAVE_PENDING)`; the Today's Availability card opens an empty modal while data is still loading. Fix by gating the keyboard handler the same way — `onKeyDown={isExpandable ? (e) => { if (shouldActivateButton(e.key)) onExpand?.(); } : undefined}` — and preferably derive the interactive a11y attributes from `isExpandable` too (e.g. `tabIndex={isExpandable ? (accessibility?.tabIndex ?? 0) : -1}`, `role={isExpandable ? (accessibility?.role ?? "button") : "group"}`) so a non-actionable card is never announced as a button.

frontend/src/community/leave/components/organisms/LeaveDashboard/LeaveDashboard.tsx
```diff
@@ -138,7 +175,15 @@ const LeaveDashboard = (): JSX.Element => {
       </Box>
       <Grid container spacing={1}>
         <Grid sx={{ width: { xs: "100%", md: "32.5%" } }}>
-          <AnalyticCard title={translateText(["todayAvailability"]) ?? ""}>
+          <AnalyticCard
+            title={translateText(["todayAvailability"]) ?? ""}
+            isExpandable={!!todaysAvailability}
```
ThinuwanW, line 178: 🤖 **Claude** · 🔴 **Important**

`isExpandable={!!todaysAvailability}` is gated on the wrong data source. `todaysAvailability` comes from `useGetAvailability` (counts only), while the modal is populated from `todaysAwayEmployees`, which is derived from the separate `useGetResourceAvailability` query. Since `select` returns `data?.results`, an empty array is truthy, so the card advertises itself as expandable and opens the OnLeaveModal with `[]` whenever nobody is away, today is a holiday, or the resource-availability query has not resolved yet. Gate on the actual modal payload instead: `isExpandable={todaysAwayEmployees.length > 0}` (this also matches the sibling card's `pendingLeaves?.[0]?.items?.length > 0` pattern) and pass the `accessibility` prop only when the card is actually expandable.

frontend/src/community/leave/components/organisms/LeaveDashboard/LeaveDashboard.tsx
```diff
@@ -66,6 +89,20 @@ const LeaveDashboard = (): JSX.Element => {
   const newPendingRequests =
     pendingLeaves?.[0]?.items?.length - viewedPendingLeaveCount;
 
+  const todaysAwayEmployees: LeaveRequest[] =
+    todaysResourceAvailability?.find(
+      (record: ResourceAvailabilityRecord) => record.actualDate === today
+    )?.leaveRequests ??
+    todaysResourceAvailability?.[0]?.leaveRequests ??
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The `?? todaysResourceAvailability?.[0]?.leaveRequests` fallback can silently show the wrong day under an "Away Today" title. The query is called with `startDate === endDate === today`, so `find(record => record.actualDate === today)` should always match; if it ever doesn't (date-format drift, backend returning a wider range), falling back to index 0 renders another day's employees while the modal header still says today. Drop the index-0 fallback and keep `?? []`. Also, `todaysResourceAvailability` is untyped (`useQuery` has no generics, so it is `any`) — the `(record: ResourceAvailabilityRecord)` annotation is only a local assumption. Type the hook's return as `ResourceAvailabilityRecord[]` in `~community/leave/api/LeaveDashboard.ts` so this is checked, and wrap the derivation in `useMemo` since it re-runs on every render.

frontend/src/community/leave/components/molecules/AvailabilityCalendar/AvailabilityCalendarCard.tsx
```diff
@@ -55,10 +55,21 @@ const AvailabilityCalendarCard = ({
     setTodaysAvailability: state.setTodaysAvailability
   }));
 
+  const getOnLeaveModalTitle = () => {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The relative-day title logic is duplicated in spirit across two files: `getOnLeaveModalTitle()` here resolves today/yesterday/tomorrow/other, while `LeaveDashboard.handleTodaysAvailabilityModalOpen` hardcodes `availabilityModalText(["awayToday"])`. Extract a shared helper (e.g. `getAwayModalTitle(actualDate, translateText)` in `~community/leave/utils/dashboardUtils`) and call it from both, so the labelling stays consistent if a case is added later. Also note `awayOnDate` interpolates `actualDate` in raw `yyyy-MM-dd` form — consider formatting it with `MEDIUM_DATE_FORMAT` for a user-facing string (this matches the pre-existing behaviour of the removed `onLeaveTitle` key, so it is not a regression, just an opportunity).

frontend/src/community/leave/components/organisms/LeaveManagerModalController/LeaveManagerModalController.tsx
```diff
@@ -136,6 +136,18 @@ const LeaveManagerModalController = () => {
     return null;
   };
 
+  if (popupType === LeaveExtraPopupTypes.ON_LEAVE_MODAL) {
+    return (
+      <LargeModal
```
ThinuwanW, line 141: 🤖 **Claude** · 🟡 **Suggestion**

This `LargeModal` omits the sizing/backdrop props that every other `LargeModal` usage in the codebase passes: `MyRequestModalController` and `PolicyLeaveModalController` use `backdropVariant="dark"` + `className="w-[75vw] max-w-[1100px]"`, and `DirectoryPopupController` uses an explicit width/max-height with `overflow-hidden`. `OnLeaveModal` renders a 4-column table (name / type / duration / reviewer), so without an explicit width and `max-h-[85vh] overflow-hidden` it will fall back to the component default and can render too narrow or overflow the viewport when many employees are away. Add the same `className`/`backdropVariant` props for consistency, and verify against a long list.

frontend/src/community/attendance/components/molecules/AttendanceCard/AttendanceCard.tsx
```diff
@@ -31,15 +31,35 @@ const AttendanceCard: FC<Props> = ({
...
+
+    router.replace(ROUTES.DASHBOARD.ATTENDANCE.CLOCK_IN_SUMMARY);
+  };
+
   return (
     <>
       <Box
+        tabIndex={0}
+        role="button"
+        aria-label={iconAriaLabel}
```
ThinuwanW, line 52: 🤖 **Claude** · 🟡 **Suggestion**

`aria-label={iconAriaLabel}` moved from the small icon wrapper to the whole card. On an element with `role="button"`, `aria-label` overrides the element's inner text for the accessible name, so screen-reader users now hear only "Today's clock ins extended view" and lose the card title and the clock-in counts that sighted users see. Either compose a card-level label (e.g. `aria-label={`${title}: ${analytic1 ?? "--"}${type === ClockInOutGraphTypes.CLOCK_IN ? ` of ${analytic2 ?? "--"}` : ""}. ${iconAriaLabel}`}`) or use `aria-describedby`/`aria-labelledby` pointing at the title and value nodes. The prop name `iconAriaLabel` is now misleading too — consider renaming it to `cardAriaLabel`.

frontend/src/community/common/components/molecules/AnalyticCard/AnalyticCard.tsx
```diff
@@ -50,7 +51,7 @@ const AnalyticCard: FC<Props> = ({
               {title}
             </Typography>
             {isExpandable && (
-              <Box sx={{ cursor: "pointer" }} onClick={onExpand}>
+              <Box sx={{ cursor: "pointer" }}>
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Dead styling: this icon wrapper (and the equivalent one at `AttendanceCard.tsx:81`) keeps `sx={{ cursor: "pointer" }}` even though its `onClick` was removed and the parent card now sets the cursor itself (`cursor: isExpandable ? "pointer" : "default"` here, `cursor: "pointer"` in AttendanceCard). Remove the redundant `sx` and render the `Icon` directly, or keep a plain `<Box>` if the flex layout needs the wrapper.

frontend/src/community/leave/components/molecules/AvailabilityCalendar/AvailabilityCalendarCard.tsx
```diff
@@ -55,10 +55,21 @@ const AvailabilityCalendarCard = ({
...
+    const today = DateTime.now().toFormat(DATE_FORMAT);
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`getOnLeaveModalTitle` calls `DateTime.now()` three times, which can theoretically straddle a midnight boundary and, more practically, is redundant work. Hoist it: `const now = DateTime.now();` then derive `now.toFormat(DATE_FORMAT)`, `now.minus({ days: 1 }).toFormat(DATE_FORMAT)`, `now.plus({ days: 1 }).toFormat(DATE_FORMAT)`. Note the component already has an `isToday()` helper doing the same `DateTime.now().toFormat(DATE_FORMAT)` comparison — reuse it for the today branch.

frontend/src/community/leave/components/organisms/LeaveDashboard/LeaveDashboard.tsx
```diff
@@ -41,12 +51,25 @@ const LeaveDashboard = (): JSX.Element => {
...
+  const today = DateTime.now().toFormat(DATE_FORMAT);
+
   const { data: todaysAvailability } = useGetAvailability(
     teamId,
-    DateTime.now().toFormat(DATE_FORMAT),
+    today,
+    isFetchingEnabled
+  );
+
+  const { data: todaysResourceAvailability } = useGetResourceAvailability(
```
ThinuwanW, line 69: 🤖 **Claude** · 🟣 **Nit**

This adds a second request to the same `GET_RESOURCE_AVAILABILITY` endpoint. `AvailabilityCalendarWidget`, rendered by this very component, already calls `useGetResourceAvailability(teams, startDate, endDate)` with `startDate` initialised to today and `endDate` two weeks out — so today's record is already in the cache under a different query key. Consider lifting the widget's date-range state into `LeaveDashboard` (or exposing today's record from a shared hook) so the modal payload is derived from the existing cached data instead of firing a duplicate network call on every dashboard load.

frontend/src/community/common/components/molecules/AnalyticCard/AnalyticCard.tsx
```diff
@@ -31,9 +31,10 @@ const AnalyticCard: FC<Props> = ({
   return (
     <>
       <Box
-        sx={classes.card}
+        sx={{ ...classes.card, cursor: isExpandable ? "pointer" : "default" }}
```
ThinuwanW, line 34: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

No test coverage for the new interaction. `AnalyticCard.test.tsx` only asserts title/children rendering and `AttendanceCard.test.tsx` only asserts text output — neither covers the behaviour this PR introduces. Add tests using `@testing-library/react` + `MockTheme`: (1) `AnalyticCard` with `isExpandable` and `accessibility={{tabIndex:0, role:"button"}}` — `fireEvent.click(screen.getByRole("button"))` calls `onExpand`; (2) `AnalyticCard` with `isExpandable={false}` — neither click nor `keyDown({key:"Enter"})` calls `onExpand` (this test would fail today, catching the keyboard-gating bug above); (3) `AttendanceCard` — clicking anywhere on the card calls `setClockInType` with the right payload and `router.replace(ROUTES.DASHBOARD.ATTENDANCE.CLOCK_IN_SUMMARY)` for both `CLOCK_IN` and non-`CLOCK_IN` types.

frontend/src/community/leave/components/organisms/LeaveManagerModalController/LeaveManagerModalController.tsx
```diff
@@ -136,6 +136,18 @@ const LeaveManagerModalController = () => {
     return null;
   };
 
+  if (popupType === LeaveExtraPopupTypes.ON_LEAVE_MODAL) {
```
ThinuwanW, line 139: 🤖 **Claude** · 🟣 **Nit**

The title effect at line 87 has deps `[popupType, isManagerModalOpen]` and reads `onLeaveModalTitle` from the store without listing it. In the current flows the title is set before `isOnLeaveModalOpen` flips, so `popupType` changes in the same batch and the effect picks up the fresh value — but now that two different sources on the same page (this dashboard's AnalyticCard and the AvailabilityCalendarCard) write `onLeaveModalTitle`, any future flow that changes the title while `popupType` stays `ON_LEAVE_MODAL` would render a stale header. Add `onLeaveModalTitle` to the dependency array, or read it directly in the render for the ON_LEAVE_MODAL branch instead of mirroring it into `modalTitle` state.

frontend/src/community/leave/components/organisms/LeaveDashboard/LeaveDashboard.tsx
```diff
@@ -66,6 +89,20 @@ const LeaveDashboard = (): JSX.Element => {
   const newPendingRequests =
     pendingLeaves?.[0]?.items?.length - viewedPendingLeaveCount;
 
+  const todaysAwayEmployees: LeaveRequest[] =
```
ThinuwanW, line 92: 🤖 **Claude** · 🔴 **Important**

Backend contract mismatch on non-working days. `LeaveServiceImpl.getResourceAvailabilityCalendar` builds one record per day only when `timeConfigDao.findByDay(dayOfWeek) != null` — days that are not configured working days are skipped entirely. With startDate == endDate == today, the response is therefore an **empty array** on every weekend / non-configured day. Meanwhile the *other* endpoint backing the card (`/leave/analytics/onleave`) explicitly returns a synthetic `"Day Off!"` holiday DTO for the same day, so the card still renders (HolidayChip) and is clickable. Result: on a Saturday the user clicks the card and gets a modal titled "Away Today" showing the table's `allAvailable` ("All available") empty state — factually wrong copy for a non-working day. Fix: derive both the expandability and the modal from the resource-availability record itself, e.g. `const todaysRecord = todaysResourceAvailability?.find(r => r.actualDate === today)` and only enable expansion when `todaysRecord` exists (and short-circuit for holidays, which `todaysAvailability[0]?.holidayResponseDtos` already tells you).

frontend/src/community/leave/components/organisms/LeaveDashboard/LeaveDashboard.tsx
```diff
@@ -138,7 +175,15 @@ const LeaveDashboard = (): JSX.Element => {
       </Box>
       <Grid container spacing={1}>
         <Grid sx={{ width: { xs: "100%", md: "32.5%" } }}>
-          <AnalyticCard title={translateText(["todayAvailability"]) ?? ""}>
+          <AnalyticCard
```
ThinuwanW, line 176: 🤖 **Claude** · 🟡 **Suggestion**

The number on the card and the number of rows in the modal are produced by two different queries with different aggregation semantics, so they can disagree. `EmployeeRepositoryImpl.findAllEmployeesOnLeave` selects `criteriaBuilder.countDistinct(employee)` for `adminOnLeaveDto.onLeaveCount`, while `LeaveRequestRepositoryImpl.getEmployeesOnLeaveByTeamAndDate` returns raw `LeaveRequest` rows (both filter on APPROVED + PENDING overlapping the date). An employee holding two requests that both cover today (e.g. a morning half-day and an afternoon half-day) counts as 1 on the card but renders 2 rows in the modal, so the card reads "1 Away" and the modal lists two entries for the same person. Either dedupe by `employee.employeeId` before calling `setTodaysAvailability(...)`, or take the count from the same resource-availability record (`leaveCount` / `leaveRequests.length`) so both surfaces share one source of truth.

frontend/src/community/common/assets/languages/english/leaveModule.json
```diff
@@ -91,7 +91,10 @@
       "awayMembers": "Away Members",
       "fullTeamAvailable": "Full team available",
       "fullTeamAway": "Full team away",
-      "onLeaveTitle": "On leave : {{date}}",
+      "awayToday": "Away Today",
```
ThinuwanW, line 94: 🤖 **Claude** · 🟡 **Suggestion**

Namespace placement: `awayToday` / `awayYesterday` / `awayTomorrow` / `awayOnDate` are added under `leaveModule.myRequests.teamAvailabilityModal`, a section that otherwise belongs to the apply-leave "Team availability" modal (`MyLeaveRequestsModals/TeamAvailabilityModal` consumes `date`, `awayMembers`, `fullTeamAvailable`, `fullTeamAway`, `goBackBtn` there). The new keys are consumed by the leave *dashboard* and the availability calendar card, which is why `LeaveDashboard` now has to instantiate a second translator (`useTranslator("leaveModule", "myRequests", "teamAvailabilityModal")`) alongside its existing `leaveModule.dashboard` one. Move these four keys to a section that matches their consumers (e.g. `leaveModule.dashboard.onLeaveModal`) and have both `LeaveDashboard` and `AvailabilityCalendarCard` read from it — that also removes the need for the extra translator hook.

frontend/src/community/leave/components/organisms/LeaveDashboard/LeaveDashboard.tsx
```diff
@@ -41,12 +51,25 @@ const LeaveDashboard = (): JSX.Element => {
   const {
     viewedPendingLeaveCount,
     setViewedPendingLeaveCount,
-    setPendingLeaveCount
+    setPendingLeaveCount,
```
ThinuwanW, line 54: 🤖 **Claude** · 🟡 **Suggestion**

`useLeaveStore((state) => state)` subscribes this organism to the entire leave store. Now that the component also *writes* modal state (`setIsOnLeaveModalOpen`, `setIsManagerModal`, `setOnLeaveModalTitle`, `setTodaysAvailability`), opening the modal mutates the store this component is fully subscribed to, re-rendering `LeaveDashboard` and its whole subtree (including `AvailabilityCalendarWidget` and the charts) on every modal open/close. `AvailabilityCalendarCard` in this same PR does it correctly — use the same scoped selector shape: `useLeaveStore((state) => ({ setIsManagerModal: state.setIsManagerModal, setIsOnLeaveModalOpen: state.setIsOnLeaveModalOpen, ... }))`.

frontend/src/community/leave/components/organisms/LeaveDashboard/LeaveDashboard.tsx
```diff
@@ -66,6 +89,20 @@ const LeaveDashboard = (): JSX.Element => {
   const newPendingRequests =
     pendingLeaves?.[0]?.items?.length - viewedPendingLeaveCount;
 
+  const todaysAwayEmployees: LeaveRequest[] =
+    todaysResourceAvailability?.find(
```
ThinuwanW, line 93: 🤖 **Claude** · 🟡 **Suggestion**

Type safety is only cosmetic here. `useGetResourceAvailability` has no generic and its `select` returns `data?.results` untyped, so `todaysResourceAvailability` is `any`; the `: LeaveRequest[]` annotation on `todaysAwayEmployees` and the `(record: ResourceAvailabilityRecord)` parameter annotation are unchecked assertions that will not catch a backend shape change (e.g. `actualDate` returned in a different format silently makes `find` return `undefined`). Type the hook instead — `useQuery<ResourceAvailabilityApiResponse, Error, ResourceAvailabilityRecord[]>(...)` in `~community/leave/api/LeaveDashboard.ts` — so all consumers (this file and `AvailabilityCalendarWidget`) get real checking.

frontend/src/community/common/components/molecules/AnalyticCard/AnalyticCard.tsx
```diff
@@ -31,9 +31,10 @@ const AnalyticCard: FC<Props> = ({
...
         tabIndex={accessibility?.tabIndex ?? -1}
         role={accessibility?.role ?? "group"}
+        onClick={isExpandable ? onExpand : undefined}
```
ThinuwanW, line 37: 🤖 **Claude** · 🟡 **Suggestion**

The card is now a click target with `role="button"` / `tabIndex={0}` but has no accessible name: `Props.accessibility` only exposes `tabIndex` and `role`, so the accessible name falls back to the concatenated card contents ("Today availability 12 Available 3 Away"), and the decorative `NEW_WINDOW_ICON` contributes to it. This is inconsistent with `AttendanceCard`, which passes an aria label, and with the established convention in `src/community/common/assets/languages/english/aria/leave.json` (which already has a `dashboard` section and labels ending in "Click to open modal"). Add `ariaLabel?: string` to the `accessibility` prop, render it as `aria-label`, and supply a new key under `leaveAria.dashboard` from `LeaveDashboard`. `AvailabilityCalendarCard`'s `role="button"` Box has the same gap and can reuse the key.

frontend/src/community/attendance/components/molecules/AttendanceCard/AttendanceCard.tsx
```diff
@@ -31,15 +31,35 @@ const AttendanceCard: FC<Props> = ({
...
           flex: 1,
           backgroundColor: "grey.50",
           p: 1.5,
           borderRadius: 1.5,
-          height: "100%"
+          height: "100%",
+          cursor: "pointer"
+        }}
+        onClick={handleCardClick}
+        onKeyDown={(e) => {
```
ThinuwanW, line 62: 🤖 **Claude** · 🟡 **Suggestion**

`onKeyDown` activates on Space (`shouldActivateButton` matches ENTER and SPACE) but never calls `e.preventDefault()`. On a `<div role="button">` the browser's default Space behaviour (page scroll) still fires, so keyboard users get a scroll jump on activation. The codebase's other card-shaped buttons do prevent it — see `DealCard.tsx:77` and `PdfTopBar.tsx:51`. Add `e.preventDefault()` before invoking the handler here, and in the same-PR handlers at `AnalyticCard.tsx:38` and `AvailabilityCalendarCard.tsx:165` (the latter is now a much larger keyboard target than the old icon-only one).

frontend/src/community/leave/components/organisms/LeaveManagerModalController/LeaveManagerModalController.tsx
```diff
@@ -136,6 +136,18 @@ const LeaveManagerModalController = () => {
...
+    return (
+      <LargeModal
+        isOpen={isManagerModalOpen && !!popupType}
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Dead condition: this JSX is only reached inside `if (popupType === LeaveExtraPopupTypes.ON_LEAVE_MODAL)`, so `!!popupType` is unconditionally `true`. It was copied from the `SmallModal` return where the guard is meaningful. Simplify to `isOpen={isManagerModalOpen}` so the intent ("open while the manager modal is open") is not obscured by an always-true clause.

frontend/src/community/leave/components/molecules/AvailabilityCalendar/AvailabilityCalendarCard.tsx
```diff
@@ -55,10 +55,21 @@ const AvailabilityCalendarCard = ({
...
+    const yesterday = DateTime.now().minus({ days: 1 }).toFormat(DATE_FORMAT);
+    const tomorrow = DateTime.now().plus({ days: 1 }).toFormat(DATE_FORMAT);
+
+    if (actualDate === today) return translateText(["awayToday"]);
+    if (actualDate === yesterday) return translateText(["awayYesterday"]);
+    if (actualDate === tomorrow) return translateText(["awayTomorrow"]);
+    return translateText(["awayOnDate"], { date: actualDate });
```
ThinuwanW, line 65: 🤖 **Claude** · 🟣 **Nit**

`awayOnDate` interpolates `actualDate`, which is the raw API `yyyy-MM-dd` value (`DATE_FORMAT`), so the modal header for any day beyond yesterday/today/tomorrow reads "Away on 2026-08-14". This carries over from the old `onLeaveTitle`, but since the header is being reworked anyway it is the right moment to format it for humans — e.g. `DateTime.fromFormat(actualDate, DATE_FORMAT).toFormat(MEDIUM_DATE_FORMAT)` ("14 Aug 2026"), which is already exported from `~community/common/constants/timeConstants`.

frontend/src/community/leave/components/molecules/AvailabilityCalendar/AvailabilityCalendarCard.tsx
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Missing explicit return type on the new `getOnLeaveModalTitle` (guideline: every function's return value must be explicitly typed under strict mode). Declare it `(): string`. The new `handleTodaysAvailabilityModalOpen` in `LeaveDashboard.tsx:99` and `handleCardClick` in `AttendanceCard.tsx:34` have the same gap — annotate them `(): void`.

ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

Untranslated user-facing strings remain in this file even though the PR is specifically adding i18n keys to the very namespace it already uses: `resourceDetails()` hardcodes `"Away"`, `"All available"` and `"Full team away"`, while `leaveModule.myRequests.teamAvailabilityModal` already defines `fullTeamAvailable` and `fullTeamAway` (currently referenced only by `TeamAvailabilityModal` / `PolicyTeamAvailabilityStatus`) and `teamAvailabilityCard.away` exists too. Wrap them with the `translateText` already in scope while touching this file.

frontend/src/community/leave/components/organisms/LeaveDashboard/LeaveDashboard.tsx
```diff
@@ -66,6 +89,20 @@ const LeaveDashboard = (): JSX.Element => {
...
+      (record: ResourceAvailabilityRecord) => record.actualDate === today
+    )?.leaveRequests ??
+    todaysResourceAvailability?.[0]?.leaveRequests ??
+    [];
+
+  const handleTodaysAvailabilityModalOpen = () => {
+    setTodaysAvailability(todaysAwayEmployees);
```
ThinuwanW, line 98: 🤖 **Claude** · 🟣 **Nit**

`setTodaysAvailability(todaysAwayEmployees)` snapshots React Query data into Zustand, so the modal renders a frozen copy: if `useGetResourceAvailability` refetches while the modal is open (window focus, team change), the list behind the modal goes stale and nothing reconciles it. The guidelines call out "state that belongs in the React Query server cache being duplicated in Zustand". This mirrors what `AvailabilityCalendarCard` already does, so it is not a regression — but since a second producer is being added here, consider storing only the selector inputs (date + teamId) in the store and letting `OnLeaveModal` read the cached query itself.

frontend/src/community/attendance/components/molecules/AttendanceCard/AttendanceCard.tsx
```diff
@@ -31,15 +31,35 @@ const AttendanceCard: FC<Props> = ({
...
   const theme = useTheme();
 
+  const handleCardClick = () => {
+    type === ClockInOutGraphTypes.CLOCK_IN
+      ? setClockInType({})
+      : setClockInType({
+          "Clock-ins": [ClockInSummaryTypes.LATE_CLOCK_INS]
+        });
+
+    router.replace(ROUTES.DASHBOARD.ATTENDANCE.CLOCK_IN_SUMMARY);
```
ThinuwanW, line 44: 🤖 **Claude** · 🟣 **Nit**

`router.replace` was appropriate when only the small "open in new window" icon was clickable; now the entire card navigates, so an accidental click anywhere on it drops the dashboard entry from history and the browser Back button skips past it. Consider `router.push` for the card-wide target so users can return to the dashboard.

shakila-kamalasena: Just kept the existing logics as it is

frontend/src/community/attendance/components/molecules/AttendanceCard/AttendanceCard.tsx
shakila-kamalasena: This is needed

frontend/src/community/attendance/components/molecules/AttendanceCard/AttendanceCard.tsx
shakila-kamalasena: Not needed

frontend/src/community/common/assets/languages/english/leaveModule.json
shakila-kamalasena: Just a rename, and didn't changed the namespaces

frontend/src/community/common/components/molecules/AnalyticCard/AnalyticCard.tsx
shakila-kamalasena: No need tests

frontend/src/community/common/components/molecules/AnalyticCard/AnalyticCard.tsx
shakila-kamalasena: No need

frontend/src/community/common/components/molecules/AnalyticCard/AnalyticCard.tsx
shakila-kamalasena: Not needed to add like this

frontend/src/community/common/components/molecules/AnalyticCard/AnalyticCard.tsx
shakila-kamalasena: Fixed

frontend/src/community/leave/components/molecules/AvailabilityCalendar/AvailabilityCalendarCard.tsx
shakila-kamalasena: Only one usage is there

shakila-kamalasena: Added types

shakila-kamalasena: Not related to this PR

frontend/src/community/leave/components/molecules/AvailabilityCalendar/AvailabilityCalendarCard.tsx
shakila-kamalasena: Fixed

frontend/src/community/leave/components/molecules/AvailabilityCalendar/AvailabilityCalendarCard.tsx
shakila-kamalasena: Kept as per the requirement

frontend/src/community/leave/components/organisms/LeaveDashboard/LeaveDashboard.tsx
shakila-kamalasena: Not an issue

frontend/src/community/leave/components/organisms/LeaveDashboard/LeaveDashboard.tsx
shakila-kamalasena: This is needed

frontend/src/community/leave/components/organisms/LeaveDashboard/LeaveDashboard.tsx
shakila-kamalasena: Not related to this PR

frontend/src/community/leave/components/organisms/LeaveDashboard/LeaveDashboard.tsx
shakila-kamalasena: Used the same existing behaviour

frontend/src/community/leave/components/organisms/LeaveDashboard/LeaveDashboard.tsx
shakila-kamalasena: Kept as the existing behaviour

frontend/src/community/leave/components/organisms/LeaveDashboard/LeaveDashboard.tsx
shakila-kamalasena: Not an issue

frontend/src/community/leave/components/organisms/LeaveDashboard/LeaveDashboard.tsx
shakila-kamalasena: Not related to this PR

frontend/src/community/leave/components/organisms/LeaveDashboard/LeaveDashboard.tsx
shakila-kamalasena: Works as needed

frontend/src/community/leave/components/organisms/LeaveManagerModalController/LeaveManagerModalController.tsx
shakila-kamalasena: Works as needed, no need to add to dependency array

frontend/src/community/leave/components/organisms/LeaveManagerModalController/LeaveManagerModalController.tsx
shakila-kamalasena: No need

frontend/src/community/leave/components/organisms/LeaveManagerModalController/LeaveManagerModalController.tsx
shakila-kamalasena: Fixed

frontend/src/community/attendance/components/molecules/AttendanceCard/AttendanceCard.tsx
```diff
@@ -31,15 +31,35 @@ const AttendanceCard: FC<Props> = ({
   const { setClockInType } = useAttendanceStore((state) => state);
   const theme = useTheme();
 
+  const handleCardClick = (): void => {
+    type === ClockInOutGraphTypes.CLOCK_IN
+      ? setClockInType({})
```
hasalarootcode: add the empty one for the false state

frontend/src/community/attendance/components/molecules/AttendanceCard/AttendanceCard.tsx
```diff
@@ -31,15 +31,35 @@ const AttendanceCard: FC<Props> = ({
...
+      : setClockInType({
```
hasalarootcode: move that key to a enum

frontend/src/community/leave/components/organisms/LeaveDashboard/LeaveDashboard.tsx
```diff
@@ -66,6 +89,20 @@ const LeaveDashboard = (): JSX.Element => {
   const newPendingRequests =
     pendingLeaves?.[0]?.items?.length - viewedPendingLeaveCount;
 
+  const todaysAwayEmployees: LeaveRequest[] =
+    todaysResourceAvailability?.find(
+      (record: ResourceAvailabilityRecord) => record.actualDate === today
+    )?.leaveRequests ??
```
hasalarootcode: handle this in a better readable way

frontend/src/community/leave/components/organisms/LeaveDashboard/LeaveDashboard.tsx
```diff
@@ -138,7 +175,15 @@ const LeaveDashboard = (): JSX.Element => {
       </Box>
       <Grid container spacing={1}>
         <Grid sx={{ width: { xs: "100%", md: "32.5%" } }}>
-          <AnalyticCard title={translateText(["todayAvailability"]) ?? ""}>
+          <AnalyticCard
+            title={translateText(["todayAvailability"]) ?? ""}
+            isExpandable={!!todaysAvailability}
+            onExpand={handleTodaysAvailabilityModalOpen}
+            accessibility={{
+              tabIndex: 0,
```
hasalarootcode, line 181: Why tab index 0

hasalarootcode (changes requested):

frontend/src/community/attendance/components/molecules/AttendanceCard/AttendanceCard.tsx
shakila-kamalasena: Done

frontend/src/community/attendance/components/molecules/AttendanceCard/AttendanceCard.tsx
shakila-kamalasena: Addressed by using an enum

frontend/src/community/leave/components/organisms/LeaveDashboard/LeaveDashboard.tsx
shakila-kamalasena: Needs tabindex

frontend/src/community/attendance/components/molecules/AttendanceCard/AttendanceCard.tsx
shakila-kamalasena: Done

frontend/src/community/leave/components/organisms/LeaveDashboard/LeaveDashboard.tsx
shakila-kamalasena: Fixed by removing the redundant fallback

hasalarootcode (dismissed):

frontend/src/community/leave/components/molecules/AvailabilityCalendar/AvailabilityCalendarCard.tsx
```diff
@@ -55,10 +55,21 @@ const AvailabilityCalendarCard = ({
     setTodaysAvailability: state.setTodaysAvailability
   }));
 
+  const getOnLeaveModalTitle = (): string => {
+    const now = DateTime.now();
+    const yesterday = now.minus({ days: 1 }).toFormat(DATE_FORMAT);
+    const tomorrow = now.plus({ days: 1 }).toFormat(DATE_FORMAT);
```
thusala: move to date utils

thusala (changes requested):

thusala (approved):

frontend/src/community/leave/components/molecules/AvailabilityCalendar/AvailabilityCalendarCard.tsx
shakila-kamalasena: Done

akilarootcode (approved):

PR #1761, opened 2026-08-14
refactor(crm): flatten v2 company & contact metrics list responses

anusham-rc:
## What

Flattens the two v2 metrics **list** responses from composition wrappers to flat DTOs that carry the entity fields directly plus a nested `metrics` object.

| Endpoint | Before | After |
|---|---|---|
| `GET /v2/crm/company` | `CrmCompanyListItemDtoV2` = `{ company, metrics }` | `CrmCompanyMetricsResponseDtoV2` (flat + nested `metrics`) |
| `GET /v2/crm/contact` | `CrmContactListItemDtoV2` = `{ contact, metrics }` | `CrmContactMetricsResponseDtoV2` (flat + nested `metrics`) |

## Why

The old wrappers nested the entity under a `company`/`contact` key purely to sit next to `metrics`. The frontend store now consumes a flat shape, so the entity fields move to the top level of each list item and `metrics` stays as a nested object.

## Design note

The shared base DTOs — `CrmCompanyResponseDto` (v1) and `CrmContactResponseDtoV2` — are **untouched**. They remain metrics-free everywhere they're embedded (deal/task `company`/`contact`) and in `GET /v2/crm/contact/{id}`. A separate list-only type (rather than an optional `metrics` field on the shared base) avoids leaking `"metrics": null` into those other responses and keeps each response contract honest.

## Wire change (frontend impact)

- `items[].company.id` → `items[].id` (company list)
- `items[].contact.id` → `items[].id`, `items[].contact.company` → `items[].company` (contact list)
- `items[].metrics.*` — **unchanged**

## Changes

- New: `CrmCompanyMetricsResponseDtoV2`, `CrmContactMetricsResponseDtoV2`
- Removed: `CrmCompanyListItemDtoV2`, `CrmContactListItemDtoV2`
- Flattened the `cb.construct(...)` in `CrmCompanyRepositoryImpl` / `CrmContactRepositoryImpl`; updated repository + service generics
- Updated the two V2 controller integration tests to the flat shape

## Testing

`CrmCompanyControllerV2IntegrationTest` + `CrmContactControllerV2IntegrationTest` — **17 tests, 0 failures**.


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR is a clean, mechanical flattening of the two V2 metrics list responses: `CrmCompanyListItemDtoV2`/`CrmContactListItemDtoV2` are replaced by `CrmCompanyMetricsResponseDtoV2`/`CrmContactMetricsResponseDtoV2` that inline the entity fields and keep `metrics` nested. I compared the old and new `cb.construct` argument lists line by line for both repositories — positional order and types match the `@AllArgsConstructor` field order exactly in both cases, so no data is transposed, and no other callers or frontend consumers of the old class names remain (the v2 API has no frontend client yet, so the shape change is not breaking). A side benefit is that `GET /v2/crm/contact` and `GET /v2/crm/contact/{id}` now return the same flat contact shape. Remaining concerns are maintainability (both new DTOs fully duplicate the fields of an existing DTO) and test coverage for the newly flattened String fields, whose positional construction makes a future transposition silently wrong.

Found **14** new issue(s): 🔴 1 important, 🟡 6 suggestion(s), 🟣 7 nit(s)

backend/src/main/java/com/skapp/community/crmplanner/payload/response/v2/CrmContactMetricsResponseDtoV2.java
```diff
@@ -0,0 +1,35 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import lombok.Setter;
+
+import java.time.LocalDateTime;
+
+@Getter
+@Setter
+@AllArgsConstructor
+public class CrmContactMetricsResponseDtoV2 {
```
ThinuwanW, line 15: 🤖 **Claude** · 🟡 **Suggestion**

Both new DTOs duplicate every field of an existing DTO verbatim: `CrmContactMetricsResponseDtoV2` repeats all 8 fields of `CrmContactResponseDtoV2` (id, name, email, contactNumber, lastContactAt, lastModifiedDate, company, owner), and `CrmCompanyMetricsResponseDtoV2` repeats all 6 fields of `CrmCompanyResponseDto` (id, name, industry, website, address, contactNumber). This is the duplication case in the review checklist. The practical risk is drift: adding a field to `CrmContactResponseDtoV2` (returned by `GET /v2/crm/contact/{id}`) will silently not appear in the list response from `GET /v2/crm/contact`, and vice versa — the two endpoints will diverge with no compile error. Two ways to keep the flat JSON without duplicating fields: (a) keep the composed shape in Java and annotate the nested field with Jackson's `@JsonUnwrapped` (e.g. `@JsonUnwrapped private CrmContactResponseDtoV2 contact;`), which serializes flat while leaving the existing `cb.construct` nesting intact; or (b) have the metrics DTO extend the base DTO and declare one explicit constructor covering the inherited plus new fields for `cb.construct` to bind to. Option (a) also removes the need to change the repository select clauses at all.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmCompanyControllerV2IntegrationTest.java
```diff
@@ -166,12 +166,12 @@ void getCompanyMetrics_HappyPath_ReturnsNestedCompanyAndMetrics() throws Excepti
...
-			.andExpect(jsonPath("['results'][0]['items'][0]['company']['name']").value("MetricsCoV2Unique"))
-			.andExpect(jsonPath("['results'][0]['items'][0]['company']['industry']")
+			.andExpect(jsonPath("['results'][0]['items'][0]['id']").value(company.getId()))
+			.andExpect(jsonPath("['results'][0]['items'][0]['name']").value("MetricsCoV2Unique"))
+			.andExpect(jsonPath("['results'][0]['items'][0]['industry']")
 				.value(CrmIndustry.TECHNOLOGY_INFORMATION_AND_MEDIA.name()))
-			.andExpect(jsonPath("['results'][0]['items'][0]['company']['website']").value("https://metrics-v2.com"))
-			.andExpect(jsonPath("['results'][0]['items'][0]['company']['address']").value("123 Metrics St"))
+			.andExpect(jsonPath("['results'][0]['items'][0]['website']").value("https://metrics-v2.com"))
+			.andExpect(jsonPath("['results'][0]['items'][0]['address']").value("123 Metrics St"))
```
ThinuwanW, line 174: 🤖 **Claude** · 🟡 **Suggestion**

The flattened fields are bound positionally by `cb.construct`, and `website`, `address` and `contactNumber` are all `String` and now adjacent in both the DTO and the select clause — transposing any two of them compiles, runs, and returns wrong data silently. `savedCompany` seeds `contactNumber` as "94771234567" but the test never asserts it, so only two of the three String fields are pinned. Add `.andExpect(jsonPath("['results'][0]['items'][0]['contactNumber']").value("94771234567"))`. The same gap exists in `CrmContactControllerV2IntegrationTest.getContactMetrics_WithContacts_...`, where the flattened `email` and `contactNumber` (also two adjacent Strings) and the `owner` object are unasserted — add assertions for `['email']`, `['contactNumber']` and `['owner']['employeeId']` there. Given the whole point of this PR is to change which fields sit at which position, these assertions are what would catch a regression.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
```diff
@@ -111,22 +111,21 @@ public Page<CrmContactListItemDtoV2> getContactMetricsV2(CrmContactMetricRequest
...
 					cb.lessThan(overdueTask.get(CrmTask_.dueAt), cb.literal(LocalDate.now().atStartOfDay())));
 
-		query.select(cb.construct(CrmContactListItemDtoV2.class, cb.construct(CrmContactResponseDtoV2.class,
-				contact.get(CrmContact_.id), contact.get(CrmContact_.name), contact.get(CrmContact_.email),
-				contact.get(CrmContact_.contactNumber), contact.get(CrmContact_.lastContactAt),
-				contact.get(Auditable_.lastModifiedDate),
+		query.select(cb.construct(CrmContactMetricsResponseDtoV2.class, contact.get(CrmContact_.id),
+				contact.get(CrmContact_.name), contact.get(CrmContact_.email), contact.get(CrmContact_.contactNumber),
+				contact.get(CrmContact_.lastContactAt), contact.get(Auditable_.lastModifiedDate),
 				cb.construct(CrmCompanyResponseDto.class, company.get(CrmCompany_.id), company.get(CrmCompany_.name),
```
ThinuwanW, line 117: 🤖 **Claude** · 🟣 **Nit**

The select clause you touched builds `CrmCompanyResponseDto` from a LEFT JOIN, so for a contact with no company Hibernate still invokes the constructor with all-null arguments and emits `"company": {"id": null, "name": null, ...}` rather than `"company": null`. This is pre-existing behaviour, but the flattened DTO is now the response contract a new frontend will be written against, and a client cannot use `company == null` as a has-no-company signal. `getContactMetrics_ContactWithoutCompany_IsReturned` exercises exactly this row but only asserts `name`, so the contract is unpinned either way. Suggest adding an assertion for whatever the intended shape is (e.g. `jsonPath("...['company']['id']").value(nullValue())` for today's behaviour), and if `null` is the intent, null the field out in `CrmContactServiceImplV2.getContactMetrics` when `company.getId() == null`.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmContactControllerV2IntegrationTest.java
```diff
@@ -278,7 +278,7 @@ void getContactById_WithoutCrmRole_ReturnsForbidden() throws Exception {
 	// --- getContactMetrics ---
 
 	@Test
-	@DisplayName("Get contact metrics - Returns page with nested contact and zero metrics when no deals or tasks")
+	@DisplayName("Get contact metrics - Returns page with flat contact fields and zero metrics when no deals or tasks")
 	void getContactMetrics_WithContacts_ReturnsNestedContactAndMetrics() throws Exception {
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

The `@DisplayName` was correctly updated to say "flat contact fields" but the Java method name still reads `getContactMetrics_WithContacts_ReturnsNestedContactAndMetrics`. Same in `CrmCompanyControllerV2IntegrationTest` (line 155): `getCompanyMetrics_HappyPath_ReturnsNestedCompanyAndMetrics`. Rename to `...ReturnsFlatContactAndMetrics` / `...ReturnsFlatCompanyAndMetrics` so the method names don't contradict the assertions.

backend/src/main/java/com/skapp/community/crmplanner/payload/response/v2/CrmCompanyMetricsResponseDtoV2.java
```diff
@@ -1,17 +1,27 @@
 package com.skapp.community.crmplanner.payload.response.v2;
 
-import com.skapp.community.crmplanner.payload.response.CrmCompanyResponseDto;
 import com.skapp.community.crmplanner.type.CrmCompanyMetrics;
+import com.skapp.community.crmplanner.type.CrmIndustry;
```
ThinuwanW, line 4: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

The flattening stops halfway: entity fields are hoisted to the top level but `metrics` stays a nested object. The frontend v2 store entities this API is presumably being shaped for (`frontend/src/community/crm/v2/types/CrmCommonTypes.ts`) declare `CrmCompanyEntity` and `CrmContactEntity` with the metric fields flat at the top level (openTasksCount, overdue, openValue, accountValue, openDeals, closedDeals / closedDealValue, closedDealCount, openTasksCount, overdueTasksCount), so a normalizer will still have to spread `metrics` client-side. If nested `metrics` is deliberate (it does group the derived/aggregate values away from the persisted columns, and it matches the v1 `getContactMetricsById` response), ignore this; otherwise flattening the metric fields too would finish the job and match the declared client contract.

backend/src/main/java/com/skapp/community/crmplanner/payload/response/v2/CrmCompanyMetricsResponseDtoV2.java
```diff
@@ -1,17 +1,27 @@
...
 import lombok.AllArgsConstructor;
 import lombok.Getter;
 import lombok.Setter;
 
 @Getter
 @Setter
 @AllArgsConstructor
-public class CrmCompanyListItemDtoV2 {
+public class CrmCompanyMetricsResponseDtoV2 {
```
ThinuwanW, line 12: 🤖 **Claude** · 🟣 **Nit**

`CrmCompanyMetricsResponseDtoV2` now differs from the existing v1 `com.skapp.community.crmplanner.payload.response.CrmCompanyMetricsResponseDto` only by the `V2` suffix, yet the two have structurally different payloads (v1 has the six metric values flat on the DTO; v2 nests them under `metrics`). `CrmCompanyRepositoryImpl` imports both, so it is easy to reach for the wrong one when editing either query. A name that encodes the difference — e.g. `CrmCompanyWithMetricsResponseDtoV2` — would make the distinction visible at the call site.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmCompanyServiceImplV2.java
```diff
@@ -27,7 +27,7 @@ public ResponseEntityDto getCompanyMetrics(CrmCompanyMetricRequestDto requestDto
 		log.info("getCompanyMetrics: execution started");
 
 		Pageable pageable = PageRequest.of(requestDto.getPage(), requestDto.getSize());
```
ThinuwanW, line 29: 🤖 **Claude** · 🔴 **Important**

`PageRequest.of(requestDto.getPage(), requestDto.getSize())` is fed straight from unvalidated query parameters. `CrmCompanyMetricRequestDto` / `CrmContactMetricRequestDto` declare plain `int page = 0` and `int size = 10` with no bean-validation annotations, and the controllers bind them without `@Valid`. Two consequences on both endpoints touched by this PR (`CrmCompanyServiceImplV2:29` and `CrmContactServiceImplV2:40`): (1) `?page=-1` or `?size=0` makes `PageRequest.of` throw `IllegalArgumentException`, which is not a `ModuleException` and therefore escapes `GlobalExceptionHandler`'s `ErrorResponse` contract as a raw 500 — this is exactly checklist item 1 (unhandled exception -> 500 to the client); (2) `?size=100000` produces an effectively unbounded criteria query with six correlated subqueries per row. Fix: annotate the request DTOs (`@Min(0)` on `page`, `@Min(1) @Max(100)` on `size`), add `@Valid` to the controller parameters so `MethodArgumentNotValidException` flows through the existing handler, or clamp both values in the service before calling `PageRequest.of`.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
```diff
@@ -111,22 +111,21 @@ public Page<CrmContactListItemDtoV2> getContactMetricsV2(CrmContactMetricRequest
 					cb.isNotNull(overdueTask.get(CrmTask_.dueAt)),
 					cb.lessThan(overdueTask.get(CrmTask_.dueAt), cb.literal(LocalDate.now().atStartOfDay())));
```
ThinuwanW, line 112: 🤖 **Claude** · 🟡 **Suggestion**

The two endpoints this PR deliberately aligns into the same shape disagree on what 'overdue' means. `getContactMetricsV2` here uses `cb.literal(LocalDate.now().atStartOfDay())`, so a task due earlier *today* is NOT counted as overdue, and the cutoff is computed from the application server's default timezone at query-build time. `CrmCompanyRepositoryImpl.getCompanyMetricsV2` (line 144) instead uses `cb.localDateTime()`, i.e. the database's current timestamp, so the same task IS counted as overdue for the company. The result is that a company row and its contact row can report contradictory overdue counts for the same underlying task within the same page load. Pick one definition and apply it to both (the start-of-day form is the majority convention in this module — `CrmTaskRepositoryImpl:142,207` and `CrmContactRepositoryImpl:177` all use it — so changing `CrmCompanyRepositoryImpl` lines 69/144 to `cb.literal(LocalDate.now().atStartOfDay())` is the smaller, more consistent change). Also note `cb.literal(...)` inlines the timestamp into the SQL rather than binding a parameter, which defeats statement caching; `cb.parameter` / a bound value would be preferable in whichever form you standardise on.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
```diff
@@ -111,22 +111,21 @@ public Page<CrmContactListItemDtoV2> getContactMetricsV2(CrmContactMetricRequest
...
+		query.select(cb.construct(CrmContactMetricsResponseDtoV2.class, contact.get(CrmContact_.id),
+				contact.get(CrmContact_.name), contact.get(CrmContact_.email), contact.get(CrmContact_.contactNumber),
+				contact.get(CrmContact_.lastContactAt), contact.get(Auditable_.lastModifiedDate),
 				cb.construct(CrmCompanyResponseDto.class, company.get(CrmCompany_.id), company.get(CrmCompany_.name),
 						company.get(CrmCompany_.industry), company.get(CrmCompany_.website),
 						company.get(CrmCompany_.address), company.get(CrmCompany_.contactNumber)),
 				cb.construct(CrmOwnerResponseDto.class, owner.get(Employee_.employeeId), owner.get(Employee_.firstName),
-						owner.get(Employee_.lastName), owner.get(Employee_.authPic))),
+						owner.get(Employee_.lastName), owner.get(Employee_.authPic)),
 				cb.construct(CrmContactMetrics.class, closedValueSub.cast(String.class), closedCountSub, openTaskSub,
```
ThinuwanW, line 122: 🤖 **Claude** · 🟡 **Suggestion**

`getContactMetricsV2` constructs `CrmContactMetrics` from `openTaskSub`/`overdueTaskSub`, which count only tasks joined directly to the contact (`cb.equal(openTask.get(CrmTask_.contact), contact)`). `getContactMetricsById` in this same file (lines 157-177) builds the identical `CrmContactMetrics` type but counts tasks linked to the contact directly OR via the contact's deals (`cb.or(openDirectContact..., openDealContact...)`). So the list view and the detail view report different `openTasksCount`/`overdueTasksCount` for the same contact whenever deal-linked tasks exist — an unexplained asymmetry between two views of the same data returning the same DTO type (checklist item 10). Since this PR promotes `metrics` to a top-level sibling of the flattened contact fields, the two are now more likely to be consumed side by side. Decide which definition is canonical and extract the subquery construction into a shared private helper so both call sites cannot drift again.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmContactControllerV2IntegrationTest.java
```diff
@@ -288,9 +288,9 @@ void getContactMetrics_WithContacts_ReturnsNestedContactAndMetrics() throws Exce
 			.andExpect(status().isOk())
 			.andExpect(jsonPath(STATUS_PATH).value(STATUS_SUCCESSFUL))
 			.andExpect(jsonPath("['results'][0]['totalItems']").value(1))
-			.andExpect(jsonPath("['results'][0]['items'][0]['contact']['id']").value(contactId))
-			.andExpect(jsonPath("['results'][0]['items'][0]['contact']['name']").value("ZeroMetricsContactV2Unique"))
-			.andExpect(jsonPath("['results'][0]['items'][0]['contact']['company']['id']").value(companyId))
+			.andExpect(jsonPath("['results'][0]['items'][0]['id']").value(contactId))
+			.andExpect(jsonPath("['results'][0]['items'][0]['name']").value("ZeroMetricsContactV2Unique"))
+			.andExpect(jsonPath("['results'][0]['items'][0]['company']['id']").value(companyId))
```
ThinuwanW, line 297: 🤖 **Claude** · 🟡 **Suggestion**

The contact flattening moved every field's position in the `cb.construct` argument list — `owner` in particular went from being argument 8 of the inner `CrmContactResponseDtoV2` construct to argument 8 of the outer `CrmContactMetricsResponseDtoV2` construct. But the updated assertions only cover `id`, `name` and `company.id`. Nothing in this test class asserts the top-level `email`, `contactNumber`, `lastContactAt`, `lastModifiedDate`, or any field of the `owner` block on the metrics response, so a mis-ordered or dropped argument in the flattened select would pass CI silently (the same-type `name`/`email`/`contactNumber` strings are especially vulnerable). Add assertions here for `['items'][0]['email']`, `['items'][0]['contactNumber']` and `['items'][0]['owner']['employeeId']` (and `['owner']['firstName']`), mirroring what `createContact_HappyPath_ReturnsBaseContactWithFullCompanyAndOwner` already asserts for the by-id shape.

backend/src/main/java/com/skapp/community/crmplanner/payload/response/v2/CrmCompanyMetricsResponseDtoV2.java
```diff
@@ -1,17 +1,27 @@
...
+
+	private CrmIndustry industry;
+
+	private String website;
+
+	private String address;
+
+	private String contactNumber;
 
 	private CrmCompanyMetrics metrics;
```
ThinuwanW, line 26: 🤖 **Claude** · 🟡 **Suggestion**

Now that company and contact v2 metrics responses share an identical envelope (flat entity fields + a `metrics` object), the key naming inside that object is visibly inconsistent: `CrmCompanyMetrics` exposes `openTasksCount` and `overdue`, while `CrmContactMetrics` exposes `openTasksCount` and `overdueTasksCount`. The tests bake this in — `CrmCompanyControllerV2IntegrationTest:199` reads `metrics.overdue` and `CrmContactControllerV2IntegrationTest:297` reads `metrics.overdueTasksCount`. A client rendering both list types has to special-case the key. This PR is the natural point to rename `CrmCompanyMetrics.overdue` to `overdueTasksCount` (v2 is not yet consumed by the frontend — there is no v2 API layer under `frontend/src/community/crm/v2`, only store/types — so the rename is currently free); if it must stay, add `@JsonProperty("overdueTasksCount")` or document why the two differ.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmCompanyRepositoryImpl.java
```diff
@@ -172,18 +171,17 @@ public Page<CrmCompanyListItemDtoV2> getCompanyMetricsV2(Pageable pageable, Stri
 					cb.isFalse(openCountDeal.get(CrmDeal_.isDeleted)),
 					cb.not(openCountDeal.get(CrmDeal_.stage).get(CrmDealStage_.id).in(closedStageIds)));
 
-		query.select(cb.construct(CrmCompanyListItemDtoV2.class,
-				cb.construct(CrmCompanyResponseDto.class, company.get(CrmCompany_.id), company.get(CrmCompany_.name),
-						company.get(CrmCompany_.industry), company.get(CrmCompany_.website),
-						company.get(CrmCompany_.address), company.get(CrmCompany_.contactNumber)),
+		query.select(cb.construct(CrmCompanyMetricsResponseDtoV2.class, company.get(CrmCompany_.id),
```
ThinuwanW, line 175: 🤖 **Claude** · 🟣 **Nit**

`getCompanyMetricsV2` now projects the six company columns in the order `id, name, industry, website, address, contactNumber`, while `getCompanyMetrics` twenty lines above (line 99-101) projects the same six columns for the v1 flat DTO in the order `id, name, contactNumber, industry, website, address`. Two near-identical positional `cb.construct` blocks in one file with different column orders is a maintenance trap — copying a line between them silently swaps `contactNumber` with `industry`/`website`/`address`, and the compiler cannot help because they are all `String`/enum projections bound by position. Consider reordering `CrmCompanyMetricsResponseDto`'s fields to match `CrmCompanyResponseDto`/`CrmCompanyMetricsResponseDtoV2` so both projections read identically, or extract a shared `private Selection<?>[] companyColumns(Root<CrmCompany>)` helper used by both.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmCompanyControllerV2IntegrationTest.java
```diff
@@ -151,7 +151,7 @@ private void savedTask(CrmCompany company, LocalDateTime dueAt) {
 	}
 
 	@Test
-	@DisplayName("Get company metrics - Returns nested company and metrics with seeded values")
+	@DisplayName("Get company metrics - Returns flat company fields and nested metrics with seeded values")
 	void getCompanyMetrics_HappyPath_ReturnsNestedCompanyAndMetrics() throws Exception {
```

Same stale-name problem as the contact test, but in this file and not flagged in the first pass: the `@DisplayName` on line 154 was updated to 'Returns flat company fields and nested metrics', while the Java method is still `getCompanyMetrics_HappyPath_ReturnsNestedCompanyAndMetrics`. Rename to `getCompanyMetrics_HappyPath_ReturnsFlatCompanyFieldsAndMetrics` so the method name, display name and assertions agree — otherwise the method name is the only thing a developer sees in an IDE test-runner failure list and it describes the pre-PR shape.

backend/src/main/java/com/skapp/community/crmplanner/payload/response/v2/CrmCompanyMetricsResponseDtoV2.java
```diff
@@ -1,17 +1,27 @@
 package com.skapp.community.crmplanner.payload.response.v2;
 
-import com.skapp.community.crmplanner.payload.response.CrmCompanyResponseDto;
 import com.skapp.community.crmplanner.type.CrmCompanyMetrics;
```
ThinuwanW, line 3: 🤖 **Claude** · 🟣 **Nit**

`CrmCompanyMetrics` and `CrmContactMetrics` are plain response payload classes (Lombok `@Getter`/`@AllArgsConstructor` value objects serialized straight to the client), but they live in `crmplanner/type/`, which the module conventions reserve for enums; DTOs belong in `payload/` and its `response/` subdir. This was tolerable while they were an internal aggregation carrier, but after this PR they are the *only* nested object in the two `*ResponseDtoV2` classes and are effectively part of the public v2 response contract. Consider moving both to `payload/response/v2/` (and naming them `CrmCompanyMetricsDto`/`CrmContactMetricsDto`) as a follow-up so the response contract lives entirely in the payload package.

backend/src/main/java/com/skapp/community/crmplanner/payload/response/v2/CrmContactMetricsResponseDtoV2.java
anusham-rc: Intentional — a separate flat DTO keeps the shared base (`CrmCompanyResponseDto` / `CrmContactResponseDtoV2`) metrics-free where it's embedded (deal/task) and in `GET /v2/crm/contact/{id}`. The field duplication is the accepted tradeoff over coupling the two contracts.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
anusham-rc: Pre-existing LEFT-JOIN behavior, unchanged by this PR — noting for a separate follow-up rather than reshaping the query here.

backend/src/main/java/com/skapp/community/crmplanner/payload/response/v2/CrmCompanyMetricsResponseDtoV2.java
anusham-rc: Nesting `metrics` is deliberate: it matches the single-resource metrics shape and groups the aggregates away from the persisted columns (same rationale as PR #1742, comment #10). Not flattening the metric fields in this PR.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmCompanyServiceImplV2.java
anusham-rc: Minor correction: this isn't a raw 500 — `IllegalArgumentException` from `PageRequest.of` is caught by the global `@ExceptionHandler(Exception.class)` and returned as a structured `ErrorResponse`. The `PageRequest.of` call is unchanged by this PR (pre-existing), and input-bounds validation is an app-wide concern best handled centrally; tracking `@Min`/`@Max` separately.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
anusham-rc: Real but pre-existing (`cb.localDateTime()` on the company side vs `LocalDate.now().atStartOfDay()` here), not introduced by this PR — will align both under one definition in a separate change.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
anusham-rc: Pre-existing, and already adjudicated in PR #1742 (comment #15): metric task-counts attribute by direct link only; aligning task attribution (company vs deal vs contact) across all metric queries is a broader change, out of scope here.

backend/src/main/java/com/skapp/community/crmplanner/payload/response/v2/CrmCompanyMetricsResponseDtoV2.java
anusham-rc: Already addressed in commit 004cdc06 — `CrmCompanyMetrics.overdue` was renamed to `overdueTasksCount` (plus `openDealsCount` / `closedDealsCount`), and the test assertions updated. This review predates that commit.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmCompanyRepositoryImpl.java
anusham-rc: Keeping the metric criteria queries self-contained by design — same decision as PR #1742 (comment #1); no shared column/subquery builder.

backend/src/main/java/com/skapp/community/crmplanner/payload/response/v2/CrmCompanyMetricsResponseDtoV2.java
anusham-rc: Follows the existing convention — the sibling `CrmContactMetrics` also lives in `crmplanner/type/` and is returned directly (see PR #1742, comment #3). Keeping the placement consistent.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmContactControllerV2IntegrationTest.java
anusham-rc: Renamed to getContactMetrics_WithContacts_ReturnsFlatContactAndMetrics so the method name, @DisplayName and assertions agree (8a8115960).

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmCompanyControllerV2IntegrationTest.java
anusham-rc: Renamed to getCompanyMetrics_HappyPath_ReturnsFlatCompanyFieldsAndMetrics to match the @DisplayName and assertions (8a8115960).

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmCompanyControllerV2IntegrationTest.java
anusham-rc: Added an assertion on contactNumber so all three adjacent String fields (website/address/contactNumber) are pinned against a positional transposition (8a8115960).

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmContactControllerV2IntegrationTest.java
anusham-rc: Added assertions for email, contactNumber and owner.employeeId on the flat contact metrics response (seeding a contactNumber in the test helper) so a mis-ordered or dropped construct argument fails CI (8a8115960).

backend/src/main/java/com/skapp/community/crmplanner/payload/response/v2/CrmCompanyMetricsResponseDtoV2.java
anusham-rc: Keeping the ...ResponseDtoV2 name for consistency with its sibling CrmContactMetricsResponseDtoV2 and the rest of the v2 payload package, where V2 is the established marker for the flattened responses. The v1 vs v2 distinction is already carried by the package (payload.response vs payload.response.v2). Open to standardising the naming across the whole v2 package in a follow-up.

akilarootcode (dismissed):

akilarootcode (approved):

PR #1762, opened 2026-08-14
fix: Auto-enable birthday notification sub-toggles on main toggle enable

SanithuGunasena07:
In people configuration when birthday notifications are turned on the sub notification under it also should turn on automatically fixed this by setting the two to true when notifications are enabled.

### DIFF
--- a/frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
+++ b/frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
@@ -106,6 +106,14 @@ const PeopleConfigurations: FC = () => {
     }
   };
 
+  const handleMainToggleChange = (checked: boolean) => {
+    birthdayFormik.setFieldValue("isTurnedOn", checked);
+    if (checked) {
+      birthdayFormik.setFieldValue("isTeamWide", true);
+      birthdayFormik.setFieldValue("isOrganizationWide", true);
+    }
+  };
+
   const subOptionsAriaMessage = birthdayNotificationConfig
     ? translateText([
         "aria",
@@ -154,9 +162,7 @@ const PeopleConfigurations: FC = () => {
             labelId="birthday-notification"
             arialabel={translateText(["aria", "mainToggle"])}
             checked={birthdayFormik.values.isTurnedOn}
-            onChange={(checked) =>
-              birthdayFormik.setFieldValue("isTurnedOn", checked)
-            }
+            onChange={handleMainToggleChange}
           />
           {birthdayFormik.values.isTurnedOn && (
             <>
--- a/frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
+++ b/frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
@@ -106,11 +106,20 @@ const PeopleConfigurations: FC = () => {
     }
   };
 
-  const handleMainToggleChange = (checked: boolean) => {
+  const handleBirthdayNotificationToggle = (checked: boolean): void => {
     birthdayFormik.setFieldValue("isTurnedOn", checked);
     if (checked) {
       birthdayFormik.setFieldValue("isTeamWide", true);
       birthdayFormik.setFieldValue("isOrganizationWide", true);
+    } else {
+      birthdayFormik.setFieldValue(
+        "isTeamWide",
+        birthdayFormik.initialValues.isTeamWide
+      );
+      birthdayFormik.setFieldValue(
+        "isOrganizationWide",
+        birthdayFormik.initialValues.isOrganizationWide
+      );
     }
   };
 
@@ -162,7 +171,7 @@ const PeopleConfigurations: FC = () => {
             labelId="birthday-notification"
             arialabel={translateText(["aria", "mainToggle"])}
             checked={birthdayFormik.values.isTurnedOn}
-            onChange={handleMainToggleChange}
+            onChange={handleBirthdayNotificationToggle}
           />
           {birthdayFormik.values.isTurnedOn && (
             <>


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR auto-enables both birthday-notification sub-toggles when the main toggle is switched on. The change is small but interacts badly with the backend's scope resolution: `PeopleServiceImpl.resolveBirthdayNotificationScope` treats `isOrganizationWide` and `isTeamWide` as mutually exclusive (organization wins), so forcing both to `true` renders the visibly-ON "team only" toggle a no-op and silently overwrites the admin's persisted preference. There is also a phantom-dirty-state issue when the main toggle is switched back off, since the sub-values are never restored.

Found **11** new issue(s): 🔴 5 important, 🟡 3 suggestion(s), 🟣 3 nit(s)

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
```diff
@@ -106,6 +106,14 @@ const PeopleConfigurations: FC = () => {
     }
   };
 
+  const handleMainToggleChange = (checked: boolean) => {
+    birthdayFormik.setFieldValue("isTurnedOn", checked);
+    if (checked) {
+      birthdayFormik.setFieldValue("isTeamWide", true);
```
ThinuwanW, line 112: 🤖 **Claude** · 🔴 **Important**

Setting `isTeamWide` AND `isOrganizationWide` to `true` produces a contradictory config that the backend cannot honour. In `backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java:1642-1651` (`resolveBirthdayNotificationScope`) the flags are resolved as a single mutually-exclusive scope: `isOrganizationWide` -> ORGANIZATION, else `isTeamWide` -> TEAM, else SELF. `EmployeeRepositoryImpl.buildBirthdayScopePredicate` then returns `cb.conjunction()` (no filter -> everyone) for ORGANIZATION. So when both are true the effective behaviour is always organization-wide, while the UI shows "Send notification to team only" switched ON — a toggle the user can flip with zero effect until they also turn the organisation toggle off. Pick a single default scope instead, e.g. `setFieldValue("isOrganizationWide", true)` and `setFieldValue("isTeamWide", false)` (or vice-versa), or better, model the two switches as mutually exclusive so turning one on turns the other off, matching the backend contract.

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
```diff
@@ -106,6 +106,14 @@ const PeopleConfigurations: FC = () => {
     }
   };
 
+  const handleMainToggleChange = (checked: boolean) => {
+    birthdayFormik.setFieldValue("isTurnedOn", checked);
+    if (checked) {
```
ThinuwanW, line 111: 🤖 **Claude** · 🔴 **Important**

The auto-enable is unconditional, so it silently overwrites a persisted admin preference. Scenario: saved config is `{isTurnedOn: true, isTeamWide: true, isOrganizationWide: false}`. The admin toggles the main switch off (intending to disable notifications), changes their mind and toggles it back on — `isOrganizationWide` is now `true` even though they never touched that switch, the form is dirty, and clicking Save silently broadens birthday notifications from team-only to the entire organisation. Guard the defaulting so it only applies when there is no existing scope, e.g. `if (checked && !birthdayFormik.values.isTeamWide && !birthdayFormik.values.isOrganizationWide) { ...set the default scope... }`.

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
```diff
@@ -106,6 +106,14 @@ const PeopleConfigurations: FC = () => {
...
+      birthdayFormik.setFieldValue("isTeamWide", true);
+      birthdayFormik.setFieldValue("isOrganizationWide", true);
+    }
```
ThinuwanW, line 123: 🤖 **Claude** · 🔴 **Important**

There is no `else` branch, so the mutated sub-values are never restored when the main toggle is switched back off. Scenario: saved config is all `false`. The admin toggles the main switch on, then immediately off. Form values are now `{isTurnedOn: false, isTeamWide: true, isOrganizationWide: true}` — the sub-switches are hidden (line 167 gates them on `isTurnedOn`), the UI looks identical to the saved state, yet `birthdayFormik.dirty` is `true`, so the Cancel/Save buttons (lines 200/208, gated on `isAnyChanged`) become enabled with nothing visibly changed. Saving then persists sub-toggle values the admin never chose. Add an else branch restoring the sub-fields from `birthdayFormik.initialValues.isTeamWide` / `.isOrganizationWide`.

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
```diff
@@ -106,6 +106,14 @@ const PeopleConfigurations: FC = () => {
     }
   };
 
+  const handleMainToggleChange = (checked: boolean) => {
+    birthdayFormik.setFieldValue("isTurnedOn", checked);
```
ThinuwanW, line 110: 🤖 **Claude** · 🟡 **Suggestion**

Three sequential `setFieldValue` calls trigger three separate Formik state updates and three `validateForm` passes, each computed against the same pre-update `state.values` snapshot (Formik v2 captures `state.values` in the closure of `setFieldValue` for validation). It works today only because this form has no Yup schema; if validation is added later the intermediate error state will be computed from partially-applied values. Prefer a single atomic update: `birthdayFormik.setValues({ ...birthdayFormik.values, isTurnedOn: checked, ...(checked && { isTeamWide: true, isOrganizationWide: true }) })`. This also makes the "all three fields change together" intent explicit.

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
```diff
@@ -106,6 +106,14 @@ const PeopleConfigurations: FC = () => {
     }
   };
 
+  const handleMainToggleChange = (checked: boolean) => {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

No test coverage for this behaviour. There is no `PeopleConfigurations.test.tsx` anywhere in the repo (only `SwitchRow.test.tsx` for the atom), and this PR introduces non-trivial conditional form logic that has several edge cases (re-enable after disable, toggle off restoring values, dirty-state calculation). Add a `PeopleConfigurations.test.tsx` alongside the component, wrapped in the `MockTheme` provider used by the other component tests, asserting with `screen.getByRole("switch", { name: ... })` that: (1) enabling the main toggle sets the intended default scope, (2) disabling it does not leave the form dirty, (3) an existing saved scope is preserved across an off/on cycle.

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
```diff
@@ -106,6 +106,14 @@ const PeopleConfigurations: FC = () => {
...
+    birthdayFormik.setFieldValue("isTurnedOn", checked);
+    if (checked) {
+      birthdayFormik.setFieldValue("isTeamWide", true);
+      birthdayFormik.setFieldValue("isOrganizationWide", true);
+    }
+  };
+
   const subOptionsAriaMessage = birthdayNotificationConfig
```
ThinuwanW, line 126: 🤖 **Claude** · 🟣 **Nit**

The `aria-live` region only announces that the sub-options are now "available"/"hidden" (`subOptionsAvailable` / `subOptionsUnavailable`). With this PR the act of enabling the main toggle now also changes the value of both sub-toggles, which is a state change a screen-reader user has no way to perceive — they will hear "options are now available" but not that both were switched on for them. Consider adding a translation key (e.g. `aria.subOptionsEnabledByDefault`) and announcing it when the auto-enable actually fires.

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
```diff
@@ -106,6 +106,14 @@ const PeopleConfigurations: FC = () => {
     }
   };
```
ThinuwanW, line 107: 🤖 **Claude** · 🔴 **Important**

`handleSaveAll` awaits `birthdayFormik.submitForm()` with no try/catch. Formik 2.x re-throws whatever `onSubmit` rejects with (only the internal `handleSubmit` swallows it with a console.warn — a direct `submitForm()` call propagates), and `onSubmit` here awaits `updateConfigAsync`, which rejects on any HTTP/network failure (and also on the global `onMutate` offline guard in TanStackProvider). Two consequences: (1) the click handler returns a rejected promise that nothing catches → unhandled promise rejection; (2) execution never reaches `setIsWorkspaceSaveTriggered(true)`, so for an enterprise super admin who edited both the birthday toggles and the Google Workspace toggles and pressed the single combined Save, the workspace changes are silently discarded — the only toast shown is the birthday error, which reads as "birthday save failed" not "workspace save never ran". This is pre-existing code, but the PR routes new state through it. Fix: wrap the submit in try/catch (the mutation's `onError` already toasts, so an empty catch is acceptable) and trigger the workspace save unconditionally afterwards, e.g. `try { if (birthdayFormik.dirty) await birthdayFormik.submitForm(); } catch { /* toast handled in onError */ } finally { if (isWorkspaceChanged) setIsWorkspaceSaveTriggered(true); }`.

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
ThinuwanW, line 111: 🤖 **Claude** · 🔴 **Important**

The invariant this PR is establishing ("notifications on ⇒ at least one scope selected") is enforced only inside the main-toggle click handler, so the bad state remains reachable by two other paths. (a) After the auto-enable, nothing stops the admin from turning BOTH sub-toggles off and pressing Save — that persists `{isTurnedOn: true, isTeamWide: false, isOrganizationWide: false}`, and `PeopleServiceImpl.resolveBirthdayNotificationScope` then falls through to `BirthdayNotificationScope.SELF`, i.e. notifications are "on" but effectively nobody is notified: exactly the condition the PR is fixing. (b) On reload of that persisted config, `initialValues` hydrates both sub-toggles to false and the handler never runs, so the UI happily renders the invalid state. Any non-UI client (mobile/API) can also create it, because `PeopleConfigServiceImpl.updateBirthdayNotificationConfigs` applies each field independently with no cross-field validation. Fix declaratively rather than in an event handler: add a Yup `validationSchema` to `birthdayFormik` asserting `isTurnedOn === false || isTeamWide || isOrganizationWide` (surface it via the existing `error` prop on SwitchRow and gate the Save button on `birthdayFormik.isValid`), and/or normalise in `onSubmit`. Note that if you add a schema you must also handle the failure path in `handleSaveAll`, since `submitForm()` resolves without submitting when validation fails and the user would get no feedback. Mirroring the rule server-side in `updateBirthdayNotificationConfigs` is the only durable guard.

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

The birthday form logic (Formik instance, success/error toast handlers, and now the toggle-dependency rule) lives inline in the organism, while the sibling section rendered by this very same component follows the module's established pattern: `GoogleWorkspaceSyncSettings` delegates all of its form logic to `usePeopleConfigurationsForm` (frontend/src/enterprise/configurations/hooks/usePeopleConfigurationsForm.ts) and keeps the component presentational. This PR pushes the file to 222 lines, past the >200-line decomposition guideline, and adds cross-field business logic to the layer that is hardest to test — which is why the behaviour has no unit test. Extract a `useBirthdayNotificationConfigForm` hook (and optionally a `BirthdayNotificationSection` molecule) exposing `formik`, `isPending`, `submit`, and the toggle handler; the dependency rule then becomes directly unit-testable without rendering the enterprise workspace section or mocking `useSessionData`/`useGetEnvironment`.

ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Declaration hygiene on the new handler: (1) no explicit return type — strict mode plus the guideline that every return value be typed, and the neighbouring `usePeopleConfigurationsForm` hook annotates its handlers (`handleUpdateSuccess = (): void`, `handleSubmit = (): Promise<void>`); add `: void`. (2) The name is under-scoped for this component, which hosts both the birthday section and the enterprise Google Workspace section — "main toggle" is only meaningful from inside the birthday block. `handleBirthdayNotificationToggle` (matching the `birthdayFormik` naming) reads unambiguously at the call site on line 165. (3) Optional: the function is recreated every render and passed as a prop; `useCallback` would only matter if `SwitchRow` were wrapped in `React.memo`, which it currently is not — so this is stylistic today.

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
```diff
@@ -154,9 +162,7 @@ const PeopleConfigurations: FC = () => {
             labelId="birthday-notification"
             arialabel={translateText(["aria", "mainToggle"])}
             checked={birthdayFormik.values.isTurnedOn}
-            onChange={(checked) =>
-              birthdayFormik.setFieldValue("isTurnedOn", checked)
-            }
+            onChange={handleMainToggleChange}
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

The two values this handler writes are transient form state, and this form runs with `enableReinitialize: true` against a query that opts out of nothing: `useGetBirthdayNotificationConfig` in PeopleApi.ts sets neither `refetchOnWindowFocus: false` nor `staleTime`, unlike most sibling queries in the same file which explicitly pass `refetchOnWindowFocus: false`, and there is no global default (TanStackProvider only configures `mutations`). Formik's reinitialize effect is deep-equality guarded, so an identical refetch is harmless — but if the server value actually changes (another people-admin saves, or the auth interceptor's blanket `queryClient.invalidateQueries()` fires after a token refresh) the form is reset and the just-auto-enabled sub-toggles plus the admin's pending main-toggle change vanish with no warning. Low frequency, but cheap to close: add `refetchOnWindowFocus: false` (and a small `staleTime`) to `useGetBirthdayNotificationConfig` for consistency with its neighbours.

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
SanithuGunasena07: As of requirement both toggles should turn on

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
SanithuGunasena07: Per the requirement, enabling the main toggle should always turn on both sub-toggles

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
SanithuGunasena07: Added an else branch restoring isTeamWide/isOrganizationWide to birthdayFormik.initialValues when the main toggle is switched off

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
SanithuGunasena07: not required

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
SanithuGunasena07: Keeping as-is cause this matches existing patterns already used elsewhere in the codebase.

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
SanithuGunasena07: not required

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
SanithuGunasena07: Pre existing not code touched in the PR hence out of scope

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
SanithuGunasena07: pre existing behaviour out of scope for PR

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
SanithuGunasena07: Not required

SanithuGunasena07: Renamed to handleBirthdayNotificationToggle with explicit : void return type

frontend/src/community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations.tsx
SanithuGunasena07: Not required

hasalarootcode (approved):

akilarootcode (approved):

PR #1763, opened 2026-08-14
fix: apply org theme primary color to birthday modal background

SanithuGunasena07:
The Birthday notification modal bg colour was hard coded and not set to organisation colour changed it to organisation colour

### DIFF
--- a/frontend/src/community/people/assets/images/BirthdayCelebration.tsx
+++ b/frontend/src/community/people/assets/images/BirthdayCelebration.tsx
@@ -13,7 +13,7 @@ const BirthdayCelebration = (): JSX.Element => {
       <g clipPath="url(#clip0_929_1835)">
         <path
           d="M-0.078125 42.69C-0.078125 19.113 19.0348 0 42.6119 0H498.022V603H42.6119C19.0349 603 -0.078125 583.887 -0.078125 560.31V42.69Z"
-          fill="#FFFBF0"
+          fill="var(--color-primary)"
         />
         <g clipPath="url(#clip1_929_1835)">
           <rect
--- a/frontend/src/community/people/assets/images/BirthdayCelebration.tsx
+++ b/frontend/src/community/people/assets/images/BirthdayCelebration.tsx
@@ -13,7 +13,7 @@ const BirthdayCelebration = (): JSX.Element => {
       <g clipPath="url(#clip0_929_1835)">
         <path
           d="M-0.078125 42.69C-0.078125 19.113 19.0348 0 42.6119 0H498.022V603H42.6119C19.0349 603 -0.078125 583.887 -0.078125 560.31V42.69Z"
-          fill="var(--color-primary)"
+          fill="var(--color-primary-background)"
         />
         <g clipPath="url(#clip1_929_1835)">
           <rect


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR is a single-line change swapping the hardcoded `#FFFBF0` cream fill on the birthday modal's background panel for `var(--color-primary)` so it picks up the org theme. The mechanism is sound — `--color-primary` is a real token defined in `frontend/styles/global.css` `@theme` and overridden at runtime by `themeSelector.ts`, and `fill="var(--color-primary-*)"` on inline SVGs is an established codebase pattern. My concerns are about token semantics (the codebase reserves `--color-primary-background` for surfaces and uses `--color-primary` only for icon fills) and the visual consequences of putting a saturated pastel where a near-white tint used to be.

Found **10** new issue(s): 🟡 2 suggestion(s), 🟣 2 nit(s)

frontend/src/community/people/assets/images/BirthdayCelebration.tsx
```diff
@@ -13,7 +13,7 @@ const BirthdayCelebration = (): JSX.Element => {
       <g clipPath="url(#clip0_929_1835)">
         <path
           d="M-0.078125 42.69C-0.078125 19.113 19.0348 0 42.6119 0H498.022V603H42.6119C19.0349 603 -0.078125 583.887 -0.078125 560.31V42.69Z"
-          fill="#FFFBF0"
+          fill="var(--color-primary)"
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Token semantics: `--color-primary` is the theme's `primary.main` (a mid-tone pastel: #93C5FD / #A3E635 / #FDBA74 / #D8B4FE / #FDA4AF / #FBBF24 per `themesData.ts`), whereas the value it replaces (`#FFFBF0`) was a near-white tint. The codebase already has a dedicated surface token for exactly this: `--color-primary-background` (`secondary.main` — #DBEAFE / #FEF3C7 / #ECFFC5 etc.), and every background usage in the repo goes through it (`bg-primary-background` in `PeopleSyncBanner.tsx:39`, `ConnectGoogleWorkspaceModal.tsx:105`, `QuickSetupSuccessImage.tsx:7`, `ClockWidget.tsx:151`, ...). `--color-primary` appears exactly once as a `fill`, on a small icon (`PricingCard.tsx:135`) — this PR is the first use of it as a full-panel background. Suggest `fill="var(--color-primary-background)"` for both consistency and a result much closer to the original design intent.

ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Contrast regression on the warm themes. The confetti and dot decorations layered on top of this path use fixed literals (#FF2D78, #FFD700, #00BCD4, #FF6B35, #9C27B0) that were chosen against the near-white #FFFBF0. Against `primary.main` they collide: gold confetti #FFD700 on the YELLOW theme's #FBBF24, and orange confetti #FF6B35 on the ORANGE theme's #FDBA74, are both close to invisible; #FF2D78 on ROSE's #FDA4AF is also heavily muted. Please verify the modal visually across all six themes (YELLOW/BLUE/ROSE/GREEN/PURPLE/ORANGE). Switching to `var(--color-primary-background)` largely resolves this since those values stay in the near-white range; otherwise the confetti palette needs to be re-tuned or themed too.

ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Corner-radius mismatch may now become visible. This path carries its own 42.69px left-corner rounding baked into the `d` attribute, but the SVG is rendered with `preserveAspectRatio="xMidYMid slice"` at 100%×100% inside a container that is itself `rounded-l-[42.69px] overflow-hidden` (`BirthdayModalShell.tsx:36`) over a `bg-white` div (`BirthdayModalShell.tsx:43`). Because `slice` scales the 499-wide viewBox up to cover the panel, the SVG's own arc scales past 42.69px and no longer coincides with the container's clip — leaving a thin sliver of the white wrapper along the left corners. That sliver was invisible when the fill was cream-on-white; with a saturated pastel it will read as an artifact. Simplest fix: make this path a plain full-bleed rect (drop the baked-in radius) and let the container's `rounded-l` + `overflow-hidden` do the clipping.

ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Hydration flash: `setTailwindPrimary` in `themeSelector.ts:26` only sets `--color-primary` when `typeof window !== "undefined"`, so the server-rendered/first-paint value falls back to the Tailwind `@theme` default `#93c5fd` (`styles/global.css:14`). Non-blue orgs will briefly see a blue panel before the theme applies. This pre-exists for the small `--color-primary-text` icon fills, but at full-panel scale it becomes noticeable. Worth a quick check that the theme is applied before this modal can mount; if not, gating the modal render on theme readiness would avoid the flash.

ThinuwanW: 🤖 **Claude** · 💬 **Note**

📁 **File-level observation**

Missing paired change on the consumer. `BirthdayModalShell.tsx:43` renders this SVG inside `<div className="relative h-full w-full bg-white">`. That hardcoded `bg-white` was consistent while the panel was cream `#FFFBF0`, but it is now the only surface behind a theme-colored panel, so any area the SVG does not cover (rounded-corner region, and the whole panel if the fill fails to resolve) renders pure white against the org theme color. Also note the codebase's established pattern for a themed modal illustration panel is the opposite of what this PR does: `frontend/src/enterprise/common/assets/images/QuickSetupSuccessImage.tsx:7` puts the theme color on the wrapper div via a Tailwind utility (`bg-primary-background`) and keeps the artwork itself untouched. Preferred fix: revert the `fill` on this path to transparent/none and change the shell wrapper to the themed utility (e.g. `bg-primary-background`) — one line, no risk of the SVG and its backdrop disagreeing. Minimum fix if the fill stays here: update `bg-white` in `BirthdayModalShell.tsx:43` to the same token so the backdrop matches.

ThinuwanW: 🤖 **Claude** · 💬 **Note**

Fragile way to consume the token: a raw `var()` inside an SVG presentation attribute, with no fallback. Two failure modes. (1) If `--color-primary` is not defined in the render context — the component rendered in isolation (a future Jest/RTL or Storybook render that does not load `styles/global.css`, an OG-image/print render), the declaration is invalid at computed-value time and behaves as `unset`; `fill` is an inherited property and the root `<svg>` sets `fill="none"`, so the entire 499x603 panel paints transparent and you see the shell's white div instead of any illustration background. (2) Presentation-attribute `var()` support is weaker than `style`/class support in older engines, and when the attribute value is rejected you get the same transparent result. Fix: use the Tailwind utility generated from the token instead of the attribute — `--color-primary` is declared in `@theme` in `frontend/styles/global.css:14`, so `className="fill-primary"` exists and resolves through the same runtime override that `themeSelector.ts` sets on `documentElement`; this also matches how the rest of the app consumes these tokens (`bg-primary-background`, `border-primary-accent`, `text-primary-text`). If you keep the attribute form, at least declare an explicit fallback preserving the previous behaviour: `fill="var(--color-primary, #FFFBF0)"`.

frontend/src/community/people/assets/images/BirthdayCelebration.tsx
```diff
@@ -13,7 +13,7 @@ const BirthdayCelebration = (): JSX.Element => {
       <g clipPath="url(#clip0_929_1835)">
         <path
           d="M-0.078125 42.69C-0.078125 19.113 19.0348 0 42.6119 0H498.022V603H42.6119C19.0349 603 -0.078125 583.887 -0.078125 560.31V42.69Z"
```
ThinuwanW, line 15: 🤖 **Claude** · 💬 **Note**

Duplicated geometry, now with divergent fills. This exact `d` string is repeated verbatim inside `clipPath id="clip0_929_1835"` (line 593), which is what clips the whole illustration group. After this PR the two copies no longer look alike (`var(--color-primary)` here, dead `fill="white"` there — `fill` is ignored inside `clipPath`), which makes it easy for a future corner-radius or size tweak to be applied to one copy only, silently clipping the artwork against a differently shaped background. Extract the geometry to a module-level constant, e.g. `const PANEL_PATH = "M-0.078125 42.69C...";`, use `d={PANEL_PATH}` in both places, and drop the meaningless `fill="white"` from the clipPath copy.

frontend/src/community/people/assets/images/BirthdayCelebration.tsx
ThinuwanW: 🤖 **Claude** · 💬 **Note**

📁 **File-level observation**

Bundle cost of keeping this as an inline SVG component. `BirthdayModalController` is statically imported by `BaseLayout.tsx:12` and rendered at `BaseLayout.tsx:115`, which wraps every authenticated page, so this 618-line SVG (600+ hardcoded shape nodes, tens of KB of JS) ships in the shared client bundle for every page load even though the modal is shown at most once a day and for most users once a year. The only justification for the inline form is runtime theming, and the wrapper-div approach described above achieves that with a static asset. Recommend either moving the artwork to `~public` (`` + themed wrapper, matching `QuickSetupSuccessImage`), or at minimum lazy-loading the controller in `BaseLayout` with `next/dynamic(..., { ssr: false })` so the illustration is fetched only when a birthday notification actually exists.

frontend/src/community/people/assets/images/BirthdayCelebration.tsx
```diff
@@ -13,7 +13,7 @@ const BirthdayCelebration = (): JSX.Element => {
       <g clipPath="url(#clip0_929_1835)">
```
ThinuwanW, line 13: 🤖 **Claude** · 💬 **Note**

Decorative SVG is not hidden from assistive tech. This artwork carries no information — the modal already exposes its own label (`ariaLabel={heading}` in `BirthdayModalShell.tsx:41`) and the greeting text lives in the content slot — but the `<svg>` has no `role`/`aria-hidden`, so screen readers may announce it as an unlabelled graphic and some AT expose the child shapes. Add `aria-hidden="true"` and `focusable="false"` to the root `<svg>` element (lines 5-12). This keeps the modal's announced content limited to the translated heading/body, consistent with the guideline that every graphic either has a translated accessible label or is explicitly marked decorative.

frontend/src/community/people/assets/images/BirthdayCelebration.tsx
ThinuwanW: 🤖 **Claude** · 💬 **Note**

📁 **File-level observation**

No component-level test coverage for the birthday modal UI. The feature's utils are tested (`frontend/src/community/people/utils/birthdayNotificationUtils.test.ts`), but there is no `.test.tsx` anywhere under `BirthdayModals/` or for this asset, so the theming intent introduced by this PR is unverified and a future refactor can silently revert it. Suggest a small `BirthdayModalShell.test.tsx` using `@testing-library/react` with the `MockTheme` wrapper that renders the shell and asserts the panel path resolves to the theme token (and that the heading/body render), rather than a full SVG snapshot which would be noisy for 600 shape nodes.

SanithuGunasena07: Updated fill to var(--color-primary-background)

SanithuGunasena07: Updated fill to var(--color-primary-background)

SanithuGunasena07: Pre existing - out of scope for PR

SanithuGunasena07: Pre existing - out of scope for PR

SanithuGunasena07: not required

SanithuGunasena07: not required

frontend/src/community/people/assets/images/BirthdayCelebration.tsx
SanithuGunasena07: pre existing code not touched in this PR hense out of scope

frontend/src/community/people/assets/images/BirthdayCelebration.tsx
SanithuGunasena07: pre existing code not touched in this PR hense out of scope

frontend/src/community/people/assets/images/BirthdayCelebration.tsx
SanithuGunasena07: pre existing behaviour out of scope for PR

SanithuGunasena07: not required

hasalarootcode (approved):

akilarootcode (approved):

PR #1764, opened 2026-08-14
fix: show confetti animation for self birthday notification

SanithuGunasena07:
Confetti wasn't wired into the birthday modal at all, so it never showed for anyone. Added it to fire only when the viewer is the birthday person, not for colleague notifications.


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds an optional `showConfetti` prop to the shared `BirthdayModalShell` and enables it for `SelfBirthdayModal`, rendering the enterprise `Confetti` component alongside the modal. The change is small and the prop plumbing is type-safe, but it introduces the first community→enterprise import of `Confetti` without a corresponding `src/fallback/` stub, which will break the community build since `src/enterprise` is a git submodule that is not present there. Secondary concerns: the confetti's `requestAnimationFrame` loop now runs unbounded for the entire (potentially indefinite) lifetime of the modal, diverging from the established 2-second-timeout pattern used by `FinishQuickSetupModal`.

Found **9** new issue(s): 🔴 1 important, 🟡 6 suggestion(s), 🟣 2 nit(s)

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
```diff
@@ -6,6 +6,7 @@ import { useTranslator } from "~community/common/hooks/useTranslator";
 import { concatStrings } from "~community/common/utils/commonUtil";
 import BirthdayCelebration from "~community/people/assets/images/BirthdayCelebration";
 import { EmployeeBirthdayType } from "~community/people/types/BirthdayNotificationTypes";
+import Confetti from "~enterprise/common/components/Confetti/Confetti";
```

Missing fallback stub — this will break the community build. `frontend/src/enterprise` is a private git submodule (see `.gitmodules`, `frontend-src` → `skapp-ep-fe-src`) that is not checked out in community builds. `tsconfig.json` maps `~enterprise/*` to `["./src/enterprise/*", "./src/fallback/*"]`, and `jest.config.js` maps `^~enterprise(.*)$` unconditionally to `<rootDir>/src/fallback$1`. Every other community file that imports from `~enterprise/common/components/*` has a matching stub under `src/fallback/` (UpgradeOverlay, Badge, LogoColorLoader, QuickSetupContainer, SubmitRequestModalController, the CommonModalController document modals, etc.), but there is no `frontend/src/fallback/common/components/Confetti/Confetti.tsx`. Since `BirthdayModalShell` is reachable from `BaseLayout` → `BirthdayModalController`, community builds will fail module resolution, and any future Jest test that renders those trees will fail too. Fix: add `frontend/src/fallback/common/components/Confetti/Confetti.tsx` following the existing stub convention (`const Confetti = () => <></>; export default Confetti;`) — or better, see the architecture comment about relocating Confetti to `~community`.

📁 **File-level observation**

Architecture: `Confetti` is a generic, purely visual component with zero enterprise-specific dependencies — its only non-React import is `ZIndexEnums` from `~community/common/enums/CommonEnums`. Keeping it under `~enterprise` and adding an empty fallback stub means community users get the birthday modal (a fully community feature) with no confetti, silently turning a cosmetic detail into an enterprise-gated one. Consider moving it to `~community/common/components/atoms/Confetti/Confetti.tsx` and updating the three existing enterprise consumers (`FinishQuickSetupModal`, `PeopleSyncImportSuccessState`, and this file) to import from there. That removes the need for a stub entirely and gives both editions the intended behavior. Separately, if the enterprise import is kept, consider rendering `<Confetti />` inside `SelfBirthdayModal` rather than adding a `showConfetti` flag to the shared shell — that keeps the enterprise dependency out of the shell that `ColleagueBirthdayModal` also uses.

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
```diff
@@ -24,57 +26,61 @@ const BirthdayModalShell: FC<Props> = ({
...
-      }
-      content={
-        <div className="flex h-[555px] flex-col items-center justify-center gap-4 text-center">
-          {total > 1 && (
-            <p className="sr-only">
-              {translateAria(["notificationPosition"], {
-                current: position,
-                total
+    <>
+      {showConfetti && <Confetti />}
```

Unbounded animation loop. `Confetti` starts a `requestAnimationFrame` loop that only stops on unmount; its 200 particles all fall past the canvas within roughly a second, after which the loop keeps calling `clearRect` at 60fps over an empty array indefinitely. Here the modal has no auto-dismiss (`useBirthdayNotifications` only advances on user `onDismiss`), so if the user leaves the tab open the loop burns CPU/battery for the whole session. The established pattern in `FinishQuickSetupModal.tsx` handles exactly this by unmounting the confetti after a timer (`useEffect` with `setTimeout(() => setShowGif(false), 2000)`). Mirror that here — hold `showConfetti` in local state gated by a ~2s timeout (with `clearTimeout` cleanup), or add a self-terminating condition inside `Confetti` (`if (particles.length === 0) return;` before scheduling the next frame).

Accessibility of the new animation. Two gaps worth covering now that confetti is shown to community users on a modal that also carries `sr-only` announcements: (1) no `prefers-reduced-motion` handling — guard the render with `window.matchMedia("(prefers-reduced-motion: reduce)").matches` (read inside an effect so SSR is safe) and skip the animation when set; (2) the confetti `<canvas>` is portaled into `document.body` after MUI's modal manager has already applied `aria-hidden` to existing body children, so it is not covered by that pass — add `aria-hidden="true"` to the canvas element to keep this purely decorative node out of the accessibility tree.

frontend/src/community/people/components/molecules/BirthdayModals/SelfBirthdayModal.tsx
```diff
@@ -35,6 +35,7 @@ const SelfBirthdayModal: FC<Props> = ({
       position={position}
       total={total}
       onDismiss={onDismiss}
+      showConfetti
```
ThinuwanW, line 35: 🤖 **Claude** · 🟣 **Nit**

No test coverage for the new branch. The `BirthdayModals` directory has no `.test.tsx` files at all, and this PR introduces conditional rendering (`showConfetti` true for self, absent for colleague) that is easy to regress. Consider adding `BirthdayModalShell.test.tsx` with the `MockTheme` wrapper, asserting the confetti canvas is present when `showConfetti` is passed and absent otherwise. Note this test will only be meaningful once the `src/fallback` Confetti stub exists — under Jest, `~enterprise` always resolves to `src/fallback`, so a stub returning an empty fragment would make the presence assertion untestable; that is another argument for moving `Confetti` into `~community`.

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx

Eager import puts the confetti canvas code in the shared bundle for every route. `BirthdayModalShell` is reachable from `BaseLayout` (BaseLayout -> BirthdayNotificationProvider -> BirthdayModalController -> SelfBirthdayModal -> BirthdayModalShell), and BaseLayout wraps the whole app. A static `import Confetti from ...` therefore ships the particle/canvas code to every user on every page load, even though it renders on at most one day per year per user. Load it lazily instead: `const Confetti = dynamic(() => import("~enterprise/common/components/Confetti/Confetti"), { ssr: false });` — this also makes the component's internal `isMounted` SSR guard redundant, and keeps the cost off the critical path. (Same import-graph reach is why the missing fallback stub is fatal rather than route-local.)

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx

The confetti canvas is sized once at mount and never re-sized, and this call site is the first one where that matters. In `Confetti`, `canvas.width = window.innerWidth` / `canvas.height = window.innerHeight` run once inside the mount effect with no `resize` listener. The existing enterprise call sites are short-lived (FinishQuickSetupModal unmounts the confetti after a 2s `setTimeout`), but the birthday modal has `isOpen` hardcoded true and stays mounted until the user clicks close — there is no auto-dismiss — so a window resize, sidebar toggle, devtools open, or mobile orientation change during the celebration leaves the drawing buffer at the old dimensions while the CSS `width:100%/height:100%` stretches it, and particle spawn/despawn bounds (`canvas.width`, `canvas.height`) no longer match the viewport, so particles are drawn in the wrong place. Add a `window.addEventListener("resize", ...)` in the Confetti effect that re-sets `canvas.width/height` (and remove it in the cleanup). While there, scale by `window.devicePixelRatio` and `ctx.scale(dpr, dpr)` — on HiDPI screens the current 1x buffer renders visibly blurry squares.

Latent bug in the component this PR newly depends on: `Confetti.update()` calls `particles.splice(index, 1)` on the same array that the animation loop is iterating with `particles.forEach(...)`. Removing an element during `forEach` shifts every later element down one index, so the particle that moves into the freed slot is skipped for that frame (not updated and not drawn) — it visibly stutters/flickers, and with several particles exiting in the same frame the skipping compounds. Fix by not mutating during iteration: either mark particles with an `isDead` flag in `update()` and do `particles = particles.filter(p => !p.isDead)` after the loop, or iterate backwards with `for (let i = particles.length - 1; i >= 0; i--)`. Flagging as a nit because the file lives in the `frontend/src/enterprise` submodule and can't be changed in this PR — but it should be tracked there, since this change makes the animation visible to a much wider audience.

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
```diff
@@ -15,6 +16,7 @@ interface Props {
   position: number;
   total: number;
   onDismiss: () => void;
+  showConfetti?: boolean;
```
ThinuwanW, line 20: 🤖 **Claude** · 🟡 **Suggestion**

Prefer inversion of control over a boolean flag on the shared shell. `showConfetti?: boolean` hard-codes one specific decoration into a component whose job is layout, and it forces the shell — which is also the render path for `ColleagueBirthdayModal`, the case that must NOT have confetti — to carry the dependency. A `decoration?: ReactNode` prop (rendered as `{decoration}` before `<LargeModal>`) keeps `BirthdayModalShell` presentation-agnostic, lets `SelfBirthdayModal` own the celebration by passing `decoration={<Confetti />}`, and means any future variant (e.g. a work-anniversary modal) can supply its own visual without another boolean. It also makes the behaviour trivially testable — a test can pass a sentinel node instead of having to mock the module. Note this is about the shell's API shape, independent of where the `Confetti` file itself should live.

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
SanithuGunasena07: Moved to community

SanithuGunasena07: Moved to community

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
SanithuGunasena07: added the same 2s setTimeout + clearTimeout pattern as FinishQuickSetupModal

SanithuGunasena07: Pre Existing out of scope for PR

frontend/src/community/people/components/molecules/BirthdayModals/SelfBirthdayModal.tsx
SanithuGunasena07: not required

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
SanithuGunasena07: next/dynamic isnt used any where previously in the Skapp codebase leaving it as it is

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
SanithuGunasena07: since 2s time out is added now this is not needed

SanithuGunasena07: Pre Existing out of scope for PR

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
SanithuGunasena07: leaving as it is

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
```diff
@@ -24,57 +26,72 @@ const BirthdayModalShell: FC<Props> = ({
...
   const translateAria = useTranslator("peopleAria", "birthdayNotifications");
   const imageUrl = useGetImageUrl(employee.authPic ?? "");
+  const [isConfettiVisible, setIsConfettiVisible] = useState(showConfetti);
+
+  useEffect(() => {
+    if (!showConfetti) return;
+
+    const timer = setTimeout(() => {
+      setIsConfettiVisible(false);
+    }, 4000);
```
hasalarootcode: move magic numbers to constants

hasalarootcode (dismissed):

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
SanithuGunasena07: Extracted 4000 into const CONFETTI_VISIBLE_DURATION_MS = 4000;

hasalarootcode (dismissed):

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
```diff
@@ -1,11 +1,14 @@
...
+import { FC, useEffect, useState } from "react";
 
 import useGetImageUrl from "~community/common/hooks/useGetImageUrl";
 import { useTranslator } from "~community/common/hooks/useTranslator";
 import { concatStrings } from "~community/common/utils/commonUtil";
 import BirthdayCelebration from "~community/people/assets/images/BirthdayCelebration";
 import { EmployeeBirthdayType } from "~community/people/types/BirthdayNotificationTypes";
+import Confetti from "~community/common/components/atoms/Confetti/Confetti";
+
+const CONFETTI_VISIBLE_DURATION_MS = 4000;
```
akilarootcode: move to constants

akilarootcode (changes requested):

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
SanithuGunasena07: moved

hasalarootcode (dismissed):
Approved with sonar issues due to a existing file movement

hasalarootcode (approved):

PR #1765, opened 2026-08-17
Feat/leave policy carry over implementation - FE

hasalarootcode:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR replaces the fixed carryover-date dropdown with a free date picker and renames `carryOverDate`/`carryoverDate` to `carryoverExpiryDate` across form data, request payload and response DTO, plus adds 0.5-step validation for day fields. The frontend work itself is coherent, but the payload/response rename does not match the backend, which still exposes and requires `carryoverDate` (`LeavePolicyAccrualDetailDto`, `LeavePolicyResponseDto`, `LeavePolicyValidationUtil.validateCarryoverSetup`) — as it stands every carryover-enabled policy creation will be rejected. There are also mismatches between the new step validation and the payload mapping/backend types (waiting period), and several consistency/UX gaps around the new date picker.

Found **27** new issue(s): 🔴 5 important, 🟡 9 suggestion(s), 🟣 13 nit(s)

frontend/src/community/leave/utils/leavePolicy/leavePolicyUtils.ts
```diff
@@ -84,7 +124,10 @@ export const mapLeavePolicyFormToPayload = (
         ? Number(formData.accrualCapDays)
         : undefined,
       isCarryoverEnabled: formData.canCarryOver,
-      carryoverDate: formData.canCarryOver ? formData.carryOverDate : undefined,
+      carryoverExpiryDate:
```
ThinuwanW, line 129: 🤖 **Claude** · 🔴 **Important**

API contract break: the payload key was renamed to `carryoverExpiryDate`, but the backend request DTO still declares `carryoverDate` (backend/src/main/java/com/skapp/community/leaveplanner/payload/request/LeavePolicyAccrualDetailDto.java:23), and `LeavePolicyValidationUtil.validateCarryoverSetup` (line 73) throws `LEAVE_ERROR_LEAVE_POLICY_CARRYOVER_DATE_REQUIRED` when `carryoverDate` is null/blank. With this change Jackson will drop the unknown `carryoverExpiryDate` property, so every carryover-enabled policy creation fails at validation. The same applies in reverse for reads: `LeavePolicyResponseDto` (backend) still returns `carryoverDate`, so the renamed `LeavePolicyResponseDto.carryoverExpiryDate` / `LeavePolicyType.carryoverExpiryDate` fields in LeavePolicyTypes.ts will always be `undefined` at runtime. Either land the matching backend rename (DTOs, entity/column, validation util, service impl) in the same release, or keep sending/reading `carryoverDate` on the wire and only rename the frontend-internal `LeavePolicyFormData` field.

frontend/src/community/leave/utils/validations.ts
```diff
@@ -179,33 +195,49 @@ export const leavePolicyWizardValidation = (
...
             is: true,
             then: (schema) =>
-              schema.test(
-                "max-carryover-days-valid",
-                translateText(["errors", "maxCarryOverDaysInvalid"]),
-                (value) =>
-                  !value ||
-                  isNumberInRange(value, MIN_POLICY_DAYS, MAX_POLICY_DAYS)
-              )
+              schema
```
ThinuwanW, line 240: 🤖 **Claude** · 🔴 **Important**

The carryover expiry date is now optional on the frontend — `mapLeavePolicyFormToPayload` omits it when empty, the tooltip says "Leave it empty if carried over days should never expire", and SummaryStep renders "Never expires". The backend does not support this: `LeavePolicyValidationUtil.validateCarryoverSetup` rejects a null/blank carryover date whenever `isCarryoverEnabled` is true. Either add a Yup rule requiring `carryoverExpiryDate` when `canCarryOver` is true (and add it to `STEP_FIELDS[ENTITLEMENT_SETUP]` in LeavePolicyWizard.tsx so the step blocks on it), or relax the backend validation to allow a null expiry date. Note `carryoverExpiryDate` currently has no validation entry at all in this schema.

frontend/src/community/leave/utils/validations.ts
```diff
@@ -179,33 +195,49 @@ export const leavePolicyWizardValidation = (
...
-                (value) => isNumberInRange(value, MIN_WAITING_PERIOD_DAYS)
-              )
+              schema
+                .test(
+                  "waiting-period-days-valid",
+                  translateText(["errors", "waitingPeriodDaysRequired"]),
+                  (value) => isNumberInRange(value, MIN_WAITING_PERIOD_DAYS)
+                )
+                .test(
+                  "waiting-period-days-step-valid",
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The new `waiting-period-days-step-valid` test allows waiting period days in multiples of 0.5, but waiting period is a whole-day concept everywhere else: `mapLeavePolicyFormToPayload` does `Math.round(Number(formData.waitingPeriodDays))` (leavePolicyUtils.ts:121) and the backend field is `Integer waitingPeriodDays`. So a user who enters 1.5 passes validation and silently gets 2 stored, with no feedback. Validate waiting period days as an integer (e.g. `Number.isInteger(Number(value))` with a dedicated message) instead of reusing the 0.5-step check.

frontend/src/community/leave/components/organisms/LeavePolicyWizard/EntitlementSetupStep.tsx
```diff
@@ -71,7 +93,6 @@ const EntitlementSetupStep: FC<Props> = ({
             <InputField
```
ThinuwanW, line 93: 🤖 **Claude** · 🟡 **Suggestion**

`type="number"` was removed from the `accrualDays` input only; `waitingPeriodDays` (line 154), `accrualCapDays` (line 192) and `maxCarryOverDays` (line 242) still use `type="number"`. This is inconsistent for four fields that share the same numeric semantics and the same new 0.5-step validation, and it also means `accrualDays` now accepts arbitrary text on mobile (no numeric keypad) — non-numeric input falls through to the range test and surfaces "Days accrued must be greater than 0.", which is misleading. Pick one approach for all four (either drop `type="number"` everywhere and rely on Yup, or keep it and set `step="0.5"`).

frontend/src/community/leave/utils/validations.ts
```diff
@@ -179,33 +195,49 @@ export const leavePolicyWizardValidation = (
...
-                translateText(["errors", "maxCarryOverDaysInvalid"]),
-                (value) =>
-                  !value ||
-                  isNumberInRange(value, MIN_POLICY_DAYS, MAX_POLICY_DAYS)
-              )
+              schema
+                .required(
+                  translateText(["errors", "maxCarryOverDaysRequired"])
+                )
+                .test(
```
ThinuwanW, line 244: 🤖 **Claude** · 🟡 **Suggestion**

`maxCarryOverDays` is the only day field that did not get the new `isPolicyDaysStepValid` test, even though it is validated against the same `MIN_POLICY_DAYS` (0.5) / `MAX_POLICY_DAYS` (365) bounds as `accrualDays`. A user can enter 3.2 carry-over days and it will pass. Add a `max-carryover-days-step-valid` test mirroring the other three (a `maxCarryOverDaysStepInvalid` key would need to be added to leaveModule.json). Also, the `!value ||` guard inside `max-carryover-days-valid` is now dead since `.required()` was added above it — it can be dropped.

frontend/src/community/leave/components/organisms/LeavePolicyWizard/EntitlementSetupStep.tsx
```diff
@@ -192,21 +213,26 @@ const EntitlementSetupStep: FC<Props> = ({
...
               <div className="flex flex-1 flex-col gap-1.5">
-                <Dropdown
-                  id="leave-policy-carryover-date"
-                  label={translateText(["carryOverDateLabel"])}
-                  value={formData.carryOverDate}
-                  placeholder={translateText(["carryOverDatePlaceholder"])}
-                  options={carryoverDateOptions}
-                  onChange={(value: string) =>
-                    onChange({ carryOverDate: value })
+                <InputDate
```
ThinuwanW, line 220: 🤖 **Claude** · 🟡 **Suggestion**

`InputDate` offers no way to clear a selected date — its `handleDateChange` early-returns on `null` and there is no clear button. Once an admin picks an expiry date they cannot get back to the "Never expires" state that the tooltip advertises (short of toggling carry over off and back on, which also wipes `maxCarryOverDays`). Either add a clear affordance (e.g. a small "Clear"/reset control that calls `setCarryoverExpiryDate(undefined)` and `onChange({ carryoverExpiryDate: "" })`) or reword the tooltip so it does not promise an unreachable state.

frontend/src/community/leave/constants/leavePolicyConstants.ts
```diff
@@ -69,12 +71,11 @@ export const accrualFrequencyItemList = [
...
-  { id: "january-1", labelKey: "january1", value: "01-01" },
-  { id: "april-1", labelKey: "april1", value: "04-01" },
-  { id: "july-1", labelKey: "july1", value: "07-01" },
-  { id: "october-1", labelKey: "october1", value: "10-01" }
-];
+export const CARRYOVER_EXPIRY_DATE_FORMAT = "MM-dd";
+
+export const CARRYOVER_EXPIRY_DISPLAY_FORMAT = "dd MMMM";
+
+export const CARRYOVER_EXPIRY_REFERENCE_YEAR = 2025;
```
ThinuwanW, line 80: 🤖 **Claude** · 🟣 **Nit**

`CARRYOVER_EXPIRY_REFERENCE_YEAR = 2025` is a bare magic number with a load-bearing property: 2025 is not a leap year, so 29 February can never be selected or parsed as a carryover expiry date (`parseCarryoverExpiryDate("02-29")` returns `undefined` and `formatCarryoverExpiryDate` returns ""). Add a short comment stating that the year is only a placeholder for a recurring month/day and that a non-leap year is chosen deliberately — otherwise a future "let's use the current year" change would silently allow 02-29 values that stop rendering in non-leap years.

frontend/src/community/leave/utils/leavePolicy/leavePolicyUtils.ts
```diff
@@ -1,15 +1,55 @@
...
+  CARRYOVER_EXPIRY_DISPLAY_FORMAT,
+  CARRYOVER_EXPIRY_REFERENCE_YEAR
+} from "~community/leave/constants/leavePolicyConstants";
 import {
   AddLeavePolicyPayload,
   LeavePolicyFormData,
   PolicyType
 } from "~community/leave/types/LeavePolicyTypes";
 
+export const toCarryoverExpiryMonthDay = (isoDate: string): string => {
```
ThinuwanW, line 19: 🤖 **Claude** · 🟡 **Suggestion**

The three new date helpers (`toCarryoverExpiryMonthDay`, `parseCarryoverExpiryDate`, `formatCarryoverExpiryDate`) and the new `isPolicyDaysStepValid` logic have no tests, although `src/community/leave/utils/__test__/` already exists with `validations.test.ts` and `accrualPreviewUtils.test.ts`. Worth adding cases for: valid ISO round-trip, invalid/empty input, the 02-29 edge case, and 0.5-step accept/reject for the validation helper.

frontend/src/community/leave/utils/leavePolicy/leavePolicyUtils.ts
```diff
@@ -1,15 +1,55 @@
...
+};
+
+export const getCarryoverExpiryReferenceDate = (): DateTime =>
+  DateTime.fromObject({
+    year: CARRYOVER_EXPIRY_REFERENCE_YEAR,
+    month: 1,
+    day: 1
+  });
+
+export const formatCarryoverExpiryDate = (monthDay: string): string =>
```
ThinuwanW, line 48: 🤖 **Claude** · 🟣 **Nit**

`formatCarryoverExpiryDate` returns an empty string for unparseable input, and SummaryStep only falls back to "Never expires" when `formData.carryoverExpiryDate` is falsy — so a stored-but-unparseable month/day (e.g. "02-29", or a value coming back from the API in a different shape) renders as a blank summary row rather than any fallback. Consider returning the raw `monthDay` (or a "-" sentinel) when parsing fails so the UI never shows an empty value.

frontend/src/community/leave/utils/validations.ts
```diff
@@ -179,33 +195,49 @@ export const leavePolicyWizardValidation = (
...
             then: (schema) =>
-              schema.test(
-                "waiting-period-days-valid",
-                translateText(["errors", "waitingPeriodDaysRequired"]),
-                (value) => isNumberInRange(value, MIN_WAITING_PERIOD_DAYS)
-              )
+              schema
+                .test(
+                  "waiting-period-days-valid",
+                  translateText(["errors", "waitingPeriodDaysRequired"]),
```
ThinuwanW, line 211: 🤖 **Claude** · 🟣 **Nit**

These blocks were restructured here, so worth fixing while in the area: the range tests reuse the "...Required" translation keys as their failure messages. Entering 0.5 for the accrual cap (min is `MIN_ACCRUAL_CAP_DAYS` = 1) shows "Please enter the maximum accrual balance.", and 0 for the waiting period shows "Please enter the number of waiting period days." — both are wrong for a value that was in fact entered. Add distinct `waitingPeriodDaysInvalid` / `accrualCapInvalid` messages that state the actual bound.

frontend/src/community/leave/components/organisms/LeavePolicyWizard/SummaryStep.tsx
```diff
@@ -4,15 +4,17 @@ import { useTranslator } from "~community/common/hooks/useTranslator";
...
-  carryoverDateItemList,
   firstAccrualItemList,
   receiveAccruedTimeItemList
 } from "~community/leave/constants/leavePolicyConstants";
 import {
   LeavePolicyFormData,
   LeavePolicyWizardSteps
 } from "~community/leave/types/LeavePolicyTypes";
-import { buildTranslatedOptionList } from "~community/leave/utils/leavePolicy/leavePolicyUtils";
+import {
```
ThinuwanW, line 14: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

Now that `maxCarryOverDays` is `.required()` when `canCarryOver` is true, the `carryOverNoLimit` ("No limit") fallback on the max-carry-over summary row (around line 156-161) is unreachable for carryover-enabled policies — the only remaining path to it is the carryover-disabled case, where showing "No limit" reads oddly. Either drop the branch or change the disabled case to render "-".

frontend/src/community/leave/components/organisms/LeavePolicyWizard/EntitlementSetupStep.tsx
```diff
@@ -42,16 +49,31 @@ const EntitlementSetupStep: FC<Props> = ({
     "options"
   );
 
+  const [carryoverExpiryDate, setCarryoverExpiryDate] = useState<
```
ThinuwanW, line 52: 🤖 **Claude** · 🟣 **Nit**

`carryoverExpiryDate` is held in component-local `useState` that is only seeded once from `formData.carryoverExpiryDate` via the lazy initializer, duplicating Formik state. It happens to work today because `LeavePolicyStepContent` unmounts this step when navigating to Summary (so it re-seeds on remount), but it will silently drift if the step is ever kept mounted or Formik values are reset externally. Consider deriving the `DateTime` from `formData.carryoverExpiryDate` with `useMemo` and passing a setter that writes back through `onChange`, keeping Formik as the single source of truth.

frontend/src/community/leave/utils/leavePolicy/leavePolicyUtils.ts
ThinuwanW, line 129: 🤖 **Claude** · 🔴 **Important**

The submit path has no user-actionable error for the exact state this PR introduces. `backend/.../util/LeavePolicyValidationUtil.java:73` throws `LEAVE_ERROR_LEAVE_POLICY_CARRYOVER_DATE_REQUIRED` (`api.error.leave.leave-policy-carryover-date-required`) whenever `isCarryoverEnabled` is true and the date is null/blank — which is precisely what `mapLeavePolicyFormToPayload` now sends when the admin follows the new tooltip ("Leave it empty if carried over days should never expire"). `getLeavePolicyErrorToastKeys` (same file) only switches on `LEAVE_ERROR_LEAVE_POLICY_ALREADY_EXISTS` and `COMMON_ERROR_ACCESS_DENIED`, so this falls through to the default branch and the admin sees "Oops! Something went wrong / Failed to create the policy. Please try again or contact support." after clicking "Create policy" on step 3, with no hint that the expiry date is the problem. Also note `LEAVE_ERROR_LEAVE_POLICY_CARRYOVER_DATE_INVALID` and `LEAVE_ERROR_LEAVE_POLICY_MAX_CARRYOVER_DAYS_INVALID` are unmapped. Fix: either make the backend accept a null carryover date (the "never expires" semantics this PR promises — note nothing in the backend currently *consumes* `LeavePolicy.carryoverDate`, so the semantics are unimplemented either way), or add `LEAVE_ERROR_LEAVE_POLICY_CARRYOVER_DATE_REQUIRED`/`_INVALID` to `~community/common/constants/errorMessageKeys.ts`, map them in the switch, and add matching toast title/description keys under `createPolicy`.

frontend/src/community/leave/types/LeavePolicyTypes.ts
```diff
@@ -50,7 +50,7 @@ export interface LeavePolicyType {
   waitingPeriodDays?: number | null;
   accrualCapDays?: number | null;
   isCarryoverEnabled?: boolean | null;
-  carryoverDate?: string | null;
+  carryoverExpiryDate?: string | null;
```
ThinuwanW, line 99: 🤖 **Claude** · 🔴 **Important**

Distinct from the request-payload mismatch: the *response*-side field was renamed too, and TypeScript cannot catch this one. `LeavePolicyType.carryoverExpiryDate` (line 53) and `LeavePolicyResponseDto.carryoverExpiryDate` (line 159) model objects deserialized from the backend, whose `LeavePolicyResponseDto` still declares `private String carryoverDate;` (verified in backend/src/main/java/.../payload/response/LeavePolicyResponseDto.java:39, no `@JsonProperty` alias). Because these come off the wire as untyped JSON, the compiler will happily let any future reader do `policy.carryoverExpiryDate` and silently get `undefined` — no build error, no runtime error, just a missing value. It happens to be latent today only because nothing in the frontend reads the field yet (`buildAccrualPreview` and `LeavePoliciesTable` don't touch it), which makes it a trap for the next person. Fix: rename the backend field (request DTO, response DTO, and the `LeavePolicy` entity/column) in the same change, or revert these two to `carryoverDate` and keep the rename confined to `LeavePolicyFormData`.

frontend/src/community/leave/utils/validations.ts
```diff
@@ -32,6 +33,16 @@ const isNumberInRange = (
   );
 };
 
+const isPolicyDaysStepValid = (value: string | undefined): boolean => {
```
ThinuwanW, line 36: 🤖 **Claude** · 🟡 **Suggestion**

The new "multiples of 0.5" business rule is enforced only in the browser. `backend/.../constant/LeavePolicyConstant.java` defines only `MIN_DAYS=0.5F`, `MAX_DAYS=365F`, `MIN_WAITING_PERIOD_DAYS=1`, `MIN_ACCRUAL_CAP_DAYS=1F`, and `LeavePolicyValidationUtil.validateAccrualSetup` does only range checks — there is no step/modulo check. So a direct POST to the add-policy endpoint with `accrualDays: 1.3` or `accrualCapDays: 2.7` is accepted and persisted, and the projected balances in `buildAccrualPreview` will then show values the UI claims are impossible. Since the guidelines treat backend validation as the source of truth for field constraints, add the same step check to `validateAccrualSetup` (e.g. a `POLICY_DAYS_STEP` constant plus a `Math.abs(value % STEP) > EPSILON` guard) with a matching `LeaveMessageConstant` key, and surface it via the toast mapping.

frontend/src/community/leave/utils/validations.ts
```diff
@@ -171,6 +182,11 @@ export const leavePolicyWizardValidation = (
               translateText(["errors", "accrualDaysInvalid"]),
               (value) =>
                 isNumberInRange(value, MIN_POLICY_DAYS, MAX_POLICY_DAYS)
+            )
+            .test(
```
ThinuwanW, line 196: 🤖 **Claude** · 🟡 **Suggestion**

Now that `accrualDays` is a free-text input (`type="number"` was dropped from it in this PR), its validation relies entirely on `Number()` coercion, which accepts several non-decimal literal forms: `"1e2"` → 100, `"0x1F"` → 31, `"0b101"` → 5, `"+2"` → 2, `" 1.5 "` → 1.5, and `"Infinity"` (rejected only incidentally by the range test). All of these pass both `isNumberInRange` and `isPolicyDaysStepValid`, so an admin who fat-fingers `0x10` silently creates a policy accruing 16 days. The same file already has the right tool: `allowsOnlyNumbersAndOptionalDecimal()` from `~community/common/regex/regexPatterns` is imported at line 3 and used by `customLeaveAllocationValidation` at line 61. Add a `.test("accrual-days-numeric", ..., (value) => !value || allowsOnlyNumbersAndOptionalDecimal().test(value))` ahead of the range test (and apply it to the other day fields for consistency), or restore `type="number"` on the input.

frontend/src/community/leave/components/organisms/LeavePolicyWizard/EntitlementSetupStep.tsx
ThinuwanW, line 220: 🤖 **Claude** · 🟡 **Suggestion**

The carryover expiry date is the only field in this step with no error plumbing at all, so nothing can ever be reported on it. `InputDate` exposes an `error?: string` prop (InputDate.tsx:59, rendered as a `role="alert"` Typography at line 455) and the component already receives Formik `errors`/`touched`, yet neither is passed — unlike `accrualDays`, `accrualFrequency`, `waitingPeriodDays`, `accrualCapDays` and `maxCarryOverDays`, which all wire `state`/`errorMessage` or `variant`/`errorMessage`. Compounding it, `carryoverExpiryDate` is absent from `STEP_FIELDS[ENTITLEMENT_SETUP]` in LeavePolicyWizard.tsx:41-47, which is the list `handleNext` checks — so even if a Yup rule were added, it would neither block "Next" nor mark the field touched. Fix: add `error={touched.carryoverExpiryDate ? errors.carryoverExpiryDate : undefined}` and add `"carryoverExpiryDate"` to `STEP_FIELDS` so the wiring exists before/when the field's required-vs-optional question is settled.

frontend/src/community/leave/components/organisms/LeavePolicyWizard/EntitlementSetupStep.tsx
```diff
@@ -192,21 +213,26 @@ const EntitlementSetupStep: FC<Props> = ({
...
-                  placeholder={translateText(["carryOverDatePlaceholder"])}
-                  options={carryoverDateOptions}
-                  onChange={(value: string) =>
-                    onChange({ carryOverDate: value })
+                <InputDate
+                  label={translateText(["carryoverExpiryDateLabel"])}
+                  placeholder={translateText([
+                    "carryoverExpiryDatePlaceholder"
+                  ])}
+                  tooltip={translateText(["carryoverExpiryDateTooltip"])}
```
ThinuwanW, line 225: 🤖 **Claude** · 🟡 **Suggestion**

The newly added tooltip is not announced to screen readers, and it's broken in two independent ways in the component you're reusing. (1) `InputDate.tsx:311` passes `aria-label={...}` to the `Tooltip` atom, but `Tooltip`'s Props declares `ariaLabel` (camelCase, Tooltip.tsx:24) and does not spread rest props — TypeScript does not flag unknown *hyphenated* JSX attributes, so the value is accepted and silently discarded. (2) That value comes from `translateAria(["ariaLabel"])`, but `frontend/src/community/common/assets/languages/english/aria/common.json:25-27` defines only `tooltip` and `calendarIcon` under `components.inputDate` — no `ariaLabel`. Since `useTranslator` returns `t(key)` and i18next echoes the key when missing, it would render the literal string `commonAria.components.inputDate.ariaLabel` if it were forwarded at all. This is pre-existing InputDate/Tooltip code, but this PR is the first community-side consumer to render an InputDate tooltip, so the defect ships here. Fix: change InputDate to `ariaLabel={...}` and either add an `ariaLabel` key under `components.inputDate` or reuse the existing `tooltip` key.

frontend/src/community/common/assets/languages/english/leaveModule.json
```diff
@@ -434,9 +430,13 @@
         "leaveTypeRequired": "Please select a leave type.",
         "accrualDaysRequired": "Please enter the number of days employees accrue.",
         "accrualDaysInvalid": "Days accrued must be greater than 0.",
+        "accrualDaysStepInvalid": "Days accrued must be in multiples of 0.5.",
```
ThinuwanW, line 470: 🤖 **Claude** · 🟡 **Suggestion**

`accrualDaysStepInvalid`, `waitingPeriodDaysStepInvalid` and `accrualCapStepInvalid` each hardcode the literal "0.5", duplicating the new `POLICY_DAYS_STEP` constant three times (four once `maxCarryOverDaysStepInvalid` is added per the first-pass finding). Change `POLICY_DAYS_STEP` and the constant and the copy drift apart silently. `TranslatorFunctionType` in validations.ts:16-19 already supports interpolation values, so use e.g. `"accrualDaysStepInvalid": "Days accrued must be in multiples of {{step}}."` and call `translateText(["errors", "accrualDaysStepInvalid"], { step: String(POLICY_DAYS_STEP) })`. Better still, collapse the three near-identical strings into one `policyDaysStepInvalid` key with a `{{field}}` placeholder.

frontend/src/community/leave/types/LeavePolicyTypes.ts
```diff
@@ -156,7 +156,7 @@ export interface LeavePolicyResponseDto {
   waitingPeriodDays: number | null;
   accrualCapDays: number | null;
   isCarryoverEnabled: boolean | null;
-  carryoverDate: string | null;
+  carryoverExpiryDate: string | null;
```
ThinuwanW, line 205: 🤖 **Claude** · 🟣 **Nit**

`LeavePolicyType` (line 39) and `LeavePolicyResponseDto` (line 146) are the same 14-field shape, differing only in optionality — which is why this one rename had to be applied in two places here plus `AddLeavePolicyAccrualPayload` (line 124), and why it is easy to update one and miss another (as happened on the backend side). Derive one from the other, e.g. `export interface LeavePolicyType extends Partial<Omit<LeavePolicyResponseDto, "id"|"name"|"leaveTypeId"|"leaveTypeName"|"leaveTypeEmoji"|"policyType"|"status">>, Pick<LeavePolicyResponseDto, "id"|"name"|...> {}`, so the accrual field names live in exactly one place.

frontend/src/community/leave/types/LeavePolicyTypes.ts
```diff
@@ -109,7 +109,7 @@ export interface LeavePolicyFormData {
   hasAccrualCap: boolean;
   accrualCapDays: string;
   canCarryOver: boolean;
-  carryOverDate: string;
+  carryoverExpiryDate: string;
```
ThinuwanW, line 158: 🤖 **Claude** · 🟣 **Nit**

Casing inconsistency introduced by the rename: `carryoverExpiryDate` (lowercase "o") now sits directly between `canCarryOver` and `maxCarryOverDays` (capital "O") in `LeavePolicyFormData`. Before this PR the form field was `carryOverDate`, which matched its siblings. The mixed casing propagates into `handleCarryOverChange`'s clearing object (EntitlementSetupStep.tsx:62 — `{ carryoverExpiryDate: "", maxCarryOverDays: "" }`) and into `mapLeavePolicyFormToPayload`, and is an easy source of typos that TypeScript catches only for known keys. Pick one convention for the form type: either `carryOverExpiryDate` to match `maxCarryOverDays`/`canCarryOver`, or rename the siblings to the backend's `carryover*` spelling.

frontend/src/community/leave/components/organisms/LeavePolicyWizard/EntitlementSetupStep.tsx
```diff
@@ -42,16 +49,31 @@ const EntitlementSetupStep: FC<Props> = ({
     "options"
   );
 
+  const [carryoverExpiryDate, setCarryoverExpiryDate] = useState<
+    DateTime | undefined
+  >(() => parseCarryoverExpiryDate(formData.carryoverExpiryDate));
+
+  const handleCarryOverChange = (value: boolean) => {
```
ThinuwanW, line 56: 🤖 **Claude** · 🟣 **Nit**

`handleCarryOverChange` and `handleCarryoverExpiryDateChange` have no explicit return type annotation, whereas every handler in the sibling `LeavePolicyWizard.tsx` (`handleClose`, `handleBack`, `handleCancel`, `handleFieldsChange`, `handleEditFromSummary`) is annotated `: void`. Guidelines checklist item 1 requires an explicit type for every return value. Add `: void` to both.

frontend/src/community/leave/components/organisms/LeavePolicyWizard/EntitlementSetupStep.tsx
```diff
@@ -42,16 +49,31 @@ const EntitlementSetupStep: FC<Props> = ({
...
+    if (!value) {
+      setCarryoverExpiryDate(undefined);
+    }
+    onChange({
+      canCarryOver: value,
+      ...(value ? {} : { carryoverExpiryDate: "", maxCarryOverDays: "" })
```
ThinuwanW, line 62: 🤖 **Claude** · 🟣 **Nit**

`handleCarryOverChange` clears the two values but nothing resets their Formik `touched` flags, which now matters because this PR made `maxCarryOverDays` `.required()`. Repro: enable carryover, click "Next" without filling the field (handleNext sets `touched.maxCarryOverDays = true` and returns), switch to "No, it's use it or lose it", then switch back to "Yes" — the field remounts empty with `touched` still true, so the red "Please enter the maximum number of carry over days." error is shown before the admin has typed anything. Fix: have `onChange` also clear touched for the fields it resets (e.g. extend the wizard's `handleFieldsChange` to accept fields to un-touch, or call `formik.setFieldTouched(field, false)` alongside the reset).

frontend/src/community/leave/constants/leavePolicyConstants.ts
```diff
@@ -69,12 +71,11 @@ export const accrualFrequencyItemList = [
   }
 ];
 
-export const carryoverDateItemList = [
-  { id: "january-1", labelKey: "january1", value: "01-01" },
-  { id: "april-1", labelKey: "april1", value: "04-01" },
-  { id: "july-1", labelKey: "july1", value: "07-01" },
-  { id: "october-1", labelKey: "october1", value: "10-01" }
-];
+export const CARRYOVER_EXPIRY_DATE_FORMAT = "MM-dd";
```
ThinuwanW, line 78: 🤖 **Claude** · 🟣 **Nit**

The three new SCREAMING_CASE constants were inserted mid-file, between `accrualFrequencyItemList` (ends line 72) and `firstAccrualItemList` (starts line 80), because that is where the deleted `carryoverDateItemList` happened to sit. The file otherwise groups all scalar constants at the top (lines 11-44) and all option lists after. Move `CARRYOVER_EXPIRY_DATE_FORMAT`, `CARRYOVER_EXPIRY_DISPLAY_FORMAT` and `CARRYOVER_EXPIRY_REFERENCE_YEAR` up next to `POLICY_DAYS_STEP`/`MIN_WAITING_PERIOD_DAYS` so the file keeps one shape.

frontend/src/community/leave/constants/leavePolicyConstants.ts
```diff
@@ -69,12 +71,11 @@ export const accrualFrequencyItemList = [
...
+
+export const CARRYOVER_EXPIRY_DISPLAY_FORMAT = "dd MMMM";
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`CARRYOVER_EXPIRY_DISPLAY_FORMAT = "dd MMMM"` zero-pads the day, so the summary and the input now read "01 January" / "01 April" where the removed dropdown read "1st of January". Luxon's `d MMMM` gives "1 January", which reads more naturally and matches the un-padded style of `formatISODateWithSuffix` in `~community/common/utils/dateTimeUtils`. Worth confirming against the design; the constant is used both here (via `formatCarryoverExpiryDate`) and as `InputDate`'s `inputFormat`, so changing it fixes both call sites at once.

frontend/src/community/leave/components/organisms/LeavePolicyWizard/EntitlementSetupStep.tsx
```diff
@@ -192,21 +213,26 @@ const EntitlementSetupStep: FC<Props> = ({
...
+                  inputFormat={CARRYOVER_EXPIRY_DISPLAY_FORMAT}
+                  isYearHidden
+                  selectedDate={carryoverExpiryDate}
+                  setSelectedDate={setCarryoverExpiryDate}
+                  initialMonthlyView={
+                    carryoverExpiryDate ?? getCarryoverExpiryReferenceDate()
                   }
-                  width="100%"
+                  onchange={handleCarryoverExpiryDateChange}
+                  componentStyle={{ mt: "0rem" }}
```
ThinuwanW, line 234: 🤖 **Claude** · 🟣 **Nit**

Two small regressions from swapping `Dropdown` for `InputDate` here. (1) `componentStyle={{ mt: "0rem" }}` is an inline sx magic value whose only purpose is to cancel InputDate's default top margin so it lines up with the sibling `maxCarryOverDays` InputField — the guidelines ask for this kind of override to live in a colocated `styles.ts`, and a comment or a named style would make the intent obvious. (2) The removed `Dropdown` carried `id="leave-policy-carryover-date"`; `InputDate` accepts no `id` (and its declared `name?: string` prop at InputDate.tsx:79 is never destructured or used), so this field now has no stable DOM hook for e2e tests or automation, and its `component="label"` Typography has no `htmlFor` target. Consider adding an `id` pass-through to InputDate, or at least a `dataTestId`.

frontend/src/community/leave/utils/leavePolicy/leavePolicyUtils.ts
hasalarootcode: Already there, https://github.com/SkappHQ/skapp/pull/1766

frontend/src/community/leave/utils/validations.ts
hasalarootcode: Not needed

frontend/src/community/leave/utils/validations.ts
hasalarootcode: Fixed

frontend/src/community/leave/components/organisms/LeavePolicyWizard/EntitlementSetupStep.tsx
hasalarootcode: added

frontend/src/community/leave/utils/validations.ts
hasalarootcode: Not needed

frontend/src/community/leave/components/organisms/LeavePolicyWizard/EntitlementSetupStep.tsx
hasalarootcode: Not needed

frontend/src/community/leave/constants/leavePolicyConstants.ts
hasalarootcode: Not needed

frontend/src/community/leave/utils/leavePolicy/leavePolicyUtils.ts
hasalarootcode: Not needed

frontend/src/community/leave/utils/leavePolicy/leavePolicyUtils.ts
hasalarootcode: Not needed

frontend/src/community/leave/utils/validations.ts
hasalarootcode: Not needed

frontend/src/community/leave/components/organisms/LeavePolicyWizard/SummaryStep.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/organisms/LeavePolicyWizard/EntitlementSetupStep.tsx
hasalarootcode: Not needed

frontend/src/community/leave/utils/leavePolicy/leavePolicyUtils.ts
hasalarootcode: Not needed

frontend/src/community/leave/types/LeavePolicyTypes.ts
hasalarootcode: Already there, https://github.com/SkappHQ/skapp/pull/1766

frontend/src/community/leave/utils/validations.ts
hasalarootcode: Not needed

frontend/src/community/leave/utils/validations.ts
hasalarootcode: Restored

frontend/src/community/leave/components/organisms/LeavePolicyWizard/EntitlementSetupStep.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/organisms/LeavePolicyWizard/EntitlementSetupStep.tsx
hasalarootcode: Not needed

frontend/src/community/common/assets/languages/english/leaveModule.json
hasalarootcode: Not needed

frontend/src/community/leave/types/LeavePolicyTypes.ts
hasalarootcode: Not needed

frontend/src/community/leave/types/LeavePolicyTypes.ts
hasalarootcode: Already there, https://github.com/SkappHQ/skapp/pull/1766

frontend/src/community/leave/components/organisms/LeavePolicyWizard/EntitlementSetupStep.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/organisms/LeavePolicyWizard/EntitlementSetupStep.tsx
hasalarootcode: Not needed

frontend/src/community/leave/constants/leavePolicyConstants.ts
hasalarootcode: Not needed

frontend/src/community/leave/constants/leavePolicyConstants.ts
hasalarootcode: Not needed

frontend/src/community/leave/components/organisms/LeavePolicyWizard/EntitlementSetupStep.tsx
hasalarootcode: Not needed

frontend/src/community/leave/constants/leavePolicyConstants.ts
```diff
@@ -75,12 +71,11 @@ export const accrualFrequencyItemList = [
...
+
+export const CARRYOVER_EXPIRY_REFERENCE_YEAR = 2025;
```
thusala, line 80: is this hard code 2025 is okay? currently we are in 2026, let's say in 2030 it shows 2025. feels a bit outdated?

frontend/src/community/leave/constants/leavePolicyConstants.ts
```diff
@@ -75,12 +71,11 @@ export const accrualFrequencyItemList = [
...
 
-export const carryoverDateItemList = [
-  { id: "january-1", labelKey: "january1", value: "01-01" },
-  { id: "april-1", labelKey: "april1", value: "04-01" },
-  { id: "july-1", labelKey: "july1", value: "07-01" },
-  { id: "october-1", labelKey: "october1", value: "10-01" }
-];
+export const CARRYOVER_EXPIRY_DATE_FORMAT = "MM-dd";
+
+export const CARRYOVER_EXPIRY_DISPLAY_FORMAT = "dd MMMM";
```
thusala: dont we have already defined type for this?

thusala (changes requested):

frontend/src/community/leave/constants/leavePolicyConstants.ts
hasalarootcode: Yes, this is just needed to exclude 29th Feb from the date picker

frontend/src/community/leave/constants/leavePolicyConstants.ts
hasalarootcode: Fixed

shakila-kamalasena (dismissed):

thusala (dismissed):

akilarootcode (dismissed):

shakila-kamalasena (approved):

PR #1766, opened 2026-08-17
Feat/leave policy carry over implementation - BE

hasalarootcode:
- Renamed `carryover_date` to `carryover_expiry_date` in LeavePolicy model and related DTOs.
- Updated validation messages and constants to reflect the new carryover expiry date terminology.
- Introduced new utility methods for handling carryover expiry logic in PolicyLeaveAccrualUtil.
- Added tests for carryover functionality to ensure correct behavior with the new expiry date logic.
- Updated database schema to rename the column in the leave policy table.

### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR renames the leave-policy carryover date column to `carryover_expiry_date`, makes it optional, and adds real carryover-into-cycle accounting anchored to the org's configured leave cycle. The accrual/carryover math is well factored, but there are three blocking issues: the balance card still resolves the cycle from the calendar year (so orgs with a non-January leave cycle get a different cycle on the balance card than on leave application), the request/response field rename silently breaks the existing frontend contract (the UI still sends/reads `carryoverDate`), and the Liquibase changeset was edited in place rather than adding a rename migration. There are also correctness and performance concerns in the carryover lookback loop, and no unit tests cover the new math.

Found **22** new issue(s): 🔴 7 important, 🟡 10 suggestion(s), 🟣 5 nit(s)

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
```diff
@@ -706,15 +713,39 @@ private EmployeeLeavePolicy resolveActiveAssignment(Employee employee, Long poli
 			.orElseThrow(() -> new ModuleException(LeaveMessageConstant.LEAVE_ERROR_POLICY_LEAVE_POLICY_NOT_ASSIGNED));
 	}
 
-	private PolicyLeaveBalanceDto calculateBalanceForYear(EmployeeLeavePolicy assignment, int year) {
-		return calculateBalance(assignment, PolicyLeaveAccrualUtil.resolveCycle(year));
+	private PolicyLeaveBalanceDto calculateBalanceForYear(EmployeeLeavePolicy assignment, int year, LocalDate today,
+			MonthDay cycleAnchor) {
+		PolicyLeaveDateWindowDto cycle = PolicyLeaveAccrualUtil.resolveCycle(year, cycleAnchor);
```
ThinuwanW, line 722: 🤖 **Claude** · 🔴 **Important**

The cycle is now anchored to the org leave-cycle start (`cycleAnchor`), but the year fed into `resolveCycle` still comes from `resolveYear(year)` = `DateTimeUtils.getCurrentUtcDate().getYear()` (calendar year). For any org whose leave cycle does not start on Jan 1, these disagree. Example: anchor = Apr 1, today = 2026-02-10. `getCurrentUserPolicyBalances(null)` builds cycle 2026-04-01 .. 2027-03-31 — a cycle that has not started yet — and `clampToCycle` then pins `asOf` to 2026-04-01. Meanwhile `calculateBalanceForDate` (used by `checkPolicyLeaveAvailability` and `applyPolicyLeaveRequest`) correctly resolves 2025-04-01 .. 2026-03-31 for the same day. The employee therefore sees a balance card for the wrong (future) cycle, with `validFrom`/`validTo` that do not match the window leave is actually validated against. Fix: derive the default from the containing cycle, e.g. `int resolvedYear = year != null ? year : PolicyLeaveAccrualUtil.resolveCycleContaining(today, cycleAnchor).getStartDate().getYear();` and document that the `year` query param now means "the cycle that starts in that year". Related: `getCurrentUserPolicyLeaveRequests` (line 257) still filters with `DateTimeUtils.getStartOfYear/getEndOfYear`, so the request list window is now inconsistent with the anchored balance cycle for the same `year` value — worth aligning in the same change.

backend/src/main/java/com/skapp/community/leaveplanner/payload/request/LeavePolicyAccrualDetailDto.java
```diff
@@ -20,7 +20,7 @@ public class LeavePolicyAccrualDetailDto {
 
 	private Boolean isCarryoverEnabled;
 
-	private String carryoverDate;
+	private String carryoverExpiryDate;
```
ThinuwanW, line 23: 🤖 **Claude** · 🔴 **Important**

Renaming this request field (and the matching field on `LeavePolicyResponseDto`) is a breaking API contract change, and the frontend in this repo has not been updated: `frontend/src/community/leave/utils/leavePolicy/leavePolicyUtils.ts:87` still builds the payload with `carryoverDate`, and `frontend/src/community/leave/types/LeavePolicyTypes.ts:53,124,159` still declare/read `carryoverDate`. Because Spring Boot leaves `FAIL_ON_UNKNOWN_PROPERTIES` disabled, the incoming `carryoverDate` will be silently dropped — every policy created/updated from the current UI will save a null expiry, which now means "carryover never expires" (see `PolicyLeaveAccrualUtil.neverExpires`). That is silent data loss with a direct effect on employee balances. Either land the frontend rename in the same PR, or accept both names on the request DTO for one release (e.g. keep a deprecated `carryoverDate` setter annotated with `@JsonAlias("carryoverDate")` on the new field) and keep emitting both on the response until the UI is migrated.

backend/src/main/resources/community/db/changelog/ddl-script/leave-ddl-script-v1-create-leave-policy-tables.sql
```diff
@@ -22,24 +22,24 @@ CREATE TABLE IF NOT EXISTS `lv_leave_type`
...
+    `name`                  text        NOT NULL,
+    `leave_type_id`         bigint      NOT NULL,
+    `policy_type`           text        NOT NULL,
+    `status`                text        NOT NULL,
+    `accrual_days`          float,
+    `frequency`             text,
+    `waiting_period_days`   int,
+    `accrual_cap_days`      float,
+    `is_carryover_enabled`  boolean     NOT NULL,
+    `carryover_expiry_date` text,
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

This changeset (`leave-ddl-script-v1-create-leave-policy-tables`) is already included in `db.changelog.yml` and has therefore already run on every environment that is past that point. Editing it in place has two consequences: (1) Liquibase recomputes the checksum and will fail startup with a ValidationFailedException on any database where the changeset is recorded as EXECUTED (there is no `runOnChange` / `validCheckSum` on it); (2) because the statement is `CREATE TABLE IF NOT EXISTS`, existing databases keep the old `carryover_date` column, so Hibernate will fail resolving `LeavePolicy.carryoverExpiryDate`. Fresh databases work, which is why this may pass locally. Fix: revert this file to its original content and add a new changeset file (e.g. `leave-ddl-script-v1-alter-lv-leave-policy-rename-carryover-date.sql`) containing `ALTER TABLE lv_leave_policy CHANGE COLUMN carryover_date carryover_expiry_date text;` with a rollback, and register it in `db.changelog.yml`. The column-alignment reformatting of the untouched rows also inflates the diff — keep it out of the changeset that already shipped.

backend/src/main/java/com/skapp/community/leaveplanner/util/PolicyLeaveAccrualUtil.java
```diff
@@ -96,6 +99,81 @@ public static float applyAccrualCap(LeavePolicy policy, float totalDaysAllocated
...
+
+	public static float carriedOverInto(LeavePolicy policy, LocalDate effectiveFrom, PolicyLeaveDateWindowDto cycle,
+			MonthDay cycleAnchor, PolicyLeaveUsageLookup usageLookup) {
+		if (!Boolean.TRUE.equals(policy.getIsCarryoverEnabled())) {
+			return 0f;
+		}
+
+		int targetYear = cycle.getStartDate().getYear();
+		int firstYear = resolveCycleContaining(effectiveFrom, cycleAnchor).getStartDate().getYear();
+		int lookbackCycles = neverExpires(policy) ? PolicyLeaveConstant.MAX_CARRYOVER_LOOKBACK_CYCLES : 1;
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

For policies with an expiry date the lookback is hard-coded to 1 cycle, but the recursion needs at least 2 to compute the prior cycle correctly. `unusedAtCycleEnd` for the single prior cycle is evaluated with `carriedOverDays = 0`, while `usageLookup.usedBetween(...)` for that cycle still includes leave that was actually consumed from the (ignored) carryover — so unused days at that cycle end are understated and the employee silently loses balance. Concrete case: accrual 12/yr, maxCarryover 5, expiry 03-31; cycle 2024 opened with 5 carried over, accrued 12, used 15 (5 of them before Mar 31). True unused at 2024 year end = 12 + 5 - 15 = 2, so 2 should carry into 2025; this code computes max(0, 12 + 0 - 15) = 0. Fix: use at least 2 lookback cycles for expiring policies (or simply use `MAX_CARRYOVER_LOOKBACK_CYCLES` for both branches and let the expiry logic zero out the older carryover naturally).

backend/src/main/java/com/skapp/community/leaveplanner/util/PolicyLeaveAccrualUtil.java
```diff
@@ -96,6 +99,81 @@ public static float applyAccrualCap(LeavePolicy policy, float totalDaysAllocated
...
+		int fromYear = Math.max(firstYear, targetYear - lookbackCycles);
+
+		float carriedOverDays = 0f;
+		for (int year = fromYear; year < targetYear; year++) {
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

This loop turns balance calculation into an N+1 query pattern. Each iteration calls `unusedAtCycleEnd`, which issues one `sumCommittedDaysForPolicyInCycle` query for the cycle total and potentially a second one inside `usableCarryoverDays` — up to ~2 queries per prior cycle, i.e. ~12 queries per assignment for a never-expiring policy (5 lookback cycles + the target cycle). `getCurrentUserPolicyBalances` runs this for every active assignment, so an employee with 6 policies triggers ~70 round trips on a single page load. Consider fetching all committed days for the employee/policy across the full lookback window in one query and letting `PolicyLeaveUsageLookup` resolve windows in memory, or at minimum memoising `usedBetween` per (from,to) inside the lambda in `PolicyLeaveServiceImpl.calculateBalance`. Note `accruedUpTo` is also re-walked period-by-period for each cycle, which for DAILY-frequency policies with a multi-year `effectiveFrom` is thousands of iterations per assignment.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
```diff
@@ -706,15 +713,39 @@ private EmployeeLeavePolicy resolveActiveAssignment(Employee employee, Long poli
...
+		return calculateBalance(assignment, PolicyLeaveAccrualUtil.resolveCycleContaining(date, cycleAnchor),
+				cycleAnchor, date);
+	}
+
+	private MonthDay resolveCycleAnchor() {
+		try {
+			LeaveCycleDetailsDto leaveCycle = leaveCycleService.getLeaveCycleConfigs();
+			return MonthDay.of(leaveCycle.getStartMonth(), leaveCycle.getStartDate());
+		}
+		catch (Exception e) {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Catching bare `Exception` here violates the exception-handling guideline and, more importantly, makes a silent-wrong-answer path: any failure (missing LEAVE_CYCLE org config, malformed JSON, a transient DB error) falls back to the calendar year, which silently changes every employee's cycle window and therefore their balances and leave-period validation — with only a WARN line to show for it. `getLeaveCycleConfigs` NPEs when the config row is absent (`leaveCycleJson.get(...)` on a NullNode), and `MonthDay.of` throws `DateTimeException` on an invalid month/day, so catch those two specifically (`NullPointerException`/`DateTimeException`) and let anything else propagate to `GlobalExceptionHandler`. Also log with the exception rather than `e.getMessage()` so the stack trace is retained: `log.warn("resolveCycleAnchor: falling back to the calendar year", e)`.

backend/src/main/java/com/skapp/community/leaveplanner/constant/PolicyLeaveConstant.java
```diff
@@ -30,4 +30,6 @@ public class PolicyLeaveConstant {
 
 	public static final int MAX_SEARCH_KEYWORD_LENGTH = 100;
 
+	public static final int MAX_CARRYOVER_LOOKBACK_CYCLES = 5;
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`MAX_CARRYOVER_LOOKBACK_CYCLES = 5` silently truncates the carryover chain for never-expiring policies: an employee assigned to such a policy for 8 years gets carryover computed from only the last 5 cycles, understating their balance with no indication anywhere in the response or the logs. Add a Javadoc comment explaining the bound and its trade-off, and consider logging (or exposing) when the assignment's `effectiveFrom` predates the lookback window so the truncation is at least observable. A cheaper alternative is to cap the accumulated carryover only by `maxCarryoverDays` and note that when `maxCarryoverDays` is set, iterating more cycles converges anyway.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
```diff
@@ -739,6 +774,7 @@ private PolicyLeaveBalanceDto calculateBalance(EmployeeLeavePolicy assignment, P
 		balance.setCycleEnd(cycle.getEndDate());
 		balance.setUsableFrom(resolveUsableFrom(effectiveFrom, cycle));
 		balance.setAccruedDays(accruedDays);
+		balance.setCarriedOverDays(carriedOverDays);
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`carriedOverDays` is computed and stored on `PolicyLeaveBalanceDto` but never surfaced: `toBalanceCard` does not copy it, and `EmployeePolicyBalanceResponseDto` has no such field, so no client can see it. If the intent is to show "X days carried over" on the balance card, add the field to the response DTO and set it in `toBalanceCard`. Also note the value stored here is the pre-expiry carryover, while `totalDaysAllocated` uses the post-expiry `usableCarryoverDays` — if it is exposed as-is, `accruedDays + carriedOverDays` will not reconcile with `totalDaysAllocated` after the expiry date. Either store the usable figure or expose both explicitly.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
```diff
@@ -747,10 +783,22 @@ private PolicyLeaveBalanceDto calculateBalance(EmployeeLeavePolicy assignment, P
 		return balance;
 	}
 
+	private float accruedInCycle(EmployeeLeavePolicy assignment, PolicyLeaveDateWindowDto cycle) {
```
ThinuwanW, line 783: 🤖 **Claude** · 🟡 **Suggestion**

`accruedInCycle` here plus the `applyAccrualCap` call at line 762 is exactly `PolicyLeaveAccrualUtil.accrualAllocationInCycle(policy, effectiveFrom, cycle)`, which this PR adds and which the carryover loop already uses. Keeping two copies of the same formula risks them drifting (e.g. if rounding or cap ordering changes, the target-cycle allocation and the prior-cycle allocation would silently diverge). Drop `accruedInCycle` and use the util for the allocation, keeping only the un-capped `roundToHalfDay(accruedWithinCycle(...))` call needed for the `accruedDays` display field.

backend/src/main/java/com/skapp/community/leaveplanner/util/PolicyLeaveAccrualUtil.java
```diff
@@ -96,6 +99,81 @@ public static float applyAccrualCap(LeavePolicy policy, float totalDaysAllocated
...
+			LocalDate asOf, PolicyLeaveUsageLookup usageLookup) {
+		if (carriedOverDays <= 0f) {
+			return 0f;
+		}
+
+		LocalDate expiresOn = resolveCarryoverExpiry(policy, cycle);
+		if (expiresOn == null || !asOf.isAfter(expiresOn)) {
+			return carriedOverDays;
+		}
+		return Math.min(carriedOverDays, usageLookup.usedBetween(cycle.getStartDate(), expiresOn));
```
ThinuwanW, line 147: 🤖 **Claude** · 🟣 **Nit**

`usedBetween(cycle.getStartDate(), expiresOn)` resolves to `sumCommittedDaysForPolicyInCycle`, whose Criteria predicate is `between(startDate, from, to)` — it filters on the request's start date only. A leave request that starts one day before the expiry date and runs for two weeks past it therefore counts its full duration as "used before expiry", inflating the retained carryover. Consider using a true overlap predicate (or pro-rating by working days inside the window) for this expiry check, since unlike the whole-cycle sum this window can land in the middle of a request.

backend/src/main/java/com/skapp/community/leaveplanner/util/LeavePolicyValidationUtil.java
```diff
@@ -70,11 +70,10 @@ private static void validateAccrualSetup(LeavePolicyAccrualDetailDto accrual) {
 	}
 
 	private static void validateCarryoverSetup(LeavePolicyAccrualDetailDto accrual) {
-		if (accrual.getCarryoverDate() == null || accrual.getCarryoverDate().isBlank()) {
-			throw new ModuleException(LeaveMessageConstant.LEAVE_ERROR_LEAVE_POLICY_CARRYOVER_DATE_REQUIRED);
-		}
-		if (!DateTimeUtils.isValidMonthDay(accrual.getCarryoverDate())) {
-			throw new ModuleException(LeaveMessageConstant.LEAVE_ERROR_LEAVE_POLICY_CARRYOVER_DATE_INVALID);
+		String carryoverExpiryDate = accrual.getCarryoverExpiryDate();
```
ThinuwanW, line 73: 🤖 **Claude** · 🟣 **Nit**

Making the expiry date optional changes the meaning of an absent value from "rejected" to "carryover never expires" (`neverExpires` returns true for null/blank), and combined with a null `maxCarryoverDays` (`capCarryover` returns uncapped) an admin can now create a policy with unbounded, never-expiring carryover by simply omitting two optional fields. That may well be intended, but it is a significant policy semantic that is now implicit. Please confirm it is deliberate and document it (Javadoc on `neverExpires`/`capCarryover`, and ideally an explicit "never expires" option in the UI) rather than leaving it as an emergent property of two null checks.

backend/src/main/java/com/skapp/community/leaveplanner/util/PolicyLeaveAccrualUtil.java
```diff
@@ -25,13 +26,15 @@ public static LocalDate resolveAccrualStartDate(LeavePolicy policy, LocalDate ef
 		return effectiveFrom.plusDays(waitingPeriodDays);
 	}
 
-	public static PolicyLeaveDateWindowDto resolveCycle(int year) {
-		MonthDay anchor = DateTimeUtils.CALENDAR_YEAR_START;
+	public static PolicyLeaveDateWindowDto resolveCycle(int year, MonthDay cycleStart) {
+		MonthDay anchor = cycleStart == null ? DateTimeUtils.CALENDAR_YEAR_START : cycleStart;
```
ThinuwanW, line 29: 🤖 **Claude** · 🟣 **Nit**

If the org configures a leave cycle starting Feb 29, `anchor.atYear(year)` resolves to Feb 28 in non-leap years, so consecutive cycles produced by `resolveCycle` can overlap or leave a one-day gap (e.g. 2024-02-29..2025-02-27 followed by 2025-02-28..2026-02-27), and `resolveCycleContaining` can then place a date in no cycle boundary you'd expect. Worth either rejecting Feb 29 as a cycle anchor upstream or normalising the anchor to Feb 28, with a short comment here recording the decision.

backend/src/main/java/com/skapp/community/leaveplanner/util/PolicyLeaveAccrualUtil.java
```diff
@@ -1,6 +1,7 @@
 package com.skapp.community.leaveplanner.util;
 
 import com.skapp.community.common.util.DateTimeUtils;
+import com.skapp.community.leaveplanner.constant.PolicyLeaveConstant;
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

No tests accompany this change. The new carryover math (`carriedOverInto`, `usableCarryoverDays`, `unusedAtCycleEnd`, `capCarryover`, `resolveCarryoverExpiry`) and the anchored `resolveCycle`/`resolveCycleContaining` are pure functions with a `PolicyLeaveUsageLookup` seam that makes them trivial to unit test, yet `backend/src/test/.../leaveplanner` currently contains only controller integration tests and none of them seed carryover data. Given the FIFO/expiry/cap interactions and the multi-cycle recursion, please add unit tests covering at minimum: non-January cycle anchor; carryover capped by `maxCarryoverDays`; carryover partially consumed before the expiry date; carryover fully expired; never-expiring carryover accumulating across cycles; and an assignment whose `effectiveFrom` falls mid-cycle.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
```diff
@@ -726,10 +757,14 @@ private PolicyLeaveBalanceDto calculateBalance(EmployeeLeavePolicy assignment, P
...
-				PolicyLeaveAccrualUtil.accruedWithinCycle(policy, accrualStartDate, cycle, cycle.getEndDate()));
-		float totalDaysAllocated = PolicyLeaveAccrualUtil.applyAccrualCap(policy, accruedDays);
+		PolicyLeaveUsageLookup usageLookup = (from, to) -> usedInWindow(assignment, from, to);
+		float accruedDays = accruedInCycle(assignment, cycle);
+		float accrualAllocation = PolicyLeaveAccrualUtil.applyAccrualCap(policy, accruedDays);
+		float carriedOverDays = PolicyLeaveAccrualUtil.carriedOverInto(policy, effectiveFrom, cycle, cycleAnchor,
+				usageLookup);
+		float usableCarryoverDays = PolicyLeaveAccrualUtil.usableCarryoverDays(policy, cycle, carriedOverDays, asOf,
+				usageLookup);
+		float totalDaysAllocated = accrualAllocation + usableCarryoverDays;
```
ThinuwanW, line 764: 🤖 **Claude** · 🔴 **Important**

The accrual cap is now applied only to the accrued component and carryover is added afterwards, so `totalDaysAllocated` can exceed the policy's configured cap. Before this change `totalDaysAllocated` was exactly `applyAccrualCap(policy, accruedDays)`, so the cap was a hard ceiling on the cycle allocation. Now a policy with accrualCapDays = 20 that accrues 24 and carries 5 forward yields 25 allocated days. The message key backing that field reads "Maximum accrual balance must be at least 1", which suggests it is meant to be a balance ceiling, not a per-cycle accrual ceiling. Worse, the same shape exists in `PolicyLeaveAccrualUtil.unusedAtCycleEnd` (accrualAllocation + usableCarryover - used), so the uncapped surplus rolls into the next cycle and compounds across the lookback chain. Decide which semantics the cap has and make it explicit: either apply `applyAccrualCap` to the combined `accrualAllocation + usableCarryoverDays` (and to the leftover inside `unusedAtCycleEnd`), or rename the field/message so it is clearly an accrual-only cap. Either way add a test pinning the chosen behaviour.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
```diff
@@ -134,13 +140,14 @@ public ResponseEntityDto getCurrentUserPolicyBalances(Integer year) {
 		int resolvedYear = resolveYear(year);
 		boolean hasSupervisor = !peopleService.getCurrentEmployeeManagers().isEmpty();
 		LocalDate today = DateTimeUtils.getCurrentUtcDate();
+		MonthDay cycleAnchor = resolveCycleAnchor();
```
ThinuwanW, line 144: 🤖 **Claude** · 🔴 **Important**

Missing paired change: balances are now anchored to the org leave cycle, but `getCurrentUserPolicyLeaveRequests` still bounds its query with `DateTimeUtils.getStartOfYear(resolvedYear)` / `getEndOfYear(resolvedYear)`, i.e. Jan 1 – Dec 31. With a non-calendar anchor (say April 1) the balance card for year 2026 covers 2026-04-01..2027-03-31 while the request feed for the same year returns 2026-01-01..2026-12-31. The controller documents that feed as what the apply-leave calendar uses to show requests already raised for a year, so the calendar and the balance card will disagree about which requests consumed the shown allocation. `toBalanceCard` also labels the card `year = <cycle start year>` while `validFrom`/`validTo` straddle two calendar years, which the frontend year dropdown is not built for. Resolve the cycle window once (via the same anchor) and pass its start/end into `findMyRequests` instead of the calendar-year helpers.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
```diff
@@ -706,15 +713,39 @@ private EmployeeLeavePolicy resolveActiveAssignment(Employee employee, Long poli
...
-		return calculateBalance(assignment, PolicyLeaveAccrualUtil.resolveCycle(year));
+	private PolicyLeaveBalanceDto calculateBalanceForYear(EmployeeLeavePolicy assignment, int year, LocalDate today,
+			MonthDay cycleAnchor) {
+		PolicyLeaveDateWindowDto cycle = PolicyLeaveAccrualUtil.resolveCycle(year, cycleAnchor);
+		return calculateBalance(assignment, cycle, cycleAnchor, clampToCycle(today, cycle));
 	}
 
 	private PolicyLeaveBalanceDto calculateBalanceForDate(EmployeeLeavePolicy assignment, LocalDate date) {
-		return calculateBalance(assignment, PolicyLeaveAccrualUtil.resolveCycleContaining(date));
+		MonthDay cycleAnchor = resolveCycleAnchor();
```
ThinuwanW, line 727: 🤖 **Claude** · 🟡 **Suggestion**

`resolveCycleAnchor` duplicates logic `LeaveCycleService` already owns — `getLeaveCycleStartDate()` reads the same LEAVE_CYCLE org config and already has a fallback to Jan 1 — creating a second source of truth for the cycle anchor (and silently ignoring the configured cycle END, which `getLeaveCycleEndDate()` exposes). It is also called per invocation from `calculateBalanceForDate`, so every availability check and every apply-leave call performs an extra org-config row read plus a full JSON parse; note `LeaveCycleServiceImpl.getLeaveCycleConfigs()` is not annotated `@Transactional(readOnly = true)`. Prefer reusing/extending `LeaveCycleService` with a `getLeaveCycleAnchor()`-style method that returns a `MonthDay`, mark it `@Transactional(readOnly = true)`, and resolve the anchor once per service call the way `getCurrentUserPolicyBalances` already does.

backend/src/main/java/com/skapp/community/leaveplanner/util/PolicyLeaveAccrualUtil.java
```diff
@@ -96,6 +99,81 @@ public static float applyAccrualCap(LeavePolicy policy, float totalDaysAllocated
 		return Math.min(totalDaysAllocated, cap);
 	}
 
+	public static float capCarryover(LeavePolicy policy, float unusedDays) {
+		float carriedOverDays = Math.max(0f, unusedDays);
+		Float maxCarryoverDays = policy.getMaxCarryoverDays();
```
ThinuwanW, line 103: 🤖 **Claude** · 🟡 **Suggestion**

`capCarryover` returns the full unused amount whenever `maxCarryoverDays` is null, and `LeavePolicyValidationUtil.validateCarryoverSetup` now requires nothing at all when carryover is enabled (both the expiry date and the max are optional). So an admin who ticks "carryover enabled" and leaves both fields blank gets uncapped, never-expiring carryover whose only bound is the lookback constant. Previously carryover had no effect on the balance so this was harmless; it now directly inflates `totalDaysAllocated`. Either require `maxCarryoverDays` when carryover is enabled (reusing `LEAVE_ERROR_LEAVE_POLICY_MAX_CARRYOVER_DAYS_INVALID` or a new required-message constant), or apply a documented default ceiling in `capCarryover` rather than treating null as unlimited.

backend/src/main/java/com/skapp/community/leaveplanner/util/PolicyLeaveAccrualUtil.java
```diff
@@ -96,6 +99,81 @@ public static float applyAccrualCap(LeavePolicy policy, float totalDaysAllocated
...
+		return maxCarryoverDays == null ? carriedOverDays : Math.min(carriedOverDays, maxCarryoverDays);
+	}
+
+	public static LocalDate resolveCarryoverExpiry(LeavePolicy policy, PolicyLeaveDateWindowDto cycle) {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Over-broad visibility: `capCarryover`, `resolveCarryoverExpiry`, `neverExpires` and `accrualAllocationInCycle` are all declared `public static` but grep shows every call site is inside `PolicyLeaveAccrualUtil` itself (only `carriedOverInto` and `usableCarryoverDays` are called from `PolicyLeaveServiceImpl`). Make the four internal helpers `private static` so the carryover math has a single, small public surface — otherwise they read as supported API and will attract callers that bypass the ordering constraints between them (e.g. `capCarryover` must be applied to the result of `unusedAtCycleEnd`, not to a raw balance).

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
ThinuwanW, line 722: 🤖 **Claude** · 🟡 **Suggestion**

The `year` request param on `/v1/leave/policy-leave/balances` is an unvalidated `Integer` that flows straight into `resolveCycle(year, cycleAnchor)`, which evaluates `anchor.atYear(year + 1)`. A caller passing 999999999 (or `Integer.MAX_VALUE`, which overflows to `Integer.MIN_VALUE`) triggers an uncaught `DateTimeException` and a 500 rather than a `ModuleException`/`ErrorResponse`. Add a bounds check in `resolveYear` (e.g. reject anything outside current year ± a small window) and throw a `ModuleException` with a new message constant added to `LeaveMessageConstant` and `leave-messages.properties`.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`carriedOverDays` is set to the *raw* carryover from `carriedOverInto`, but the value that actually feeds `totalDaysAllocated` is `usableCarryoverDays` (post-expiry reduction). Once this field is surfaced on the response DTO, `accruedDays + carriedOverDays` will not reconcile to `totalDaysAllocated` for any policy whose carryover has expired — a support-ticket generator. Either store `usableCarryoverDays` here, or carry both values (e.g. `carriedOverDays` and `usableCarryoverDays`) so the breakdown is self-consistent and the UI can explain the expired portion.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
```diff
@@ -747,10 +783,22 @@ private PolicyLeaveBalanceDto calculateBalance(EmployeeLeavePolicy assignment, P
...
+				assignment.getEffectiveFrom());
+		return PolicyLeaveAccrualUtil.roundToHalfDay(
+				PolicyLeaveAccrualUtil.accruedWithinCycle(policy, accrualStartDate, cycle, cycle.getEndDate()));
+	}
+
 	private float totalDaysUsedInCycle(EmployeeLeavePolicy assignment, PolicyLeaveDateWindowDto cycle) {
+		return usedInWindow(assignment, cycle.getStartDate(), cycle.getEndDate());
+	}
+
+	private float usedInWindow(EmployeeLeavePolicy assignment, LocalDate from, LocalDate to) {
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`usedInWindow` now calls `sumCommittedDaysForPolicyInCycle` with arbitrary sub-cycle windows (carryover expiry ranges, prior-cycle ranges), but the repository method and its `cycleStart`/`cycleEnd` parameters still say "cycle". That naming is what makes the boundary semantics easy to misread at the new call sites. Rename the custom repository method to something like `sumCommittedDaysForPolicyInWindow` with `windowStart`/`windowEnd` parameters, in both `PolicyLeaveRequestRepository` and `PolicyLeaveRequestRepositoryImpl`, and document that the predicate matches on the request start date only.

backend/src/main/java/com/skapp/community/leaveplanner/constant/PolicyLeaveConstant.java
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Placement: `PolicyLeaveConstant` holds request-side constants (attachment limits, search keyword length, status transition maps), while every policy-configuration bound the accrual/validation code uses — MIN_DAYS, MAX_DAYS, MIN_ACCRUAL_CAP_DAYS, MIN_WAITING_PERIOD_DAYS — lives in `LeavePolicyConstant`. `MAX_CARRYOVER_LOOKBACK_CYCLES` is a policy-accrual tuning value consumed only by `PolicyLeaveAccrualUtil`, so it belongs in `LeavePolicyConstant` alongside the others (and drops the new cross-package import this adds to the util). If it is meant to be operator-tunable rather than a compile-time constant, move it to `application.yml` instead.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
hasalarootcode: Fixed. `resolveYear` is now `resolveCycleYear` and derives the default from the cycle containing today, so an omitted year no longer resolves to a future cycle on a non-January anchor. Documented on the endpoint that the year identifies the cycle starting in that year.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
hasalarootcode: Fixed. `getCurrentUserPolicyLeaveRequests` resolves the cycle through the same anchor and passes its start/end into `findMyRequests` instead of `getStartOfYear`/`getEndOfYear`.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
hasalarootcode: Fixed. `resolveCycleYear` rejects anything more than `PolicyLeaveConstant.MAX_CYCLE_YEAR_OFFSET` (10) either side of the current year with `LEAVE_ERROR_POLICY_LEAVE_INVALID_YEAR`. The comparison is done in long arithmetic - `Math.abs(Integer.MIN_VALUE)` is itself negative, so an int subtraction would have passed the bound.

backend/src/main/resources/community/db/changelog/ddl-script/leave-ddl-script-v1-create-leave-policy-tables.sql
hasalarootcode: Fixed. Reverted the create-table changeset to what is on develop and added `leave-ddl-script-v1-alter-leave-policy-rename-carryover-date.sql` with a rollback, registered at the end of `db.changelog.yml`. The reformatting went with the revert.

backend/src/main/java/com/skapp/community/leaveplanner/util/PolicyLeaveAccrualUtil.java
hasalarootcode: Fixed. Dropped the ternary, both branches now use `MAX_CARRYOVER_LOOKBACK_CYCLES`. The expiry reduction inside `usableCarryoverDays` zeroes the stale carryover per cycle anyway, so the extra cycles are safe. This does make the query count worse for expiring policies - noted on the N+1 thread.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
hasalarootcode: Fixed. Narrowed to `NullPointerException | DateTimeException` and the exception is passed to `log.warn` rather than `e.getMessage()`. A Jackson parse failure now propagates to `GlobalExceptionHandler` instead of silently falling back.

backend/src/main/java/com/skapp/community/leaveplanner/util/PolicyLeaveAccrualUtil.java
hasalarootcode: Fixed. `capCarryover`, `resolveCarryoverExpiry` and `neverExpires` are private now. `accrualAllocationInCycle` stays public - it is the call the service uses after removing the duplicated formula on the other thread.

backend/src/main/java/com/skapp/community/leaveplanner/constant/PolicyLeaveConstant.java
hasalarootcode: Fixed. Javadoc on the constant covers the bound, what gets dropped for an older assignment, and why the understatement is limited - `maxCarryoverDays` is required now, so the chain converges on that ceiling. Left out the logging; it would fire per assignment per page load for every long-tenured employee.

hasalarootcode: Moved to `LeavePolicyConstant`, next to the other policy day bounds. Drops the cross-package import from the util.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
hasalarootcode: Fixed. `accrualAllocation` comes from `PolicyLeaveAccrualUtil.accrualAllocationInCycle` now. `accruedInCycle` stays for the un-capped `accruedDays` display field only.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
hasalarootcode: Fixed. Renamed to `sumCommittedDaysForPolicyInWindow` with `windowStart`/`windowEnd` in both the interface and the impl, and the Javadoc records that the predicate matches on the request start date only.

backend/src/main/java/com/skapp/community/leaveplanner/util/PolicyLeaveAccrualUtil.java
hasalarootcode: Fixed. `validateCarryoverSetup` requires `maxCarryoverDays` when carryover is enabled, with a new `LEAVE_ERROR_LEAVE_POLICY_MAX_CARRYOVER_DAYS_REQUIRED` key. That matches the FE, which already has it as `.required()`. So carryover can be never-expiring but never uncapped.

backend/src/main/java/com/skapp/community/leaveplanner/util/LeavePolicyValidationUtil.java
hasalarootcode: The optional expiry date is deliberate - it is the Never expires option on the wizard. Documented it on `neverExpires`, and the unbounded half is gone: `maxCarryoverDays` is required whenever carryover is enabled now.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
hasalarootcode: Fixed. Stores `usableCarryoverDays`. Note the breakdown still will not add up exactly when the accrual cap bites, since `totalDaysAllocated` uses the capped allocation while `accruedDays` is un-capped - that is the cap-semantics question on the other thread.

backend/src/main/java/com/skapp/community/leaveplanner/payload/request/LeavePolicyAccrualDetailDto.java
hasalarootcode: The matching FE rename is on #1765 - `carryoverExpiryDate` in the payload, the types and the form data. Both PRs need to land together, which is why this branch alone looks like a contract break. No alias needed if they go in as a pair.

backend/src/main/java/com/skapp/community/leaveplanner/util/PolicyLeaveAccrualUtil.java
hasalarootcode: The clamping applies to both boundaries, so the cycles stay contiguous - your own example shows it: `2024-02-29..2025-02-27` is followed by `2025-02-28..2026-02-27`, no gap and no overlap. `resolveCycle(y)` is always `[A(y), A(y+1)-1]`, and since `A` stays inside calendar year `y`, `resolveCycleContaining` picks exactly one cycle for any date. The only real effect is that a Feb 29 anchor behaves as Feb 28 in non-leap years, which is more forgiving than the existing leave cycle code where `LocalDate.of` would throw. Happy to add a comment recording that if you want it stated.

backend/src/main/java/com/skapp/community/leaveplanner/util/PolicyLeaveAccrualUtil.java
hasalarootcode: Real, and slightly worse after the lookback fix on the other thread - expiring policies now walk 5 cycles too. Memoising `usedBetween` buys little since each cycle asks for a different window. The fix that works is one query for the whole lookback span per assignment and letting `PolicyLeaveUsageLookup` slice it in memory, which means a new repository method returning the rows rather than a sum. Want that here, or as a follow-up before the screens go live?

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
hasalarootcode: Agree it is a second source of truth. `getLeaveCycleStartDate` returns a `LocalDate` pinned to the current year though, so it cannot be used as-is for an arbitrary cycle year - this needs a `getLeaveCycleAnchor()` returning `MonthDay` added to `LeaveCycleService`, plus `@Transactional(readOnly = true)` on `getLeaveCycleConfigs`, which is shared with the entitlement and analytics services. Shall I add it in this PR?

backend/src/main/java/com/skapp/community/leaveplanner/util/PolicyLeaveAccrualUtil.java
hasalarootcode: Correct, the start-date-only predicate over-counts a request straddling the expiry date. Changing it to an overlap predicate would change `sumCommittedDaysForPolicyInWindow` for the whole-cycle callers too, and pro-rating needs the working-day calculation that currently lives in the service. I have documented the boundary on the repository method for now - want a separate overlap-aware method for the expiry check?

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
hasalarootcode: The cap is currently accrual-only, which is why carryover is added after it. Applying it to `accrualAllocation + usableCarryoverDays` would make it a balance ceiling and would also change `unusedAtCycleEnd`, so it changes what every existing carryover-enabled policy allocates. The message copy reads like a balance ceiling to me too. Which did you intend? I will make it explicit and pin it with a test once you confirm.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
hasalarootcode: Deliberate for now - the field is populated so the balance card can show it, but the response DTO change was going to wait for the FE to have somewhere to put it. Want `carriedOverDays` on `EmployeePolicyBalanceResponseDto` and `toBalanceCard` in this PR?

backend/src/main/java/com/skapp/community/leaveplanner/util/PolicyLeaveAccrualUtil.java
hasalarootcode: Fair, and the seam is there for it. Before I write them: the cap semantics question on `totalDaysAllocated` is still open, and pinning the current behaviour in tests first would mean rewriting them if the cap becomes a balance ceiling. Want the tests here once that is settled, or a follow-up alongside the batching change?

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
```diff
@@ -706,15 +716,38 @@ private EmployeeLeavePolicy resolveActiveAssignment(Employee employee, Long poli
...
+		return calculateBalance(assignment, PolicyLeaveAccrualUtil.resolveCycleContaining(date, cycleAnchor),
+				cycleAnchor, date);
+	}
+
+	private MonthDay resolveCycleAnchor() {
+		try {
+			LeaveCycleDetailsDto leaveCycle = leaveCycleService.getLeaveCycleConfigs();
+			return MonthDay.of(leaveCycle.getStartMonth(), leaveCycle.getStartDate());
+		}
+		catch (NullPointerException | DateTimeException e) {
```
shakila-kamalasena: DO not use warn logs

shakila-kamalasena (changes requested):

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
hasalarootcode: Done

shakila-kamalasena (dismissed):

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
```diff
@@ -747,11 +780,39 @@ private PolicyLeaveBalanceDto calculateBalance(EmployeeLeavePolicy assignment, P
...
+				assignment.getEffectiveFrom());
+		return PolicyLeaveAccrualUtil.roundToHalfDay(
+				PolicyLeaveAccrualUtil.accruedWithinCycle(policy, accrualStartDate, cycle, cycle.getEndDate()));
+	}
+
+	/**
+	 * Loads every committed request the balance walk can ask about in one query, so
+	 * replaying prior cycles for carryover costs no extra round trips.
+	 */
+	private PolicyLeaveUsageLookup buildUsageLookup(EmployeeLeavePolicy assignment, PolicyLeaveDateWindowDto cycle,
```
akilarootcode, line 791: remove comments

akilarootcode (changes requested):

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
hasalarootcode: Fixed

akilarootcode (approved):

shakila-kamalasena (approved):

PR #1767, opened 2026-08-17
fix: hide management tabs on a manager's own employee profile

Deshan1100:
### Summary

-
image (3)

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/frontend/src/community/people/components/molecules/DirectorySteppers/DirectorySteppers.tsx
+++ b/frontend/src/community/people/components/molecules/DirectorySteppers/DirectorySteppers.tsx
@@ -32,7 +32,8 @@ const DirectorySteppers = ({
 
   const { user } = useAuth();
 
-  const { isSuperAdmin, isPeopleAdmin, isESignSender } = useSessionData();
+  const { isSuperAdmin, isPeopleAdmin, isESignSender, userId } =
+    useSessionData();
 
   const { setNextStep, currentStep } = usePeopleStore((state) => state);
 
@@ -54,6 +55,7 @@ const DirectorySteppers = ({
   );
 
   const isEditView = !isIndividualView && !isAccountView;
+  const isSelfView = Number(employeeId) === Number(userId);
 
   useEffect(() => {
     if (supervisedData && !supervisorDataLoading) {
@@ -86,14 +88,18 @@ const DirectorySteppers = ({
     ...(isEditView
       ? [translateText(["editAllInfo", "systemPermissions"])]
       : []),
-    ...(isEditView ? [translateText(["editAllInfo", "timeline"])] : []),
+    ...(isEditView && (isSuperAdmin || isPeopleAdmin)
+      ? [translateText(["editAllInfo", "timeline"])]
+      : []),
     ...(isLeaveTabVisible &&
     !isAccountView &&
+    (!isSelfView || isLeaveAdmin) &&
     user?.roles?.includes(EmployeeTypes.LEAVE_EMPLOYEE)
       ? [translateText(["editAllInfo", "leave"])]
       : []),
     ...(isTimeTabVisible &&
     !isAccountView &&
+    (!isSelfView || isAttendanceAdmin) &&
     user?.roles?.includes(EmployeeTypes.ATTENDANCE_EMPLOYEE)
       ? [translateText(["editAllInfo", "timesheet"])]
       : []),
--- a/frontend/src/community/people/components/molecules/DirectorySteppers/DirectorySteppers.tsx
+++ b/frontend/src/community/people/components/molecules/DirectorySteppers/DirectorySteppers.tsx
@@ -55,7 +55,7 @@ const DirectorySteppers = ({
   );
 
   const isEditView = !isIndividualView && !isAccountView;
-  const isSelfView = Number(employeeId) === Number(userId);
+  const isSelfView = employeeId === userId;
 
   useEffect(() => {
     if (supervisedData && !supervisorDataLoading) {


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR hides the Timeline tab from non-admins and hides the Leave/Timesheet tabs when a user views their own profile in DirectorySteppers. The self-view guard itself is correctly derived, but it only removes the stepper *label* — the corresponding sections are still rendered by PeopleFormSections/PeopleIndividualSection whenever `currentStep` points at them, and `useDefaultTabNavigation` sets that step straight from the `?tab=leave|timesheet` query param with no permission check, so the management view remains reachable. The Timeline change is also not scoped to self-view, which silently removes Timeline for People Managers on every profile, not just their own.

Found **14** new issue(s): 🔴 4 important, 🟡 5 suggestion(s), 🟣 5 nit(s)

frontend/src/community/people/components/molecules/DirectorySteppers/DirectorySteppers.tsx
```diff
@@ -86,14 +88,18 @@ const DirectorySteppers = ({
     ...(isEditView
       ? [translateText(["editAllInfo", "systemPermissions"])]
       : []),
-    ...(isEditView ? [translateText(["editAllInfo", "timeline"])] : []),
+    ...(isEditView && (isSuperAdmin || isPeopleAdmin)
+      ? [translateText(["editAllInfo", "timeline"])]
+      : []),
     ...(isLeaveTabVisible &&
     !isAccountView &&
+    (!isSelfView || isLeaveAdmin) &&
```
ThinuwanW, line 96: 🤖 **Claude** · 🔴 **Important**

Hiding the step label does not hide the step content, so the self-view restriction is bypassable. `DirectoryEditSectionWrapper` and `IndividualSectionWrapper` both call `useDefaultTabNavigation()`, which reads `router.query.tab` and calls `setNextStep(EditPeopleFormTypes.leave)` / `.timesheet` with no permission check; `PeopleFormSections.getEditFlowSection()` (and `PeopleIndividualSection.getSections()`) then render `IndividualEmployeeLeaveReportSection` / `IndividualEmployeeTimeReportSection` purely off `currentStep`. Concrete repro: a LEAVE_MANAGER + PEOPLE_MANAGER who is a team supervisor of their own team opens /leave/leave-requests, searches for themselves in the autocomplete (handleRowClick in pages/community/leave/leave-requests.tsx pushes `<PEOPLE.EDIT(ownId)>?tab=leave`) — the Leave step label is now gone but the leave management section still renders. A second path: `currentStep` lives in the module-level Zustand people slice and `resetPeopleSlice` is only ever called from tests, so selecting Leave on employee A's profile and then client-side navigating to your own profile leaves `currentStep === leave` and renders the section with no matching tab. Fix by making the visibility rules the single source of truth: compute the allowed steps (e.g. export a `useVisibleDirectorySteps(employeeId, view)` hook) and (a) have `PeopleFormSections`/`PeopleIndividualSection` fall back to `personal` when `currentStep` is not in that list, and (b) have `useDefaultTabNavigation` only honour `?tab=` when that step is allowed. As a secondary effect, when `currentStep` is not present in `steps`, BoxStepper gives every tab `aria-selected={false}` and `tabIndex={-1}`, so the whole tablist becomes unreachable by keyboard — the fallback above fixes that too.

frontend/src/community/people/components/molecules/DirectorySteppers/DirectorySteppers.tsx
```diff
@@ -86,14 +88,18 @@ const DirectorySteppers = ({
     ...(isEditView
       ? [translateText(["editAllInfo", "systemPermissions"])]
       : []),
-    ...(isEditView ? [translateText(["editAllInfo", "timeline"])] : []),
+    ...(isEditView && (isSuperAdmin || isPeopleAdmin)
```
ThinuwanW, line 91: 🤖 **Claude** · 🔴 **Important**

The Timeline gate is not scoped to self-view, unlike the two changes below it, so it goes further than the stated intent ("hide management tabs on a manager's own employee profile"). Previously every edit-view visitor saw Timeline; now a PEOPLE_MANAGER (who is allowed on /people/directory/edit/[id] — `employeeRestrictedRoutes` only blocks non-PEOPLE_MANAGERs) loses Timeline on *every* employee they manage, not just their own profile. If the intent is self-profile only, this should read `isEditView && (!isSelfView || isSuperAdmin || isPeopleAdmin)`. If hiding Timeline from managers everywhere is genuinely intended, please say so in the PR description, since it is a separate permission change. Related consistency point: `isSuperAdmin` is included here but not in the Leave/Timesheet self-view exceptions — and per backend AuthorityServiceImpl.addSuperAdminAuthority, SUPER_ADMIN does not imply LEAVE_ADMIN/ATTENDANCE_ADMIN, so a super admin whose leave role is only LEAVE_MANAGER will lose those tabs on their own profile. Confirm that asymmetry is deliberate.

frontend/src/community/people/components/molecules/DirectorySteppers/DirectorySteppers.tsx
```diff
@@ -32,7 +32,8 @@ const DirectorySteppers = ({
 
   const { user } = useAuth();
 
-  const { isSuperAdmin, isPeopleAdmin, isESignSender } = useSessionData();
+  const { isSuperAdmin, isPeopleAdmin, isESignSender, userId } =
```
ThinuwanW, line 35: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

No test coverage was added for the new visibility rules, and the existing DirectorySteppers.test.tsx cannot cover them: it mocks `next-auth/react`'s `useSession`, but the component reads roles via `useAuth()` from ~community/auth/providers/AuthProvider and `useSessionData()`. `useAuth` throws "useAuth must be used within an AuthProvider" when no provider is present (AuthProvider.tsx:166-168) and jest.config.js registers no setupFiles that mock it, so the existing suite for this file is already broken/ineffective rather than asserting the old behaviour. Please fix the mock (mock `~community/auth/providers/AuthProvider` and `~community/common/hooks/useSessionData`, or wrap in a provider) and add cases for: self-view + LEAVE_MANAGER only (Leave hidden), self-view + LEAVE_ADMIN (Leave shown), self-view + ATTENDANCE_MANAGER only (Timesheet hidden), other-employee view (both shown), and edit view as PEOPLE_MANAGER (Timeline hidden).

frontend/src/community/people/components/molecules/DirectorySteppers/DirectorySteppers.tsx
```diff
@@ -54,6 +55,7 @@ const DirectorySteppers = ({
   );
 
   const isEditView = !isIndividualView && !isAccountView;
+  const isSelfView = Number(employeeId) === Number(userId);
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`Number(employeeId) === Number(userId)` relies on NaN !== NaN to be safe when the session has not loaded (`userId` is typed `number | undefined` in authUtils). That happens to fail open (tab shown), which is the wrong default for a permission gate, and the `Number()` wrapper on `employeeId` is redundant since the prop is already typed `number`. Prefer an explicit check such as `const isSelfView = userId !== undefined && Number(employeeId) === userId;` and consider gating the stepper render on the session being resolved (`sessionStatus === "authenticated"` from useSessionData) so a step never flashes before roles are known.

frontend/src/community/people/components/molecules/DirectorySteppers/DirectorySteppers.tsx
```diff
@@ -86,14 +88,18 @@ const DirectorySteppers = ({
     ...(isEditView
       ? [translateText(["editAllInfo", "systemPermissions"])]
       : []),
-    ...(isEditView ? [translateText(["editAllInfo", "timeline"])] : []),
+    ...(isEditView && (isSuperAdmin || isPeopleAdmin)
+      ? [translateText(["editAllInfo", "timeline"])]
+      : []),
     ...(isLeaveTabVisible &&
```
ThinuwanW, line 94: 🤖 **Claude** · 🟣 **Nit**

The Leave and Timesheet entries are now 4-term boolean chains inline in the array literal, which is hard to read and easy to get wrong on the next change. Extract named booleans above the array, e.g. `const canSeeLeaveTab = isLeaveTabVisible && !isAccountView && (!isSelfView || isLeaveAdmin) && isLeaveEmployee;` and the analogous `canSeeTimesheetTab` / `canSeeTimelineTab`, then spread on those. This also gives you the values you need to reuse for the content-level guard described in the other comment.

frontend/src/community/people/components/molecules/DirectorySteppers/DirectorySteppers.tsx
```diff
@@ -54,6 +55,7 @@ const DirectorySteppers = ({
   );
```
ThinuwanW, line 55: 🤖 **Claude** · 🟣 **Nit**

Pre-existing but now load-bearing for the new admin-vs-manager distinction: `user?.roles?.includes(ManagerTypes.LEAVE_MANAGER || AdminTypes.LEAVE_ADMIN)` evaluates the `||` before `includes`, so it is just `includes(LEAVE_MANAGER)` — the LEAVE_ADMIN half is dead. Same for the ATTENDANCE variant. It is currently harmless because the backend role hierarchy (AuthorityServiceImpl.addRoleHierarchy) grants LEAVE_MANAGER to LEAVE_ADMINs, but the expression reads as if it checks both and will mislead the next reader of this now-more-complex condition. Replace with `includes(ManagerTypes.LEAVE_MANAGER) || includes(AdminTypes.LEAVE_ADMIN)`, or better, use `isLeaveManager`/`isLeaveAdmin`/`isAttendanceManager` from `useSessionData()` (already exported there, and `userId` is now being pulled from that hook anyway) instead of re-deriving roles from `useAuth().user` in this component — only `isAttendanceAdmin` is missing and could be added to useSessionData.

frontend/src/community/people/components/molecules/DirectorySteppers/DirectorySteppers.tsx
ThinuwanW, line 35: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

The same stepper-visibility logic is duplicated in two legacy pages that were not updated by this PR and now diverge: pages/community/people/directory/edit-all-information/[id].tsx (lines ~181-194 and ~592-614) and pages/community/people/individual.tsx (lines ~81-92 and ~171-193). Both build the Leave/Timesheet steps from the identical `useGetSupervisedByMe` + role checks with no self-view guard, and both set the form type straight from `?tab=`. Nothing links to `ROUTES.PEOPLE.EDIT_ALL_INFORMATION` or bare `/people/individual` any more, but they are still real Next.js routes, and `/people/directory/edit-all-information/` is not listed in `employeeRestrictedRoutes` (only `/people/directory/edit/` is), so they are directly navigable. Either delete these legacy pages plus the unused `EDIT_ALL_INFORMATION` route constant, or apply the same guard there.

frontend/src/community/people/components/molecules/DirectorySteppers/DirectorySteppers.tsx
ThinuwanW, line 96: 🤖 **Claude** · 🔴 **Important**

Missing paired change in `useDefaultTabNavigation` — the hidden steps are still reachable automatically, with no user action. `pages/community/leave/leave-requests.tsx:56,61` and `pages/community/timesheet/all-timesheets.tsx:43,48` push `?tab=leave` / `?tab=timesheet`, and `useDefaultTabNavigation` (called by both `DirectoryEditSectionWrapper` and `IndividualSectionWrapper`) unconditionally does `setNextStep(EditPeopleFormTypes.leave/timesheet)` in a mount-only effect that knows nothing about the new visibility rules. Concrete repro: a leave manager who is a team supervisor of their own team clicks their own row in All Leave Requests -> `/people/individual/<selfId>?tab=leave` -> the step is forced to `leave`, the Leave report renders, but the Leave tab is now absent from the stepper, so the page shows management content with no matching tab. Fix by making the visibility decision the single source of truth: lift the leave/timesheet/timeline visibility booleans out of `DirectorySteppers` (a `usePeopleStepVisibility(employeeId)` hook, or store them on the people slice) and have `useDefaultTabNavigation` only honour the `tab` query param when that step is visible, falling back to `EditPeopleFormTypes.personal` otherwise.

frontend/src/community/people/components/molecules/DirectorySteppers/DirectorySteppers.tsx
ThinuwanW, line 91: 🤖 **Claude** · 🔴 **Important**

No fallback when `currentStep` is no longer present in `steps`, which breaks the tablist for keyboard and screen-reader users. `BoxStepper` implements a roving tabindex: `tabIndex={activeStep === index || activeStep === step ? 0 : -1}` and `aria-selected={activeStep === index || activeStep === step}` (BoxStepper.tsx:80,110). Since `activeStep` is `currentStep` and `useStringIdentifier` matches on the label string, removing the currently-active step (Timeline for a people manager, Leave/Timesheet on one's own profile, plus the deep-link case) leaves every tab at `tabIndex={-1}` and no tab with `aria-selected="true"` — the `role="tablist"` becomes completely unreachable by Tab (WCAG 2.1.1) and AT announces no selected tab. Fix in this component by adding an effect that resets to a visible step, e.g. `useEffect(() => { if (steps.length && !steps.includes(currentStep)) setNextStep(steps[0] as EditPeopleFormTypes); }, [steps, currentStep])`; ideally also harden `BoxStepper` so that when no step matches `activeStep`, index 0 gets `tabIndex={0}`.

frontend/src/community/people/components/molecules/DirectorySteppers/DirectorySteppers.tsx
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Third competing implementation of the same "is this my own profile" predicate, with a third comparison idiom. `PeopleFormSections.tsx:36,44` uses `userId !== employeeId` / `userId === employeeId` (strict, no coercion) and `pages/community/people/directory/edit-all-information/[id].tsx:313` uses `currentEmployeeDetails?.employeeId === (id?.toString() ?? "")` (string compare). This PR adds `Number(employeeId) === Number(userId)`. Three idioms for one concept will drift, and the `Number()` wrapper is also redundant here since `Props.employeeId` is already typed `number` — the coercion only papers over callers passing `Number(router.query.id)` (NaN before hydration). Extract a single `useIsOwnProfile(employeeId: number): boolean` hook under `~community/people/hooks/` that handles the not-yet-loaded session explicitly (e.g. `userId !== undefined && employeeId === userId`), and use it in all three places. Also consider naming it consistently with the existing `isPeopleAdminViewingOwnProfile` (e.g. `isViewingOwnProfile`) rather than `isSelfView`, which reads too much like the neighbouring `isIndividualView` prop.

frontend/src/community/people/components/molecules/DirectorySteppers/DirectorySteppers.tsx
ThinuwanW, line 35: 🤖 **Claude** · 🟡 **Suggestion**

The component now straddles two sources for the same session data. This line pulls `userId` from `useSessionData()`, yet lines 45-55 still derive `isLeaveAdmin`, `isAttendanceAdmin`, `isLeaveManager` and `isAttendanceManager` from `useAuth().user?.roles`, and lines 97/103 read `user?.roles` directly. `useSessionData` already exports `isLeaveAdmin`, `isLeaveManager`, `isAttendanceManager`, `isLeaveEmployee` and `isAttendanceEmployee` as memoised values (useSessionData.ts:94-117); `isAttendanceAdmin` is the only one missing. Add `isAttendanceAdmin` to `useSessionData` alongside the existing `isLeaveAdmin`, then destructure all role flags from that one hook and drop the `useAuth` import — that removes the local duplicates, removes the un-memoised `roles.includes` calls on every render, and keeps role derivation in one place for the whole app.

frontend/src/community/people/components/molecules/DirectorySteppers/DirectorySteppers.tsx
```diff
@@ -86,14 +88,18 @@ const DirectorySteppers = ({
...
+      : []),
     ...(isLeaveTabVisible &&
     !isAccountView &&
+    (!isSelfView || isLeaveAdmin) &&
     user?.roles?.includes(EmployeeTypes.LEAVE_EMPLOYEE)
       ? [translateText(["editAllInfo", "leave"])]
       : []),
     ...(isTimeTabVisible &&
     !isAccountView &&
+    (!isSelfView || isAttendanceAdmin) &&
```
ThinuwanW, line 102: 🤖 **Claude** · 🟡 **Suggestion**

The self-view escape hatch grants module admins but not super admins, which contradicts the Timeline gate 11 lines above. Line 91 uses `isSuperAdmin || isPeopleAdmin`, while lines 96/102 use only `isLeaveAdmin` / `isAttendanceAdmin`. Backend role assignment makes these independent: `AuthorityServiceImpl.addSuperAdminAuthority` adds `ROLE_SUPER_ADMIN` on its own, and module authorities come separately from `employeeRole.getLeaveRole()` / `getAttendanceRole()` — so a super admin whose leave role is `LEAVE_MANAGER` (not `LEAVE_ADMIN`) has `ROLE_LEAVE_MANAGER` but not `ROLE_LEAVE_ADMIN`, and if they supervise their own team they lose the Leave/Timesheet tabs on their own profile that they had before this PR. Decide the intent and make it explicit — most likely `(!isSelfView || isSuperAdmin || isLeaveAdmin)` and `(!isSelfView || isSuperAdmin || isAttendanceAdmin)` — so all three gates read consistently.

frontend/src/community/people/components/molecules/DirectorySteppers/DirectorySteppers.tsx
```diff
@@ -86,14 +88,18 @@ const DirectorySteppers = ({
     ...(isEditView
```
ThinuwanW, line 88: 🤖 **Claude** · 🟡 **Suggestion**

The `systemPermissions` step was left ungated while Timeline became admin-only, so the PR's stated rule is applied inconsistently across the management tabs. This step is shown for any `isEditView` user, meaning a People Manager viewing their own profile still gets the System Permissions tab (rendered read-only via `PeopleFormSections.tsx:38` `isSystemPermissionsReadOnly = !isPeopleAdmin`, but still a management surface), and a People Admin on their own profile can edit their own permissions there except the super-admin toggle (`SystemPermissionFormSection.tsx:271`). Either bring this step under the same rule as the others (e.g. `isEditView && isPeopleAdmin && (!isSelfView || isSuperAdmin)`) or leave Timeline as-is — right now the two neighbouring management tabs follow different rules with no comment explaining why.

frontend/src/community/people/components/molecules/DirectorySteppers/DirectorySteppers.tsx
ThinuwanW, line 35: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

`useGetSupervisedByMe(Number(employeeId))` (line 42-43) still fires on every profile view, including account view (`!isAccountView` already excludes those steps) and now self-view for non-admins, where the response can no longer make either tab visible. Its `enabled` is only `!!employeeId` (PeopleApi.ts:788), so this is one wasted request per profile open. Consider passing the new condition into the hook's `enabled`. Related: the hook is typed `UseQueryResult<any>`, so the `supervisedData.isPrimaryManager / isSecondaryManager / isTeamSupervisor` reads in the effect (lines 66-68, 76-78) are unchecked `any` — now that this data gates tab visibility, giving it a real response type (`PrimarySecondaryOrTeamSupervisorResponseDto`) would catch backend shape drift at compile time instead of silently yielding `undefined` and hiding tabs.

Deshan1100: No test needed

Deshan1100: out of scope only need this change

Deshan1100: need user id

Deshan1100: need user id

frontend/src/community/people/components/molecules/DirectorySteppers/DirectorySteppers.tsx
Deshan1100: kept as previous

frontend/src/community/people/components/molecules/DirectorySteppers/DirectorySteppers.tsx
Deshan1100: need this check

frontend/src/community/people/components/molecules/DirectorySteppers/DirectorySteppers.tsx
Deshan1100: kept as previous

frontend/src/community/people/components/molecules/DirectorySteppers/DirectorySteppers.tsx
Deshan1100: yes need to not visible there own tabs to managers

Deshan1100: No need fallback kept as previous

frontend/src/community/people/components/molecules/DirectorySteppers/DirectorySteppers.tsx
Deshan1100: kept as previous

frontend/src/community/people/components/molecules/DirectorySteppers/DirectorySteppers.tsx
Deshan1100: requirement is to hide the tab view

frontend/src/community/people/components/molecules/DirectorySteppers/DirectorySteppers.tsx
Deshan1100: Admins need to view all tabs related to atendance

frontend/src/community/people/components/molecules/DirectorySteppers/DirectorySteppers.tsx
Deshan1100: out of scope only need to restrict to Managers

frontend/src/community/people/components/molecules/DirectorySteppers/DirectorySteppers.tsx
Deshan1100: use this when data loaded

shakila-kamalasena: Remove the number cast

shakila-kamalasena (changes requested):

frontend/src/community/people/components/molecules/DirectorySteppers/DirectorySteppers.tsx
Deshan1100: updated

shakila-kamalasena (approved):

thusala (approved):

akilarootcode (approved):

PR #1768, opened 2026-08-17
fix: persist dismissed birthday modals to prevent them reappearing on refresh

SanithuGunasena07:
Fixes birthday notification modals reappearing after a page refresh once already dismissed, by persisting dismissed employee IDs to localStorage (scoped by user + date) and filtering them out when the queue rebuilds on mount.

### DIFF
--- a/frontend/src/community/people/constants/stringConstants.ts
+++ b/frontend/src/community/people/constants/stringConstants.ts
@@ -8,6 +8,8 @@ export const EMAIL_MAX_LENGTH = 100;
 export const BIRTHDAY_NOTIFICATION_VIEW_STATE_CACHE_KEY =
   "birthdayNotificationViewState";
 
+export const BIRTHDAY_DISMISSED_ENTRIES_CACHE_KEY = "birthdayDismissedEntries";
+
 export const PAYROLL_ID_LENGTH = 50;
 
 export const TIN_LENGTH = 50;
--- a/frontend/src/community/people/hooks/useBirthdayNotifications.ts
+++ b/frontend/src/community/people/hooks/useBirthdayNotifications.ts
@@ -15,6 +15,11 @@ import {
   MarkBirthdayNotificationsViewedResponse
 } from "~community/people/types/BirthdayNotificationTypes";
 import {
+  getDismissedEmployeeIds,
+  readDismissedCache,
+  writeDismissedCache
+} from "~community/people/utils/birthdayNotificationCacheUtils";
+import {
   buildBirthdayQueue,
   normalizeEmployeeId
 } from "~community/people/utils/birthdayNotificationUtils";
@@ -79,7 +84,16 @@ const useBirthdayNotifications = (): BirthdayNotificationsType => {
       currentEmployeeId
     );
 
-    if (birthdayQueue.length === 0) {
+    const dismissedEmployeeIds = getDismissedEmployeeIds(
+      readDismissedCache(),
+      today,
+      userId
+    );
+    const pendingQueue = birthdayQueue.filter(
+      (entry) => !dismissedEmployeeIds.includes(entry.employee.employeeId)
+    );
+
+    if (pendingQueue.length === 0) {
       if (data.lastViewedDate !== today) {
         markViewed();
       }
@@ -91,12 +105,13 @@ const useBirthdayNotifications = (): BirthdayNotificationsType => {
         ? document.activeElement
         : null;
     isShowingRef.current = true;
-    setQueue(birthdayQueue);
+    setQueue(pendingQueue);
     setCursor(0);
   }, [
     data,
     shouldEvaluate,
     today,
+    userId,
     currentEmployeeId,
     isCurrentEmployeeResolved,
     markViewed,
@@ -119,6 +134,25 @@ const useBirthdayNotifications = (): BirthdayNotificationsType => {
   }, [markViewed]);
 
   const onDismiss = useCallback(() => {
+    const dismissedEntry = queue[cursor];
+
+    if (dismissedEntry && userId !== undefined) {
+      const existingDismissedIds = getDismissedEmployeeIds(
+        readDismissedCache(),
+        today,
+        userId
+      );
+      const dismissedEmployeeId = dismissedEntry.employee.employeeId;
+
+      if (!existingDismissedIds.includes(dismissedEmployeeId)) {
+        writeDismissedCache({
+          userId,
+          date: today,
+          dismissedEmployeeIds: [...existingDismissedIds, dismissedEmployeeId]
+        });
+      }
+    }
+
     const nextCursor = cursor + 1;
 
     if (nextCursor < queue.length) {
@@ -127,7 +161,7 @@ const useBirthdayNotifications = (): BirthdayNotificationsType => {
     }
 
     finishSequence();
-  }, [cursor, queue.length, finishSequence]);
+  }, [cursor, queue, userId, today, finishSequence]);
 
   useEffect(() => {
     if (evaluationTick === 0) return;
--- a/frontend/src/community/people/hooks/useBirthdayNotifications.ts
+++ b/frontend/src/community/people/hooks/useBirthdayNotifications.ts
@@ -15,6 +15,7 @@ import {
   MarkBirthdayNotificationsViewedResponse
 } from "~community/people/types/BirthdayNotificationTypes";
 import {
+  clearDismissedCache,
   getDismissedEmployeeIds,
   readDismissedCache,
   writeDismissedCache
@@ -72,6 +73,7 @@ const useBirthdayNotifications = (): BirthdayNotificationsType => {
     if (!shouldEvaluate) return;
     if (isShowingRef.current) return;
     if (!isCurrentEmployeeResolved) return;
+    if (userId === undefined) return;
 
     seededDataRef.current = data;
 
@@ -136,7 +138,7 @@ const useBirthdayNotifications = (): BirthdayNotificationsType => {
   const onDismiss = useCallback(() => {
     const dismissedEntry = queue[cursor];
 
-    if (dismissedEntry && userId !== undefined) {
+    if (dismissedEntry && userId !== undefined && today) {
       const existingDismissedIds = getDismissedEmployeeIds(
         readDismissedCache(),
         today,
@@ -180,6 +182,7 @@ const useBirthdayNotifications = (): BirthdayNotificationsType => {
     restoreFocusRef.current = null;
     setQueue([]);
     setCursor(0);
+    clearDismissedCache();
   }, [userId]);
 
   return {
--- a/frontend/src/community/people/utils/birthdayNotificationCacheUtils.ts
+++ b/frontend/src/community/people/utils/birthdayNotificationCacheUtils.ts
@@ -12,61 +12,22 @@ import {
   BirthdayNotificationViewStateType
 } from "~community/people/types/BirthdayNotificationTypes";
 
-const readCache = <T>(
-  key: string,
-  isValid: (value: unknown) => value is T
-): T | null => {
+export const readViewedCache = (): BirthdayNotificationViewStateType | null => {
   try {
-    const cachedValue = getDataFromLocalStorage(key);
+    const cachedValue = getDataFromLocalStorage(
+      BIRTHDAY_NOTIFICATION_VIEW_STATE_CACHE_KEY
+    );
 
-    if (!isValid(cachedValue)) return null;
+    if (!cachedValue || typeof cachedValue !== "object") return null;
+    if (typeof cachedValue.userId !== "number") return null;
+    if (typeof cachedValue.lastViewedDate !== "string") return null;
 
-    return cachedValue;
+    return cachedValue as BirthdayNotificationViewStateType;
   } catch {
-    removeDataFromLocalStorage(key);
+    removeDataFromLocalStorage(BIRTHDAY_NOTIFICATION_VIEW_STATE_CACHE_KEY);
     return null;
   }
 };
-
-const isViewedCacheShape = (
-  value: unknown
-): value is BirthdayNotificationViewStateType => {
-  const cache = value as BirthdayNotificationViewStateType;
-
-  return (
-    !!cache &&
-    typeof cache === "object" &&
-    typeof cache.userId === "number" &&
-    typeof cache.lastViewedDate === "string"
-  );
-};
-
-const isDismissedCacheShape = (
-  value: unknown
-): value is BirthdayDismissedCacheType => {
-  const cache = value as BirthdayDismissedCacheType;
-
-  return (
-    !!cache &&
-    typeof cache === "object" &&
-    typeof cache.userId === "number" &&
-    typeof cache.date === "string" &&
-    Array.isArray(cache.dismissedEmployeeIds)
-  );
-};
-
-const matchesUserAndDate = (
-  cacheUserId: number,
-  cacheDate: string,
-  today: string | null,
-  currentUserId: number | undefined
-): boolean => {
-  if (!today || currentUserId === undefined) return false;
-  return cacheUserId === currentUserId && cacheDate === today;
-};
-
-export const readViewedCache = (): BirthdayNotificationViewStateType | null =>
-  readCache(BIRTHDAY_NOTIFICATION_VIEW_STATE_CACHE_KEY, isViewedCacheShape);
 
 export const writeViewedCache = (
   cache: BirthdayNotificationViewStateType
@@ -78,12 +39,29 @@ export const isViewedToday = (
   cache: BirthdayNotificationViewStateType | null,
   today: string | null,
   currentUserId: number | undefined
-): boolean =>
-  !!cache &&
-  matchesUserAndDate(cache.userId, cache.lastViewedDate, today, currentUserId);
+): boolean => {
+  if (!cache || !today || currentUserId === undefined) return false;
+  if (cache.userId !== currentUserId) return false;
+  return cache.lastViewedDate === today;
+};
 
-export const readDismissedCache = (): BirthdayDismissedCacheType | null =>
-  readCache(BIRTHDAY_DISMISSED_ENTRIES_CACHE_KEY, isDismissedCacheShape);
+export const readDismissedCache = (): BirthdayDismissedCacheType | null => {
+  try {
+    const cachedValue = getDataFromLocalStorage(
+      BIRTHDAY_DISMISSED_ENTRIES_CACHE_KEY
+    );
+
+    if (!cachedValue || typeof cachedValue !== "object") return null;
+    if (typeof cachedValue.userId !== "number") return null;
+    if (typeof cachedValue.date !== "string") return null;
+    if (!Array.isArray(cachedValue.dismissedEmployeeIds)) return null;
+
+    return cachedValue as BirthdayDismissedCacheType;
+  } catch {
+    removeDataFromLocalStorage(BIRTHDAY_DISMISSED_ENTRIES_CACHE_KEY);
+    return null;
+  }
+};
 
 export const writeDismissedCache = (
   cache: BirthdayDismissedCacheType
@@ -99,7 +77,9 @@ export const getDismissedEmployeeIds = (
   cache: BirthdayDismissedCacheType | null,
   today: string | null,
   currentUserId: number | undefined
-): number[] =>
-  cache && matchesUserAndDate(cache.userId, cache.date, today, currentUserId)
-    ? cache.dismissedEmployeeIds
-    : [];
+): number[] => {
+  if (!cache || !today || currentUserId === undefined) return [];
+  if (cache.userId !== currentUserId) return [];
+  if (cache.date !== today) return [];
+  return cache.dismissedEmployeeIds;
+};
--- a/frontend/src/community/people/utils/birthdayNotificationCacheUtils.ts
+++ b/frontend/src/community/people/utils/birthdayNotificationCacheUtils.ts
@@ -3,8 +3,14 @@ import {
   removeDataFromLocalStorage,
   setDataToLocalStorage
 } from "~community/common/utils/accessLocalStorage";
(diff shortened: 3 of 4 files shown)


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds a localStorage-backed record of dismissed birthday modals so a mid-sequence page refresh no longer replays already-dismissed entries. The core logic is sound: the queue is filtered at seed time, the dismissal is persisted before `finishSequence()` runs (so the last entry is covered), and both the user-id and date guards correctly invalidate the cache across users and days. I found no hard correctness bug, but several robustness, cleanup, and test-coverage gaps are worth addressing before this ships.

Found **14** new issue(s): 🔴 2 important, 🟡 7 suggestion(s), 🟣 5 nit(s)

frontend/src/community/people/utils/birthdayNotificationCacheUtils.ts
```diff
@@ -38,3 +44,38 @@ export const isViewedToday = (
...
+export const readDismissedCache = (): BirthdayDismissedCacheType | null => {
+  try {
+    const cachedValue = getDataFromLocalStorage(
+      BIRTHDAY_DISMISSED_ENTRIES_CACHE_KEY
+    );
+
+    if (!cachedValue || typeof cachedValue !== "object") return null;
+    if (typeof cachedValue.userId !== "number") return null;
+    if (typeof cachedValue.date !== "string") return null;
+    if (!Array.isArray(cachedValue.dismissedEmployeeIds)) return null;
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`readDismissedCache` validates that `dismissedEmployeeIds` is an array but never validates its elements, then uses `as BirthdayDismissedCacheType` to assert the shape. If localStorage ever holds `{"userId":3,"date":"2026-08-17","dismissedEmployeeIds":["12"]}` (hand-edited, corrupted, or a future schema change), the cast succeeds and `dismissedEmployeeIds.includes(12)` returns false — the filter silently no-ops and the modal reappears, i.e. exactly the bug this PR fixes, with no error surfaced. Tighten the guard: `if (!Array.isArray(cachedValue.dismissedEmployeeIds) || !cachedValue.dismissedEmployeeIds.every((id: unknown) => typeof id === "number")) return null;`. The other three fields are already validated this carefully, so this is the one hole in the validator.

frontend/src/community/people/utils/birthdayNotificationCacheUtils.ts
```diff
@@ -3,8 +3,14 @@ import {
   removeDataFromLocalStorage,
   setDataToLocalStorage
 } from "~community/common/utils/accessLocalStorage";
-import { BIRTHDAY_NOTIFICATION_VIEW_STATE_CACHE_KEY } from "~community/people/constants/stringConstants";
-import { BirthdayNotificationViewStateType } from "~community/people/types/BirthdayNotificationTypes";
+import {
```
ThinuwanW, line 6: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

No tests were added for the three new exported functions, even though a colocated `birthdayNotificationCacheUtils.test.ts` already exists and covers the sibling `readViewedCache` / `isViewedToday` with exactly the cases that matter here (valid shape, missing key, wrong field types, unparseable JSON, user mismatch, date mismatch, undefined viewer). Mirror those describe blocks for `readDismissedCache` and `getDismissedEmployeeIds` — the date-rollover and different-user-on-same-browser cases in particular, since those are the guards that keep the new cache from suppressing legitimate notifications. A test for the new queue-filtering behaviour in `useBirthdayNotifications` (dismiss one entry, re-seed, assert the remaining queue) would also protect the actual regression this PR fixes.

frontend/src/community/people/hooks/useBirthdayNotifications.ts
```diff
@@ -14,6 +14,11 @@ import {
   BirthdayQueueEntryType,
   MarkBirthdayNotificationsViewedResponse
 } from "~community/people/types/BirthdayNotificationTypes";
+import {
```
ThinuwanW, line 17: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

The new `birthdayDismissedEntries` key is written but never removed. The reset effect that fires when `userId` becomes undefined (logout) clears `seededDataRef`, `isShowingRef`, `queue` and `cursor`, but leaves the dismissed record in localStorage indefinitely. Two consequences: (1) the record persists on the device long after the session ends, and (2) because the key stores a single `{ userId, date, dismissedEmployeeIds }` object rather than a per-user map, a second user logging in on the same browser and dismissing one modal overwrites the first user's dismissals — if user A dismissed 1 of 3 and logs back in the same day, they see all 3 again. Add `removeDataFromLocalStorage(BIRTHDAY_DISMISSED_ENTRIES_CACHE_KEY)` to the `userId === undefined` reset effect (the existing `AnnouncementProvider` clears its cache on de-auth in the same way), and/or have `readDismissedCache` drop a record whose `date` is not today so stale entries self-clean on read.

frontend/src/community/people/hooks/useBirthdayNotifications.ts
```diff
@@ -119,6 +134,25 @@ const useBirthdayNotifications = (): BirthdayNotificationsType => {
   }, [markViewed]);
 
   const onDismiss = useCallback(() => {
+    const dismissedEntry = queue[cursor];
+
+    if (dismissedEntry && userId !== undefined) {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The persistence branch is gated on `userId !== undefined` and fails silently otherwise. `User.userId` is declared optional (`userId?: number` in `~community/auth/utils/authUtils`) and is populated from JWT claims, while `isEligible` in `BirthdayNotificationProvider` only requires `!!user`. So for a session whose token lacks a `userId` claim, the modals still render but no dismissal is ever recorded and the pre-fix behaviour returns with no warning. Either assert the invariant explicitly (log/no-op the whole notification flow when `userId` is undefined, consistent with `cacheLastViewedDate`'s early return), or fall back to a non-user-scoped cache entry. At minimum this deserves a comment so the silent branch is intentional rather than an oversight.

frontend/src/community/people/utils/birthdayNotificationCacheUtils.ts
```diff
@@ -38,3 +44,38 @@ export const isViewedToday = (
...
+  }
+};
+
+export const writeDismissedCache = (
+  cache: BirthdayDismissedCacheType
+): void => {
+  setDataToLocalStorage(BIRTHDAY_DISMISSED_ENTRIES_CACHE_KEY, cache);
+};
+
+export const getDismissedEmployeeIds = (
```
ThinuwanW, line 76: 🤖 **Claude** · 🟣 **Nit**

`getDismissedEmployeeIds` and `isViewedToday` are structurally identical — both do `if (!cache || !today || currentUserId === undefined) return <empty>; if (cache.userId !== currentUserId) return <empty>;` then a date comparison — and `readDismissedCache` / `readViewedCache` differ only in the key and the field validator. Consider extracting a shared `readCache<T>(key, isValid)` helper and an `isCacheForUserAndDate(cache, today, userId)` guard so the two caches cannot drift apart (e.g. someone fixing the user-mismatch handling in one and forgetting the other). This is the checklist's "2+ hooks/utils with the same logic → extract" item.

frontend/src/community/people/utils/birthdayNotificationCacheUtils.ts
```diff
@@ -38,3 +44,38 @@ export const isViewedToday = (
...
+  cache: BirthdayDismissedCacheType | null,
+  today: string | null,
+  currentUserId: number | undefined
+): number[] => {
+  if (!cache || !today || currentUserId === undefined) return [];
+  if (cache.userId !== currentUserId) return [];
+  if (cache.date !== today) return [];
+  return cache.dismissedEmployeeIds;
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`return cache.dismissedEmployeeIds;` hands back the array held inside the parsed cache object rather than a copy, so a caller could mutate the cache's internals in place. Today's callers only read and spread it, so there is no live bug, but returning `[...cache.dismissedEmployeeIds]` (or typing the return as `readonly number[]`) makes the util safe against future in-place `push` usage.

frontend/src/community/people/hooks/useBirthdayNotifications.ts
```diff
@@ -91,12 +105,13 @@ const useBirthdayNotifications = (): BirthdayNotificationsType => {
         ? document.activeElement
         : null;
     isShowingRef.current = true;
-    setQueue(birthdayQueue);
+    setQueue(pendingQueue);
```
ThinuwanW, line 110: 🤖 **Claude** · 🟣 **Nit**

Seeding with `pendingQueue` changes the semantics of `total` (and therefore `position`) after a refresh: a user who saw "1 of 3" and dismissed one will see "1 of 2" rather than "2 of 3" on reload, because `total` is now the pending count instead of today's birthday count. This flows into the screen-reader-only `peopleAria.birthdayNotifications.notificationPosition` string in `BirthdayModalShell`, and the `total > 1` condition means the counter disappears entirely once only one entry remains. This is probably the intended trade-off, but if the counter is meant to convey "how many birthdays today" rather than "how many left to dismiss", keep the full `birthdayQueue.length` as `total` and offset `position` by the number already dismissed.

frontend/src/community/people/hooks/useBirthdayNotifications.ts
```diff
@@ -91,12 +105,13 @@ const useBirthdayNotifications = (): BirthdayNotificationsType => {
...
     setCursor(0);
   }, [
     data,
     shouldEvaluate,
     today,
+    userId,
```
ThinuwanW, line 116: 🤖 **Claude** · 🔴 **Important**

The `userId` dependency added here cannot change the outcome, so it is effectively dead — and it hides a real race. The effect assigns `seededDataRef.current = data` before it reads the dismissed cache, and the guard `if (seededDataRef.current === data) return;` short-circuits every subsequent run for the same `data` object. So if `userId` is still `undefined` on the first run, `getDismissedEmployeeIds(...)` returns `[]`, `pendingQueue` becomes the full queue, and previously dismissed birthdays are shown again — and the re-run triggered by `userId` resolving does nothing. This is reachable: `isEligible` only requires `!!user`, while `userId` is `user?.userId`, declared optional in `User` (`frontend/src/community/auth/utils/authUtils.ts:67`) and populated from JWT claims. Once the modal is showing, `isShowingRef.current` also blocks the refetch path, so there is no recovery until the sequence finishes. Fix: mirror the existing `isCurrentEmployeeResolved` guard — add `if (userId === undefined) return;` alongside the other early returns (before the `seededDataRef.current = data` assignment), so seeding is deferred until the viewer id is known.

frontend/src/community/people/utils/birthdayNotificationCacheUtils.ts
```diff
@@ -38,3 +44,38 @@ export const isViewedToday = (
...
+  } catch {
+    removeDataFromLocalStorage(BIRTHDAY_DISMISSED_ENTRIES_CACHE_KEY);
+    return null;
+  }
+};
+
+export const writeDismissedCache = (
+  cache: BirthdayDismissedCacheType
+): void => {
+  setDataToLocalStorage(BIRTHDAY_DISMISSED_ENTRIES_CACHE_KEY, cache);
```
ThinuwanW, line 69: 🤖 **Claude** · 🔴 **Important**

`writeDismissedCache` has no error handling, unlike its sibling `readDismissedCache` which is fully wrapped in try/catch. `setDataToLocalStorage` is declared `async`, so a `QuotaExceededError` or `SecurityError` (storage full, Safari private browsing, cookies/storage disabled by policy) is thrown inside the async function and surfaces as an unhandled promise rejection — a `try/catch` at the call site in `onDismiss` would not catch it either, because the function returns before the rejection occurs. The user-visible effect is that the dismissal is silently lost and the modal reappears on the next reload, which is exactly the bug this PR is meant to fix. Fix: `void setDataToLocalStorage(KEY, cache).catch(() => undefined);` (and return a boolean/void so callers can react if desired). The same applies to the `removeDataFromLocalStorage(...)` call in the `catch` block on line 61 — it is also async and its rejection is unhandled.

frontend/src/community/people/hooks/useBirthdayNotifications.ts
```diff
@@ -119,6 +134,25 @@ const useBirthdayNotifications = (): BirthdayNotificationsType => {
   }, [markViewed]);
 
   const onDismiss = useCallback(() => {
+    const dismissedEntry = queue[cursor];
```
ThinuwanW, line 139: 🤖 **Claude** · 🟡 **Suggestion**

Dismissal is now durable for the rest of the day, which changes the cost of an accidental close. `BirthdayModalShell` wires `onDismiss` directly into `LargeModal`'s `onClose` prop (`BirthdayModalShell.tsx:39`), and modal components in this stack normally fire `onClose` for backdrop clicks and the Escape key in addition to the close button. Before this PR a stray backdrop click only advanced the in-memory cursor, so a reload re-showed the full queue; now it writes the employee id to localStorage and that birthday is suppressed until midnight. Consider only persisting on an explicit dismiss (e.g. have `LargeModal` pass a close reason, or expose a separate `onDismissConfirmed` from the hook and keep backdrop/ESC as a non-persisting advance). At minimum, verify the `LargeModal` close semantics before merging and document the intent.

frontend/src/community/people/hooks/useBirthdayNotifications.ts
```diff
@@ -119,6 +134,25 @@ const useBirthdayNotifications = (): BirthdayNotificationsType => {
...
+    if (dismissedEntry && userId !== undefined) {
+      const existingDismissedIds = getDismissedEmployeeIds(
+        readDismissedCache(),
+        today,
+        userId
+      );
+      const dismissedEmployeeId = dismissedEntry.employee.employeeId;
+
+      if (!existingDismissedIds.includes(dismissedEmployeeId)) {
+        writeDismissedCache({
```
ThinuwanW, line 150: 🤖 **Claude** · 🟡 **Suggestion**

The write path persists `date: today` without validating it, while the read path explicitly rejects a falsy `today` (`getDismissedEmployeeIds` returns `[]` when `!today`). `today` is not guaranteed non-empty: `getDateForPeriod("day", "start")` returns `""` when `toISODate()` yields null (`dateTimeUtils.ts:440`), and `defaultBirthdayNotificationContext.today` is also `""`. In that case this writes a cache entry that the reader will never honour, so dismissals are silently discarded while still consuming the storage key. Add `if (!today) return;` (or fold it into the existing `dismissedEntry && userId !== undefined` condition) so the writer and reader agree on what a valid date is.

frontend/src/community/people/hooks/useBirthdayNotifications.ts
```diff
@@ -79,7 +84,16 @@ const useBirthdayNotifications = (): BirthdayNotificationsType => {
       currentEmployeeId
     );
 
-    if (birthdayQueue.length === 0) {
+    const dismissedEmployeeIds = getDismissedEmployeeIds(
```
ThinuwanW, line 89: 🤖 **Claude** · 🟡 **Suggestion**

The dismissed cache is accessed through raw util calls sprinkled across the hook — the identical `getDismissedEmployeeIds(readDismissedCache(), today, userId)` expression appears both here and in `onDismiss`, each one re-reading and re-parsing localStorage. This is inconsistent with how the sibling viewed cache is handled: that one is owned by a dedicated `useBirthdayViewedCache(today)` hook which loads once into state and exposes `isCacheLoaded`/`cacheLastViewedDate`. Extract a matching `useBirthdayDismissedCache(today, userId)` hook under `~community/people/hooks/` that owns the read/write for this key and exposes `dismissedEmployeeIds` plus a `markDismissed(employeeId)` callback. That removes the duplicated expression, gives you a natural place for the `isCacheLoaded` deferral that fixes the seeding race, and keeps `useBirthdayNotifications` free of storage details.

frontend/src/community/people/constants/stringConstants.ts
```diff
@@ -8,6 +8,8 @@ export const EMAIL_MAX_LENGTH = 100;
 export const BIRTHDAY_NOTIFICATION_VIEW_STATE_CACHE_KEY =
   "birthdayNotificationViewState";
 
+export const BIRTHDAY_DISMISSED_ENTRIES_CACHE_KEY = "birthdayDismissedEntries";
```
ThinuwanW, line 11: 🤖 **Claude** · 🟣 **Nit**

Naming drifts from the sibling it sits next to. The existing constant is `BIRTHDAY_NOTIFICATION_VIEW_STATE_CACHE_KEY` -> `"birthdayNotificationViewState"`; the new one drops the `NOTIFICATION` segment: `BIRTHDAY_DISMISSED_ENTRIES_CACHE_KEY` -> `"birthdayDismissedEntries"`. Since both keys belong to the same feature and share a storage namespace, prefer `BIRTHDAY_NOTIFICATION_DISMISSED_ENTRIES_CACHE_KEY` -> `"birthdayNotificationDismissedEntries"`. Related: in `BirthdayNotificationTypes.ts:45`, `BirthdayDismissedCacheType` names its date field `date` while `BirthdayNotificationViewStateType` uses the descriptive `lastViewedDate`; `dismissedDate` would read consistently and avoid the bare `date` name. Note that changing the string literal orphans any already-written cache entries, so if you rename, do it now before this ships.

frontend/src/community/people/utils/birthdayNotificationCacheUtils.ts
ThinuwanW, line 6: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

This is the second independent implementation of "remember which modals this user already dismissed" in the codebase — `src/enterprise/common/providers/AnnouncementProvider.tsx` solves the same problem with its own `announcementsCache` localStorage entry and its own dismissed-id tracking. The two implementations have diverged in validation, error handling and lifecycle management, so a fix to one will not benefit the other. Consider promoting a small generic dismissed-ids cache helper into `~community/common/utils/` (enterprise can import from community, so both call sites could use it) rather than growing a third variant next time. Not blocking for this PR, but worth a follow-up ticket.

SanithuGunasena07: not required

frontend/src/community/people/utils/birthdayNotificationCacheUtils.ts
SanithuGunasena07: keeping as it is cause s the existing patternAnnouncementProvider.tsx does the same shallow Array.isArray check with no element validation

frontend/src/community/people/hooks/useBirthdayNotifications.ts
SanithuGunasena07: added clearDismissedCache() now fires in the logout reset effect alongside the existing in-memory resets, same pattern AnnouncementProvider uses on de-auth."

frontend/src/community/people/hooks/useBirthdayNotifications.ts
SanithuGunasena07: keeping as it is Same pattern as the existing cacheLastViewedDate guard in useBirthdayViewedCache.ts

frontend/src/community/people/utils/birthdayNotificationCacheUtils.ts
SanithuGunasena07: extracted readCache<T> and matchesUserAndDate as shared helpers; both cache pairs now go through the same validation/guard logic instead of duplicating it

frontend/src/community/people/utils/birthdayNotificationCacheUtils.ts
SanithuGunasena07: not required

frontend/src/community/people/hooks/useBirthdayNotifications.ts
SanithuGunasena07: The behaviour is ok hence leaving as it is

frontend/src/community/people/hooks/useBirthdayNotifications.ts
SanithuGunasena07: if (userId === undefined) return; before the seededDataRef.current = data assignment, mirroring the existing isCurrentEmployeeResolved guard

frontend/src/community/people/utils/birthdayNotificationCacheUtils.ts
SanithuGunasena07: Same pattern as the existing writeViewedCache/readViewedCache not new to this PR

frontend/src/community/people/hooks/useBirthdayNotifications.ts
SanithuGunasena07: not required current functionality alligns with requirement

frontend/src/community/people/utils/birthdayNotificationCacheUtils.ts
SanithuGunasena07: not required

frontend/src/community/people/hooks/useBirthdayNotifications.ts
SanithuGunasena07: added  onDismiss's guard now includes && today

frontend/src/community/people/hooks/useBirthdayNotifications.ts
SanithuGunasena07: out of scope for PR

frontend/src/community/people/constants/stringConstants.ts
SanithuGunasena07: not required

hasalarootcode (approved):

PR #1769, opened 2026-08-17
Fix/3000 Leave policy - Accrual policy type card icon does not follow theme color and description is missing a full stop

hasalarootcode:
…lor in PolicyTypeCard

### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/frontend/src/community/common/assets/languages/english/leaveModule.json
+++ b/frontend/src/community/common/assets/languages/english/leaveModule.json
@@ -367,7 +367,7 @@
       "basicInfo": {
         "policyTypeTitle": "Policy type",
         "accrualTitle": "Accrual",
-        "accrualDescription": "Time taken is deducted from a balance that accrues on a schedule or is set manually",
+        "accrualDescription": "Time taken is deducted from a balance that accrues on a schedule or is set manually.",
         "flexibleTitle": "Flexible (unlimited)",
         "flexibleDescription": "There is no balance to deduct from, but I still want leave requests to be approved and tracked.",
         "basicDetailsTitle": "Basic details",
--- a/frontend/src/community/common/assets/languages/english/leaveModule.json
+++ b/frontend/src/community/common/assets/languages/english/leaveModule.json
@@ -368,7 +368,7 @@
         "policyTypeTitle": "Policy type",
         "accrualTitle": "Accrual",
         "accrualDescription": "Time taken is deducted from a balance that accrues on a schedule or is set manually.",
-        "flexibleTitle": "Flexible (unlimited)",
+        "flexibleTitle": "Flexible (Unlimited)",
         "flexibleDescription": "There is no balance to deduct from, but I still want leave requests to be approved and tracked.",
         "basicDetailsTitle": "Basic details",
         "policyNameLabel": "Policy name",
--- a/frontend/src/community/leave/components/molecules/PolicyTypeSelectionModal/PolicyTypeCard.tsx
+++ b/frontend/src/community/leave/components/molecules/PolicyTypeSelectionModal/PolicyTypeCard.tsx
@@ -22,7 +22,7 @@ const PolicyTypeCard: FC<Props> = ({ icon, title, description, onSelect }) => (
     className="flex-1 cursor-pointer rounded-xl aspect-133.75/109.5 bg-white hover:shadow-md hover:outline-primary-accent"
   >
     <div className="flex h-full flex-col items-center justify-center gap-8 text-center">
-      <span className="text-primary-accent" aria-hidden="true">
+      <span className="text-black" aria-hidden="true">
         {icon}
       </span>
       <span className="flex flex-col gap-4">
--- a/frontend/src/community/leave/components/molecules/PolicyTypeSelectionModal/PolicyTypeCard.tsx
+++ b/frontend/src/community/leave/components/molecules/PolicyTypeSelectionModal/PolicyTypeCard.tsx
@@ -19,7 +19,7 @@ const PolicyTypeCard: FC<Props> = ({ icon, title, description, onSelect }) => (
         onSelect();
       }
     }}
-    className="flex-1 cursor-pointer rounded-xl aspect-133.75/109.5 bg-white hover:shadow-md hover:outline-primary-accent"
+    className="flex-1 cursor-pointer rounded-xl aspect-133.75/109.5 bg-white hover:bg-tertiary-background"
   >
     <div className="flex h-full flex-col items-center justify-center gap-8 text-center">
       <span className="text-black" aria-hidden="true">
--- a/frontend/src/community/leave/components/molecules/PolicyTypeSelectionModal/PolicyTypeSelectionModal.tsx
+++ b/frontend/src/community/leave/components/molecules/PolicyTypeSelectionModal/PolicyTypeSelectionModal.tsx
@@ -31,13 +31,13 @@ const PolicyTypeSelectionModal: FC<Props> = ({ isOpen, onClose, onSelect }) => {
       content={
         <div className="flex flex-col gap-4 pt-2 md:flex-row md:gap-8">
           <PolicyTypeCard
-            icon={<RefreshIcon className="size-16" />}
+            icon={<RefreshIcon className="size-16" stroke="currentColor" />}
             title={translateText(["basicInfo", "accrualTitle"])}
             description={translateText(["basicInfo", "accrualDescription"])}
             onSelect={() => onSelect(PolicyType.ACCRUAL)}
           />
           <PolicyTypeCard
-            icon={<CalendarIcon className="size-16"/>}
+            icon={<CalendarIcon className="size-16" />}
             title={translateText(["basicInfo", "flexibleTitle"])}
             description={translateText(["basicInfo", "flexibleDescription"])}
             onSelect={() => onSelect(PolicyType.FLEXIBLE)}


ThinuwanW (commented):
## 🤖 Claude Code Review

This is a small cosmetic PR: it adds a trailing period to the `accrualDescription` i18n string, switches the PolicyTypeCard icon wrapper from `text-primary-accent` to `text-black`, and adds `stroke="currentColor"` to `RefreshIcon` (plus a Prettier whitespace fix on `CalendarIcon`). The i18n change is safe — English is the only locale directory, the key is used in exactly one place, and the JSON structure is unchanged. The main concerns are the asymmetric icon-coloring mechanism (only one of the two icons gets an explicit `stroke` override, so the two cards may not render the same color) and pre-existing accessibility/hover-styling gaps in the card this PR is polishing.

Found **13** new issue(s): 🔴 1 important, 🟡 5 suggestion(s), 🟣 7 nit(s)

frontend/src/community/leave/components/molecules/PolicyTypeSelectionModal/PolicyTypeSelectionModal.tsx
```diff
@@ -31,13 +31,13 @@ const PolicyTypeSelectionModal: FC<Props> = ({ isOpen, onClose, onSelect }) => {
       content={
         <div className="flex flex-col gap-4 pt-2 md:flex-row md:gap-8">
           <PolicyTypeCard
-            icon={<RefreshIcon className="size-16" />}
+            icon={<RefreshIcon className="size-16" stroke="currentColor" />}
```
ThinuwanW, line 34: 🤖 **Claude** · 🟡 **Suggestion**

Inconsistent icon-coloring mechanism between the two cards: `RefreshIcon` gets an explicit `stroke="currentColor"` while `CalendarIcon` relies purely on inheriting the wrapper's `text-black`. Please verify visually that both icons actually render black — if `RefreshIcon` needed the override because it hardcodes a stroke color internally, `CalendarIcon` may hardcode its color too (or be fill-based and need `fill="currentColor"`), which would leave the two cards visually mismatched after this change. Two further points: (1) note that `stroke` on the root `<svg>` only takes effect if the icon's inner `<path>` elements don't set their own `stroke` attribute — worth confirming rather than assuming; (2) every other skapp-ui icon call site in this repo colors icons purely via Tailwind text utilities (e.g. `<SearchIcon className="size-4 text-secondary-icon" />`, `<CloseIcon className="size-4 text-black" />` in LeavePolicyWizard.tsx:183) and never passes a `stroke` prop. The durable fix is to make `RefreshIcon` in `@rootcodelabs/skapp-ui` default to `currentColor` so all call sites benefit; patching it at this one call site means the next consumer hits the same surprise. Also confirm `stroke` is part of `RefreshIcon`'s exported prop type (i.e. it extends `SVGProps<SVGSVGElement>`) so this compiles under TS strict — I could not verify this locally because node_modules is not installed in this checkout.

frontend/src/community/leave/components/molecules/PolicyTypeSelectionModal/PolicyTypeCard.tsx
```diff
@@ -22,7 +22,7 @@ const PolicyTypeCard: FC<Props> = ({ icon, title, description, onSelect }) => (
     className="flex-1 cursor-pointer rounded-xl aspect-133.75/109.5 bg-white hover:shadow-md hover:outline-primary-accent"
   >
     <div className="flex h-full flex-col items-center justify-center gap-8 text-center">
-      <span className="text-primary-accent" aria-hidden="true">
+      <span className="text-black" aria-hidden="true">
```
ThinuwanW, line 25: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

The card sets `role="button"` and an `onKeyDown` handler for Enter/Space, but never sets `tabIndex={0}`. Unless the skapp-ui `Card` internally adds a tabindex when a `role` is passed, the card is not keyboard-focusable, so the `onKeyDown` handler is unreachable dead code and keyboard-only users cannot select a policy type at all. The established pattern elsewhere in this repo pairs the two explicitly — see `src/enterprise/sign/components/molecules/AddRecipientRow/RecipientRowDetails.tsx:352-353` and `TemplateRoleRowDetails.tsx:109-110`, both of which use `role="button"` + `tabIndex={0}` + `onKeyDown`. Fix: add `tabIndex={0}` to the `Card` props (and see the related note about adding a `focus-visible` outline so the focused card is visible). Raising this because the PR is deliberately polishing this card's presentation/a11y layer.

frontend/src/community/leave/components/molecules/PolicyTypeSelectionModal/PolicyTypeCard.tsx
```diff
@@ -22,7 +22,7 @@ const PolicyTypeCard: FC<Props> = ({ icon, title, description, onSelect }) => (
     className="flex-1 cursor-pointer rounded-xl aspect-133.75/109.5 bg-white hover:shadow-md hover:outline-primary-accent"
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`hover:outline-primary-accent` only sets `outline-color`; it does not set an outline width or style, so with Tailwind 4 the browser default `outline-style: none` applies and the intended hover outline almost certainly never renders (the `hover:shadow-md` alone provides the hover affordance). The working pattern in this codebase pairs a width with the color — see `src/community/configurations/components/molecules/WorkLocationEmployeeSelector/WorkLocationEmployeeSelector.tsx:314`: `focus:outline-1 focus:outline-primary-accent focus:-outline-offset-[2px]`. Fix: change to `hover:outline-1 hover:outline-primary-accent hover:-outline-offset-[2px]`, and consider mirroring it as `focus-visible:` so the keyboard-focused card is visible too. Pre-existing, but it's on the card whose styling this PR is correcting, and it means one of the two declared hover states is silently inert.

frontend/src/community/leave/components/molecules/PolicyTypeSelectionModal/PolicyTypeCard.tsx
ThinuwanW, line 25: 🤖 **Claude** · 🟣 **Nit**

This swaps a theme token (`text-primary-accent`, mapped to `--color-primary-accent` in `styles/global.css`) for the fixed Tailwind `black`, which is not defined in the `@theme` block and therefore won't follow any future brand-color or surface-theming change. It is internally consistent with line 29 (`subtitle2 text-black`) and `text-black` is used elsewhere in the repo, so this is acceptable as-is — but if a token like `text-primary-text` matches the design intent, prefer it over a fixed color.

frontend/src/community/common/assets/languages/english/leaveModule.json
```diff
@@ -367,7 +367,7 @@
       "basicInfo": {
         "policyTypeTitle": "Policy type",
         "accrualTitle": "Accrual",
-        "accrualDescription": "Time taken is deducted from a balance that accrues on a schedule or is set manually",
+        "accrualDescription": "Time taken is deducted from a balance that accrues on a schedule or is set manually.",
```
ThinuwanW, line 370: 🤖 **Claude** · 🟣 **Nit**

The punctuation fix is correct and now matches `flexibleDescription`, but the two sibling strings still disagree in voice: `accrualDescription` is neutral third-person ("Time taken is deducted from...") while `flexibleDescription` switches to first person ("...but I still want leave requests to be approved and tracked."). Since this PR is already normalizing this copy block, consider rewording the flexible description to match, e.g. "There is no balance to deduct from, but leave requests are still approved and tracked." Also confirmed no action needed on other locales — `assets/languages/` contains only `english`, and this key is consumed in exactly one place (PolicyTypeSelectionModal.tsx:36).

frontend/src/community/leave/components/molecules/PolicyTypeSelectionModal/PolicyTypeCard.tsx
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The card is interactive (role="button", onClick, onKeyDown) but the className only defines hover-state affordances (`hover:shadow-md hover:outline-primary-accent`) — there is no `focus`/`focus-visible` variant anywhere. Combined with the fact that skapp-ui's `Card` renders a plain div (confirmed by `PeopleSyncOptionCard` and `DurationOptionCard` both having to set `tabIndex` manually), a keyboard user tabbing to this card gets no visible focus indicator, which is a WCAG 2.4.7 failure. This is distinct from the hover-outline-width problem: even with a correct hover outline, focus is still unstyled. The codebase already has the right pattern at WorkLocationEmployeeSelector.tsx:314 — mirror it: add `focus:outline-1 focus:outline-primary-accent focus:-outline-offset-[2px]` (or `focus-visible:` equivalents) to the card className.

frontend/src/community/leave/components/molecules/PolicyTypeSelectionModal/PolicyTypeSelectionModal.tsx
ThinuwanW, line 34: 🤖 **Claude** · 🟣 **Nit**

`stroke="currentColor"` is the only place in the entire frontend that passes a `stroke` prop to an icon — every other color override on an icon (30+ call sites, e.g. Icon name=SEND_ICON fill="currentColor", CloseIcon fill="var(--color-primary-text)", DeleteButtonIcon fill=...) uses `fill`. Two things worth confirming before merge, since `@rootcodelabs/skapp-ui` is pinned to a beta (1.1.1-beta.10) and its types can't be checked from this repo: (a) that `RefreshIcon`'s prop type actually declares `stroke` — if it's a narrow custom prop type rather than `SVGProps<SVGSVGElement>`, this is a strict-mode TS compile error; and (b) that the prop lands where it needs to — a `stroke` on the root `<svg>` does not override child `<path stroke="…">` attributes, so if the SVG hardcodes stroke on its paths this is a silent no-op. Please verify by rendering both cards side by side; if `fill` is the supported prop, switch to it for consistency with the rest of the codebase.

frontend/src/community/leave/components/molecules/PolicyTypeSelectionModal/PolicyTypeCard.tsx
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`aspect-133.75/109.5` is the only `aspect-*` utility used anywhere in the frontend, and it uses a bare fraction with decimal operands, which is outside Tailwind v4's documented `aspect-<ratio>` bare-value grammar. If the class fails to generate, it fails silently — the card just loses its aspect ratio and collapses to content height, with no build error to catch it. Two fixes: use the explicit arbitrary-value form `aspect-[133.75/109.5]`, which is unambiguous; and consider replacing the raw design-tool pixel numbers with a named ratio or a plain `min-h-*`, since 133.75/109.5 is a magic number that no future reader can interpret. Please verify the generated CSS actually contains an `aspect-ratio` declaration for this card.

frontend/src/community/leave/components/molecules/PolicyTypeSelectionModal/PolicyTypeCard.tsx
ThinuwanW, line 25: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

`PolicyTypeCard` is near-identical to `src/enterprise/people/components/atoms/PeopleSyncOptionCard/PeopleSyncOptionCard.tsx`: both are a skapp-ui `Card` with `role="button"`, an Enter/Space `onKeyDown` handler, a centered 16-unit icon slot, a `subtitle*` title and a `body* text-secondary-text` description, `rounded-xl bg-white`, `cursor-pointer`, and `gap-8`/`gap-4` spacing. Per the duplication guideline (2+ components with near-identical markup → extract a shared component), this should be one component under `~community/common/components/molecules/` (community, so enterprise can import it) with optional `disabled`/badge/tooltip props. Notably, the enterprise version already gets `tabIndex` and disabled handling right, so extracting would also fix the a11y gaps here rather than leaving two copies to drift.

frontend/src/community/leave/components/molecules/PolicyTypeSelectionModal/PolicyTypeSelectionModal.tsx
ThinuwanW, line 34: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

`closeButtonAriaLabel={translateText(["closeBtnAriaLabel"])}` (line 30) resolves to "Close create policy" (leaveModule.json:357), but this is the policy-*type-selection* modal, not the create-policy wizard — screen-reader users are told they're closing something they haven't opened yet. The same key is also consumed by LeavePolicyWizard.tsx:187 for the wizard's own close button, so the two modals are coupled: rewording the label for one silently changes the other. Suggest adding a dedicated key such as `createPolicy.policyTypeCloseBtnAriaLabel` ("Close policy type selection") and pointing this modal at it.

frontend/src/community/leave/components/molecules/PolicyTypeSelectionModal/PolicyTypeCard.tsx
```diff
@@ -22,7 +22,7 @@ const PolicyTypeCard: FC<Props> = ({ icon, title, description, onSelect }) => (
...
   >
     <div className="flex h-full flex-col items-center justify-center gap-8 text-center">
-      <span className="text-primary-accent" aria-hidden="true">
+      <span className="text-black" aria-hidden="true">
         {icon}
       </span>
       <span className="flex flex-col gap-4">
```
ThinuwanW, line 28: 🤖 **Claude** · 🟣 **Nit**

The title and description are rendered as nested `<span>` elements inside another `<span>` that is styled `flex flex-col`. These are block-level text runs, not inline phrasing content, and the analogous `PeopleSyncOptionCard` uses `<p>` for both. Using `<div>`/`<p>` here would be more semantically accurate and avoids relying on Tailwind's `display: flex` to override the inline default. Purely cosmetic in rendering terms, so treat as optional.

frontend/src/community/leave/components/molecules/PolicyTypeSelectionModal/PolicyTypeCard.tsx
ThinuwanW, line 25: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

Neither `PolicyTypeCard.tsx` nor `PolicyTypeSelectionModal.tsx` has a colocated `.test.tsx`, even though the repo's convention is colocated component tests (e.g. AttendanceCard.test.tsx, DescribedSelection.test.tsx, IconButton.test.tsx). Caveat: the entire `src/community/leave/components` tree currently has zero component tests — only util tests — so this is a module-wide gap rather than something this PR introduced. Given the card carries hand-rolled keyboard-activation logic that the guidelines flag as a common defect area, a small test asserting `screen.getByRole("button", { name: title })` is reachable and that Enter/Space call `onSelect` exactly once would be cheap insurance.

frontend/src/community/common/assets/languages/english/leaveModule.json
hasalarootcode: Not needed

frontend/src/community/leave/components/molecules/PolicyTypeSelectionModal/PolicyTypeSelectionModal.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/molecules/PolicyTypeSelectionModal/PolicyTypeCard.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/molecules/PolicyTypeSelectionModal/PolicyTypeCard.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/molecules/PolicyTypeSelectionModal/PolicyTypeCard.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/molecules/PolicyTypeSelectionModal/PolicyTypeCard.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/molecules/PolicyTypeSelectionModal/PolicyTypeSelectionModal.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/molecules/PolicyTypeSelectionModal/PolicyTypeCard.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/molecules/PolicyTypeSelectionModal/PolicyTypeCard.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/molecules/PolicyTypeSelectionModal/PolicyTypeSelectionModal.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/molecules/PolicyTypeSelectionModal/PolicyTypeCard.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/molecules/PolicyTypeSelectionModal/PolicyTypeCard.tsx
hasalarootcode: Not needed

shakila-kamalasena (approved):

thusala (approved):

akilarootcode (approved):

PR #1771, opened 2026-08-17
Fix/Nav bar notification count badge doesn't update when the underlying action is completed elsewhere

umeshaliyanage:
### Summary

The fix wraps that page in NotificationReadProvider flipping the is leave notifocation read to true . Now opening it sends the same message, and the count drops.

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR wraps the pending-leave page in `NotificationReadProvider` so visiting it marks LEAVE_REQUEST notifications as read — consistent with the existing pattern in `leave-requests.tsx` and `all-timesheets.tsx`. However, the change appears to be the product of a botched merge resolution: the JSX is left unbalanced with a stray `</ContentLayout>`, which will fail the build, and the page's content component was reverted from `PendingLeaveRequestsSection` back to `PendingLeaveRequestTable`, silently dropping the leave-policies branching that was on the base branch. Both must be fixed before merge.

Found **7** new issue(s): 🔴 4 important, 🟡 1 suggestion(s), 🟣 2 nit(s)

frontend/pages/community/leave/pending-leave.tsx
```diff
@@ -18,31 +21,36 @@ const PendingLeave: NextPage = () => {
...
+              value={searchTerm}
+              setSearchTerm={setSearchTerm}
+              placeHolder={translateText(["searchBoxPlaceholder"])}
+            />
+          </Box>
+          <PendingLeaveRequestTable searchTerm={searchTerm} />
+        </>
+      </ContentLayout>
+    </NotificationReadProvider>
     </ContentLayout>
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Unbalanced JSX — this file will not compile. The new tree closes with `</ContentLayout>` (line 52) then `</NotificationReadProvider>` (line 53), but the original `</ContentLayout>` on line 54 was left behind, leaving a closing tag with no matching opening tag. TypeScript/SWC will fail with an unexpected-closing-tag parse error and the whole page build breaks. Fix: delete line 54 so the return ends with `</NotificationReadProvider>` followed by `);`. This looks like leftover conflict debris from the `Merge branch 'develop'` commit — worth re-checking the rest of that merge resolution too (see the separate finding on line 50).

frontend/pages/community/leave/pending-leave.tsx
```diff
@@ -18,31 +21,36 @@ const PendingLeave: NextPage = () => {
...
+      >
+        <>
+          <Box mb={2}>
+            <SearchBox
+              value={searchTerm}
+              setSearchTerm={setSearchTerm}
+              placeHolder={translateText(["searchBoxPlaceholder"])}
+            />
+          </Box>
+          <PendingLeaveRequestTable searchTerm={searchTerm} />
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Functional regression: the page's content was changed from `<PendingLeaveRequestsSection searchTerm={searchTerm} />` (base) to `<PendingLeaveRequestTable searchTerm={searchTerm} />`. `PendingLeaveRequestsSection` (src/community/leave/components/organisms/PendingLeaveRequestsSection/PendingLeaveRequestsSection.tsx) is not a rename — it adds behaviour that is now lost: (1) it renders `<TableSkeleton rows={LEAVE_REQUESTS_SKELETON_ROWS} />` while `useLeavePoliciesEnabled()` is loading, so the page no longer has a loading state and will flash the legacy empty table; (2) when `isLeavePoliciesEnabled` is true it renders `PolicyPendingLeaveRequests`, so tenants with leave policies enabled will now see the old non-policy table instead of the policy-aware one. After this change `PendingLeaveRequestsSection` and `PolicyPendingLeaveRequests` have zero remaining references in the codebase (verified by grep) — i.e. a whole feature path became dead code, which is a strong signal this was an accidental revert during the develop merge rather than an intentional change. Also note the import of `PendingLeaveRequestsSection` on line 13 is now unused (dead import / lint failure). Fix: restore `<PendingLeaveRequestsSection searchTerm={searchTerm} />` here, keep the line 13 import, and drop the now-unnecessary `PendingLeaveRequestTable` import on line 12 — this PR should only be adding the `NotificationReadProvider` wrapper, nothing else.

frontend/pages/community/leave/pending-leave.tsx
```diff
@@ -18,31 +21,36 @@ const PendingLeave: NextPage = () => {
...
-        {
-          label: translateBreadcrumbText(["leaveRequests.title"])
-        }
-      ]}
-      pageHead={translateText(["pageHead"])}
-      title={translateText(["title"])}
-      isDividerVisible={true}
-      isBackButtonVisible={true}
-      onBackClick={() => router.replace(ROUTES.DASHBOARD.BASE)}
+    <NotificationReadProvider
```
ThinuwanW, line 23: 🤖 **Claude** · 🔴 **Important**

Mounting NotificationReadProvider here leaves the notification bell out of sync. `markNotificationTypeAsRead` in NotificationServiceImpl (backend, ~line 326) calls `markTypeViewedByUserIdAndTypes`, which sets `isViewed = true` on the actual Notification rows — so this page's mount permanently changes the user's unread notification state, not just a separate summary counter. But `useMarkNotificationSummaryAsRead` (frontend/src/community/common/api/notificationsApi.ts:147-166) invalidates only `notificationsQueryKeys.GET_NOTIFICATION_SUMMARY`. The bell badge renders `notifyData.unreadCount` (AppBar.tsx:129), which is hydrated from `useGetUnreadNotificationsCount()` → `GET_NOTIFICATIONS_COUNT` (AppBar.tsx:76-86), and the dropdown list uses `GET_NOTIFICATIONS`. Neither key is invalidated, and the Zustand `notifyData.unreadCount` is never adjusted — contrast `useMarkAllNotificationsAsRead` (resets the store + `resetUnreadCount()`) and `useMarkNotificationAsRead` (decrements the store and rewrites the `unReadMsgCount` localStorage key used in enterprise). Result: after visiting this page the bell keeps showing the pre-read count (and enterprise keeps a stale `unReadMsgCount`) until a hard reload, and reopening the list shows items as unread that the server now considers read. Fix in `useMarkNotificationSummaryAsRead.onSuccess`: also `invalidateQueries` for `notificationsQueryKeys.GET_NOTIFICATIONS` and `GET_NOTIFICATIONS_COUNT`, and reconcile `notifyData.unreadCount` in `useCommonStore` the same way the other two mark-as-read mutations do. This affects the two pre-existing call sites too, but this PR adds a third and makes it reachable from the dashboard widget.

frontend/pages/community/leave/pending-leave.tsx
```diff
@@ -7,6 +7,9 @@ import SearchBox from "~community/common/components/molecules/SearchBox/SearchBo
 import ContentLayout from "~community/common/components/templates/ContentLayout/ContentLayout";
 import ROUTES from "~community/common/constants/routes";
 import { useTranslator } from "~community/common/hooks/useTranslator";
+import NotificationReadProvider from "~community/common/providers/NotificationReadProvider";
+import { NotificationSummaryType } from "~community/common/types/notificationTypes";
+import PendingLeaveRequestTable from "~community/leave/components/molecules/PendingLeaveRequestTable/PendingLeaveRequestTable";
 import PendingLeaveRequestsSection from "~community/leave/components/organisms/PendingLeaveRequestsSection/PendingLeaveRequestsSection";
```
ThinuwanW, line 12: 🤖 **Claude** · 🔴 **Important**

The `PendingLeaveRequestsSection` import is now unreferenced. `unused-imports/no-unused-imports` is configured as "error" in frontend/.eslintrc.json:42, so lint/CI fails on this file independently of the JSX problem. Worth noting when fixing: a repo-wide grep shows this page was the *only* consumer of `PendingLeaveRequestsSection`, and that section is the only consumer of `PolicyPendingLeaveRequests` and of the `useLeavePoliciesEnabled` + `TableSkeleton` gate — so dropping it orphans that whole subtree app-wide (no route can reach the leave-policy pending-requests UI anymore). Restore the section as the rendered child rather than just deleting the import.

frontend/pages/community/leave/pending-leave.tsx
```diff
@@ -18,31 +21,36 @@ const PendingLeave: NextPage = () => {
...
-          label: translateBreadcrumbText(["leaveRequests.title"])
-        }
-      ]}
-      pageHead={translateText(["pageHead"])}
-      title={translateText(["title"])}
-      isDividerVisible={true}
-      isBackButtonVisible={true}
-      onBackClick={() => router.replace(ROUTES.DASHBOARD.BASE)}
+    <NotificationReadProvider
+      notificationType={NotificationSummaryType.LEAVE_REQUEST}
```
ThinuwanW, line 24: 🤖 **Claude** · 🟡 **Suggestion**

The LEAVE_REQUEST badge this clears is not attached to this route. getDrawerRoutes.ts:339-345 attaches `notificationLeaveCount` only to sub-route id "2B" — "All Requests" / `ROUTES.LEAVE.LEAVE_REQUESTS` (see src/community/common/utils/data/routes.ts:111-114) — and `/leave/pending-leave` has no drawer entry at all, so the only way in is the dashboard pending-requests widget. As written, landing here from the dashboard silently zeroes a badge on a nav item the user never opened, and the list shown here comes from the widget feed (`/leave/pending-requests`, rendered with pagination disabled), so the badge can be cleared before the user has seen everything it counted. If clearing from this page is deliberate, please state that in the PR description and consider also badging this route; otherwise keep the read-marking on the badged "All Requests" page, which already wraps itself in this provider (pages/community/leave/leave-requests.tsx:22).

frontend/pages/community/leave/pending-leave.tsx
ThinuwanW, line 23: 🤖 **Claude** · 🟣 **Nit**

The mark-as-read is fire-and-forget and unconditional: the provider's `useEffect` fires on mount before the pending-request query resolves, so it also fires when the list is empty or the fetch fails, and on failure the mutation only `console.error`s (notificationsApi.ts:162-164) with no retry and no user-visible feedback — the badge just silently stays. Consider gating the mutation on a successful load of the pending requests (or at minimum setting `retry` on the mutation) so the persisted read state reflects what the user actually saw. Flagging as a nit because a toast would be wrong for a background side effect; the timing/retry aspect is the substantive part.

frontend/pages/community/leave/pending-leave.tsx
```diff
@@ -7,6 +7,9 @@ import SearchBox from "~community/common/components/molecules/SearchBox/SearchBo
 import ContentLayout from "~community/common/components/templates/ContentLayout/ContentLayout";
 import ROUTES from "~community/common/constants/routes";
 import { useTranslator } from "~community/common/hooks/useTranslator";
+import NotificationReadProvider from "~community/common/providers/NotificationReadProvider";
```
ThinuwanW, line 10: 🤖 **Claude** · 🟣 **Nit**

Using a Provider for this is more machinery than the job needs: `NotificationReadContext` in NotificationReadProvider.tsx is created with a `null` value, is never exported, and has no consumer or `useX` accessor — the component exists purely to run a mount effect, so it adds a context layer around the entire page for no data flow. A small hook (e.g. `useMarkNotificationSummaryAsReadOnMount(NotificationSummaryType.LEAVE_REQUEST)`) called in the page body would express the intent directly and remove the extra tree level. Acknowledging this is an established pattern already used by two other pages, so this is only worth doing as a follow-up cleanup across all call sites, not in this PR.

frontend/pages/community/leave/pending-leave.tsx
umeshaliyanage: removed the extra content layout

frontend/pages/community/leave/pending-leave.tsx
umeshaliyanage: issues came when merging fixed, and only added the NotificationReadProvider wrapper only

frontend/pages/community/leave/pending-leave.tsx
umeshaliyanage: restored

frontend/pages/community/leave/pending-leave.tsx
umeshaliyanage: markTypeViewedByUserIdAndTypes doesn't touch isViewed, it sets a different column Check NotificationRepositiryImpl
 
There are two independent flags Notification, is_viewed drives the bell badge, is_type_viewed drives the summary. Disjoint columns, so mounting the provider only moves the summary, invalidating just GET_NOTIFICATION_SUMMARY is the right scope

frontend/pages/community/leave/pending-leave.tsx
umeshaliyanage: Deliberate, and it's the point of this PR. The nav badge wasn't clearing when requests were handled via the dashboard pending-requests widget, which routes to /leave/pending-leave rather than All Requests so the count stuck around after the user had already dealt with everything.

frontend/pages/community/leave/pending-leave.tsx
umeshaliyanage: out of scope for the bug i fixed

frontend/pages/community/leave/pending-leave.tsx
umeshaliyanage: out of scope

Deshan1100 (approved):

akilarootcode (approved):

PR #1772, opened 2026-08-17
Fix/2947: Leave policy - mandatory fields missing red asterisk in create policy flow

hasalarootcode:
…eave Policy Wizard

### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/frontend/src/community/leave/components/organisms/LeavePolicyWizard/BasicInfoStep.tsx
+++ b/frontend/src/community/leave/components/organisms/LeavePolicyWizard/BasicInfoStep.tsx
@@ -58,6 +58,7 @@ const BasicInfoStep: FC<Props> = ({ formData, onChange, errors, touched }) => {
             label={translateText(["basicInfo", "policyNameLabel"])}
             name="policyName"
             type="text"
+            required
             value={formData.policyName}
             placeholder={translateText(["basicInfo", "policyNamePlaceholder"])}
             state={policyNameError ? "error" : "default"}
@@ -71,6 +72,7 @@ const BasicInfoStep: FC<Props> = ({ formData, onChange, errors, touched }) => {
             <Dropdown
               id="leave-policy-leave-type"
               label={translateText(["basicInfo", "leaveTypeLabel"])}
+              required
               value={formData.leaveType}
               placeholder={translateText(["basicInfo", "leaveTypePlaceholder"])}
               options={leaveTypeOptions}


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds the `required` prop to the policy name `InputField` and leave type `Dropdown` in the leave policy wizard's Basic Info step. I verified `required` is a valid, established prop on both `@rootcodelabs/skapp-ui` `InputField` and `Dropdown` (same usage in `DealNameStageSection.tsx` and `TaskModalForm.tsx`), that the labels in `leaveModule.json` do not already contain an asterisk (no double marker), and that `LeavePolicyWizard` renders no `<form>` element so a native `required` attribute cannot interfere with the button-driven step navigation. The change is correct and low risk; my only concern is that it leaves the required-field marking inconsistent across the rest of the wizard and the edit modal.

Found **8** new issue(s): 🔴 1 important, 🟡 3 suggestion(s), 🟣 4 nit(s)

frontend/src/community/leave/components/organisms/LeavePolicyWizard/BasicInfoStep.tsx
```diff
@@ -58,6 +58,7 @@ const BasicInfoStep: FC<Props> = ({ formData, onChange, errors, touched }) => {
             label={translateText(["basicInfo", "policyNameLabel"])}
             name="policyName"
             type="text"
+            required
```
ThinuwanW, line 61: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

The required markers are now applied only to step 1, which makes the rest of the leave-policy flow read as optional when it isn't. In the same schema (`leavePolicyWizardValidation` in `~community/leave/utils/validations.ts`), `accrualDays` and `accrualFrequency` are `.required()` whenever `isAccrual` is true, and `waitingPeriodDays` / `accrualCapDays` are effectively required via their `.when(...)` tests once the corresponding toggle is on — yet none of those fields in `EntitlementSetupStep.tsx` carry `required`. `EditLeavePolicyModal.tsx` has the same gap: its `policyName` field is `.required()` in `editLeavePolicyValidation` but renders without the marker. Partial asterisks are worse than none, because users infer that unmarked fields are optional and then get blocked by validation on Next. Suggest adding `required` to the `accrualDays` InputField and `accrualFrequency` Dropdown in `EntitlementSetupStep.tsx` (and to the conditionally-rendered `waitingPeriodDays` / `accrualCapDays` inputs, which are only rendered when their toggle makes them mandatory), plus the `policyName` InputField in `EditLeavePolicyModal.tsx`, so the whole flow is consistent.

frontend/src/community/leave/components/organisms/LeavePolicyWizard/BasicInfoStep.tsx
```diff
@@ -71,6 +72,7 @@ const BasicInfoStep: FC<Props> = ({ formData, onChange, errors, touched }) => {
             <Dropdown
               id="leave-policy-leave-type"
               label={translateText(["basicInfo", "leaveTypeLabel"])}
+              required
```
ThinuwanW, line 75: 🤖 **Claude** · 🟣 **Nit**

Minor UX edge case: the leave type Dropdown is now unconditionally marked required, including in the empty-catalogue state where `leaveTypeOptions.length === 0` and the `noLeaveTypesWarning` alert is shown. The user then sees a mandatory-field marker on a control with nothing selectable. Not a regression in behaviour (validation already blocked this case), but you may want to consider hiding the dropdown or disabling Next in that state so the mandatory marker is always satisfiable. Flagging as a nit — your call whether it's worth handling here.

ThinuwanW, line 75: 🤖 **Claude** · 🔴 **Important**

Query error is silently rendered as an empty catalogue, and marking the field `required` makes that worse. `useGetPolicyLeaveTypes` (src/community/leave/api/PolicyLeaveTypeApi.ts) exposes no error handling and the component only destructures `data` and `isLoading`. If the request fails, React Query v5 sets `isLoading` to false with `data` undefined, so `leaveTypeOptions` is `[]` and the branch at line 84 renders "No leave types available. Please configure leave types before creating a policy." The user is now told a field is mandatory, given zero options, and blamed for a configuration problem that is actually a network/server failure — and `Next` will keep failing validation with "Please select a leave type." with no way forward. Fix: also destructure `isError` (and `refetch`) from the hook, gate the existing warning on `!isError`, and render a distinct error state with a retry action (plus a new `leaveTypesFetchError` key in leaveModule.json under `createPolicy`), or surface a toast via `useToast` on failure as the guidelines require for failed data loads.

frontend/src/community/leave/components/organisms/LeavePolicyWizard/BasicInfoStep.tsx
```diff
@@ -71,6 +72,7 @@ const BasicInfoStep: FC<Props> = ({ formData, onChange, errors, touched }) => {
             <Dropdown
               id="leave-policy-leave-type"
               label={translateText(["basicInfo", "leaveTypeLabel"])}
```
ThinuwanW, line 74: 🤖 **Claude** · 🟡 **Suggestion**

No loading affordance on the now-required dropdown. `isLoading` is consumed only to suppress the empty-state warning (line 84); the `Dropdown` itself is rendered enabled with `options={[]}` while the request is in flight, so a user who lands on step 1 sees a field flagged as mandatory that opens to nothing, with no indication that data is still arriving. The sibling form in the same module handles this properly (src/community/leave/components/organisms/LeavePolicyForm/../PolicyLeaveTypeForm/PolicyLeaveTypeForm.tsx:222 renders `ColorPaletteSkeleton` while `isPolicyLeaveTypeLoading`). Fix: pass a disabled/loading state to the `Dropdown` while `isLoading` (e.g. `disabled={isLoading}` or a skeleton wrapper), so the required marker only becomes actionable once the options exist.

frontend/src/community/leave/components/organisms/LeavePolicyWizard/BasicInfoStep.tsx
ThinuwanW, line 61: 🤖 **Claude** · 🟡 **Suggestion**

The PR surfaces the "this field is mandatory" constraint but leaves the length constraint on the same field invisible. `leavePolicyWizardValidation` (src/community/leave/utils/validations.ts:155-161) enforces `.max(MAX_POLICY_NAME_LENGTH)` (100) with the `policyNameMaxLength` message, yet this `InputField` has no `maxLength`, so a user can type 300 characters and only discover the problem after clicking Next. The equivalent field in the same module does cap input (PolicyLeaveTypeForm.tsx:202 passes `maxLength={MAX_POLICY_LEAVE_TYPE_NAME_LENGTH}` alongside its `required`). Fix: import `MAX_POLICY_NAME_LENGTH` from `~community/leave/constants/leavePolicyConstants` and add `maxLength={MAX_POLICY_NAME_LENGTH}` next to the new `required` prop.

frontend/src/community/leave/components/organisms/LeavePolicyWizard/BasicInfoStep.tsx
```diff
@@ -71,6 +72,7 @@ const BasicInfoStep: FC<Props> = ({ formData, onChange, errors, touched }) => {
...
+              required
               value={formData.leaveType}
```
ThinuwanW, line 76: 🤖 **Claude** · 🟣 **Nit**

Accessibility gap opened by the new `required` prop: the explanatory message rendered below the dropdown (the `<p role="alert">` with `noLeaveTypesWarning`) is a sibling `<p>` with no `id` and no `aria-describedby` link back to `leave-policy-leave-type`. A screen-reader user focused on the control now hears that it is required but never hears why it cannot be satisfied. Fix: give the paragraph a stable `id` (e.g. `leave-policy-leave-type-warning`) and pass `aria-describedby` to the `Dropdown` when the warning is visible. Separately, `role="alert"` is assertive and interrupts the user for what is a static advisory — `role="status"` (polite) is the better fit here.

frontend/src/community/leave/components/organisms/LeavePolicyWizard/BasicInfoStep.tsx
ThinuwanW, line 61: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

Performance nit in the component this PR touches: `leaveTypeOptions` (line 33) is rebuilt and `handleLeaveTypeChange` (line 44) is re-created on every render. Because the parent wizard calls `formik.setValues` on every keystroke in the policy name field (LeavePolicyWizard.tsx:128-130), the whole step re-renders per character, re-mapping the leave type list and re-running `getEmoji` for every entry. Wrap the mapping in `useMemo` keyed on `policyLeaveTypes?.items` and the handler in `useCallback` keyed on `[onChange, leaveTypeOptions]`, per the performance guideline on memoizing computations and callbacks passed to children.

ThinuwanW, line 61: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

No test coverage exists for the entire `LeavePolicyWizard` folder — there is no `.test.tsx` for `BasicInfoStep`, `EntitlementSetupStep`, or `LeavePolicyWizard`, so this required-field change ships unverified and the empty/loading/error branches of the leave type dropdown are untested. Flagging as a nit since the components predate this PR, but a small `BasicInfoStep.test.tsx` (rendered with the `MockTheme` wrapper, asserting via `screen.getByRole("textbox", { name: /policy name/i })` that the fields expose their required state, plus a case where `useGetPolicyLeaveTypes` returns an empty list) would lock in the behaviour this PR is introducing.

hasalarootcode: Not needed

frontend/src/community/leave/components/organisms/LeavePolicyWizard/BasicInfoStep.tsx
hasalarootcode: Not needed

hasalarootcode: Not needed

frontend/src/community/leave/components/organisms/LeavePolicyWizard/BasicInfoStep.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/organisms/LeavePolicyWizard/BasicInfoStep.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/organisms/LeavePolicyWizard/BasicInfoStep.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/organisms/LeavePolicyWizard/BasicInfoStep.tsx
hasalarootcode: Not needed

hasalarootcode: Not needed

shakila-kamalasena (approved):

thusala (approved):

akilarootcode (approved):

PR #1773, opened 2026-08-17
Fix/3010: Leave Policy - Duplicate policy name error shows as a toast instead of an inline field error

hasalarootcode:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/frontend/src/community/leave/components/organisms/LeavePolicyWizard/LeavePolicyWizard.tsx
+++ b/frontend/src/community/leave/components/organisms/LeavePolicyWizard/LeavePolicyWizard.tsx
@@ -6,7 +6,7 @@ import {
   IconButton
 } from "@rootcodelabs/skapp-ui";
 import { AxiosError } from "axios";
-import { useFormik } from "formik";
+import { FormikProps, useFormik } from "formik";
 import { useRouter } from "next/router";
 import { FC, useRef, useState } from "react";
 
@@ -24,6 +24,7 @@ import {
 } from "~community/leave/types/LeavePolicyTypes";
 import {
   getLeavePolicyErrorToastKeys,
+  isDuplicatePolicyNameError,
   mapLeavePolicyFormToPayload
 } from "~community/leave/utils/leavePolicy/leavePolicyUtils";
 import { leavePolicyWizardValidation } from "~community/leave/utils/validations";
@@ -73,6 +74,8 @@ const LeavePolicyWizard: FC<Props> = ({ policyType }) => {
 
   const submittedNameRef = useRef<string>("");
 
+  const formikRef = useRef<FormikProps<LeavePolicyFormData> | null>(null);
+
   const steps = [
     translateText(["steps", "basicInfo"]),
     translateText(["steps", "entitlementSetup"]),
@@ -100,6 +103,16 @@ const LeavePolicyWizard: FC<Props> = ({ policyType }) => {
   };
 
   const handleError = (error: AxiosError): void => {
+    if (isDuplicatePolicyNameError(error)) {
+      formikRef.current?.setFieldTouched("policyName", true, false);
+      formikRef.current?.setFieldError(
+        "policyName",
+        translateText(["duplicateToastDescription"])
+      );
+      setActiveStep(LeavePolicyWizardSteps.BASIC_INFO);
+      return;
+    }
+
     const { title, description } = getLeavePolicyErrorToastKeys(error);
 
     setToastMessage({
@@ -124,6 +137,8 @@ const LeavePolicyWizard: FC<Props> = ({ policyType }) => {
       addLeavePolicy(mapLeavePolicyFormToPayload(values));
     }
   });
+
+  formikRef.current = formik;
 
   const handleFieldsChange = (values: Partial<LeavePolicyFormData>): void => {
     formik.setValues({ ...formik.values, ...values });
--- a/frontend/src/community/leave/utils/leavePolicy/leavePolicyUtils.ts
+++ b/frontend/src/community/leave/utils/leavePolicy/leavePolicyUtils.ts
@@ -19,6 +19,10 @@ interface LeavePolicyErrorToastKeys {
   description: string;
 }
 
+export const isDuplicatePolicyNameError = (error: AxiosError): boolean =>
+  (error?.response?.data as LeavePolicyErrorData | undefined)?.results?.[0]
+    ?.messageKey === LEAVE_ERROR_LEAVE_POLICY_ALREADY_EXISTS;
+
 export const getLeavePolicyErrorToastKeys = (
   error: AxiosError
 ): LeavePolicyErrorToastKeys => {


ThinuwanW (commented):
## 🤖 Claude Code Review

The PR adds inline field-level surfacing of the backend duplicate-policy-name error in the create-policy wizard, replacing a toast. The detection helper is correct — it matches `results[0].messageKey` against the enum name that `ErrorResponse` serialises (`messageKey.name()`), and the backend does throw `LEAVE_ERROR_LEAVE_POLICY_ALREADY_EXISTS` on the create path. However, the manually-set Formik error is wiped by `handleNext`'s own `formik.validateForm()` call, so the error silently disappears the moment the user clicks Next without changing anything; the ref is also mutated during render, and the change leaves the edit-policy path and a11y announcement inconsistent.

Found **12** new issue(s): 🔴 2 important, 🟡 8 suggestion(s), 🟣 2 nit(s)

frontend/src/community/leave/components/organisms/LeavePolicyWizard/LeavePolicyWizard.tsx
```diff
@@ -100,6 +103,16 @@ const LeavePolicyWizard: FC<Props> = ({ policyType }) => {
   };
 
   const handleError = (error: AxiosError): void => {
+    if (isDuplicatePolicyNameError(error)) {
```
ThinuwanW, line 106: 🤖 **Claude** · 🔴 **Important**

The inline error set here is erased by the wizard's own navigation handler. `handleNext` calls `await formik.validateForm()`, and in Formik 2.4 `validateForm` dispatches `SET_ERRORS` with the freshly computed Yup results — which replaces the whole `errors` object, discarding the manually set `policyName` error (Yup has no knowledge of the server-side uniqueness constraint).

Repro (accrual policy): submit from Summary with a name that collides → server 409 → user is bounced to BASIC_INFO with the inline error visible → user clicks "Next" without editing anything → `validateForm()` returns no `policyName` error → `hasStepError` is false → the error vanishes from the UI and the wizard advances to step 2. The user gets no signal until they reach Summary and eat a second failed round-trip. The same wipe happens via `handleFieldsChange` → `formik.setValues` (validateOnChange defaults to true), so editing an unrelated field on the step also clears it.

Fix: don't store this in Formik's error slice at all — hold the rejected value in component state, e.g. `const [duplicateName, setDuplicateName] = useState<string | null>(null)`, set it to `submittedNameRef.current` in `handleError`, clear it in `handleFieldsChange` when `policyName` or `leaveType` changes, treat it as a step error in `handleNext` (`if (duplicateName && formik.values.policyName.trim() === duplicateName) return;`), and merge it into what is handed to the step content: `errors={{ ...formik.errors, ...(duplicateName ? { policyName: message } : {}) }}` plus the matching `touched` override. That both survives `validateForm()` and actually blocks progression until the name changes.

frontend/src/community/leave/components/organisms/LeavePolicyWizard/LeavePolicyWizard.tsx
```diff
@@ -125,6 +138,8 @@ const LeavePolicyWizard: FC<Props> = ({ policyType }) => {
     }
   });
 
+  formikRef.current = formik;
```
ThinuwanW, line 141: 🤖 **Claude** · 🟡 **Suggestion**

`formikRef.current = formik` writes to a ref during the render phase, which breaks React's purity rule — a render that React discards (Suspense retry, concurrent interruption, StrictMode double-invoke) still leaves its formik bag in the ref, and React Compiler / `react-hooks` lint rules flag render-phase ref mutation. It works today only because the assignment happens to be idempotent.

The state-based approach described in the `handleError` comment removes the ref (and the `FormikProps` import at line 9) entirely, which is the cleaner fix. If you want to keep the imperative shape, move the assignment into an effect: `useEffect(() => { formikRef.current = formik; });` — safe here because the mutation callback can only fire after at least one commit.

frontend/src/community/leave/components/organisms/LeavePolicyWizard/LeavePolicyWizard.tsx
```diff
@@ -100,6 +103,16 @@ const LeavePolicyWizard: FC<Props> = ({ policyType }) => {
   };
 
   const handleError = (error: AxiosError): void => {
+    if (isDuplicatePolicyNameError(error)) {
+      formikRef.current?.setFieldTouched("policyName", true, false);
+      formikRef.current?.setFieldError(
+        "policyName",
+        translateText(["duplicateToastDescription"])
```
ThinuwanW, line 110: 🤖 **Claude** · 🟡 **Suggestion**

i18n: `duplicateToastDescription` is a toast body being reused as an inline field error. The English value ("A policy with this name already exists for the selected leave type. Please use a different name.") reads long for a field-level message, and the key name now lies about its usage context. Add a dedicated key under `leaveModule.leavePolicies.createPolicy`, e.g. `"duplicateNameFieldError": "A policy with this name already exists for this leave type."`, and reference that instead.

Related dead key: with this early return, `createPolicy.duplicateToastTitle` (leaveModule.json:480) is now unreachable from the create flow — `getLeavePolicyErrorToastKeys`'s duplicate branch is never hit from this component anymore. It is still used by the edit/activate flows, so the util branch stays, but the `createPolicy.duplicateToastTitle` entry should be removed if no dedicated field-error key reuses it.

frontend/src/community/leave/components/organisms/LeavePolicyWizard/LeavePolicyWizard.tsx
```diff
@@ -100,6 +103,16 @@ const LeavePolicyWizard: FC<Props> = ({ policyType }) => {
...
+      );
+      setActiveStep(LeavePolicyWizardSteps.BASIC_INFO);
```
ThinuwanW, line 112: 🤖 **Claude** · 🟡 **Suggestion**

Accessibility regression from dropping the toast: the toast was an announced, attention-grabbing surface; the replacement silently changes `activeStep` and sets a field error with no focus move and no live-region announcement. A keyboard or screen-reader user who submits from the Summary step is teleported back to step 1 with no indication of why.

Suggest moving focus to the policy-name input when the duplicate error is set (a ref on the `InputField` in `BasicInfoStep`, or an `autoFocus`-style effect keyed on the error), and verifying the `InputField` `errorMessage` renders in a live region — `BasicInfoStep.tsx:83` already uses `role="alert"` for the no-leave-types warning, so mirror that if the shared component doesn't provide it.

frontend/src/community/leave/utils/leavePolicy/leavePolicyUtils.ts
```diff
@@ -19,6 +19,10 @@ interface LeavePolicyErrorToastKeys {
   description: string;
 }
 
+export const isDuplicatePolicyNameError = (error: AxiosError): boolean =>
```
ThinuwanW, line 22: 🤖 **Claude** · 🟡 **Suggestion**

This duplicates the error-shape extraction already present in `getLeavePolicyErrorToastKeys` immediately below (same `as LeavePolicyErrorData | undefined` cast, same `results?.[0]?.messageKey` walk). Extract a single private helper and build both on top of it:

`const getLeavePolicyErrorMessageKey = (error: AxiosError): string | undefined => (error?.response?.data as LeavePolicyErrorData | undefined)?.results?.[0]?.messageKey;`

Then `isDuplicatePolicyNameError` becomes a one-liner over it and `getLeavePolicyErrorToastKeys` switches on its result. This keeps the two in sync if the error envelope shape ever changes, and confines the `as` assertion to one place.

frontend/src/community/leave/utils/leavePolicy/leavePolicyUtils.ts
```diff
@@ -19,6 +19,10 @@ interface LeavePolicyErrorToastKeys {
...
+  (error?.response?.data as LeavePolicyErrorData | undefined)?.results?.[0]
```
ThinuwanW, line 23: 🤖 **Claude** · 🟡 **Suggestion**

Inconsistent UX across the two paths that can produce this exact error. The backend raises `LEAVE_ERROR_LEAVE_POLICY_ALREADY_EXISTS` from both `addLeavePolicy` (LeavePolicyServiceImpl.java:87) and `updateLeavePolicy` (LeavePolicyServiceImpl.java:111), but only the create wizard gets the inline treatment — `EditLeavePolicyModal.tsx:61` still falls through to a generic toast when a user renames a policy into a collision.

Since `isDuplicatePolicyNameError` is exported, apply it there too. The edit modal is actually the easier case: it already renders `errorMessage={errors.policyName}` without a `touched` gate (EditLeavePolicyModal.tsx:124-125) and its `useFormik` bag is in scope by the time `onUpdateError` runs, so a plain `setFieldError("policyName", ...)` works with no ref indirection.

frontend/src/community/leave/utils/leavePolicy/leavePolicyUtils.ts
```diff
@@ -19,6 +19,10 @@ interface LeavePolicyErrorToastKeys {
...
+    ?.messageKey === LEAVE_ERROR_LEAVE_POLICY_ALREADY_EXISTS;
```
ThinuwanW, line 24: 🤖 **Claude** · 🟣 **Nit**

No test coverage for the new exported util. `leavePolicyUtils.ts` has no `.test.ts`, while the sibling module `policyLeaveTypes/policyLeaveTypeUtils.test.ts` already tests exactly this error-envelope mapping and even has a reusable `buildError = (messageKey?: string): AxiosError` fixture you can copy. `isDuplicatePolicyNameError` is a pure function over an easily-faked shape — add `leavePolicyUtils.test.ts` covering: matching messageKey → true; a different messageKey (e.g. `COMMON_ERROR_ACCESS_DENIED`) → false; `data` present but `results` missing/empty → false; and no `response` at all (network error) → false. That last case is the one most likely to regress if the optional chaining is ever tightened.

frontend/src/community/leave/components/organisms/LeavePolicyWizard/LeavePolicyWizard.tsx
```diff
@@ -100,6 +103,16 @@ const LeavePolicyWizard: FC<Props> = ({ policyType }) => {
   };
 
   const handleError = (error: AxiosError): void => {
+    if (isDuplicatePolicyNameError(error)) {
+      formikRef.current?.setFieldTouched("policyName", true, false);
```
ThinuwanW, line 107: 🤖 **Claude** · 🔴 **Important**

The `formikRef` indirection is unnecessary and it turns the duplicate branch into a path that can silently produce zero user feedback. `handleError` is only ever invoked asynchronously from React Query's `onError`; `useAddLeavePolicy` passes `onError` straight through to `useMutation` (LeavePolicyApi.ts:92), and React Query 5 re-applies options on every render, so the latest `handleError` closure is called. That closure can reference the `formik` const declared below it directly — no TDZ issue, because by call time the binding is initialised — and Formik 2.4's `setFieldError`/`setFieldTouched` are `useEventCallback`-stable, so even a stale closure would mutate the current form state correctly. This is exactly what the established precedent in the codebase does: `AddEditTeamModal.tsx:108-115` destructures `setFieldError` from `useFormik` (declared at line 190, *after* the error handler) and calls it directly from `onAddError`. Separately, because the branch uses `formikRef.current?.setFieldError(...)` and then `return`s unconditionally, if the ref is ever null the field error is dropped AND the fallback toast is skipped — the user sees the wizard silently jump to step 1 with no message. Fix: delete `formikRef` and the `FormikProps` import (line 9), call `formik.setFieldTouched`/`formik.setFieldError` directly, and keep the toast reachable as a fallback rather than returning past it unconditionally. This also removes the render-phase ref write entirely.

frontend/src/community/leave/components/organisms/LeavePolicyWizard/LeavePolicyWizard.tsx
```diff
@@ -100,6 +103,16 @@ const LeavePolicyWizard: FC<Props> = ({ policyType }) => {
...
+      formikRef.current?.setFieldError(
```
ThinuwanW, line 108: 🤖 **Claude** · 🟡 **Suggestion**

Stale-name race: the inline error can be pinned to a policy name the user has already changed. For flexible (non-accrual) policies `isLastStep` is true on BASIC_INFO, so submission happens while `BasicInfoStep`'s policy-name `InputField` is still mounted — and that input is not disabled during the request (only the Back/Next buttons get `disabled={isPending}`). If the user edits the name while the POST is in flight, the server's duplicate response for the *old* name is attached to the *new* (possibly perfectly valid) name, telling the user a name already exists when it does not. The component already captures the submitted value in `submittedNameRef.current` (set in `onSubmit`), so guard on it: only apply the inline error when `formik.values.policyName.trim() === submittedNameRef.current`, otherwise fall through to the existing error toast. Alternatively pass `isPending` down and disable the name input while the mutation is pending.

frontend/src/community/leave/components/organisms/LeavePolicyWizard/LeavePolicyWizard.tsx
ThinuwanW, line 110: 🤖 **Claude** · 🟡 **Suggestion**

Dead i18n key and a divergent second representation of the same error. `LeavePolicyWizard` is the only component that calls `getLeavePolicyErrorToastKeys` with the `leaveModule.leavePolicies.createPolicy` namespace (verified: the other callers — LeavePoliciesTable, DeactivateLeavePolicyModal, EditLeavePolicyModal — use their own namespaces). Because the new branch returns before reaching it, `createPolicy.duplicateToastTitle` ("Duplicate policy name", leaveModule.json:480) is now unreachable, and the `LEAVE_ERROR_LEAVE_POLICY_ALREADY_EXISTS` case of the switch is dead for this flow — so a future edit to that switch will silently not affect the wizard. Also, the key placement violates the module's own convention: every other inline field error for this form lives under `createPolicy.errors.*` (validations.ts:157 uses `translateText(["errors", "policyNameRequired"])`, and BasicInfoStep renders those through the same `errorMessage` prop). Fix: add `createPolicy.errors.policyNameDuplicate` to leaveModule.json, reference it as `translateText(["errors", "policyNameDuplicate"])`, and delete the now-unused `createPolicy.duplicateToastTitle` / `duplicateToastDescription` pair.

frontend/src/community/leave/components/organisms/LeavePolicyWizard/LeavePolicyWizard.tsx
ThinuwanW, line 112: 🤖 **Claude** · 🟣 **Nit**

Field attribution is incomplete for a composite uniqueness constraint. The backend check is `existsByNameIgnoreCaseAndLeaveType_Id(name, leaveTypeId)` (LeavePolicyServiceImpl.java:86), i.e. uniqueness is scoped to the (name, leaveType) pair, and the message itself says "...already exists for the selected leave type". The error is surfaced only on `policyName`, so a user who would rather resolve the collision by picking a different leave type gets no signal on that field. Consider also marking `leaveType` as errored/touched, or wording the inline message so it's unambiguous that either field can be changed to resolve it. Flagging as a nit since attributing it to the name field is a defensible default.

frontend/src/community/leave/utils/leavePolicy/leavePolicyUtils.ts
ThinuwanW, line 22: 🤖 **Claude** · 🟡 **Suggestion**

The `as` cast is avoidable and the optional chaining on `error` is dead code. Per the type-safety guideline, prefer fixing the type over asserting: declare the parameter as `error: AxiosError<LeavePolicyErrorData>` and the body reduces to `error.response?.data?.results?.[0]?.messageKey === LEAVE_ERROR_LEAVE_POLICY_ALREADY_EXISTS` with no assertion at all (`AxiosError`'s first generic is the response-data type). `error?.` is also redundant given the non-optional `error: AxiosError` signature — it silences nothing and suggests a nullable input that the type forbids. Additionally, `LeavePolicyErrorData` re-declares a shape that already exists as `ErrorResponse` in `~community/common/types/CommonTypes` (line 39-48, `results: { message?; code?; messageKey? }[]`); reusing or deriving from that shared type would keep the error contract in one place across modules.

frontend/src/community/leave/components/organisms/LeavePolicyWizard/LeavePolicyWizard.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/organisms/LeavePolicyWizard/LeavePolicyWizard.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/organisms/LeavePolicyWizard/LeavePolicyWizard.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/organisms/LeavePolicyWizard/LeavePolicyWizard.tsx
hasalarootcode: Not needed

frontend/src/community/leave/utils/leavePolicy/leavePolicyUtils.ts
hasalarootcode: Not needed

frontend/src/community/leave/utils/leavePolicy/leavePolicyUtils.ts
hasalarootcode: Not needed

frontend/src/community/leave/utils/leavePolicy/leavePolicyUtils.ts
hasalarootcode: Not needed

frontend/src/community/leave/components/organisms/LeavePolicyWizard/LeavePolicyWizard.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/organisms/LeavePolicyWizard/LeavePolicyWizard.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/organisms/LeavePolicyWizard/LeavePolicyWizard.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/organisms/LeavePolicyWizard/LeavePolicyWizard.tsx
hasalarootcode: Not needed

frontend/src/community/leave/utils/leavePolicy/leavePolicyUtils.ts
hasalarootcode: Not needed

shakila-kamalasena (approved):

thusala (approved):

akilarootcode (approved):

PR #1775, opened 2026-08-17
Fix/3015: Leave policy - Mouse scroll and trackpad scroll change values in day input fields on Accrual step 2

hasalarootcode:
…upStep

### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/frontend/src/community/leave/components/organisms/LeavePolicyWizard/EntitlementSetupStep.tsx
+++ b/frontend/src/community/leave/components/organisms/LeavePolicyWizard/EntitlementSetupStep.tsx
@@ -71,7 +71,6 @@ const EntitlementSetupStep: FC<Props> = ({
             <InputField
               label={translateText(["employeesAccrueLabel"])}
               name="accrualDays"
-              type="number"
               value={formData.accrualDays}
               placeholder={translateText(["employeesAccruePlaceholder"])}
               state={
@@ -130,7 +129,6 @@ const EntitlementSetupStep: FC<Props> = ({
               <InputField
                 label={translateText(["waitingPeriodDaysLabel"])}
                 name="waitingPeriodDays"
-                type="number"
                 value={formData.waitingPeriodDays}
                 placeholder={translateText(["waitingPeriodDaysPlaceholder"])}
                 state={
@@ -168,7 +166,6 @@ const EntitlementSetupStep: FC<Props> = ({
               <InputField
                 label={translateText(["accrualCapDaysLabel"])}
                 name="accrualCapDays"
-                type="number"
                 value={formData.accrualCapDays}
                 placeholder={translateText(["accrualCapDaysPlaceholder"])}
                 state={
@@ -213,7 +210,6 @@ const EntitlementSetupStep: FC<Props> = ({
                 <InputField
                   label={translateText(["maxCarryOverDaysLabel"])}
                   name="maxCarryOverDays"
-                  type="number"
                   value={formData.maxCarryOverDays}
                   placeholder={translateText(["maxCarryOverDaysPlaceholder"])}
                   state={


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR removes `type="number"` from the four numeric "days" inputs in the accrual entitlement step, turning them into plain text fields. The change itself is reasonable (it avoids the `type=number` empty-value/spinner pitfalls and the form already stores these as strings in `LeavePolicyFormData`), but it is incomplete: nothing was added to replace the numeric affordances the browser was providing, and the Yup layer that must now be the only guard relies on `Number()` coercion, which accepts free-text forms the number input previously blocked. No sibling code, payload mapping, or tests break, but I'd want the input-restriction and validation-tightening items addressed before this ships.

Found **14** new issue(s): 🔴 4 important, 🟡 8 suggestion(s), 🟣 2 nit(s)

frontend/src/community/leave/components/organisms/LeavePolicyWizard/EntitlementSetupStep.tsx
```diff
@@ -71,7 +71,6 @@ const EntitlementSetupStep: FC<Props> = ({
             <InputField
```
ThinuwanW, line 71: 🤖 **Claude** · 🔴 **Important**

All four day fields (accrualDays L71, waitingPeriodDays L129, accrualCapDays L166, maxCarryOverDays L210) now render as plain text inputs with no numeric affordance at all: no `inputMode`, no key/paste restriction, no `maxLength`. Concretely this regresses (a) mobile/tablet UX — the OS shows the full alphabetic keyboard instead of the numeric pad for a field labelled "Enter days", and (b) input hygiene — a user can type or paste "abc", "--", or a 40-character string and only discovers the problem when they press Next. This module already has an established pattern for exactly this case: see LeaveTypeForm.tsx:475-480, which uses `inputType="text"` + `inputMode="numeric"` + `onKeyDown={carryForwardKeyDownRestriction}` + `onPaste={carryForwardPasteRestriction}` (from ~community/leave/utils/leaveTypes/LeaveTypeUtils), plus `maxLength={3}`. Suggest applying the same here: `inputMode="decimal"` for accrualDays and maxCarryOverDays (validation allows 0.5-day values via MIN_POLICY_DAYS = 0.5), `inputMode="numeric"` for waitingPeriodDays and accrualCapDays, a keydown/paste restriction on each, and a `maxLength` (3 is enough given MAX_POLICY_DAYS = 365).

frontend/src/community/leave/components/organisms/LeavePolicyWizard/EntitlementSetupStep.tsx
```diff
@@ -168,7 +166,6 @@ const EntitlementSetupStep: FC<Props> = ({
               <InputField
```
ThinuwanW, line 166: 🤖 **Claude** · 🔴 **Important**

With the number input gone, `leavePolicyWizardValidation` in ~community/leave/utils/validations.ts is now the only gate on these values, and its `isNumberInRange` helper validates via bare `Number(value)`. That coercion accepts strings a number input would never have produced: `Number("0x1F")` = 31, `Number("1e2")` = 100, `Number(" 5 ")` = 5, `Number("Infinity")` = Infinity. Two concrete consequences: (1) SummaryStep.tsx:88-121 renders the raw form string, so the user is shown "0x1F days" / "1e2 days" on the confirmation screen while mapLeavePolicyFormToPayload sends 31 / 100 — a review-then-save mismatch; (2) `waitingPeriodDays` and `accrualCapDays` have no `max` in the schema, so "1e400" coerces to Infinity, passes `>= min`, and `JSON.stringify` serializes it as `null` in the request body — an unintended null reaching the backend for a field the UI presented as valid. Fix by adding a format test alongside the range test rather than relying on `Number()`: `allowsOnlyNumbersAndOptionalDecimal()` is already imported in this same validations.ts file (line 3, used at line 50) and matches the 0.5-step semantics for accrualDays/maxCarryOverDays; use a plain integer check for waitingPeriodDays/accrualCapDays. Also worth adding an explicit upper bound (MAX_POLICY_DAYS) to accrualCapDays and waitingPeriodDays while you're in there.

frontend/src/community/leave/components/organisms/LeavePolicyWizard/EntitlementSetupStep.tsx
```diff
@@ -130,7 +129,6 @@ const EntitlementSetupStep: FC<Props> = ({
...
```
ThinuwanW, line 129: 🤖 **Claude** · 🟡 **Suggestion**

waitingPeriodDays is validated only as `isNumberInRange(value, MIN_WAITING_PERIOD_DAYS)` (no integer constraint), but mapLeavePolicyFormToPayload wraps it in `Math.round(...)` (leavePolicyUtils.ts:81). So entering "2.5" passes validation, SummaryStep shows "2.5 days" on the confirm screen, and the backend receives 3. This mismatch existed before, but the removal of the number input makes decimal entry easier and there is no longer any browser-level step hint for the field. Recommend validating waitingPeriodDays as an integer (e.g. a `/^\d+$/` format test with a dedicated error message) so the value the user confirms is the value that is saved, and dropping the now-redundant `Math.round`.

frontend/src/community/leave/components/organisms/LeavePolicyWizard/EntitlementSetupStep.tsx
```diff
@@ -213,7 +210,6 @@ const EntitlementSetupStep: FC<Props> = ({
                 <InputField
```
ThinuwanW, line 210: 🤖 **Claude** · 🟡 **Suggestion**

The `type` prop was deleted outright rather than switched to `type="text"`, which now depends on whatever default the `@rootcodelabs/skapp-ui` InputField applies internally. Every sibling usage in this feature passes it explicitly — BasicInfoStep.tsx:60, EditLeavePolicyModal.tsx:122, PolicyLeaveTypeForm.tsx:199 all use `type="text"`. Pass `type="text"` on all four fields here for consistency and so the intent (numeric value captured as text, validated by Yup) is explicit and not silently changed by a future design-system default change.

frontend/src/community/leave/components/organisms/LeavePolicyWizard/EntitlementSetupStep.tsx
ThinuwanW, line 71: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

There is no test file anywhere under LeavePolicyWizard/ (no EntitlementSetupStep.test.tsx, no LeavePolicyWizard.test.tsx), so this behavioural change is entirely uncovered. Since the browser no longer rejects non-numeric characters, the Yup rules are the only safety net — a small test rendering EntitlementSetupStep with MockTheme, typing a non-numeric value into the "Employees accrue" field (`screen.getByRole('textbox', { name: /employees accrue/i })`) and asserting the error message renders would lock in the new contract cheaply.

frontend/src/community/leave/components/organisms/LeavePolicyWizard/EntitlementSetupStep.tsx
ThinuwanW, line 129: 🤖 **Claude** · 🔴 **Important**

`waitingPeriodDays` (this field) and `accrualCapDays` (line 166) have NO upper bound. In `~community/leave/utils/validations.ts` they call `isNumberInRange(value, MIN_WAITING_PERIOD_DAYS)` and `isNumberInRange(value, MIN_ACCRUAL_CAP_DAYS)` with the third `max` argument omitted, unlike `accrualDays` and `maxCarryOverDays` which both pass `MAX_POLICY_DAYS`. I checked the server side: `backend/.../payload/request/LeavePolicyAccrualDetailDto.java` declares plain `Integer waitingPeriodDays` / `Float accrualCapDays` with no `@Min`/`@Max`/`@NotNull`, and `LeavePolicyServiceImpl.applyAccrualDetail` copies them straight onto the entity — so there is no server-side range check anywhere. While the browser number input existed this was hard to hit; now a user can type `1e10`, which for `waitingPeriodDays` exceeds `Integer.MAX_VALUE` and is rejected by Jackson as a 400 that surfaces only as the generic error toast, and for `accrualCapDays` fits in a Float and silently persists a ten-billion-day cap. Fix: pass `MAX_POLICY_DAYS` as the third argument to both `isNumberInRange` calls (and ideally add `@Min`/`@Max` to the backend DTO so the client is not the only gate).

frontend/src/community/leave/components/organisms/LeavePolicyWizard/EntitlementSetupStep.tsx
ThinuwanW, line 71: 🤖 **Claude** · 🔴 **Important**

`isNumberInRange` validates via `Number(value)`, which accepts several string forms the old number input could never produce — and the payload mapper then re-coerces with `Number()`, so what validates is not what the user typed. Concretely: `Number("0x1F")` is 31, `Number("1e2")` is 100, `Number(" 5 ")` and `Number("+5")` are 5, `Number("007")` is 7. Worst case, `Number("Infinity")` is `Infinity`, which passes the unbounded `waitingPeriodDays`/`accrualCapDays` tests (see the separate finding), survives `Math.round`, and is serialised by `JSON.stringify` as `null` — so the policy saves with the waiting period / cap unset while the wizard showed a value and reported success. Fix: gate on a strict decimal string before coercing. The codebase already has `allowsOnlyNumbersAndOptionalDecimal()` in `~community/common/regex/regexPatterns`, and `customLeaveAllocationValidation` in the same validations file already uses it for its days field — reuse it here (it also matches the whole-or-half-day semantics implied by `MIN_POLICY_DAYS` of 0.5) so the wizard is consistent with the rest of the leave module.

frontend/src/community/leave/components/organisms/LeavePolicyWizard/EntitlementSetupStep.tsx
ThinuwanW, line 210: 🤖 **Claude** · 🟡 **Suggestion**

Inconsistent reset behaviour across the three radio groups. The `hasWaitingPeriod` handler clears `waitingPeriodDays` and the `hasAccrualCap` handler clears `accrualCapDays` when toggled to No, but the `canCarryOver` handler only sets `canCarryOver` and leaves `maxCarryOverDays` untouched. The payload mapper guards on `canCarryOver` so nothing wrong is submitted, but the stale value stays in form state: with the number input gone a user can now leave arbitrary text such as `abc` in this field, toggle carry-over off, and have that text reappear when they toggle it back on. Fix: mirror the other two handlers — spread `...(value ? {} : { maxCarryOverDays: "" })` into the `canCarryOver` onChange payload.

frontend/src/community/leave/components/organisms/LeavePolicyWizard/EntitlementSetupStep.tsx
ThinuwanW, line 71: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Display/payload divergence between this step and the summary step. `SummaryStep.tsx` interpolates the raw form strings (`formData.accrualDays`, `waitingPeriodDays`, `accrualCapDays`, `maxCarryOverDays`) directly into the `{{days}}` placeholders, while `mapLeavePolicyFormToPayload` sends `Number(...)` of the same strings. While these were number inputs the two always agreed; as free-text fields they can disagree — entering `0x1F` shows "0x1F days" on the summary card but persists 31, and `007` shows "007 days" but persists 7. The user confirms one value and the system stores another. Fix: normalise the value once on change (e.g. store the sanitised numeric string in form state) so the summary, validation, and payload all read the same canonical value.

ThinuwanW, line 71: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

`leavePolicyWizardValidation` has no test coverage, even though this PR makes it the sole enforcement point for these four values (the backend DTO has no validation annotations). Note this is not the same gap as the missing component tests: `frontend/src/community/leave/utils/__test__/validations.test.ts` already exists and is the natural home, but grepping it for `leavePolicyWizardValidation`, `waitingPeriodDays`, `accrualCapDays`, and `maxCarryOverDays` returns nothing. Fix: add cases to that existing file covering the boundaries (`MIN_POLICY_DAYS` 0.5, `MAX_POLICY_DAYS` 365, `MIN_WAITING_PERIOD_DAYS`, `MIN_ACCRUAL_CAP_DAYS`), the conditional `.when()` branches for each radio toggle, and the now-reachable non-numeric inputs (`abc`, `1e2`, `0x1F`, `Infinity`, whitespace-padded values).

frontend/src/community/leave/components/organisms/LeavePolicyWizard/EntitlementSetupStep.tsx
ThinuwanW, line 166: 🤖 **Claude** · 🟡 **Suggestion**

Error copy is now misleading because non-numeric input is reachable. `accrualCapDays` failing `isNumberInRange` renders `errors.accrualCapRequired` — "Please enter the maximum accrual balance." — which is shown even when the field visibly contains `abc`, telling the user to enter something they have already entered. The same applies to `waitingPeriodDaysRequired` at line 129, and `accrualDaysInvalid` ("Days accrued must be greater than 0.") at line 71 is equally wrong for a non-numeric entry. Fix: add a distinct key such as `errors.mustBeValidNumber` in `leaveModule.json` and split the Yup rules into a required check and a separate numeric-format check so each condition gets accurate copy.

frontend/src/community/leave/components/organisms/LeavePolicyWizard/EntitlementSetupStep.tsx
ThinuwanW, line 129: 🤖 **Claude** · 🟡 **Suggestion**

Stale `touched` makes conditional fields render pre-errored. `handleNext` in `LeavePolicyWizard.tsx` marks every field in `STEP_FIELDS[ENTITLEMENT_SETUP]` as touched whenever any one of them errors — including `waitingPeriodDays` and `accrualCapDays`, which may be hidden at that moment. Repro: leave `accrualDays` blank, click Next (all five fields become touched), fill in `accrualDays`, then switch the waiting-period radio to Yes. The field is revealed with an empty value, validation now applies, and because `touched.waitingPeriodDays` is already true it renders immediately in `state="error"` with a red message before the user has typed a character. Fix: either restrict the `setTouched` reduce to fields that are actually visible for the current form state, or clear the touched flag alongside the value in the radio onChange handlers.

frontend/src/community/leave/components/organisms/LeavePolicyWizard/EntitlementSetupStep.tsx
ThinuwanW, line 71: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

The four day-input blocks are now byte-for-byte identical apart from the field name and translation keys — same label/name/value/placeholder/state/errorMessage/onChange/fullWidth shape. That this PR had to make the exact same one-line edit in all four places is the signal. Per the duplication guideline, extract a small colocated component (e.g. `PolicyDaysInputField`) taking the field name plus the label/placeholder keys and deriving `state`/`errorMessage` from the passed `errors`/`touched`. That also gives a single place to add the numeric handling the other findings call for (sanitising onChange, `inputMode`, `maxLength`) instead of repeating it four times.

ThinuwanW, line 71: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

Possible missing paired change. After this PR, `frontend/src/enterprise/invoice/components/molecules/ProjectMembersTable/ProjectMembersTable.tsx:174` is the only remaining `type="number"` usage of the skapp-ui `InputField` in the repo. If the motivation here was a defect in how that component handles `type="number"`, the invoice billable-rate field has the same exposure — and it is arguably worse there, since its onChange does `Number(e.target.value)` straight into state, so clearing the field writes 0 rather than an empty value. Worth confirming whether that call site needs the same treatment, or documenting why it is exempt. Separately, `BasicInfoStep.tsx:60` passes `type="text"` explicitly, so the wizard is now internally inconsistent about whether `type` is specified at all.

hasalarootcode: Not needed

frontend/src/community/leave/components/organisms/LeavePolicyWizard/EntitlementSetupStep.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/organisms/LeavePolicyWizard/EntitlementSetupStep.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/organisms/LeavePolicyWizard/EntitlementSetupStep.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/organisms/LeavePolicyWizard/EntitlementSetupStep.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/organisms/LeavePolicyWizard/EntitlementSetupStep.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/organisms/LeavePolicyWizard/EntitlementSetupStep.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/organisms/LeavePolicyWizard/EntitlementSetupStep.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/organisms/LeavePolicyWizard/EntitlementSetupStep.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/organisms/LeavePolicyWizard/EntitlementSetupStep.tsx
hasalarootcode: Not needed

hasalarootcode: Not needed

shakila-kamalasena (approved):

thusala (approved):

akilarootcode (approved):

PR #1776, opened 2026-08-17
[Chore] Bump @rootcodelabs/skapp-ui to 1.1.1-beta.13

thusala:


akilarootcode (approved):

ThinuwanW (commented):
## 🤖 Claude Code Review

This PR is a pure dependency bump of the internal design-system package `@rootcodelabs/skapp-ui` from 1.1.1-beta.10 to 1.1.1-beta.13 in both `frontend/package.json` and `frontend/package-lock.json`. The lockfile is internally consistent — the root `packages[""].dependencies` entry, the `node_modules/@rootcodelabs/skapp-ui` version, the `resolved` GitHub Packages URL and the `integrity` hash were all updated together, and the package's `peerDependencies` block is unchanged, so no new peer requirements are introduced (dnd-kit, react-day-picker, dompurify and the tiptap peers are already satisfied by the existing tree). No source files are touched. The exact-pin style (no caret) matches how this package was previously declared. No correctness issues found in the diff itself; remaining comments are process/verification concerns.

Found **8** new issue(s): 🔴 2 important, 🟡 4 suggestion(s), 🟣 2 nit(s)

frontend/package.json
```diff
@@ -29,7 +29,7 @@
     "@mui/base": "^5.0.0-beta.58",
     "@mui/x-date-pickers": "^7.29.1",
     "@next/third-parties": "^15.3.1",
-    "@rootcodelabs/skapp-ui": "1.1.1-beta.10",
+    "@rootcodelabs/skapp-ui": "1.1.1-beta.13",
```
ThinuwanW, line 32: 🤖 **Claude** · 🟡 **Suggestion**

The bump skips two prerelease versions (beta.11, beta.12) and ships with zero accompanying source changes, yet `@rootcodelabs/skapp-ui` is imported by 100+ components across community and enterprise modules (leave policy wizard/modals, CRM deal panels, people/configurations tables, enterprise OAuth cards and invoice tables). A prerelease design-system jump can change component prop contracts, default styling, or internal DOM structure. Please (a) link the release notes / changelog diff for beta.11→beta.13 in the PR description, and (b) confirm the consuming screens were smoke-tested. `next build` type-checking will catch removed/renamed props, but it will not catch runtime, layout, or a11y regressions (e.g. changed default variants or altered roles), and several component tests here assert on rendered structure — worth running the full Jest suite plus a visual pass on the leave-policy and CRM flows before merge.

frontend/package-lock.json
```diff
@@ -7651,9 +7651,9 @@
       "peer": true
     },
     "node_modules/@rootcodelabs/skapp-ui": {
-      "version": "1.1.1-beta.10",
-      "resolved": "https://npm.pkg.github.com/download/@rootcodelabs/skapp-ui/1.1.1-beta.10/88a6f4bba392083c190b33f7ba19c471000a4204",
-      "integrity": "sha512-8TzMsA0b41h/uzNlxABwMkjfFfS5YL1oMWjncKaxjlLdCD4DG+3WafvkrEp+IwPdTqrsWa62tEzNRreaw44Vqg==",
+      "version": "1.1.1-beta.13",
+      "resolved": "https://npm.pkg.github.com/download/@rootcodelabs/skapp-ui/1.1.1-beta.13/5cbe426143804e600c6b91265020f07ef950faf9",
```
ThinuwanW, line 7655: 🤖 **Claude** · 🟣 **Nit**

Reproducibility gap this PR relies on: `.github/workflows/build-fe.yml` runs `npm install` (not `npm ci`) in the frontend build job. `npm install` is allowed to rewrite `package-lock.json` and re-resolve transitive `^` ranges, so the pinned `resolved`/`integrity` values added here are never actually enforced or validated by CI — a lockfile that drifts from `package.json`, or an integrity mismatch, would pass the PR build silently. Since this PR's entire payload is lockfile integrity metadata, consider switching the workflow's install step to `npm ci` so the committed lockfile is the source of truth for CI and container builds.

frontend/package.json
ThinuwanW, line 32: 🤖 **Claude** · 🔴 **Important**

This bump crosses three prereleases of a library imported by 476 files in this repo, yet nothing in CI can catch a breaking API change. `.github/workflows/build-fe.yml` is the only frontend job and it runs just `npm install` + `npm run build` — no `check-lint`, no `test`. `next.config.mjs` then disables both gates that build would otherwise provide: `typescript: { ignoreBuildErrors: true }` (line 466-468) and `eslint: { ignoreDuringBuilds: true }` (line 463-465). There is also no `typecheck` script in package.json. Net effect: if beta.13 renamed a prop, tightened a prop type, or dropped an export, the build still goes green and the failure surfaces at runtime in production. Fix: add a `"typecheck": "tsc --noEmit"` script and a CI step that runs it (this is the single highest-value change here — it costs one line and actually validates the bump), and paste the `tsc --noEmit` output in the PR description as evidence. Longer term, remove `ignoreBuildErrors`/`ignoreDuringBuilds` so `npm run build` is a real gate.

ThinuwanW, line 32: 🤖 **Claude** · 🔴 **Important**

📁 **File-level observation**

The one CI job this PR relies on for validation runs on a Node version the toolchain declares unsupported, and it does not match the production runtime. `.github/workflows/build-fe.yml` pins `node-version: '18'`, but `@tailwindcss/oxide@4.2.1` (a transitive dep of `@tailwindcss/postcss`, package-lock line 9263-9270) declares `engines: { node: '>= 20' }`, and `frontend/Dockerfile` runs `node:20-alpine3.19`. npm only emits an EBADENGINE warning rather than failing, so this passes silently. The `engines` field here (`node: '>=18.0'`) is also now inconsistent with what the dependency tree actually requires. Fix: bump the workflow to `node-version: '20'`, and raise `engines.node` to `>=20.0` so a mismatched local Node is caught at install time rather than producing a subtly different CSS build than production.

frontend/package-lock.json
ThinuwanW, line 7655: 🤖 **Claude** · 🟡 **Suggestion**

This is the only dependency in the tree whose `resolved` host is `npm.pkg.github.com` rather than `registry.npmjs.org`, but there is no `.npmrc` committed anywhere in the repo (I checked repo root and frontend/, and `.gitignore` does not exclude one), and `build-fe.yml` uses `actions/setup-node@v4` with no `registry-url` and no `NODE_AUTH_TOKEN` env. Installs work only as long as npm can reuse the lockfile's `resolved` entry verbatim; the moment npm has to re-resolve the `@rootcodelabs` scope (lockfile regeneration, adding a dep, a peer conflict), it queries the default registry where this package does not exist and fails with E404 — and GitHub Packages' npm endpoint requires a token even for public packages. This PR makes it concrete because the beta.13 tarball is a fresh fetch that cannot be served from any warm cache. Fix: commit a `frontend/.npmrc` with `@rootcodelabs:registry=https://npm.pkg.github.com` and `//npm.pkg.github.com/:_authToken=${NODE_AUTH_TOKEN}`, and set `registry-url` + `NODE_AUTH_TOKEN` in the workflow's setup-node step. Note the README only links to external docs (docs.skapp.com) and says nothing about this, so external contributors to this open-source repo currently cannot install the frontend.

frontend/package-lock.json
```diff
@@ -7651,9 +7651,9 @@
...
+      "integrity": "sha512-mffc1H1+G6GHoEjkM53WOOeDj1ZZ9KHdgFtZxDBE5fxXeGAo9rV6aR0w6Cq5ewcqnXJFPD0Cap231DjvxFjpQA==",
       "license": "ISC",
       "peerDependencies": {
         "@dnd-kit/core": "^6.3.1",
```
ThinuwanW, line 7659: 🤖 **Claude** · 🟡 **Suggestion**

This package declares 20 peerDependencies, and the entire `@tiptap/*` v3 stack plus `tiptap-extension-resize-image` are not declared in `frontend/package.json` at all — they exist only as npm-7 auto-installed peers (each is marked `"peer": true` in the lockfile, e.g. `@tiptap/core` at line 9673-9686). They resolved to 3.13.0 from a `^3.3.0` range, meaning ~20 packages that ship into the production bundle float on a range this repo does not control and never reviewed. No application code imports `@tiptap/*` directly (I checked), so nothing is broken today, but a lockfile regeneration silently picks up whatever the newest 3.x is. Fix: promote the peers that actually ship to the browser into explicit `dependencies` in package.json at the versions currently resolved, so they are visible to review and to dependency scanning.

frontend/package.json
ThinuwanW, line 32: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

The `test-all` script (line 15) chains its three commands with a single `&` instead of `&&`. Under the POSIX shell npm uses, `&` backgrounds `check-format` and `check-lint` and the script's exit code reflects only the final `npm run test` — so format and lint failures are silently discarded and the script reports success. This matters for this PR specifically because `test-all` is the natural one-shot local gate a reviewer would run to sanity-check a dependency bump, and it gives a false pass. Fix: change the separators to `&&` (or use `npm-run-all -s` if sequencing with clearer output is preferred).

ThinuwanW, line 32: 🤖 **Claude** · 🟣 **Nit**

`jest.config.js` maps only the `~community`/`~enterprise`/`~public`/`~styles`/`~i18n`/`~middleware` aliases, leaves `next/jest`'s default `transformIgnorePatterns` (which skips all of node_modules) in place, and there is no `__mocks__` directory or manual mock for this package. If beta.13 changed its distribution format — ESM-only output, a new `exports` map, or untranspiled syntax — every one of the 222 test suites that transitively imports it would fail to parse. I could not verify the shipped format because the package is private and not installed locally. Since no CI job runs `npm test`, this would only be discovered by whoever next runs tests locally. Worth confirming `npm test` passes on beta.13 before merging, and if the package is ESM, adding `transformIgnorePatterns: ['/node_modules/(?!@rootcodelabs/skapp-ui)']` to the Jest config.

PR #1777, opened 2026-08-17
fix: leave policy bug fixes

thusala:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR extracts the policy leave balance calculation into a new `PolicyLeaveBalanceService` and surfaces balances on the assigned-policy cards. The extraction is not behaviour-preserving: the moved `calculateBalance` gained a `creditedUpTo` parameter that changes `balanceInDays` from full-cycle allocation to accrued-to-date and clamps it at zero, which silently alters the existing apply-leave balance cards and the apply/availability validation, and returns a zero balance for the next-year option the UI already offers. The frontend accrual preview also switched to per-event half-day rounding, which systematically diverges from the backend's round-once-at-the-end accrual, and the assign modal's join-date button is now mislabelled "Save".

Found **31** new issue(s): 🔴 8 important, 🟡 15 suggestion(s), 🟣 8 nit(s)

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveBalanceServiceImpl.java
```diff
@@ -0,0 +1,122 @@
...
+		PolicyLeaveBalanceDto balance = new PolicyLeaveBalanceDto();
+		balance.setPolicy(policy);
+		balance.setEffectiveFrom(effectiveFrom);
+		balance.setCycleStart(cycle.getStartDate());
+		balance.setCycleEnd(cycle.getEndDate());
+		balance.setUsableFrom(resolveUsableFrom(effectiveFrom, cycle));
+		balance.setAccruedDays(accruedDays);
+		balance.setTotalDaysAllocated(totalDaysAllocated);
+		balance.setTotalDaysUsed(totalDaysUsed);
+		balance.setBalanceInDays(Math.max(0f, usableDays - totalDaysUsed));
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Not a pure extraction — the balance semantics changed for the existing apply-leave flow. Old code: `accruedDays` was computed with `asOf = cycle.getEndDate()` and `balanceInDays = totalDaysAllocated - totalDaysUsed` (full-cycle allocation, allowed to go negative). New code: `accruedDays` uses `creditedUpTo`, `usableDays = cap(accruedDays)`, and `balanceInDays = Math.max(0f, usableDays - totalDaysUsed)`. Two consequences on screens this PR did not intend to touch: (1) `getCurrentUserPolicyBalances` -> `toBalanceCard` now feeds `balanceInDays` = accrued-so-far into `LeavePolicyCard`, so an employee on a monthly-accrual policy sees ~1/12 of what the card showed yesterday, and the `LOW_BALANCE_WARNING_DAYS` warning fires for most of the year; (2) `checkPolicyLeaveAvailability`/`applyPolicyLeaveRequest` now reject requests that draw on not-yet-accrued days, which previously succeeded. The `Math.max(0f, ...)` clamp additionally masks over-drawn balances, so the `LEAVE_ERROR_POLICY_LEAVE_INSUFFICIENT_BALANCE` message reports "0.0" instead of the real deficit. If accrue-as-you-go is the intended product behaviour, say so in the PR and update the balance-card/insufficient-balance messaging; if not, pass `cycle.getEndDate()` as `creditedUpTo` from `calculateBalanceForYear` so the existing flow is preserved and only the new card consumes the accrued-to-date figure.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveBalanceServiceImpl.java
```diff
@@ -0,0 +1,122 @@
...
+@Service
+@RequiredArgsConstructor
+public class PolicyLeaveBalanceServiceImpl implements PolicyLeaveBalanceService {
+
+	private final PolicyLeaveRequestDao policyLeaveRequestDao;
+
+	@Override
+	@Transactional(readOnly = true)
+	public PolicyLeaveBalanceDto calculateBalanceForYear(EmployeeLeavePolicy assignment, int year) {
+		return calculateBalance(assignment, PolicyLeaveAccrualUtil.resolveCycle(year),
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`calculateBalanceForYear` hardcodes `creditedUpTo = DateTimeUtils.getCurrentUtcDate()`, which breaks the future-year case. `PolicyLeaveController.getCurrentUserPolicyBalances` takes an unbounded `@RequestParam Integer year`, and the UI year selector (`getAdjacentYearsWithCurrent` in dateTimeUtils.ts returns [currentYear-1, currentYear, currentYear+1]) lets a user pick next year. For year = currentYear+1, `accruedWithinCycle` computes `windowEnd = min(today, cycleEnd) = today`, which is before `cycle.getStartDate()`, so it returns 0 -> `accruedDays = 0`, `usableDays = 0`, `balanceInDays = 0`, while `totalDaysAllocated` is still the full cycle. The card renders "0 / 12" and `resolveDisabledReason` returns `FULLY_UTILIZED`, disabling next-year leave entirely; before this PR it showed the full allocation and was actionable. Clamp `creditedUpTo` to the cycle, e.g. `LocalDate creditedUpTo = today.isBefore(cycle.getStartDate()) ? cycle.getEndDate() : today;`, or pass the cycle end for any year other than the current one.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
```diff
@@ -490,7 +492,7 @@ private void validateReviewerComment(String reviewerComment) {
 	private EmployeePolicyBalanceResponseDto toBalanceCard(EmployeeLeavePolicy assignment, int year,
 			boolean hasSupervisor, LocalDate today) {
 		LeavePolicy policy = assignment.getPolicy();
-		PolicyLeaveBalanceDto balance = calculateBalanceForYear(assignment, year);
+		PolicyLeaveBalanceDto balance = policyLeaveBalanceService.calculateBalanceForYear(assignment, year);
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`resolveDisabledReason` now mislabels un-accrued policies. With `balanceInDays` switched to accrued-to-date, a freshly assigned policy (or one whose first credit date has not yet passed, e.g. `AccrualTiming.PERIOD_END` monthly in the first week of a period) has `balanceInDays == 0`, so `hasBalance(balance)` is false and the card is disabled with `PolicyBalanceDisabledReason.FULLY_UTILIZED` — telling the employee they have used up leave they have never accrued. Add a distinct reason (e.g. `NOT_YET_ACCRUED` / reuse the `usableFrom` check) for the case where `totalDaysUsed == 0` but `accruedDays == 0`, and map it to appropriate copy on the frontend.

frontend/src/community/leave/utils/accrualPreviewUtils.ts
```diff
@@ -135,11 +136,12 @@ const toPreviewRows = (
   const rows: AccrualPreviewRow[] = [];
   let balance = 0;
   for (const event of accrualEvents) {
-    balance = roundToTwoDecimals(balance + event.days);
+    const days = roundToHalfDay(event.days);
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The new per-event half-day rounding makes the preview diverge from what the backend actually credits, and the comment on line 30-31 asserting parity is incorrect. `PolicyLeaveAccrualUtil.accruedUpTo` sums the raw `accrualDays` for every period and `PolicyLeaveBalanceServiceImpl` calls `roundToHalfDay` once on the total; this code rounds every event and then rounds the running balance again. For `accrualDays = 1.2` monthly, each event rounds to 1.0 (Math.round(2.4)/2), so after 12 periods the preview shows 12 days while the backend credits round(14.4) = 14.5 — a 2.5-day error that grows with the number of rows. Keep the un-rounded running sum and only round for display, i.e. accumulate `balance += event.days` and render `roundToHalfDay(balance)`, or drop back to `roundToTwoDecimals`. Note `firstPeriodDays` already rounds on line 59, so the extra `roundToHalfDay(event.days)` on line 139 is also redundant.

frontend/src/community/leave/components/molecules/AssignLeavePolicyModal/AssignLeavePolicyModal.tsx
```diff
@@ -254,13 +262,16 @@ const AssignLeavePolicyModal: FC<Props> = ({
                 variant: "primary",
                 onClick: () => setIsSetJoinDateOpen(true),
                 disabled: isEmployeeLoading,
-                children: translateText(["assignModal", "setJoinDateBtnTxt"])
+                children: translateText(["assignModal", "saveBtnTxt"])
```
ThinuwanW, line 264: 🤖 **Claude** · 🔴 **Important**

The `needsJoinDate` branch button label was changed from `setJoinDateBtnTxt` ("Set a join date") to `saveBtnTxt` ("Save"), but its `onClick` is still `() => setIsSetJoinDateOpen(true)` — it opens the Set Join Date modal and does not save the assignment. Both branches now render an identical "Save" button with completely different behaviour. Restore a distinct label for this branch (the `setJoinDateBtnTxt` key was removed from leaveModule.json in this PR and would need to be re-added), or make the button actually save.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/EmployeeLeavePolicyServiceImpl.java
```diff
@@ -202,9 +207,14 @@ public ResponseEntityDto getEmployeeLeavePolicies(Long employeeId, EmployeeLeave
 			.findByEmployee_EmployeeIdAndStatusOrderByEffectiveFromDesc(employeeId, EmployeeLeavePolicyStatus.ACTIVE,
 					pageable);
 
+		int currentYear = DateTimeUtils.getCurrentUtcDate().getYear();
+		List<EmployeeLeavePolicyResponseDto> items = activeEmployeeLeavePolicies.getContent()
+			.stream()
+			.map(assignment -> toEmployeeLeavePolicyWithBalance(assignment, currentYear))
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

N+1 query introduced on a list endpoint. `toEmployeeLeavePolicyWithBalance` is called per row and each call runs `policyLeaveRequestDao.sumCommittedDaysForPolicyInCycle`, plus lazy loads of `assignment.getEmployee()`/`getPolicy()`/`getLeaveType()`. `UserLeavePolicies` requests 6 per page, but `AssignLeavePolicyModal` calls `useGetEmployeeLeavePolicies(employeeId, 0, UNPAGINATED_SIZE, isOpen)` with `UNPAGINATED_SIZE = -1`, which hits the `Pageable.unpaged()` branch on line 204 — so opening the assign modal now fires one balance query per assigned policy purely to compute a conflict warning that never reads the balance fields. Consider a single batched `sumCommittedDaysForPoliciesInCycle(employeeId, policyIds, ...)` query returning a policyId -> days map, and/or skipping balance enrichment when the caller does not need it.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveBalanceServiceImpl.java
```diff
@@ -0,0 +1,122 @@
...
+
+@Service
+@RequiredArgsConstructor
+public class PolicyLeaveBalanceServiceImpl implements PolicyLeaveBalanceService {
+
+	private final PolicyLeaveRequestDao policyLeaveRequestDao;
+
+	@Override
+	@Transactional(readOnly = true)
+	public PolicyLeaveBalanceDto calculateBalanceForYear(EmployeeLeavePolicy assignment, int year) {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

New service class has no `@Slf4j` and no entry logging. The repo convention (and the review guidelines) is `log.info("methodName: execution started")` / `"...execution ended"` on service methods — see `PolicyLeaveServiceImpl` and `EmployeeLeavePolicyServiceImpl`. Add `@Slf4j` to the class and entry/exit logs to `calculateBalanceForYear` and `calculateBalanceForDate`.

frontend/src/community/leave/types/LeavePolicyTypes.ts
```diff
@@ -234,6 +234,10 @@ export interface EmployeeLeavePolicyType {
   effectiveDateType: EffectiveDateType;
   effectiveFrom: string;
   status: EmployeeLeavePolicyStatus;
+  totalDaysAllocated: number;
```
ThinuwanW, line 237: 🤖 **Claude** · 🟡 **Suggestion**

The four new fields are declared as required (`number` / `boolean`), but the backend only populates them in `getEmployeeLeavePolicies`. `assignLeavePolicy` and `unassignLeavePolicy` return the same `EmployeeLeavePolicyResponseDto` straight from the mapper (EmployeeLeavePolicyServiceImpl lines 100 and 196), so those responses serialise `totalDaysAllocated`/`totalDaysUsed`/`balanceInDays`/`isUnlimited` as `null`. Anything that renders an assignment from a mutation response would hit `formatDays(undefined)` -> `Cannot read properties of undefined (reading 'toFixed')`. Either mark the fields optional here and guard in `LeavePolicyCard`, or populate them in the assign/unassign responses too so the DTO contract is uniform.

frontend/src/community/leave/components/molecules/AssignLeavePolicyModal/AssignLeavePolicyModal.tsx
```diff
@@ -254,13 +262,16 @@ const AssignLeavePolicyModal: FC<Props> = ({
...
               }
             : {
                 variant: "primary",
                 onClick: handleSave,
                 disabled: isSaveDisabled,
                 isLoading: isPending,
+                title: selectedPolicyId
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The `title` attribute is only set when `!selectedPolicyId`, which is exactly the condition that makes `isSaveDisabled` true. Chromium suppresses pointer events (and therefore the native title tooltip) on disabled form controls, so `saveDisabledTooltip` will never be shown to most users. This PR already imports the `Tooltip` component from `@rootcodelabs/skapp-ui` in AssignLeavePolicyForm — use that with a non-disabled wrapper element instead of the native `title`, so the hint is consistent with the rest of the modal and actually renders.

frontend/src/community/leave/components/molecules/AssignLeavePolicyModal/AssignLeavePolicyModal.tsx
```diff
@@ -158,6 +158,13 @@ const AssignLeavePolicyModal: FC<Props> = ({
     });
   }, [selectedPolicy, existingAssignmentsPage, employeeSubject, translateText]);
 
+  const joinDateWarning =
+    !isEmployeeLoading && !joinedDate
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`joinDateWarning` is computed from `!isEmployeeLoading && !joinedDate` only, unlike `needsJoinDate` (line 58) which also requires `effectiveDateType === EffectiveDateType.JOIN_DATE`. So after the user follows the advice and switches to "Specific Date", the banner still tells them to "choose a specific date instead". Gate it on the same condition as `needsJoinDate` (or reuse `needsJoinDate` directly).

frontend/src/community/leave/components/molecules/AssignLeavePolicyModal/AssignLeavePolicyForm.tsx
```diff
@@ -95,123 +96,139 @@ const AssignLeavePolicyForm: FC<Props> = ({
...
+        {isFlexiblePolicy && (
+          <InfoTipBanner
+            status="info"
+            description={translateText(["assignModal", "flexibleInfoLabel"])}
+          />
+        )}
+
+        {accrualPreview.length > 0 && (
+          <div className="flex flex-col gap-2">
+            <Tooltip
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The accrual-preview `Tooltip` wraps a plain `<p>`, which is neither focusable nor described by anything. Keyboard and screen-reader users can never reach `accrualPreviewTooltip`. Attach the tooltip to a focusable info affordance (e.g. an icon button with an `aria-label`), or render the text as an `aria-describedby`-linked description instead of a hover-only tooltip.

frontend/src/community/leave/components/molecules/AssignLeavePolicyModal/AssignLeavePolicyForm.tsx
```diff
@@ -95,123 +96,139 @@ const AssignLeavePolicyForm: FC<Props> = ({
...
-            <RadioButton
-              isSelected={effectiveDateType === EffectiveDateType.JOIN_DATE}
-              variant="dot"
-            />
-            <span className="body1 flex items-center gap-1.5 text-black">
-              {translateText(["assignModal", "joinDateOption"])}
-              {joinDateLabel && (
-                <span className="body2 text-secondary-text">
-                  ({joinDateLabel})
+            <Tooltip
```
ThinuwanW, line 121: 🤖 **Claude** · 🟣 **Nit**

The `Tooltip` wrapper is inserted between the `role="radiogroup"` container and the `role="radio"` button, so the radio is no longer a direct child of the group. ARIA requires radios to be owned by their radiogroup; the intervening generic element can break the group/position-in-set relationship for some assistive tech. Either move the tooltip inside the button (around the label span) or add `aria-owns`/restructure so both radios remain direct children. Also worth noting the asymmetry: only the join-date option gets a tooltip, the specific-date option gets none.

frontend/src/community/common/utils/commonUtil.ts
```diff
@@ -623,4 +623,4 @@ export const concatStrings = (args: string[], separator: string = " ") =>
   args.join(separator);
 
 export const formatDays = (value: number): string =>
-  Number.isInteger(value) ? String(value) : value.toFixed(2);
+  Number.isInteger(value) ? String(value) : value.toFixed(1);
```
ThinuwanW, line 626: 🤖 **Claude** · 🟣 **Nit**

`formatDays` is a shared `commonUtil` export, but its precision is being changed from 2dp to 1dp to suit the single current call site (`LeavePolicyCard`). Since the backend rounds to half days, 1dp is right for policy balances, but a future caller with quarter-day values would silently get "1.3" for 1.25. Consider naming the intent (e.g. `formatHalfDays`) or accepting a precision argument so the shared helper does not encode one screen's assumption.

frontend/src/community/leave/components/molecules/UserLeavePolicies/LeavePolicyCard.tsx
```diff
@@ -46,39 +39,36 @@ const LeavePolicyCard: FC<Props> = ({
...
-                {formatDays(usage.remaining)}
-              </span>
-              <span className="body2 text-secondary-text">
-                /{formatDays(usage.total)}
-              </span>
-            </div>
-          )
+          <div
+            className="flex shrink-0 items-baseline gap-0.5"
+            aria-label={translateText(["leavesRemainingLabel"], {
```
ThinuwanW, line 41: 🤖 **Claude** · 🟣 **Nit**

The visible numbers go through `formatDays` but the `aria-label` interpolates the raw values, so a screen reader announces "1.5 / 12 days remaining" vs "1.5 / 12" today but would announce e.g. "1.25" if the backend ever returns a quarter day while the text reads "1.3". Pass `formatDays(policy.balanceInDays)` / `formatDays(policy.totalDaysAllocated)` into `translateText` so the announced and rendered values always match.

frontend/src/community/common/assets/languages/english/leaveModule.json
```diff
@@ -495,35 +495,36 @@
...
+      "accrualPreviewTitle": "Accrual Preview",
+      "accrualPreviewTooltip": "This is a projection based on the selected policy and effective date. Actual balances may vary.",
       "flexibleInfoLabel": "This is a flexible policy. Leave taken will be recorded but is not deducted from any balance.",
       "conflictWarning": "{{employeeName}} already has a {{leaveType}} policy. Assigning this policy will replace their existing {{leaveType}} policy.",
-      "colDate": "Date",
-      "colDays": "Days",
-      "colBalance": "Balance",
-      "setJoinDateBtnTxt": "Set a join date",
+      "joinDateMissingLabel": "{{employeeName}} doesn't have a join date yet. Set one to accrue leave from their join date, or choose a specific date instead.",
+      "colDate": "DATE",
```
ThinuwanW, line 527: 🤖 **Claude** · 🟣 **Nit**

`colDate`/`colDays`/`colBalance` bake ALL-CAPS presentation into the translation values. Casing is locale-specific (and some scripts have no case), so this belongs in CSS (`uppercase` on the TableView header) with sentence-case copy in the JSON. Related: this PR moves `sectionTitle` to Title Case ("Leave Policies") while moving `assignModal.title` to sentence case ("Assign leave policy") — worth settling on one convention.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/EmployeeLeavePolicyServiceImpl.java
```diff
@@ -202,9 +207,14 @@ public ResponseEntityDto getEmployeeLeavePolicies(Long employeeId, EmployeeLeave
...
+			.toList();
+
 		PageDto pageDto = new PageDto();
-		pageDto.setItems(leaveMapper
-			.employeeLeavePolicyListToEmployeeLeavePolicyResponseDtoList(activeEmployeeLeavePolicies.getContent()));
+		pageDto.setItems(items);
```
ThinuwanW, line 219: 🤖 **Claude** · 🟣 **Nit**

Replacing the list-mapper call leaves `LeaveMapper.employeeLeavePolicyListToEmployeeLeavePolicyResponseDtoList` with no callers anywhere in the backend (verified by grep — only the interface declaration remains). Remove the now-dead mapper method.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveBalanceServiceImpl.java
```diff
@@ -0,0 +1,122 @@
+package com.skapp.community.leaveplanner.service.impl;
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

No tests accompany this change. There is no unit test for the new `PolicyLeaveBalanceServiceImpl` (the only leave-policy tests are the controller integration tests, which do not assert balance fields), and no frontend test for the reworked `roundToHalfDay` in accrualPreviewUtils or for `LeavePolicyCard` now that the `usage` prop guard was removed. Given the semantic changes to `balanceInDays`/`accruedDays` described above, please add cases covering: current year mid-cycle, past year, next year, effective-from after cycle end, flexible policy, and used > accrued.

frontend/src/community/leave/components/molecules/IndividualEmployeeLeaveReportSection/IndividualEmployeeLeaveReportSection.tsx
```diff
@@ -76,8 +76,6 @@ const IndividualEmployeeLeaveReportSection: FC<Props> = ({
     >
       <UpgradeOverlay customContainerStyles={classes.customContainerStyles}>
         <>
-          <h2 className="h2 text-black">{translateText(["pageHead"])}</h2>
-
           {!isLeavePolicyConfigLoading &&
```
ThinuwanW, line 79: 🤖 **Claude** · 🔴 **Important**

Deleting the `<h2>{translateText(["pageHead"])}</h2>` removes the only visible heading on this page. `PeopleLayout` is called with `title={""}` and only renders `pageHead` inside `<Head><title>` (PeopleLayout.tsx:57) — it is never a visible/DOM heading. I grepped the three sibling sections rendered here (`UserAssignedLeaveTypes`, `UserLeaveUtilization`, `UserLeaveHistory`) and none of them contains any `<h1>`–`<h4>` element. Consequences: (1) when `isLeavePoliciesEnabled` is false the whole tab now has zero heading elements — screen-reader heading navigation and the document outline are broken and the visible "Leave" section label disappears for orgs still on the legacy leave-types flow; (2) even on the policies path there is no heading at all while `isLeavePolicyConfigLoading` is true, because the promoted `<h2>Leave Policies</h2>` in `UserLeavePolicies.tsx:54` is inside the loading-gated branch. Fix: keep the page-level `<h2>{translateText(["pageHead"])}</h2>` outside the loading gate and revert `UserLeavePolicies` back to `<h3>` so the outline stays h2 → h3, or give `UserAssignedLeaveTypes` its own `<h2>` and render the page heading unconditionally.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/EmployeeLeavePolicyServiceImpl.java
```diff
@@ -202,9 +207,14 @@ public ResponseEntityDto getEmployeeLeavePolicies(Long employeeId, EmployeeLeave
 			.findByEmployee_EmployeeIdAndStatusOrderByEffectiveFromDesc(employeeId, EmployeeLeavePolicyStatus.ACTIVE,
 					pageable);
 
+		int currentYear = DateTimeUtils.getCurrentUtcDate().getYear();
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The leave cycle is now selected with `DateTimeUtils.getCurrentUtcDate().getYear()`, and `PolicyLeaveBalanceServiceImpl.calculateBalanceForYear` derives `creditedUpTo` from `DateTimeUtils.getCurrentUtcDate()` as well. Both are UTC-based, but the rest of this module resolves "now" against the organization timezone (see `PolicyLeaveServiceImpl.calculateRequestedDays`, which passes `organizationService.getOrganizationTimeZone()`). Two concrete failures: (a) an org at UTC+13 opening this page on 1 Jan local time is still 31 Dec in UTC, so every card shows last year's cycle balance; (b) because `creditedUpTo` now actually changes the accrued amount (it no longer defaults to the cycle end), a PERIOD_END monthly policy credits a day late or early relative to local month-end. Resolve the current date through the organization timezone in both places, the same way `calculateRequestedDays` does.

frontend/src/community/leave/components/molecules/UserLeavePolicies/LeavePolicyCard.tsx
```diff
@@ -46,39 +39,36 @@ const LeavePolicyCard: FC<Props> = ({
...
+              remaining: policy.balanceInDays,
+              total: policy.totalDaysAllocated
+            })}
+          >
+            <span className="h1b text-black">
+              {formatDays(policy.balanceInDays)}
```
ThinuwanW, line 47: 🤖 **Claude** · 🔴 **Important**

There is no "balance unavailable" state on this card. `PolicyLeaveBalanceServiceImpl.emptyBalance()` returns all-zero figures whenever `cycle.getEndDate().isBefore(effectiveFrom)` (an assignment whose effective date falls in a future cycle), so such a policy renders "0 / 0" — visually identical to a policy the employee has fully consumed. The existing sibling card handles exactly this: `PolicyLeaveBalanceCard.tsx` reads `policyBalance.isBalanceAvailable` and routes through `getPolicyBalanceLabel` (policyLeaveUtils.ts:114) which renders "—" when the balance is not derivable. That signal exists on the backend (`PolicyLeaveBalanceDto.isDerived`, surfaced as `isBalanceAvailable` on `EmployeePolicyBalanceResponseDto`) but was not carried onto `EmployeeLeavePolicyResponseDto`. Add an `isBalanceAvailable` field to `EmployeeLeavePolicyResponseDto` (set from `balance.isDerived()` in `toEmployeeLeavePolicyWithBalance`) and reuse `getPolicyBalanceLabel` here instead of calling `formatDays` unconditionally.

frontend/src/community/leave/components/molecules/UserLeavePolicies/LeavePolicyCard.tsx
```diff
@@ -46,39 +39,36 @@ const LeavePolicyCard: FC<Props> = ({
...
-            >
-              <span className="text-2xl text-black">
-                {formatDays(usage.remaining)}
-              </span>
-              <span className="body2 text-secondary-text">
-                /{formatDays(usage.total)}
-              </span>
-            </div>
-          )
+          <div
```
ThinuwanW, line 39: 🤖 **Claude** · 🟡 **Suggestion**

The PR adds `isUnlimited` and `totalDaysUsed` to both the backend DTO and `EmployeeLeavePolicyType`, but this card never reads either one — `totalDaysUsed` is entirely dead payload, and the unlimited case is still decided locally by `policy.policyType === PolicyType.FLEXIBLE` (line 28). That is two sources of truth for the same fact: the backend derives `isUnlimited` from `PolicyType.FLEXIBLE` in `PolicyLeaveBalanceServiceImpl.calculateBalance`, so the moment that rule gains a second condition (or a FLEXIBLE policy is ever assigned a tracked balance) the card silently falls into the numeric branch and renders "0 / 0". Switch line 28 to `const isFlexiblePolicy = policy.isUnlimited;` and either consume `totalDaysUsed` (e.g. in the aria-label) or drop it from the DTO and the type.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveBalanceServiceImpl.java
```diff
@@ -0,0 +1,122 @@
...
+
+		if (cycle.getEndDate().isBefore(effectiveFrom)) {
+			return emptyBalance(policy, effectiveFrom, cycle);
+		}
+
+		LocalDate accrualStartDate = PolicyLeaveAccrualUtil.resolveAccrualStartDate(policy, effectiveFrom);
+
+		float accruedDays = PolicyLeaveAccrualUtil
+			.roundToHalfDay(PolicyLeaveAccrualUtil.accruedWithinCycle(policy, accrualStartDate, cycle, creditedUpTo));
+		float accruedDaysForCycle = PolicyLeaveAccrualUtil.roundToHalfDay(
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`calculateBalance` now runs two `accruedWithinCycle` calls (`accruedDays` at line 57 and `accruedDaysForCycle` at line 59), and each of those internally invokes `PolicyLeaveAccrualUtil.accruedUpTo` twice (once for `windowEnd`, once for `cycle.getStartDate().minusDays(1)`). That is four period walks per assignment where the old code did two. For an `AccrualFrequency.DAILY` policy each walk iterates roughly 365 times, so this is ~1,460 loop iterations per assignment — and it now runs once per row of a paginated list plus once per assignment in the assign-modal's `size=-1` (unpaged) fetch. Short-circuit the second walk: when `!creditedUpTo.isBefore(cycle.getEndDate())` the two values are identical, so set `accruedDaysForCycle = accruedDays` instead of recomputing.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveBalanceServiceImpl.java
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`balanceInDays` is now clamped with `Math.max(0f, usableDays - totalDaysUsed)`; the extracted original was the unclamped `totalDaysAllocated - totalDaysUsed`. This silently hides over-drawn balances, which are reachable (approved leave later in the cycle counts fully toward `totalDaysUsed` while accrual is only credited up to `creditedUpTo`, and `applyAccrualCap` can lower `usableDays` if the cap is edited downward). Two visible consequences: the insufficient-balance error interpolates `balance.getBalanceInDays()` (`PolicyLeaveServiceImpl:596`) so the user is told "0 days remaining" instead of the real deficit, and `resolveDisabledReason`/`hasBalance` can no longer distinguish "exactly used up" from "over-drawn". If the clamp is intentional, keep the raw value on the DTO and clamp only at the presentation boundary; otherwise drop the `Math.max` and state the change in the PR description.

frontend/src/community/leave/components/molecules/AssignLeavePolicyModal/AssignLeavePolicyForm.tsx
```diff
@@ -95,123 +96,139 @@ const AssignLeavePolicyForm: FC<Props> = ({
...
-        >
-          <button
-            type="button"
-            role="radio"
-            aria-checked={effectiveDateType === EffectiveDateType.JOIN_DATE}
-            onClick={() =>
-              onEffectiveDateTypeChange(EffectiveDateType.JOIN_DATE)
-            }
-            className="flex w-fit cursor-pointer items-center gap-3"
+      <div className="flex max-h-[55vh] flex-col gap-4 overflow-y-auto pr-2">
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The new inner scroll box `max-h-[55vh] ... overflow-y-auto` now encloses two positioned overlays that were previously free of it, or that the PR deliberately moved out. The `Dropdown` and the conflict banner were correctly hoisted above this container, but the `DatePicker` with `popperProps={{ position: "bottom-start" }}` (line 158) and the accrual-preview `Tooltip position="top"` (line 210) are still inside it. Unless skapp-ui portals these to `document.body`, an `overflow-y-auto` ancestor clips them: the date input sits near the bottom of a 55vh box so a `bottom-start` popper opens into the clipped region, and the preview tooltip opens upward from an element the user has scrolled to. Please verify both components portal; if they do not, hoist the datepicker out of the scroll container or give it a portal target.

backend/src/main/java/com/skapp/community/leaveplanner/service/PolicyLeaveBalanceService.java
```diff
@@ -0,0 +1,14 @@
+package com.skapp.community.leaveplanner.service;
+
+import com.skapp.community.leaveplanner.model.EmployeeLeavePolicy;
+import com.skapp.community.leaveplanner.payload.PolicyLeaveBalanceDto;
+
+import java.time.LocalDate;
+
+public interface PolicyLeaveBalanceService {
+
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The new public service contract takes the JPA entity `EmployeeLeavePolicy` as its parameter, and the impl is annotated `@Transactional(readOnly = true)`. That annotation advertises the method as safely callable standalone, but it is not: `calculateBalance` dereferences `assignment.getPolicy()` and `totalDaysUsedInCycle` dereferences `assignment.getEmployee()`, so a caller outside an active persistence context passing a detached entity gets a `LazyInitializationException`, not a clean result. It works today only because both call sites (`EmployeeLeavePolicyServiceImpl.getEmployeeLeavePolicies` and the three `PolicyLeaveServiceImpl` methods) already run inside a transaction. Either change the signature to take the assignment id and load the entity inside the transactional method, or document on the interface that the caller must supply a managed entity within an open transaction.

frontend/src/community/leave/components/molecules/UserLeavePolicies/LeavePolicyCard.tsx
```diff
@@ -35,7 +28,7 @@ const LeavePolicyCard: FC<Props> = ({
   const isFlexiblePolicy = policy.policyType === PolicyType.FLEXIBLE;
 
   return (
-    <Card className="flex h-full flex-row items-center justify-between gap-4 bg-white p-6!">
+    <Card className="flex h-[96px] flex-row items-center justify-between gap-4 bg-white p-6!">
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

There are now two distinct components named `LeavePolicyCard` under `~community/leave/components/molecules/` — this one and `molecules/LeavePolicyCard/LeavePolicyCard.tsx` — and after this PR they render the same concept: leave type name, policy name, emoji, and `balance / totalDaysAllocated` with an unlimited variant. The other one already routes its balance text through the shared `getPolicyBalanceLabel` helper (policyLeaveUtils.ts:114), which handles the unlimited and not-available cases; this one re-implements that inline. Beyond the duplicated markup the identical filenames make imports and stack traces ambiguous. Rename this file to something scoped (e.g. `AssignedLeavePolicyCard.tsx`) and reuse `getPolicyBalanceLabel` for the balance text rather than duplicating the branching.

ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`h-full` was replaced with a hard `h-[96px]`, while the parent grid still uses `auto-rows-fr` (UserLeavePolicies.tsx:84) which already equalises row heights — so the fixed height buys nothing and introduces a clipping risk. With `p-6!` (48px vertical padding) only ~48px of content box remains, and the left column now stacks two lines with `gap-2` plus a `size-5 text-xl leading-5` emoji. At 200% browser zoom or with a user font-size override the second line is cut off, and `truncate` does not help with vertical overflow. Use `min-h-[96px]` (or drop the height and keep `h-full` with `auto-rows-fr`).

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/EmployeeLeavePolicyServiceImpl.java
ThinuwanW, line 219: 🤖 **Claude** · 🟡 **Suggestion**

`getEmployeeLeavePolicies` never calls a feature-flag guard, unlike every read path in `PolicyLeaveServiceImpl`, each of which begins with `requireLeavePoliciesEnabled()` (`leavePolicyService.isLeavePoliciesEnabled()`). That gap was tolerable while this endpoint returned only assignment metadata; now that it computes and returns accrual balances it is serving the same class of data as `getCurrentUserPolicyBalances`, which is flag-guarded. Add `requireLeavePoliciesEnabled()`-equivalent gating here (via the injected `LeavePolicyService`) so the two balance-producing endpoints behave consistently when the org has leave policies switched off.

frontend/src/community/leave/components/molecules/AssignLeavePolicyModal/AssignLeavePolicyForm.tsx
```diff
@@ -75,13 +78,11 @@ const AssignLeavePolicyForm: FC<Props> = ({
...
-    <div className="flex max-h-[73vh] flex-col gap-4 overflow-y-auto pr-2">
-      <div className="flex flex-col gap-1.5">
-        <p className="body2 text-secondary-text">
-          {translateText(["assignModal", "policyLabel"])}
-        </p>
+    <div className="flex flex-col gap-4">
+      <div className="flex flex-col gap-1.5"> 
         <Dropdown
           id="assign-leave-policy-dropdown"
+          label={translateText(["assignModal", "policyLabel"])}
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`label` and `ariaLabel` are both set to the same `policyLabel` string. Now that the `Dropdown` renders a visible label, `aria-label` overrides it as the accessible name — harmless while the strings match, but it makes the two silently diverge-able and defeats WCAG 2.5.3 (Label in Name) for speech-input users the moment they do. Drop the `ariaLabel` prop and let the visible `label` provide the accessible name.

frontend/src/community/leave/components/molecules/AssignLeavePolicyModal/AssignLeavePolicyForm.tsx
```diff
@@ -75,13 +78,11 @@ const AssignLeavePolicyForm: FC<Props> = ({
   );
 
   return (
-    <div className="flex max-h-[73vh] flex-col gap-4 overflow-y-auto pr-2">
-      <div className="flex flex-col gap-1.5">
-        <p className="body2 text-secondary-text">
-          {translateText(["assignModal", "policyLabel"])}
-        </p>
+    <div className="flex flex-col gap-4">
+      <div className="flex flex-col gap-1.5">
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

The `<div className="flex flex-col gap-1.5">` wrapper now has a single child — the `<p>` label it used to space was folded into the `Dropdown`'s `label` prop — so the `flex flex-col gap-1.5` classes are dead. (The line also carries trailing whitespace after the `>`.) Remove the wrapper and render the `Dropdown` directly under the outer `flex flex-col gap-4` container.

frontend/src/community/common/assets/languages/english/leaveModule.json
```diff
@@ -495,35 +495,36 @@
     "noPoliciesYetTitle": "No leave policies yet. Create your first policy to get started."
   },
   "leavePolicyAssignment": {
-    "sectionTitle": "Leave policies",
+    "sectionTitle": "Leave Policies",
```
ThinuwanW, line 502: 🤖 **Claude** · 🟣 **Nit**

The two casing changes in this PR pull in opposite directions: `sectionTitle` goes "Leave policies" → "Leave Policies" (Title Case) while `assignModal.title` goes "Assign Leave Policy" → "Assign leave policy" (sentence case), in the same feature and the same file. Pick one convention for headings and modal titles and apply it to both so the surface reads consistently.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveBalanceServiceImpl.java
thusala: Accrue-as-you-go is the intended behaviour here - that is what the `fix: balance calculation issue` commit is for, so the apply-leave flow should also stop granting days that have not been credited yet. The one genuine regression you spotted, the next-year card, is fixed in 4bbcb3cd2: a cycle that starts in the future is now credited up to its cycle end instead of to today.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveBalanceServiceImpl.java
thusala: Fixed in 4bbcb3cd2 - `calculateBalanceForYear` now clamps `creditedUpTo` to the cycle end when the cycle starts after today, so next year renders the full allocation and stays actionable.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
thusala: Not fixing here - a `NOT_YET_ACCRUED` reason needs a new enum value plus new frontend copy, which is a design/copy call outside this bugfix. Raising it as a follow-up.

frontend/src/community/leave/utils/accrualPreviewUtils.ts
thusala: Fixed in 4bbcb3cd2 - the running total is accumulated unrounded and only rounded to a half day for display, the per-event days column is back on `roundToTwoDecimals`, and the misleading parity comment is corrected.

frontend/src/community/leave/components/molecules/AssignLeavePolicyModal/AssignLeavePolicyModal.tsx
thusala: Intentional - the modal now has a single Save CTA, and the inline warning banner tells the user the join date has to be set first, so the button opens that step. The banner carries that state rather than a second button wording.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/EmployeeLeavePolicyServiceImpl.java
thusala: Agreed in principle, but batching needs a new `sumCommittedDaysForPoliciesInCycle` DAO query and a caller-side opt-out flag - too much for this bugfix.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveBalanceServiceImpl.java
thusala: Moot now - the separate `PolicyLeaveBalanceService` is gone in 4bbcb3cd2 and the balance calculation lives back in `PolicyLeaveServiceImpl`, which already has `@Slf4j` and entry/exit logging on its public methods.

frontend/src/community/leave/types/LeavePolicyTypes.ts
thusala: Not an actual crash path - neither `useAssignLeavePolicy` nor `useUnassignLeavePolicy` renders the mutation response, they only invalidate the query keys, so these fields are always read from `getEmployeeLeavePolicies`, which populates them.

frontend/src/community/leave/components/molecules/AssignLeavePolicyModal/AssignLeavePolicyModal.tsx
thusala: Fixed in 4bbcb3cd2 by removing the dead `title` and its unused `saveDisabledTooltip` key. `SmallModal`'s `buttons` prop only takes plain button props, so a `Tooltip` wrapper is not available there.

frontend/src/community/leave/components/molecules/AssignLeavePolicyModal/AssignLeavePolicyModal.tsx
thusala: Fixed in 4bbcb3cd2 - `joinDateWarning` is now derived from `needsJoinDate`, so it disappears as soon as the user switches to Specific Date.

frontend/src/community/leave/components/molecules/AssignLeavePolicyModal/AssignLeavePolicyForm.tsx
thusala: Fixed in 4bbcb3cd2 - the title is now focusable (`tabIndex={0}`), so `Tooltip`'s `onFocus` / `aria-describedby` wiring reaches keyboard and screen-reader users.

frontend/src/community/leave/components/molecules/AssignLeavePolicyModal/AssignLeavePolicyForm.tsx
thusala: Fixed in 4bbcb3cd2 - the `Tooltip` moved inside the button around the label span, so both radios are direct children of the radiogroup again.

frontend/src/community/common/utils/commonUtil.ts
thusala: `formatDays` has exactly one call site (`LeavePolicyCard`) and the backend rounds policy balances to half days, so 1dp is correct for every value it can currently receive. Happy to add a precision argument the moment a second caller needs one.

frontend/src/community/leave/components/molecules/UserLeavePolicies/LeavePolicyCard.tsx
thusala: Fixed in 4bbcb3cd2 - the aria-label now interpolates `formatDays(...)` for both values so the announced and rendered numbers match.

frontend/src/community/common/assets/languages/english/leaveModule.json
thusala: Fixed in 4bbcb3cd2 - the JSON is back to sentence case and the casing moved to `className: "uppercase"` on the `GridHeader` entries.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/EmployeeLeavePolicyServiceImpl.java
thusala: Fixed in 4bbcb3cd2 - the dead `employeeLeavePolicyListToEmployeeLeavePolicyResponseDtoList` mapper method is removed.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveBalanceServiceImpl.java
thusala: Acknowledged - the module only has controller integration tests today, and adding a balance test layer is a bigger piece of work than this fix. Tracking it as a follow-up rather than bolting a partial suite onto this PR.

frontend/src/community/leave/components/molecules/IndividualEmployeeLeaveReportSection/IndividualEmployeeLeaveReportSection.tsx
thusala: Fixed in 4bbcb3cd2 - the page-level `<h2>` is restored on the legacy `UserAssignedLeaveTypes` branch, which was the path that lost its only heading. On the policies path `UserLeavePolicies` renders its own `<h2>` unconditionally (it sits above the loading gate, not inside it), so that branch was already covered.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/EmployeeLeavePolicyServiceImpl.java
thusala: Matching the module convention here - `PolicyLeaveServiceImpl` already resolves `today` with `DateTimeUtils.getCurrentUtcDate()` for the balance cards, and the org timezone is only used for working-day duration maths. Moving the module to org-local dates is a separate change.

frontend/src/community/leave/components/molecules/UserLeavePolicies/LeavePolicyCard.tsx
thusala: `isDerived` is set to `true` on every branch of the balance calculation, including `emptyBalance()`, so surfacing it as `isBalanceAvailable` would not distinguish the future-effective case from a consumed one. A real "not yet effective" signal needs a new flag - out of scope for this fix.

frontend/src/community/leave/components/molecules/UserLeavePolicies/LeavePolicyCard.tsx
thusala: Fixed in 4bbcb3cd2 - the card now reads `policy.isUnlimited` instead of re-deriving it from `policyType`. `totalDaysUsed` stays on the DTO so it mirrors the balance-card contract.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveBalanceServiceImpl.java
thusala: Fixed in 4bbcb3cd2 - the second walk is skipped when `creditedUpTo` already reaches the cycle end, so the common case is back to two period walks.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveBalanceServiceImpl.java
thusala: The clamp is intentional - a negative remaining balance is not meaningful on a card or in the error copy, and requests that exceed the accrued days are still rejected by `canAccommodate`. Carrying both a raw and a clamped value on the DTO would add a second source of truth for a number no screen shows.

frontend/src/community/leave/components/molecules/AssignLeavePolicyModal/AssignLeavePolicyForm.tsx
thusala: Verified - neither portals: `Tooltip` renders an absolutely positioned child, and `Popper` defaults to `positionStrategy: "absolute"` in place. Fixed in 4bbcb3cd2 by scoping the scroll box to the accrual table only, so the dropdown, radios and datepicker sit outside any clipping ancestor.

backend/src/main/java/com/skapp/community/leaveplanner/service/PolicyLeaveBalanceService.java
thusala: The separate service is gone in 4bbcb3cd2 - `calculateBalanceForYear` now sits on `PolicyLeaveService` with a javadoc stating the caller must pass an assignment attached to an open persistence context.

frontend/src/community/leave/components/molecules/UserLeavePolicies/LeavePolicyCard.tsx
thusala: Out of scope for this fix - renaming the file and folding both cards onto `getPolicyBalanceLabel` is a refactor of the other card too. Noted as a follow-up.

thusala: Fixed in 4bbcb3cd2 - changed to `min-h-[96px]` so `auto-rows-fr` still equalises rows without clipping at large text sizes.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/EmployeeLeavePolicyServiceImpl.java
thusala: Leaving as is - this endpoint also backs the assign modal and the admin assignment flows, so making it throw when policies are disabled would change behaviour well beyond the balance fields. Worth settling module-wide rather than here.

frontend/src/community/leave/components/molecules/AssignLeavePolicyModal/AssignLeavePolicyForm.tsx
thusala: Fixed in 4bbcb3cd2 - the redundant `ariaLabel` is dropped so the visible `label` provides the accessible name.

frontend/src/community/leave/components/molecules/AssignLeavePolicyModal/AssignLeavePolicyForm.tsx
thusala: Fixed in 4bbcb3cd2 - the single-child wrapper and its trailing whitespace are gone, the `Dropdown` renders directly under the outer container.

frontend/src/community/common/assets/languages/english/leaveModule.json
thusala: Both strings match the copy in the design, which uses title case for section headings and sentence case for modal titles. Leaving them as designed rather than picking a new convention in this PR.

frontend/src/community/common/assets/languages/english/leaveModule.json
thusala: Keeping the casing in the copy - the ALL-CAPS values are what the design specifies for these column headers, so the translation stays the single place that owns the string. Reverted in beea22889.

frontend/src/community/leave/components/molecules/AssignLeavePolicyModal/AssignLeavePolicyForm.tsx
thusala: Reverted the `tabIndex` fix in beea22889 - `tabIndex` on a non-interactive `<p>` trips `jsx-a11y/no-noninteractive-tabindex`. Reaching the tooltip properly needs a focusable info affordance next to the title, which is a design call; tracking it as a follow-up.

shakila-kamalasena (dismissed):

hasalarootcode (dismissed):

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/EmployeeLeavePolicyServiceImpl.java
akilarootcode: this need to address

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
```diff
@@ -741,7 +757,7 @@ private PolicyLeaveBalanceDto calculateBalance(EmployeeLeavePolicy assignment, P
 		balance.setAccruedDays(accruedDays);
 		balance.setTotalDaysAllocated(totalDaysAllocated);
 		balance.setTotalDaysUsed(totalDaysUsed);
-		balance.setBalanceInDays(totalDaysAllocated - totalDaysUsed);
+		balance.setBalanceInDays(Math.max(0f, usableDays - totalDaysUsed));
```
akilarootcode: remove max

akilarootcode (changes requested):

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/EmployeeLeavePolicyServiceImpl.java
thusala: Addressed in dcf66e0 (took it on rather than deferring). `sumCommittedDaysForPolicyInCycle` is now a grouped criteria query `sumCommittedDaysForPoliciesInCycle(employeeId, policyIds, ...)` returning a policyId -> days map, and `calculateBalanceForYear` became `calculateBalancesForYear(employeeId, assignments, year)` — both the assignment list and the balance-card endpoint now resolve every assignment in one query. The paged assignment lookup also moved to a criteria query that fetch-joins employee/policy/leaveType, so the per-row lazy loads are gone too. Verified with SQL logging: 6 assignments on a page now fire 1 grouped sum + 1 fetch-join select (was 1 + 6 + 12). Skipped the caller-side opt-out flag since a flat 1 extra query no longer justifies it.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
thusala: Removed

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/EmployeeLeavePolicyServiceImpl.java
thusala: Correction to my previous reply — 933e8a33d scopes this back so Hasala's code is untouched. `sumCommittedDaysForPolicyInCycle` is restored verbatim and still backs the balance-card and apply-leave flows; the batched `sumCommittedDaysForPoliciesInCycle` is a separate new query used only by the list endpoint flagged here, via a new `calculateBalancesForYear(employeeId, assignments, year)` next to the existing `calculateBalanceForYear`. The paged assignment lookup also moved to a criteria query that fetch-joins employee/policy/leaveType. Verified with SQL logging: a 6-row page now fires 1 fetch-join select + 1 grouped sum (was 1 + 6 + per-row lazy loads). `getCurrentUserPolicyBalances` keeps its existing per-assignment path.

thusala: Done. Added a grouped criteria query `findCommittedUsageForPoliciesInWindow(employeeId, policyIds, ...)` returning the committed usage rows keyed by policy, so `calculateBalancesForYear` resolves a whole page from one query instead of one `findCommittedUsageForPolicyInWindow` per assignment.

hasalarootcode (dismissed):

backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/PolicyLeaveRequestRepositoryImpl.java
```diff
@@ -141,6 +145,30 @@ public List<PolicyLeaveUsageDto> findCommittedUsageForPolicyInWindow(Long employ
 		return entityManager.createQuery(criteriaQuery).getResultList();
 	}
 
+	@Override
+	public Map<Long, List<PolicyLeaveUsageDto>> findCommittedUsageForPoliciesInWindow(Long employeeId,
+			Collection<Long> policyIds, Collection<LeaveRequestStatus> statuses, LocalDate windowStart,
```
akilarootcode: remove collection to List

backend/src/main/java/com/skapp/community/leaveplanner/repository/PolicyLeaveRequestRepository.java
```diff
@@ -25,6 +26,10 @@ Page<PolicyLeaveRequest> findSupervisedRequests(Long supervisorEmployeeId, Polic
 	List<PolicyLeaveUsageDto> findCommittedUsageForPolicyInWindow(Long employeeId, Long policyId,
 			Collection<LeaveRequestStatus> statuses, LocalDate windowStart, LocalDate windowEnd);
 
+	Map<Long, List<PolicyLeaveUsageDto>> findCommittedUsageForPoliciesInWindow(Long employeeId,
+			Collection<Long> policyIds, Collection<LeaveRequestStatus> statuses, LocalDate windowStart,
```
akilarootcode: change collection to List

akilarootcode (dismissed):

akilarootcode (approved):

PR #1778, opened 2026-08-17
Fix/2898 employee documents empty state

shakila-kamalasena:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/frontend/src/community/common/assets/languages/english/peopleModule.json
+++ b/frontend/src/community/common/assets/languages/english/peopleModule.json
@@ -212,7 +212,7 @@
     "teams": "Teams :",
     "joinedDate": "Joined Date :",
     "jobFamily": "Job Family :",
-    "systemPermissions": "System Permissions",
+    "systemPermissions": "Permissions",
     "primarySupervisor": "Primary Supervisor :",
     "entryAdded": "Entry added",
     "by": "by {{name}}",
--- a/frontend/src/community/common/assets/languages/english/peopleModule.json
+++ b/frontend/src/community/common/assets/languages/english/peopleModule.json
@@ -556,7 +556,7 @@
       "emergency": "Emergency",
       "employment": "Employment",
       "entitlements": "Entitlements",
-      "systemPermissions": "System Permissions",
+      "systemPermissions": "Permissions",
       "entitlementErrorMessage": "Entitlements could not be added. Please try again later.",
       "resourceSuccessMessage": "Resource added successfully",
       "editToastTitle": "Updated successfully",
@@ -934,7 +934,7 @@
       "head": "Directory | Skapp"
     },
     "systemPermissions": {
-      "title": "System Permissions",
+      "title": "Permissions",
       "superAdmin": "Super Admin",
       "people": "People",
       "leave": "Leave",
--- a/frontend/src/community/people/types/EditEmployeeInfoTypes.ts
+++ b/frontend/src/community/people/types/EditEmployeeInfoTypes.ts
@@ -33,7 +33,7 @@ export enum EditAllInformationType {
   personal = "Personal",
   emergency = "Emergency",
   employment = "Employment",
-  permission = "System Permissions",
+  permission = "Permissions",
   timeline = "Timeline",
   leave = "Leave",
   timesheeet = "Timesheet"
--- a/frontend/src/community/people/types/PeopleEditTypes.ts
+++ b/frontend/src/community/people/types/PeopleEditTypes.ts
@@ -2,7 +2,7 @@ export enum EditPeopleFormTypes {
   personal = "Personal",
   emergency = "Emergency",
   employment = "Employment",
-  permission = "System Permissions",
+  permission = "Permissions",
   timeline = "Timeline",
   leave = "Leave",
   timesheet = "Timesheet",


ThinuwanW (commented):
## 🤖 Claude Code Review

This one-line i18n change renames the `editAllInfo.systemPermissions` label from "System Permissions" to "Permissions", but that string is not purely cosmetic — the People edit steppers use `BoxStepper` with `useStringIdentifier`, so the *translated label* is passed back as the step identifier and compared against `EditPeopleFormTypes.permission` / `EditAllInformationType.permission`, both of which are the literal `"System Permissions"`. As shipped, clicking the Permissions tab in both the new edit flow (`DirectoryEditSectionWrapper` → `PeopleFormSections`) and the legacy `edit-all-information/[id].tsx` page will fall through to the `default` branch and render nothing, the tab will never highlight as active, and `useFormChangeDetector` will stop detecting permission changes. The rename is also only partially applied — the add-people flow stepper and the section title still read "System Permissions".

Found **9** new issue(s): 🔴 3 important, 🟡 3 suggestion(s), 🟣 3 nit(s)

frontend/src/community/common/assets/languages/english/peopleModule.json
```diff
@@ -212,7 +212,7 @@
     "teams": "Teams :",
     "joinedDate": "Joined Date :",
     "jobFamily": "Job Family :",
-    "systemPermissions": "System Permissions",
+    "systemPermissions": "Permissions",
```
ThinuwanW, line 215: 🤖 **Claude** · 🔴 **Important**

This label is used as a step *identifier*, not just display text, so renaming it silently breaks the System Permissions tab.

Trace: `DirectorySteppers.tsx:87` builds `steps` from `translateText(["editAllInfo", "systemPermissions"])` and renders `BoxStepper` with `useStringIdentifier` (line 125). `BoxStepper.tsx:85` then calls `onStepClick(step)` with the *translated string*, which `DirectorySteppers.tsx:124` casts to `EditPeopleFormTypes` and feeds into `setNextStep`. `DirectoryEditSectionWrapper.tsx:67` promotes it to `currentStep`, and `PeopleFormSections.tsx:64-89` switches on `currentStep` against `EditPeopleFormTypes.permission`, which is the literal `"System Permissions"` (`src/community/people/types/PeopleEditTypes.ts:5`).

With the value now `"Permissions"` the match fails, so:
1. `getEditFlowSection()` hits `default: return null` → blank content area when the Permissions tab is clicked.
2. Same on the legacy page: `pages/.../edit-all-information/[id].tsx:185` feeds the label into `formType`, and `getComponent()` (line 511 `case EditAllInformationType.permission`, enum value also `"System Permissions"` in `types/EditEmployeeInfoTypes.ts:36`) falls to a `default` branch that doesn't even `return` → renders nothing.
3. `BoxStepper.tsx:80/94/110` compares `activeStep === step`, so the tab never renders as selected and `aria-selected`/`tabIndex` are wrong for a11y.
4. `useFormChangeDetector.tsx:65` (`case EditPeopleFormTypes.permission`) no longer fires → `hasChanged` stays false on that step, so the unsaved-changes guard is bypassed and the mutation payload comes back empty.

Fix: change both enum values to match the new string exactly — `permission = "Permissions"` in `src/community/people/types/PeopleEditTypes.ts:5` and in `src/community/people/types/EditEmployeeInfoTypes.ts:36`. Better still, decouple the identifier from the label: pass a stable key array to `BoxStepper` alongside the display labels (or index-based `onStepClick`) so future copy changes can't break navigation. Note every other `editAllInfo.*` step label (`Personal`, `Emergency`, `Employment`, `Timeline`, `Leave`, `Timesheet`, `Documents`) currently matches its enum value verbatim — this implicit contract is what the change violates.

ThinuwanW, line 215: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

The rename is only partially applied, so the product now shows two different names for the same concept. `addResource.commonText.systemPermissions` (line 559) still reads "System Permissions" and drives the add-people stepper (`hooks/useStepper.tsx:24` and `AddNewResourceFlow.tsx:71`), and `addResource.systemPermissions.title` (line 937) still renders "System Permissions" as the section heading in `AddNewResourceFlow/SystemPermissionSection/SystemPermissionForm.tsx:261`. If "Permissions" is the intended product wording, update those two values as well; if the rename was meant to be scoped to the edit-flow stepper only, please confirm that intent in the PR description.

ThinuwanW, line 215: 🤖 **Claude** · 🟣 **Nit**

No existing test can catch this class of regression: `DirectorySteppers.test.tsx:37-39` mocks `useTranslator` to return `key.join(".")`, so the assertion on line 53 checks the literal string `"editAllInfo.systemPermissions"` and is blind to the real JSON value. Consider adding a small unit test that asserts the i18n label values equal the corresponding `EditPeopleFormTypes` / `EditAllInformationType` enum values (e.g. `expect(peopleModule.editAllInfo.systemPermissions).toBe(EditPeopleFormTypes.permission)`) so any future copy change fails loudly instead of producing a blank tab.

ThinuwanW, line 215: 🤖 **Claude** · 🔴 **Important**

Silent data loss, separate from the blank-tab render. `useFormChangeDetector` (src/community/people/hooks/useFormChangeDetector.tsx:34-99) switches on `currentStep` against `EditPeopleFormTypes` and has `case EditPeopleFormTypes.permission` (still `"System Permissions"`). Once the stepper emits `"Permissions"`, that switch falls through to `default: break`, so `newApiPayload` stays `{}` and `hasChanged` is always `false` while the user is on the permissions tab. `DirectoryEditSectionWrapper.tsx:63-68` then takes the `else` branch (`setCurrentStep(nextStep)`) instead of opening `UnsavedChangesModal`, so a user who edits roles and clicks another tab loses the edits with no prompt and no error. The same `hasChanged` value feeds the save/submit enablement. Fixing only the stepper label lookup is not enough — the enum value in `PeopleEditTypes.ts:5` must change in lockstep with this JSON value, or the rename must be reverted and applied only to a display-only key.

ThinuwanW, line 215: 🤖 **Claude** · 🔴 **Important**

📁 **File-level observation**

Missing paired change in a second, independent consumer. `EditAllInformationType.permission` in src/community/people/types/EditEmployeeInfoTypes.ts:36 is a *separate* enum that also hardcodes `"System Permissions"`, and pages/community/people/directory/edit-all-information/[id].tsx:185 builds its `BoxStepper` steps from this same `editAllInfo.systemPermissions` key with `useStringIdentifier`. After the rename, `formType` becomes `"Permissions"` and `getComponent()` (same file, lines 443-546) misses `case EditAllInformationType.permission`. Worse, its `default:` branch writes a bare JSX expression statement without `return`, so the function returns `undefined` and the page renders an empty panel with the tab still highlighted. A dev fixing only `EditPeopleFormTypes` will not notice this. Decide explicitly: either update `EditAllInformationType.permission` too and add a `return` to that default branch, or delete this page. Note it appears orphaned — `ROUTES.PEOPLE.EDIT_ALL_INFORMATION` (src/community/common/constants/routes.ts:72) has zero call sites, the live flow is `/people/directory/edit/{id}` → `DirectoryEditSectionWrapper`, and this legacy page renders `SystemPermissionForm` without the `isReadOnly`/`isPeopleAdminViewingOwnProfile` gating that `PeopleFormSections.tsx:82-89` applies, so it is a diverged duplicate that is still reachable by direct URL.

ThinuwanW, line 215: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

A third stale copy of the old label, distinct from the `addResource.commonText` one. `addResource.systemPermissions.title` (line 937) is still `"System Permissions"` and is rendered as the section heading by `SystemPermissionFormSection.tsx:261` (`title={translateText(["title"])}` on `PeopleFormSectionWrapper`) — i.e. the very panel this renamed tab opens. Result: the user clicks a tab labelled "Permissions" and lands on a page headed "System Permissions". If "Permissions" is the intended product wording, update line 937 as well (and line 559 for the add-resource stepper); if not, revert line 215.

ThinuwanW, line 215: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Root cause is a type assertion that defeats strict mode — worth fixing so this class of change stops being silently breaking. `DirectorySteppers.tsx:124` does `onStepClick={(step) => handleStepClick(step as EditPeopleFormTypes)}` and `edit-all-information/[id].tsx:637-639` does the same with `EditAllInformationType`; `BoxStepper`'s `onStepClick` is typed `(step: number | string) => void`, so the `as` cast launders an arbitrary translated string into an enum-typed slot and the compiler cannot catch a label/enum divergence. Per the guidelines ("Using type assertions (`as`) to silence errors instead of fixing the underlying type"), replace the string-label identity with a stable id: have the stepper build `{ id: EditPeopleFormTypes; label: string }[]`, pass ids to `onStepClick`, and let `BoxStepper` render `label` only. Then translation values are free to change without touching behaviour.

ThinuwanW, line 215: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

Defensive gap that made this rename undetectable. `PeopleFormSections.getEditFlowSection()` (src/.../PeopleFormSections/PeopleFormSections.tsx:110-111) returns `null` on an unrecognised `currentStep`, and `BoxStepper` compares `activeStep === step` as plain strings so the tab still renders as selected/highlighted — no console error, no error boundary, nothing a smoke test would trip on. Consider having the default branch log or render an explicit error/empty state in non-production builds so future step-identifier drift surfaces immediately rather than as a blank panel.

ThinuwanW, line 215: 🤖 **Claude** · 🟣 **Nit**

Key/value drift: the key is `systemPermissions` but the value is now `"Permissions"`, while two sibling keys with the identical name (line 559, line 937) still hold `"System Permissions"`. Three same-named keys with two different values in one file is a maintenance trap — the next translator or dev has no way to tell which divergence is intentional. If the split is deliberate, rename this key (e.g. `permissionsTab`) so the name reflects the value and the difference is self-documenting.

shakila-kamalasena: Fixed

shakila-kamalasena: Fixed

shakila-kamalasena: No need

shakila-kamalasena: Already fixed from above comments

shakila-kamalasena: No need as its a legacy page

shakila-kamalasena: Not related to this PR

shakila-kamalasena: Not related to this PR

shakila-kamalasena: Not an issue

shakila-kamalasena: Already fixed

hasalarootcode (dismissed):

hasalarootcode (approved):

akilarootcode (approved):

PR #1780, opened 2026-08-17
[Fix] My Leave Requests inherits page size and date filter from All Leave Requests

akilarootcode:
## Summary

The My Leave Requests table showed a different row count and a different set of requests each time it was revisited, without the user ever leaving the page. It read the whole shared `leaveRequestParams` store object, which the All Leave Requests screen writes `size`, `startDate` and `endDate` into on mount. Returning to My Leave Requests therefore ran the query with a page size of 6 and a current-year date window it never asked for. The query hook now takes its params as an argument, so each caller declares exactly what it needs.

- `useGetEmployeeLeaveRequests` takes a `MyLeaveRequestParamsType` argument instead of reading `leaveRequestParams` out of the store. `MyRequestApi.ts` no longer imports `useLeaveStore` at all.
- `LeaveRequests` (My Leave Requests) passes only `status`, `leaveType`, `page`, `sortKey` and its own `MY_LEAVE_REQUESTS_PER_PAGE`. It never asks for `startDate`/`endDate`, so the date window cannot reach it.
- `AddEditTimeEntry` (timesheet modal) passes its own status filter instead of mutating shared store state on mount to steer the query. Removes a `useEffect`, a `setLeaveRequestParams` call and the `useLeaveStore` import — and one more writer to the shared object.
- `leaveRequestParams.page` typed as a required `number` instead of `number | string`, removing an `as number` cast at both read sites. The only string writer set `page` immediately before `setPagination` overwrote it with a number, so that redundant write is gone too.
- New `MY_LEAVE_REQUESTS_PER_PAGE = 4` constant, alongside the existing `MANAGER_LEAVE_REQUESTS_PER_PAGE`. Value matches what the store initialised to, so a fresh load is unchanged.

## Reproduction

From the reported recording, URL bar reads `/leave/my-requests` in every frame — the user never navigates away:

| Time | Same table |
|------|-----------|
| 0:04 | page 3 of **7**, **4 rows**/page |
| 0:13 | page 7 of 7, last row `28th Oct 2025` |
| 0:25 | page 2 of **2**, still 4 rows |
| 0:34 | page 1 of 2, **6 rows**, all rows 2026 |

Two independent bleeds, landing as separate store updates: the page count collapsing from 7 to 2 (and the 2025 request disappearing) is `startDate`/`endDate`; 4 rows becoming 6 is `size`.

## Architecture / flow

```mermaid
flowchart TB
    subgraph before["Before — one shared object"]
        M1[All Leave Requests<br/>mount effect]
        S1[("leaveRequestParams<br/>status, leaveType, page, sortKey<br/>size, startDate, endDate")]
        L1[My Leave Requests]
        M1 -->|"writes size=6<br/>startDate, endDate"| S1
        S1 -->|"reads ALL fields"| L1
        L1 -.->|"GET /leave?size=6<br/>&startDate&endDate"| X[wrong rows]
    end

    subgraph after["After — caller declares its params"]
        S2[("leaveRequestParams")]
        L2[My Leave Requests]
        T2[Timesheet modal]
        H2["useGetEmployeeLeaveRequests(params)"]
        S2 -->|"reads status, leaveType,<br/>page, sortKey only"| L2
        L2 -->|"+ own page size"| H2
        T2 -->|"own status filter"| H2
        H2 --> Y[GET /leave]
    end
```

- **Backend query fixes.** The `findAllRequestsByEmployee` pagination sort has no unique tiebreaker (tied rows can reshuffle between pages on MySQL) and the date predicate is `start BETWEEN a AND b OR end BETWEEN a AND b`, which drops a leave spanning the whole window. Both are real but neither causes this bug. Held for a separate backend PR.
- **`status`, `leaveType` and `page` are still shared** with All Leave Requests, so applying a status filter there still affects My Leave Requests on return. Not what this ticket or the recording shows; separating the remaining state is a larger refactor.
- **`UserLeaveHistory` writes `size`/`startDate`/`endDate`** into the same shared object, but its own table never reads them (`useGetEmployeeLeaveHistory` takes explicit args). Those writes are dead for that screen and now only affect All Leave Requests.
- **The timesheet modal reuses `MY_LEAVE_REQUESTS_PER_PAGE`** for its date-picker query. It only ever received one page before (size 4 or 6 depending on navigation history), so this is no worse and is now deterministic — but it arguably wants all active leaves unpaged. Left alone to keep this PR to the reported bug.

- **Submodule hash / pointer bumps** are intentionally excluded. `backend/src/main/java/com/skapp/enterprise` and `frontend/src/enterprise` moved when the branch was fast-forwarded onto latest `develop`; those gitlinks are left unstaged and out of both commits. Gitlink updates happen at deployment time, not in feature PRs.

## Testing

| Check | Ran? | Result |
|-------|------|--------|
| **e2e (Playwright)** | **yes** | **pass on this branch, fails on `develop`** — see below |
| tsc (`--noEmit`) | yes | pass — 0 new errors, diffed by `file(line,col): TScode` against a stashed baseline on the same `develop` (656 pre-existing both sides) |
| eslint | yes | pass — 0 errors; 8 pre-existing `exhaustive-deps` warnings on effects this PR does not touch |
| prettier | yes | pass — changed files only |
| `next build` | yes | pass — compiled successfully |
| Unit tests | n/a | no jest specs cover the changed files, and per the repo rules FE component tests are not added |
| Formatter / tsc-lint (BE only) | n/a | frontend-only change |

### e2e regression guard

Added `tests/leave/my-requests-consistency.spec.ts` to the e2e suite (`leave` feature). It seeds eight PENDING requests through the suite's tenant-scoped DB helper — six in the current year, two in the previous one — then walks My Leave Requests → All Leave Requests → My Leave Requests **via the sidebar**, and asserts the row count, the pager and the page-size/date-window parameters are unchanged.

The navigation has to be an in-app route change. An earlier version used `page.goto`, which reloads the app, rebuilds the zustand store from its defaults and made the test pass against the unfixed code — a false green. The previous-year rows matter for the same reason: without them a current-year window changes nothing visible.

Run against both branches, backend and data identical:

| | `develop` (unfixed) | this branch |
|---|---|---|
| rows, 1st visit → 2nd | 5 → **7** | 5 → 5 |
| pager, 1st → 2nd | `1 2 3` → **`1 2`** | `1 2 3` → `1 2 3` |
| `size` | 4 → **6** | 4 → 4 |
| `startDate` / `endDate` | none → **2026-01-01 / 2026-12-31** | none → none |
| result | ✘ fails | ✓ passes |

That reproduces the reported recording: rows growing, the page count collapsing, and the previous-year request dropping out of its own list.

### One residual leak this surfaced

The run also shows `status=PENDING` becoming `status=PENDING,` after the trip to All Requests — a trailing comma from `setLeaveRequestParams`' array serialisation. It is the `status`-still-shared case listed under Exclusions above, does not change the result set, and is out of scope here. The spec logs the full query string but scopes its assertion to the parameters this PR owns, so the leak is recorded rather than hidden.

### DIFF
--- a/frontend/src/community/attendance/components/molecules/AttendanceModals/AddEditTimeEntry/AddEditTimeEntry.tsx
+++ b/frontend/src/community/attendance/components/molecules/AttendanceModals/AddEditTimeEntry/AddEditTimeEntry.tsx
@@ -47,8 +47,8 @@ import {
 } from "~community/common/utils/dateTimeUtils";
 import { useDefaultCapacity } from "~community/configurations/api/timeConfigurationApi";
 import { useGetEmployeeLeaveRequests } from "~community/leave/api/MyRequestApi";
+import { MY_LEAVE_REQUESTS_PER_PAGE } from "~community/leave/constants/stringConstants";
 import { LeaveStatusEnums } from "~community/leave/enums/MyRequestEnums";
-import { useLeaveStore } from "~community/leave/store/store";
 import { useGetAllHolidaysInfinite } from "~community/people/api/HolidayApi";
 
 import styles from "./styles";
@@ -75,16 +75,10 @@ const AddEditTimeEntry = ({ setFromDateTime, setToDateTime }: Props) => {
     currentYear.toString()
   );
 
-  const { data: leaveRequests } = useGetEmployeeLeaveRequests();
-
-  const { setLeaveRequestParams } = useLeaveStore((state) => state);
-
-  useEffect(() => {
-    setLeaveRequestParams("status", [
-      LeaveStatusEnums.APPROVED,
-      LeaveStatusEnums.PENDING
-    ]);
-  }, []);
+  const { data: leaveRequests } = useGetEmployeeLeaveRequests({
+    status: `${LeaveStatusEnums.APPROVED},${LeaveStatusEnums.PENDING}`,
+    size: MY_LEAVE_REQUESTS_PER_PAGE
+  });
 
   const {
     selectedDailyRecord,
--- a/frontend/src/community/leave/api/MyRequestApi.ts
+++ b/frontend/src/community/leave/api/MyRequestApi.ts
@@ -23,12 +23,12 @@ import {
 import { LeaveEntitlementBalanceType } from "~community/leave/types/LeaveEntitlementTypes";
 import {
   LeaveRequestPayloadType,
+  MyLeaveRequestParamsType,
   MyLeaveRequestPayloadType,
   ResourceAvailabilityParamTypes,
   ResourceAvailabilityPayload
 } from "~community/leave/types/MyRequests";
 
-import { useLeaveStore } from "../store/store";
 import { LeaveRequestItemsType } from "../types/LeaveRequestTypes";
 
 export const useGetLeaveAllocation = (selectedYear: string, enabled = true) => {
@@ -150,9 +150,9 @@ export const useApplyLeave = (
   });
 };
 
-export const useGetEmployeeLeaveRequests = () => {
-  const params = useLeaveStore((state) => state.leaveRequestParams);
-
+export const useGetEmployeeLeaveRequests = (
+  params: MyLeaveRequestParamsType
+) => {
   return useQuery({
     queryKey: [leaveQueryKeys.EMPLOYEE_LEAVE_REQUESTS, params],
     queryFn: async () => {
--- a/frontend/src/community/leave/components/molecules/LeaveRequests/LeaveRequests.tsx
+++ b/frontend/src/community/leave/components/molecules/LeaveRequests/LeaveRequests.tsx
@@ -15,6 +15,7 @@ import {
   useGetEmployeeLeaveRequests
 } from "~community/leave/api/MyRequestApi";
 import MyLeaveRequestFilterBody from "~community/leave/components/molecules/MyLeaveRequestFilterBody/MyLeaveRequestFilterBody";
+import { MY_LEAVE_REQUESTS_PER_PAGE } from "~community/leave/constants/stringConstants";
 import { useLeaveStore } from "~community/leave/store/store";
 import { LeaveRequestDataType } from "~community/leave/types/EmployeeLeaveRequestTypes";
 import { generateMyLeaveRequestAriaLabel } from "~community/leave/utils/accessibilityUtils";
@@ -26,9 +27,8 @@ import styles from "./styles";
 const LeaveRequests: FC = () => {
   const classes = styles();
 
-  const currentPage = useLeaveStore((state) => state.leaveRequestParams.page);
-  const leaveRequestSort = useLeaveStore(
-    (state) => state.leaveRequestParams.sortKey
+  const { status, leaveType, page, sortKey } = useLeaveStore(
+    (state) => state.leaveRequestParams
   );
   const {
     setPagination,
@@ -40,7 +40,13 @@ const LeaveRequests: FC = () => {
     leaveRequestsFilter
   } = useLeaveStore((state) => state);
 
-  const { data: leaveRequests, isLoading } = useGetEmployeeLeaveRequests();
+  const { data: leaveRequests, isLoading } = useGetEmployeeLeaveRequests({
+    status,
+    leaveType,
+    page,
+    sortKey,
+    size: MY_LEAVE_REQUESTS_PER_PAGE
+  });
 
   const {
     refetch,
@@ -188,14 +194,14 @@ const LeaveRequests: FC = () => {
       onRowClick={handleRowClick}
       pagination={{
         totalPages: leaveRequests?.totalPages,
-        currentPage: currentPage as number,
+        currentPage: page,
         onPageChange: setPagination
       }}
       toolbar={{
         dropdown: {
           id: "my-leave-requests-sort",
           options: sortOptions,
-          value: leaveRequestSort,
+          value: sortKey,
           onChange: handleSortChange,
           renderSelectedValue: renderSelectedSortValue,
           width: "auto",
--- a/frontend/src/community/leave/components/molecules/ManagerLeaveRequests/ManagerLeaveRequest.tsx
+++ b/frontend/src/community/leave/components/molecules/ManagerLeaveRequests/ManagerLeaveRequest.tsx
@@ -85,9 +85,7 @@ const ManagerLeaveRequest: FC<Props> = ({
     leaveRequestsFilter: state.leaveRequestsFilter
   }));
 
-  const currentPage: number = useLeaveStore(
-    (state) => state.leaveRequestParams.page
-  ) as number;
+  const currentPage = useLeaveStore((state) => state.leaveRequestParams.page);
 
   const leaveRequestSort = leaveRequestParams.sortKey;
 
@@ -238,17 +236,10 @@ const ManagerLeaveRequest: FC<Props> = ({
   useEffect(() => {
     if (employeeLeaveRequests?.length === 0 && totalPages === 0) {
       if (currentPage !== 0) {
-        setLeaveRequestParams("page", (currentPage - 1).toString());
         setPagination(currentPage - 1);
       }
     }
-  }, [
-    currentPage,
-    employeeLeaveRequests?.length,
-    setLeaveRequestParams,
-    setPagination,
-    totalPages
-  ]);
+  }, [currentPage, employeeLeaveRequests?.length, setPagination, totalPages]);
 
   useEffect(() => {
     if (getLeaveByIdSuccess && getLeaveByIdData) {
--- a/frontend/src/community/leave/constants/stringConstants.ts
+++ b/frontend/src/community/leave/constants/stringConstants.ts
@@ -8,6 +8,8 @@ export const POLICY_LEAVE_REQUESTS_PER_PAGE = 10;
 
 export const MANAGER_LEAVE_REQUESTS_PER_PAGE = 6;
 
+export const MY_LEAVE_REQUESTS_PER_PAGE = 4;
+
 export const LEAVE_REQUESTS_SKELETON_ROWS = 5;
 
 export const MAX_ALLOWED_UPLOADS = 5;
--- a/frontend/src/community/leave/types/MyRequests.ts
+++ b/frontend/src/community/leave/types/MyRequests.ts
@@ -64,6 +64,14 @@ export interface ResourceAvailabilityPayload {
   holidays: HolidayType[];
 }
 
+export interface MyLeaveRequestParamsType {
+  status?: string;
+  leaveType?: string;
+  page?: number;
+  sortKey?: string;
+  size: number;
+}
+
 export interface MyLeaveRequestPayloadType {
   leaveRequestId: number;
   startDate: string;
--- a/frontend/src/community/leave/types/StoreTypes.ts
+++ b/frontend/src/community/leave/types/StoreTypes.ts
@@ -205,7 +205,7 @@ export interface LeaveStore extends actionsTypes {
     startDate?: string;
     endDate?: string;
     sortKey?: string;
-    page?: number | string;
+    page: number;
     size?: string;
     managerType?: string[];
   };


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR correctly untangles `useGetEmployeeLeaveRequests` from the shared Zustand `leaveRequestParams` slice by making params caller-supplied, which fixes a real cross-component bug (the Add/Edit Time Entry modal was mutating the leave store's `status`/`page` and corrupting the My Leave Requests and manager leave queries). The two call sites now get distinct React Query cache entries, the `page: number | string` union is properly narrowed, and the redundant `setLeaveRequestParams("page", ...)` in ManagerLeaveRequest is removed without behavior change. However, the decoupling stops half-way: `LeaveRequests` still reads `status`/`leaveType`/`page`/`sortKey` from the *shared* manager slice so cross-page bleed remains, and the attendance date-picker now explicitly requests only 4 leave requests via a table page-size constant, so leave markers on the calendar are truncated.

Found **22** new issue(s): 🔴 4 important, 🟡 13 suggestion(s), 🟣 5 nit(s)

frontend/src/community/attendance/components/molecules/AttendanceModals/AddEditTimeEntry/AddEditTimeEntry.tsx
```diff
@@ -75,16 +75,10 @@ const AddEditTimeEntry = ({ setFromDateTime, setToDateTime }: Props) => {
...
-
-  useEffect(() => {
-    setLeaveRequestParams("status", [
-      LeaveStatusEnums.APPROVED,
-      LeaveStatusEnums.PENDING
-    ]);
-  }, []);
+  const { data: leaveRequests } = useGetEmployeeLeaveRequests({
+    status: `${LeaveStatusEnums.APPROVED},${LeaveStatusEnums.PENDING}`,
+    size: MY_LEAVE_REQUESTS_PER_PAGE
```
ThinuwanW, line 80: 🤖 **Claude** · 🔴 **Important**

`leaveRequests?.items` is fed to `InputDate` as `myLeaveRequests`, which InputDate uses to mark leave days across the whole calendar (see InputDate.tsx:144-151, `isLeave`). Capping the request at `MY_LEAVE_REQUESTS_PER_PAGE` (4) means at most 4 leave days can ever be highlighted, and since no `sortKey`/`sortOrder` is passed the backend default (`LeaveRequestFilterDto`: `sortKey = START_DATE`, `sortOrder = DESC`) returns the 4 *latest-starting* leaves — so near-term/past leave days silently go unmarked. Reusing the My-Leave-Requests **table** page size for a calendar-marker query is also semantically wrong: the two have no reason to move together. Fix: request the full set the calendar can display, e.g. pass `startDate`/`endDate` bounded to the picker's year (`getMinDateOfYear()`..today, which the component already computes) with a large `size`, or add a dedicated constant such as `LEAVE_MARKERS_FETCH_SIZE`. If the intent really was to preserve the old (equally-truncated) behaviour, please add a comment saying so.

frontend/src/community/leave/components/molecules/LeaveRequests/LeaveRequests.tsx
```diff
@@ -26,9 +27,8 @@ import styles from "./styles";
 const LeaveRequests: FC = () => {
   const classes = styles();
 
-  const currentPage = useLeaveStore((state) => state.leaveRequestParams.page);
-  const leaveRequestSort = useLeaveStore(
-    (state) => state.leaveRequestParams.sortKey
+  const { status, leaveType, page, sortKey } = useLeaveStore(
```
ThinuwanW, line 30: 🤖 **Claude** · 🔴 **Important**

The PR's goal is to stop unrelated components from driving this query, but `LeaveRequests` still sources `status`, `leaveType`, `page` and `sortKey` from `leaveRequestParams` — the same slice that `ManagerLeaveRequest.tsx` (lines 223-233), `UserLeaveHistory.tsx` (line 471-483), `useManagerLeaveRequestFilters` and `useGetManagerAssignedLeaveRequests` (LeaveApi.ts:311) all read/write. Concrete failure: a manager paginates All Leave Requests to page 4 (size 6), then navigates to My Requests; nothing resets `leaveRequestParams.page`, so this component fetches page 4 at size 4 and renders an empty table — and unlike `ManagerLeaveRequest` (lines 236-242) there is no "empty page → step back" effect here to recover. Same bleed applies to `status`/`leaveType` set by the manager filter. Recommend finishing the decoupling: give My Leave Requests its own slice (e.g. `myLeaveRequestParams` + `setMyLeaveRequestsPagination`) or local component state, so the manager and employee tables no longer share pagination and filter state.

frontend/src/community/leave/api/MyRequestApi.ts
```diff
@@ -150,9 +150,9 @@ export const useApplyLeave = (
   });
 };
 
-export const useGetEmployeeLeaveRequests = () => {
-  const params = useLeaveStore((state) => state.leaveRequestParams);
-
+export const useGetEmployeeLeaveRequests = (
```
ThinuwanW, line 153: 🤖 **Claude** · 🟡 **Suggestion**

The params are now typed but the return value still isn't: there is no `UseQueryResult<...>` annotation and `select` returns `response?.data?.results[0]`, which is `any`. Consumers therefore access `leaveRequests.items` and `leaveRequests.totalPages` with zero type checking (contrast `useGetResourceAvailability`/`useGetMyRequests` in the same file, which do annotate). Since this PR is already tightening the contract, add a paginated response type (e.g. `{ items: LeaveRequestDataType[]; totalPages: number; totalItems: number; currentPage: number }`) and annotate the hook. Also note `results[0]` is not null-safe — `response?.data?.results` guards `data` but not `results`; use `response?.data?.results?.[0]`.

frontend/src/community/leave/constants/stringConstants.ts
```diff
@@ -8,6 +8,8 @@ export const POLICY_LEAVE_REQUESTS_PER_PAGE = 10;
 
 export const MANAGER_LEAVE_REQUESTS_PER_PAGE = 6;
 
+export const MY_LEAVE_REQUESTS_PER_PAGE = 4;
```
ThinuwanW, line 11: 🤖 **Claude** · 🟡 **Suggestion**

`MY_LEAVE_REQUESTS_PER_PAGE = 4` is introduced but the value is still duplicated as a magic string in `leaveRequestFilterSlice.tsx` (`size: "4"` in both the initial state at line 21 and `resetLeaveRequestParams` at line 81), and `ManagerLeaveRequest.tsx:223` still hardcodes `setLeaveRequestParams("size", "6")` even though `MANAGER_LEAVE_REQUESTS_PER_PAGE = 6` already exists. That leaves three sources of truth for two page sizes, with a string/number type split on top (`StoreTypes.leaveRequestParams.size?: string` vs `MyLeaveRequestParamsType.size: number`). Replace the literals with `String(MY_LEAVE_REQUESTS_PER_PAGE)` / `String(MANAGER_LEAVE_REQUESTS_PER_PAGE)` (or better, change the store's `size` to `number` now that `page` was normalized the same way).

frontend/src/community/leave/types/StoreTypes.ts
```diff
@@ -205,7 +205,7 @@ export interface LeaveStore extends actionsTypes {
     startDate?: string;
     endDate?: string;
     sortKey?: string;
-    page?: number | string;
+    page: number;
```
ThinuwanW, line 208: 🤖 **Claude** · 🟡 **Suggestion**

Narrowing `page` to `number` is the right call, but the setter that writes it is still untyped: `setLeaveRequestParams: (key: string, value: string | string[])` (line 98) spreads `[key]: types` into `leaveRequestParams`, so `setLeaveRequestParams("page", "2")` compiles fine and puts a string back into a field the interface now declares as `number` — exactly the call this PR just removed from ManagerLeaveRequest. Consider making the setter key-safe, e.g. `<K extends Exclude<keyof LeaveStore["leaveRequestParams"], "page">>(key: K, value: ...)`, so `setPagination` remains the only way to write `page` and the invariant is enforced by the compiler rather than by convention.

frontend/src/community/leave/types/MyRequests.ts
```diff
@@ -64,6 +64,14 @@ export interface ResourceAvailabilityPayload {
   holidays: HolidayType[];
 }
 
+export interface MyLeaveRequestParamsType {
```
ThinuwanW, line 67: 🤖 **Claude** · 🟣 **Nit**

Two things on the new `MyLeaveRequestParamsType`: (1) the optionality is inverted relative to the store — `size` is required while `page` is optional, even though `StoreTypes` now guarantees `page: number`; making `page` required (and `size` optional with a default in the hook) would better match how both call sites actually behave. (2) `status?: string` / `leaveType?: string` are looser than the backend contract — `LeaveRequestFilterDto` binds `status` to `List<LeaveRequestStatus>` and `leaveType` to `List<Long>`, so typing these as `LeaveStatusEnums[] | string` / `string[]` would catch malformed values at compile time. Worth noting the type also omits `startDate`/`endDate`/`sortOrder`, which the old store-sourced object used to forward; that drop looks intentional for My Requests (its filter body has no date filter) but should be confirmed.

frontend/src/community/leave/components/molecules/LeaveRequests/LeaveRequests.tsx
```diff
@@ -40,7 +40,13 @@ const LeaveRequests: FC = () => {
     leaveRequestsFilter
   } = useLeaveStore((state) => state);
 
-  const { data: leaveRequests, isLoading } = useGetEmployeeLeaveRequests();
+  const { data: leaveRequests, isLoading } = useGetEmployeeLeaveRequests({
+    status,
+    leaveType,
+    page,
+    sortKey,
+    size: MY_LEAVE_REQUESTS_PER_PAGE
```
ThinuwanW, line 48: 🤖 **Claude** · 🟣 **Nit**

`MyLeaveRequestsSection` renders either this component or `PolicyLeaveRequests` for the same "My Leave Requests" section, but the policy variant pages at `POLICY_LEAVE_REQUESTS_PER_PAGE = 10` while this one now pages at 4 — the same UI shows very different page sizes depending on whether leave policies are enabled. Also `skeletonRows={5}` (line 189) is now larger than the page size, so the loading skeleton is taller than the loaded table. Worth aligning these three numbers, or at least confirming 4 is a deliberate product decision.

frontend/src/community/leave/components/molecules/LeaveRequests/LeaveRequests.tsx
```diff
@@ -40,7 +40,13 @@ const LeaveRequests: FC = () => {
     leaveRequestsFilter
   } = useLeaveStore((state) => state);
```
ThinuwanW, line 41: 🤖 **Claude** · 🟣 **Nit**

Lines 30-31 were just refactored to a granular selector, but this second `useLeaveStore((state) => state)` right below still subscribes to the entire leave store, so this component re-renders on any unrelated leave-store change (leave types, entitlements, modals, analytics — all slices live in the same store). Since you're already touching this block, pull the same fields via a narrow selector (or `useShallow`) as `ManagerLeaveRequest.tsx:74-86` does.

frontend/src/community/leave/api/MyRequestApi.ts
```diff
@@ -23,12 +23,12 @@ import {
 import { LeaveEntitlementBalanceType } from "~community/leave/types/LeaveEntitlementTypes";
 import {
   LeaveRequestPayloadType,
+  MyLeaveRequestParamsType,
```
ThinuwanW, line 26: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

No test coverage accompanies the signature change. `useGetEmployeeLeaveRequests` went from store-sourced to caller-supplied params and there is no `.test.tsx`/`.test.ts` anywhere for the hook, `LeaveRequests`, or `AddEditTimeEntry`, so the regression this PR fixes (one component's store write changing another component's query) has nothing guarding it. A small test asserting that the two call sites produce distinct query keys / request params — and that `AddEditTimeEntry` no longer mutates `leaveRequestParams` — would lock in the fix.

frontend/src/community/leave/components/molecules/LeaveRequests/LeaveRequests.tsx
```diff
@@ -188,14 +194,14 @@ const LeaveRequests: FC = () => {
       onRowClick={handleRowClick}
       pagination={{
         totalPages: leaveRequests?.totalPages,
-        currentPage: currentPage as number,
+        currentPage: page,
```
ThinuwanW, line 197: 🤖 **Claude** · 🔴 **Important**

`page` is now an explicit query param, but it is still sourced from the shared `leaveRequestParams` slice that the manager table also writes via `setPagination` — and this component has no page clamp. Repro: on All Leave Requests (LegacyAllLeaveRequests → ManagerLeaveRequest, 6 rows/page) go to page 2, then client-side navigate to /community/leave/my-requests. LeaveRequests mounts with page=1 and size=4; a user with ≤4 requests gets totalPages=1, the backend returns an empty items array, and the table renders the "no leave requests" empty state forever. ManagerLeaveRequest has a dedicated clamp effect for exactly this case (it decrements the page when items are empty and totalPages is 0); LeaveRequests has none, and nothing on the my-requests page resets the params on mount. Fix: either hold `page` in local component state (consistent with the direction of this PR), or reset it to 0 on mount, or add the same empty-page clamp effect.

frontend/src/community/leave/api/MyRequestApi.ts
ThinuwanW, line 26: 🤖 **Claude** · 🔴 **Important**

📁 **File-level observation**

The store-driven-params anti-pattern this PR removes was only deleted from one of three sites, so the defect class is still live. (1) `useGetManagerAssignedLeaveRequests` in LeaveApi.ts:311 still does `useLeaveStore((state) => state.leaveRequestParams)` — the exact pattern deleted here. (2) LegacyAllLeaveRequests.tsx:19 still runs a mount effect writing `status` into the shared bag to drive that query — byte-for-byte the pattern removed from AddEditTimeEntry. (3) ManagerLeaveRequest.tsx:223/232/233 writes `size`, `startDate` and `endDate` into the same bag. Because LeaveRequests.tsx still reads `status`, `leaveType` and `sortKey` from that bag, filtering or sorting All Leave Requests silently re-filters/re-sorts My Leave Requests on the next navigation. Fix in the same pass: give the manager query a caller-supplied params type (mirroring `MyLeaveRequestParamsType`) plus a dedicated slice or local state, and delete the LegacyAllLeaveRequests mount effect — otherwise the PR's stated goal isn't reached and a future reader will assume the coupling is gone.

frontend/src/community/attendance/components/molecules/AttendanceModals/AddEditTimeEntry/AddEditTimeEntry.tsx
```diff
@@ -75,16 +75,10 @@ const AddEditTimeEntry = ({ setFromDateTime, setToDateTime }: Props) => {
...
-  const { setLeaveRequestParams } = useLeaveStore((state) => state);
-
-  useEffect(() => {
-    setLeaveRequestParams("status", [
-      LeaveStatusEnums.APPROVED,
-      LeaveStatusEnums.PENDING
-    ]);
-  }, []);
+  const { data: leaveRequests } = useGetEmployeeLeaveRequests({
+    status: `${LeaveStatusEnums.APPROVED},${LeaveStatusEnums.PENDING}`,
```
ThinuwanW, line 79: 🤖 **Claude** · 🟡 **Suggestion**

There are now three different ways to build the comma-separated `status` param for the same endpoint: the new hand-rolled template literal here, the loop inside `setLeaveRequestParams` (leaveRequestFilterSlice.tsx:45-55), and `getPolicyLeaveRequestQueryParams` in policyLeaveUtils.ts:286 which already does it correctly with `join(",")`. The store setter's loop appends a trailing comma, so applying a My Leave Requests status filter (useMyLeaveRequestFilters.ts:39-40) still sends a value ending in a comma to the same LeaveController GET endpoint; Spring's comma-delimited binding turns the empty trailing segment into a `null` element in `List<LeaveRequestStatus>`/`List<Long>`, which then lands in the criteria IN-list (LeaveRequestRepositoryImpl.buildEmployeeLeavePredicates). It does not currently throw, but it is a latent trap. Fix: replace the loop body in `setLeaveRequestParams` with `value.join(",")` (returning `undefined` for an empty array, as the policy util does) and reuse that single helper here instead of a template literal.

frontend/src/community/leave/api/MyRequestApi.ts
ThinuwanW, line 153: 🤖 **Claude** · 🟡 **Suggestion**

Now that the params are caller-supplied, this is the moment to move the query key onto the module's factory convention. `leaveQueryKeys.EMPLOYEE_LEAVE_REQUESTS` is a bare array, so the key evaluates to a nested array wrapping the params object, while every comparable hook in this module uses a factory: `myRequestsQueryKeys.RESOURCE_AVAILABILITY(...)` in this same file, and `policyLeaveQueryKeys.MY_POLICY_LEAVE_REQUESTS_PAGE(queryParams)` for the policy equivalent of this exact query. Add e.g. `EMPLOYEE_LEAVE_REQUESTS_PAGE: (params: MyLeaveRequestParamsType) => [...leaveQueryKeys.EMPLOYEE_LEAVE_REQUESTS, "page", params]` in QueryKeys.ts and use it here; the existing prefix-based invalidations in `useApplyLeave` and `useCancelLeaveRequest` keep working.

frontend/src/community/attendance/components/molecules/AttendanceModals/AddEditTimeEntry/AddEditTimeEntry.tsx
```diff
@@ -75,16 +75,10 @@ const AddEditTimeEntry = ({ setFromDateTime, setToDateTime }: Props) => {
...
-
-  const { setLeaveRequestParams } = useLeaveStore((state) => state);
-
-  useEffect(() => {
-    setLeaveRequestParams("status", [
-      LeaveStatusEnums.APPROVED,
-      LeaveStatusEnums.PENDING
-    ]);
-  }, []);
+  const { data: leaveRequests } = useGetEmployeeLeaveRequests({
```
ThinuwanW, line 78: 🤖 **Claude** · 🟡 **Suggestion**

Missing `enabled` flag on a query that depends on runtime state. This component is rendered for five `EmployeeTimesheetModalTypes` values, but `leaveRequests` is consumed only by the `InputDate` block, which renders solely for ADD_TIME_ENTRY and EDIT_AVAILABLE_TIME_ENTRY — and in the latter the picker is `readOnly`, so the leave markers are cosmetic. The fetch therefore fires needlessly for ADD_LEAVE_TIME_ENTRY, EDIT_LEAVE_TIME_ENTRY and ADD_TIME_ENTRY_BY_TABLE. Add `enabled: employeeTimesheetModalType === EmployeeTimesheetModalTypes.ADD_TIME_ENTRY` (extend the hook's signature with an `enabled` arg like `useGetLeaveAllocation` already does), and move the `useAttendanceStore` destructure above this call so `employeeTimesheetModalType` is in scope.

frontend/src/community/attendance/components/molecules/AttendanceModals/AddEditTimeEntry/AddEditTimeEntry.tsx
ThinuwanW, line 80: 🤖 **Claude** · 🟡 **Suggestion**

Semantic misuse of a cross-module constant: `MY_LEAVE_REQUESTS_PER_PAGE` is defined as the My Leave Requests *table* page size, and it is now the fetch limit for the attendance date picker's leave markers. These are unrelated concerns — raising the leave table to 10 rows per page would silently change how many days the timesheet calendar marks as leave, and dropping it to 2 would silently stop marking days. Declare a purpose-named constant in the attendance constants (e.g. `TIME_ENTRY_LEAVE_LOOKUP_SIZE`) and use it here, so the two values can diverge intentionally.

frontend/src/community/leave/components/molecules/LeaveRequests/LeaveRequests.tsx
```diff
@@ -40,7 +40,13 @@ const LeaveRequests: FC = () => {
     leaveRequestsFilter
   } = useLeaveStore((state) => state);
 
-  const { data: leaveRequests, isLoading } = useGetEmployeeLeaveRequests();
+  const { data: leaveRequests, isLoading } = useGetEmployeeLeaveRequests({
```
ThinuwanW, line 43: 🤖 **Claude** · 🟡 **Suggestion**

Only `isLoading` is destructured; `isError` is discarded. When the request fails, `leaveRequests` is undefined, the rows fall back to an empty array, and the user sees the "no leave requests" empty state — a fetch failure is indistinguishable from genuinely having no leave. Note the hook's `select` also indexes `results[0]` without guarding `results`, so a malformed payload surfaces as a query error rather than data. `PolicyLeaveRequests.tsx:142-156` — the other branch of the very same MyLeaveRequestsSection — already destructures `isError`/`isFetching`/`refetch` and renders `PolicyLeaveErrorState` with a retry button. Mirror that here so both variants of the section behave the same on failure.

frontend/src/community/leave/types/MyRequests.ts
```diff
@@ -64,6 +64,14 @@ export interface ResourceAvailabilityPayload {
...
+  status?: string;
+  leaveType?: string;
+  page?: number;
+  sortKey?: string;
```
ThinuwanW, line 71: 🤖 **Claude** · 🟡 **Suggestion**

`sortKey?: string` and `status?: string` are loose strings in a brand-new type, whereas the sibling `PolicyLeaveRequestQueryParams` (PolicyLeaveTypes.ts:143) already types the same fields as `sortKey: SortKeyTypes` / `sortOrder: SortOrderTypes`. The backend binds this param to the `LeaveRequestSort` enum (CREATED_DATE / START_DATE only), so a typo or an unrelated `SortKeyTypes` member such as NAME or TEAM_NAME compiles fine here and fails at runtime instead. Type `sortKey?: SortKeyTypes` and, if you keep the comma-joined form for `status`, add a comment or a branded alias documenting that it is a joined `LeaveStatusEnums` list.

frontend/src/community/leave/types/StoreTypes.ts
```diff
@@ -205,7 +205,7 @@ export interface LeaveStore extends actionsTypes {
...
     size?: string;
```
ThinuwanW, line 209: 🤖 **Claude** · 🟡 **Suggestion**

Missed paired change in the same block: `page` was narrowed to `number` but `size?: string` was left as a string, even though the new `MyLeaveRequestParamsType.size` is `number` and both page-size constants (`MY_LEAVE_REQUESTS_PER_PAGE`, `MANAGER_LEAVE_REQUESTS_PER_PAGE`) are numbers. The slice now holds a params bag with mixed representations of two adjacent pagination fields, which is exactly the ambiguity this PR set out to remove. Narrow `size` to `number` too and update the two writers that pass a string literal (ManagerLeaveRequest.tsx:223 and UserLeaveHistory.tsx:471) plus the slice default and reset.

frontend/src/community/leave/components/molecules/ManagerLeaveRequests/ManagerLeaveRequest.tsx
```diff
@@ -85,9 +85,7 @@ const ManagerLeaveRequest: FC<Props> = ({
     leaveRequestsFilter: state.leaveRequestsFilter
   }));
 
-  const currentPage: number = useLeaveStore(
-    (state) => state.leaveRequestParams.page
-  ) as number;
+  const currentPage = useLeaveStore((state) => state.leaveRequestParams.page);
```
ThinuwanW, line 88: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Missing paired change for the manager table. `MANAGER_LEAVE_REQUESTS_PER_PAGE = 6` already exists in the same constants file this PR edits (currently used only by policyLeaveReviewSlice.ts:16), yet the legacy manager path still hardcodes the page size as a string literal in the mount effect here (line 223) and again in UserLeaveHistory.tsx:471. Since the PR added and wired up the `MY_...` analogue, wire up the manager one in the same pass — otherwise the constant looks authoritative while two components silently define the real value.

ThinuwanW, line 88: 🤖 **Claude** · 🟡 **Suggestion**

With the `as number` cast gone, this line is a redundant third subscription to the same store: `leaveRequestParams` is already selected in the object selector just above (line 84), so `currentPage` is just `leaveRequestParams.page`. Delete this line and read it from `leaveRequestParams` — the clamp effect and the pagination prop both work unchanged. While in here: that object selector also selects `resetLeaveRequestParams` and `leaveRequestFilterOrder`, neither of which is destructured or used anywhere in the component — dead entries worth removing.

frontend/src/community/leave/types/MyRequests.ts
ThinuwanW, line 67: 🤖 **Claude** · 🟡 **Suggestion**

The new params type deliberately omits `startDate`/`endDate`, which the store bag used to supply. Two consequences to confirm/document. (a) Behaviour change: after visiting All Leave Requests or the employee leave-history page — both of which write current-year bounds into the shared bag — My Leave Requests used to be silently scoped to the current year and now returns all-time. The backend only applies its date predicate when both bounds are non-null (LeaveRequestRepositoryImpl.buildEmployeeLeavePredicates), so this is safe, but it changes what page 1 shows for long-tenured employees. (b) The my-requests page has a year dropdown (`selectedYear`), and the policy variant threads that year into its query params; the community table now has no field through which it could ever be scoped. If the table is meant to follow the year selector, add the year/date fields here; if it is intentionally all-time, say so in the PR description so the two variants' divergence is a recorded decision.

frontend/src/community/leave/components/molecules/LeaveRequests/LeaveRequests.tsx
```diff
@@ -15,6 +15,7 @@ import {
   useGetEmployeeLeaveRequests
 } from "~community/leave/api/MyRequestApi";
 import MyLeaveRequestFilterBody from "~community/leave/components/molecules/MyLeaveRequestFilterBody/MyLeaveRequestFilterBody";
+import { MY_LEAVE_REQUESTS_PER_PAGE } from "~community/leave/constants/stringConstants";
```
ThinuwanW, line 18: 🤖 **Claude** · 🟣 **Nit**

While adding this import from stringConstants, also replace the hardcoded skeleton row count at line 189 with `LEAVE_REQUESTS_SKELETON_ROWS` from the same file — it holds the identical value and is already used by MyLeaveRequestsSection.tsx:17, AllLeaveRequestsSection.tsx:15 and PolicyLeaveRequests.tsx:171, making this the only remaining magic number for that value.

frontend/src/community/attendance/components/molecules/AttendanceModals/AddEditTimeEntry/AddEditTimeEntry.tsx
akilarootcode: Not in this PR. The 4-item cap is pre-existing - the shared store already supplied size 4 or 6 depending on navigation. The ordering shift is real though: sortKey is no longer sent, so the backend default START_DATE DESC applies. Both tracked with the dedicated-constant change.

frontend/src/community/leave/components/molecules/LeaveRequests/LeaveRequests.tsx
akilarootcode: Not in this PR - status, leaveType and page stay shared by design, listed under Exclusions. Separating them is the follow-up this PR was deliberately scoped out of.

frontend/src/community/leave/api/MyRequestApi.ts
akilarootcode: Deferred - the response type and the results?.[0] guard land with the typing pass.

frontend/src/community/leave/constants/stringConstants.ts
akilarootcode: Deferred - wiring MANAGER_LEAVE_REQUESTS_PER_PAGE and collapsing the string/number split touches the manager path this PR deliberately leaves alone.

frontend/src/community/leave/types/StoreTypes.ts
akilarootcode: Correct, the setter can still write a string into page. Deferred - a key-safe setter lands with the slice separation.

frontend/src/community/leave/types/MyRequests.ts
akilarootcode: Deferred - page/size optionality and enum-typed status/leaveType go in with the typing pass.

frontend/src/community/leave/components/molecules/LeaveRequests/LeaveRequests.tsx
akilarootcode: Page size 4 is deliberate: it is the value the store already defaulted to, so a fresh load is unchanged. Skeleton row alignment deferred.

frontend/src/community/leave/components/molecules/LeaveRequests/LeaveRequests.tsx
akilarootcode: Deferred - narrowing the whole-store selector is worth doing but is unrelated to the params fix.

frontend/src/community/leave/api/MyRequestApi.ts
akilarootcode: Not doing this - CLAUDE.md forbids FE component tests. The regression is covered by an e2e guard instead, e2e tests/leave/my-requests-consistency.spec.ts, which fails on develop and passes on this branch.

frontend/src/community/leave/components/molecules/LeaveRequests/LeaveRequests.tsx
akilarootcode: Confirmed - there is no clamp here and ManagerLeaveRequest has one. Not fixed in this PR; it is the next change on this file.

frontend/src/community/leave/api/MyRequestApi.ts
akilarootcode: Confirmed - LegacyAllLeaveRequests.tsx:19 and LeaveApi.ts:311 both still use the store-driven pattern. Out of scope here; that is the follow-up.

frontend/src/community/attendance/components/molecules/AttendanceModals/AddEditTimeEntry/AddEditTimeEntry.tsx
akilarootcode: Confirmed - an e2e run on this branch logs status=PENDING with the trailing comma. Deferred; the fix belongs in the slice setter, not at this call site.

frontend/src/community/leave/api/MyRequestApi.ts
akilarootcode: Deferred - the key already lives in QueryKeys.ts per CLAUDE.md; moving it to a factory is a follow-up.

frontend/src/community/attendance/components/molecules/AttendanceModals/AddEditTimeEntry/AddEditTimeEntry.tsx
akilarootcode: Deferred - the query already fired unconditionally before this PR, so gating it is a separate change.

frontend/src/community/attendance/components/molecules/AttendanceModals/AddEditTimeEntry/AddEditTimeEntry.tsx
akilarootcode: Deferred - already called out in the PR description as a known follow-up; the purpose-named constant lands with it.

frontend/src/community/leave/components/molecules/LeaveRequests/LeaveRequests.tsx
akilarootcode: Correct, and CLAUDE.md already requires distinguishing error state from empty state. Deferred rather than declined - it needs the PolicyLeaveErrorState treatment to match the sibling branch.

frontend/src/community/leave/types/MyRequests.ts
akilarootcode: Deferred - sortKey should be SortKeyTypes per CLAUDE.md's enum rule; lands with the typing pass.

frontend/src/community/leave/types/StoreTypes.ts
akilarootcode: Deferred - narrowing size to number means touching ManagerLeaveRequest and UserLeaveHistory, both out of scope here.

frontend/src/community/leave/components/molecules/ManagerLeaveRequests/ManagerLeaveRequest.tsx
akilarootcode: Deferred - wiring the manager constant changes the manager path this PR deliberately leaves alone.

akilarootcode: Correct on both counts. Deferred - the redundant selector and the dead resetLeaveRequestParams/leaveRequestFilterOrder entries go with the manager cleanup.

frontend/src/community/leave/types/MyRequests.ts
akilarootcode: Intentional - dropping startDate/endDate is the fix itself, and it is documented in the PR description. The backend only applies its date predicate when both bounds are non-null, so all-time is the correct result.

frontend/src/community/leave/components/molecules/LeaveRequests/LeaveRequests.tsx
akilarootcode: Deferred - the LEAVE_REQUESTS_SKELETON_ROWS swap goes with the skeleton/page-size alignment above.

hasalarootcode (approved):

PR #1781, opened 2026-08-18
[Fix] Stabilise My Leave Requests pagination and date range filtering

akilarootcode:
## Summary

Two correctness problems in the query behind **My Leave Requests** (`GET /leave` → `LeaveRequestRepositoryImpl.findAllRequestsByEmployee`): pagination has no deterministic ordering, and the date filter silently drops leaves that span the whole requested range. Both are longstanding and independent of any UI change.

- **Deterministic paging.** The query ordered only by the requested sort column — `startDate` or `createdDate` (`LeaveRequestSort`). Neither is unique, so rows that tie have no defined order and MySQL is free to return a different slice for the same `LIMIT/OFFSET` across calls. `leaveRequestId` is now appended as a final ordering, extracted into `buildStableOrders`.
- **Correct range overlap.** The filter matched `startDate BETWEEN range OR endDate BETWEEN range`, which misses a leave that starts *before* the range and ends *after* it — a long leave spanning the window disappears from its own date filter. Replaced with the standard overlap test already used by the sibling queries in this class (`setDateRangeFiltration`): `startDate <= rangeEnd AND endDate >= rangeStart`.

## Impact

```mermaid
flowchart TB
    subgraph sort["Ordering — ties had no tiebreaker"]
        A["ORDER BY startDate DESC<br/>LIMIT 4 OFFSET 4"] --> B{"rows tie on startDate"}
        B -->|"call 1"| C["rows A, B, C, D"]
        B -->|"call 2"| D["rows A, C, B, E"]
        C -.->|"same query,<br/>different page"| D
        E["ORDER BY startDate DESC,<br/>leaveRequestId DESC"] --> F["stable slice"]
    end

    subgraph range["Range — spanning leave was dropped"]
        G["filter: Mar 1 – Mar 31"]
        H["leave: Feb 20 – Apr 10"]
        G --> I{"start BETWEEN or<br/>end BETWEEN?"}
        I -->|"neither is inside"| J["excluded"]
        G --> K{"start <= Mar 31<br/>AND end >= Mar 1?"}
        K -->|"yes"| L["included"]
    end
```

- **The frontend fix for the reported record-count bug** is separate. That bug was shared client state, not this query; the two are unrelated and are kept apart so each can be reviewed on its own.
- **Only `findAllRequestsByEmployee` is touched.** `findAllLeaveRequests`, `findAllRequestAssignedToManager` and `getLeaveRequestHistoryByTeam` have the same missing tiebreaker, but they are not what this change is about and widening it would make the diff harder to reason about. Worth a follow-up.
- The metamodel `String` constants (`LeaveRequest_.START_DATE`) already used in this method are left as-is except on the lines being changed, which now use the typed attributes (`LeaveRequest_.startDate`) so the comparisons stay type-checked.

- **Submodule hash / pointer bumps** are intentionally excluded. Gitlink updates are done at **deployment time**, not in feature PRs.

## Testing

| Check | Ran? | Result |
|-------|------|--------|
| Unit / integration tests | yes | **pass** — `LeaveControllerIntegrationTest` 16/16 (9 of them `Get Current User Leave Requests`, covering date-range, pagination, sorting and all-filters-combined — the exact paths changed) |
| Formatter (checkstyle) | yes | pass — `mvn checkstyle:check` BUILD SUCCESS |
| Compile | yes | pass — `mvn compile` BUILD SUCCESS on latest `develop` |
| tsc + lint | n/a | backend-only change |
| e2e (Playwright) | no | suite has no leave specs; the running local mirror currently fronts another workspace's dev server, so a run would not have exercised this branch |

Note: an `okr_company_objective` H2 DDL warning (`year` is reserved) appears during the test run. It is present on clean `develop`, unrelated to this change, and non-fatal — the suite passes.

### DIFF
--- a/backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/LeaveRequestRepositoryImpl.java
+++ b/backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/LeaveRequestRepositoryImpl.java
@@ -56,6 +56,7 @@ import jakarta.persistence.criteria.CriteriaQuery;
 import jakarta.persistence.criteria.Expression;
 import jakarta.persistence.criteria.Join;
 import jakarta.persistence.criteria.JoinType;
+import jakarta.persistence.criteria.Order;
 import jakarta.persistence.criteria.Predicate;
 import jakarta.persistence.criteria.Root;
 import jakarta.persistence.criteria.Subquery;
@@ -771,7 +772,7 @@ public class LeaveRequestRepositoryImpl implements LeaveRequestRepository {
 		Root<LeaveRequest> root = criteriaQuery.from(LeaveRequest.class);
 		buildEmployeeLeavePredicates(criteriaBuilder, root, predicates, employeeId, leaveRequestFilterDto);
 		criteriaQuery.where(predicates.toArray(new Predicate[0]));
-		criteriaQuery.orderBy(QueryUtils.toOrders(page.getSort(), root, criteriaBuilder));
+		criteriaQuery.orderBy(buildStableOrders(criteriaBuilder, root, page));
 
 		TypedQuery<LeaveRequest> query = entityManager.createQuery(criteriaQuery);
 		if (page.isPaged()) {
@@ -793,6 +794,12 @@ public class LeaveRequestRepositoryImpl implements LeaveRequestRepository {
 		return new PageImpl<>(results, page, results.size());
 	}
 
+	private List<Order> buildStableOrders(CriteriaBuilder criteriaBuilder, Root<LeaveRequest> root, Pageable page) {
+		List<Order> orders = new ArrayList<>(QueryUtils.toOrders(page.getSort(), root, criteriaBuilder));
+		orders.add(criteriaBuilder.desc(root.get(LeaveRequest_.leaveRequestId)));
+		return orders;
+	}
+
 	private void buildEmployeeLeavePredicates(CriteriaBuilder criteriaBuilder, Root<LeaveRequest> root,
 			List<Predicate> predicates, Long employeeId, LeaveRequestFilterDto leaveRequestFilterDto) {
 		predicates.add(criteriaBuilder.equal(root.get(LeaveRequest_.EMPLOYEE).get(Employee_.EMPLOYEE_ID), employeeId));
@@ -808,12 +815,12 @@ public class LeaveRequestRepositoryImpl implements LeaveRequestRepository {
 		}
 
 		if (leaveRequestFilterDto.getStartDate() != null && leaveRequestFilterDto.getEndDate() != null) {
-			Predicate dateBetween = criteriaBuilder.or(
-					criteriaBuilder.between(root.get(LeaveRequest_.START_DATE), leaveRequestFilterDto.getStartDate(),
+			Predicate dateOverlap = criteriaBuilder.and(
+					criteriaBuilder.lessThanOrEqualTo(root.get(LeaveRequest_.startDate),
 							leaveRequestFilterDto.getEndDate()),
-					criteriaBuilder.between(root.get(LeaveRequest_.END_DATE), leaveRequestFilterDto.getStartDate(),
-							leaveRequestFilterDto.getEndDate()));
-			predicates.add(dateBetween);
+					criteriaBuilder.greaterThanOrEqualTo(root.get(LeaveRequest_.endDate),
+							leaveRequestFilterDto.getStartDate()));
+			predicates.add(dateOverlap);
 		}
 	}
 


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR makes two targeted fixes to `findAllRequestsByEmployee` (the "My Leave Requests" endpoint): it appends a `leaveRequestId DESC` tie-breaker to the ORDER BY so paging is deterministic, and it replaces the `startDate BETWEEN … OR endDate BETWEEN …` date filter with a proper overlap predicate (`startDate <= filterEnd AND endDate >= filterStart`). Both changes are correct — the overlap form is a strict superset that now also returns leaves fully spanning the filter window, `startDate`/`endDate` are `nullable = false` so no NULL rows are silently dropped, the count query is rebuilt from the same `buildEmployeeLeavePredicates` helper (which is side-effect free, unlike `setDateRangeFiltration`), and existing integration tests in `LeaveControllerIntegrationTest` still hold against `data.sql`. The remaining issues are that the new overlap predicate is verbatim duplicated from an existing private helper, the ordering fix is applied to only 1 of the 5 paginated queries with the same defect, and neither behaviour change is covered by a test.

Found **10** new issue(s): 🟡 6 suggestion(s), 🟣 4 nit(s)

backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/LeaveRequestRepositoryImpl.java
```diff
@@ -808,12 +815,12 @@ private void buildEmployeeLeavePredicates(CriteriaBuilder criteriaBuilder, Root<
 		}
 
 		if (leaveRequestFilterDto.getStartDate() != null && leaveRequestFilterDto.getEndDate() != null) {
-			Predicate dateBetween = criteriaBuilder.or(
-					criteriaBuilder.between(root.get(LeaveRequest_.START_DATE), leaveRequestFilterDto.getStartDate(),
+			Predicate dateOverlap = criteriaBuilder.and(
```
ThinuwanW, line 818: 🤖 **Claude** · 🟡 **Suggestion**

The new `dateOverlap` predicate is byte-for-byte identical to the predicate already built in `setDateRangeFiltration` (same file, lines 659-664: `and(lessThanOrEqualTo(startDate, filterEnd), greaterThanOrEqualTo(endDate, filterStart))`). That is duplicated query logic in two places that must now be kept in sync.

You can't simply call `setDateRangeFiltration` here — it defaults null dates from the leave-cycle config AND mutates the passed `LeaveRequestFilterDto` via `setStartDate`/`setEndDate`, which would be a behaviour change and is also risky given `buildEmployeeLeavePredicates` is invoked twice (data query + count query) with the same DTO instance.

Instead extract just the predicate construction into a small private helper, e.g. `private Predicate buildDateOverlapPredicate(CriteriaBuilder cb, Root<LeaveRequest> root, LocalDate from, LocalDate to)` returning `cb.and(cb.lessThanOrEqualTo(root.get(LeaveRequest_.startDate), to), cb.greaterThanOrEqualTo(root.get(LeaveRequest_.endDate), from))`, and call it from both `buildEmployeeLeavePredicates` and `setDateRangeFiltration`.

backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/LeaveRequestRepositoryImpl.java
```diff
@@ -793,6 +794,12 @@ public Page<LeaveRequest> findAllRequestsByEmployee(Long employeeId, LeaveReques
 		return new PageImpl<>(results, page, results.size());
 	}
 
+	private List<Order> buildStableOrders(CriteriaBuilder criteriaBuilder, Root<LeaveRequest> root, Pageable page) {
```
ThinuwanW, line 797: 🤖 **Claude** · 🟡 **Suggestion**

The non-deterministic-paging bug this helper fixes is not unique to `findAllRequestsByEmployee`. Four sibling methods in this same class paginate a `Root<LeaveRequest>` with the exact same non-unique sort keys (`LeaveRequestSort` only offers `createdDate` and `startDate`) and still call `QueryUtils.toOrders(page.getSort(), root, criteriaBuilder)` with no tie-breaker:

- line 509 — team leave history
- line 551 — `findAllLeaveRequestsByEmployeeId`
- line 896 — `findAllLeaveRequests`
- line 979 — `findAllRequestAssignedToManager`

All four take `Root<LeaveRequest> root` and `Pageable page`, so `buildStableOrders(criteriaBuilder, root, page)` drops in unchanged. Users of those tables will still see rows duplicated across pages or skipped entirely whenever several requests share the same `start_date`/`created_date`. Line 979 is arguably the worst case since it pairs an unstable order with a separately-computed count query. Recommend applying the same one-line change to all four (or explicitly scoping them to a follow-up ticket in the PR description).

backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/LeaveRequestRepositoryImpl.java
```diff
@@ -56,6 +56,7 @@
 import jakarta.persistence.criteria.Expression;
 import jakarta.persistence.criteria.Join;
 import jakarta.persistence.criteria.JoinType;
+import jakarta.persistence.criteria.Order;
```
ThinuwanW, line 59: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Neither behaviour change is covered by a test, even though `LeaveControllerIntegrationTest` already exercises this exact endpoint (`/v1/leave`) with date-range, pagination and sort cases. The two changes are precisely the ones that need regression protection:

1. Overlap semantics — add a case seeding/creating a leave that fully spans the filter window (e.g. leave 2024-12-20 → 2025-02-10 filtered by 2025-01-01 → 2025-01-31) and assert it is now returned. Under the old OR-of-BETWEENs it would have been missed; nothing today would catch a regression back to the old predicate. Note the current `data.sql` (line 299) has no year-spanning row, so the fixture needs one (verify it doesn't break `getLeaveRequests_FutureDateRange_ReturnsEmptyResults`, which asserts `hasSize(0)` for next year).
2. Stable ordering — add a case with several requests sharing the same `start_date`, page through with `size=1`, and assert the union of pages contains no duplicates and no omissions.

backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/LeaveRequestRepositoryImpl.java
```diff
@@ -793,6 +794,12 @@ public Page<LeaveRequest> findAllRequestsByEmployee(Long employeeId, LeaveReques
 		return new PageImpl<>(results, page, results.size());
 	}
 
+	private List<Order> buildStableOrders(CriteriaBuilder criteriaBuilder, Root<LeaveRequest> root, Pageable page) {
+		List<Order> orders = new ArrayList<>(QueryUtils.toOrders(page.getSort(), root, criteriaBuilder));
+		orders.add(criteriaBuilder.desc(root.get(LeaveRequest_.leaveRequestId)));
```
ThinuwanW, line 799: 🤖 **Claude** · 🟣 **Nit**

The tie-breaker is hardcoded to `desc` regardless of the caller's requested `sortOrder`. This is still fully deterministic so pagination is correct either way, but when a user sorts ASC by `startDate` the secondary ordering runs backwards, which reads oddly in the UI for rows sharing a date. Consider mirroring the primary direction, e.g. derive it from `page.getSort()` (`page.getSort().stream().findFirst().map(Sort.Order::isAscending).orElse(false)`) and emit `criteriaBuilder.asc(...)`/`desc(...)` accordingly. Also worth guarding against a duplicate ORDER BY entry if `buildStableOrders` is ever reused for a query whose sort key can be `leaveRequestId` — harmless today since `LeaveRequestSort` only exposes `createdDate` and `startDate`.

backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/LeaveRequestRepositoryImpl.java
```diff
@@ -808,12 +815,12 @@ private void buildEmployeeLeavePredicates(CriteriaBuilder criteriaBuilder, Root<
 		}
 
 		if (leaveRequestFilterDto.getStartDate() != null && leaveRequestFilterDto.getEndDate() != null) {
```
ThinuwanW, line 817: 🤖 **Claude** · 🟣 **Nit**

The date filter is applied only when BOTH `startDate` and `endDate` are non-null; if a caller supplies just one, the filter is silently dropped and the full unfiltered history is returned. This is pre-existing, but the PR is specifically reworking this branch so it's worth resolving now. It's also asymmetric with `setDateRangeFiltration` (line 630) and `setDateRangeBetweenFiltration` (line 671), which both fall back to leave-cycle bounds when a bound is missing — so the same one-sided filter behaves differently on "My Leave Requests" vs. the manager/history endpoints. Either handle the one-sided cases explicitly (only the supplied bound constrains the overlap) or reject the partial input with a `ModuleException` and a message constant.

backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/LeaveRequestRepositoryImpl.java
```diff
@@ -771,7 +772,7 @@ public Page<LeaveRequest> findAllRequestsByEmployee(Long employeeId, LeaveReques
 		Root<LeaveRequest> root = criteriaQuery.from(LeaveRequest.class);
 		buildEmployeeLeavePredicates(criteriaBuilder, root, predicates, employeeId, leaveRequestFilterDto);
 		criteriaQuery.where(predicates.toArray(new Predicate[0]));
-		criteriaQuery.orderBy(QueryUtils.toOrders(page.getSort(), root, criteriaBuilder));
+		criteriaQuery.orderBy(buildStableOrders(criteriaBuilder, root, page));
 
 		TypedQuery<LeaveRequest> query = entityManager.createQuery(criteriaQuery);
```
ThinuwanW, line 777: 🤖 **Claude** · 🟡 **Suggestion**

N+1 on the query this PR is reworking. `LeaveRequest` declares `employee`, `reviewer` and `leaveType` as `@ManyToOne(fetch = FetchType.EAGER)`, and this criteria query selects the entity with no fetch join, so Hibernate issues a secondary SELECT per row per association (plus whatever `Employee` eagerly pulls). The sibling method in this same class (`findAllRequestAssignedToManager`) already guards against this with `root.fetch(LeaveRequest_.leaveType, JoinType.INNER)`. It is worse here because the caller (`LeaveServiceImpl.getCurrentUserLeaveRequests`) uses `Pageable.unpaged(sort)` when `isExport=true`, so the export path runs the N+1 over the employee's entire history with no page cap — and this PR's overlap predicate widens that result set further. Fix: add `root.fetch(LeaveRequest_.leaveType, JoinType.INNER)` and `root.fetch(LeaveRequest_.reviewer, JoinType.LEFT)` (LEFT, since `reviewer_id` is nullable) to the data query only — do not add them to the count query. Note the mapper is what dereferences these, so verify `leaveRequestToLeaveRequestResponseDto` before deciding which associations to fetch.

backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/LeaveRequestRepositoryImpl.java
```diff
@@ -771,7 +772,7 @@ public Page<LeaveRequest> findAllRequestsByEmployee(Long employeeId, LeaveReques
...
 		if (page.isPaged()) {
```
ThinuwanW, line 778: 🤖 **Claude** · 🟡 **Suggestion**

The paging inputs on this endpoint are unbounded on the upper end, which undercuts the 'stabilise pagination' goal. Two concrete problems in the `page.isPaged()` branch: (1) the offset is computed as `page.getPageNumber() * page.getPageSize()` in int arithmetic. `LeaveRequestFilterDto.page` is only constrained by `@Min(0)`, so e.g. `?page=300000000&size=10` overflows to a negative int and `TypedQuery.setFirstResult(...)` throws `IllegalArgumentException`, which is not a `ModuleException` and surfaces as an opaque 500 rather than a 400. Line 244 of this same file already uses the correct `pageable.getOffset()`; use `(int) page.getOffset()` here (or clamp) for consistency. (2) `LeaveRequestFilterDto.size` has `@Min(1)` but no `@Max`, so a single request can ask for an arbitrarily large page, which combined with the widened overlap predicate and the N+1 above is a cheap memory/DB amplification. Add `@Max(<sensible bound>)` to `size` and consider `@Max` on `page`.

backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/LeaveRequestRepositoryImpl.java
ThinuwanW, line 59: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

After this PR there are three implementations of the same date-overlap concept in this one class, not two. Besides the byte-identical copy in `setDateRangeFiltration`, `setDateRangeBetweenFiltration` (lines 704-725) expresses the *same* predicate as a four-branch OR — contained, spanning, starts-before/ends-inside, starts-inside/ends-after — whose union is exactly `start <= filterEnd AND end >= filterStart`. It is therefore 20 lines of redundant SQL that the optimiser has to unpick, and it is used by `getLeaveRequestHistoryByTeam`, `findAllLeaveRequestsByEmployeeId`, `getLeaveRequestsByTeamId` and `findLeaveRequestsByDateRange`. When extracting the shared overlap helper, collapse this one into it too so all leave-list endpoints demonstrably share one definition; otherwise the next semantic tweak has to be made in three places and will drift. If the two helpers are kept separate, the only real difference left is the leave-cycle defaulting (`startMonth`/`endMonth` vs `startMonth - 1`/`endMonth - 1`), which should be the parameter, not the predicate.

backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/LeaveRequestRepositoryImpl.java
ThinuwanW, line 818: 🤖 **Claude** · 🟣 **Nit**

Mixed metamodel reference styles now coexist inside a single method. The new date predicate correctly uses the typed attributes (`LeaveRequest_.startDate`, `LeaveRequest_.endDate`, and `LeaveRequest_.leaveRequestId` in `buildStableOrders`), which is a genuine improvement over the removed `LeaveRequest_.START_DATE` String constants because `Path.get(SingularAttribute)` is type-checked at compile time while `Path.get(String)` resolves at runtime and returns `Path<Object>`. But lines 805, 808-810 and 814 in this same method still use the untyped String constants (`LeaveRequest_.EMPLOYEE`, `LEAVE_TYPE`, `TYPE_ID`, `STATUS`, `Employee_.EMPLOYEE_ID`). Since the method is already being edited, convert those to the typed attributes as well so a future entity rename is caught by the compiler rather than at query-build time.

backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/LeaveRequestRepositoryImpl.java
```diff
@@ -793,6 +794,12 @@ public Page<LeaveRequest> findAllRequestsByEmployee(Long employeeId, LeaveReques
 		return new PageImpl<>(results, page, results.size());
 	}
 
+	private List<Order> buildStableOrders(CriteriaBuilder criteriaBuilder, Root<LeaveRequest> root, Pageable page) {
+		List<Order> orders = new ArrayList<>(QueryUtils.toOrders(page.getSort(), root, criteriaBuilder));
+		orders.add(criteriaBuilder.desc(root.get(LeaveRequest_.leaveRequestId)));
+		return orders;
+	}
+
 	private void buildEmployeeLeavePredicates(CriteriaBuilder criteriaBuilder, Root<LeaveRequest> root,
```
ThinuwanW, line 803: 🤖 **Claude** · 🟣 **Nit**

`buildEmployeeLeavePredicates` dereferences `leaveRequestFilterDto` unconditionally (`getLeaveType()`, `getStatus()`, `getStartDate()`, `getEndDate()`), while both sibling date helpers in this class open with `if (leaveRequestFilterDto != null)`. `findAllRequestsByEmployee` is declared on the public `LeaveRequestRepository` interface (line 58), so a null filter is reachable by contract even though the sole current caller always passes a Spring-bound instance. With this PR the helper is now invoked twice per call (data + count query), so a null would NPE out of the repository as a raw `NullPointerException` -> 500 rather than a `ModuleException`. Either add the same null guard for consistency with `setDateRangeFiltration`/`setDateRangeBetweenFiltration`, or annotate the parameter `@NonNull` (the class already imports `lombok.NonNull` and uses it on other repository methods) so the intent is explicit.

backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/LeaveRequestRepositoryImpl.java
akilarootcode: Deferred - extracting the shared overlap predicate is right, but it touches setDateRangeFiltration's callers, which this PR keeps out.

backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/LeaveRequestRepositoryImpl.java
akilarootcode: Deferred - already listed under Exclusions. buildStableOrders is named to be reused by the four siblings when that happens.

backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/LeaveRequestRepositoryImpl.java
akilarootcode: Correct - the existing 16 tests pass with or without this change, so they do not protect it. Deferred; a spanning-leave case and a tied-sort pagination case are the next commit here.

backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/LeaveRequestRepositoryImpl.java
akilarootcode: Deferred - paging is deterministic either way, so the mirrored direction is cosmetic.

backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/LeaveRequestRepositoryImpl.java
akilarootcode: Deferred - the single-bound behaviour predates this PR and changing it alters what existing callers get back.

backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/LeaveRequestRepositoryImpl.java
akilarootcode: Deferred - the EAGER associations are pre-existing; adding fetch joins changes the query shape beyond this fix.

backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/LeaveRequestRepositoryImpl.java
akilarootcode: Deferred - the offset overflow predates this PR; switching to pageable.getOffset() belongs with a bound on the page input.

backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/LeaveRequestRepositoryImpl.java
akilarootcode: Deferred - collapsing all three overlap expressions is the right end state but is a larger refactor of this class.

backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/LeaveRequestRepositoryImpl.java
akilarootcode: The mixed styles are real. Deferred - converting the rest of the method to typed attributes goes with the shared-predicate extraction.

backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/LeaveRequestRepositoryImpl.java
akilarootcode: Not doing this - CLAUDE.md says to check the API does not already prevent the case before guarding it. The only caller is the Spring-bound controller path, which never supplies null.

hasalarootcode (approved):

PR #1782, opened 2026-08-18
fix: birthday modal name display and overflow/scroll for long names

SanithuGunasena07:
Fixes the Birthday Notification modal so the self view shows only the first name instead of the full name, and long names now wrap and truncate cleanly instead of overflowing and forcing the modal to scroll.

### DIFF
--- a/frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
+++ b/frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
@@ -33,7 +33,7 @@ const BirthdayModalShell: FC<Props> = ({
     <LargeModal
       id={id}
       isOpen
-      className="relative h-[603px] max-h-[85vh] w-[1107px] max-w-[92vw] overflow-hidden rounded-l-[42.69px] rounded-r-[24px] shadow-[0_20px_40px_rgba(0,0,0,0.15)]"
+      className="relative h-auto min-h-[603px] max-h-[85vh] w-[1107px] max-w-[92vw] overflow-hidden rounded-l-[42.69px] rounded-r-[24px] shadow-[0_20px_40px_rgba(0,0,0,0.15)] [&_.overflow-y-auto]:overflow-visible"
       imagePosition="left"
       backdropVariant="dark"
       onClose={onDismiss}
@@ -45,7 +45,7 @@ const BirthdayModalShell: FC<Props> = ({
         </div>
       }
       content={
-        <div className="flex h-[555px] flex-col items-center justify-center gap-4 text-center">
+        <div className="flex min-h-[555px] flex-col items-center justify-center gap-4 text-center">
           {total > 1 && (
             <p className="sr-only">
               {translateAria(["notificationPosition"], {
@@ -66,7 +66,7 @@ const BirthdayModalShell: FC<Props> = ({
                 : employee.firstName
             })}
           />
-          <h2 className="h1 leading-6 tracking-[0.07px] wrap-break-word text-black">
+          <h2 className="h1 line-clamp-2 w-full pr-1 leading-tight tracking-[0.07px] break-words text-black">
             {heading}
           </h2>
           <p className="body2 max-w-104 leading-normal text-secondary-text">
--- a/frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
+++ b/frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
@@ -33,7 +33,7 @@ const BirthdayModalShell: FC<Props> = ({
     <LargeModal
       id={id}
       isOpen
-      className="relative h-auto min-h-[603px] max-h-[85vh] w-[1107px] max-w-[92vw] overflow-hidden rounded-l-[42.69px] rounded-r-[24px] shadow-[0_20px_40px_rgba(0,0,0,0.15)] [&_.overflow-y-auto]:overflow-visible"
+      className="relative h-[603px] max-h-[85vh] w-[1107px] max-w-[92vw] overflow-hidden rounded-l-[42.69px] rounded-r-[24px] shadow-[0_20px_40px_rgba(0,0,0,0.15)]"
       imagePosition="left"
       backdropVariant="dark"
       onClose={onDismiss}
@@ -45,7 +45,7 @@ const BirthdayModalShell: FC<Props> = ({
         </div>
       }
       content={
-        <div className="flex min-h-[555px] flex-col items-center justify-center gap-4 text-center">
+        <div className="flex h-[555px] flex-col items-center justify-center gap-4 text-center">
           {total > 1 && (
             <p className="sr-only">
               {translateAria(["notificationPosition"], {
@@ -66,7 +66,7 @@ const BirthdayModalShell: FC<Props> = ({
                 : employee.firstName
             })}
           />
-          <h2 className="h1 line-clamp-2 w-full pr-1 leading-tight tracking-[0.07px] break-words text-black">
+          <h2 className="h1 line-clamp-2 w-full px-1 leading-tight tracking-[0.07px] wrap-break-word text-black">
             {heading}
           </h2>
           <p className="body2 max-w-104 leading-normal text-secondary-text">
--- a/frontend/src/community/people/components/molecules/BirthdayModals/SelfBirthdayModal.tsx
+++ b/frontend/src/community/people/components/molecules/BirthdayModals/SelfBirthdayModal.tsx
@@ -1,7 +1,6 @@
 import { FC } from "react";
 
 import { useTranslator } from "~community/common/hooks/useTranslator";
-import { concatStrings } from "~community/common/utils/commonUtil";
 import BirthdayModalShell from "~community/people/components/molecules/BirthdayModals/BirthdayModalShell";
 import { EmployeeBirthdayType } from "~community/people/types/BirthdayNotificationTypes";
 
@@ -27,9 +26,7 @@ const SelfBirthdayModal: FC<Props> = ({
       id={MODAL_ID}
       employee={employee}
       heading={translateText(["self", "heading"], {
-        name: employee.lastName
-          ? concatStrings([employee.firstName, employee.lastName])
-          : employee.firstName
+        name: employee.firstName
       })}
       body={translateText(["self", "body"])}
       position={position}


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR makes the birthday modal grow for long headings, clamps the heading to two lines, and drops the last name from the self-birthday heading (making it consistent with ColleagueBirthdayModal, which already used first name only). The i18n keys, aria label wiring and type usage are all correct, and removing the now-unused `concatStrings` import is clean. My main concern is the sizing change on BirthdayModalShell line 36: `min-h-[603px]` overrides `max-h-[85vh]` per the CSS box-sizing rules, and the same class list explicitly disables the modal's internal scroll container, so on short viewports content becomes clipped and unreachable — a regression versus the previous fixed-height + internal-scroll behaviour.

Found **12** new issue(s): 🔴 2 important, 🟡 5 suggestion(s), 🟣 5 nit(s)

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
```diff
@@ -33,7 +33,7 @@ const BirthdayModalShell: FC<Props> = ({
     <LargeModal
       id={id}
       isOpen
-      className="relative h-[603px] max-h-[85vh] w-[1107px] max-w-[92vw] overflow-hidden rounded-l-[42.69px] rounded-r-[24px] shadow-[0_20px_40px_rgba(0,0,0,0.15)]"
+      className="relative h-auto min-h-[603px] max-h-[85vh] w-[1107px] max-w-[92vw] overflow-hidden rounded-l-[42.69px] rounded-r-[24px] shadow-[0_20px_40px_rgba(0,0,0,0.15)] [&_.overflow-y-auto]:overflow-visible"
```

`min-h-[603px]` silently defeats `max-h-[85vh]`. CSS resolves used height as `max(min-height, min(max-height, height))`, so min-height always wins: the modal will render 603px tall even when 85vh is smaller (viewport shorter than ~709px). That alone would be tolerable if the content could still scroll, but the same class list adds `[&_.overflow-y-auto]:overflow-visible`, which kills the scroll container, while the root keeps `overflow-hidden`. Net result: on a viewport shorter than 603px (resized browser window, landscape tablet/phone, split-screen), the modal is taller than the screen, content (body copy, close button) is clipped, and there is no way to scroll to it. Previously `h-[603px] max-h-[85vh]` clamped correctly and the internal `overflow-y-auto` handled the overflow. Suggested fix: keep a real upper bound and keep scrolling available — e.g. use `h-auto max-h-[85vh]` (no `min-h`) and drop the `[&_.overflow-y-auto]:overflow-visible` override, letting the inner scroller take over only in the rare case content exceeds 85vh. If a minimum visual height is genuinely required, express it as `min-h-[min(603px,85vh)]` so it can never exceed the viewport cap. Note every other `LargeModal` in the codebase (e.g. UpgradeToCoreModal.tsx:56) uses the fixed `h-[…] max-h-[85vh] overflow-hidden` pattern; this is the only deviation.

`[&_.overflow-y-auto]:overflow-visible` reaches into the internals of a third-party design-system component (`@rootcodelabs/skapp-ui@1.1.1-beta.13`) by matching one of its Tailwind utility class names. Two problems: (1) it is a descendant selector with no scoping, so it disables scrolling on *any* `overflow-y-auto` element rendered anywhere inside this modal, including future content or nested library components; (2) it breaks silently on any skapp-ui version bump that renames or restructures that wrapper — no type error, no test failure, just a layout regression. This is the only such library-internal override in `src/` apart from SupervisorReassignmentModalSection.tsx:147. Prefer requesting a first-class API from skapp-ui (a `contentClassName`/`disableContentScroll` prop) and, until then, at minimum add a code comment naming the skapp-ui version this selector depends on so the coupling is discoverable.

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
```diff
@@ -66,7 +66,7 @@ const BirthdayModalShell: FC<Props> = ({
                 : employee.firstName
             })}
           />
-          <h2 className="h1 leading-6 tracking-[0.07px] wrap-break-word text-black">
+          <h2 className="h1 line-clamp-2 w-full pr-1 leading-tight tracking-[0.07px] break-words text-black">
```

`break-words` is the deprecated spelling. This project is on Tailwind 4.2 (frontend/package.json), where `break-words` was deprecated in 4.1 in favour of `wrap-break-word` — which is exactly what this line used before the change, and what the rest of the codebase uses (e.g. DealCard.tsx:72). Revert to `wrap-break-word` so the codebase stays on one spelling and does not accumulate a utility that a future Tailwind major will drop.

`pr-1` on a `w-full` heading whose text is centred (inherited `text-center` from the parent) makes the content box asymmetric, so the heading renders ~2px left of true centre relative to the avatar and body copy below it. If the padding is there to stop the descender/ellipsis touching the edge, use `px-1` instead of `pr-1` to keep it visually centred.

`line-clamp-2` truncates a long heading with an ellipsis and gives sighted users no way to read the full text. Screen-reader users are fine because the untruncated `heading` is still passed as `ariaLabel` on line 41, but consider adding `title={heading}` to this `h2` so hovering reveals the full string. Worth confirming with design that silent truncation (rather than shrinking the font or allowing 3 lines) is the intended treatment, given the modal is now free to grow via `h-auto`.

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
```diff
@@ -45,7 +45,7 @@ const BirthdayModalShell: FC<Props> = ({
         </div>
```
ThinuwanW, line 45: 🤖 **Claude** · 🟣 **Nit**

The image pane uses `h-full w-full`, i.e. `height: 100%`, which only resolves against a parent with a definite height. Now that the root modal is `h-auto` rather than `h-[603px]`, this only keeps working if skapp-ui's image column is a flex/grid item that stretches; otherwise `height: 100%` falls back to auto and the `BirthdayCelebration` illustration can collapse. Please verify visually with both a short heading (1 line) and a long clamped heading (2 lines) before merging, and if it collapses, add `self-stretch` or a `min-h-[inherit]` on this wrapper.

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
```diff
@@ -66,7 +66,7 @@ const BirthdayModalShell: FC<Props> = ({
                 : employee.firstName
```
ThinuwanW, line 66: 🤖 **Claude** · 🟣 **Nit**

`employee.lastName` is typed as a required, non-nullable `string` in `EmployeeBirthdayType` (BirthdayNotificationTypes.ts:6), so this `employee.lastName ? concatStrings([...]) : employee.firstName` ternary is unreachable-by-type dead code. Now that SelfBirthdayModal has dropped its copy of the same ternary, this is the last one left. Either make the type honest (`lastName?: string` / `string | null`, if the backend can actually omit it — check the birthday response DTO) and keep the guard, or drop the ternary and just use `concatStrings([employee.firstName, employee.lastName])`. Leaving the type and the runtime guard disagreeing is the pattern the strict-mode checklist calls out.

frontend/src/community/people/components/molecules/BirthdayModals/SelfBirthdayModal.tsx
```diff
@@ -27,9 +26,7 @@ const SelfBirthdayModal: FC<Props> = ({
       id={MODAL_ID}
       employee={employee}
       heading={translateText(["self", "heading"], {
-        name: employee.lastName
-          ? concatStrings([employee.firstName, employee.lastName])
-          : employee.firstName
+        name: employee.firstName
```
ThinuwanW, line 29: 🤖 **Claude** · 🟡 **Suggestion**

This changes user-visible copy (self heading now renders first name only, matching ColleagueBirthdayModal) but there is no test anywhere for the BirthdayModals directory — a repo-wide search for a `*.test.tsx` referencing Birthday returns nothing. Since the heading string is the actual subject of this fix, add a `SelfBirthdayModal.test.tsx` (and ideally a `BirthdayModalShell.test.tsx`) using `@testing-library/react` with the `MockTheme` wrapper, asserting via `screen.getByRole("heading")` that an employee with a last name renders only the first name in the heading while the avatar `alt` still carries the full name. That pins both halves of this change so a future refactor cannot silently reintroduce the last name.

ThinuwanW, line 29: 🤖 **Claude** · 🔴 **Important**

Regression for employees with a blank/missing first name. `partitionBirthdays` (frontend/src/community/people/utils/birthdayNotificationUtils.ts:29) keeps an employee when EITHER name is non-empty — `!!employee.firstName?.trim() || !!employee.lastName?.trim()` — and it defensively uses `?.` even though `EmployeeBirthdayType.firstName` is declared as a non-nullable `string`. The backend backs this up: `Employee.first_name` and `last_name` are both nullable columns (`@Column(name = "first_name", length = 50)`, no `nullable = false`), and `findEmployeeBirthdaysOnByViewerAndScope` applies no name predicate. So `{firstName: "", lastName: "Perera"}` is queued today. Before this change the self heading used `employee.lastName ? concatStrings([firstName, lastName]) : firstName`, which still produced a usable name in that case; now it interpolates `employee.firstName` only, so the user sees "Happy birthday, !" (or a leaked `{{name}}` placeholder if the field is absent rather than empty). The Avatar initials degrade the same way. Note the existing unit tests only cover the both-names-empty case (birthdayNotificationUtils.test.ts "excludes employees with no displayable name"), so this path is untested. Fix: resolve a display name once rather than reading `firstName` raw — e.g. a small shared helper `getBirthdayDisplayName(employee)` returning `employee.firstName?.trim() || employee.lastName?.trim() || ""` — and use it for both modal headings and the Avatar `alt` in BirthdayModalShell. ColleagueBirthdayModal.tsx:29 already has the identical defect ("Today is 's birthday!"), so fixing it in one shared place covers both. Also consider tightening `EmployeeBirthdayType` to `firstName?: string` (or making `partitionBirthdays` require a non-blank `firstName`) so the type stops contradicting the runtime shape.

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
```diff
@@ -45,7 +45,7 @@ const BirthdayModalShell: FC<Props> = ({
         </div>
       }
       content={
-        <div className="flex h-[555px] flex-col items-center justify-center gap-4 text-center">
+        <div className="flex min-h-[555px] flex-col items-center justify-center gap-4 text-center">
```

The two halves of this fix work against each other, which makes the height/overflow changes inert. `line-clamp-2` on the heading hard-caps the only variable-length content in this pane at two lines; the body copy is a static translated string with no interpolation. Worst case the content is roughly avatar + 2 lines of 24px/30px heading + ~4 lines of 14px body + two `gap-4` gaps, i.e. well under 300px — it can never reach the 555px floor, so `min-h-[555px]` is always the used height, `h-auto` on the root never grows anything, and the `[&_.overflow-y-auto]:overflow-visible` override on line 36 is never needed. Net effect: the modal still renders at exactly the old fixed size, while the design-system scroll container has been permanently neutralised for this instance with nothing gained. Pick one strategy: either (a) let it grow and scroll — drop `line-clamp-2`, leave the design-system `overflow-y-auto` alone, and remove the `overflow-visible` override; or (b) clamp — keep `line-clamp-2` plus a `title`/tooltip for the full text and revert `h-auto`/`min-h-[603px]`/`min-h-[555px]`/`overflow-visible` back to the fixed heights. If LargeModal exposes a prop for body scroll behaviour, prefer that over the arbitrary descendant selector. Also worth cleaning up while here: 603px is now hard-coded in three coupled places — the root `min-h-[603px]`, the implicit 555+48 split on this line, and `viewBox="0 0 499 603"` in BirthdayCelebration.tsx — so changing one silently drifts the layout. Extract a single constant or drive the content pane from the parent instead of a second magic number.

frontend/src/community/people/components/molecules/BirthdayModals/SelfBirthdayModal.tsx
ThinuwanW, line 29: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Removing the last-name branch leaves SelfBirthdayModal.tsx and ColleagueBirthdayModal.tsx character-for-character identical apart from two literals: `MODAL_ID` ("birthday-self" vs "birthday-colleague") and the translation segment ("self" vs "colleague"). Same imports, same Props interface, same body. Per the review guidelines (2+ components with near-identical markup → extract a shared component), collapse them: BirthdayModalController already switches on `entry.variant`, so it can derive the id and key segment from the variant and render BirthdayModalShell directly, or a single `BirthdayModal` taking `variant: BirthdayModalVariant` can map variant → `{ modalId, translationKey }`. This also removes the maintenance trap that caused the bug above — the two headings drifted apart precisely because the name-resolution logic was duplicated in two files.

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx

Latent cascade trap worth knowing about when overriding typography on `.h1`. global.css imports `tailwindcss` on line 3 and then declares `.h1 { font-size: 24px; font-weight: 700 }` unlayered on line 91. Tailwind 4 emits its utilities inside `@layer utilities`, and per the CSS cascade-layer rules unlayered declarations beat layered ones regardless of source order or specificity — so `.h1` unconditionally wins for every property it declares. Today that is only font-size/font-weight, so `leading-tight`, `line-clamp-2`, `break-words` and `text-black` all apply fine. But the moment someone adds a `line-height` to `.h1` in global.css, the `leading-tight` here (and `leading-normal` on line 72) will silently stop taking effect with no build or lint error. This is also the only `leading-tight` in the whole frontend, so it is a one-off value; if the design calls for a specific heading leading, prefer setting it on the `.h1` typography class itself (or use the explicit `leading-[30px]` the design specifies) rather than layering a Tailwind utility on top of an unlayered global class.

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
SanithuGunasena07: reverted back

SanithuGunasena07: this was reverted to original from previous comment not valid anymore

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
SanithuGunasena07: reverted to wrap-break-word

SanithuGunasena07: Fixed

SanithuGunasena07: confiremed with design

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
SanithuGunasena07: Not valid as reverted h-auto back to h-[603px], so h-full resolves fine

frontend/src/community/people/components/molecules/BirthdayModals/SelfBirthdayModal.tsx
SanithuGunasena07: not required

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
SanithuGunasena07: not code touched in this PR

frontend/src/community/people/components/molecules/BirthdayModals/SelfBirthdayModal.tsx
SanithuGunasena07: used existing already implemented behaviour from ColleagueBirthdayModal.tsx

SanithuGunasena07: not required for this PR

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
SanithuGunasena07: Stale due to previous comment fix

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
SanithuGunasena07: Valid heads-up for the future, no change needed now.

shakila-kamalasena (approved):

akilarootcode (approved):

PR #1783, opened 2026-08-18
Fix/leave policy demo bug fixes

hasalarootcode:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR migrates the Leave Policies table to the shared `TableView`/`TableV2` component with a popover filter, gates the Leave configurations tab and the Leave Entitlements drawer route behind the leave-policies feature flag, and applies assorted UI polish to the policy wizard, edit modal and policy card. The TableView migration closely follows established patterns in the codebase (CompanyTable, PolicyLeaveRequests) and the removed i18n keys have no dangling references. Two issues need fixing before merge: the Configurations page can end up with no selected tab once the async feature flag resolves, and the new empty-state signal treats a failed policy fetch as "no policies", hiding the page's action buttons and showing a misleading empty state.

Found **26** new issue(s): 🔴 7 important, 🟡 10 suggestion(s), 🟣 9 nit(s)

frontend/pages/community/configurations/index.tsx
```diff
@@ -30,10 +32,12 @@ const Configurations: NextPage = () => {
...
-      tab.requiredRoles.some((role) => userRoles.includes(role))
+    return allTabs.filter(
+      (tab) =>
+        tab.requiredRoles.some((role) => userRoles.includes(role)) &&
+        !(tab.id === "leave" && isLeavePoliciesEnabled)
     );
-  }, [allTabs, user?.roles]);
+  }, [allTabs, user?.roles, isLeavePoliciesEnabled]);
 
   const [activeTab, setActiveTab] = useState(visibleTabs[0]?.id);
```
ThinuwanW, line 42: 🤖 **Claude** · 🔴 **Important**

`activeTab` is initialised once from `visibleTabs[0]?.id`, but `visibleTabs` now changes *asynchronously*: `isLeavePoliciesEnabled` is false while `useGetLeavePolicyConfig` is in flight, so the `leave` tab is present on first render and removed once the query resolves. For a user whose first visible tab is `leave` this leaves `activeTab === "leave"` pointing at a tab that no longer exists, so `visibleTabs.find((tab) => tab.id === activeTab)?.component` is `undefined` and the content area renders blank with no tab highlighted. This is reachable in community mode: `getConfigurationTabs` is not sorted, and for a user holding only `LEAVE_ADMIN` (or `LEAVE_ADMIN` + `CRM_ADMIN`) the role filter drops organization/time/attendance, making `leave` the first visible tab. Fix by keeping the selection in sync, e.g. derive the rendered tab with a fallback (`const currentTab = visibleTabs.find((tab) => tab.id === activeTab) ?? visibleTabs[0]`) and add an effect that calls `setActiveTab(visibleTabs[0]?.id)` when `activeTab` is no longer in `visibleTabs`.

frontend/src/community/leave/components/molecules/LeavePoliciesTable/LeavePoliciesTable.tsx
```diff
@@ -128,134 +115,132 @@ const LeavePoliciesTable: FC<Props> = ({ onCreatePolicy }) => {
...
+                )
+              }
+            : {})
+        };
+      }),
+    [policies, translateText, canManagePolicies, openKebabMenuId, handleActivate]
+  );
 
   const isFiltering = Boolean(debouncedSearch.trim() || leaveTypeFilter);
+  const isEmpty = !isLoading && !isFiltering && policies.length === 0;
```
ThinuwanW, line 217: 🤖 **Claude** · 🔴 **Important**

`isEmpty` conflates "no data" with "fetch failed". `useGetLeavePoliciesInfinite` exposes `isError`, but it is never destructured or used here, so when the policies request fails (network error/5xx) `isLoading` is false and `policies` is `[]`, which makes `isEmpty` true. The consequences are now worse than before this PR: the table renders the "No leave policies yet. / Create your first policy to get started." empty state (telling the admin their data is gone), and the new `onEmptyStateChange` callback also hides the page-level "Create policy" and "Bulk assign" buttons. Destructure `isError` and exclude it (`const isEmpty = !isLoading && !isError && !isFiltering && policies.length === 0;`), and render an explicit error state (or fire an error toast) instead of the empty state when the query fails — per the review checklist, loading *and* error states must be represented in the UI.

frontend/src/community/leave/components/molecules/LeavePoliciesTable/LeavePoliciesTable.tsx
```diff
@@ -38,9 +43,13 @@ import { getLeavePolicyErrorToastKeys } from "~community/leave/utils/leavePolicy
 
 interface Props {
   onCreatePolicy: () => void;
+  onEmptyStateChange?: (isEmpty: boolean) => void;
```
ThinuwanW, line 46: 🤖 **Claude** · 🟡 **Suggestion**

`onEmptyStateChange` pushes derived server state upward from a presentational child through an effect, which is an inverted data flow: the parent re-renders a second time after the child's effect fires, and the flag is duplicated in `useState` in the page while the source of truth lives in the React Query cache. Prefer having the page own the emptiness signal — either call `useGetLeavePoliciesInfinite` in `pages/community/leave/policies/index.tsx` (React Query dedupes the same query key, so no extra request) and compute `isEmpty` there, or expose it via the existing leave store. That also removes the initial flash where the header buttons render and then disappear once the first page resolves.

frontend/src/community/common/utils/getDrawerRoutes.ts
```diff
@@ -336,6 +336,10 @@ const getDrawerRoutes = ({
               return null;
             }
 
+            if (subRoute.id === "2C" && isLeavePoliciesEnabled) {
```
ThinuwanW, line 339: 🤖 **Claude** · 🟡 **Suggestion**

Hiding sub-route `2C` (Leave Entitlements) from the drawer only removes the nav entry — the page itself is still fully reachable and functional when leave policies are enabled. `pages/community/leave/entitlements/leave-entitlements.tsx` still renders the entitlement table and `LeaveCarryForward` (it only hides the bulk-upload button and `CustomLeaveAllocationContent` behind the flag), and `pages/community/leave/entitlements/carry-forward-balances.tsx:165,175` still routes users back to `ROUTES.LEAVE.LEAVE_ENTITLEMENTS`. Bookmarks and the browser back button reach it too. Add a page-level guard (redirect to `ROUTES.LEAVE.LEAVE_POLICIES` when `isLeavePoliciesEnabled` and not loading) so nav visibility and route access agree.

frontend/src/community/leave/components/molecules/LeavePolicyCard/LeavePolicyCard.tsx
```diff
@@ -118,15 +118,10 @@ const LeavePolicyCard = forwardRef<HTMLDivElement, Props>(
...
-          <Stack>
-            <Typography variant="body1" sx={classes.policyName}>
-              {policyName} &nbsp;
-              {isMouseOn && isActionable && getEmoji(leaveType.emojiCode)}
-            </Typography>
-            <Typography variant="caption" sx={classes.leaveTypeName}>
-              {leaveType.name}
-            </Typography>
-          </Stack>
+          <Typography variant="body1">
```
ThinuwanW, line 121: 🤖 **Claude** · 🟡 **Suggestion**

Two behavioural regressions from collapsing the name block into a single `Typography`: (1) the deleted `classes.policyName` carried `whiteSpace: nowrap / overflow: hidden / textOverflow: ellipsis`, so long names were truncated; the new plain `Typography variant="body1"` has no truncation and the parent `leftContent` only sets `overflow: hidden`, so long leave type names will now be clipped mid-glyph or push the layout instead of ellipsising. (2) The card no longer displays `policyName` at all, yet `policyName` is still used in the `cardAriaLabel` and in the disabled-reason toast — if an employee is ever assigned two policies for the same leave type the cards become visually indistinguishable while screen readers still announce distinct names. Re-add the truncation sx to the retained `Typography`, and confirm with design that dropping the policy name from the card face is intended.

frontend/src/community/leave/components/organisms/LeavePolicyWizard/EntitlementSetupStep.tsx
```diff
@@ -234,18 +236,14 @@ const EntitlementSetupStep: FC<Props> = ({
...
-      <WizardSection title={translateText(["fineTuningTitle"])}>
-        <div className="flex max-w-3xl flex-col gap-4">
           <Dropdown
             id="leave-policy-first-accrual"
             label={translateText(["firstAccrualLabel"])}
             value={formData.firstAccrual}
             options={firstAccrualOptions}
             onChange={(value: string) => onChange({ firstAccrual: value })}
-            width="100%"
+            width="50%"
```
ThinuwanW, line 245: 🤖 **Claude** · 🟡 **Suggestion**

`width="50%"` on the `firstAccrual` and `receiveAccruedTime` dropdowns is a fixed percentage with no responsive handling, so these controls stay half-width on small viewports where every other field in this step is full width (`fullWidth`) or uses a breakpoint-aware class (`w-full md:w-64`). Prefer keeping `width="100%"` and wrapping each dropdown in a responsive container (e.g. `<div className="w-full md:w-1/2">`) so the layout matches the rest of the wizard on mobile.

frontend/src/community/common/components/molecules/Stepper/styles.ts
```diff
@@ -11,6 +11,10 @@ const styles = (theme: Theme): StyleProps => ({
       border: `.0625rem dashed ${theme.palette.grey.A100}`,
       borderRadius: 1
     },
+    "& .MuiStepConnector-root.Mui-active .MuiStepConnector-line, & .MuiStepConnector-root.Mui-completed .MuiStepConnector-line":
```
ThinuwanW, line 14: 🤖 **Claude** · 🟡 **Suggestion**

This changes the active/completed connector colour on the *shared* `Stepper` molecule, which is consumed by four flows beyond the leave policy wizard: `AddNewResourceFlow`, `DirectoryAddSectionWrapper`, `LeavePolicyWizard`, and the enterprise `CreateDocumentContainer`. Please verify the other three visually (or scope the override via an `sx` prop passed from `LeavePolicyWizard`) so this doesn't become an unintended cross-module restyle.

frontend/src/community/leave/components/molecules/LeavePoliciesTable/LeavePoliciesTable.tsx
```diff
@@ -270,56 +255,51 @@ const LeavePoliciesTable: FC<Props> = ({ onCreatePolicy }) => {
...
+          searchBar: {
+            value: searchTerm,
+            onChange: handleSearchChange,
+            placeholder: translateText(["searchPlaceholder"]),
+            "aria-label": translateText(["searchPlaceholder"])
+          }
+        }}
+        filter={{
+          filterCount: leaveTypeFilter ? 1 : 0,
+          filterButtonAriaLabel: translateText(["leaveTypeFilterLabel"]),
```
ThinuwanW, line 276: 🤖 **Claude** · 🟣 **Nit**

`filterButtonAriaLabel` is set to `translateText(["leaveTypeFilterLabel"])` ("Leave Type"), so screen readers announce the filter *button* as "Leave Type" rather than describing the control. Every other `TableView` consumer passes a dedicated aria-namespace key for this (`translateAria(["filterButton"])` in `ManagerTimesheetRequestTable`/`EmployeeTimesheetRequestTable`, `translateAria(["filterSection"])` in `PolicyLeaveRequests`). Add a `filterBtn`/`filterButton` key under the leave aria namespace and use that here; keep `leaveTypeFilterLabel` for the `SelectableItemList` title inside the popover, where it already is.

frontend/src/community/leave/components/molecules/LeavePoliciesTable/LeavePoliciesTable.tsx
```diff
@@ -270,56 +255,51 @@ const LeavePoliciesTable: FC<Props> = ({ onCreatePolicy }) => {
...
-      <Table<TableRow>
-        columns={columns}
-        data={tableData}
-        tableAriaLabel={translateText(["title"])}
+      <TableView
+        tableName={TableNames.LEAVE_POLICIES}
+        ariaLabel={{ regionAriaLabel: translateText(["title"]) }}
+        headers={tableHeaders}
+        rows={tableRows}
         isLoading={isLoading}
```
ThinuwanW, line 265: 🤖 **Claude** · 🟣 **Nit**

The deleted `LeavePoliciesTableSkeletonLoader` rendered `LEAVE_POLICY_SKELETON_ROW_COUNT` (8) skeleton rows; no replacement is passed, so `TableView` falls back to its `skeletonRows = 4` default while `LEAVE_POLICY_PAGE_SIZE` is 10. The loading placeholder will now be noticeably shorter than the loaded table, causing a layout jump. Pass `skeletonRows={LEAVE_POLICY_PAGE_SIZE}` (or a `loader` like `CompanyTable` does) to keep the height stable.

frontend/src/community/leave/components/molecules/LeavePoliciesTable/LeavePoliciesTable.tsx
```diff
@@ -128,134 +115,132 @@ const LeavePoliciesTable: FC<Props> = ({ onCreatePolicy }) => {
...
+                            onClick: () => handleActivate(policy)
+                          }
+                    ]}
+                  />
+                )
+              }
+            : {})
+        };
+      }),
+    [policies, translateText, canManagePolicies, openKebabMenuId, handleActivate]
```
ThinuwanW, line 213: 🤖 **Claude** · 🟣 **Nit**

`translateText` is in the dependency arrays of both `tableHeaders` and `tableRows`, but `useTranslator` returns a freshly-created closure on every render (see `src/community/common/hooks/useTranslator.tsx` — it builds and returns a new arrow function each call). The dependency therefore changes every render and both `useMemo`s recompute unconditionally, so the memoisation added here buys nothing (the whole row array, including every `KebabMenu`, is rebuilt on each keystroke in the search box). Either wrap the returned translator in `useCallback` inside `useTranslator`, or drop the `useMemo` wrappers so the code doesn't imply a guarantee it doesn't provide.

frontend/src/community/leave/components/molecules/EditLeavePolicyModal/EditLeavePolicyModal.tsx
```diff
@@ -164,9 +166,13 @@ const EditLeavePolicyModal: FC<EditLeavePolicyModalProps> = ({
         buttonRight: {
           variant: "primary",
           onClick: handleSave,
-          disabled: isPending || !isChanged,
+          disabled: isSaveDisabled,
           isLoading: isPending,
-          icon: <SaveIcon />,
+          icon: (
```
ThinuwanW, line 171: 🤖 **Claude** · 🟣 **Nit**

Dimming the icon with a hand-rolled `opacity-50` class re-implements disabled styling that the design-system button should own — if `SmallModal`'s right button already reduces opacity on `disabled`, the icon ends up double-dimmed relative to the label. Also note that when `isPending` is true the button is in `isLoading` state, so this only really matters for the `!isChanged` case. Prefer fixing the disabled icon treatment in the shared button component (or a theme token) rather than per-call-site.

frontend/src/community/leave/components/molecules/LeavePoliciesTable/LeavePoliciesTable.tsx
```diff
@@ -270,56 +255,51 @@ const LeavePoliciesTable: FC<Props> = ({ onCreatePolicy }) => {
...
+        }}
+        filter={{
+          filterCount: leaveTypeFilter ? 1 : 0,
+          filterButtonAriaLabel: translateText(["leaveTypeFilterLabel"]),
+          popoverId: "leave-policy-leave-type-filter",
+          filterContent: renderFilterContent
+        }}
+        emptyState={
+          isFiltering
+            ? { title: translateText(["noSearchResultsTitle"]) }
```
ThinuwanW, line 282: 🤖 **Claude** · 🟣 **Nit**

`isFiltering` is true for either a search term *or* a leave-type filter, but the shared branch only shows `noSearchResultsTitle` ("No policies found. Try a different search term."). A user who cleared the search and only applied a leave-type chip is told to change their search term. Consider a separate message for the filter-only case, or reword the key to something neutral like "No policies match your search or filters."

frontend/src/community/leave/components/molecules/LeavePolicyFilterBody/LeavePolicyFilterBody.tsx
```diff
@@ -0,0 +1,85 @@
+import {
```
ThinuwanW, line 1: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

New component with no colocated `LeavePolicyFilterBody.test.tsx`. The behaviour worth covering is cheap and non-obvious: chip click toggles selection off when the same leave type is re-clicked, Apply propagates the selected id and closes, Reset clears both local and applied state. Note the sibling filter bodies (`CustomLeaveAllocationFilterBody`, `PolicyLeaveRequestFilterBody`) are also untested, so this is consistent with the codebase rather than a regression — flagging per the testing checklist item. If you add one, remember the `MockTheme` wrapper.

frontend/pages/community/leave/policies/index.tsx
```diff
@@ -22,6 +22,9 @@ const LeavePolicies: NextPage = () => {
     useState<boolean>(false);
   const [isBulkAssignModalOpen, setIsBulkAssignModalOpen] =
     useState<boolean>(false);
+  const [isPoliciesEmpty, setIsPoliciesEmpty] = useState<boolean>(false);
+
+  const showTopActionButtons = canManagePolicies && !isPoliciesEmpty;
```
ThinuwanW, line 27: 🤖 **Claude** · 🔴 **Important**

`showTopActionButtons` is driven by `isEmpty` from the table, which is computed as `!isLoading && !isFiltering && policies.length === 0`. Because `isFiltering` is part of that expression, the signal is wrong in two directions on a tenant with zero policies: (1) the toolbar search box is always rendered by `TableView`, so as soon as the admin types one character `isFiltering` becomes true, `isEmpty` flips back to false, and the top "Create policy" / "Bulk upload" buttons reappear while the table simultaneously says "No policies found. Try a different search term."; (2) `isPoliciesEmpty` is initialised to `false` and `isEmpty` is forced false while `isLoading`, so both buttons render on first paint and then disappear once the first page resolves — a visible flash on every visit. Fix: report emptiness independently of the filter/search state (e.g. `!isLoading && !debouncedSearch && !leaveTypeFilter ? policies.length === 0 : previousValue`, or better, have the parent read the unfiltered total from the query result / a separate count) and treat "still loading" as "not yet known" rather than "not empty" so the buttons don't flash.

frontend/src/community/common/utils/getDrawerRoutes.ts
ThinuwanW, line 339: 🤖 **Claude** · 🔴 **Important**

The new `2C` rule only checks `isLeavePoliciesEnabled`, while the rule immediately above it for `2E` deliberately also consults `isLeavePoliciesConfigError` so it fails open. `isLeavePoliciesEnabled` is `Boolean(data?.isEnabled)`, which is `false` both while the `/leave/policies/config` request is in flight and when it errors. Consequences: (a) on every page load the drawer renders "Leave Entitlements", then removes it a moment later when the config resolves — a nav item that visibly disappears under the cursor; (b) when the config request fails, the drawer shows *both* "Leave Entitlements" and "Leave Policies", i.e. two mutually exclusive modes at once. Fix: thread an `isLeavePoliciesConfigLoading` flag into `Props` (the hook already returns `isLoading` and `Drawer.tsx` already destructures it) and skip the `2C` decision until the config has resolved, mirroring the loading guards already used in `pages/community/leave/types/index.tsx` and `pages/community/leave/entitlements/leave-entitlements.tsx`; also decide explicitly what the error case should show so `2C`/`2E` can't both appear.

frontend/src/community/leave/components/molecules/LeavePolicyCard/LeavePolicyCard.tsx
ThinuwanW, line 121: 🤖 **Claude** · 🔴 **Important**

`policyName` is no longer rendered anywhere in the card, but it is still destructured and still injected into the card's `aria-label` (line 113) and the disabled-reason toast. That leaves two problems: (1) the accessible name announces a policy name that is nowhere in the visible text, which is exactly the WCAG "label in name" mismatch; (2) two policies attached to the same leave type (a supported case — the policy name is the only thing distinguishing them) now render as visually identical cards with no way for a sighted user to tell them apart. The parallel card at `src/community/leave/components/molecules/UserLeavePolicies/LeavePolicyCard.tsx:67-82` shows the intended shape: leave type name as the primary line, policy name as a secondary `truncate`d line. Fix: keep the policy name as a secondary line (with truncation) rather than deleting it, so the visible text and the `aria-label` stay in sync.

frontend/src/community/leave/components/molecules/LeavePolicyFilterBody/LeavePolicyFilterBody.tsx
```diff
@@ -0,0 +1,85 @@
...
+const LeavePolicyFilterBody: FC<Props> = ({
+  appliedLeaveTypeId,
+  onApply,
+  onReset,
+  onClose
+}) => {
+  const translateText = useTranslator("leaveModule", "leavePolicies");
+  const translateFilterText = useTranslator("commonComponents", "filterButton");
+
+  const { data: policyLeaveTypes } = useGetPolicyLeaveTypes({
```
ThinuwanW, line 27: 🤖 **Claude** · 🔴 **Important**

`useGetPolicyLeaveTypes` result is consumed as `policyLeaveTypes?.items ?? []` with no handling of `isLoading`, `isError`, or the genuinely-empty case. Before this PR the leave-type list was fetched by the table on page load and rendered in an always-visible dropdown; now it is fetched lazily when the popover first mounts, so the first open reliably shows an empty chip area with no spinner and no message, and a failed request is indistinguishable from "this tenant has no leave types". Fix: destructure `isLoading`/`isError` and render a loader plus an explicit message — `BasicInfoStep.tsx:83` already does this for the same query with the `createPolicy.noLeaveTypesWarning` key — and/or set `filter.isDisabled` on the `TableView` when there are no leave types (the pattern `CompanyTable.tsx:172-180` uses) so the button isn't offered when it can do nothing.

frontend/src/community/common/assets/languages/english/leaveModule.json
```diff
@@ -488,11 +487,11 @@
       "cancelDismissBtnTxt": "Dismiss",
       "savingBtnTxt": "Saving..."
     },
-    "policyTypeHeader": "POLICY TYPE",
+    "entitlementTypeHeader": "ENTITLEMENT TYPE",
```
ThinuwanW, line 490: 🤖 **Claude** · 🔴 **Important**

The "policy type" → "entitlement type" rename and the "Flexible (unlimited)" → "Flexible" shortening were applied only to the table header (this line) and the shared `flexible` value (line 307), but three sibling strings in the same create-policy flow still use the old wording: `createPolicy.basicInfo.policyTypeTitle` = "Policy type" (the header of `PolicyTypeSelectionModal`, the very first thing the user sees when creating a policy), `createPolicy.basicInfo.flexibleTitle` = "Flexible (unlimited)" (the card the user clicks to pick that type), and `createPolicy.summary.policyTypeLabel` = "Policy Type" (the review step). So the user picks "Flexible (unlimited)" under a heading "Policy type", confirms it as "Policy Type", and then sees it listed as "Flexible" under "ENTITLEMENT TYPE". Fix: update those three keys to match (or revert the header rename) so one concept has one name.

frontend/pages/community/configurations/index.tsx
```diff
@@ -30,10 +32,12 @@ const Configurations: NextPage = () => {
 
   const visibleTabs = useMemo(() => {
     const userRoles = user?.roles || [];
-    return allTabs.filter((tab) =>
-      tab.requiredRoles.some((role) => userRoles.includes(role))
+    return allTabs.filter(
+      (tab) =>
+        tab.requiredRoles.some((role) => userRoles.includes(role)) &&
+        !(tab.id === "leave" && isLeavePoliciesEnabled)
```
ThinuwanW, line 38: 🤖 **Claude** · 🟡 **Suggestion**

This hides the Leave tab only for `isLeavePoliciesEnabled`, but `LeaveConfigurations` itself already returns `null` for four conditions (`!canManageLeavePolicies || isLoading || isError || isLeavePoliciesEnabled`). The gate is now duplicated in two places with different predicates, and the three conditions this filter doesn't cover produce a visible-but-blank tab: a SUPER_ADMIN sees a "Leave" tab whose panel renders nothing while the config request is in flight, and permanently if that request fails. It also can't be reached at all by a PEOPLE_ADMIN even though `requiredRoles` lists them via `AdminTypes.LEAVE_ADMIN`… (`canManageLeavePolicies` is super/leave admin only). Fix: derive one predicate from `useLeavePoliciesEnabled` — hide the tab when `isLeavePoliciesEnabled || isError || !canManageLeavePolicies`, and render a loader (or hold the tab list) while `isLoading` — and drop the duplicated early-return in `LeaveConfigurations`, or expose a single shared `shouldShowLeaveConfigTab` helper used by both. Also consider putting the tab id in a constant instead of the inline `"leave"` string literal.

frontend/pages/community/configurations/index.tsx
```diff
@@ -19,6 +20,7 @@ const Configurations: NextPage = () => {
   const translateText = useTranslator("configurations");
   const environment = useGetEnvironment();
   const isEnterprise = environment === appModes.ENTERPRISE;
+  const { isLeavePoliciesEnabled } = useLeavePoliciesEnabled();
```
ThinuwanW, line 23: 🤖 **Claude** · 🟡 **Suggestion**

`useLeavePoliciesEnabled()` is called with the default `enabled = true`, so `GET /leave/policies/config` now fires for every user who opens the Configurations page — attendance admins, people admins, CRM admins — even though the result is only used to hide a tab that requires SUPER_ADMIN/LEAVE_ADMIN. `Drawer.tsx:133` deliberately gates the same hook (`useLeavePoliciesEnabled(canViewLeavePolicies)`); this page should do the same. The endpoint is also `@PreAuthorize("hasAnyRole('ROLE_LEAVE_EMPLOYEE')")`, so any account without a leave role would take an unnecessary 403 on page load. Fix: `useLeavePoliciesEnabled(useCanManageLeavePolicies())` (or `useCanViewLeavePolicies()`), matching the guideline that a query depending on runtime state gets an `enabled` flag.

frontend/src/community/leave/components/molecules/LeavePolicyFilterBody/LeavePolicyFilterBody.tsx
```diff
@@ -0,0 +1,85 @@
...
+    onReset();
+    onClose();
+  };
+
+  return (
+    <BasicFilterStructure
+      title={translateFilterText(["title"])}
+      resetButtonProps={{
+        onClick: handleReset,
+        disabled: !selectedLeaveTypeId,
```
ThinuwanW, line 65: 🤖 **Claude** · 🟡 **Suggestion**

`resetButtonProps.disabled: !selectedLeaveTypeId` is computed from the *pending* selection rather than the *applied* one, and because this is a single-select toggle the user can easily reach a dead end: with a filter applied, open the popover, click the selected chip to deselect it — local state becomes `""`, so "Reset filters" greys out, while the applied filter is still active and the filter badge still reads 1. The only way out is "Apply filters" with nothing selected, which is unintuitive. Fix: `disabled: !selectedLeaveTypeId && !appliedLeaveTypeId` so Reset stays available whenever something is actually applied.

frontend/src/community/leave/components/molecules/LeavePoliciesTable/LeavePoliciesTable.tsx
```diff
@@ -128,134 +115,132 @@ const LeavePoliciesTable: FC<Props> = ({ onCreatePolicy }) => {
...
-        leaveType: policy,
-        policyType:
-          policy.policyType === PolicyType.ACCRUAL
-            ? translateText(["accrual"])
-            : translateText(["flexible"]),
-        status: policy.status,
-        actions: policy
-      })),
-    [policies, translateText]
+  const handleActivate = useCallback(
```
ThinuwanW, line 118: 🤖 **Claude** · 🟡 **Suggestion**

`handleActivate` was refactored into a `useCallback` but keeps the fragile pattern of stashing the policy name in component state (`setActivatingPolicyName`) purely so that `handleActivateSuccess` can interpolate it into the toast. The success handler closes over `activatingPolicyName`, so correctness depends on the re-render caused by `setActivatingPolicyName` landing before the mutation resolves — and if two activations are ever triggered in quick succession the toast names the wrong policy. Since this code was being touched anyway, prefer passing the name through the mutation instead of through state: either use the per-call form `activateLeavePolicy(policy.id, { onSuccess: () => showToast(policy.name) })`, or read it from the `variables` argument React Query hands to `onSuccess`. That also lets `activatingPolicyName` state be deleted.

frontend/src/community/leave/components/molecules/LeavePoliciesTable/LeavePoliciesTable.tsx
```diff
@@ -270,56 +255,51 @@ const LeavePoliciesTable: FC<Props> = ({ onCreatePolicy }) => {
...
-          <LeavePoliciesTableSkeletonLoader
-            rowCount={LEAVE_POLICY_SKELETON_ROW_COUNT}
-            showActionsColumn={canManagePolicies}
-          />
+        toolbar={{
+          searchBar: {
+            value: searchTerm,
+            onChange: handleSearchChange,
+            placeholder: translateText(["searchPlaceholder"]),
+            "aria-label": translateText(["searchPlaceholder"])
```
ThinuwanW, line 271: 🤖 **Claude** · 🟡 **Suggestion**

The `searchBar` config omits `ariaLabelClearButton`, so the search field's clear button ships with no accessible name. Every other `TableView` search bar in the codebase supplies it (`ContactTable.tsx:226`, `CompanyTable.tsx:169`, `CustomerTable.tsx:168`). Add a translated key (e.g. `leavePolicies.clearSearchAriaLabel`) and pass it, per the guideline that accessibility labels must be translated rather than absent.

frontend/src/community/leave/components/molecules/EditLeavePolicyModal/EditLeavePolicyModal.tsx
```diff
@@ -127,23 +128,24 @@ const EditLeavePolicyModal: FC<EditLeavePolicyModalProps> = ({
             fullWidth
           />
           <div className="flex flex-col gap-1.5">
-            <p className="body2 text-secondary-text">
+            <p className="subtitle1 text-secondary-text">
```
ThinuwanW, line 131: 🤖 **Claude** · 🟣 **Nit**

The two read-only pseudo-labels moved from `body2` (14px / weight 400) to `subtitle1` (16px / weight 500), which makes them larger and heavier than the real `InputField` label rendered directly above them in the same modal, and diverges from the identical label-plus-readonly-value pattern in `AssignLeavePolicyForm.tsx:80,99` which still uses `body2 text-secondary-text`. If the intent was to make them look like form labels, `subtitle3` (14px / 500) is the closer token; otherwise keep `body2` so the module stays internally consistent.

frontend/src/community/leave/components/molecules/LeaveTypeChip/LeaveTypeChip.tsx
```diff
@@ -6,15 +6,17 @@ interface Props {
   name: string;
   emojiCode: string | null;
   className?: string;
+  isDisabled?: boolean;
```
ThinuwanW, line 9: 🤖 **Claude** · 🟣 **Nit**

`isDisabled` is misleading on a non-interactive `<span>` — nothing is actually disabled and no `aria-disabled` is emitted; the prop is purely a muted/read-only visual for the edit modal. `isMuted` or `isReadOnly` would describe it accurately. Separately, applying `grayscale` to the whole chip desaturates the caller-supplied background too, so any future caller that passes a coloured `className` will get an unexpected wash-out; scoping the filter to the emoji span (which is the only thing that actually needs desaturating) would be safer.

frontend/src/community/leave/components/molecules/LeavePoliciesTable/LeavePoliciesTable.tsx
```diff
@@ -270,56 +255,51 @@ const LeavePoliciesTable: FC<Props> = ({ onCreatePolicy }) => {
...
-        hasMore={hasNextPage ?? false}
-        noDataState={noDataState}
-        noSearchResultsState={{
-          title: translateText(["noSearchResultsTitle"])
+        infiniteScroll={{
+          isEnabled: true,
+          height: "34.5rem",
+          hasMore: hasNextPage,
+          isFetchingNextPage,
+          onLoadMore: handleLoadMore
```
ThinuwanW, line 302: 🤖 **Claude** · 🟣 **Nit**

`onLoadMore` is typed `() => void` but `handleLoadMore` is `async`, so the returned promise is silently discarded; the other infinite-scroll `TableView` consumers use the explicit `onLoadMore: () => { void fetchNextPage(); }` form (`ContactTable.tsx:216`, `CompanyTable.tsx:159`). More substantively, there is still no user-visible feedback if fetching a subsequent page fails — the spinner just stops and the list silently ends. Consider matching the sibling call shape and surfacing a toast/retry via the query's error state, consistent with how the activate/deactivate mutations already report failures.

frontend/pages/community/configurations/index.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/molecules/LeavePoliciesTable/LeavePoliciesTable.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/molecules/LeavePoliciesTable/LeavePoliciesTable.tsx
hasalarootcode: Not needed

frontend/src/community/common/utils/getDrawerRoutes.ts
hasalarootcode: Not needed

frontend/src/community/leave/components/molecules/LeavePolicyCard/LeavePolicyCard.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/organisms/LeavePolicyWizard/EntitlementSetupStep.tsx
hasalarootcode: Not needed

frontend/src/community/common/components/molecules/Stepper/styles.ts
hasalarootcode: Not needed

frontend/src/community/leave/components/molecules/LeavePoliciesTable/LeavePoliciesTable.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/molecules/LeavePoliciesTable/LeavePoliciesTable.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/molecules/LeavePoliciesTable/LeavePoliciesTable.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/molecules/LeavePoliciesTable/LeavePoliciesTable.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/molecules/LeavePolicyFilterBody/LeavePolicyFilterBody.tsx
hasalarootcode: Not needed

frontend/pages/community/leave/policies/index.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/molecules/LeavePolicyCard/LeavePolicyCard.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/molecules/LeavePolicyFilterBody/LeavePolicyFilterBody.tsx
hasalarootcode: Not needed

frontend/src/community/common/assets/languages/english/leaveModule.json
hasalarootcode: Not needed

frontend/pages/community/configurations/index.tsx
hasalarootcode: Not needed

frontend/pages/community/configurations/index.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/molecules/LeavePolicyFilterBody/LeavePolicyFilterBody.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/molecules/LeavePoliciesTable/LeavePoliciesTable.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/molecules/LeavePoliciesTable/LeavePoliciesTable.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/molecules/EditLeavePolicyModal/EditLeavePolicyModal.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/molecules/LeaveTypeChip/LeaveTypeChip.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/molecules/LeavePoliciesTable/LeavePoliciesTable.tsx
hasalarootcode: Not needed

frontend/src/community/common/utils/getDrawerRoutes.ts
hasalarootcode: Not needed

frontend/src/community/leave/components/molecules/EditLeavePolicyModal/EditLeavePolicyModal.tsx
hasalarootcode: Not needed

shakila-kamalasena (approved):

akilarootcode (approved):

PR #1784, opened 2026-08-18
fix: tooltip not announced by screen readers

SanithuGunasena07:
Fixed the shared Tooltip component announcing blank instead of its actual text to screen readers

### DIFF
--- a/frontend/src/community/common/components/atoms/Tooltip/Tooltip.tsx
+++ b/frontend/src/community/common/components/atoms/Tooltip/Tooltip.tsx
@@ -95,8 +95,8 @@ const Tooltip: FC<Props> = ({
       title={title}
       TransitionComponent={Fade}
       PopperProps={PopperProps}
-      {...(ariaLabel ? { "aria-label": ariaLabel } : {})}
-      {...(ariaDescription ? { "aria-description": ariaDescription } : {})}
+      ariaLabel={ariaLabel ?? ""}
+      aria-description={ariaDescription ?? ""}
       customstyles={{
         tooltip: {
           maxWidth: maxWidth,
--- a/frontend/src/community/common/components/atoms/Tooltip/Tooltip.tsx
+++ b/frontend/src/community/common/components/atoms/Tooltip/Tooltip.tsx
@@ -95,8 +95,8 @@ const Tooltip: FC<Props> = ({
       title={title}
       TransitionComponent={Fade}
       PopperProps={PopperProps}
-      aria-label={ariaLabel ?? ""}
-      aria-description={ariaDescription ?? ""}
+      {...(ariaLabel ? { "aria-label": ariaLabel } : {})}
+      {...(ariaDescription ? { "aria-description": ariaDescription } : {})}
       customstyles={{
         tooltip: {
           maxWidth: maxWidth,
@@ -131,6 +131,7 @@ const Tooltip: FC<Props> = ({
                   ? theme.palette.error.contrastText
                   : iconColor
             }
+            svgProps={{ "aria-hidden": true, focusable: false }}
           />
         )}
       </span>
--- a/frontend/src/community/common/components/molecules/InputDate/InputDate.tsx
+++ b/frontend/src/community/common/components/molecules/InputDate/InputDate.tsx
@@ -308,7 +308,7 @@ const InputDate: FC<Props> = ({
             id="emoji-field"
             title={tooltip}
             isDisabled={disabled}
-            aria-label={`${lowerCaseLabel} ${translateAria(["ariaLabel"])}`}
+            ariaLabel={`${lowerCaseLabel} ${translateAria(["ariaLabel"])}`}
           />
         )}
       </Stack>


ThinuwanW (commented):
## 🤖 Claude Code Review

This is a focused, correct a11y fix: replacing the unconditional `aria-label={ariaLabel ?? ""}` / `aria-description={ariaDescription ?? ""}` with conditional spreads is exactly right, because MUI v6 merges Tooltip's extra props into the cloned child *after* its own `nameOrDescProps`, so an explicit empty `aria-label` was clobbering MUI's title-derived accessible name — that is the root cause of "tooltip not announced". Marking the decorative info icon `aria-hidden`/`focusable=false` is also correct and does not remove any name the trigger previously had. Remaining concerns are about whether the restored name actually reaches assistive tech: the trigger `<span>` is focusable but role-less (aria-label on `role=generic` is ARIA-prohibited and inconsistently exposed), JSX titles still leave the trigger unnamed now that the icon is hidden, and none of the new behaviour is covered by tests.

Found **13** new issue(s): 🔴 2 important, 🟡 7 suggestion(s), 🟣 4 nit(s)

frontend/src/community/common/components/atoms/Tooltip/Tooltip.tsx
```diff
@@ -95,8 +95,8 @@ const Tooltip: FC<Props> = ({
       title={title}
       TransitionComponent={Fade}
       PopperProps={PopperProps}
-      aria-label={ariaLabel ?? ""}
-      aria-description={ariaDescription ?? ""}
+      {...(ariaLabel ? { "aria-label": ariaLabel } : {})}
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Restoring MUI's derived name only half-solves the reported problem, because the element it lands on is a role-less span. MUI v6 builds the child props as `{...nameOrDescProps, ...other, ...children.props}`, so with `aria-label` no longer forced to "", `nameOrDescProps['aria-label'] = title` (string titles) now reaches the wrapper `<span>` on line 107. That span has `tabIndex=0` but no `role`, i.e. `role=generic`, where `aria-label` is prohibited by ARIA 1.2 and exposed inconsistently across browser/AT combinations (and axe's `aria-prohibited-attr` rule will now start firing on every Tooltip in the app, where the previous empty value was exempt). Give the trigger a real role so the name is valid and reliably announced: for the icon-only variant `role="img"`, or — since the span already handles Enter to toggle via handleKeyDown — `role="button"` with `aria-expanded={isOpen}`. Worth re-verifying with NVDA/VoiceOver after the change rather than assuming the name is picked up.

frontend/src/community/common/components/atoms/Tooltip/Tooltip.tsx
```diff
@@ -131,6 +131,7 @@ const Tooltip: FC<Props> = ({
                   ? theme.palette.error.contrastText
                   : iconColor
             }
+            svgProps={{ "aria-hidden": true, focusable: false }}
```
ThinuwanW, line 134: 🤖 **Claude** · 🟡 **Suggestion**

Hiding the icon closes the last fallback for the JSX-title case, leaving a focusable element with no accessible name at all. MUI only derives `aria-label` from the title when `typeof title === "string"`; for a JSX title it sets `aria-labelledby` **only while the tooltip is open** and `aria-label = null` otherwise. So for `<Tooltip title={<SomeJsx/>} />` with no `ariaLabel`, the closed trigger is a nameless tab stop and the svg is now `aria-hidden`. Real consumers hit this: DropdownList.tsx:135 and MultivalueDropdownList.tsx both type `tooltip?: string | JSX.Element` and forward it as `title` without passing `ariaLabel`. Suggest a fallback in this component — e.g. compute `const resolvedAriaLabel = ariaLabel ?? (typeof title === "string" ? undefined : translateAria(["moreInformation"]))` and spread that — so the icon-only trigger always has a name regardless of title type.

frontend/src/community/common/components/atoms/Tooltip/Tooltip.tsx
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

No test covers the behaviour this PR fixes, so the regression can silently come back. Tooltip.test.tsx currently only asserts rendering/placement. Add cases with `MockTheme`: (1) `render(<Tooltip title="Test"><button/></Tooltip>)` → the wrapper must expose the title as its accessible name and must NOT carry `aria-label=""`; (2) with `ariaLabel="Custom"` → the explicit label wins over the title; (3) the default info-icon svg has `aria-hidden="true"` and `focusable="false"`. Also worth running the full suite once: because MUI forwards these props onto the trigger span, tooltip wrappers now have accessible names and `getByLabelText`/`getByRole(..., {name})` queries elsewhere (e.g. SetupOrganizationForm.test.tsx) could newly match more than one element.

frontend/src/community/common/components/atoms/Tooltip/Tooltip.tsx
ThinuwanW, line 134: 🤖 **Claude** · 🟣 **Nit**

`dataTestId="tooltip-icon"` on this Icon never reaches the DOM, which will bite when writing the test for the new `aria-hidden` attribute. Icon.tsx:367 passes `data-testid` down, but InformationIcon (like most icons) destructures only `{ fill, width, height, id, svgProps, onClick }` from `IconProps` and drops the rest — and TS does not catch it because hyphenated JSX attributes bypass excess-property checking. `screen.getByTestId("tooltip-icon")` therefore fails today. Either add `"data-testid": dataTestId` into the `svgProps` object here, or fix `IconProps`/the icon components to forward `data-testid`.

frontend/src/community/common/components/atoms/Tooltip/Tooltip.tsx
```diff
@@ -95,8 +95,8 @@ const Tooltip: FC<Props> = ({
       title={title}
       TransitionComponent={Fade}
       PopperProps={PopperProps}
-      aria-label={ariaLabel ?? ""}
-      aria-description={ariaDescription ?? ""}
+      {...(ariaLabel ? { "aria-label": ariaLabel } : {})}
+      {...(ariaDescription ? { "aria-description": ariaDescription } : {})}
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Given the stated goal ("tooltip not announced by screen readers"), note that `aria-description` is an ARIA 1.3 *draft* attribute with essentially no support in NVDA, JAWS or VoiceOver — anything passed via `ariaDescription` is still not announced after this change; it just stops emitting an empty attribute. It is an established pattern across this codebase (IconButton, KebabMenu, TableBody, IconChip), so no need to change it here, but if the description content matters for the affected screens, switch to `aria-describedby` pointing at a visually hidden element (or fold the text into the tooltip `title`).

ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Now that the blanket empty `aria-label` is gone, MUI derives the child's `aria-label` from a string `title`, so the two heaviest callers announce the same text twice. InputField.tsx passes `title={tooltip}` and `ariaDescription={tooltip}` (lines 231/234) and DropdownList.tsx does the same with `ariaDescription={tooltip as string}` (lines 135/139) — the wrapper span ends up with name = tooltip text and description = the identical tooltip text, so a screen reader reads it as "<text>, <text>". Fix at the call sites: drop `ariaDescription` where it is just a copy of `title` (the derived name already conveys it), or keep `ariaDescription` only when it adds information beyond the title. Worth doing in this PR since the duplication only becomes audible because of this change.

frontend/src/community/common/components/atoms/Tooltip/Tooltip.tsx
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Behavioural drift for the `children` call sites: MUI spreads its own props (including the title-derived name) onto the child it clones, which here is the wrapper span rendered at line 107 — and that span is a tab stop by default (`tabIndex` defaults to 0 when `isDisabled` is false). Previously the span was nameless (`aria-label=""`), now it is named with the tooltip text while the control inside it already carries its own label: Timer.tsx:92/98 (span "clockOut" then button "clockOut"), ClockWidget.tsx:145/155, AvatarChip.tsx:133/136. Result is an extra named tab stop that repeats the inner control's name. Note PolicyLeaveBalanceCard.tsx:74 and LeaveEntitlementBalanceCard.tsx:73 already pass `tabIndex={-1}` for exactly this reason, so the mitigation exists but is applied inconsistently. Suggested fix inside this component: default `tabIndex` to -1 when `children` is provided (the child is expected to be the focusable/named element) and keep 0 only for the internal-icon branch.

ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

`handleKeyDown`, `handleFocus` and `handleBlur` (lines 49-66) call `setOpen`, but `isOpen` at line 86 ignores internal state whenever `controlledOpen` is defined — so for every controlled call site these handlers are dead code and the tooltip cannot be opened by keyboard at all. InputField.tsx is the concrete case: it passes `open={open}` and drives it only from `onMouseOver`/`onMouseLeave` on the wrapping Box (lines 225-226), with no focus handling, so a keyboard user can focus the trigger and never see the tooltip. Since this PR is about making the tooltip reachable by assistive tech, either guard the focus/keydown handlers the way `handleMouseEnter` does (making the no-op explicit), or expose `onOpen`/`onClose` callbacks so controlled parents can react to focus as well as hover.

ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Missing paired change: InputDate.tsx:311 passes `aria-label={...}` (hyphenated DOM form) to this component, but `Props` only declares `ariaLabel`, so the value is destructured away and never rendered — and TypeScript does not catch it because JSX attribute names containing a hyphen are exempt from excess-property checking. That tooltip trigger therefore has no author-supplied name and now depends entirely on MUI deriving one from `title`. Rename it to `ariaLabel` at that call site, and consider grepping for other hyphenated `aria-*` props handed to this atom since they will always be silently dropped.

frontend/src/community/common/components/atoms/Tooltip/Tooltip.tsx
ThinuwanW, line 134: 🤖 **Claude** · 🟡 **Suggestion**

The decorative-icon hiding only covers the fallback branch of this component. Callers that supply `children` render the very same information icon themselves and leave it exposed — PolicyLeaveBalanceCard.tsx:85 and LeaveEntitlementBalanceCard.tsx:85 both render the icon inside a button/`role="button"` with no `svgProps` — so the fix is half-applied across the tooltip trigger surface. Also note the `ariaLabel` prop on Icon (Icon.tsx:173) is forwarded as `aria-label` to the leaf icon components at Icon.tsx:370, but `IconProps` has no such field and leaf icons like InformationIcon only spread `svgProps`, so that prop is inert — `svgProps` is currently the only channel that works. Cleanest fix is central: in Icon.tsx default the rendered svg to `aria-hidden` when no accessible name is supplied, and make the leaf icons accept/forward a real label, rather than hand-marking each call site.

frontend/src/community/common/components/atoms/Tooltip/Tooltip.tsx
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Two WCAG 1.4.13 (Content on Hover or Focus) gaps remain in the same handlers this PR is fixing around: `handleKeyDown` (line 52) handles only Enter, so there is no Escape-to-dismiss for the tooltip content, and `handleMouseLeave` closes immediately, so the pointer cannot travel into the tooltip to read/select long content (LeaveEntitlementBalanceCard and PolicyLeaveBalanceCard render multi-row tables in `title`). Add an `event.key === "Escape"` branch that calls `setOpen(false)` and stops propagation, and consider a small close delay or letting MUI's own hover handling manage the popper for the uncontrolled case.

ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

The conditional object spreads can be simplified — React already omits an attribute whose value is `undefined`, so `aria-label={ariaLabel || undefined}` and the equivalent for `aria-description` give identical DOM output with less noise, and keep the props readable in place. This is also how the rest of the codebase does it (KebabMenu.tsx:83 and IconButton.tsx:66 pass the optional value straight through).

ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

There is a second Tooltip in play: EmployeeTimeRecordsTable.tsx:2/209 uses the design-system `Tooltip` from `@rootcodelabs/skapp-ui` (with a `content` prop) rather than this atom. If that component has the same empty-`aria-label` pattern, this fix leaves that call site broken and the two will drift further apart. Worth confirming with the design-system package and filing the paired fix upstream, or noting in the PR description that only the community atom is in scope.

SanithuGunasena07: Tested with Narrator and the label announces fine as-is. Skipping role="button" since a tooltip isn't really a button.

SanithuGunasena07: not required

frontend/src/community/common/components/atoms/Tooltip/Tooltip.tsx
SanithuGunasena07: Pre-existing for JSX titles the old code forced aria-label="" regardless of title type, so this isn't something the diff introduced. No current caller passes JSX to tooltip either. Leaving out of scope for this fix.

SanithuGunasena07: Pre-existing dataTestId="tooltip-icon" was already on this line before our diff, and the drop happens in IconProps/icon components generally, not something we introduced. Out of scope for this fix.

frontend/src/community/common/components/atoms/Tooltip/Tooltip.tsx
SanithuGunasena07: matches the existing pattern elsewhere (IconButton, KebabMenu, etc.), so not changing it here.

SanithuGunasena07: out of scope for PR

frontend/src/community/common/components/atoms/Tooltip/Tooltip.tsx
SanithuGunasena07: Out of scope for this PR touches shared Tooltip behavior across multiple consumers,

SanithuGunasena07: pre-existing and separate from this fix doesn't affect the aria-label announcement since that's independent of the popup's open state.

SanithuGunasena07: renamed to ariaLabel

frontend/src/community/common/components/atoms/Tooltip/Tooltip.tsx
SanithuGunasena07: both the per-call-site icon hiding (PolicyLeaveBalanceCard, LeaveEntitlementBalanceCard) and the central Icon.tsx refactor are pre-existing, unrelated to this diff, and the pattern isn't consistently reachable via props today.

frontend/src/community/common/components/atoms/Tooltip/Tooltip.tsx
SanithuGunasena07: pre-existing and structural (same handlers as the earlier controlled/focus finding), not introduced by this diff.

SanithuGunasena07: Not safe undefined still overwrites MUI's title-derived default in its internal spread merge, reintroducing the bug. Keeping the conditional spread.

SanithuGunasena07: different component from a separate design-system package (@rootcodelabs/skapp-ui), not this atom out of scope for this PR

frontend/src/community/common/components/atoms/Tooltip/Tooltip.tsx
shakila-kamalasena: Check if we can use undefined instead of spreading

shakila-kamalasena (changes requested):

frontend/src/community/common/components/atoms/Tooltip/Tooltip.tsx
```diff
@@ -95,8 +95,8 @@ const Tooltip: FC<Props> = ({
       title={title}
       TransitionComponent={Fade}
       PopperProps={PopperProps}
-      aria-label={ariaLabel ?? ""}
```
akilarootcode, line 98: "aria-label"

shakila-kamalasena (approved):

frontend/src/community/common/components/atoms/Tooltip/Tooltip.tsx
SanithuGunasena07: outdated

akilarootcode (approved):

ThinuwanW (approved):

PR #1785, opened 2026-08-18
fix: leave policy bulk upload icon change

thusala:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds a new ExportArrowIcon component, registers it in the IconName enum and IconMapping, and swaps the Leave Policies 'Bulk upload' button from UP_ARROW_ICON to the new icon. The implementation is correct and follows the established icon conventions exactly (IconProps signature, svgProps spread, alphabetical import, Partial<Record<>> mapping — no typing or dead-code issues, and UP_ARROW_ICON remains in use elsewhere so nothing is orphaned). All findings are consistency/polish concerns rather than defects: the same 'bulk upload' button on four other pages still renders the old icon, and the new icon's default size and fill deviate from the icon it replaces.

frontend/pages/community/leave/policies/index.tsx
```diff
@@ -46,7 +46,7 @@ const LeavePolicies: NextPage = () => {
         canManagePolicies ? translateText(["bulkAssignBtnTxt"]) : undefined
       }
       secondaryBtnVariant="tertiary"
-      secondaryBtnIconName={IconName.UP_ARROW_ICON}
+      secondaryBtnIconName={IconName.EXPORT_ARROW_ICON}
```
ThinuwanW, line 49: 🤖 **Claude** · 🟡 **Suggestion**

Only this page was switched to EXPORT_ARROW_ICON, but four other buttons with identical 'bulk upload' semantics still render IconName.UP_ARROW_ICON: pages/community/leave/entitlements/leave-entitlements.tsx:83 (`leaveEntitlements.bulkUploadBtnTxt`), pages/community/people/directory/index.tsx:58 and pages/community/people/directory/pending.tsx:75 (both `peoples.addBulkPeople`), and pages/enterprise/people/directory/index.tsx:176. After this change the same user action shows two different glyphs depending on which page you are on. If the new icon is the intended design for bulk upload, update those four call sites in this PR too; if the change is intentionally scoped to Leave Policies only, it's worth confirming with design, since the divergence will be visible to users navigating between Leave and People.

frontend/src/community/common/assets/Icons/ExportArrowIcon.tsx
```diff
@@ -0,0 +1,32 @@
+import { JSX } from "react";
+
+import { IconProps } from "~community/common/types/IconTypes";
+
+const ExportArrowIcon = ({
+  fill = "currentColor",
+  width = "14",
```
ThinuwanW, line 7: 🤖 **Claude** · 🟣 **Nit**

Default dimensions are 14x14, while the UpArrowIcon this replaces defaults to width="16" height="17". ContentLayout.tsx renders the secondary-button icon as `<Icon name={secondaryBtnIconName} fill={secondaryBtnIconFill} />` without passing width/height, so these defaults are what actually render — the button icon will be ~2px smaller here than the visually equivalent buttons on the other pages listed above. If the smaller size is deliberate per the design spec, ignore; otherwise default to width="16" height="16" (the viewBox is square, so no distortion) to match the surrounding buttons.

frontend/src/community/common/assets/Icons/ExportArrowIcon.tsx
```diff
@@ -0,0 +1,32 @@
+import { JSX } from "react";
+
+import { IconProps } from "~community/common/types/IconTypes";
+
+const ExportArrowIcon = ({
+  fill = "currentColor",
```
ThinuwanW, line 6: 🤖 **Claude** · 🟣 **Nit**

`fill = "currentColor"` deviates from the convention used by essentially every other icon in this directory (UpArrowIcon and ShareIcon default to "black", FileUploadIcon to "#71717A"); StarIcon is the only other currentColor user. Inheriting the button's text color is arguably the better behavior — it makes the icon track hover/disabled states automatically — but ContentLayout passes `fill={secondaryBtnIconFill}` which is undefined at this call site, so the default is what renders, and ButtonV2 comes from the external `@rootcodelabs/skapp-ui` package, so the effective inherited color for the `tertiary` variant can't be verified from this repo. Please confirm visually that the icon is legible in the tertiary button's default, hover, and disabled states before merging.

frontend/src/community/common/assets/Icons/ExportArrowIcon.tsx
```diff
@@ -0,0 +1,32 @@
+import { JSX } from "react";
```
ThinuwanW, line 1: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

Naming: the component/enum is 'ExportArrow', but the only consumer is the Leave Policies button labelled 'Bulk upload' (leaveModule.json: `bulkAssignBtnTxt` = "Bulk upload"), which opens BulkAssignPolicyModal — an instructions -> CSV upload -> summary import flow. The glyph (arrow out of a tray) is the conventional upload/share mark, so the visual is fine, but the name reads as the opposite direction of the action. Since this icon is now globally registered and future devs will pick it by name from the IconName enum, consider naming it UPLOAD_ARROW_ICON / UploadArrowIcon (or a direction-neutral SHARE_ARROW) to avoid someone later reaching for it expecting a data-export affordance.

frontend/src/community/common/assets/Icons/ExportArrowIcon.tsx
```diff
@@ -0,0 +1,32 @@
...
+  id,
+  svgProps,
+  onClick
+}: IconProps): JSX.Element => {
+  return (
+    <svg
+      id={id}
+      width={width}
+      height={height}
+      viewBox="0 0 14 14"
```
ThinuwanW, line 18: 🤖 **Claude** · 🟡 **Suggestion**

This glyph is the vertical mirror of the existing DownloadIcon (arrow out of a tray vs arrow into a tray), but it was authored on a different frame with a different technique, so the two will not read as a pair. DownloadIcon uses a 16x16 viewBox, three `stroke={fill}` paths at strokeWidth 1.25 with round caps/joins, and defaults to 16x16; this one uses a 14x14 viewBox, a single solid `fill` path, and defaults to 14x14. That matters concretely because pages/enterprise/people/directory/index.tsx:173-177 toggles the *same* secondary button between IconName.DOWNLOAD_ICON and the bulk-upload arrow — once EXPORT_ARROW_ICON replaces UP_ARROW_ICON there (and on the other bulk-upload buttons), the download and upload states of one button will render at different pixel sizes and different optical weights (solid vs 1.25pt hairline). Separately, the exported path's bounding box is only ~13.33 x 13.35 units inside the 14x14 viewBox, and it sits flush against the top-left (x starts at 0, y starts at -0.017), leaving all the slack on the right and bottom — the glyph is not optically centred in its own frame. Fix: re-export the artwork on a 16x16 frame as the true mirror of DownloadIcon (stroke paths, strokeWidth 1.25, strokeLinecap/strokeLinejoin round) and default width/height to "16"; at minimum, normalise the viewBox to the real bbox (or translate the paths by ~0.33) so it is centred.

frontend/src/community/common/assets/Icons/ExportArrowIcon.tsx
```diff
@@ -0,0 +1,32 @@
...
+const ExportArrowIcon = ({
+  fill = "currentColor",
+  width = "14",
+  height = "14",
+  id,
+  svgProps,
+  onClick
+}: IconProps): JSX.Element => {
+  return (
+    <svg
```
ThinuwanW, line 14: 🤖 **Claude** · 🟣 **Nit**

Icon.tsx forwards `data-testid` (Icon.tsx:369) and `aria-label` (Icon.tsx:372) into the mapped icon component, but IconProps declares neither and this component destructures a closed prop list with no rest spread — so both are silently dropped and `<Icon name={IconName.EXPORT_ARROW_ICON} dataTestId="..." ariaLabel="..." />` is a no-op with no type error to warn you. Also, the glyph here is purely decorative (the ButtonV2 it sits in already has the visible label "Bulk upload"), and the svg accepts an `onClick` without any keyboard affordance. Suggest adding `aria-hidden="true"` and `focusable="false"` to the svg so assistive tech and tab order skip it, and either accepting `dataTestId`/`ariaLabel` explicitly or spreading a rest object. I verified this is the pattern in all 179 files in this directory (0 forward data-testid or aria-label), so it is pre-existing rather than introduced — flagging only because this is a brand-new file and the cheapest place to stop propagating it.

frontend/pages/community/leave/policies/index.tsx
ThinuwanW, line 49: 🤖 **Claude** · 🟣 **Nit**

With the new icon defaulting to `currentColor` and no `secondaryBtnIconFill` passed here, ContentLayout renders `<Icon name={secondaryBtnIconName} fill={undefined} />` (ContentLayout.tsx:353-356), so the icon colour is decided entirely by whatever `color` the tertiary ButtonV2 variant resolves to at that DOM node — including across its hover/focus/disabled states. That is not verifiable from this repo, since ButtonV2 comes from the out-of-tree `@rootcodelabs/skapp-ui` package. The only bulk-upload button in the codebase that pins its icon colour does so explicitly: pages/enterprise/people/directory/index.tsx:178 passes `secondaryBtnIconFill="var(--color-primary-text)"`. Suggest passing an explicit `secondaryBtnIconFill` token here as well so the rendered colour is deterministic and matches the design system rather than inheriting whatever the button variant happens to set.

thusala: No need. this is the requirement in here

frontend/src/community/common/assets/Icons/ExportArrowIcon.tsx
thusala: this is an existing icon

frontend/src/community/common/assets/Icons/ExportArrowIcon.tsx
thusala: this is the existing way

frontend/src/community/common/assets/Icons/ExportArrowIcon.tsx
thusala: this is comes from skapp ui

frontend/src/community/common/assets/Icons/ExportArrowIcon.tsx
thusala: this is comes from skapp ui

frontend/src/community/common/assets/Icons/ExportArrowIcon.tsx
thusala: no need

frontend/pages/community/leave/policies/index.tsx
thusala: no need

shakila-kamalasena (approved):

akilarootcode (approved):

PR #1786, opened 2026-08-18
Fix: people configurations UI changes

SanithuGunasena07:
Switched Row label color to #52525B

### DIFF
--- a/frontend/src/community/common/components/atoms/SwitchRow/SwitchRow.tsx
+++ b/frontend/src/community/common/components/atoms/SwitchRow/SwitchRow.tsx
@@ -61,7 +61,7 @@ const SwitchRow: FC<SwitchComponentProps> = ({
             ...classes.label,
             color: disabled
               ? theme.palette.grey.A100
-              : theme.palette.common.black
+              : theme.palette.text.secondary
           }}
           id={labelId}
           component="label"


ThinuwanW (commented):
## 🤖 Claude Code Review

Single-line change swapping the SwitchRow label color from `theme.palette.common.black` to `theme.palette.text.secondary`. The change is correct and idiomatic: `text.secondary` resolves to `#52525B` (frontend/src/community/common/theme/theme.ts:323), exactly matching the commit's stated intent, and it is stable across all six themes since `themeSelector.ts`/`muiThemeOverride` only overrides `palette.primary` and `palette.secondary` and leaves `palette.text` untouched. No test asserts on the label color (SwitchRow.test.tsx only checks label presence, toggle behavior, disabled state, error text, and wrapper styles), and #52525B on white is ~7.5:1 contrast, so WCAG AA is preserved. The only remaining concerns are design-system consistency with sibling form-label atoms and the unchanged error-state branch of the same ternary.

Found **7** new issue(s): 🔴 1 important, 🟡 4 suggestion(s), 🟣 2 nit(s)

frontend/src/community/common/components/atoms/SwitchRow/SwitchRow.tsx
```diff
@@ -61,7 +61,7 @@ const SwitchRow: FC<SwitchComponentProps> = ({
             ...classes.label,
             color: disabled
               ? theme.palette.grey.A100
-              : theme.palette.common.black
+              : theme.palette.text.secondary
```
ThinuwanW, line 64: 🤖 **Claude** · 🟡 **Suggestion**

SwitchRow's label now uses `theme.palette.text.secondary` (#52525B), but every sibling form-label atom/molecule still renders labels in `common.black`: InputField.tsx:200 (`theme.palette.common.black`), molecules/ColorInputField/styles.ts:11, molecules/MultiSelectChipInput/styles.ts:7, molecules/SearchBox/styles.ts:6, and atoms/ColorPaletteSkeleton/styles.ts:9. Several screens render these side by side in the same form — e.g. community/attendance/.../AttendanceConfiguration.tsx, community/leave/.../LeaveTypeForm.tsx, community/configurations/.../PeopleConfigurations.tsx and the enterprise settings organisms (GeoFencingSettings, FingerprintSettings, GoogleWorkspaceSyncSettings, signConfigurations) — so after this change toggle labels will visibly differ in tone from adjacent text-field labels in the same form. If the design intent is that all field labels move to #52525B, apply the same token change to the atoms/molecules listed above in this PR so the design system stays coherent; if only toggles are meant to be de-emphasized, no change is needed but it's worth confirming with design, since the divergence is currently undocumented.

frontend/src/community/common/components/atoms/SwitchRow/SwitchRow.tsx
```diff
@@ -61,7 +61,7 @@ const SwitchRow: FC<SwitchComponentProps> = ({
             ...classes.label,
             color: disabled
```
ThinuwanW, line 62: 🤖 **Claude** · 🟣 **Nit**

The ternary being modified handles only `disabled` vs default, with no `error` branch — so when the `error` prop is set, the label stays #52525B while InputField.tsx:196-200 recolors its label to `theme.palette.error.contrastText` in the same situation. This is pre-existing behavior, but the change makes the label lower-contrast than before, which slightly weakens the visual error signal for a component that already renders the error text separately. If the design system expects errored field labels to turn red, consider extending this to `disabled ? theme.palette.grey.A100 : error ? theme.palette.error.contrastText : theme.palette.text.secondary`. Flagging as a nit since the error message itself is announced via `role="alert"`/`aria-live="assertive"` on line 84-85, so accessibility is not impacted — this is purely a visual-consistency judgment call for the developer/designer.

frontend/src/community/common/components/atoms/SwitchRow/SwitchRow.tsx
ThinuwanW, line 64: 🤖 **Claude** · 🔴 **Important**

The tooltip icon and the optional `icon` rendered *inside* this label were not updated with it, so an enabled SwitchRow now shows a #52525B label next to a solid black info icon. `Tooltip` is rendered without `iconColor`, so its `InformationIcon` falls back to its own default (`fill = "black"` in src/community/common/assets/Icons/InformationIcon.tsx), and `<Icon name={icon} />` likewise gets no `fill`. Note that the disabled state was deliberately paired — Tooltip.tsx uses `theme.palette.grey.A100` for the icon when `isDisabled`, exactly matching the disabled label color here — so this PR breaks that pairing only for the enabled state. Real call sites are affected: PeopleConfigurations.tsx:167 and :177 both pass `tooltip`. Fix: compute the label color once into a local (e.g. `const labelColor = disabled ? theme.palette.grey.A100 : theme.palette.text.secondary`), use it for the Typography, and pass it down as `iconColor={labelColor}` to `Tooltip` and `fill={labelColor}` to `Icon`.

ThinuwanW, line 64: 🤖 **Claude** · 🟡 **Suggestion**

Now that the enabled branch uses a semantic text token, the disabled branch of this same ternary is the odd one out: it still uses `theme.palette.grey.A100` (#A1A1AA) while the codebase's semantic disabled-text token is `theme.palette.text.disabled` (#71717A), used by the closest analogue (DropdownList/styles.ts `labelStyle`, which does `isDisabled ? theme.palette.text.disabled : ...`) plus ~10 other input/label components (Search, FilterSearch, InputPhoneNumber, DropdownAutocomplete, MultiValueDropdownList, signConfigurations...). Switching the disabled branch to `theme.palette.text.disabled` in the same PR keeps the atom internally consistent — but if you do, also update the disabled icon color in Tooltip.tsx:129 (`grey.A100`) so the label and its tooltip icon stay in step.

ThinuwanW, line 64: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

This atom has 26 call sites (community + the `src/enterprise` git submodule, e.g. ModulePageCard, GeoFencingSettings, GoogleWorkspaceSyncSettings, signConfigurations, RecipientDetailsSection, TemplateRolesSection), and the new color is hard-coded inline with no override path: the only styling prop is `wrapperStyles`, which lands on the parent `Stack`, and a parent's inherited `color` cannot override the child Typography's explicit `color`. Any call site that needs the previous emphasis has no way to opt out, and the enterprise submodule cannot adapt within this PR. Follow the established DropdownList pattern instead: move the state-dependent color into the colocated styles.ts as a parameterized entry (`labelStyle: (isDisabled: boolean, error: boolean) => ({...})`, matching DropdownList/styles.ts:15 — note this means dropping the `StyleProps = Record<string, SxProps>` return annotation, as ToggleSwitch/styles.ts already does), and add an optional `labelStyles?: SxProps<Theme>` prop merged last via the existing `mergeSx` helper.

ThinuwanW, line 64: 🤖 **Claude** · 🟡 **Suggestion**

This label Typography has no `variant`, so it renders as `body1` (400 / 1rem per theme.ts). With the color now `text.secondary`, the label becomes pixel-identical to section description text in at least one consumer: GoogleWorkspaceSyncSettings.tsx renders `<Typography variant="body1" color={theme.palette.text.secondary}>` for the section description immediately above two SwitchRows, so the field labels and the descriptive paragraph are now the same size, weight, and color, collapsing the visual hierarchy of that block. Worth confirming with the design spec whether the #52525B token was intended for switch *labels* or only for secondary/descriptive text; if labels are meant to stay distinguishable, either keep an emphasis difference on the label or update the adjacent description text to a lighter token.

frontend/src/community/common/components/atoms/SwitchRow/SwitchRow.tsx
```diff
@@ -61,7 +61,7 @@ const SwitchRow: FC<SwitchComponentProps> = ({
...
               ? theme.palette.grey.A100
-              : theme.palette.common.black
+              : theme.palette.text.secondary
           }}
           id={labelId}
```
ThinuwanW, line 66: 🤖 **Claude** · 🟣 **Nit**

While you're in this element: `labelId` is rendered as the label's `id` but nothing ever references it — the `Toggle` receives `ariaLabel={getAriaLabel()}` rather than `aria-labelledby={labelId}`, and `component="label"` has no `htmlFor`, so the visible label is not programmatically associated with the control. The prop is effectively dead, and ModulePageCard.tsx:89-93 shows the intent leaking out: it passes `labelId={module.title}` with no `label` (so the id is never rendered here) and duplicates `id={module.title}` on its own Typography. If the skapp-ui `Toggle` supports it, wire `aria-labelledby={labelId}` (or `htmlFor` plus a real input id) and drop the `arialabel` fallback for the labelled case; otherwise remove `labelId` from the props so it doesn't read as a working association.

frontend/src/community/common/components/atoms/SwitchRow/SwitchRow.tsx
SanithuGunasena07: Out of scope for this fix this PR only addresses the Birthday Notification labels per the ticket.

frontend/src/community/common/components/atoms/SwitchRow/SwitchRow.tsx
SanithuGunasena07: pre-existing, not introduced by this change.

frontend/src/community/common/components/atoms/SwitchRow/SwitchRow.tsx
SanithuGunasena07: Icon color is intentionally left as black as its used like that in other places as well

SanithuGunasena07: pre-existing disabled styling, out of scope for this ticket

SanithuGunasena07: out of scope for PR

SanithuGunasena07: Not required matches the ticket spec (#52525B for labels)

frontend/src/community/common/components/atoms/SwitchRow/SwitchRow.tsx
SanithuGunasena07: Not required pre-existing accessibility structure,

thusala (approved):

akilarootcode (approved):

PR #1787, opened 2026-08-18
[Fix] Reject re-uploaded error reports in leave policy bulk upload

thusala:
## Summary

On the leave policy bulk upload flow, re-uploading a downloaded **error report** was accepted as if it were a filled-in template: the rows were submitted and a success toast was shown. The error report carries the three template columns **plus an extra `Error` column**, and header validation only checked that the required columns were *present* — extra columns were ignored. This PR makes the upload validate against the template schema and reject any file carrying columns outside it, so an error report is refused before it reaches the API.


ThinuwanW (commented):
## 🤖 Claude Code Review

The change is small, well-typed, and test-covered: a new UNEXPECTED_COLUMNS validation state threaded consistently through the enum, validation result type, util, and UI switch, with the blank-trailing-header edge case handled and tested. My main concern is scope: the check rejects any unrecognized column, not just the error-report column, so CSVs with harmless extra columns (which previously validated fine because buildBulkAssignPayload only reads the three known headers) are now hard-rejected, and the error message misdiagnoses that case. The rest are duplication/polish items in the growing validation-result boilerplate and display of raw file-supplied header names.

Found **13** new issue(s): 🔴 2 important, 🟡 4 suggestion(s), 🟣 7 nit(s)

frontend/src/community/leave/utils/bulkAssignPolicyUtils.ts
```diff
@@ -80,6 +90,20 @@ export const validateBulkAssignCsv = (
     return {
       error: BulkAssignCsvError.MISSING_COLUMNS,
       missingColumns,
+      unexpectedColumns: [],
+      payload: null
+    };
+  }
+
+  const unexpectedColumns = getUnexpectedBulkAssignHeaders(
```
ThinuwanW, line 98: 🤖 **Claude** · 🔴 **Important**

The new check is broader than the stated intent ("reject re-uploaded error reports"): it rejects ANY column that isn't one of the three template headers. Before this PR, extra columns were harmless — buildBulkAssignPayload only reads employeeName/policyName/effectiveDate via getCell, so a CSV with e.g. a 'Department', 'Notes' or 'Employee ID' column validated and produced a correct payload. Those uploads now fail outright, which is a behavioural regression for a common real-world case (spreadsheets exported from other HR tools, or admins keeping their own bookkeeping column). It also blocks the natural remediation loop: download error report → fix the failing rows → re-upload, since the user must now manually delete the 'Error' column first. If the goal is only to stop a verbatim error-report re-upload, narrow the detection to the error-report column, e.g. compare against the translated errorHeader value (pass it into validateBulkAssignCsv the same way downloadBulkAssignErrorReport receives it) and return UNEXPECTED_COLUMNS only when that column is present, otherwise keep ignoring extras. If a blanket reject really is the product decision, please confirm it and update the instructions text (leavePolicies.bulkAssignModal.addPoliciesDescription) to state that no extra columns are allowed, since today it only says to keep the header names unchanged.

frontend/src/community/common/assets/languages/english/leaveModule.json
```diff
@@ -286,6 +286,7 @@
       "unreadableFileError": "The uploaded file could not be read. Please try again with a valid CSV file.",
       "tooManyRowsError": "The uploaded file has more than {{maxRows}} rows. Please split it into smaller files.",
       "missingColumnsError": "The uploaded file is missing required columns: {{columns}}. Please use the CSV template.",
+      "unexpectedColumnsError": "The uploaded file has columns that are not in the CSV template: {{columns}}. Please upload a filled-in template, not an error report.",
```
ThinuwanW, line 292: 🤖 **Claude** · 🟡 **Suggestion**

The message asserts a cause the check cannot actually determine: it fires for any unrecognized column, so a user who kept a 'Notes' or 'Department' column is told "Please upload a filled-in template, not an error report", which is confusing and doesn't tell them what to do. Reword to be generic and actionable, e.g. "The uploaded file has columns that are not in the CSV template: {{columns}}. Remove the extra columns, or download the template and fill it in again." If the error-report round-trip is the case you specifically want to call out, only mention it when the error column is detected (see the related comment in bulkAssignPolicyUtils.ts).

frontend/src/community/leave/utils/bulkAssignPolicyUtils.ts
```diff
@@ -80,6 +90,20 @@ export const validateBulkAssignCsv = (
     return {
       error: BulkAssignCsvError.MISSING_COLUMNS,
       missingColumns,
+      unexpectedColumns: [],
```
ThinuwanW, line 93: 🤖 **Claude** · 🟡 **Suggestion**

The validation-result boilerplate is now duplicated six times, and this PR had to touch every branch just to add one field — the next field will cost six more edits and the same risk of missing one. Extract a small factory, e.g. `const invalidResult = (error: BulkAssignCsvError, overrides: Partial<BulkAssignCsvValidation> = {}): BulkAssignCsvValidation => ({ error, missingColumns: [], unexpectedColumns: [], payload: null, ...overrides });` then each branch becomes `return invalidResult(BulkAssignCsvError.UNEXPECTED_COLUMNS, { unexpectedColumns });`. This also makes the success branch's explicit empty arrays unnecessary noise.

frontend/src/community/leave/utils/bulkAssignPolicyUtils.ts
```diff
@@ -57,6 +57,16 @@ const getMissingBulkAssignHeaders = (
   );
 };
 
+const getUnexpectedBulkAssignHeaders = (
```
ThinuwanW, line 60: 🤖 **Claude** · 🟣 **Nit**

getUnexpectedBulkAssignHeaders and getMissingBulkAssignHeaders are the same set-difference in opposite directions, and validateBulkAssignCsv now evaluates `parseResult.meta.fields ?? []` twice. Consider hoisting `const fields = parseResult.meta.fields ?? [];` once in validateBulkAssignCsv and collapsing the two helpers into one shared diff helper (e.g. `const diffHeaders = (candidates: string[], against: string[]): string[] => { const set = new Set(against.map(normalizeHeader)); return candidates.filter((c) => !set.has(normalizeHeader(c))); }`), with the blank-field filter applied at the call site for the unexpected case.

frontend/src/community/leave/utils/bulkAssignPolicyUtils.ts
```diff
@@ -57,6 +57,16 @@ const getMissingBulkAssignHeaders = (
...
+  fields: string[],
+  headers: BulkAssignCsvHeaders
+): string[] => {
+  const expected = new Set(toCsvValues(headers).map(normalizeHeader));
+  return fields.filter(
```
ThinuwanW, line 65: 🤖 **Claude** · 🟣 **Nit**

The returned column names are raw, file-supplied strings rendered straight into the inline upload error. Three small hardening/polish points: (1) they aren't trimmed, so a header like " Error " is displayed with its surrounding whitespace even though it's compared normalized; (2) they aren't deduplicated, so a file with both 'Error' and 'error' lists the column twice; (3) the list is unbounded, so a wide/garbage CSV can produce a very long error string in DragAndDropField's customError. Suggest `.map((field) => field.trim())` plus deduping by normalized value, and capping the displayed list in the component (e.g. first 5 then "+N more").

frontend/src/community/leave/utils/bulkAssignPolicyUtils.ts
```diff
@@ -80,6 +90,20 @@ export const validateBulkAssignCsv = (
...
+      unexpectedColumns: [],
+      payload: null
+    };
+  }
+
+  const unexpectedColumns = getUnexpectedBulkAssignHeaders(
+    parseResult.meta.fields ?? [],
+    headers
+  );
+  if (unexpectedColumns.length > 0) {
```
ThinuwanW, line 102: 🤖 **Claude** · 🟣 **Nit**

Ordering consideration: the unexpected-columns check now runs before the `parseResult.errors.length > 0` (MALFORMED_ROWS) check. For a file whose quoting is genuinely broken, papaparse can produce nonsense field names, and the user will now get "columns that are not in the CSV template: <garbage>" instead of the more accurate "some rows could not be read" message. Moving the unexpected-columns check after the MALFORMED_ROWS check would keep parse failures reported as parse failures; the missing-columns check should stay first as it is today.

frontend/src/community/leave/utils/__test__/bulkAssignPolicyUtils.test.ts
```diff
@@ -128,6 +130,36 @@ describe("validateBulkAssignCsv", () => {
...
+        ["Employee Name", "Policy Name", "Effective Date", "", "  "]
+      ),
+      headers
+    );
+
+    expect(validation.error).toBeNull();
+    expect(validation.payload).not.toBeNull();
+  });
+
   it("reports missing columns before any other problem", () => {
```
ThinuwanW, line 163: 🤖 **Claude** · 🟣 **Nit**

The precedence test ("reports missing columns before any other problem") uses fields with no unexpected column, so it doesn't actually cover the new branch's placement. Add a case where a file has both problems — e.g. fields ['Error'] only — asserting `error === MISSING_COLUMNS` and `unexpectedColumns` is `[]`. That's the realistic "uploaded a completely different CSV" path, and it pins down the intentional decision that the unexpected list is not surfaced in the missing-columns branch.

frontend/src/community/leave/components/molecules/BulkAssignPolicyModals/BulkAssignPolicyUploadStep.tsx
```diff
@@ -97,6 +97,10 @@ const BulkAssignPolicyUploadStep: FC<Props> = ({ onComplete, onBack }) => {
         return translateText(["missingColumnsError"], {
           columns: validation.missingColumns.join(CSV_DELIMITER)
         });
+      case BulkAssignCsvError.UNEXPECTED_COLUMNS:
```
ThinuwanW, line 100: 🤖 **Claude** · 🔴 **Important**

Adding UNEXPECTED_COLUMNS required a manual edit to this switch, and nothing forces that edit. The switch ends in `default: return ""` (line 112), so any BulkAssignCsvError member that is not explicitly cased falls through to an empty string. Because `handleParseComplete` does `setFileError(getValidationErrorText(validation))` and `setAssignmentPayload(validation.payload)` (null on every error), an unmapped error produces a silent dead end: no inline message in DragAndDropField, and the Confirm button disabled with no explanation. This PR happened to remember the case; the next added member may not. Fix: make the mapping exhaustive at compile time. Either replace the switch body with a `Record<BulkAssignCsvError, (v: BulkAssignCsvValidation) => string>` lookup (TypeScript then errors on a missing key when the enum grows), or keep the switch but handle `error === null` explicitly up front and end with an exhaustiveness guard in the default branch that assigns `validation.error` to a `never`-typed local before returning "". A stronger variant is to model BulkAssignCsvValidation as a discriminated union on `error` so each error variant carries only its own payload.

frontend/src/community/leave/utils/__test__/bulkAssignPolicyUtils.test.ts
```diff
@@ -128,6 +130,36 @@ describe("validateBulkAssignCsv", () => {
...
+
+    expect(validation).toEqual({
+      error: BulkAssignCsvError.UNEXPECTED_COLUMNS,
+      missingColumns: [],
+      unexpectedColumns: ["Error"],
+      payload: null
+    });
+  });
+
+  it("ignores blank trailing headers that spreadsheets append", () => {
```
ThinuwanW, line 150: 🤖 **Claude** · 🟡 **Suggestion**

The "ignores blank trailing headers that spreadsheets append" test proves less than its name claims, and the behaviour it asserts likely does not hold end to end. `buildParseResult` hand-builds the ParseResult with `errors: []`, so the test only exercises the new `field.trim() !== ""` filter — it never verifies what PapaParse actually emits for a header row with trailing commas. In a real run with `header: true`, a blank trailing header column means the header row has more fields than data rows that omit that trailing cell, and PapaParse pushes a FieldMismatch/TooFewFields entry into `parseResult.errors`. `validateBulkAssignCsv` checks `parseResult.errors.length > 0` immediately after the unexpected-columns check, so the file the carve-out was written to accept would still be rejected as MALFORMED_ROWS ("Some rows in the uploaded file could not be read") — the carve-out only helps for exports that pad every row to the same column count. Fix: add a test that feeds a real CSV string through `parse()` with the production options (`header: true`, `skipEmptyLines: "greedy"`) and asserts the file is accepted, and if FieldMismatch does fire, either drop the blank-header carve-out as ineffective or filter FieldMismatch errors attributable to blank trailing columns before the MALFORMED_ROWS check.

frontend/src/community/common/assets/languages/english/leaveModule.json
ThinuwanW, line 292: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Missing paired copy change for the partial-failure recovery flow. `summaryPartial`, `successToastPartialDescription` and `errorToastAllFailedDescription` all direct the user to the error report, and BulkAssignPolicySummaryStep renders the "Download Error Report" button, which writes a CSV containing the three template headers plus the `errorHeader` ("Error") column. That file is now guaranteed to be rejected on re-upload, yet no copy anywhere tells the user what to do instead — `addPoliciesDescription` only says "Keep the column headers as they are", which a user reading it while holding the error report will read as instruction to keep the Error column too. Add explicit recovery guidance: either extend `addPoliciesDescription`/`summaryPartial` to say the error report is for reference only and failed rows must be re-entered in a fresh template (or the Error column deleted before re-upload), or, if the retry loop is meant to be supported, have the validator strip the known `errorHeader` column rather than reject the file.

frontend/src/community/leave/utils/bulkAssignPolicyUtils.ts
```diff
@@ -57,6 +57,16 @@ const getMissingBulkAssignHeaders = (
   );
 };
 
+const getUnexpectedBulkAssignHeaders = (
+  fields: string[],
+  headers: BulkAssignCsvHeaders
+): string[] => {
+  const expected = new Set(toCsvValues(headers).map(normalizeHeader));
+  return fields.filter(
+    (field) => field.trim() !== "" && !expected.has(normalizeHeader(field))
```
ThinuwanW, line 66: 🤖 **Claude** · 🟣 **Nit**

Inconsistent notion of "blank header" inside a file that otherwise routes every header comparison through `normalizeHeader`. This filter uses a raw `field.trim() !== ""` while the very next clause on the same line calls `normalizeHeader(field)`, so there are now two definitions of header canonicalisation in one expression. Use `normalizeHeader(field) !== ""` so the blank check and the equality check agree by construction and any future change to `normalizeHeader` (e.g. stripping zero-width or non-breaking characters that Excel emits) applies to both.

frontend/src/community/leave/components/molecules/BulkAssignPolicyModals/BulkAssignPolicyUploadStep.tsx
```diff
@@ -97,6 +97,10 @@ const BulkAssignPolicyUploadStep: FC<Props> = ({ onComplete, onBack }) => {
...
+        return translateText(["unexpectedColumnsError"], {
+          columns: validation.unexpectedColumns.join(CSV_DELIMITER)
```
ThinuwanW, line 102: 🤖 **Claude** · 🟣 **Nit**

`CSV_DELIMITER` is a file-format constant (defined as "," in leavePolicyConstants alongside MAX_BULK_ASSIGN_ROWS and used by `toCsvRow` to build CSV output), and it is being reused here as a presentation-layer list separator. This copies the existing `missingColumnsError` line rather than fixing it, and it renders as `Error,Notes,Department` with no spaces, which reads poorly in a prose sentence. It also means a future change to the CSV delimiter would silently reformat user-facing error text. Introduce a small display separator (`", "`) or use `Intl.ListFormat` for the join, and update the existing `missingColumns` join at line 98 the same way.

frontend/src/community/leave/components/molecules/BulkAssignPolicyModals/BulkAssignPolicyUploadStep.tsx
ThinuwanW, line 100: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

No component-level test coverage exists for this modal step — there is no `__test__` directory anywhere under frontend/src/community/leave/components, so BulkAssignPolicyUploadStep has never had a test. This PR adds a new user-facing error branch whose only coverage is the pure validator; the translation-key lookup, the interpolation of `{{columns}}`, and the fact that the message actually reaches DragAndDropField's `customError` (and therefore the aria-live region) are all unverified, and a typo in the `unexpectedColumnsError` key would ship as a raw key string on screen. Consider adding a `BulkAssignPolicyUploadStep.test.tsx` with the MockTheme wrapper that drops a CSV containing an extra column and asserts via `screen.getByText`/`getByRole` that the rendered message names the offending column and that the Confirm button is disabled.

frontend/src/community/leave/utils/bulkAssignPolicyUtils.ts
thusala: Agreed, and fixed by narrowing the check rather than confirming the blanket reject.

`getUnexpectedBulkAssignHeaders` is gone. It is replaced by `hasErrorReportColumn(fields, errorHeader)` (`bulkAssignPolicyUtils.ts:64`), which matches only the translated `errorHeader` value, normalised through the same `normalizeHeader` the rest of the file uses. `errorHeader` is now a third parameter on `validateBulkAssignCsv`, passed from the upload step the same way `downloadBulkAssignErrorReport` receives it.

Net behaviour:

| CSV columns | Before this fix | Now |
|---|---|---|
| template + `Error` | rejected | rejected (`ERROR_REPORT_UPLOAD`) |
| template + `Department`, `Notes` | rejected | accepted, extras ignored |
| template + blank trailing column | accepted | accepted |
| template only | accepted | accepted |

So the third-party-HR-export and own-bookkeeping-column cases you called out are no longer a regression, and `addPoliciesDescription` stays accurate as written since extra columns are allowed again. The enum member was renamed `UNEXPECTED_COLUMNS` to `ERROR_REPORT_UPLOAD` and `BulkAssignCsvValidation.unexpectedColumns` was dropped, so the validation shape is back to what it was before this PR plus one error member.

New test pins the accepted case: "accepts extra columns other than the error column".

frontend/src/community/common/assets/languages/english/leaveModule.json
thusala: Fixed. The key is renamed and the message no longer has to guess at a cause, because the check now only fires for the error-report column (see the reply on `bulkAssignPolicyUtils.ts`).

`leaveModule.json:289`:

> "This looks like a downloaded error report, which cannot be uploaded. Re-enter the failed rows in a fresh copy of the CSV template and upload that instead."

No `{{columns}}` interpolation any more, since there is exactly one column it can be about. A file with a `Notes` or `Department` column never reaches this message now.

frontend/src/community/leave/utils/bulkAssignPolicyUtils.ts
thusala: Fixed. `bulkAssignPolicyUtils.ts:87`:

```ts
const invalidCsv = (
  error: BulkAssignCsvError,
  missingColumns: string[] = []
): BulkAssignCsvValidation => ({ error, missingColumns, payload: null });
```

Every failure branch is now one line, e.g. `return invalidCsv(BulkAssignCsvError.MISSING_COLUMNS, missingColumns);`. Only the success branch spells the object out, which is the branch that actually carries a payload.

Since the narrowing removed `unexpectedColumns` from `BulkAssignCsvValidation` entirely, the field that prompted the six edits is gone too, but the factory is worth keeping for the reason you gave.

frontend/src/community/leave/utils/bulkAssignPolicyUtils.ts
thusala: Both parts fixed, though the second one resolved itself.

`parseResult.meta.fields ?? []` is now evaluated once, hoisted at `bulkAssignPolicyUtils.ts:97`, and passed to the checks below it.

The shared diff helper is not needed: narrowing the detection to the error-report column deleted `getUnexpectedBulkAssignHeaders` outright, so there is no opposite-direction set difference left to collapse. `getMissingBulkAssignHeaders` is the only header comparison remaining, and `hasErrorReportColumn` is a single normalised equality check rather than a set difference.

frontend/src/community/leave/utils/bulkAssignPolicyUtils.ts
thusala: No fix needed, all three points are moot after the narrowing.

The validator no longer returns a list of column names. `BulkAssignCsvValidation.unexpectedColumns` is removed, and `errorReportUploadError` interpolates nothing, so there is no file-supplied string rendered into `DragAndDropField`'s `customError` on this path. That removes the untrimmed-display, duplicate-listing and unbounded-length concerns together, rather than hardening against them.

The one place a file-supplied list still reaches the UI is `missingColumnsError`, but those names come from `toCsvValues(headers)`, our own translated header values, not from the uploaded file.

frontend/src/community/leave/utils/bulkAssignPolicyUtils.ts
thusala: No fix needed, because narrowing the check removes the failure mode.

The concern was that garbage field names from broken quoting would be reported as unexpected columns. `hasErrorReportColumn` now only returns true when a field normalises to exactly the translated `errorHeader` ("Error"), so nonsense field names fall through to the `parseResult.errors.length > 0` check and are still reported as `MALFORMED_ROWS`.

I kept the error-report check ahead of `MALFORMED_ROWS` deliberately: what the file *is* is more useful to the user than a row-level complaint about it. A real error report parses cleanly anyway, so in practice only a file that is both an error report and malformed is affected, and for that file "this is an error report" is the more actionable message.

Added a test that pins this: "reports rows short of the header count as malformed, not as an error report", driving a real `parse()` whose rows are one field short.

frontend/src/community/leave/utils/__test__/bulkAssignPolicyUtils.test.ts
thusala: Fixed. The precedence test now uses `fields: ["Error"]`, the "uploaded a completely different CSV" path, and asserts the full validation object, so it pins both the error and the absence of any error-report signal in that branch:

```ts
it("reports missing columns before the error report check", () => {
  const validation = validate(buildParseResult([], ["Error"]));

  expect(validation).toEqual({
    error: BulkAssignCsvError.MISSING_COLUMNS,
    missingColumns: ["Employee Name", "Policy Name", "Effective Date"],
    payload: null
  });
});
```

`unexpectedColumns` no longer exists to assert on, so the object assertion covers the equivalent intent.

frontend/src/community/leave/components/molecules/BulkAssignPolicyModals/BulkAssignPolicyUploadStep.tsx
thusala: Fixed, using the `Record` variant you suggested. `BulkAssignPolicyUploadStep.tsx:94`:

```ts
const errorTextByType: Record<
  BulkAssignCsvError,
  (validation: BulkAssignCsvValidation) => string
> = {
  [BulkAssignCsvError.MISSING_COLUMNS]: (validation) => ...,
  [BulkAssignCsvError.ERROR_REPORT_UPLOAD]: () => ...,
  [BulkAssignCsvError.MALFORMED_ROWS]: () => ...,
  [BulkAssignCsvError.EMPTY_FILE]: () => ...,
  [BulkAssignCsvError.TOO_MANY_ROWS]: () => ...
};

const getValidationErrorText = (validation: BulkAssignCsvValidation): string =>
  validation.error ? errorTextByType[validation.error](validation) : "";
```

There is no `default` branch left to swallow an unmapped member: adding one to `BulkAssignCsvError` without a matching key is now a compile error, and `error === null` is handled explicitly up front instead of falling through to `""`. Verified by adding a throwaway enum member locally, `tsc` failed on the Record until the key was added.

I did not go as far as the discriminated union. `MISSING_COLUMNS` is the only member carrying a payload, so the union would add five variants to remove one array field; happy to revisit if a second payload-bearing error shows up.

frontend/src/community/leave/utils/__test__/bulkAssignPolicyUtils.test.ts
thusala: Good catch on the test proving less than its name, but I checked what PapaParse actually emits before changing behaviour, and the carve-out does hold for the case it was written for. Driving the production options (`header: true`, `skipEmptyLines: "greedy"`):

| CSV | `meta.fields` | `errors` |
|---|---|---|
| `Employee Name,Policy Name,Effective Date,` + row `John Doe,Annual,01/06/2026,` | `[..., ""]` | `[]` |
| same header, row `John Doe,Annual,01/06/2026` (unpadded) | `[..., ""]` | `FieldMismatch/TooFewFields` |

So the padded case, which is what Excel and Google Sheets emit since they pad every row to the sheet's column count, is accepted end to end. `FieldMismatch` only fires when rows are genuinely short of the header count, and that file was already rejected as `MALFORMED_ROWS` before this PR, so it is not something the carve-out was meant to rescue.

I kept the carve-out and added the test you asked for, plus its negative. A new `describe("validateBulkAssignCsv with a real parsed CSV")` feeds real CSV strings through `parse()` with the production options:

- "accepts a spreadsheet export whose rows are padded to a blank trailing column" — asserts `error` is null and the payload is built
- "reports rows short of the header count as malformed, not as an error report" — asserts `MALFORMED_ROWS`
- "rejects the CSV that downloadBulkAssignErrorReport produces" — builds the CSV with `toCsvRow`, so the writer and validator are checked against each other rather than against a hand-written string

The blank-header guard also does real work now for a second reason: `hasErrorReportColumn` returns early when the normalised `errorHeader` is empty, otherwise a missing translation would make every blank trailing column look like an error report. Covered by "does not treat a blank error header as a match on a blank column".

frontend/src/community/common/assets/languages/english/leaveModule.json
thusala: Addressed, though not by adding copy to the instructions step.

The recovery guidance now lives in the message the user hits at the moment they attempt the wrong thing (`leaveModule.json:289`):

> "This looks like a downloaded error report, which cannot be uploaded. Re-enter the failed rows in a fresh copy of the CSV template and upload that instead."

I preferred that over extending `addPoliciesDescription`, because the instructions step is read before the first upload, long before a user has an error report in hand, and pre-emptive prose about a file they do not have yet is noise for the common path.

`addPoliciesDescription` also no longer misleads: after the narrowing, "Keep the column headers as they are" is accurate, and extra columns really are allowed. The only file that is rejected is one carrying the `Error` column, and that rejection now explains itself.

`summaryPartial`, `successToastPartialDescription` and `errorToastAllFailedDescription` are left alone, since they say to download the report "for details", which already frames it as reference rather than as a re-upload artefact. Say the word if you would rather have that made explicit in the summary step too.

frontend/src/community/leave/utils/bulkAssignPolicyUtils.ts
thusala: Fixed, there is now one definition of header canonicalisation on this path. `bulkAssignPolicyUtils.ts:64`:

```ts
const hasErrorReportColumn = (fields: string[], errorHeader: string): boolean => {
  const errorColumn = normalizeHeader(errorHeader);
  // A blank errorHeader would match the empty trailing header spreadsheets append.
  if (errorColumn === "") {
    return false;
  }
  return fields.some((field) => normalizeHeader(field) === errorColumn);
};
```

The raw `field.trim() !== ""` is gone. The blank check is now `normalizeHeader(errorHeader) === ""` and the equality check is `normalizeHeader(field) === errorColumn`, so both go through `normalizeHeader` and any future change to it, zero-width or non-breaking characters as you mention, applies to both by construction.

frontend/src/community/leave/components/molecules/BulkAssignPolicyModals/BulkAssignPolicyUploadStep.tsx
thusala: Fixed, including the pre-existing `missingColumns` join.

`leavePolicyConstants.ts` now carries a separate display separator next to `CSV_DELIMITER`:

```ts
// Separator for column names listed inside a user-facing sentence. Kept apart from
// CSV_DELIMITER so changing the file format never reformats an error message.
export const COLUMN_LIST_SEPARATOR = ", ";
```

`BulkAssignPolicyUploadStep.tsx` no longer imports `CSV_DELIMITER` at all, and `missingColumnsError` reads `Policy Name, Effective Date` rather than `Policy Name,Effective Date`.

I went with the plain separator over `Intl.ListFormat`: the sentence already ends in "Please use the CSV template", so an "and"-joined list adds locale machinery without changing how the message reads. The `unexpectedColumnsError` join is gone entirely, that message no longer interpolates a column list.

frontend/src/community/leave/components/molecules/BulkAssignPolicyModals/BulkAssignPolicyUploadStep.tsx
thusala: Not done, and worth being explicit about what does and does not cover the gap you describe.

Closed by the `Record` change (see the reply above): a missing enum member is a compile error rather than a blank message, so the silent-dead-end path is gone by construction.

Still uncovered: that `errorReportUploadError` resolves to real copy and reaches `DragAndDropField`'s `customError` and its aria-live region. A typo in the key would still ship as a raw key string. I checked the key by hand against `leaveModule.json:289`, but there is no automated guard and I have not run this in a browser.

I did not add `BulkAssignPolicyUploadStep.test.tsx` in this PR because it would be the first component test under `frontend/src/community/leave/components`: it needs `useBulkAssignLeavePolicies`, `ToastProvider` and a `DragAndDropField` drop simulation mocked before it asserts anything, and that scaffolding is a larger change than the fix this PR is carrying. The `attendance` module's `MockTheme` tests are the pattern to copy.

Happy to do it as a follow-up if you would rather it land with this work than after, say so and I will add it here instead.

shakila-kamalasena (approved):

hasalarootcode (approved):

akilarootcode (approved):

ThinuwanW (approved):

PR #1788, opened 2026-08-18
fix: exclude pending employees from birthday notification modal

SanithuGunasena07:
The Birthday Notification modal was incorrectly displaying birthdays for employees still in Pending status, exposing onboarding-in-progress employees to other users before their accounts were fully active. This happened because the repository query backing the notification (findEmployeeBirthdaysOnByViewerAndScope in EmployeeRepositoryImpl) filtered on both ACTIVE and PENDING account statuses instead of ACTIVE alone. This change restricts the status predicate to ACTIVE only, ensuring pending employees are excluded from birthday notifications until their onboarding is completed, matching the expected behavior.

### DIFF
--- a/backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/EmployeeRepositoryImpl.java
+++ b/backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/EmployeeRepositoryImpl.java
@@ -1681,7 +1681,7 @@ public class EmployeeRepositoryImpl implements EmployeeRepository {
 
 		List<Predicate> predicates = new ArrayList<>();
 		predicates.add(cb.isTrue(userJoin.get(User_.isActive)));
-		predicates.add(cb.equal(root.get(Employee_.accountStatus), AccountStatus.ACTIVE));
+		predicates.add(root.get(Employee_.accountStatus).in(Set.of(AccountStatus.ACTIVE)));
 		predicates.add(PeopleUtil.notGuestEmployeePredicate(cb, roleJoin));
 
 		predicates.add(cb.isNotNull(personalInfoJoin.get(EmployeePersonalInfo_.birthDate)));
--- a/backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/EmployeeRepositoryImpl.java
+++ b/backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/EmployeeRepositoryImpl.java
@@ -1681,7 +1681,7 @@ public class EmployeeRepositoryImpl implements EmployeeRepository {
 
 		List<Predicate> predicates = new ArrayList<>();
 		predicates.add(cb.isTrue(userJoin.get(User_.isActive)));
-		predicates.add(root.get(Employee_.accountStatus).in(Set.of(AccountStatus.ACTIVE)));
+		predicates.add(cb.equal(root.get(Employee_.accountStatus), AccountStatus.ACTIVE));
 		predicates.add(PeopleUtil.notGuestEmployeePredicate(cb, roleJoin));
 
 		predicates.add(cb.isNotNull(personalInfoJoin.get(EmployeePersonalInfo_.birthDate)));
--- a/backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/EmployeeRepositoryImpl.java
+++ b/backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/EmployeeRepositoryImpl.java
@@ -1681,7 +1681,7 @@ public class EmployeeRepositoryImpl implements EmployeeRepository {
 
 		List<Predicate> predicates = new ArrayList<>();
 		predicates.add(cb.isTrue(userJoin.get(User_.isActive)));
-		predicates.add(root.get(Employee_.accountStatus).in(Set.of(AccountStatus.ACTIVE, AccountStatus.PENDING)));
+		predicates.add(cb.equal(root.get(Employee_.accountStatus), AccountStatus.ACTIVE));
 		predicates.add(PeopleUtil.notGuestEmployeePredicate(cb, roleJoin));
 
 		predicates.add(cb.isNotNull(personalInfoJoin.get(EmployeePersonalInfo_.birthDate)));
--- a/backend/src/test/java/com/skapp/community/peopleplanner/controller/v1/PeopleBirthdayNotificationControllerIntegrationTest.java
+++ b/backend/src/test/java/com/skapp/community/peopleplanner/controller/v1/PeopleBirthdayNotificationControllerIntegrationTest.java
@@ -276,7 +276,7 @@ class PeopleBirthdayNotificationControllerIntegrationTest {
 		}
 
 		@Test
-		@DisplayName("Get today's birthdays with ORGANIZATION scope - Returns all matching employees, including pending ones, ordered by name")
+		@DisplayName("Get today's birthdays with ORGANIZATION scope - Returns matching employees ordered by name, excluding pending ones")
 		void getTodayBirthdayNotifications_WithOrganizationScope_ReturnsAllMatchingEmployeesOrderedByName()
 				throws Exception {
 			seedConfig(true, true, false);
@@ -291,10 +291,9 @@ class PeopleBirthdayNotificationControllerIntegrationTest {
 			});
 
 			assertSuccessful(performGetTodayRequest(currentUserToken))
-				.andExpect(jsonPath(BIRTHDAYS_COUNT_PATH).value(3))
-				.andExpect(jsonPath(birthdayFieldPath(0, "employeeId")).value(SECOND_OUTSIDER_EMPLOYEE_ID.intValue()))
-				.andExpect(jsonPath(birthdayFieldPath(1, "employeeId")).value(OUTSIDER_EMPLOYEE_ID.intValue()))
-				.andExpect(jsonPath(birthdayFieldPath(2, "employeeId")).value(CURRENT_EMPLOYEE_ID.intValue()));
+				.andExpect(jsonPath(BIRTHDAYS_COUNT_PATH).value(2))
+				.andExpect(jsonPath(birthdayFieldPath(0, "employeeId")).value(OUTSIDER_EMPLOYEE_ID.intValue()))
+				.andExpect(jsonPath(birthdayFieldPath(1, "employeeId")).value(CURRENT_EMPLOYEE_ID.intValue()));
 		}
 
 		@Test
--- a/backend/src/test/java/com/skapp/community/peopleplanner/controller/v1/PeopleBirthdayNotificationControllerIntegrationTest.java
+++ b/backend/src/test/java/com/skapp/community/peopleplanner/controller/v1/PeopleBirthdayNotificationControllerIntegrationTest.java
@@ -276,7 +276,7 @@ class PeopleBirthdayNotificationControllerIntegrationTest {
 		}
 
 		@Test
-		@DisplayName("Get today's birthdays with ORGANIZATION scope - Returns matching employees ordered by name, excluding pending ones")
+		@DisplayName("Get today's birthdays with ORGANIZATION scope - Returns all matching employees ordered by name")
 		void getTodayBirthdayNotifications_WithOrganizationScope_ReturnsAllMatchingEmployeesOrderedByName()
 				throws Exception {
 			seedConfig(true, true, false);
@@ -285,15 +285,13 @@ class PeopleBirthdayNotificationControllerIntegrationTest {
 			giveBirthdayOn(SECOND_OUTSIDER_EMPLOYEE_ID, today);
 
 			mutateEmployee(OUTSIDER_EMPLOYEE_ID, employee -> employee.setFirstName("alpha"));
-			mutateEmployee(SECOND_OUTSIDER_EMPLOYEE_ID, employee -> {
-				employee.setFirstName("Alpha");
-				employee.setAccountStatus(AccountStatus.PENDING);
-			});
+			mutateEmployee(SECOND_OUTSIDER_EMPLOYEE_ID, employee -> employee.setFirstName("Alpha"));
 
 			assertSuccessful(performGetTodayRequest(currentUserToken))
-				.andExpect(jsonPath(BIRTHDAYS_COUNT_PATH).value(2))
-				.andExpect(jsonPath(birthdayFieldPath(0, "employeeId")).value(OUTSIDER_EMPLOYEE_ID.intValue()))
-				.andExpect(jsonPath(birthdayFieldPath(1, "employeeId")).value(CURRENT_EMPLOYEE_ID.intValue()));
+				.andExpect(jsonPath(BIRTHDAYS_COUNT_PATH).value(3))
+				.andExpect(jsonPath(birthdayFieldPath(0, "employeeId")).value(SECOND_OUTSIDER_EMPLOYEE_ID.intValue()))
+				.andExpect(jsonPath(birthdayFieldPath(1, "employeeId")).value(OUTSIDER_EMPLOYEE_ID.intValue()))
+				.andExpect(jsonPath(birthdayFieldPath(2, "employeeId")).value(CURRENT_EMPLOYEE_ID.intValue()));
 		}
 
 		@Test


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR narrows the birthday-notification query so only ACTIVE employees are returned, excluding PENDING ones, and updates the corresponding ORGANIZATION-scope integration test. The change itself is a one-line predicate narrowing and is behaviorally sound in the common case, but it diverges from the codebase-wide convention that treats ACTIVE+PENDING as the 'live employee' set, and PENDING is not reliably cleared on community credential sign-in — so a real, signed-in employee can remain PENDING indefinitely and silently lose birthday visibility. The test update also folds the new exclusion rule into an ordering-focused test rather than the existing disqualifying-conditions test, and in doing so drops the only coverage of the name tie-break ordering.

Found **9** new issue(s): 🔴 2 important, 🟡 4 suggestion(s), 🟣 3 nit(s)

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/EmployeeRepositoryImpl.java
```diff
@@ -1681,7 +1681,7 @@ public List<Employee> findEmployeeBirthdaysOnByViewerAndScope(LocalDate date, Lo
 
 		List<Predicate> predicates = new ArrayList<>();
 		predicates.add(cb.isTrue(userJoin.get(User_.isActive)));
-		predicates.add(root.get(Employee_.accountStatus).in(Set.of(AccountStatus.ACTIVE, AccountStatus.PENDING)));
+		predicates.add(root.get(Employee_.accountStatus).in(Set.of(AccountStatus.ACTIVE)));
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Excluding PENDING here diverges from how the rest of the codebase defines a 'live' employee, and PENDING is not reliably cleared for credential-based users. Everywhere else the live set is ACTIVE+PENDING: same file lines 1577, 1598 and 1623 (countActiveEmployeesExcludingGuests, work-location counts), AppSetupServiceImpl:31, RolesServiceImpl:292 and :370 (super-admin count), PeopleServiceImpl:1514, and enterprise AiNudgeServiceImpl:48. More importantly, PENDING -> ACTIVE only happens in AuthServiceImpl.employeeResetPassword (line 389) and in the SSO paths (EpAuthServiceImpl:505/:809, EpAuthServiceV2Impl:262/:594). The community credential path AuthServiceImpl.performSignIn (line 166) does NOT promote PENDING -> ACTIVE; it only checks user.getIsActive(). Since employee creation emails a working temp password and leaves accountStatus PENDING (PeopleServiceImpl:609-610 for single create, :2387-2388 for bulk), an employee who signs in with that temp password and never completes the reset stays PENDING forever while being a fully authenticated daily user. Consequences: (a) that person's birthday is never announced to anyone, and (b) under SELF scope the ACTIVE predicate is ANDed with the self predicate, so they get an empty response on their own birthday. Please confirm the intent is 'invited but never onboarded' rather than 'never completed a password reset'. If the intent is the former, the more robust fix is to also promote PENDING -> ACTIVE on successful credential sign-in (or key off user.getIsPasswordChangedForTheFirstTime()). Either way, add a short comment at this predicate explaining why birthday notifications intentionally use a narrower definition of active than the rest of the app, so the next reader does not 'fix' it back to ACTIVE+PENDING.

backend/src/test/java/com/skapp/community/peopleplanner/controller/v1/PeopleBirthdayNotificationControllerIntegrationTest.java
```diff
@@ -291,10 +291,9 @@ void getTodayBirthdayNotifications_WithOrganizationScope_ReturnsAllMatchingEmplo
 			});
 
 			assertSuccessful(performGetTodayRequest(currentUserToken))
-				.andExpect(jsonPath(BIRTHDAYS_COUNT_PATH).value(3))
-				.andExpect(jsonPath(birthdayFieldPath(0, "employeeId")).value(SECOND_OUTSIDER_EMPLOYEE_ID.intValue()))
-				.andExpect(jsonPath(birthdayFieldPath(1, "employeeId")).value(OUTSIDER_EMPLOYEE_ID.intValue()))
-				.andExpect(jsonPath(birthdayFieldPath(2, "employeeId")).value(CURRENT_EMPLOYEE_ID.intValue()));
+				.andExpect(jsonPath(BIRTHDAYS_COUNT_PATH).value(2))
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The new PENDING exclusion is asserted inside an ordering-focused test, and doing so silently removes the only coverage of the name tie-break. Before this change, employee 3 ("alpha") and employee 4 ("Alpha") tied on lower(firstName), so the assertion that 4 came before 3 was what exercised the cb.asc(cb.lower(lastName)) tie-breaker in the orderBy at EmployeeRepositoryImpl:1693-1694. Now employee 4 is filtered out and the two remaining rows differ on firstName, so the lastName tie-break (and the lower() applied to it) is no longer covered by any test. Suggested fix: keep SECOND_OUTSIDER_EMPLOYEE_ID ACTIVE with firstName "Alpha" in this test so the 3-element ordering assertion is restored, and move the setAccountStatus(AccountStatus.PENDING) mutation into getTodayBirthdayNotifications_WithDisqualifyingConditions_ExcludesAllOfThem (line 341), which is the purpose-built test that already enumerates every other exclusion rule (deactivated user, TERMINATED, PM guest, null birth date) and currently has no PENDING case. That keeps each test asserting one concern and gives the new predicate a home alongside its siblings.

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/EmployeeRepositoryImpl.java
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

A single-valued IN is better expressed as an equality. Replace `predicates.add(root.get(Employee_.accountStatus).in(Set.of(AccountStatus.ACTIVE)));` with `predicates.add(cb.equal(root.get(Employee_.accountStatus), AccountStatus.ACTIVE));`. This is already the established idiom in this same file — see findAllActiveEmployeesExcludingRole at line 1642 (`cb.equal(root.get(Employee_.ACCOUNT_STATUS), AccountStatus.ACTIVE)`) and line 456 (`criteriaBuilder.equal(root.get(Employee_.ACCOUNT_STATUS), AccountStatus.PENDING)`). It also produces a plain `=` in the generated SQL instead of a one-element `IN (...)`, and makes it obvious at a glance that the set is intentionally a single status rather than a set someone forgot to finish editing.

backend/src/test/java/com/skapp/community/peopleplanner/controller/v1/PeopleBirthdayNotificationControllerIntegrationTest.java
```diff
@@ -276,7 +276,7 @@ void getTodayBirthdayNotifications_WithTeamScope_ReturnsOnlyTeammates() throws E
 		}
 
 		@Test
-		@DisplayName("Get today's birthdays with ORGANIZATION scope - Returns all matching employees, including pending ones, ordered by name")
+		@DisplayName("Get today's birthdays with ORGANIZATION scope - Returns matching employees ordered by name, excluding pending ones")
 		void getTodayBirthdayNotifications_WithOrganizationScope_ReturnsAllMatchingEmployeesOrderedByName()
```
ThinuwanW, line 280: 🤖 **Claude** · 🟣 **Nit**

The @DisplayName was updated to say pending employees are excluded, but the method name still reads getTodayBirthdayNotifications_WithOrganizationScope_ReturnsAllMatchingEmployeesOrderedByName, which claims it returns "all matching" employees. Since the surrounding tests follow a consistent method-name-mirrors-DisplayName style (e.g. _WithDisqualifyingConditions_ExcludesAllOfThem, _WhenAlreadyViewedToday_ReturnsEmptyList), rename it to match — for example getTodayBirthdayNotifications_WithOrganizationScope_ExcludesPendingAndOrdersByName. If you take the restructuring suggested above and move the PENDING case to the disqualifying-conditions test, the original method name becomes accurate again and no rename is needed.

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/EmployeeRepositoryImpl.java
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The account-status predicate is AND-ed with the scope predicate, so it now filters the viewer's OWN row as well. `buildBirthdayScopePredicate` builds a `selfPredicate` (`employeeId = viewerEmployeeId`) that is the entire predicate for `BirthdayNotificationScope.SELF` and one arm of the OR for `TEAM` — that guarantee is silently broken by this change. Concrete failure: an employee created by an admin starts at PENDING (PeopleServiceImpl lines 254/308/524) and is only flipped to ACTIVE by `AuthServiceImpl.employeeResetPassword`; sign-in itself only blocks TERMINATED/DELETED, so such a user is fully logged in. On their birthday, with the default SELF scope (neither org-wide nor team-wide configured), `getTodayBirthdayNotifications` now returns an empty list — the user cannot see their own birthday in the one scope that exists solely to show it to them. If the intent is 'don't broadcast not-yet-onboarded people to others', the viewer's own row should be exempt, e.g. build the eligibility predicate as `cb.or(cb.equal(root.get(Employee_.employeeId), viewerEmployeeId), statusPredicate)` or apply the status filter only when the scope is TEAM/ORGANIZATION. There is also no test covering a PENDING viewer or PENDING exclusion under SELF/TEAM scope — the only new assertion is on ORGANIZATION scope.

ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The eligibility rule is inlined as an anonymous `Set.of(...)` literal with no name or comment explaining why this query's definition of 'eligible employee' differs from its neighbours. Two lines below, the guest rule for this exact query is already centralized as `PeopleUtil.notGuestEmployeePredicate(cb, roleJoin)` — the status rule should follow the same pattern. Within this one file there are now three different notions of 'active': `findActiveEmployeesExcludingGuests` (line 1598) and `countActiveEmployeesExcludingGuests` (line 1623) use ACTIVE+PENDING, `findAllActiveEmployeesExcludingRole` (line 1642) uses ACTIVE only, and this method now uses ACTIVE only. Extract a named helper (e.g. `PeopleUtil.notificationEligibleStatusPredicate(cb, root)`) or at minimum a named constant such as `BIRTHDAY_NOTIFICATION_ELIGIBLE_STATUSES`, so the deliberate divergence is documented at the definition site rather than having to be inferred from a diff.

backend/src/test/java/com/skapp/community/peopleplanner/controller/v1/PeopleBirthdayNotificationControllerIntegrationTest.java
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The new exclusion is asserted only indirectly, via `BIRTHDAYS_COUNT_PATH == 2` plus positional employeeId checks. If the ordering predicate ever changes so the pending employee lands at index 0 while some other employee drops out, the count and both index assertions could still pass. Add an explicit absence assertion for the pending employee, e.g. `.andExpect(jsonPath(BIRTHDAYS_PATH + "[*]['employeeId']", not(hasItem(SECOND_OUTSIDER_EMPLOYEE_ID.intValue()))))` (org.hamcrest.Matchers.not / hasItem), so the test names the behaviour it is protecting instead of inferring it from a length.

backend/src/test/java/com/skapp/community/peopleplanner/controller/v1/PeopleBirthdayNotificationControllerIntegrationTest.java
```diff
@@ -291,10 +291,9 @@ void getTodayBirthdayNotifications_WithOrganizationScope_ReturnsAllMatchingEmplo
 			});
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Dead test setup: the `setFirstName("Alpha")` call on SECOND_OUTSIDER_EMPLOYEE_ID now has no observable effect, because the same mutation sets the account status to PENDING and the employee is filtered out of the result entirely. The paired `setFirstName("alpha")` on OUTSIDER_EMPLOYEE_ID (line 287) existed only to form an 'Alpha'/'alpha' pair for the case-insensitive `cb.lower(firstName)` ordering, so it is now just an arbitrary rename. Either drop the `setFirstName("Alpha")` line so the setup states only what it tests, or move the PENDING flag to a different employee and keep the name pair intact.

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/EmployeeRepositoryImpl.java
```diff
@@ -1681,7 +1681,7 @@ public List<Employee> findEmployeeBirthdaysOnByViewerAndScope(LocalDate date, Lo
 
 		List<Predicate> predicates = new ArrayList<>();
 		predicates.add(cb.isTrue(userJoin.get(User_.isActive)));
```
ThinuwanW, line 1683: 🤖 **Claude** · 🟣 **Nit**

Pre-existing issue in the method being modified: lines 1674-1676 call `root.fetch(...)` for personalInfo, user and employeeRole, and lines 1678-1680 then call `root.join(...)` for the same three associations. In JPA Criteria these are distinct join nodes, so Hibernate emits six SQL joins instead of three for every invocation of this query. Since the predicate on line 1683 (and 1685, 1687-1688) needs handles to those associations, the idiomatic fix is to reuse the fetch nodes, e.g. `Join<Employee, User> userJoin = (Join<Employee, User>) root.fetch(Employee_.user);` rather than declaring separate joins. All three associations are `@OneToOne`, so this is a cost/plan concern rather than a row-duplication bug.

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/EmployeeRepositoryImpl.java
SanithuGunasena07: intentionally left out for pending as per requirement

SanithuGunasena07: now uses cb.equal(root.get(Employee_.accountStatus), AccountStatus.ACTIVE), matching the existing idiom in this file.

backend/src/test/java/com/skapp/community/peopleplanner/controller/v1/PeopleBirthdayNotificationControllerIntegrationTest.java
SanithuGunasena07: reverted that test back to its original DisplayName

backend/src/test/java/com/skapp/community/peopleplanner/controller/v1/PeopleBirthdayNotificationControllerIntegrationTest.java
SanithuGunasena07: All three employees in that test are ACTIVE (no PENDING mutation), so the fix won't exclude any of them test should pass.

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/EmployeeRepositoryImpl.java
SanithuGunasena07: per the requirement, only ACTIVE employees should ever appear in birthday notifications, including for their own SELF-scope view.

SanithuGunasena07: it's a pre-existing inconsistency across the file's various "active" definitions, not something introduced by this change.

backend/src/test/java/com/skapp/community/peopleplanner/controller/v1/PeopleBirthdayNotificationControllerIntegrationTest.java
SanithuGunasena07: test was reverted back to the original 3-employee tie-break assertion with no PENDING mutation at all

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/EmployeeRepositoryImpl.java
SanithuGunasena07: not code touched by this pr

backend/src/test/java/com/skapp/community/peopleplanner/controller/v1/PeopleBirthdayNotificationControllerIntegrationTest.java
SanithuGunasena07: not code touched by this pr

akilarootcode (approved):

PR #1790, opened 2026-08-18
feat(crm): add batch companies-by-ids endpoint for store hydration

anusham-rc:
## Summary
Adds a batch `POST /v2/crm/company/by-ids` endpoint that returns the base details of companies matching a list of ids, used to hydrate the client's company store. Soft-deleted companies are omitted.

Split out from the v2 scalar-keys refactor branch into its own PR to ease review and merge independently.

## Changes
- `CrmCompanyControllerV2` — new `getCompaniesByIds` endpoint (`POST /by-ids`), restricted to `ROLE_CRM_SALES_REPRESENTATIVE`
- `CrmCompanyBatchRequestDto` — request payload carrying the id list
- `CrmCompanyDao` / `CrmCompanyServiceV2` / `CrmCompanyServiceImplV2` — batch fetch wiring, soft-deleted rows excluded
- `CrmCompanyControllerV2IntegrationTest` — integration coverage

## Testing
- `mvn -Dtest=CrmCompanyControllerV2IntegrationTest test` → **10 passed, 0 failures, 0 errors**

### DIFF
--- a/backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmCompanyController.java
+++ b/backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmCompanyController.java
@@ -16,6 +16,7 @@ import org.springframework.web.bind.annotation.RequestParam;
 import org.springframework.web.bind.annotation.RestController;
 
 import com.skapp.community.common.payload.response.ResponseEntityDto;
+import com.skapp.community.crmplanner.payload.request.CrmCompanyBatchRequestDto;
 import com.skapp.community.crmplanner.payload.request.CrmCompanyCreateDto;
 import com.skapp.community.crmplanner.payload.request.CrmCompanyDomainSearchRequestDto;
 import com.skapp.community.crmplanner.payload.request.CrmCompanyEditDto;
@@ -94,6 +95,16 @@ public class CrmCompanyController {
 	@PreAuthorize("hasAnyRole('ROLE_CRM_SALES_REPRESENTATIVE')")
 	public ResponseEntity<ResponseEntityDto> getCompanyById(@PathVariable Long id) {
 		ResponseEntityDto responseDto = companyService.getCompanyById(id);
+		return new ResponseEntity<>(responseDto, HttpStatus.OK);
+	}
+
+	@Operation(summary = "Get companies by ids",
+			description = "Returns the base details of the companies matching the given ids, "
+					+ "used to hydrate the client's company store. Soft-deleted companies are omitted.")
+	@PostMapping("/batch")
+	@PreAuthorize("hasAnyRole('ROLE_CRM_SALES_REPRESENTATIVE')")
+	public ResponseEntity<ResponseEntityDto> getCompaniesByIds(@RequestBody CrmCompanyBatchRequestDto requestDto) {
+		ResponseEntityDto responseDto = companyService.getCompaniesByIds(requestDto);
 		return new ResponseEntity<>(responseDto, HttpStatus.OK);
 	}
 
--- a/backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmCompanyController.java
+++ b/backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmCompanyController.java
@@ -16,7 +16,7 @@ import org.springframework.web.bind.annotation.RequestParam;
 import org.springframework.web.bind.annotation.RestController;
 
 import com.skapp.community.common.payload.response.ResponseEntityDto;
-import com.skapp.community.crmplanner.payload.request.CrmCompanyBatchRequestDto;
+import com.skapp.community.crmplanner.payload.request.CrmCompanyIdsRequestDto;
 import com.skapp.community.crmplanner.payload.request.CrmCompanyCreateDto;
 import com.skapp.community.crmplanner.payload.request.CrmCompanyDomainSearchRequestDto;
 import com.skapp.community.crmplanner.payload.request.CrmCompanyEditDto;
@@ -101,9 +101,9 @@ public class CrmCompanyController {
 	@Operation(summary = "Get companies by ids",
 			description = "Returns the base details of the companies matching the given ids, "
 					+ "used to hydrate the client's company store. Soft-deleted companies are omitted.")
-	@PostMapping("/batch")
+	@PostMapping("/by-ids")
 	@PreAuthorize("hasAnyRole('ROLE_CRM_SALES_REPRESENTATIVE')")
-	public ResponseEntity<ResponseEntityDto> getCompaniesByIds(@RequestBody CrmCompanyBatchRequestDto requestDto) {
+	public ResponseEntity<ResponseEntityDto> getCompaniesByIds(@RequestBody CrmCompanyIdsRequestDto requestDto) {
 		ResponseEntityDto responseDto = companyService.getCompaniesByIds(requestDto);
 		return new ResponseEntity<>(responseDto, HttpStatus.OK);
 	}
--- a/backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmCompanyController.java
+++ b/backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmCompanyController.java
@@ -101,7 +101,7 @@ public class CrmCompanyController {
 	@Operation(summary = "Get companies by ids",
 			description = "Returns the base details of the companies matching the given ids, "
 					+ "used to hydrate the client's company store. Soft-deleted companies are omitted.")
-	@PostMapping("/by-ids")
+	@PostMapping("/ids")
 	@PreAuthorize("hasAnyRole('ROLE_CRM_SALES_REPRESENTATIVE')")
 	public ResponseEntity<ResponseEntityDto> getCompaniesByIds(@RequestBody CrmCompanyIdsRequestDto requestDto) {
 		ResponseEntityDto responseDto = companyService.getCompaniesByIds(requestDto);
--- a/backend/src/main/java/com/skapp/community/crmplanner/payload/request/CrmCompanyBatchRequestDto.java
+++ b/backend/src/main/java/com/skapp/community/crmplanner/payload/request/CrmCompanyBatchRequestDto.java
@@ -1,0 +1,14 @@
+package com.skapp.community.crmplanner.payload.request;
+
+import lombok.Getter;
+import lombok.Setter;
+
+import java.util.List;
+
+@Getter
+@Setter
+public class CrmCompanyBatchRequestDto {
+
+	private List<Long> ids;
+
+}
--- a/backend/src/main/java/com/skapp/community/crmplanner/payload/request/CrmCompanyIdsRequestDto.java
+++ b/backend/src/main/java/com/skapp/community/crmplanner/payload/request/CrmCompanyIdsRequestDto.java
@@ -7,7 +7,7 @@ import java.util.List;
 
 @Getter
 @Setter
-public class CrmCompanyBatchRequestDto {
+public class CrmCompanyIdsRequestDto {
 
 	private List<Long> ids;
 
--- a/backend/src/main/java/com/skapp/community/crmplanner/repository/CrmCompanyDao.java
+++ b/backend/src/main/java/com/skapp/community/crmplanner/repository/CrmCompanyDao.java
@@ -4,12 +4,15 @@ import com.skapp.community.crmplanner.model.CrmCompany;
 import org.springframework.data.jpa.repository.JpaRepository;
 import org.springframework.stereotype.Repository;
 
+import java.util.List;
 import java.util.Optional;
 
 @Repository
 public interface CrmCompanyDao extends JpaRepository<CrmCompany, Long>, CrmCompanyRepository {
 
 	Optional<CrmCompany> findByIdAndIsDeletedFalse(Long id);
+
+	List<CrmCompany> findByIdInAndIsDeletedFalse(List<Long> ids);
 
 	boolean existsByNameIgnoreCaseAndIsDeletedFalse(String name);
 
--- a/backend/src/main/java/com/skapp/community/crmplanner/repository/CrmCompanyDao.java
+++ b/backend/src/main/java/com/skapp/community/crmplanner/repository/CrmCompanyDao.java
@@ -12,7 +12,7 @@ public interface CrmCompanyDao extends JpaRepository<CrmCompany, Long>, CrmCompa
 
 	Optional<CrmCompany> findByIdAndIsDeletedFalse(Long id);
 
-	List<CrmCompany> findByIdInAndIsDeletedFalse(List<Long> ids);
+	List<CrmCompany> findByIdInAndIsDeletedFalseOrderByIdAsc(List<Long> ids);
 
 	boolean existsByNameIgnoreCaseAndIsDeletedFalse(String name);
 
--- a/backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmCompanyServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmCompanyServiceImpl.java
@@ -12,7 +12,7 @@ import com.skapp.community.common.util.MessageUtil;
 import com.skapp.community.crmplanner.constant.CrmMessageConstant;
 import com.skapp.community.crmplanner.mapper.CrmMapper;
 import com.skapp.community.crmplanner.model.CrmCompany;
-import com.skapp.community.crmplanner.payload.request.CrmCompanyBatchRequestDto;
+import com.skapp.community.crmplanner.payload.request.CrmCompanyIdsRequestDto;
 import com.skapp.community.crmplanner.payload.request.CrmCompanyCreateDto;
 import com.skapp.community.crmplanner.payload.request.CrmCompanyDomainSearchRequestDto;
 import com.skapp.community.crmplanner.payload.request.CrmCompanyEditDto;
@@ -145,7 +145,7 @@ public class CrmCompanyServiceImpl implements CrmCompanyService {
 
 	@Override
 	@Transactional(readOnly = true)
-	public ResponseEntityDto getCompaniesByIds(CrmCompanyBatchRequestDto requestDto) {
+	public ResponseEntityDto getCompaniesByIds(CrmCompanyIdsRequestDto requestDto) {
 		log.info("getCompaniesByIds: execution started");
 
 		if (requestDto.getIds() == null || requestDto.getIds().isEmpty()) {
@@ -153,10 +153,10 @@ public class CrmCompanyServiceImpl implements CrmCompanyService {
 			return new ResponseEntityDto(false, Collections.emptyList());
 		}
 
-		List<CrmCompanyResponseDto> companies = crmCompanyDao.findByIdInAndIsDeletedFalse(requestDto.getIds())
-			.stream()
-			.map(crmCompanyMapper::crmCompanyToCrmCompanyResponseDto)
-			.toList();
+		CrmValidations.validateCompanyIds(requestDto.getIds());
+
+		List<CrmCompanyResponseDto> companies = crmCompanyMapper.crmCompaniesToCrmCompanyResponseDtos(
+				crmCompanyDao.findByIdInAndIsDeletedFalseOrderByIdAsc(requestDto.getIds()));
 
 		log.info("getCompaniesByIds: execution ended with {} result(s)", companies.size());
 		return new ResponseEntityDto(false, companies);
@@ -184,9 +184,7 @@ public class CrmCompanyServiceImpl implements CrmCompanyService {
 		List<CrmCompany> companies = crmCompanyDao.findCompaniesByWebsiteDomain(requestDto.getDomain(),
 				requestDto.getLimit());
 
-		List<CrmCompanyResponseDto> companyDtos = companies.stream()
-			.map(crmCompanyMapper::crmCompanyToCrmCompanyResponseDto)
-			.toList();
+		List<CrmCompanyResponseDto> companyDtos = crmCompanyMapper.crmCompaniesToCrmCompanyResponseDtos(companies);
 
 		CrmCompanyDomainSearchResponseDto responseDto = new CrmCompanyDomainSearchResponseDto();
 		responseDto.setCompanies(companyDtos);
--- a/backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmCompanyServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmCompanyServiceImpl.java
@@ -12,6 +12,7 @@ import com.skapp.community.common.util.MessageUtil;
 import com.skapp.community.crmplanner.constant.CrmMessageConstant;
 import com.skapp.community.crmplanner.mapper.CrmMapper;
 import com.skapp.community.crmplanner.model.CrmCompany;
+import com.skapp.community.crmplanner.payload.request.CrmCompanyBatchRequestDto;
 import com.skapp.community.crmplanner.payload.request.CrmCompanyCreateDto;
 import com.skapp.community.crmplanner.payload.request.CrmCompanyDomainSearchRequestDto;
 import com.skapp.community.crmplanner.payload.request.CrmCompanyEditDto;
@@ -30,6 +31,7 @@ import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.data.domain.PageRequest;
 
+import java.util.Collections;
 import java.util.List;
 
 @Service
@@ -139,6 +141,25 @@ public class CrmCompanyServiceImpl implements CrmCompanyService {
 
 		log.info("getCompanyById: execution ended");
 		return new ResponseEntityDto(false, response);
+	}
+
+	@Override
+	@Transactional(readOnly = true)
+	public ResponseEntityDto getCompaniesByIds(CrmCompanyBatchRequestDto requestDto) {
+		log.info("getCompaniesByIds: execution started");
+
+		if (requestDto.getIds() == null || requestDto.getIds().isEmpty()) {
+			log.info("getCompaniesByIds: no ids provided, returning empty list");
+			return new ResponseEntityDto(false, Collections.emptyList());
+		}
+
+		List<CrmCompanyResponseDto> companies = crmCompanyDao.findByIdInAndIsDeletedFalse(requestDto.getIds())
+			.stream()
+			.map(crmCompanyMapper::crmCompanyToCrmCompanyResponseDto)
+			.toList();
+
+		log.info("getCompaniesByIds: execution ended with {} result(s)", companies.size());
+		return new ResponseEntityDto(false, companies);
 	}
 
 	@Override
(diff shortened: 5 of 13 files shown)


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds a `POST /v1/crm/company/batch` endpoint that returns base company details for a list of ids, backed by a new derived Dao query, request DTO, service method and five integration tests. Layering, `@Transactional(readOnly = true)`, entry/exit logging, MapStruct mapping, DTO usage and the `@PreAuthorize` role (consistent with `getCompanyById`, and CRM_ADMIN/SALES_MANAGER inherit the representative authority) all follow the module conventions, and the v2 endpoint was fully removed with no leftovers. The main gap is that the `ids` list is completely unvalidated and unbounded, so a single authenticated request can drive an arbitrarily large `IN (...)` query.

Found **13** new issue(s): 🔴 1 important, 🟡 4 suggestion(s), 🟣 8 nit(s)

backend/src/main/java/com/skapp/community/crmplanner/payload/request/CrmCompanyIdsRequestDto.java
```diff
@@ -0,0 +1,14 @@
...
+import lombok.Getter;
+import lombok.Setter;
+
+import java.util.List;
+
+@Getter
+@Setter
+public class CrmCompanyBatchRequestDto {
+
+	private List<Long> ids;
```
ThinuwanW, line 12: 🤖 **Claude** · 🔴 **Important**

The `ids` list has no bean-validation annotations and the controller takes the body without `@Valid`, so the batch size is unbounded. No global JSON request-body size limit is configured (only `spring.servlet.multipart.max-file-size` in application-*.yml, which does not apply to JSON), so an authenticated CRM user can POST tens of thousands of ids and force a single huge `IN (...)` query — slow query, large persistence-context/heap usage, and potentially a packet-size failure surfacing as a 500 via the `DataAccessException` handler. `@Valid` is used in 190+ places elsewhere in the codebase, so omitting it here is also a convention deviation. Fix: annotate the field (e.g. `@Size(max = CrmConstants.COMPANY_BATCH_MAX_IDS) private List<@NotNull Long> ids;` with a new constant in `CrmConstants`) and add `@Valid` to the `@RequestBody` parameter in `CrmCompanyController.getCompaniesByIds`, or alternatively add a `CrmValidations.validateCompanyIds(...)` check throwing `ModuleException` with a new `CrmMessageConstant` key (matching how the rest of this module validates). Keep the current empty-list-returns-empty behaviour (do not add `@NotEmpty`, the tests rely on it). Consider also de-duplicating the ids (e.g. `new LinkedHashSet<>(ids)` before the query) so a client repeating the same id does not inflate the query.

backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmCompanyController.java
```diff
@@ -97,6 +98,16 @@ public ResponseEntity<ResponseEntityDto> getCompanyById(@PathVariable Long id) {
 		return new ResponseEntity<>(responseDto, HttpStatus.OK);
 	}
 
+	@Operation(summary = "Get companies by ids",
+			description = "Returns the base details of the companies matching the given ids, "
+					+ "used to hydrate the client's company store. Soft-deleted companies are omitted.")
+	@PostMapping("/batch")
+	@PreAuthorize("hasAnyRole('ROLE_CRM_SALES_REPRESENTATIVE')")
+	public ResponseEntity<ResponseEntityDto> getCompaniesByIds(@RequestBody CrmCompanyBatchRequestDto requestDto) {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

A malformed body on this new endpoint returns 500, not 400. `GlobalExceptionHandler` has no `@ExceptionHandler(HttpMessageNotReadableException.class)` and its catch-all `@ExceptionHandler(Exception.class)` maps to `INTERNAL_SERVER_ERROR`, so a payload such as `{"ids":["abc"]}` (or a truncated/absent body) fails Jackson deserialization of `List<Long>` and the client gets an opaque 500 instead of a validation error. This is pre-existing infrastructure behaviour that also affects `createCompany`/`editCompany`, but this endpoint newly exposes a typed numeric collection that clients are likely to get wrong. Suggest adding a `HttpMessageNotReadableException` handler to `GlobalExceptionHandler` returning `HttpStatus.BAD_REQUEST` with the standard `ErrorResponse` shape.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmCompanyServiceImpl.java
```diff
@@ -141,6 +143,25 @@ public ResponseEntityDto getCompanyById(Long id) {
 		return new ResponseEntityDto(false, response);
 	}
 
+	@Override
+	@Transactional(readOnly = true)
+	public ResponseEntityDto getCompaniesByIds(CrmCompanyBatchRequestDto requestDto) {
+		log.info("getCompaniesByIds: execution started");
+
+		if (requestDto.getIds() == null || requestDto.getIds().isEmpty()) {
```
ThinuwanW, line 151: 🤖 **Claude** · 🟣 **Nit**

The `requestDto.getIds() == null` branch is not covered by any test — only the empty-list case (`getCompaniesByIds_EmptyIds_ReturnsEmptyList`) is. Since a client omitting the `ids` field entirely (`{}`) hits a different code path than `{"ids":[]}`, add a test that posts a body with a null/absent `ids` and asserts 200 with an empty `results` array, so the intended silent-empty behaviour is locked in.

backend/src/test/java/com/skapp/community/crmplanner/controller/v1/CrmCompanyControllerIntegrationTest.java
```diff
@@ -844,4 +847,84 @@ void getCompanyMetricsById_WithoutCrmRole_ReturnsForbidden() throws Exception {
...
+	@DisplayName("Get companies by ids - Excludes soft-deleted companies")
+	void getCompaniesByIds_SoftDeleted_Excluded() throws Exception {
+		CrmCompany company = savedBatchCompany("BatchDeletedCo");
+		company.setIsDeleted(true);
+		crmCompanyDao.save(company);
+
+		performBatchRequest(List.of(company.getId())).andDo(print())
+			.andExpect(status().isOk())
+			.andExpect(jsonPath(STATUS_PATH).value(STATUS_SUCCESSFUL))
+			.andExpect(jsonPath("['results']").isEmpty());
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`jsonPath("['results']")` is hardcoded here and at line 919 while `TestConstants.RESULTS_PATH` already exists and `STATUS_PATH`/`RESULTS_0_PATH` are used from the constants class in the same tests. Use `RESULTS_PATH` for consistency (the `"['results'].length()"` expressions can be built as `RESULTS_PATH + ".length()"`).

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmCompanyServiceImpl.java
```diff
@@ -141,6 +143,25 @@ public ResponseEntityDto getCompanyById(Long id) {
...
+			log.info("getCompaniesByIds: no ids provided, returning empty list");
+			return new ResponseEntityDto(false, Collections.emptyList());
+		}
+
+		List<CrmCompanyResponseDto> companies = crmCompanyDao.findByIdInAndIsDeletedFalse(requestDto.getIds())
+			.stream()
+			.map(crmCompanyMapper::crmCompanyToCrmCompanyResponseDto)
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The `.stream().map(crmCompanyMapper::crmCompanyToCrmCompanyResponseDto).toList()` block is a verbatim copy of the one in `searchCompaniesByDomain` (lines 187-189), and it bypasses the mapper convention this module already follows: `CrmMapper` declares list-mapping methods for the other aggregates (`crmDealsToCrmDealResponseDtos`, `crmDealStagesToCrmDealStageResponseDtos`, `crmTasksToCrmTaskResponseDtos`) and `CrmDealStageServiceImpl.getDealStages` — the closest analogue to this new method — calls `crmMapper.crmDealStagesToCrmDealStageResponseDtos(...)` directly. Add `List<CrmCompanyResponseDto> crmCompaniesToCrmCompanyResponseDtos(List<CrmCompany> companies);` to `CrmMapper` and use it here and in `searchCompaniesByDomain`, removing the duplicated stream.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmCompanyServiceImpl.java
ThinuwanW, line 151: 🤖 **Claude** · 🟡 **Suggestion**

Returning a 200 `successful` empty payload when `ids` is null means a client that omits a required field gets a silent success instead of an error, and it contradicts this module's own precedent for list-shaped requests: `CrmValidations.validateDealStageReorderRequest` throws `ModuleException(CRM_ERROR_DEAL_STAGE_REORDER_INVALID_REQUEST)` for a null-or-empty list. It also collides with the fix proposed for the missing bean validation: once `@NotEmpty`/`@Valid` is added to the DTO, this whole branch becomes unreachable dead code. Pick one strategy explicitly — either (a) validate at the boundary (`@NotEmpty` + `@Valid`) and delete lines 151-154, or (b) keep the service-layer guard but throw a `ModuleException` with a message constant for null while still short-circuiting on an explicitly empty list.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmCompanyServiceImpl.java
```diff
@@ -141,6 +143,25 @@ public ResponseEntityDto getCompanyById(Long id) {
...
+	@Transactional(readOnly = true)
+	public ResponseEntityDto getCompaniesByIds(CrmCompanyBatchRequestDto requestDto) {
+		log.info("getCompaniesByIds: execution started");
+
+		if (requestDto.getIds() == null || requestDto.getIds().isEmpty()) {
+			log.info("getCompaniesByIds: no ids provided, returning empty list");
+			return new ResponseEntityDto(false, Collections.emptyList());
+		}
+
+		List<CrmCompanyResponseDto> companies = crmCompanyDao.findByIdInAndIsDeletedFalse(requestDto.getIds())
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The request ids are passed straight into the derived `IN` query with no sanitization, and this is the only company operation that uses neither bean validation nor `CrmValidations` (compare `deleteCompany`/`editCompany`, which both call `CrmValidations.validateCompanyId`, and `searchCompaniesByDomain`, which calls `validateDomain`). Concrete consequences: `{"ids":[1,null]}` binds a null into the IN list (wasted parameter, and driver/Hibernate-version dependent behaviour); duplicate ids make the response length silently smaller than the requested length, which is confusing for a store-hydration caller; and every distinct list size compiles a distinct SQL string, churning the statement/plan cache. Add a `CrmValidations.validateCompanyIds(List<Long> ids)` helper next to `validateCompanyId` that rejects null/non-positive elements, and dedupe before querying (e.g. `new LinkedHashSet<>(requestDto.getIds())`).

backend/src/main/java/com/skapp/community/crmplanner/repository/CrmCompanyDao.java
```diff
@@ -4,13 +4,16 @@
...
 
+import java.util.List;
 import java.util.Optional;
 
 @Repository
 public interface CrmCompanyDao extends JpaRepository<CrmCompany, Long>, CrmCompanyRepository {
 
 	Optional<CrmCompany> findByIdAndIsDeletedFalse(Long id);
 
+	List<CrmCompany> findByIdInAndIsDeletedFalse(List<Long> ids);
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Two small signature issues on the new derived query. (1) Declare the parameter as `Collection<Long>` rather than `List<Long>` so the service can pass a deduplicated `Set` without converting back to a list. (2) There is no ordering, so the row order for an `IN` query is whatever MySQL returns — it neither matches the order of the requested ids nor is guaranteed stable between calls. The analogous list query in this module is explicitly ordered (`CrmDealStageDao.findAllByIsDeletedFalseOrderByOrderIndexAsc`); consider `findByIdInAndIsDeletedFalseOrderByIdAsc` for a deterministic response, or state in the `@Operation` description that the order is unspecified so clients key by id instead of position.

backend/src/main/java/com/skapp/community/crmplanner/payload/request/CrmCompanyBatchRequestDto.java
```diff
@@ -0,0 +1,14 @@
+package com.skapp.community.crmplanner.payload.request;
+
+import lombok.Getter;
+import lombok.Setter;
+
+import java.util.List;
+
+@Getter
+@Setter
+public class CrmCompanyBatchRequestDto {
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

This DTO is field-for-field identical to the existing `com.skapp.community.common.payload.request.ReInvitationRequestDto` (same `private List<Long> ids;`, same `@Getter`/`@Setter`), so the codebase now has two copies of the same shape and will grow a third the moment contacts or deals need the same batch lookup. Also, `Batch` names the transport mechanism rather than the operation, which breaks from the module's descriptive naming (`CrmCompanyDomainSearchRequestDto`, `CrmCompanyMetricRequestDto`). Prefer either a single shared `IdListRequestDto` in `common/payload/request` reused by both call sites, or at minimum rename this to `CrmCompanyIdsRequestDto` to match the `getCompaniesByIds` operation.

backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmCompanyController.java
```diff
@@ -97,6 +98,16 @@ public ResponseEntity<ResponseEntityDto> getCompanyById(@PathVariable Long id) {
 		return new ResponseEntity<>(responseDto, HttpStatus.OK);
 	}
 
+	@Operation(summary = "Get companies by ids",
+			description = "Returns the base details of the companies matching the given ids, "
+					+ "used to hydrate the client's company store. Soft-deleted companies are omitted.")
+	@PostMapping("/batch")
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

A pure read is being exposed as `POST /batch`. Every other read on this controller is a GET with a descriptive kebab-case path (`/lookup`, `/exists`, `/search-by-domain`, `/{id}/metrics`), and POST-for-read gives up HTTP caching/ETags and shows up as a mutation in access-log and audit tooling. If the id list is bounded (which it should be per the validation fix), `GET /by-ids?ids=1,2,3` fits the existing conventions better; if POST is deliberate to avoid URL-length limits, at least rename the path to `/by-ids` so it matches the method name `getCompaniesByIds` and says what it returns rather than how it is transported.

backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmCompanyController.java
```diff
@@ -97,6 +98,16 @@ public ResponseEntity<ResponseEntityDto> getCompanyById(@PathVariable Long id) {
 		return new ResponseEntity<>(responseDto, HttpStatus.OK);
 	}
 
+	@Operation(summary = "Get companies by ids",
```
ThinuwanW, line 101: 🤖 **Claude** · 🟣 **Nit**

The `@Operation` description states this exists to "hydrate the client's company store", but nothing in `frontend/` references this route (no `crm/company/batch` endpoint constant, no query hook, and the `crm/v2` `companies: CrmCompanyRecord` slice has no fetcher). Confirm the consuming frontend change is landing in a follow-up rather than being forgotten — otherwise this ships as unused, untestable-in-practice API surface. Separately, consider rewording the description in terms of the API contract ("returns base details for the given company ids; unknown and soft-deleted ids are omitted") instead of naming a specific client's internal store, since Swagger is the public contract.

backend/src/test/java/com/skapp/community/crmplanner/controller/v1/CrmCompanyControllerIntegrationTest.java
```diff
@@ -844,4 +847,84 @@ void getCompanyMetricsById_WithoutCrmRole_ReturnsForbidden() throws Exception {
...
+
+	private ResultActions performBatchRequest(List<Long> ids) throws Exception {
+		CrmCompanyBatchRequestDto requestDto = new CrmCompanyBatchRequestDto();
+		requestDto.setIds(ids);
+		return performRequest(post(BATCH_PATH).contentType(MediaType.APPLICATION_JSON)
+			.content(objectMapper.writeValueAsString(requestDto))
+			.accept(MediaType.APPLICATION_JSON));
+	}
+
+	private CrmCompany savedBatchCompany(String name) {
```
ThinuwanW, line 862: 🤖 **Claude** · 🟣 **Nit**

`savedBatchCompany` is the third near-identical company factory in this one test class — `createMetricsCompany` (line 630) and the inline block in `getCompanyById_HappyPath_ReturnsCompany` (lines 770-776) build the same entity with the same industry. Consolidate into a single parameterized helper (e.g. `createCompany(String name, String website, String address, String contactNumber)`) and have the existing helpers delegate to it, so future field additions to `CrmCompany` only need one fixture update.

backend/src/test/java/com/skapp/community/crmplanner/controller/v1/CrmCompanyControllerIntegrationTest.java
```diff
@@ -844,4 +847,84 @@ void getCompanyMetricsById_WithoutCrmRole_ReturnsForbidden() throws Exception {
...
+	@Test
+	@DisplayName("Get companies by ids - Returns matching companies and ignores unknown ids")
+	void getCompaniesByIds_WithUnknownIds_ReturnsOnlyExisting() throws Exception {
+		CrmCompany companyA = savedBatchCompany("BatchCoA");
+		CrmCompany companyB = savedBatchCompany("BatchCoB");
+
+		performBatchRequest(List.of(companyA.getId(), companyB.getId(), 999999L)).andDo(print())
+			.andExpect(status().isOk())
+			.andExpect(jsonPath(STATUS_PATH).value(STATUS_SUCCESSFUL))
+			.andExpect(jsonPath("['results'].length()").value(2));
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`getCompaniesByIds_WithUnknownIds_ReturnsOnlyExisting` only asserts `results.length() == 2`; it never checks that the two returned rows are companyA and companyB, so a filter regression that returned two arbitrary companies would still pass. Assert the ids explicitly, e.g. `jsonPath("['results'][*]['id']")` with `containsInAnyOrder(companyA.getId().intValue(), companyB.getId().intValue())`. Two behaviours are also untested and worth adding while you are here: a request containing the same id twice (documents whether the response is deduplicated), and a single request mixing a live id with a soft-deleted id (the current soft-delete test uses a soft-deleted id only, so it would still pass if the filter dropped every row).

backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmCompanyController.java
anusham-rc: Agreed this is accurate, but it's pre-existing global behaviour, not something this endpoint introduces — `GlobalExceptionHandler` has only the catch-all `@ExceptionHandler(Exception.class)`, so every JSON endpoint (`createCompany`, `editCompany`, etc.) returns 500 on an unreadable body today. Adding a `HttpMessageNotReadableException` handler is a cross-cutting change to shared infrastructure and shouldn't ride along in a single-endpoint feature PR. Tracking it separately.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmCompanyServiceImpl.java
anusham-rc: Leaving the lenient behaviour as-is. This is a store-hydration endpoint, and returning an empty result for null/empty ids is intentional — the integration tests (`getCompaniesByIds_EmptyIds_ReturnsEmptyList`) encode it. This also directly conflicts with the guidance in the important comment above, which says to *keep* the empty-returns-empty path and not add `@NotEmpty`. Picking that strategy: lenient at the boundary, no `ModuleException` for null/empty.

backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmCompanyController.java
anusham-rc: POST is deliberate here. The id list is client-driven and unbounded in length, so a GET query string risks URL-length limits — a request body is the right transport for a batch lookup. Caching/ETags aren't a concern for a hydration call whose payload varies per client. Keeping `POST /batch`.

backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmCompanyController.java
anusham-rc: Confirmed intentional — as noted in the PR summary, this was split out of the v2 scalar-keys refactor branch to merge independently; the consuming frontend fetcher lands in a follow-up. Not forgotten API surface.

backend/src/main/java/com/skapp/community/crmplanner/payload/request/CrmCompanyIdsRequestDto.java
anusham-rc: Not actioning this. Two reasons:

- This module doesn't use bean validation anywhere — no `@Valid`, `@Size`, or `jakarta.validation` annotations exist across the CRM controllers or DTOs; validation is done exclusively through `CrmValidations.*` service-layer calls. So `@Valid`/`@Size` would introduce a pattern the module doesn't otherwise follow.
- The endpoint is role-restricted to authenticated CRM users and is a store-hydration call: the ids come from companies the client already legitimately holds, so the list is bounded in practice by the caller's own dataset. The unbounded-`IN` risk is low enough that a cap isn't warranted here.

Leaving the lenient empty/null short-circuit as the only guard, consistent with the tests.

backend/src/test/java/com/skapp/community/crmplanner/controller/v1/CrmCompanyControllerIntegrationTest.java
anusham-rc: Not doing this one — it's scoped outside this PR's changes. Consolidating `savedBatchCompany` would mean rewriting the pre-existing `createMetricsCompany` and the inline fixture in `getCompanyById_HappyPath` to delegate to a shared helper, i.e. refactoring passing tests this PR never touched, for no behaviour change. Keeping the batch tests' fixture local and the diff tight. A fixture-consolidation pass across this test class is better done on its own.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmCompanyServiceImpl.java
anusham-rc: Done — added `getCompaniesByIds_NullIds_ReturnsEmptyList`, which posts a null `ids` and asserts 200 with an empty `results`, locking in the intended silent-empty behaviour on the `== null` branch.

backend/src/test/java/com/skapp/community/crmplanner/controller/v1/CrmCompanyControllerIntegrationTest.java
anusham-rc: Done — switched the batch tests to the `RESULTS_PATH` constant (added the import); `["['results'].length()"]` is now `RESULTS_PATH + ".length()"`. No bare `['results']` literals remain in the class.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmCompanyServiceImpl.java
anusham-rc: Done — added `List<CrmCompanyResponseDto> crmCompaniesToCrmCompanyResponseDtos(List<CrmCompany>)` to `CrmMapper` and used it in both `getCompaniesByIds` and `searchCompaniesByDomain`, removing the duplicated `.stream().map(...).toList()` from both.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmCompanyServiceImpl.java
anusham-rc: Partially done — added `CrmValidations.validateCompanyIds(List<Long>)` (mirroring `validateCompanyId`) that rejects null or non-positive elements with `CRM_ERROR_COMPANY_NOT_FOUND` → 400, called after the empty short-circuit so null/empty still returns empty. Skipped the dedup: the derived `IN` query already returns each company once (added a duplicate-ids test documenting this), so deduping the request adds nothing for this caller.

backend/src/main/java/com/skapp/community/crmplanner/repository/CrmCompanyDao.java
anusham-rc: Done (ordering) — renamed the query to `findByIdInAndIsDeletedFalseOrderByIdAsc` so the response is deterministically id-ordered. Kept the parameter as `List` rather than `Collection`, since we decided not to dedupe into a `Set`, so there's no caller that needs the wider type.

backend/src/main/java/com/skapp/community/crmplanner/payload/request/CrmCompanyBatchRequestDto.java
anusham-rc: Done (rename) — renamed `CrmCompanyBatchRequestDto` → `CrmCompanyIdsRequestDto` (via `git mv`, references updated across controller/service/impl/test). Left the shared `IdListRequestDto` idea out on purpose: folding this into a common id-bag DTO would couple CRM to the invitation payload and lose the per-operation context in Swagger for a single-field type.

backend/src/test/java/com/skapp/community/crmplanner/controller/v1/CrmCompanyControllerIntegrationTest.java
anusham-rc: Done — `getCompaniesByIds_WithUnknownIds_ReturnsOnlyExisting` now asserts the returned ids via `containsInAnyOrder(companyA, companyB)`, not just the count. Also added the two suggested cases: a duplicate-id request (documents that the company comes back once) and a request mixing a live id with a soft-deleted id (asserts only the live one returns).

backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmCompanyController.java
anusham-rc: Update — reversed course on this: the path is now `POST /v1/crm/company/by-ids` (was `/batch`), matching the `getCompaniesByIds` method and the existing `/search-by-domain` convention. The request-body transport stays (id list is client-driven and unbounded in length, so GET query-string limits are a real risk), but the path now names the operation rather than the mechanism.

backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmCompanyController.java
```diff
@@ -97,6 +98,16 @@ public ResponseEntity<ResponseEntityDto> getCompanyById(@PathVariable Long id) {
...
+			description = "Returns the base details of the companies matching the given ids, "
+					+ "used to hydrate the client's company store. Soft-deleted companies are omitted.")
+	@PostMapping("/by-ids")
```
akilarootcode: remove by

akilarootcode (changes requested):

backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmCompanyController.java
anusham-rc: updated to ids
now: company/ids

akilarootcode (approved):

ThinuwanW (approved):

PR #1791, opened 2026-08-18
Fix/3041 business unit name tooltip

shakila-kamalasena:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/frontend/src/community/configurations/components/molecules/DeleteBusinessUnitModal/DeleteBusinessUnitTransferContent.tsx
+++ b/frontend/src/community/configurations/components/molecules/DeleteBusinessUnitModal/DeleteBusinessUnitTransferContent.tsx
@@ -36,7 +36,11 @@ const DeleteBusinessUnitTransferContent: FC<Props> = ({
     },
     ...(otherUnits?.map((unit) => ({
       id: String(unit.businessUnitId),
-      label: unit.name,
+      label: (
+        <span title={unit.name} className="min-w-0 flex-1 truncate text-left">
+          {unit.name}
+        </span>
+      ),
       value: String(unit.businessUnitId)
     })) ?? [])
   ];
@@ -51,6 +55,7 @@ const DeleteBusinessUnitTransferContent: FC<Props> = ({
       </p>
       <Dropdown
         id="business-unit-transfer-target"
+        variant="jsx-content"
         label={translateText(["deleteModal", "transferToLabel"])}
         tooltipContent={translateText(["deleteModal", "transferTooltip"])}
         ariaLabelTooltipButton={translateText([
--- a/frontend/src/community/configurations/components/molecules/DeleteBusinessUnitModal/DeleteBusinessUnitTransferContent.tsx
+++ b/frontend/src/community/configurations/components/molecules/DeleteBusinessUnitModal/DeleteBusinessUnitTransferContent.tsx
@@ -37,7 +37,7 @@ const DeleteBusinessUnitTransferContent: FC<Props> = ({
     ...(otherUnits?.map((unit) => ({
       id: String(unit.businessUnitId),
       label: (
-        <span title={unit.name} className="min-w-0 flex-1 truncate text-left">
+        <span title={unit.name} className="min-w-0 flex-1 truncate">
           {unit.name}
         </span>
       ),


hasalarootcode (approved):

ThinuwanW (commented):
## 🤖 Claude Code Review

This PR converts the business-unit transfer dropdown option labels from plain strings to a JSX span with `title` + Tailwind `truncate`, and switches the skapp-ui `Dropdown` to `variant="jsx-content"`. The `title`+`truncate` idiom matches established patterns in the codebase (BusinessUnitCard, ContactOptionItem, CompanyTable) and `variant="jsx-content"` matches PriorityDropdown, so the direction is sound. Remaining concerns are all in the details: only the business-unit options get truncation (the unassign option is left as a bare string), the `flex-1` utility silently depends on the library rendering the label inside a flex container (and `truncate` on an inline `span` is a no-op otherwise), the JSX option array is rebuilt on every render unlike the memoized precedent, and the menu width is not pinned to the trigger so the truncation may never actually engage.

Found **12** new issue(s): 🔴 1 important, 🟡 7 suggestion(s), 🟣 4 nit(s)

frontend/src/community/configurations/components/molecules/DeleteBusinessUnitModal/DeleteBusinessUnitTransferContent.tsx
```diff
@@ -36,7 +36,11 @@ const DeleteBusinessUnitTransferContent: FC<Props> = ({
     },
     ...(otherUnits?.map((unit) => ({
       id: String(unit.businessUnitId),
-      label: unit.name,
+      label: (
+        <span title={unit.name} className="min-w-0 flex-1 truncate">
```
ThinuwanW, line 40: 🤖 **Claude** · 🟡 **Suggestion**

`min-w-0 flex-1 truncate` makes truncation dependent on skapp-ui internals. Tailwind's `truncate` expands to `overflow:hidden; text-overflow:ellipsis; white-space:nowrap` but does NOT set `display` — those properties have no effect on an inline element. The span only ellipsises here because `flex-1` blockifies it as a flex item, which is true only if the library renders the option content (and the closed trigger content) inside a `display:flex` container. If either the menu item wrapper or the trigger is not a flex container, `flex-1` is inert and the long name will overflow instead of truncating, with no compile-time or test signal. Prefer a layout-independent form that matches the existing patterns in this repo, e.g. `className="block w-full truncate"` (see CompanyTable.tsx:94 `body2 block w-full truncate` and ContactOptionItem.tsx:23 `body2 w-full truncate`), optionally keeping `min-w-0`. Please also verify visually in both places the label renders: inside the open menu AND inside the collapsed trigger once a unit is selected.

frontend/src/community/configurations/components/molecules/DeleteBusinessUnitModal/DeleteBusinessUnitTransferContent.tsx
```diff
@@ -36,7 +36,11 @@ const DeleteBusinessUnitTransferContent: FC<Props> = ({
     },
     ...(otherUnits?.map((unit) => ({
       id: String(unit.businessUnitId),
-      label: unit.name,
+      label: (
```
ThinuwanW, line 39: 🤖 **Claude** · 🟡 **Suggestion**

Inconsistent option labels: only the mapped business-unit options are wrapped in the truncating span, while the first "unassign" option (line 34) stays a bare translated string. Under `variant="jsx-content"` that string still renders, but it gets no `truncate`, no `min-w-0`/`flex-1` and no `title`, so it is styled differently from its siblings and can overflow the narrow SmallModal — `unassignOption` is already "Don't transfer (unassign)" in English and translations are typically longer. Extract a tiny local helper and use it for every option, e.g. `const renderOptionLabel = (text: string) => (<span title={text} className="block w-full truncate">{text}</span>)`, then apply it to both the unassign option and the mapped units. That also removes the `string | JSX.Element` union in the inferred option type.

frontend/src/community/configurations/components/molecules/DeleteBusinessUnitModal/DeleteBusinessUnitTransferContent.tsx
```diff
@@ -36,7 +36,11 @@ const DeleteBusinessUnitTransferContent: FC<Props> = ({
     },
     ...(otherUnits?.map((unit) => ({
```
ThinuwanW, line 37: 🤖 **Claude** · 🟡 **Suggestion**

`transferOptions` is rebuilt on every render, and now that labels are JSX it allocates brand-new React elements each time, so every option's identity changes on any parent re-render (this component re-renders on each `onChange`/selection change via the parent's `transferTargetValue` state). Any memoization inside the skapp-ui Dropdown/menu is defeated. The established pattern for jsx-content options in this codebase memoizes the array — see `useGetPriorityOptions.tsx`, which wraps the identical JSX-label mapping in `useMemo(..., [translateText])`. Wrap this array in `useMemo` with deps `[businessUnits, currentBusinessUnitId, translateText]` (the `otherUnits` filter can move inside the same memo).

frontend/src/community/configurations/components/molecules/DeleteBusinessUnitModal/DeleteBusinessUnitTransferContent.tsx
```diff
@@ -51,6 +55,7 @@ const DeleteBusinessUnitTransferContent: FC<Props> = ({
       </p>
       <Dropdown
         id="business-unit-transfer-target"
+        variant="jsx-content"
```
ThinuwanW, line 58: 🤖 **Claude** · 🟣 **Nit**

Consider also passing `menuWidth="match"` alongside `variant="jsx-content"`, as PriorityDropdown.tsx:28 does. `width="100%"` sizes the trigger, but if the popup menu sizes to its content (the default in other usages is explicitly `menuWidth: "content"`, e.g. ContactTable.tsx:234), the menu will simply grow to fit the longest business-unit name — the new `truncate` would never engage in the open menu and the popup could overflow the SmallModal instead. Pinning the menu to the trigger width is what makes the truncation meaningful for long names.

frontend/src/community/configurations/components/molecules/DeleteBusinessUnitModal/DeleteBusinessUnitTransferContent.tsx
```diff
@@ -51,6 +55,7 @@ const DeleteBusinessUnitTransferContent: FC<Props> = ({
...
         label={translateText(["deleteModal", "transferToLabel"])}
```
ThinuwanW, line 59: 🤖 **Claude** · 🟣 **Nit**

Accessibility check worth confirming: `label` is no longer a string, so anything the library derives from it as text (the trigger's accessible name, `aria-label`, typeahead/keyboard search, or a `String(label)` interpolation) will now receive a React element rather than the unit name — a `String(...)` path would render "[object Object]". Please verify with a screen reader / the accessibility inspector that the trigger and the listbox options still announce the business-unit name after selection. Note also that the native `title` attribute is not exposed on keyboard focus, so keyboard-only users lose the full-name affordance; that matches the existing convention in BusinessUnitCard.tsx:50 so it is fine to keep, but the module has no test coverage (no `.test.tsx` under community configurations) that would catch a regression in either behaviour.

frontend/src/community/configurations/components/molecules/DeleteBusinessUnitModal/DeleteBusinessUnitTransferContent.tsx
ThinuwanW, line 58: 🤖 **Claude** · 🔴 **Important**

`variant="jsx-content"` looks unnecessary here, and switching to it changes the field's visual variant rather than just enabling JSX labels. Evidence from the codebase: `DealPropertiesSidebar.tsx:117-126` and `DealNameStageSection.tsx:93-104` both pass option `label`s that are JSX elements (`<StageLabel .../>`) while using `variant="primary"` / `"primary-error"` — so skapp-ui's Dropdown already renders ReactNode labels under the primary variant. Meanwhile the only other `jsx-content` consumer, `PriorityDropdown.tsx:23-34`, is a chrome-less inline chip dropdown that has to re-add its own presentation (`hideArrowIcon`, `padding="py-2 px-1"`, `height="min-h-8"`, `className="... bg-transparent border-none rounded-lg"`). That pattern suggests `jsx-content` strips the standard field treatment (border, height, padding) that a labeled form dropdown expects — and this one is a labeled modal form field with `label`, `tooltipContent`, and only `className="rounded-lg"`, so it has nothing to fall back on. Fix: remove `variant="jsx-content"` (or pass `variant="primary"` explicitly, matching `CompanyModalForm.tsx:128`), keep only the label-wrapping change, and confirm in the browser that the trigger still renders with its border/height and that the options display correctly. If `jsx-content` really is required for this case, add the compensating style props the way PriorityDropdown does.

frontend/src/community/configurations/components/molecules/DeleteBusinessUnitModal/DeleteBusinessUnitTransferContent.tsx
ThinuwanW, line 40: 🤖 **Claude** · 🟡 **Suggestion**

The new span carries no typography class. Every comparable truncating element in this codebase pairs `truncate` with an explicit typography token — `body2` in `CompanyTable.tsx:94`, `ContactOptionItem.tsx:23` and `StageLabel.tsx:16`, `subtitle4` in `ContactOptionItem.tsx:28`, `subtitle1` in `BusinessUnitCard.tsx:49`. Because the business-unit options are now raw JSX while the sibling 'Don't transfer (unassign)' option (line 34) is still a plain string styled by the library's own option renderer, the two can end up at different font sizes/weights in the same menu. Add the matching token (most likely `body2`) to the span so all options render identically.

frontend/src/community/configurations/components/molecules/DeleteBusinessUnitModal/DeleteBusinessUnitTransferContent.tsx
ThinuwanW, line 37: 🤖 **Claude** · 🟡 **Suggestion**

`transferOptions` has no explicit type. skapp-ui exports `DropdownOption` and the codebase already uses it as the canonical option shape (`crmUtil.ts:85`, `dealUtil.ts:13-14`, `ContactOptionItem.tsx:8`). Right now TypeScript infers a union element type — `{ label: string; ... } | { label: JSX.Element; ... }` — so nothing anchors these objects to the library contract; if a future skapp-ui version narrows `DropdownOption.label`, the mismatch surfaces as an opaque error at the `options={...}` prop instead of at the construction site. Change to `const transferOptions: DropdownOption[] = [...]` and import `DropdownOption` from `@rootcodelabs/skapp-ui`, per the strict-typing guideline that every variable have an explicit type.

frontend/src/community/configurations/components/molecules/DeleteBusinessUnitModal/DeleteBusinessUnitTransferContent.tsx
ThinuwanW, line 40: 🤖 **Claude** · 🟡 **Suggestion**

The native `title` attribute only surfaces on mouse hover, so it does not recover the truncated text for keyboard-only users, who reach dropdown options with arrow keys rather than a pointer. That matters more here than in the table cells where this pattern is normally used: business unit names can be up to 100 characters (`BUSINESS_UNIT_NAME_MAX_LENGTH` in `commonConstants.ts:39`), and choosing the wrong transfer target silently reassigns every employee of the unit being deleted. Consider letting the option text wrap instead of truncating (e.g. `line-clamp-2` / `whitespace-normal break-words`), or render the full name through the design-system tooltip already used on this same Dropdown via `tooltipContent`, so the value is reachable without hovering.

frontend/src/community/configurations/components/molecules/DeleteBusinessUnitModal/DeleteBusinessUnitTransferContent.tsx
```diff
@@ -36,7 +36,11 @@ const DeleteBusinessUnitTransferContent: FC<Props> = ({
...
     ...(otherUnits?.map((unit) => ({
       id: String(unit.businessUnitId),
-      label: unit.name,
+      label: (
+        <span title={unit.name} className="min-w-0 flex-1 truncate">
+          {unit.name}
+        </span>
+      ),
       value: String(unit.businessUnitId)
     })) ?? [])
```
ThinuwanW, line 45: 🤖 **Claude** · 🟣 **Nit**

Dead defensive code around the option list the PR is editing. `businessUnits` is declared as a required, non-nullable `BusinessUnit[]` in `Props` (line 9), and the only caller already short-circuits to a skeleton when it is falsy (`DeleteBusinessUnitModal.tsx:116-128` returns early on `!businessUnits`). So the `?.` in `businessUnits?.filter(...)` on line 27 is the sole reason `otherUnits` is typed `BusinessUnit[] | undefined`, which in turn is the only reason the `?? []` fallback on this line exists — neither branch is reachable. Drop the `?.` and the `?? []` so the type reflects reality, or make the prop `businessUnits?: BusinessUnit[]` if the nullability is genuinely expected at runtime.

frontend/src/community/configurations/components/molecules/DeleteBusinessUnitModal/DeleteBusinessUnitTransferContent.tsx
ThinuwanW, line 39: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Duplication: this is roughly the twelfth hand-rolled `<span className="…truncate" title={value}>{value}</span>` in the frontend — see `CompanyTable.tsx:94`, `ContactTable.tsx:137,149`, `ContactOptionItem.tsx:23,28`, `DealCard.tsx:88,96`, `StageLabel.tsx:16`, `TaskModalForm.tsx:194,207`, `GuestRequestsTable.tsx:111`, `BusinessUnitCard.tsx:49` — and each site picks a slightly different class combination (`block`, `w-full`, `min-w-0`, `flex-1`, `max-w-[200px]`). Per the code-duplication guideline, extract a small `TruncatedText` atom under `~community/common/components/atoms` that owns the display/min-width classes plus the `title` wiring, and use it here. That would also remove the per-site guesswork about which Tailwind classes actually make truncation take effect, which is exactly the ambiguity this change introduces (note `StageLabel.tsx:11` sidesteps it by supplying its own `flex min-w-0` wrapper rather than depending on the parent).

ThinuwanW, line 39: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

No test coverage: there is not a single `.test.tsx` file anywhere under `src/community/configurations`, yet this component drives a destructive flow (deleting a business unit and reassigning all its employees). A rendering-only change like this one is precisely the kind that regresses silently. Add `DeleteBusinessUnitTransferContent.test.tsx` wrapped in `MockTheme`, asserting with `screen.getByRole("option", { name: ... })` that the unassign option plus every other business unit is offered, that the current unit is excluded, and that selecting an option calls `onChange` with the stringified `businessUnitId`.

frontend/src/community/configurations/components/molecules/DeleteBusinessUnitModal/DeleteBusinessUnitTransferContent.tsx
shakila-kamalasena: No need

frontend/src/community/configurations/components/molecules/DeleteBusinessUnitModal/DeleteBusinessUnitTransferContent.tsx
shakila-kamalasena: No need

frontend/src/community/configurations/components/molecules/DeleteBusinessUnitModal/DeleteBusinessUnitTransferContent.tsx
shakila-kamalasena: No need

frontend/src/community/configurations/components/molecules/DeleteBusinessUnitModal/DeleteBusinessUnitTransferContent.tsx
shakila-kamalasena: No need

frontend/src/community/configurations/components/molecules/DeleteBusinessUnitModal/DeleteBusinessUnitTransferContent.tsx
shakila-kamalasena: No need

frontend/src/community/configurations/components/molecules/DeleteBusinessUnitModal/DeleteBusinessUnitTransferContent.tsx
shakila-kamalasena: No need

frontend/src/community/configurations/components/molecules/DeleteBusinessUnitModal/DeleteBusinessUnitTransferContent.tsx
shakila-kamalasena: No need

frontend/src/community/configurations/components/molecules/DeleteBusinessUnitModal/DeleteBusinessUnitTransferContent.tsx
shakila-kamalasena: No need

frontend/src/community/configurations/components/molecules/DeleteBusinessUnitModal/DeleteBusinessUnitTransferContent.tsx
shakila-kamalasena: No need

frontend/src/community/configurations/components/molecules/DeleteBusinessUnitModal/DeleteBusinessUnitTransferContent.tsx
shakila-kamalasena: No need

frontend/src/community/configurations/components/molecules/DeleteBusinessUnitModal/DeleteBusinessUnitTransferContent.tsx
shakila-kamalasena: No need

shakila-kamalasena: No need

akilarootcode (approved):

PR #1794, opened 2026-08-18
feat: Add board initialization data retrieval and update task service…

YasinduRC:
### Summary

Reshapes the CRM **v2 task API** around id references and adds a **v2 board init-data** endpoint that does the same. Backend only. First slice of the CRM v2 refactor — the batch lookup endpoints that resolve these ids follow in a separate PR.

- **`CrmTaskResponseDtoV2` returns ids, not nested objects** — `typeId`, `ownerId`, `contactId`, `companyId`, `dealId` replace the embedded `type` / `owner` / `contact` / `company` / `deal`.
- **Task endpoints collapsed 3 → 2:**

  | Before | After |
  |---|---|
  | `GET /v2/crm/task` (open only) | `GET /v2/crm/task` — `isCompleted` selects completed / open / both |
  | `GET /v2/crm/task/completed` | *removed, folded into the above* |
  | `GET /v2/crm/task/related?contactId=&dealId=` | `GET /v2/crm/task/{id}/related` |

- `GET /v2/crm/task` also gains `sortKey` (`DUE_AT` default, or `LAST_MODIFIED_DATE`) and `sortOrder`; `size < 0` returns every match unpaged.
- `GET /v2/crm/task/{id}/related` takes the source task's id and derives the links server-side: returns tasks sharing **every** link the source has — contact *and* deal when it carries both (AND, not OR). Company is not a link.
- **New `GET /v2/crm/board/init-data`** — the shared lookup set (stages, contacts, owners, task types, roles). Delegates to the v1 service and re-shapes only contacts, so stage visibility and everything else stay identical to v1; the sole difference is `companyId` instead of a nested `company`.

### Breaking changes

All v2-only; no v1 endpoint changes. The v1 board init-data endpoint is deliberately untouched because the v1 kanban reads `contact.company.name` from it — hence a new v2 endpoint rather than an in-place change.

1. `CrmTaskResponseDtoV2` consumers must read `*Id` fields.
2. `GET /v2/crm/task/completed` → use `GET /v2/crm/task?isCompleted=true`.
3. `GET /v2/crm/task/related` → `GET /v2/crm/task/{id}/related`, no `contactId` / `dealId` params.

### How to test

1. `cd backend && mvn test -Dtest='Crm*'` — 347 tests, all green.
2. `GET /v2/crm/task?isCompleted=false` → items carry `*Id` fields, no nested objects.
3. `GET /v2/crm/task?size=-1` → all matches in one page.
4. Task linked to a contact **and** a deal → `GET /v2/crm/task/{id}/related` returns only tasks sharing both.
5. `GET /v2/crm/board/init-data` → contacts have `companyId`, no `company` object.
6. No v1 regression: deals kanban still shows company names.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
- [x] Code is formatted with `npm run format` *(backend: `spring-javaformat:validate` clean)*

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information

No frontend changes — nothing calls these endpoints yet.

`CrmTaskControllerV2IntegrationTest` grows 11 → 22 tests; `CrmBoardControllerV2IntegrationTest` is new (3). Worth a close look: `getRelatedTasks_ContactAndDealSource_RequiresBothToMatch` and `getRelatedTasks_SharesCompanyOnly_ReturnsNoMatches` pin down the AND semantics.


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR reshapes the CRM v2 task API to id-reference DTOs, merges the open/completed task endpoints into a single paginated `GET /v2/crm/task`, re-anchors related-tasks to `/{id}/related` with a proper ownership gate, and adds a `/v2/crm/board/init-data` v2 endpoint. The direction is good — the new related-tasks endpoint actually adds an authorization check that the old query-param version lacked, and the criteria projection is now much leaner. The main concerns are a 500-producing gap in the new `toPageable` helper (`size=0` / negative `page`), a genuinely unbounded query mode exposed via `size < 0`, a brittle unchecked downcast when the v2 board service re-wraps the v1 response, and missing test coverage for the new ownership gate.

Found **22** new issue(s): 🔴 3 important, 🟡 14 suggestion(s), 🟣 5 nit(s)

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmTaskServiceImplV2.java
```diff
@@ -131,6 +116,10 @@ public ResponseEntityDto editTask(Long id, CrmTaskEditRequestDto requestDto) {
 		return new ResponseEntityDto(false, CrmUtil.toTaskResponseDtoV2(crmMapperV2, updatedTask));
 	}
 
+	private Pageable toPageable(int page, int size) {
+		return size < 0 ? Pageable.unpaged() : PageRequest.of(page, size);
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`toPageable` only treats `size < 0` as the unpaged sentinel, so `size = 0` falls through to `PageRequest.of(page, 0)`, which throws `IllegalArgumentException("Page size must not be less than one")`. The same happens for any negative `page` (`"Page index must not be less than zero"`). Neither exception is handled specifically — `GlobalExceptionHandler` only catches `BindException`/`ModuleException`/etc., so both fall through to `@ExceptionHandler(Exception.class)` and return a 500 to the client. This is reachable from both `GET /v2/crm/task?size=0` and `GET /v2/crm/task/{id}/related?size=0` (and `?page=-1`). Fix by making the sentinel inclusive and guarding page, e.g. `return size <= 0 ? Pageable.unpaged() : PageRequest.of(Math.max(page, 0), size);` — or, better, reject out-of-range values up front with bean validation so the client gets a 422 instead.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmTaskServiceImplV2.java
```diff
@@ -131,6 +116,10 @@ public ResponseEntityDto editTask(Long id, CrmTaskEditRequestDto requestDto) {
 		return new ResponseEntityDto(false, CrmUtil.toTaskResponseDtoV2(crmMapperV2, updatedTask));
 	}
 
+	private Pageable toPageable(int page, int size) {
```
ThinuwanW, line 123: 🤖 **Claude** · 🔴 **Important**

`size < 0` disables pagination entirely, and `findTasksV2` then runs the projection query with no `setMaxResults`. For a non-sales-rep caller (CRM_ADMIN / CRM_SALES_MANAGER, who get `ownerId == null` via `CrmUtil.isCrmSalesRepresentative`) that is a full, unfiltered scan of `crm_task` materialised into DTOs in one response — the review guidelines call out unbounded queries without pagination explicitly. Any authenticated CRM user can trigger it with a single query param. Suggest either dropping the unpaged mode, or clamping it to a documented maximum (e.g. `PageRequest.of(0, MAX_PAGE_SIZE)` with `MAX_PAGE_SIZE` as a constant / `application.yml` value). Note also that when unpaged the separate count query still executes, which is pure waste since `content.size()` is already the total.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmDealServiceImplV2.java
```diff
@@ -37,6 +43,31 @@ public class CrmDealServiceImplV2 implements CrmDealServiceV2 {
 
 	private final UserService userService;
 
+	@Override
+	@Transactional(readOnly = true)
+	public ResponseEntityDto getBoardInitData() {
+		log.info("getBoardInitData: execution started");
+
+		ResponseEntityDto v1Response = crmDealService.getBoardInitData();
+		CrmBoardInitDataResponseDto v1InitData = (CrmBoardInitDataResponseDto) v1Response.getResults().getFirst();
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`(CrmBoardInitDataResponseDto) v1Response.getResults().getFirst()` reaches into the v1 response envelope and blind-casts it. Two failure modes, both surfacing as a 500 via the generic handler: `getFirst()` throws `NoSuchElementException` if `putToResults` ever receives null (it silently skips nulls), and the cast throws `ClassCastException` if `CrmDealServiceImpl.getBoardInitData()` ever changes its payload type. Nothing in the type system stops either. Prefer extracting the build logic into a typed method on the `CrmDealService` interface — e.g. `CrmBoardInitDataResponseDto buildBoardInitData()` — and have both `CrmDealServiceImpl.getBoardInitData()` and this v2 method call it. That also removes the need for v2 to know how v1 wraps its results.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmTaskRepositoryImpl.java
```diff
@@ -262,94 +243,91 @@ public Page<CrmTaskResponseDtoV2> findCompletedTasksV2(Long ownerId, CrmTaskComp
...
+		CrmTaskSort key = sortKey == null ? CrmTaskSort.DUE_AT : sortKey;
+
+		List<Order> orders = new ArrayList<>();
+		if (key == CrmTaskSort.LAST_MODIFIED_DATE) {
+			Expression<?> lastModified = task.get(Auditable_.lastModifiedDate);
+			orders.add(direction.isAscending() ? cb.asc(lastModified) : cb.desc(lastModified));
+			orders.add(direction.isAscending() ? cb.asc(task.get(CrmTask_.id)) : cb.desc(task.get(CrmTask_.id)));
+		}
+		else {
+			orders.add(cb.asc(cb.selectCase().when(cb.isNull(task.get(CrmTask_.dueAt)), 1).otherwise(0)));
```
ThinuwanW, line 268: 🤖 **Claude** · 🟡 **Suggestion**

The two branches of `buildTaskOrder` apply `sortOrder` inconsistently. In the `LAST_MODIFIED_DATE` branch the `id` tie-break follows the requested direction; in the `DUE_AT` branch both the null-grouping expression (`cb.asc(case when dueAt is null ...)`) and the `id` tie-break are hard-coded ASC. So `sortKey=DUE_AT&sortOrder=DESC` still pushes null-due-date tasks to the *end* and breaks ties ascending, which is likely to surprise a caller who asked for a full reversal. Pick one convention (nulls-last always, or nulls follow direction) and apply the same tie-break rule in both branches, then state it in the `@Operation` description on `CrmTaskControllerV2.getTasks`.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmTaskRepositoryImpl.java
```diff
@@ -262,94 +243,91 @@ public Page<CrmTaskResponseDtoV2> findCompletedTasksV2(Long ownerId, CrmTaskComp
...
 		Root<CrmTask> countRoot = countQuery.from(CrmTask.class);
 		countQuery.select(cb.count(countRoot))
-			.where(buildRelatedTaskPredicates(cb, countRoot, params).toArray(new Predicate[0]));
+			.where(buildRelatedTaskPredicates(cb, countRoot, taskId, params).toArray(new Predicate[0]));
 		Long total = entityManager.createQuery(countQuery).getSingleResult();
 
 		return new PageImpl<>(content, pageable, total);
 	}
 
+	private List<Predicate> buildRelatedTaskPredicates(CriteriaBuilder cb, Root<CrmTask> root, Long taskId,
```
ThinuwanW, line 309: 🤖 **Claude** · 🟡 **Suggestion**

There are now two overloads named `buildRelatedTaskPredicates` in this class with *opposite* matching semantics: the new 4-arg version (line 297) ANDs the contact and deal predicates, while the pre-existing 3-arg version (line 415, used by v1 `findRelatedTasks`) ORs them. Same name, same package, same params object, contradictory behaviour — this is an easy trap for the next person who picks the wrong overload via IDE completion. Rename them to encode the semantics, e.g. `buildRelatedTaskPredicatesMatchingAllLinks(...)` for the new one and `buildRelatedTaskPredicatesMatchingAnyLink(...)` for the v1 one.

backend/src/main/java/com/skapp/community/crmplanner/payload/request/CrmTaskFilterDtoV2.java
```diff
@@ -0,0 +1,30 @@
...
+
+	private Long companyId;
+
+	private Boolean isCompleted;
+
+	private CrmTaskSort sortKey;
+
+	private Sort.Direction sortOrder;
+
+	private int page = 0;
```
ThinuwanW, line 26: 🤖 **Claude** · 🟡 **Suggestion**

`page` and `size` carry no bean validation, so out-of-range values reach `PageRequest.of` and surface as 500s (see the `toPageable` finding) rather than a clean 422 via the `BindException` handler. Add `@Min(0)` on `page` and a bounded `@Max` on `size` (with the unpaged sentinel handled explicitly, e.g. `@Min(-1) @Max(500)`), and annotate the controller class with `@Validated` plus `@Valid` on the `CrmTaskFilterDtoV2` parameter so the constraints actually fire for query-param binding. The same applies to the loose `@RequestParam int page/size` on `CrmTaskControllerV2.getRelatedTasks` — those bypass the DTO entirely and have no bounds at all.

backend/src/main/java/com/skapp/community/crmplanner/payload/request/CrmTaskFilterDtoV2.java
```diff
@@ -0,0 +1,30 @@
+package com.skapp.community.crmplanner.payload.request;
+
+import com.skapp.community.crmplanner.type.CrmTaskSort;
+import lombok.Getter;
+import lombok.Setter;
+import org.springframework.data.domain.Sort;
+
+@Getter
+@Setter
+public class CrmTaskFilterDtoV2 {
```
ThinuwanW, line 10: 🤖 **Claude** · 🟡 **Suggestion**

All four of `searchKeyword`, `contactId`, `dealId`, `companyId` are copied verbatim from `CrmTaskFilterDto`. The existing `CrmTaskCompletedFilterDto` already establishes the pattern of `extends CrmTaskFilterDto` and adding only the extra fields. Extending `CrmTaskFilterDto` here and declaring only `isCompleted`, `sortKey`, `sortOrder`, `page`, `size` would remove the duplication and keep the two filter shapes from drifting apart when a filter field is added later.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmTaskControllerV2IntegrationTest.java
```diff
@@ -305,39 +319,211 @@ void getCompletedTasks_ReturnsPaginatedCompletedTasks() throws Exception {
...
+
+		performGetRelatedRequest(source.getId()).andDo(print())
+			.andExpect(status().isOk())
+			.andExpect(jsonPath(RESULTS_0_PATH + "['items'].length()").value(2))
+			.andExpect(jsonPath(RESULTS_0_PATH + "['totalItems']").value(2));
+	}
+
+	@Test
+	@DisplayName("Get related tasks for a task that does not exist - Returns Bad Request")
+	void getRelatedTasks_TaskNotFound_ReturnsBadRequest() throws Exception {
```
ThinuwanW, line 653: 🤖 **Claude** · 🟡 **Suggestion**

`getRelatedTasks` gained a brand-new authorization gate — it now loads the source task and throws `CRM_ERROR_TASK_VIEW_DENIED` when `CrmValidations.isOwnerRestrictedForRepresentative` fires — but no test exercises it. `getTaskById` has exactly this test (`getTaskById_SalesRepViewingOthersTask_ReturnsBadRequest`, line 561), so the pattern is right there to copy: flip employee 2 to `CRM_SALES_REPRESENTATIVE`, create a task owned by employee 1, re-issue the token as user2, and assert 400 + unsuccessful. Related gap: no test asserts that `getTasks` scopes results to the caller's own tasks when they are a sales rep — every current `getTasks` test runs as user1, for whom `ownerId` resolves to null, so the `ownerId` predicate path in `findTasksV2` is never covered.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmDealServiceImplV2.java
```diff
@@ -24,6 +27,9 @@
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
 
+import java.util.Collections;
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`import java.util.Collections;` was added but nothing in the file references `Collections`. Remove it. (Checkstyle here is minimal and won't catch it, so it will linger.)

backend/src/main/java/com/skapp/community/crmplanner/controller/v2/CrmBoardControllerV2.java
```diff
@@ -0,0 +1,34 @@
...
+public class CrmBoardControllerV2 {
+
+	private final CrmDealServiceV2 crmDealServiceV2;
+
+	@Operation(summary = "Get board init data",
+			description = "Returns the shared CRM lookup set: stages, contacts, CRM roles, owners and task types. "
+					+ "Related records are carried as id references only - a contact names its company by id rather "
+					+ "than nesting it.")
+	@GetMapping("/init-data")
+	@PreAuthorize("hasAnyRole('ROLE_CRM_SALES_REPRESENTATIVE')")
```
ThinuwanW, line 28: 🤖 **Claude** · 🟣 **Nit**

`hasAnyRole('ROLE_CRM_SALES_REPRESENTATIVE')` with a single role is equivalent to `hasRole(...)`. It mirrors the v1 `CrmBoardController`, so it is defensible, but every endpoint in the sibling `CrmTaskControllerV2` added/touched by this same PR uses `hasRole(...)`. Worth aligning on `hasRole` for the new v2 surface. (The authorization level itself checks out: `AuthorityServiceImpl.addRoleHierarchy` grants `ROLE_CRM_SALES_REPRESENTATIVE` to CRM_ADMIN and CRM_SALES_MANAGER too, so admins and managers are not locked out.)

backend/src/main/java/com/skapp/community/crmplanner/payload/request/CrmTaskFilterDtoV2.java
```diff
@@ -0,0 +1,30 @@
...
+
+	private Long dealId;
+
+	private Long companyId;
+
+	private Boolean isCompleted;
+
+	private CrmTaskSort sortKey;
+
+	private Sort.Direction sortOrder;
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`sortOrder` is typed as Spring Data's `org.springframework.data.domain.Sort.Direction`, which leaks a persistence-layer type into the public request contract (the sibling `sortKey` correctly uses the project's own `CrmTaskSort` enum). It also means binding goes through Spring's case-sensitive `StringToEnum` converter, so `?sortOrder=desc` (lowercase) fails with a 422 while `DESC` works — an inconsistency callers will hit. Consider a small project enum, or at minimum document the case requirement in the `@Operation` description.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmTaskRepositoryImpl.java
```diff
@@ -218,40 +214,25 @@ public CrmContactTaskMetrics findTaskMetricsByContactId(Long contactId) {
...
 		Root<CrmTask> task = query.from(CrmTask.class);
 
 		query.select(buildTaskProjectionSelection(cb, task));
 
-		CrmTaskFilterParams params = new CrmTaskFilterParams(ownerId, false, filterDto.getSearchKeyword(),
-				filterDto.getContactId(), filterDto.getDealId(), filterDto.getCompanyId());
+		CrmTaskFilterParams params = new CrmTaskFilterParams(ownerId, filterDto.getIsCompleted(),
+				filterDto.getSearchKeyword(), filterDto.getContactId(), filterDto.getDealId(),
+				filterDto.getCompanyId());
 		query.where(buildTaskPredicates(cb, task, params).toArray(new Predicate[0]))
```
ThinuwanW, line 244: 🤖 **Claude** · 🟣 **Nit**

`buildTaskProjectionSelection` already creates LEFT joins to contact, deal and company on the same root, and then `buildTaskPredicates` creates *additional* LEFT joins to contact and deal (for `searchKeyword`) and to company (for `companyId`). The result is duplicated joins in the emitted SQL for the most common list query. Results are unaffected (all are `@ManyToOne`, so no row multiplication), but it is avoidable work on the primary task-list endpoint. Threading the already-created joins into the predicate builder, or resolving joins once via a small helper that caches by attribute, would clean this up. Pre-existing, but this PR makes `getTasks` the single list endpoint so it now carries all the traffic.

backend/src/main/java/com/skapp/community/crmplanner/controller/v2/CrmTaskControllerV2.java
```diff
@@ -31,29 +30,28 @@ public class CrmTaskControllerV2 {
...
-	@GetMapping("/related")
+			description = "Returns a paginated list of tasks related to the given task - i.e. tasks that share every "
+					+ "link the source task has, both its contact and its deal when it carries both. A task with "
+					+ "neither has no related tasks. Supports pagination only (size < 0 disables pagination and "
+					+ "returns every related task).")
+	@GetMapping("/{id}/related")
 	@PreAuthorize("hasRole('ROLE_CRM_SALES_REPRESENTATIVE')")
-	public ResponseEntity<ResponseEntityDto> getRelatedTasks(CrmTaskRelatedFilterDto filterDto) {
-		ResponseEntityDto response = taskService.getRelatedTasks(filterDto);
+	public ResponseEntity<ResponseEntityDto> getRelatedTasks(@PathVariable Long id,
```
ThinuwanW, line 52: 🤖 **Claude** · 🔴 **Important**

`getRelatedTasks` takes `page` and `size` as bare `@RequestParam int` values with no bounds and no bean-validation wiring. `page=-1` reaches `toPageable(page, size)` -> `PageRequest.of(-1, 10)`, which throws `IllegalArgumentException`; GlobalExceptionHandler has no handler for it, so it falls through to the catch-all `@ExceptionHandler(Exception.class)` and the client gets a 500 instead of a 4xx. This is a separate entry point from the `CrmTaskFilterDtoV2` fields — it bypasses the DTO entirely, so adding validation annotations to the DTO does not cover it. Fix: annotate the class with `@Validated` and the params with `@Min(0)` on page (and a sensible `@Max` on size), or better, replace the three loose params with a small paging request DTO shared with the getTasks path so both endpoints validate identically.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmDealServiceImplV2.java
```diff
@@ -37,6 +43,31 @@ public class CrmDealServiceImplV2 implements CrmDealServiceV2 {
...
+		ResponseEntityDto v1Response = crmDealService.getBoardInitData();
+		CrmBoardInitDataResponseDto v1InitData = (CrmBoardInitDataResponseDto) v1Response.getResults().getFirst();
+
+		List<CrmBoardContactResponseDtoV2> contacts = v1InitData.getContacts()
+			.stream()
+			.map(contact -> new CrmBoardContactResponseDtoV2(contact.getId(), contact.getName(),
+					contact.getCompany() == null ? null : contact.getCompany().getId()))
+			.toList();
+
+		CrmBoardInitDataResponseDtoV2 responseDto = new CrmBoardInitDataResponseDtoV2();
```
ThinuwanW, line 79: 🤖 **Claude** · 🟡 **Suggestion**

The v1 -> v2 board reshape is hand-written field-by-field in the service (five setter calls plus an inline lambda for the contact projection). The review guidelines call out manual field-by-field mapping as something MapStruct should own, and `CrmMapperV2` already exists in this module and is injected into this very class. Move this to `CrmMapperV2` — e.g. `CrmBoardContactResponseDtoV2 crmBoardContactToV2(CrmBoardContactResponseDto)` with `@Mapping(target = "companyId", source = "company.id")` (MapStruct generates the null-safe nested access, replacing the ternary) plus a list method and a `CrmBoardInitDataResponseDto -> CrmBoardInitDataResponseDtoV2` mapping. That also removes the risk of a newly added v1 board field being silently dropped from v2, since the mapper will warn on unmapped targets.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmTaskServiceImplV2.java
ThinuwanW, line 123: 🤖 **Claude** · 🟡 **Suggestion**

`Pageable.unpaged()` flows into `new PageImpl<>(content, pageable, total)` and then into `toPageDto`. For an unpaged result Spring's `Chunk.getSize()` returns `content.size()`, so `Page.getTotalPages()` computes `ceil(total / content.size())` — and when the unpaged result is empty, `getSize()` is 0 and `getTotalPages()` returns 1. So an empty unpaged response reports `totalItems: 0, totalPages: 1`, whereas an empty paged response reports `totalItems: 0, totalPages: 0`. The same applies to the `Page.empty(pageable)` short-circuit in `findRelatedTasksV2` when called with an unpaged pageable. Clients that drive "load more" off `totalPages` will see an inconsistent contract. Fix: normalise in `toPageDto` — when the page is unpaged, set `totalPages` to `totalItems == 0 ? 0 : 1` explicitly rather than relying on Spring's unpaged arithmetic.

backend/src/main/java/com/skapp/community/crmplanner/repository/CrmTaskRepository.java
```diff
@@ -20,7 +22,7 @@ public interface CrmTaskRepository {
 
 	List<CrmTask> findTasks(Long ownerId, CrmTaskFilterDto filterDto);
 
-	List<CrmTaskResponseDtoV2> findTasksV2(Long ownerId, CrmTaskFilterDto filterDto);
+	Page<CrmTaskResponseDtoV2> findTasksV2(Long ownerId, CrmTaskFilterDtoV2 filterDto, Pageable pageable);
```
ThinuwanW, line 26: 🤖 **Claude** · 🟡 **Suggestion**

`findTasksV2(Long ownerId, CrmTaskFilterDtoV2 filterDto, Pageable pageable)` receives paging information twice: `filterDto` carries `page`/`size`, and `pageable` is derived from those same two fields by the service. The repository silently ignores the DTO's copies, so the two can disagree and nothing catches it. The sibling `findRelatedTasksV2(Long taskId, CrmTaskRelatedParams params, Pageable pageable)` has the opposite shape — no DTO at all. Pick one convention: either strip `page`/`size` out of `CrmTaskFilterDtoV2` into a separate paging concern (matching how `CrmTaskCompletedFilterDto` layered them on top of `CrmTaskFilterDto`), or drop the `Pageable` argument and let the repository build it. The same asymmetry shows up one layer up in `CrmTaskServiceV2`, where `getTasks(CrmTaskFilterDtoV2)` carries paging inside the DTO while `getRelatedTasks(Long, int, int)` takes loose primitives.

backend/src/main/java/com/skapp/community/crmplanner/type/CrmTaskSort.java
```diff
@@ -0,0 +1,9 @@
+package com.skapp.community.crmplanner.type;
+
+public enum CrmTaskSort {
```
ThinuwanW, line 8: 🤖 **Claude** · 🟡 **Suggestion**

`CrmTaskSort` is a bare enum, which deviates from the established sort-enum convention in this codebase: `CrmDealSort`, `NotificationSort`, `TeamSort`, `LeaveRequestSort` etc. all carry a `sortField` string and override `toString()` so callers can do `Sort.by(direction, sortKey.toString())` and hand the sort to `PageRequest`. Because this enum carries no field mapping, `CrmTaskRepositoryImpl.buildTaskOrder` has to hand-roll an if/else that will need editing for every new sort key. Related: `CrmTaskFilterDtoV2` leaves `sortKey`/`sortOrder` null and pushes the null-coalescing into the repository, whereas the sibling `CrmDealFilterDto` declares field-level defaults (`sortOrder = Sort.Direction.ASC`, `sortKey = CrmDealSort.STAGE_ORDER`). Putting the defaults on the DTO fields makes the API contract self-documenting and lets the repository stop guessing. (The DUE_AT nulls-last CASE expression still needs Criteria, so keep that branch — just drive it from the enum rather than an if/else chain.)

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmTaskServiceImplV2.java
```diff
@@ -45,48 +41,37 @@ public class CrmTaskServiceImplV2 implements CrmTaskServiceV2 {
...
 	}
 
 	@Override
 	@Transactional(readOnly = true)
-	public ResponseEntityDto getRelatedTasks(CrmTaskRelatedFilterDto filterDto) {
+	public ResponseEntityDto getRelatedTasks(Long id, int page, int size) {
 		log.info("getRelatedTasks: execution started");
 
-		CrmValidations.validateRelatedTaskContextFilter(filterDto.getContactId(), filterDto.getDealId());
+		CrmTask task = crmTaskDao.findByIdWithAssociations(id)
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`getRelatedTasks` calls `findByIdWithAssociations(id)`, which issues INNER fetch joins on type and owner plus LEFT fetch joins on contact, deal, deal.stage and deal.owner — a six-table hydration — but the method only ever reads three values off it: `owner.employeeId`, `contact.id` and `deal.id`. All three are FK columns on the task row itself. Consider a lightweight projection (a small Criteria/derived query returning those three ids plus the isDeleted guard), or reuse an existing `findByIdAndIsDeletedFalse` and read the FK ids via the proxies without triggering initialisation. Note the fetch graph also uses `JoinType.INNER` for `type`, so a task with a null type would be reported as "not found" — inherited from `getTaskById`, but now it gates a second endpoint.

backend/src/main/java/com/skapp/community/crmplanner/controller/v2/CrmBoardControllerV2.java
```diff
@@ -0,0 +1,34 @@
...
+	public ResponseEntity<ResponseEntityDto> getBoardInitData() {
```
ThinuwanW, line 29: 🤖 **Claude** · 🟡 **Suggestion**

This new endpoint delegates straight to the v1 `getBoardInitData`, which returns every contact (`findAllContactsForBoardInit()`), every CRM owner and every stage in the organisation with no owner scoping. Every other v2 CRM read path in this PR scopes a CRM_SALES_REPRESENTATIVE to their own records via `CrmUtil.isCrmSalesRepresentative(currentUser)` / `CrmValidations.isOwnerRestrictedForRepresentative`, so a sales rep who cannot read another rep's contact through `/v2/crm/contact/{id}` can still enumerate that contact's id, name and companyId here. It matches v1 behaviour, so it may well be deliberate (the board needs the full lookup set), but this PR is creating a brand-new surface for it — worth confirming the exposure is intended and, if so, adding a short comment on the service method explaining why board init data is deliberately unscoped.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmTaskControllerV2IntegrationTest.java
```diff
@@ -149,15 +149,23 @@ private ResultActions performGetTasksRequest() throws Exception {
 		return performRequest(get(BASE_PATH).accept(MediaType.APPLICATION_JSON), authToken);
 	}
 
-	private ResultActions performGetCompletedRequest() throws Exception {
+	private ResultActions performGetTasksRequest(String isCompleted) throws Exception {
```
ThinuwanW, line 153: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

The new tests cover isCompleted (true/false/absent), contactId, paging, unpaged and the related-task matrix well, but several behaviours introduced or reshaped by this PR have no test: (1) `searchKeyword` on the v2 getTasks endpoint — it now goes through a different projection with extra LEFT joins than the v1 path that was previously tested; (2) the `dealId` and `companyId` filters on v2 getTasks; (3) `sortKey=DUE_AT` with `sortOrder=DESC`, which is exactly the branch where the id tie-break stays ASC while dueAt flips — a test here would pin the intended ordering; (4) `sortKey=LAST_MODIFIED_DATE` with ASC; (5) `size=0`, which currently reaches `PageRequest.of(page, 0)`. Adding (3) and (5) in particular would turn two latent behaviours into asserted contract.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmTaskControllerV2IntegrationTest.java
```diff
@@ -209,7 +232,7 @@ private CrmTask savedTask(String name, boolean isCompleted) {
 
 	@Test
 	@DisplayName("Create task with contact and deal - Returns Created with embedded type, owner, company, contact, deal")
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Stale `@DisplayName`: the method was renamed to `createTask_WithContactAndDeal_ReturnsIdReferences` and the assertions were rewritten to assert `typeId`/`ownerId`/`companyId`/`contactId`/`dealId` and that `type`/`owner`/`deal` do not exist, but the display name still reads "Returns Created with embedded type, owner, company, contact, deal" — the exact opposite of what is now asserted. This is the string that shows up in CI reports. Update it to match, e.g. "Create task with contact and deal - Returns Created with id references only". (The sibling `getTaskById` display name was correctly updated in the same diff.)

backend/src/main/java/com/skapp/community/crmplanner/controller/v2/CrmTaskControllerV2.java
YasinduRC: Moved to pageTransformer util

backend/src/main/java/com/skapp/community/crmplanner/payload/request/CrmTaskFilterDtoV2.java
YasinduRC: This is v2 DTO and these old details will be replaced

backend/src/main/java/com/skapp/community/crmplanner/payload/request/CrmTaskFilterDtoV2.java
YasinduRC: We do not add bean validations

backend/src/main/java/com/skapp/community/crmplanner/payload/request/CrmTaskFilterDtoV2.java
YasinduRC: The codebase is using this as the convection

backend/src/main/java/com/skapp/community/crmplanner/controller/v2/CrmBoardControllerV2.java
YasinduRC: Fixed, moved to hasRole

backend/src/main/java/com/skapp/community/crmplanner/controller/v2/CrmBoardControllerV2.java
YasinduRC: Intentional, `getBoardInitData` initializes the CRM store, so it needs the full lookup data without owner scoping.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmDealServiceImplV2.java
YasinduRC: Removed

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmTaskServiceImplV2.java
YasinduRC: size=-1 is intentional , the frontend uses it to load everything, matching the existing leave-service pattern. I also fixed the unnecessary count query and the size=0 500.

YasinduRC: Moved to pageTransformer util

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmTaskServiceImplV2.java
YasinduRC: Moved to pageTransformer util

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmTaskRepositoryImpl.java
YasinduRC: ompanyId — switched it to the direct FK path. Left the search joins since they’re required for contact.name/deal.name, and avoiding a new join-sharing pattern for no functional gain.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmTaskRepositoryImpl.java
YasinduRC: The v1 version of this function will be removed after the refactor

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmTaskRepositoryImpl.java
YasinduRC: This matches v1 on purpose. v1 has no sortOrder and hardcodes two different orderings  open/related lists use nulls-last, dueAt asc, id asc, while the completed list uses lastModified desc, id desc

backend/src/main/java/com/skapp/community/crmplanner/repository/CrmTaskRepository.java
YasinduRC: This matches the existing v1 pattern. findTasksV2 follows v1 by carrying page/size in the DTO and building Pageable from them. getRelatedTasks is different because the v2 endpoint derives the related data from the task ID, so there’s no filter DTO. Leaving as-is for v1 consistency.

backend/src/main/java/com/skapp/community/crmplanner/type/CrmTaskSort.java
YasinduRC: `DUE_AT` needs custom Criteria `CASE` logic for nulls-last, so the `sortField`/`toString()` pattern doesn't fully fit. Defaults are handled in the repository, so leaving the enum simple is intentional.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmTaskServiceImplV2.java
YasinduRC: Fixed. Added a lightweight CrmTaskLinkRefs projection to fetch only the required IDs, avoiding the full association graph and the inherited INNER JOIN issue for tasks with no type.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmDealServiceImplV2.java
YasinduRC: Done, moved the v1→v2 mapping into CrmMapperV2. The service now just fetches, maps, and wraps, with MapStruct handling nulls and unmapped-field warnings.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmTaskControllerV2IntegrationTest.java
YasinduRC: Fixed the tests

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmTaskControllerV2IntegrationTest.java
YasinduRC: Fixed the tests

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmTaskControllerV2IntegrationTest.java
YasinduRC: Added the tests

backend/src/main/java/com/skapp/community/crmplanner/controller/v2/CrmBoardControllerV2.java
```diff
@@ -0,0 +1,34 @@
...
+@Tag(name = "CRM Board Controller V2", description = "Operations related to CRM board")
+public class CrmBoardControllerV2 {
+
+	private final CrmDealServiceV2 crmDealServiceV2;
+
+	@Operation(summary = "Get board init data",
+			description = "Returns the shared CRM lookup set: stages, contacts, CRM roles, owners and task types. "
+					+ "Related records are carried as id references only - a contact names its company by id rather "
+					+ "than nesting it.")
+	@GetMapping("/init-data")
```
erandi-rootcode, line 27: init-data means initial data? if the operation is like something that returns defaults values, can we look into another name> eg: default-data

backend/src/main/java/com/skapp/community/crmplanner/controller/v2/CrmTaskControllerV2.java
```diff
@@ -31,29 +30,28 @@ public class CrmTaskControllerV2 {
...
+			@RequestParam(defaultValue = "0") int page, @RequestParam(defaultValue = "10") int size) {
```
erandi-rootcode: move to a requestDto

backend/src/main/java/com/skapp/community/crmplanner/payload/response/v2/CrmBoardInitDataResponseDtoV2.java
```diff
@@ -0,0 +1,25 @@
...
+
+@Getter
+@Setter
+public class CrmBoardInitDataResponseDtoV2 {
+
+	private List<CrmBoardStageResponseDto> stages;
+
+	private List<CrmBoardContactResponseDtoV2> contacts;
+
+	private List<String> crmRoles;
```
erandi-rootcode: isn't this coming from an enum?

erandi-rootcode (changes requested):

backend/src/main/java/com/skapp/community/crmplanner/controller/v2/CrmBoardControllerV2.java
YasinduRC: Since this is an existing API we had we decided to keep it as that name

backend/src/main/java/com/skapp/community/crmplanner/controller/v2/CrmTaskControllerV2.java
YasinduRC: Moved

backend/src/main/java/com/skapp/community/crmplanner/payload/response/v2/CrmBoardInitDataResponseDtoV2.java
YasinduRC: Removed the from the response it is not used in anywhere

YasinduRC: Removed from the reponse

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmDealServiceImplV2.java
YasinduRC: Already fixed  this method no longer calls v1 at all, it builds the response directly from its own queries, so there's no envelope to unwrap or cast.

erandi-rootcode (dismissed):

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmDealServiceImplV2.java
```diff
@@ -33,10 +46,57 @@ public class CrmDealServiceImplV2 implements CrmDealServiceV2 {
...
+	public ResponseEntityDto getBoardInitData() {
+		log.info("getBoardInitData: execution started");
+
+		List<CrmDealStage> visibleStages = filterVisibleDealStages(
+				crmDealStageDao.findAllByIsDeletedFalseOrderByOrderIndexAsc());
+		List<CrmBoardStageResponseDto> stages = crmMapper.crmDealStagesToCrmBoardStageResponseDtos(visibleStages);
+
+		List<CrmBoardContactResponseDtoV2> contacts = crmContactDao.findAllContactsForBoardInit()
+			.stream()
+			.map(crmMapperV2::crmContactToCrmBoardContactResponseDtoV2)
```
akilarootcode: get directly from query

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmDealServiceImplV2.java
```diff
@@ -33,10 +46,57 @@ public class CrmDealServiceImplV2 implements CrmDealServiceV2 {
...
+		responseDto.setStages(stages);
+		responseDto.setContacts(contacts);
+		responseDto.setOwners(owners);
+		responseDto.setTaskTypes(taskTypes);
+
+		log.info("getBoardInitData: execution ended");
+		return new ResponseEntityDto(false, responseDto);
+	}
+
+	protected List<CrmDealStage> filterVisibleDealStages(List<CrmDealStage> stages) {
```
akilarootcode, line 89: check why we need this protected method

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmDealServiceImplV2.java
```diff
@@ -33,10 +46,57 @@ public class CrmDealServiceImplV2 implements CrmDealServiceV2 {
...
+	@Override
+	@Transactional(readOnly = true)
+	public ResponseEntityDto getBoardInitData() {
+		log.info("getBoardInitData: execution started");
+
+		List<CrmDealStage> visibleStages = filterVisibleDealStages(
+				crmDealStageDao.findAllByIsDeletedFalseOrderByOrderIndexAsc());
+		List<CrmBoardStageResponseDto> stages = crmMapper.crmDealStagesToCrmBoardStageResponseDtos(visibleStages);
+
+		List<CrmBoardContactResponseDtoV2> contacts = crmContactDao.findAllContactsForBoardInit()
```
akilarootcode: get from db query

backend/src/main/java/com/skapp/community/crmplanner/type/CrmTaskSort.java
```diff
@@ -0,0 +1,9 @@
...
+
+	DUE_AT,
```
akilarootcode: check how sort behavior

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmTaskRepositoryImpl.java
```diff
@@ -261,83 +260,91 @@ public Page<CrmTaskResponseDtoV2> findCompletedTasksV2(Long ownerId, CrmTaskComp
 		return new PageImpl<>(content, pageable, total);
 	}
 
+	private List<Order> buildTaskOrder(CriteriaBuilder cb, Root<CrmTask> task, CrmTaskSort sortKey,
+			Sort.Direction sortOrder) {
+		Sort.Direction direction = sortOrder == null ? Sort.Direction.ASC : sortOrder;
+		CrmTaskSort key = sortKey == null ? CrmTaskSort.DUE_AT : sortKey;
```
akilarootcode: remove null and set default sort

akilarootcode (changes requested):

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmDealServiceImplV2.java
YasinduRC: Removed the stream

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmDealServiceImplV2.java
YasinduRC: Removed the stream

backend/src/main/java/com/skapp/community/crmplanner/type/CrmTaskSort.java
YasinduRC: Added the db column to the enum

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmTaskRepositoryImpl.java
YasinduRC: default sort was added and remove the null check

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmDealServiceImplV2.java
YasinduRC: This method is a placeholder for enterprise deal stage visibility filtering

akilarootcode (approved):

erandi-rootcode (approved):

PR #1796, opened 2026-08-18
refactor: add company management components and functionalities

dulsara-manakal:
### Summary

# CRM v2 — Company

Rebuilds the CRM **Company** feature against the new v2 type + store foundation.

This is the first domain to be migrated. It is **additive and inert** — nothing in the
running application changes.

---

## Why

The CRM module's existing state layer keeps entities as denormalized arrays and looks
them up by linear search, with request/response types spread across several loosely
related files. v2 replaces this with:

- **A normalized Zustand store** — entities held as `Record<id, Entity>`, with ordered
  id arrays describing what each view shows.
- **Strict, relaxed-by-design types** — one entity type serves list rows, detail
  responses and create/edit bodies, so there is no request/response type sprawl.
- **A consistent API layer** — a standalone request function above each react-query
  hook, one typed params object per call, and v2 endpoints preferred wherever the
  backend provides them.

Company is the pilot. Contacts, deals and tasks follow the same shape.

---

## Does this affect existing flows?

**No. This PR cannot change current behaviour.**

- All new code lives under `src/community/crm/v2/`. Nothing outside that folder imports
  from it.
- `pages/community/crm/companies.tsx` **still imports v1** and is functionally
  unchanged. The only edit to it is a commented-out block at the bottom containing the
  future v2 version of the page, so cutover is delete-and-uncomment rather than a
  rewrite.
- v1 (`src/community/crm/*`) is untouched — no file outside `v2/` was modified.
- `src/enterprise/crm` (the tier limit-guard layer) is untouched. v2 consumes
  `useCrmLimitGuard` and `crmLimitationQueryKeys` exactly as v1 does.
- v2 uses its own store instance (`useCrmStoreV2`), separate from the live v1 store.

The v2 code is compiled and type-checked but never mounted. It becomes live in a later
cutover PR that repoints the four CRM page imports.

---

## Scope

**Included**

| Area | Detail |
|---|---|
| Company table | Paginated list with metrics, debounced search, infinite scroll, row click opens the side panel |
| Add company | Modal + form, Formik validation, duplicate-name check |
| Edit company | Modal + form, sends only changed fields |
| Delete company | Confirmation modal, removes the company from the store |
| Company side panel | Company metadata, contact details, metric cards, edit/delete actions |

**Deliberately excluded** — these are separate PRs:

- Task, deal and contact sections inside the company side panel. The tab strip renders
  as a shell with no content behind it; the sections attach later.
- Contact-only company endpoints (`company/lookup`, `company/search-by-domain`). They
  are consumed solely by the contact form, so they belong to the contacts PR.
- Page cutover, and deletion of v1.

---

## Architecture

**Data flow** — API returns the raw payload, the store is the single source of truth,
components read only from the store:

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR stands up a parallel `crm/v2` tree (API layer, normalized Zustand-backed company table, side panel, and add/edit/delete modals) while leaving the live page on v1 behind a commented-out cutover block. The port is a faithful copy in most places, but I found two hard data-contract defects (the nested `metrics` payload keys don't match the backend, and `CrmCompanyEntity` has no `metrics` property at all, so the code doesn't type-check under `strict`), a missing React Query invalidation strategy that lets deleted/created rows resurface, and a cutover comment that references two v2 files that don't exist. There is also substantial verbatim duplication of v1 components and zero test coverage for the new pure utilities.

Found **31** new issue(s): 🔴 7 important, 🟡 16 suggestion(s), 🟣 8 nit(s)

frontend/src/community/crm/v2/components/organisms/CompanyTable/CompanyTable.tsx
```diff
@@ -0,0 +1,208 @@
...
+        ),
+        contactNumber: (
+          <div className="flex items-baseline">
+            {formatTableValue(company.contactNumber, PHONE_NUMBER_PREFIX)}
+          </div>
+        ),
+        openTasksCount: (
+          <div className="flex flex-row items-center gap-2">
+            {formatTableValue(metrics?.openTasksCount)}
+            {metrics?.overdue !== undefined && metrics.overdue > 0 && (
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Metric field names don't match the backend contract, so three UI elements will always be empty. The backend `CrmCompanyMetrics` type (backend/.../crmplanner/type/CrmCompanyMetrics.java) serialises `openTasksCount`, `overdueTasksCount`, `openValue`, `accountValue`, `openDealsCount`, `closedDealsCount` — this is confirmed by CrmCompanyControllerV2IntegrationTest, which asserts `metrics.openDealsCount`, `metrics.closedDealsCount` and `metrics.overdueTasksCount`. The frontend `CrmCompanyMetrics` interface (v2/types/CrmCommonTypes.ts:28) declares `overdue`, `openDeals`, `closedDeals` instead — those names came from the v1 flat DTO (`CrmCompanyMetricsResponseDto`), not the v2 nested one. Result: `metrics?.overdue` here is permanently `undefined` (overdue badge never renders), `metrics?.closedDeals` on line 149 never renders the "Deals closed" sub-label, and `companyUtil.ts:42/47` make the Open deals / Closed deals side-panel metric cards permanently blank. Fix by renaming the fields in the v2 `CrmCompanyMetrics` interface to `overdueTasksCount`, `openDealsCount`, `closedDealsCount` and updating the four read sites. Worth also correcting the same three stale names on `CrmCompanyEntity` so nobody re-copies them.

frontend/src/community/crm/v2/components/organisms/CompanySidePanel/CompanySidePanel.tsx
```diff
@@ -0,0 +1,195 @@
...
+
+  const isLoading = isCompanyLoading || isMetricsLoading;
+
+  useEffect(() => {
+    if (!fetchedCompany || !fetchedMetrics) return;
+
+    setCompanies(
+      updateCompany(companies, companyId, {
+        ...fetchedCompany,
+        metrics: fetchedMetrics
```
ThinuwanW, line 83: 🤖 **Claude** · 🔴 **Important**

`CrmCompanyEntity` (v2/types/CrmCommonTypes.ts:10-26) has no `metrics` property, so this object literal, plus `company.metrics` in CompanyTable.tsx:112 and `company.metrics?.*` in companyUtil.ts:36/42/47, are type errors under `strict: true`. They only slip through CI because next.config.mjs sets `typescript.ignoreBuildErrors: true` and there is no `tsc --noEmit` script — so the whole feature ships with a broken type contract and red squiggles in every editor. Add `metrics?: CrmCompanyMetrics;` to `CrmCompanyEntity`. While there, the six flat metric fields on that same interface (`openTasksCount`, `overdue`, `openValue`, `accountValue`, `openDeals`, `closedDeals`) are now dead — the v2 endpoint returns them nested — and should be removed so the entity has one source of truth.

frontend/src/community/crm/v2/api/CompanyApi.ts
```diff
@@ -0,0 +1,183 @@
...
+export const useCreateCompany = (
+  onSuccess: (company: CrmCompanyEntity) => void,
+  onError: (error: AxiosError) => void
+): UseMutationResult<CrmCompanyEntity, AxiosError, CrmCompanyEntity> => {
+  const queryClient = useQueryClient();
+
+  return useMutation({
+    mutationFn: createCompany,
+    onSuccess: (createdCompany) => {
+      queryClient.invalidateQueries({
```
ThinuwanW, line 121: 🤖 **Claude** · 🔴 **Important**

None of the three mutations invalidate `crmCompanyQueryKeys`; only the enterprise limitation key is invalidated. v1 invalidated `companyQueryKeys.GET_COMPANY_DATA` on create and delete. Because `CompanyTable`'s effect re-derives `companyIds` from the (now stale) cached pages, this produces visible bugs: (1) create a company, then scroll to trigger `fetchNextPage` — `data` changes, the effect runs, and `setCompanyIds(normalized.companyIds)` rebuilds the list from cached pages that predate the creation, so the just-created row silently disappears; (2) delete a company, navigate away and back — the cached first page still contains it, the effect writes it back into the store, and the deleted row is clickable until the background refetch lands (clicking it hits `/company/{id}` and 404s). Also note `crmCompanyQueryKeys.ALL` is declared but never used — it is exactly the key you want here. Add `queryClient.invalidateQueries({ queryKey: crmCompanyQueryKeys.ALL })` to `useCreateCompany` and `useDeleteCompany`, and in `useEditCompany` (which currently has no `queryClient` at all) invalidate `ALL` or at minimum `DETAIL(id)` + `METRICS(id)` + `LIST`.

frontend/pages/community/crm/companies.tsx
```diff
@@ -62,3 +62,73 @@ const Companies: NextPage = () => {
...
+ * The full v2 page is below. At cutover, delete everything above and
+ * uncomment this — no reassembly needed.
+ *
+ * import { NextPage } from "next";
+ *
+ * import ContentLayout from "~community/common/components/templates/ContentLayout/ContentLayout";
+ * import { Modules } from "~community/common/enums/CommonEnums";
+ * import { useTranslator } from "~community/common/hooks/useTranslator";
+ * import { IconName } from "~community/common/types/IconTypes";
+ * import SidePanelWrapper from "~community/crm/v2/components/atoms/SidePanelWrapper/SidePanelWrapper";
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The comment promises "At cutover, delete everything above and uncomment this — no reassembly needed", but the block imports two modules that do not exist in this PR: `~community/crm/v2/components/atoms/SidePanelWrapper/SidePanelWrapper` (line 76) and `~community/crm/v2/components/organisms/TaskModalController/TaskModalController` (line 80). Neither has a v2 counterpart — only the v1 files exist. Uncommenting would fail to resolve immediately, so the block has already drifted from reality. Beyond the broken imports, parking a full alternate page implementation in a comment is dead code that no compiler, linter, or test can protect: prefer gating the two implementations behind the existing `NEXT_PUBLIC_MODE`-style env check or a feature flag so both paths type-check, or keep the v2 page on its own branch. If the comment stays for now, at least add the two missing v2 components (or point those imports at the v1 paths) so it is actually accurate.

frontend/src/community/crm/v2/regex/crmRegexPatterns.ts
```diff
@@ -0,0 +1,3 @@
+export function isValidCompanyWebsiteUrl(): RegExp {
+  return /^(https:\/\/)?(www\.)?[a-z0-9-]{1,63}(\.[a-z0-9-]{1,63}){0,9}\.[a-z]{2,63}(\/[^\s?#]*)?$/i;
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`isValidCompanyWebsiteUrl` makes the `https://` prefix optional, but the backend rejects anything without an https scheme: `CrmValidations.validateWebsite` does `new URI(website)` and throws `CRM_ERROR_WEBSITE_INVALID` unless `uri.getScheme()` equals https. So a user who types `acme.com` — which this regex accepts, and which the v1 test suite explicitly asserts is valid — passes client validation and then gets a 400 surfaced as the generic "Failed to add the company. Please try again" toast, with no indication that the scheme is the problem. This is inherited from v1, but the PR re-establishes it in a new file so it is worth fixing here: either make `https://` mandatory in the pattern (and update the placeholder/validation copy), or normalise the value in `getTrimmedCompanyValues` by prefixing `https://` when no scheme is present.

frontend/src/community/crm/v2/components/organisms/CompanyTable/CompanyTable.tsx
```diff
@@ -0,0 +1,208 @@
...
+
+  const { data, fetchNextPage, hasNextPage, isFetchingNextPage, isLoading } =
+    useGetCompaniesInfinite(companyFilters);
+
+  const hasSearchTerm = debouncedSearch !== "";
+
+  useEffect(() => {
+    if (!data) return;
+
+    const items = data.pages.flatMap((page) => page.items);
```
ThinuwanW, line 67: 🤖 **Claude** · 🟡 **Suggestion**

Null-safety that v1 had was dropped in the port. v1 used `data?.pages.flatMap((page) => page?.items ?? [])`; this version does `data.pages.flatMap((page) => page.items)`. `fetchCompanies` returns `response?.data?.results?.[0]`, which is typed as always present but is `undefined` if the envelope ever comes back empty — then `page.items` throws, `normalizeCompanies` iterates `undefined`, and the effect crashes outside any try/catch (also `getNextPageParam` would throw on `lastPage.currentPage`). Restore `page?.items ?? []`. Same class of issue at lines 111-112: `const company = companies[id]` is dereferenced (`company.name`, `company.metrics`) with no guard, so any future divergence between `companyIds` and `companies` becomes a render crash rather than a skipped row — a `if (!company) return null`-style filter or `companies[id]?.` would be cheap insurance.

frontend/src/community/crm/v2/components/organisms/CompanySidePanel/CompanySidePanel.tsx
```diff
@@ -0,0 +1,195 @@
...
+        ) : (
+          <CompanySidePanelHeaderActions
+            isCrmSalesManager={Boolean(isCrmSalesManager)}
+            menuItems={menuItems}
+          />
+        )
+      }
+    >
+      <div className="flex flex-col pb-4 gap-4">
+        {isLoading || !company ? (
```
ThinuwanW, line 172: 🤖 **Claude** · 🟡 **Suggestion**

No error state for either query. If `useGetCompanyById` or `useGetCompanyMetrics` fails (e.g. the company was deleted in another tab, or by the stale-cache path described in the CompanyApi finding), `isLoading` goes false but `company` stays undefined, so the body renders `CompanySidePanelSkeleton` forever while the header renders an empty `<h2>` — a permanently loading panel with no way to tell what went wrong. Destructure `isError` from both hooks and render an error/empty state (or fire an error toast and `handleClose()`), consistent with the toast pattern used in the modal contents.

frontend/src/community/crm/v2/api/CompanyApi.ts
```diff
@@ -0,0 +1,183 @@
...
+  );
+  return response?.data?.results?.[0];
+};
+
+export const useGetCompanyById = (
+  id: number,
+  enabled?: boolean
+): UseQueryResult<CrmCompanyEntity> =>
+  useQuery({
+    queryKey: crmCompanyQueryKeys.DETAIL(id),
```
ThinuwanW, line 63: 🤖 **Claude** · 🟡 **Suggestion**

`useGetCompanyById` and `useGetCompanyMetrics` set neither `staleTime` nor `refetchOnWindowFocus: false`, unlike `useGetCompaniesInfinite` just above (which sets `refetchOnWindowFocus: false`). There is no `defaultOptions.queries` config in TanStackProvider, so both queries refetch on every mount and every window focus. Two consequences: every tab-back fires two extra requests per open side panel, and because `EditCompanyModalContent` derives `initialValues` from the store record these queries write into and uses `enableReinitialize: true`, a refetch that returns changed server values will reset the open edit form and discard whatever the user has typed, with no warning. Add `refetchOnWindowFocus: false` (and a small `staleTime`) to both hooks for parity with the list query.

frontend/src/community/crm/v2/components/organisms/CompanySidePanel/CompanySidePanel.tsx
```diff
@@ -0,0 +1,195 @@
...
+  useEffect(() => {
+    if (!fetchedCompany || !fetchedMetrics) return;
+
+    setCompanies(
+      updateCompany(companies, companyId, {
+        ...fetchedCompany,
+        metrics: fetchedMetrics
+      })
+    );
+  }, [fetchedCompany, fetchedMetrics]);
```
ThinuwanW, line 86: 🤖 **Claude** · 🟡 **Suggestion**

The dependency array omits `companies` and `companyId`, both of which the effect reads. It works today only because the effect happens to run in the same commit that the fresh data arrives, so the captured `companies` is current — but this is load-bearing lint suppression: adding the missing deps (as `react-hooks/exhaustive-deps` will tell you to) turns it into a render loop, because the effect calls `setCompanies`, which changes `companies`. CompanyTable.tsx:72 has the identical shape (`companies` read, only `[data]` in deps). Make the read explicit instead of implicit: either pull the current record inside the effect via `useCrmStoreV2.getState().companies`, or better, add a store action (e.g. `upsertCompany(id, fields)` / `mergeCompanies(record)`) that does the merge inside `set()`. Both effects then depend only on the fetched data and correctness stops depending on a suppressed lint rule.

frontend/src/community/crm/v2/components/organisms/CompanySidePanel/CompanySidePanel.tsx
```diff
@@ -0,0 +1,195 @@
...
+          <CompanySidePanelSkeleton />
+        ) : (
+          <>
+            <SidePanelCompanyHeader company={company} />
+
+            <SidePanelMetricCards
+              metrics={getCompanyMetricItems(company, translateText)}
+            />
+            <div className="flex flex-col pt-2 w-full">
+              <Tabs
```
ThinuwanW, line 182: 🤖 **Claude** · 🟡 **Suggestion**

The `Tabs` control is rendered and `activeTab` is tracked, but nothing consumes it — v1 had a `renderTabContent()` switch below the tabs rendering `SidePanelTasksSection` / `SidePanelDealSection` / `SidePanelCompanyContacts`. As written, clicking Tasks/Deals/Contacts changes the highlight and renders nothing, which reads as a broken panel. If the tab sections are deliberately deferred to a follow-up PR, say so in a TODO here; otherwise the tabs should not ship until there is content behind them.

frontend/src/community/crm/v2/components/atoms/SkeletonShape/SkeletonShape.tsx
```diff
@@ -0,0 +1,17 @@
+import { FC } from "react";
```
ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Large-scale verbatim duplication. I diffed the v2 tree against `src/community/crm/**`: SkeletonShape, SidePanelHeaderInfoItem, SidePanelHeaderSkeleton, SidePanelHeaderActionsSkeleton, SidePanelTabsSkeleton, CrmDeleteModalContent, CompanySidePanelSkeleton and CompanySidePanelHeaderActions are byte-identical apart from the props-interface name and the `~community/crm/` → `~community/crm/v2/` import prefix; `companyValidations.ts`, `isValidCompanyWebsiteUrl` and `useGetIndustryOptions` are also exact copies. None of these have any v1/v2 API coupling — they are pure presentation/validation. Every future fix (including the https-scheme fix above and the `circle` inconsistency in SidePanelTabsSkeleton) now has to be applied twice, and one copy will inevitably be missed. Move the version-agnostic pieces to a shared location (e.g. `src/community/crm/shared/` or `~community/common/components`) and have both trees import them; keep `v2/` for things that genuinely differ (api, store, types, table, side panel).

frontend/src/community/crm/v2/components/molecules/SidePanelMetricCards/SidePanelMetricCards.tsx
```diff
@@ -0,0 +1,33 @@
...
+        <Card
+          key={metric.id}
+          className="flex flex-col gap-1 flex-1 min-w-0 p-3 overflow-hidden border border-secondary-accent rounded-lg bg-white"
+        >
+          <p className="body2 text-secondary-text">{metric.title}</p>
+          <div className="flex items-center gap-2">
+            <p className="subtitle2">
+              {metric.isCurrency
+                ? formatMonetaryValueWithDecimals(metric.amount)
+                : metric.amount}
```
ThinuwanW, line 24: 🤖 **Claude** · 🟡 **Suggestion**

Non-currency metrics render blank rather than `0` when the value is missing. `CrmMetricItem.amount` is optional and `getCompanyMetricItems` passes `company.metrics?.openDeals` / `?.closedDeals` straight through, so `undefined` renders as an empty `<p>`. v1 normalised with `String(company.openDeals ?? 0)`, so an empty card read "0". Either default in `getCompanyMetricItems` (`?? 0`) or route non-currency values through `formatTableValue` so they fall back to the `EMPTY_PLACEHOLDER` used everywhere else in the table.

frontend/src/community/crm/v2/utils/companyUtil.ts
```diff
@@ -0,0 +1,130 @@
...
+  address: company?.address ?? "",
+  contactNumber: company?.contactNumber ?? ""
+});
+
+export const getTrimmedCompanyValues = (
+  values: CrmCompanyEntity
+): CrmCompanyEntity => ({
+  name: values.name?.trim(),
+  industry: values.industry,
+  website: values.website?.trim(),
```
ThinuwanW, line 98: 🤖 **Claude** · 🟡 **Suggestion**

Cleared optional fields are now submitted as `""` where v1 submitted `null` (`values.website?.trim() || null`). The backend `CrmCompanyEditDto` models these as `JsonNullable<String>` precisely so it can distinguish "absent" from "explicitly cleared", and `validateWebsite`/`validateAddress`/`validateContactNumber` early-return on blank — so `""` is accepted and persisted as an empty string rather than SQL NULL. New and edited rows will therefore store `''` while existing rows store NULL, which is a mismatch any `IS NULL` query or domain search (`findCompaniesByWebsiteDomain`) will trip over. Map blanks back to `null` in `getTrimmedCompanyValues` (or in `getChangedCompanyFields`) as v1 did, and widen the entity field types to `string | null` to match.

frontend/src/community/crm/v2/components/molecules/CompanyModalForm/CompanyModalForm.tsx
```diff
@@ -0,0 +1,160 @@
...
+import { SEARCH_DEBOUNCE_DELAY } from "~community/common/constants/commonConstants";
+import { characterLengths } from "~community/common/constants/stringConstants";
+import useDebounce from "~community/common/hooks/useDebounce";
+import { TranslatorFunctionType } from "~community/common/types/CommonTypes";
+import { useCheckCompanyNameExists } from "~community/crm/v2/api/CompanyApi";
+import useGetIndustryOptions from "~community/crm/v2/hooks/useGetIndustryOptions";
+import { CrmCompanyEntity } from "~community/crm/v2/types/CrmCommonTypes";
+
+interface CompanyModalFormProps {
+  formik: FormikProps<CrmCompanyEntity>;
```
ThinuwanW, line 20: 🤖 **Claude** · 🟡 **Suggestion**

`CrmCompanyEntity` is being used as three different things: the API response type, the Formik value type, and the create/edit request payload type. Because it is the response shape, every field is optional and it also carries `id`, `openTasksCount`, `contactIds`, `dealIds`, `taskIds`. That weakens the form contract (hence the defensive `values.name?.trim() ?? ""` on line 46 for a field the Yup schema marks `required`) and means nothing type-prevents an unrelated field from being POSTed. v1 had a dedicated `CrmCompanyFormTypes` for exactly this. Introduce a `CrmCompanyFormValues` (required `name`/`industry`, optional `website`/`address`/`contactNumber`) plus a `CrmCompanyCreatePayload`/`CrmCompanyEditPayload`, and keep `CrmCompanyEntity` for responses only. Also worth exporting `EditCompanyVariables` from CompanyApi.ts so `EditCompanyModalContent` can annotate `editCompanyVariables` instead of relying on structural inference.

frontend/src/community/crm/v2/utils/companyUtil.ts
```diff
@@ -0,0 +1,130 @@
+import { TranslatorFunctionType } from "~community/common/types/CommonTypes";
```
ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

No tests anywhere in the new v2 tree. The v1 module it was copied from has `utils/__test__/companyUtil.test.ts` and `regex/crmRegexPatterns.test.ts`; the v2 copies of both files ship with no coverage. Everything added here is trivially testable pure logic: `normalizeCompanies` (id-undefined skip), `updateCompany`/`removeCompany` (immutability), `getChangedCompanyFields` (the diffing that decides what gets PATCHed), `getCompanyFormInitialValues`/`getTrimmedCompanyValues`, and all three formatters in `commonUtil.ts` (whose zero/blank/sub-$1 edge cases are the subject of two other comments here). Given the whole tree is currently unreachable from any page and so gets no manual QA either, unit tests on the utils are the only thing that will catch a regression before cutover. Component tests should use the existing `MockTheme` + `QueryClientProvider` wrapper pattern.

frontend/src/community/crm/v2/hooks/useGetIndustryOptions.ts
```diff
@@ -0,0 +1,24 @@
...
+import { CrmIndustryEnum } from "~community/crm/v2/enums/common";
+
+const useGetIndustryOptions = () => {
+  const translateIndustryOptions = useTranslator(
+    "crmModule",
+    "companies",
+    "industryOptions"
+  );
+
+  return useMemo(
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

This `useMemo` never hits. `useTranslator` returns a brand-new closure on every render (it builds and returns an arrow function), so `[translateIndustryOptions]` changes every render and the 21-entry options array is rebuilt each time and handed to `Dropdown` as a fresh reference. Same problem with `menuItems` in CompanySidePanel.tsx:96-128 (`[translateText]`). Either drop the `useMemo` (it is buying nothing and the deps array is misleading) or memoise on something stable — e.g. take `t` and `i18n.language` from `useTranslation` directly, or wrap the returned function in `useCallback` inside `useTranslator`.

frontend/src/community/crm/v2/components/molecules/CompanyModalForm/CompanyModalForm.tsx
```diff
@@ -0,0 +1,160 @@
...
+  const trimmedName = values.name?.trim() ?? "";
+  const trimmedOriginalName = originalName?.trim();
+  const debouncedName = useDebounce(trimmedName, SEARCH_DEBOUNCE_DELAY);
+
+  const { data: companyNameData } = useCheckCompanyNameExists(
+    debouncedName,
+    debouncedName.length > 0 && debouncedName !== trimmedOriginalName
+  );
+
+  const isAlreadyNameExists =
```
ThinuwanW, line 70: 🤖 **Claude** · 🟣 **Nit**

`isAlreadyNameExists` mixes the live `trimmedName` with `companyNameData`, which is the result for the *debounced* name. During the 500 ms debounce window after the user edits a name that already existed, `companyNameData.isExists` is still `true` from the previous query while `trimmedName` is the new value — so a false "A company with this name already exists" error shows and Save stays disabled until the debounce fires. Guard on the values matching, e.g. only trust the result when `trimmedName === debouncedName`.

frontend/src/community/crm/v2/utils/commonUtil.ts
```diff
@@ -0,0 +1,18 @@
...
+  EMPTY_PLACEHOLDER
+} from "~community/crm/v2/constants/commonConstants";
+
+const isEmptyValue = (value?: string | number) =>
+  value === undefined || Number(value) === 0;
+
+export const formatTableValue = (value?: string | number, prefix = "") =>
+  isEmptyValue(value) ? EMPTY_PLACEHOLDER : `${prefix}${value}`;
+
+export const formatMonetaryValue = (value?: string) =>
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`formatMonetaryValue` splits on `.` before the emptiness check, so any amount under $1 is treated as no-data: `"0.50"` → `"0"` → `Number("0") === 0` → `"-"`. v1 checked `value == 0` on the full value first and rendered `"$0"`. Not a big deal for pipeline totals, but a company with a single $0.75 open deal will show "-" (no data) rather than a small value. Test emptiness on the original value, then truncate.

frontend/src/community/crm/v2/components/molecules/CrmDeleteModalContent/CrmDeleteModalContent.tsx
```diff
@@ -0,0 +1,57 @@
...
+  confirmLabel,
+  cancelLabel,
+  onConfirm,
+  onClose
+}) => {
+  return (
+    <div className="flex flex-col">
+      <div className="body1">{description}</div>
+      <div className="flex flex-row justify-end py-[0.85rem] gap-[1rem]">
+        <ButtonV2
```
ThinuwanW, line 29: 🤖 **Claude** · 🟣 **Nit**

Neither button gets an `aria-label`, even though `crmModule.companies.deleteCompanyModal.ariaLabels.confirm` ("Confirm Delete Company") and `.cancel` are already defined in the locale file and `CompanyModalForm` labels its equivalent buttons. Since the visible labels here are just "Delete"/"Cancel" with no context, a screen-reader user gets no indication of what is being deleted. Add optional `confirmAriaLabel`/`cancelAriaLabel` props and pass those existing keys from `DeleteCompanyModalContent`.

frontend/src/community/crm/v2/components/molecules/SidePanelSkeleton/SidePanelTabsSkeleton.tsx
```diff
@@ -0,0 +1,111 @@
...
+          </div>
+        ))}
+      </div>
+      <hr className="border-secondary-accent" />
+    </div>
+
+    <div className="border border-secondary-accent rounded-lg divide-y divide-secondary-accent w-full overflow-hidden">
+      <div className="flex items-center justify-between px-3 py-3">
+        <div className="flex items-center gap-4">
+          <SkeletonShape circle className="h-6 w-6 shrink-0" />
```
ThinuwanW, line 40: 🤖 **Claude** · 🟣 **Nit**

The first skeleton row uses `<SkeletonShape circle .../>` for its leading shape while rows 2-4 use a non-circle shape with otherwise identical classes (lines 56, 72, 88) — almost certainly a copy-paste slip, and it will render one rounded and three square placeholders. Since all four rows are byte-identical otherwise, generate them with `Array.from({ length: 4 })` and fix the shape once. (Carried over from the v1 file, but this is a new file so it is the moment to fix it.)

frontend/src/community/crm/v2/utils/companyUtil.ts
```diff
@@ -0,0 +1,130 @@
...
+    amount: company.metrics?.openDeals
+  },
+  {
+    id: "closedDeals",
+    title: translateText(["metrics", "closedDeals"]),
+    amount: company.metrics?.closedDeals
+  }
+];
+
+export const getSelectedCompany = (
```
ThinuwanW, line 51: 🤖 **Claude** · 🟣 **Nit**

Minor type-explicitness gaps against the strict-mode convention: `normalizeCompanies` (line 8), `getSelectedCompany` (line 51) and `removeCompany` (line 62) have no declared return types while their neighbours in the same file do. Also `CrmMetricItem` (line 22) is a type declared in a `utils/` file and imported as such by SidePanelMetricCards — per the module layout it belongs in `v2/types/`. Related: CompanySidePanel.tsx:183 uses a `tabId as CrmSidePanelTabEnum` assertion to narrow the tab id; if `TabItem['id']` is `string`, a small type guard over `Object.values(CrmSidePanelTabEnum)` avoids the cast. Finally, hardcoded `text-black` (SidePanelHeaderInfoItem.tsx:33, CompanySidePanel.tsx:154) and `bg-white` (SidePanelMetricCards.tsx:17) should use theme tokens like the `text-primary-text` / `bg-tertiary-background` used elsewhere in these same files.

frontend/src/community/crm/v2/api/utils/QueryKeys.ts
```diff
@@ -0,0 +1,11 @@
+import { CrmCompanyFilterRequest } from "~community/crm/v2/types/CrmTypes";
+
+const CRM_COMPANIES = "crm-companies";
```
ThinuwanW, line 7: 🤖 **Claude** · 🔴 **Important**

The new key namespace root `"crm-companies"` is already claimed by the v1 tree, which is still live: `companyQueryKeys.CRM_COMPANIES(size)` = ["crm-companies", size] (actively used by ContactApi.ts:87 for the company lookup infinite query) and `contactQueryKeys.CRM_COMPANIES` = ["crm-companies"]. React Query matches invalidation by key prefix, so `crmCompanyQueryKeys.ALL` will also blow away the v1 company-lookup cache, and any v1 invalidation of ["crm-companies"] will wipe the new v2 list/detail/metrics entries. This is latent only because nothing currently invalidates `crmCompanyQueryKeys` at all — the moment that gap is fixed the collision becomes live. Namespace the v2 root distinctly (e.g. `const CRM_COMPANIES = "crm-companies-v2"`) for as long as both trees coexist.

frontend/src/community/crm/v2/components/organisms/CompanyTable/CompanyTable.tsx
```diff
@@ -0,0 +1,208 @@
...
+      id: "accountValue",
+      label: translateText(["table", "columns", "accountValueHeader"]),
+      width: "20%",
+      align: "right"
+    }
+  ];
+
+  const transformToTableRows = (): GridRow[] =>
+    companyIds.map((id) => {
+      const company = companies[id];
```
ThinuwanW, line 111: 🤖 **Claude** · 🔴 **Important**

`companies[id]` is dereferenced without a guard, then `company.name`, `company.contactNumber` and `company.metrics` are read. `companyIds` lives in the UI slice and `companies` in the data slice, and they are written by two separate `set()` calls in four different places (this effect, AddCompanyModalContent, DeleteCompanyModalContent, CompanySidePanel's effect) — several of them from render-captured snapshots. Any ordering where an id list is committed without its matching entity produces `TypeError: Cannot read properties of undefined (reading 'name')`, which unmounts the whole table. v1 mapped over the company array directly so this class of desync was structurally impossible. Either derive the rows defensively (`companyIds.map(id => companies[id]).filter(Boolean).map(...)`) or expose a single store action that sets `companies` and `companyIds` atomically in one `set()`. Related: the `setCompanies({ ...companies, ...normalized.companies })` merge never evicts entries, so the entity map also grows unbounded across searches.

frontend/src/community/crm/v2/components/organisms/CompanySidePanel/CompanySidePanel.tsx
```diff
@@ -0,0 +1,195 @@
...
+  const isOpen =
+    isCrmSidePanelOpen &&
+    crmSidePanelType === CrmSidePanelTypes.COMPANY_SIDE_PANEL;
+
+  const handleClose = () => {
+    setSelectedCompanyId(null);
+    closeCrmSidePanel();
+  };
+
+  const menuItems: MenuItemProps[] = useMemo(
```
ThinuwanW, line 99: 🤖 **Claude** · 🟡 **Suggestion**

This PR introduces a second source of truth for "which company". The panel now receives `companyId` as a prop and fetches/renders that company, but the Edit and Delete menu items only flip modal flags — EditCompanyModalContent and DeleteCompanyModalContent both resolve their target from `store.selectedCompanyId`. They agree today only because the (commented-out) page happens to pass `selectedCompanyId` as `companyId`. Any future caller that opens the panel for a different company (e.g. a company link from a deal or contact panel) will edit or hard-delete the wrong record with no visible cue. Either call `setSelectedCompanyId(companyId)` inside the menu `onClick` handlers, or drop the prop and read `selectedCompanyId` from the store as v1 did.

frontend/src/community/crm/v2/components/organisms/CompanySidePanel/CompanySidePanelSkeleton.tsx
```diff
@@ -0,0 +1,53 @@
...
+      </div>
+    </div>
+
+    <div className="flex gap-4 w-full">
+      <div className="flex-1 border border-secondary-accent rounded-lg p-3 flex flex-col gap-1">
+        <SkeletonShape className="h-2.5 w-14" />
+        <div className="flex items-center gap-2">
+          <SkeletonShape className="h-3 w-11" />
+          <div className="flex items-center gap-1 bg-secondary-accent rounded-full px-3 py-1">
+            <SkeletonShape className="h-3 w-3" />
```
ThinuwanW, line 34: 🤖 **Claude** · 🟡 **Suggestion**

The skeleton was copied verbatim from v1 and still renders the rounded chip placeholder (`bg-secondary-accent rounded-full px-3 py-1` with an icon + label shape) inside the first metric card, but v2's SidePanelMetricCards dropped `MetricChipLabel` and the new `CrmMetricItem` type has no `chip` field. So the loading state promises a chip that the loaded state never renders, causing a visible layout shift on every side-panel open. Either remove the chip placeholder from the skeleton or restore chip support in `CrmMetricItem` + SidePanelMetricCards if the chip is still part of the design.

frontend/pages/community/crm/companies.tsx
```diff
@@ -62,3 +62,73 @@ const Companies: NextPage = () => {
 };
 
 export default Companies;
+
```
ThinuwanW, line 117: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

I grepped `pages/` and `src/` for `crm/v2` imports: the only hits are inside the commented block and inside the v2 tree itself. That means roughly 1,100 lines of new production code (18 components, the API layer, utils, validations) have zero runtime entry point — it cannot be smoke-tested, cannot be exercised by any test, and lint/tsc will not surface the runtime issues in it. A comment block is also the one form of staging that guarantees drift (imports go stale, as the missing SidePanelWrapper/TaskModalController already show). Prefer gating the cutover on an env/feature flag so both page variants stay compiled and at least one is reachable in a non-prod environment.

frontend/src/community/crm/v2/utils/companyUtil.ts
```diff
@@ -0,0 +1,130 @@
...
+): CrmCompanyEntity => ({
+  name: values.name?.trim(),
+  industry: values.industry,
+  website: values.website?.trim(),
+  address: values.address?.trim(),
+  contactNumber: values.contactNumber?.trim()
+});
+
+export const getChangedCompanyFields = (
+  initialValues: CrmCompanyEntity,
```
ThinuwanW, line 104: 🤖 **Claude** · 🟡 **Suggestion**

`getChangedCompanyFields` can legitimately return `{}`, and nothing guards against submitting it. Formik's `dirty` compares untrimmed values, so typing a trailing space in the name enables Save while the trimmed diff is empty — the app then fires `PATCH /crm/company/{id}` with an empty body. The backend skips every field and returns 200, so the user gets a "Company updated successfully" toast for a request that changed nothing. Add an early return in `submitEditCompany` when `Object.keys(payload).length === 0` (close the modal without a toast, or without the network call).

frontend/src/community/crm/v2/components/molecules/SidePanelCompanyHeader/SidePanelCompanyHeader.tsx
```diff
@@ -0,0 +1,86 @@
...
+            <BrowserIcon
+              width="20"
+              height="20"
+              fill="var(--color-secondary-icon)"
+            />
+          }
+          value={website}
+          endIcon={IconName.POP_OUT_ICON}
+          onClick={() => openInNewTab(website)}
+        />
```
ThinuwanW, line 44: 🤖 **Claude** · 🟡 **Suggestion**

Separate from the frontend/backend validation mismatch: because the new regex makes the `https://` prefix optional, a stored website can be scheme-less (`example.com`). `openInNewTab` calls `window.open(url, "_blank", ...)` with no normalization, and a scheme-less string is resolved against the current origin — the new tab navigates to `/community/crm/example.com` (a 404 inside the app) instead of the external site. Normalize before opening (prepend `https://` when `!/^https?:\/\//i.test(website)`), and/or make the scheme mandatory in `isValidCompanyWebsiteUrl` so stored values always carry one.

frontend/src/community/crm/v2/utils/companyUtil.ts
```diff
@@ -0,0 +1,130 @@
...
+      companies[company.id] = company;
+      companyIds.push(company.id);
+    }
+  });
+
+  return { companies, companyIds };
+};
+
+export interface CrmMetricItem {
+  id: string;
```
ThinuwanW, line 23: 🤖 **Claude** · 🟡 **Suggestion**

Type declarations are placed outside the module's `types/` directory: `CrmMetricItem` is exported from `utils/companyUtil.ts` and `EditCompanyVariables` is declared inside `api/CompanyApi.ts`. The v2 module already has the right homes for both (`types/CrmCommonTypes.ts` for entities/view models, `types/CrmTypes.ts` for request/response shapes — which is where `CrmCompanyFilterRequest` and friends live), and v1 kept the equivalent `MetricItem` in `types/CommonTypes`. The consequence is that a presentational component (SidePanelMetricCards) now imports its prop type from a utils module. Move both declarations into `types/`.

frontend/src/community/crm/v2/api/utils/ApiEndpoints.ts
```diff
@@ -0,0 +1,15 @@
+import { moduleAPIPath } from "~community/common/constants/configs";
+
+export const crmCompanyEndpointsV2 = {
+  GET_COMPANIES: `${moduleAPIPath.CRM}/company`
+};
+
+export const crmCompanyEndpointsV1 = {
+  GET_COMPANY_BY_ID: (id: number) => `${moduleAPIPath.CRM}/company/${id}`,
+  GET_COMPANY_METRICS: (id: number) =>
```
ThinuwanW, line 24: 🤖 **Claude** · 🟣 **Nit**

Name reuse with different semantics across the two coexisting trees makes these easy to mix up: `crmCompanyEndpointsV1.GET_COMPANY_METRICS(id)` is the single-company metrics endpoint, while the still-live `companyEndpoints.GET_COMPANY_METRICS` is the paginated list endpoint; likewise `useGetCompanyMetrics` is exported from both `crm/api/CompanyApi` (infinite list) and `crm/v2/api/CompanyApi` (one company). Rename the v2 pair to `GET_COMPANY_METRICS_BY_ID` / `useGetCompanyMetricsById` to match the backend method name (`getCompanyMetricsById`) and remove the ambiguity.

frontend/src/community/crm/v2/api/CompanyApi.ts
```diff
@@ -0,0 +1,183 @@
...
+  params: CrmCompanyFilterRequest
+): Promise<CrmCompanyListResponse> => {
+  const response = await authFetchV2.get(crmCompanyEndpointsV2.GET_COMPANIES, {
+    params
+  });
+  return response?.data?.results?.[0];
+};
+
+export const useGetCompaniesInfinite = (
+  params: CrmCompanyFilterRequest,
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`useGetCompaniesInfinite` is the only hook in this file without an explicit return type — the other four are annotated `UseQueryResult<...>` / `UseMutationResult<...>`, so this one silently relies on inference despite the strict-mode convention; annotate it `UseInfiniteQueryResult<...>`. Its second parameter `enabled?: boolean` is also dead: the sole call site (CompanyTable.tsx:60) never passes it, so `enabled: undefined` is forwarded to `useInfiniteQuery`. Either drop the parameter or pass a real condition from the caller.

frontend/src/community/crm/v2/components/organisms/CompanyTable/CompanyTable.tsx
dulsara-manakal: Fixed this.

frontend/src/community/crm/v2/components/organisms/CompanySidePanel/CompanySidePanel.tsx
dulsara-manakal: This comment is actually stale.

frontend/src/community/crm/v2/api/CompanyApi.ts
dulsara-manakal: No need to invalidate

frontend/src/community/crm/v2/regex/crmRegexPatterns.ts
dulsara-manakal: Regex already handles that. `https://` required.

frontend/src/community/crm/v2/components/molecules/CompanyModalForm/CompanyModalForm.tsx
dulsara-manakal: Fixed this

frontend/src/community/crm/v2/utils/commonUtil.ts
dulsara-manakal: That's the requirement.

frontend/src/community/crm/v2/api/utils/QueryKeys.ts
dulsara-manakal: No need to add a prefix for v2. The import path is correct.

frontend/pages/community/crm/companies.tsx
dulsara-manakal: Remove this approach and this will handle with a hook.

frontend/src/community/crm/v2/components/organisms/CompanyTable/CompanyTable.tsx
dulsara-manakal: Invalid, and that fallback doesnt make any impact with new v2 types.

frontend/src/community/crm/v2/api/CompanyApi.ts
dulsara-manakal: No need to add custom stale time, and also that metrics endpoint should always be up-to-date.

frontend/src/community/crm/v2/components/organisms/CompanySidePanel/CompanySidePanel.tsx
dulsara-manakal: The forever-skeleton path isn't reachable: the panel renders companies[companyId] from the store, and the table has already populated that record before the panel can open.

frontend/src/community/crm/v2/components/organisms/CompanySidePanel/CompanySidePanel.tsx
dulsara-manakal: Will add it .

frontend/src/community/crm/v2/components/atoms/SkeletonShape/SkeletonShape.tsx
dulsara-manakal: This is what we discussed, so this duplication is a known one.

frontend/src/community/crm/v2/components/organisms/CompanySidePanel/CompanySidePanel.tsx
dulsara-manakal: Added.

frontend/src/community/crm/v2/components/molecules/SidePanelMetricCards/SidePanelMetricCards.tsx
dulsara-manakal: If it is undefined, it showss the EMPTY_PLACEHOLDER, which is fine with the requirements.

frontend/src/community/crm/v2/utils/companyUtil.ts
dulsara-manakal: Added

frontend/src/community/crm/v2/hooks/useGetIndustryOptions.ts
dulsara-manakal: useMemo is for values we are using, not to keep translation keys.

frontend/src/community/crm/v2/components/molecules/SidePanelSkeleton/SidePanelTabsSkeleton.tsx
dulsara-manakal: This is the finalized design

frontend/src/community/crm/v2/components/molecules/CrmDeleteModalContent/CrmDeleteModalContent.tsx
dulsara-manakal: Added

frontend/src/community/crm/v2/utils/companyUtil.ts
dulsara-manakal: This won't happen. Invalid comment.

frontend/src/community/crm/v2/components/molecules/CompanyModalForm/CompanyModalForm.tsx
dulsara-manakal: We keep a relaxed type so the entity is well-defined, therefore, there is no need for separate types.

frontend/src/community/crm/v2/components/organisms/CompanySidePanel/CompanySidePanelSkeleton.tsx
dulsara-manakal: That is the design.

frontend/pages/community/crm/companies.tsx
dulsara-manakal: This is a known thing and will be fixed after refactoring.

frontend/src/community/crm/v2/components/molecules/SidePanelCompanyHeader/SidePanelCompanyHeader.tsx
dulsara-manakal: Fixed this.

frontend/src/community/crm/v2/components/organisms/CompanySidePanel/CompanySidePanel.tsx
dulsara-manakal: No need for it, and the current approach is fine.

frontend/src/community/crm/v2/components/organisms/CompanySidePanel/CompanySidePanel.tsx
dulsara-manakal: We pass the ID to panel through prop because there can be stale IDs, but when edit and delete are happening, it wont happen because id should be there to edit or delete.

frontend/src/community/crm/v2/utils/companyUtil.ts
dulsara-manakal: Added missing types and cannot change the colors of in design.

frontend/src/community/crm/v2/components/organisms/CompanyTable/CompanyTable.tsx
dulsara-manakal: Checked all four writers: every one sets companies and companyIds together from the same normalized result, and the side panel only touches companies, so an orphan id isn't reachable in the current code.

frontend/src/community/crm/v2/utils/companyUtil.ts
dulsara-manakal: Added a guard

frontend/src/community/crm/v2/api/utils/ApiEndpoints.ts
dulsara-manakal: After refactoring, this won't get confused, so keep it as it is.

frontend/src/community/crm/v2/utils/companyUtil.ts
dulsara-manakal: Neither type exists in v2/types, so moving them would mean adding new types to those files, which we're deliberately not doing.

frontend/src/community/crm/v2/api/CompanyApi.ts
dulsara-manakal: Added the UseInfiniteQueryResult annotation. On the dead param, same issue was on useGetCompanyById and useGetCompanyMetrics, so I dropped enabled from all three.

frontend/src/community/crm/v2/api/CompanyApi.ts
```diff
@@ -0,0 +1,176 @@
...
+  CrmCompanyMetrics
+} from "~community/crm/v2/types/CrmCommonTypes";
+import {
+  CrmCompanyFilterRequest,
+  CrmCompanyListResponse,
+  CrmExistsResponse
+} from "~community/crm/v2/types/CrmTypes";
+import { crmLimitationQueryKeys } from "~enterprise/crm/api/utils/QueryKeys";
+
+interface EditCompanyVariables {
```
hasalarootcode: as remember, we discussed something for this, check with @anusham-rc

frontend/src/community/crm/v2/components/molecules/SidePanelHeaderInfoItem/SidePanelHeaderInfoItem.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { IconName } from "~community/common/types/IconTypes";
+
+interface SidePanelHeaderInfoItemProps {
+  icon: ReactElement;
+  value: string;
+  endIcon?: IconName;
+  onClick?: () => void;
+}
+
+const SidePanelHeaderInfoItem: FC<SidePanelHeaderInfoItemProps> = ({
```
hasalarootcode, line 13: this is not an atom

frontend/src/community/crm/v2/components/templates/SidePanelWrapper/SidePanelWrapper.tsx
```diff
@@ -0,0 +1,27 @@
+import { FC, ReactNode, useEffect } from "react";
+
+import { ZIndexEnums } from "~community/common/enums/CommonEnums";
+
+interface SidePanelWrapperProps {
+  children: ReactNode;
+  isOpen?: boolean;
+}
+
+const SidePanelWrapper: FC<SidePanelWrapperProps> = ({
```
hasalarootcode, line 10: Wrappers should be templates

frontend/src/community/crm/v2/constants/commonConstants.ts
```diff
@@ -0,0 +1,5 @@
+export const PHONE_NUMBER_PREFIX = "+";
+export const CURRENCY_PREFIX = "$";
+export const EMPTY_PLACEHOLDER = "-";
```
hasalarootcode: remove these constants 
export const PHONE_NUMBER_PREFIX = "+";
export const CURRENCY_PREFIX = "$";
export const EMPTY_PLACEHOLDER = "-";

frontend/src/community/crm/v2/hooks/useGetIndustryOptions.ts
```diff
@@ -0,0 +1,24 @@
+import { useMemo } from "react";
+
+import { useTranslator } from "~community/common/hooks/useTranslator";
+import { CrmIndustryEnum } from "~community/crm/v2/enums/common";
+
+const useGetIndustryOptions = () => {
```
hasalarootcode: remove this hook

hasalarootcode (changes requested):

frontend/src/community/crm/v2/components/molecules/SidePanelHeaderInfoItem/SidePanelHeaderInfoItem.tsx
dulsara-manakal: Moved to molecule level.

frontend/src/community/crm/v2/api/CompanyApi.ts
dulsara-manakal: Removed this and access the id from the payload it self.

frontend/src/community/crm/v2/components/templates/SidePanelWrapper/SidePanelWrapper.tsx
dulsara-manakal: Moved to template

frontend/src/community/crm/v2/constants/commonConstants.ts
dulsara-manakal: Removed them

frontend/src/community/crm/v2/hooks/useGetIndustryOptions.ts
dulsara-manakal: removed this

hasalarootcode (dismissed):

akilarootcode (dismissed):

akilarootcode (approved):

hasalarootcode (approved):

PR #1798, opened 2026-08-18
feat: implement batch retrieval of deals by ids 

YasinduRC:
### Summary

Adds a batch lookup so the CRM v2 client can hydrate its deal store in one request instead of fetching each deal by id. Backend only.

**`POST /v1/crm/deal/batch`** — body `{ "ids": [1, 2, 3] }`, responds with the matching deals as a bare array under `results`.

- Returns the new **`CrmDealListItemDtoV2`**, which carries related records as ids (`stageId`, `ownerId`, `companyId`, `contactId`) rather than nested objects — the client resolves them against entities it already holds.
- Unknown ids, soft-deleted deals and — for a sales representative — other owners' deals are simply **absent** from the response rather than raising an error, so one bad id in a batch cannot fail the whole request.
- An empty `ids` array short-circuits to an empty list without hitting the database.
- A soft-deleted company is reported as **no company at all** rather than a dangling `companyId`, via a `LEFT JOIN ... ON isDeleted = false` in the projection.

**Mounted on the v1 controller deliberately.** The v2 frontend API files import the default `authFetch` client, whose `baseURL` is `/v1` — a `/v2` route would 404 for them despite living in a `v2` folder. It delegates to `CrmDealServiceV2`.

### Breaking changes

None. `CrmDealResponseDtoV2` and every existing deal endpoint are untouched — this is purely additive.

### How to test

1. `cd backend && mvn test -Dtest='CrmDealControllerV2IntegrationTest'` — 16 tests, all green.
2. `POST /v1/crm/deal/batch` with `{"ids":[<two real deal ids>]}` → each result carries `stageId` / `ownerId` / `companyId` / `contactId` and **no** nested `stage` / `owner` / `company` / `contact`.
3. `{"ids":[<real id>, 999999]}` → one result, status 200, no error.
4. `{"ids":[]}` → 200 with an empty `results`.
5. As a sales representative, request a deal owned by someone else → omitted from `results`.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
- [x] Code is formatted with `npm run format` *(backend: `spring-javaformat:validate` clean)*

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information

**Base branch:** raise against `refactor/crm-716-provider-implementation-v2`, not `develop` — this branch is a single commit on top of it, so targeting develop would pull in all 17 provider commits.

Tests live in `CrmDealControllerV2IntegrationTest` (16 total, 6 new) even though the route sits on the v1 controller, because the endpoint is served by `CrmDealServiceV2` and returns the v2 id-reference shape. `CrmDealControllerIntegrationTest` is untouched.

**Known follow-up:** `CrmDealListItemDtoV2` duplicates what `CrmDealResponseDtoV2` will become once the v2 task/deal refactor flattens it. They should be merged into one DTO at that point — kept separate here so this PR stays additive and does not drag the task response changes in.


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds a `POST /v1/crm/deal/batch` endpoint that projects deals as id-references for client-store hydration, backed by a new Criteria query, DTO, and six integration tests. The query itself is correct — I verified that `stage`/`owner`/`contact` are all `nullable = false` on `CrmDeal` (so the implicit path dereferences cannot drop rows), that a live deal can never carry a soft-deleted contact or stage (contact deletion cascades the soft delete to its deals; a stage cannot be deleted while non-deleted deals reference it), so masking only the company is the correct and complete set, and that `ResponseEntityDto.putToResults` flattens the list so the `['results']` assertions are right. Two issues stand out: the v2-shaped endpoint is mounted on the v1 controller (which now injects both the v1 and v2 services), and the client-controlled `ids` list has no size bound or validation of any kind.

Found **14** new issue(s): 🔴 3 important, 🟡 5 suggestion(s), 🟣 6 nit(s)

backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmDealController.java
```diff
@@ -29,6 +31,19 @@ public class CrmDealController {
 
 	private final CrmDealService crmDealService;
 
+	private final CrmDealServiceV2 crmDealServiceV2;
+
+	@Operation(summary = "Get deals by ids",
+			description = "Returns the base details of the deals matching the given ids, used to hydrate the "
+					+ "client's deal store. Related records are carried as id references only. Unknown, deleted "
+					+ "and - for a sales representative - other owners' deals are omitted.")
+	@PostMapping("/batch")
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

A v2-shaped endpoint is being mounted on the v1 controller. `CrmDealControllerV2` already exists at `/v2/crm/deal` and already injects `CrmDealServiceV2` (as its `crmDealService` field), so this route belongs there as `POST /v2/crm/deal/batch`. As written, three conventions break at once: (1) `/v1/crm/deal/batch` returns `CrmDealListItemDtoV2` while every other `/v1/crm/deal/*` endpoint delegates to the v1 `CrmDealService` and returns v1 payloads — v1 consumers now see two different response vocabularies under one base path; (2) the v1 controller now injects two service interfaces for the same aggregate (`crmDealService` + `crmDealServiceV2`), which is the exact mixing the v1/v2 controller split exists to prevent; (3) the endpoint is effectively unversioned relative to its payload. The test's justification ("so the v2 clients reach it through their /v1 base URL") is a frontend base-URL concern — fix that on the client (point the deal store at the v2 base) rather than mis-versioning the route. Fix: move `getDealsByIds` verbatim into `CrmDealControllerV2` as `@PostMapping("/batch")`, delete the `crmDealServiceV2` field and its two imports from this class, and update `BATCH_PATH` in the test to `BASE_PATH + "/batch"`.

backend/src/main/java/com/skapp/community/crmplanner/payload/request/CrmDealIdsRequestDto.java
```diff
@@ -0,0 +1,14 @@
...
+import lombok.Getter;
+import lombok.Setter;
+
+import java.util.List;
+
+@Getter
+@Setter
+public class CrmDealBatchRequestDto {
+
+	private List<Long> ids;
```
ThinuwanW, line 12: 🤖 **Claude** · 🔴 **Important**

`ids` is a fully client-controlled, unbounded collection with no validation anywhere — no bean validation annotations on the DTO, no `@Valid` on the controller's `@RequestBody`, and no service-layer check beyond the null/empty guard. It flows straight into `deal.get(CrmDeal_.id).in(dealIds)`. Three concrete consequences: a request with, say, 500k ids produces a single `IN (...)` list that will blow past MySQL's `max_allowed_packet` / statement-parse limits and surface as a 500 (the `DataAccessException` handler); the response size is likewise unbounded, unlike `getDeals` which at least funnels through `PageRequest`; and `{"ids": [null, 1]}` pushes a null element into `in(...)`, whose Hibernate literal handling is at best undefined here. Fix, following the module's own convention (service-layer validation via `CrmValidations` + `ModuleException` + a message constant, as `validateDealStageReorderRequest` does): add `CrmConstants.DEAL_BATCH_MAX_IDS` (e.g. 200 — enough for a page-load hydration), a `CrmMessageConstant.CRM_ERROR_DEAL_BATCH_IDS_INVALID` key plus its entry in `crm-messages.properties`, and a `CrmValidations.validateDealBatchIds(List<Long> ids)` that rejects a size over the cap and any null element; call it from `getDealsByIds` before hitting the DAO. If you prefer bean validation instead, annotate the field `@Size(max = 200) private List<@NotNull Long> ids;` and add `@Valid` to the `@RequestBody` parameter — but note that requires `@Valid` to actually be wired, which no other CRM controller currently does.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmDealControllerV2IntegrationTest.java
```diff
@@ -349,4 +357,119 @@ void createDeal_WithoutCrmRole_ReturnsForbidden() throws Exception {
...
+		CrmContact contact = savedContact(company, "deal.batch.scope.v2@example.com");
+		CrmDeal deal = savedDeal("Admin Owned Deal", stage, company, contact, 1L);
+
+		authToken = jwtService.generateAccessToken(userDetailsService.loadUserByUsername("user2@gmail.com"), 1L);
+
+		performBatchRequest(List.of(deal.getId())).andDo(print())
+			.andExpect(status().isOk())
+			.andExpect(jsonPath(STATUS_PATH).value(STATUS_SUCCESSFUL))
+			.andExpect(jsonPath("['results']").isEmpty());
+	}
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The six new tests cover the happy path, unknown ids, soft-deleted deals, deleted-company masking, empty ids, and sales-rep owner scoping — but not the endpoint's primary security gate. Every other route in this class has a companion `*_WithoutCrmRole_ReturnsForbidden` test (see `createDeal_WithoutCrmRole_ReturnsForbidden`), and this is a brand-new route, so `@PreAuthorize("hasAnyRole('ROLE_CRM_SALES_REPRESENTATIVE')")` on it is currently unasserted; a future edit to that annotation would not fail any test. Add `getDealsByIds_WithoutCrmRole_ReturnsForbidden` that re-issues the token for `user2@gmail.com` (whose CRM role is left unset, as in the existing forbidden test) and expects `status().isForbidden()`. Also worth adding a case that posts `{}` (i.e. `ids` absent) — the `requestDto.getIds() == null` branch in `CrmDealServiceImplV2.getDealsByIds` is currently never exercised, only the empty-list branch is.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmDealServiceImplV2.java
```diff
@@ -58,6 +63,30 @@ public ResponseEntityDto getDeals(CrmDealFilterDto filterDto) {
...
+	 * Hydrates the client's deal store for ids it already holds references to. Unknown,
+	 * soft deleted and - for a sales representative - other owners' deals are simply
+	 * absent from the response rather than reported as errors.
+	 */
+	@Override
+	@Transactional(readOnly = true)
+	public ResponseEntityDto getDealsByIds(CrmDealBatchRequestDto requestDto) {
+		log.info("getDealsByIds: execution started");
+
+		if (requestDto.getIds() == null || requestDto.getIds().isEmpty()) {
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

The null/empty ids guard is duplicated: this early return and the identical `if (dealIds == null || dealIds.isEmpty())` check at the top of `CrmDealRepositoryImpl.findDealsByIdsV2`. The repository guard matches the pattern in `findClosedDealSummaryByContactIds` and `countDealsByStageIds`, so it should stay; the service-level one is what's redundant, and it also means the null-ids path silently skips `userService.getCurrentUser()` (harmless today, but it makes the two branches behaviourally asymmetric). Consider dropping the service guard and letting the repository return `Collections.emptyList()` — the response shape is identical either way. If you keep it (e.g. to avoid the `getCurrentUser` round trip on empty input), that's a defensible call; just note it's intentional.

backend/src/main/java/com/skapp/community/crmplanner/payload/response/v2/CrmDealListItemDtoV2.java
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * A deal carrying id references only, so the client resolves them against the entities it
+ * already holds instead of receiving another copy nested inside every deal. Used by the
+ * batch lookup, which exists to hydrate a client store that is keyed by id.
+ */
+@Getter
+@Setter
+@NoArgsConstructor
+@AllArgsConstructor
+public class CrmDealListItemDtoV2 {
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Naming: the class's own Javadoc says it is "Used by the batch lookup", and its field set is id-references-only, but the name says "ListItem" — which reads as the item type of the paginated `getDeals` list (that role is filled by `CrmDealResponseDtoV2`). Something like `CrmDealRefDtoV2` or `CrmDealBatchItemDtoV2` would describe what it actually is. Separately, note the heavy overlap with the existing `CrmDealByStageItemResponseDto` (id, name, amount, ownerId, companyId, contactId, priority) — this new DTO is that set plus description/orderIndex/closingAt/stageId, minus taskCount. Not identical, so consolidation isn't clearly right, but worth a look before a third near-duplicate appears.

backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmDealController.java
```diff
@@ -29,6 +31,19 @@ public class CrmDealController {
...
+	@PreAuthorize("hasAnyRole('ROLE_CRM_SALES_REPRESENTATIVE')")
+	public ResponseEntity<ResponseEntityDto> getDealsByIds(@RequestBody CrmDealBatchRequestDto requestDto) {
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Pre-existing gap this new endpoint newly exposes: `GlobalExceptionHandler` has no `@ExceptionHandler(HttpMessageNotReadableException.class)`, so a malformed body — e.g. `{"ids": "abc"}` or `{"ids": ["abc"]}`, both easy for a client to produce against a typed `List<Long>` — falls through to the catch-all `@ExceptionHandler(Exception.class)` and returns 500 with Jackson's parse message (which echoes a snippet of the submitted input) in `results[0].message`. This already affects `createDeal`/`editDeal`, so it isn't introduced here, but a collection-typed body widens the surface. Consider adding a `HttpMessageNotReadableException` handler that maps to 400 with `COMMON_ERROR_VALIDATION_ERROR` and a fixed message rather than `e.getMessage()`.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmDealServiceImplV2.java
```diff
@@ -58,6 +63,30 @@ public ResponseEntityDto getDeals(CrmDealFilterDto filterDto) {
...
+			return new ResponseEntityDto(false, Collections.emptyList());
+		}
+
+		User currentUser = userService.getCurrentUser();
+		Long ownerId = CrmUtil.isCrmSalesRepresentative(currentUser) ? currentUser.getEmployee().getEmployeeId() : null;
+
+		List<CrmDealListItemDtoV2> deals = crmDealDao.findDealsByIdsV2(requestDto.getIds(), ownerId);
+
+		log.info("getDealsByIds: execution ended with {} result(s)", deals.size());
+		return new ResponseEntityDto(false, deals);
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Response envelope shape differs from every other deal endpoint. `ResponseEntityDto.putToResults` unwraps any `Collection` into `results` (see `ResponseEntityDto:45-55`), so this endpoint emits `results: [deal, deal, ...]` — a flat array of deals. Every other v2 deal endpoint emits a single-element `results`: `getDeals` gives `results: [{items, currentPage, totalItems, totalPages}]` and `getDealById`/`createDeal`/`editDeal` give `results: [{deal}]`. Any shared frontend unwrapper that reads `results[0]` will get the *first deal* here rather than the payload, and a caller cannot distinguish a 1-deal response from a single-object response. The new tests hardcode `['results'][0]['id']` precisely because of this, which confirms the divergence. Fix: wrap the list in a small response object (e.g. `new ResponseEntityDto(false, new CrmDealBatchResponseDtoV2(deals))`) or reuse the `PageDto`-style `items` wrapper so `results[0]` is consistently the payload across the deal API. If the flat-array shape is deliberate, document it explicitly on the `@Operation` and confirm the client unwrapper handles it.

backend/src/main/java/com/skapp/community/crmplanner/payload/response/v2/CrmDealListItemDtoV2.java
```diff
@@ -0,0 +1,44 @@
+package com.skapp.community.crmplanner.payload.response.v2;
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Duplicates an existing id-reference deal DTO and its masking behaviour. `CrmDealByStageItemResponseDto` (payload/response/board/) already carries id, name, amount, ownerId, companyId, contactId, priority — 7 of this class's 11 fields — and `CrmUtil.toDealByStageItemDto` already implements the exact same rule (`if (hasDeletedCompany(deal)) dto.setCompanyId(null)`), only in Java rather than in the join's ON clause. The codebase now has two id-reference deal shapes and two independent implementations of deleted-company masking that can drift apart. Fix: either extend/reuse `CrmDealByStageItemResponseDto` (adding description/orderIndex/closingAt/stageId, leaving taskCount null for the batch path) or extract a single shared id-reference deal DTO used by both the board and the batch lookup, so the masking rule lives in exactly one place.

backend/src/main/java/com/skapp/community/crmplanner/payload/response/v2/CrmDealListItemDtoV2.java
```diff
@@ -0,0 +1,44 @@
...
+
+	private CrmDealPriority priority;
+
+	private String orderIndex;
+
+	private String amount;
+
+	private LocalDateTime closingAt;
+
+	private Long stageId;
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Fragile positional projection: `cb.construct(CrmDealListItemDtoV2.class, ...)` in `CrmDealRepositoryImpl.findDealsByIdsV2` binds to the Lombok `@AllArgsConstructor`, whose parameter order is the field declaration order. The last four fields — stageId, ownerId, companyId, contactId — are all `Long`, so reordering or inserting a field here still compiles and still passes constructor resolution, but silently swaps values across the wire (e.g. a deal's ownerId reported as its stageId). Nothing in the build catches it; only the one happy-path integration test would. Fix: declare an explicit projection constructor on this DTO annotated with a comment tying it to the criteria query, or drop `@AllArgsConstructor` in favour of a hand-written constructor so any field reorder becomes a compile error rather than a data-corruption bug.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmDealControllerV2IntegrationTest.java
```diff
@@ -349,4 +357,119 @@ void createDeal_WithoutCrmRole_ReturnsForbidden() throws Exception {
...
+	void getDealsByIds_EmptyIds_ReturnsEmptyList() throws Exception {
+		performBatchRequest(List.of()).andDo(print())
+			.andExpect(status().isOk())
+			.andExpect(jsonPath(STATUS_PATH).value(STATUS_SUCCESSFUL))
+			.andExpect(jsonPath("['results']").isEmpty());
+	}
+
+	@Test
+	@DisplayName("Get deals by ids as Sales Representative - Omits another owner's deal")
+	void getDealsByIds_SalesRepRequestingOthersDeal_OmitsIt() throws Exception {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The owner-scoping predicate's positive branch is never exercised. This test only asserts the negative — a sales rep asking for someone else's deal gets nothing. No test has a sales rep request their OWN deal and get it back. Every one of the six new tests would still pass if `predicates.add(cb.equal(deal.get(CrmDeal_.owner).get(Employee_.employeeId), ownerId))` in `CrmDealRepositoryImpl:189` were inverted or made unsatisfiable, because the happy-path tests all run as user1 (which takes the `ownerId == null` branch). Fix: in this test, also create a deal owned by employee 2 and assert the sales rep receives exactly that one deal while the admin-owned deal is absent — a single request asserting both halves of the filter.

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmDealControllerV2IntegrationTest.java
```diff
@@ -349,4 +357,119 @@ void createDeal_WithoutCrmRole_ReturnsForbidden() throws Exception {
...
+	void getDealsByIds_HappyPath_ReturnsIdReferences() throws Exception {
+		CrmDealStage stage = savedStage();
+		CrmCompany company = savedCompany("Batch V2 Corp");
+		CrmContact contact = savedContact(company, "deal.batch.v2@example.com");
+		CrmDeal deal = savedDeal("Batch Deal V2", stage, company, contact, 1L);
+
+		performBatchRequest(List.of(deal.getId())).andDo(print())
+			.andExpect(status().isOk())
+			.andExpect(jsonPath(STATUS_PATH).value(STATUS_SUCCESSFUL))
+			.andExpect(jsonPath("['results'].length()").value(1))
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The six new tests hardcode `"['results']"` and `"['results'][0]"` as string literals while the same file already imports and uses `RESULTS_0_PATH` from `com.skapp.support.TestConstants` throughout the pre-existing tests, and `TestConstants` also defines an unused `RESULTS_PATH` for the bare-array case. Fix: use `RESULTS_0_PATH + "['id']"`, `RESULTS_PATH + ".length()"`, `jsonPath(RESULTS_PATH).isEmpty()` etc. so the batch tests stay consistent with the rest of the suite and survive any change to the envelope path.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmDealServiceImplV2.java
```diff
@@ -58,6 +63,30 @@ public ResponseEntityDto getDeals(CrmDealFilterDto filterDto) {
...
+
+		if (requestDto.getIds() == null || requestDto.getIds().isEmpty()) {
+			log.info("getDealsByIds: no ids provided, returning empty list");
+			return new ResponseEntityDto(false, Collections.emptyList());
+		}
+
+		User currentUser = userService.getCurrentUser();
+		Long ownerId = CrmUtil.isCrmSalesRepresentative(currentUser) ? currentUser.getEmployee().getEmployeeId() : null;
+
+		List<CrmDealListItemDtoV2> deals = crmDealDao.findDealsByIdsV2(requestDto.getIds(), ownerId);
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Authorization asymmetry with the single-deal read on the same service. `getDealById` (line 101-103) throws `CrmMessageConstant.CRM_ERROR_DEAL_VIEW_DENIED` (HTTP 400) when a sales rep requests another owner's deal, while this path silently omits it. The Javadoc explains the choice, so this is a deliberate call, but it means the same authorization decision produces two different observable outcomes on the same resource, and a client hydrating its store cannot tell 'not permitted' from 'deleted' from 'never existed'. Worth confirming the frontend doesn't need to distinguish these — if it does, consider returning the omitted ids alongside the results rather than silently dropping them.

backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmDealController.java
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

POST used for a pure read, diverging from the codebase's existing by-ids lookup convention. `EpUserController` (enterprise/common/controller/v1) exposes the equivalent 'fetch these entities by id' operation as `@GetMapping` with `@RequestParam(value = "employeeIds", required = false) List<Long>` and no wrapper request DTO. A POST here makes the response uncacheable and non-idempotent in the eyes of proxies/clients for what is a read. If the id list is expected to be short (which it should be, given it comes from a client store), prefer `@GetMapping("/batch")` with a repeated `ids` param and drop `CrmDealBatchRequestDto` entirely; if POST is kept because the list can be long, note that reason in the `@Operation` description.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmDealRepositoryImpl.java
```diff
@@ -157,6 +158,42 @@ public Page<CrmDealResponseDtoV2> findDealsV2(CrmDealFilterDto filterDto, Long o
...
+				deal.get(CrmDeal_.contact).get(CrmContact_.id)));
+
+		List<Predicate> predicates = new ArrayList<>();
+		predicates.add(cb.isFalse(deal.get(CrmDeal_.isDeleted)));
+		predicates.add(deal.get(CrmDeal_.id).in(dealIds));
+		if (ownerId != null) {
+			predicates.add(cb.equal(deal.get(CrmDeal_.owner).get(Employee_.employeeId), ownerId));
+		}
+
+		query.where(predicates.toArray(new Predicate[0])).orderBy(cb.asc(deal.get(CrmDeal_.id)));
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`orderBy(cb.asc(deal.get(CrmDeal_.id)))` adds an ORDER BY to every batch query for a result set that, by this method's own contract, feeds a client store keyed by id — order is not part of the contract and no caller or test asserts it. Meanwhile the sibling `findDealsV2` orders by stage.orderIndex, deal.orderIndex, deal.id, so this ordering is not the module's display order either. Either drop the `orderBy` (the store re-keys anyway) or, if deterministic order is wanted for response stability, say so in the Javadoc so a future reader doesn't mistake it for display ordering.

backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmDealController.java
YasinduRC: Moved to v1 but used the new responseDto for deal

backend/src/main/java/com/skapp/community/crmplanner/payload/request/CrmDealIdsRequestDto.java
YasinduRC: We do not put validations on the dto and CrmValidations.validateDealIds now rejects null/non-positive elements,Also ignored the BATCH_IDS_MAX_COUNT

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmDealServiceImplV2.java
YasinduRC: Remove that guard it is covered by the validation

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmDealControllerV2IntegrationTest.java
YasinduRC: Fixed,Added both getDealsByIds_WithoutCrmRole_ReturnsForbidden and a null-ids test covering

backend/src/main/java/com/skapp/community/crmplanner/payload/response/v2/CrmDealListItemDtoV2.java
YasinduRC: Updated the name as CrmDealResponseDtoV2

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmDealControllerV2IntegrationTest.java
YasinduRC: Fixed,switched all of these over to RESULTS_PATH / RESULTS_0_PATH to match the rest of the file.

backend/src/main/java/com/skapp/community/crmplanner/payload/response/v2/CrmDealListItemDtoV2.java
YasinduRC: The query selects full CrmDeal entities now and maps them through the existing mapper, so there's no positional constructor left to silently get out of sync.

backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmDealController.java
YasinduRC: Out of scope, correctly identified as pre-existing and codebase-wide, not specific to this endpoint.

backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmDealController.java
YasinduRC: This is intentional to use POST for this endpoint

backend/src/test/java/com/skapp/community/crmplanner/controller/v2/CrmDealControllerV2IntegrationTest.java
YasinduRC: Fixed. The test now creates a second deal owned by the sales rep and asserts the batch response contains exactly that one, while the admin-owned deal is omitted.

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmDealServiceImplV2.java
YasinduRC: This is intentional, not an oversight. The batch endpoint's whole job is to hydrate a client store with whatever the caller is actually allowed to see

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmDealRepositoryImpl.java
YasinduRC: Removed the orderBy

backend/src/main/java/com/skapp/community/crmplanner/payload/response/v2/CrmDealListItemDtoV2.java
YasinduRC: The real overlap now is between this and the board's deal DTO, and we're keeping those separate on purpose board needs task counts, this doesn't

backend/src/main/java/com/skapp/community/crmplanner/service/v2/impl/CrmDealServiceImplV2.java
YasinduRC: The concern doesn't hold up in practice  for a single deal, this endpoint's `results: [deal]` is literally identical to what `getDealById` returns. Other non paginated endpoints also return without the wrapper

backend/src/main/java/com/skapp/community/crmplanner/repository/CrmDealRepository.java
```diff
@@ -18,6 +18,8 @@ public interface CrmDealRepository {
 
 	Page<CrmDealResponseDtoV2> findDealsV2(CrmDealFilterDto filterDto, Long ownerId, Pageable pageable);
 
+	List<CrmDeal> findDealsByIds(List<Long> dealIds, Long ownerId);
```
erandi-rootcode: since it's a two param search better to name something like findDealsByDealIdsAndOwner

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
```diff
@@ -387,6 +392,25 @@ public ResponseEntityDto getDealById(Long id) {
...
+	@Transactional(readOnly = true)
+	public ResponseEntityDto getDealsByIds(CrmDealIdsRequestDto requestDto) {
+		log.info("getDealsByIds: execution started");
+
+		CrmValidations.validateDealIds(requestDto.getIds());
+
+		User currentUser = userService.getCurrentUser();
+		Long ownerId = CrmUtil.isCrmSalesRepresentative(currentUser) ? currentUser.getEmployee().getEmployeeId() : null;
+
+		List<CrmDealResponseDtoV2> deals = crmDealDao.findDealsByIds(requestDto.getIds(), ownerId)
```
erandi-rootcode: MapStruct handles collections natively. try that out to see if it fits so that we could omit the stream

erandi-rootcode (changes requested):

backend/src/main/java/com/skapp/community/crmplanner/repository/CrmDealRepository.java
YasinduRC: Kept it as findDealsByIds, countDealsByStageIds right above it also takes an extra ownerId param without naming it, so this matches that existing convention rather than breaking it.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
YasinduRC: The stream isn't pure mapping, CrmUtil.toDealResponseDtoV2 also masks the company when it's soft-deleted. A generated MapStruct collection mapper would only perform the mapping and wouldn't apply this additional logic on its own.

erandi-rootcode (dismissed):

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
akilarootcode: get directly from query

akilarootcode (changes requested):

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealServiceImpl.java
YasinduRC: Fixed and removed the stream

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmDealRepositoryImpl.java
```diff
@@ -157,6 +132,39 @@ public Page<CrmDealResponseDtoV2> findDealsV2(CrmDealFilterDto filterDto, Long o
 		return new PageImpl<>(content, pageable, total);
 	}
 
+	@Override
+	public List<CrmDealResponseDtoV2> findDealsByIds(List<Long> dealIds, Long ownerId) {
+		if (dealIds == null || dealIds.isEmpty()) {
```
akilarootcode: move to service level

akilarootcode (changes requested):

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmDealRepositoryImpl.java
YasinduRC: Moved the validation to service level

akilarootcode (dismissed):

erandi-rootcode (dismissed):

erandi-rootcode (approved):

PR #1799, opened 2026-08-18
Fix/3049 business unit case sensitivity

shakila-kamalasena:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

The PR replaces two derived exists-queries with a `findByNameIgnoreCase` list lookup plus a Java-side `equals()` comparison so that business-unit name uniqueness becomes case-sensitive. The refactor itself is behaviour-preserving for the create/update duplicate checks, but it now permits rows whose names differ only by case, which breaks the remaining `Optional<BusinessUnit> findByName(...)` lookups in the bulk-upload path under MySQL's default case-insensitive collation. The added integration test also runs on H2 (case-sensitive), so it passes against the pre-fix code and does not actually guard the new behaviour; the update path's exclude-id branch is untested.

Found **12** new issue(s): 🔴 3 important, 🟡 7 suggestion(s), 🟣 2 nit(s)

backend/src/main/java/com/skapp/community/common/repository/BusinessUnitDao.java
```diff
@@ -14,8 +14,6 @@ public interface BusinessUnitDao extends JpaRepository<BusinessUnit, Long> {
 
 	Optional<BusinessUnit> findByName(String name);
 
-	boolean existsByName(String name);
-
-	boolean existsByNameAndBusinessUnitIdNot(String name, Long businessUnitId);
+	List<BusinessUnit> findByNameIgnoreCase(String name);
```
ThinuwanW, line 17: 🤖 **Claude** · 🔴 **Important**

`Optional<BusinessUnit> findByName(String name)` (line 15) is still used by the bulk-upload path — PeopleServiceImpl.setBulkEmployeeBusinessUnit (~line 1999) and PeopleServiceImpl.validateBusinessUnitInBulk (~line 2129) — and it becomes unsafe as a direct result of this change. Production runs MySQL 8 (application-prd.yml sets MySQL8Dialect) and `com_business_unit.name` is declared as plain `text` with no explicit collation in people-ddl-script-v1-create-table-com-business-unit.sql, so it inherits the DB default (utf8mb4_0900_ai_ci — case- and accent-insensitive; note another changeset explicitly overrides to utf8mb4_bin, confirming the default is CI). Before this PR, `existsByName` matched case-insensitively at the DB level, so 'Sales' and 'SALES' could never coexist. This PR deliberately allows them to coexist — and once they do, `findByName("Sales")` matches both rows and Spring Data throws IncorrectResultSizeDataAccessException, surfacing as an opaque 500 during employee bulk validation/import instead of a business error. Fix: stop using the single-result `findByName` for name lookups. Either add a shared exact-match helper built on `findByNameIgnoreCase(...)` filtered with `getName().equals(...)` returning `Optional`, or replace `findByName` with an explicit `@Query` using a case-sensitive comparison (e.g. `where binary bu.name = :name`), and update both PeopleServiceImpl call sites (their existing `.filter(unit -> unit.getName().equals(...))` guards the wrong case — it filters the result, but the exception is thrown before the filter ever runs).

backend/src/test/java/com/skapp/community/common/controller/v1/BusinessUnitControllerIntegrationTest.java
```diff
@@ -150,6 +150,22 @@ void createBusinessUnit_withDuplicateName_ReturnsBadRequest() throws Exception {
 						messageUtil.getMessage(CommonMessageConstant.COMMON_ERROR_BUSINESS_UNIT_NAME_ALREADY_EXISTS)));
 		}
 
+		@Test
```
ThinuwanW, line 153: 🤖 **Claude** · 🔴 **Important**

This test does not guard the change it is meant to cover. Integration tests run on H2 (`jdbc:h2:mem:test` in backend/src/test/resources/application.yml), where string comparison is case-sensitive by default. The pre-PR implementation called `existsByName("SALES")`, which on H2 compiles to `where name = 'SALES'` and would not match the seeded 'Sales' — so this test passes against the old code as well and gives false confidence that the case-sensitivity fix works on the production MySQL (case-insensitive collation) where the bug actually existed. Add a test that fails without the fix: a service-level unit test with a mocked `BusinessUnitDao` whose `findByNameIgnoreCase("SALES")` returns a BusinessUnit named 'Sales', asserting creation succeeds (and the inverse, that an exact-case match still throws COMMON_ERROR_BUSINESS_UNIT_NAME_ALREADY_EXISTS). Keep this integration test as a smoke check, but it should not be the only coverage.

backend/src/main/java/com/skapp/community/common/service/impl/BusinessUnitServiceImpl.java
```diff
@@ -185,6 +185,16 @@ private void validateBusinessUnitRequest(BusinessUnitRequestDto businessUnitRequ
 		}
 	}
 
+	private boolean existsByNameCaseSensitive(String name, Long excludeId) {
```
ThinuwanW, line 188: 🤖 **Claude** · 🟡 **Suggestion**

The "case-insensitive fetch, then case-sensitive `equals` compare" logic now exists in three places: this helper, PeopleServiceImpl.setBulkEmployeeBusinessUnit (~1999) and PeopleServiceImpl.validateBusinessUnitInBulk (~2129). That is exactly the kind of duplication the guidelines call out, and it is why the `findByName` bug above is easy to miss. Consolidate into one place — e.g. a `BusinessUnitDao` query with a case-sensitive comparison, or a package-visible `BusinessUnitService` method such as `Optional<BusinessUnit> findByExactName(String name)` that both the config module and the people module reuse. Also consider renaming this helper: `existsByNameCaseSensitive(name, excludeId)` reads like a Dao method and its exclusion semantics are not conveyed by the name (`isDuplicateName(name, excludeId)` would be clearer). The loop itself can be `findByNameIgnoreCase(name).stream().anyMatch(...)`.

backend/src/main/java/com/skapp/community/common/service/impl/BusinessUnitServiceImpl.java
```diff
@@ -40,7 +40,7 @@ public ResponseEntityDto createBusinessUnit(BusinessUnitRequestDto businessUnitR
 
 		String name = businessUnitRequestDto.getName();
 
-		if (businessUnitDao.existsByName(name)) {
+		if (existsByNameCaseSensitive(name, null)) {
```
ThinuwanW, line 43: 🤖 **Claude** · 🟡 **Suggestion**

Uniqueness is enforced purely in application code with a check-then-insert, and `com_business_unit` has no unique index on `name` (see people-ddl-script-v1-create-table-com-business-unit.sql). Two concurrent create requests with the identical name can both pass `existsByNameCaseSensitive` and both insert, producing true duplicates that the service can never resolve. Since this PR is specifically defining the intended uniqueness semantics, this is a good moment to back it with the schema: add a Liquibase changeset changing `name` from `text` to `varchar(<BUSINESS_UNIT_NAME_MAX_LENGTH>)` with an explicit case-sensitive collation (`utf8mb4_bin`) plus a unique index. That both closes the race and makes `findByName` single-result again, which resolves the bulk-upload issue at its root. If a schema change is out of scope for this PR, note it as follow-up.

backend/src/test/java/com/skapp/community/common/controller/v1/BusinessUnitControllerIntegrationTest.java
```diff
@@ -150,6 +150,22 @@ void createBusinessUnit_withDuplicateName_ReturnsBadRequest() throws Exception {
...
+			BusinessUnitRequestDto request = new BusinessUnitRequestDto();
+			request.setName("SALES");
+
+			performRequest(post(BASE_PATH).contentType(MediaType.APPLICATION_JSON)
+				.content(objectMapper.writeValueAsString(request))
+				.accept(MediaType.APPLICATION_JSON)).andDo(print())
+				.andExpect(status().isCreated())
+				.andExpect(jsonPath(STATUS_PATH).value(STATUS_SUCCESSFUL))
+				.andExpect(jsonPath(RESULTS_0_PATH + "['name']").value("SALES"));
+		}
```
ThinuwanW, line 167: 🤖 **Claude** · 🟡 **Suggestion**

The `excludeId` branch of the new helper is untested. `existsByNameCaseSensitive(name, id)` replaced `existsByNameAndBusinessUnitIdNot`, and the UpdateBusinessUnit nested class has no duplicate-name coverage at all. Add: (1) update a unit's name to a different casing of its own current name (e.g. 'Sales' -> 'SALES') expecting 200 — this exercises the self-exclusion path; (2) update a unit to another existing unit's exact name expecting 400 with COMMON_ERROR_BUSINESS_UNIT_NAME_ALREADY_EXISTS; (3) update a unit to a different casing of another unit's name expecting 200.

backend/src/main/java/com/skapp/community/common/service/impl/BusinessUnitServiceImpl.java
ThinuwanW, line 43: 🤖 **Claude** · 🟣 **Nit**

Neither the duplicate check nor the persisted value trims the name, while the frontend validator (frontend/src/community/configurations/utils/businessUnitValidations.ts) applies Yup `.trim()` before its own uniqueness test. So an API client (or the bulk flow) can create 'Sales ' alongside 'Sales'; the UI will then treat them as duplicates and block edits, and bulk upload's exact `equals` match on a trimmed CSV value will not find the padded row. Consider trimming once in `validateBusinessUnitRequest` / at the top of `updateBusinessUnit` and using the trimmed value for both the duplicate check and `setName`, so backend and frontend agree on what counts as the same name.

backend/src/main/java/com/skapp/community/common/repository/BusinessUnitDao.java
```diff
@@ -14,8 +14,6 @@ public interface BusinessUnitDao extends JpaRepository<BusinessUnit, Long> {
 
 	Optional<BusinessUnit> findByName(String name);
```
ThinuwanW, line 15: 🤖 **Claude** · 🔴 **Important**

Multi-row crash (distinct from the general 'bulk path still uses findByName' concern): `findByName` returns `Optional`, which Spring Data implements with a single-result query. Until this PR, `existsByName` guaranteed at most one row per case-folded name, so the Optional could never see two rows. The whole premise of this PR is that MySQL's default collation (utf8mb4_0900_ai_ci) makes `name = ?` case-insensitive — so the moment an admin creates both 'Sales' and 'SALES' (now allowed), `findByName("Sales")` matches TWO rows and Spring Data throws `IncorrectResultSizeDataAccessException` (wrapping Hibernate's `NonUniqueResultException`). Both call sites are on the employee bulk-upload path: `PeopleServiceImpl.validateBusinessUnitInBulk` (line 2129) and `setBulkEmployeeBusinessUnit` (line 1999). Neither is inside a per-row try/catch, so the whole CSV validate/import request fails with an opaque 500 instead of a row-level 'business unit not found' error. Note this is NOT hypothetical-only for MySQL: it is exactly the collation behaviour this PR exists to work around. Fix: change the DAO signature to `List<BusinessUnit> findByName(String name)` (or delete it and reuse `findByNameIgnoreCase`), and resolve the exact match in Java — e.g. a shared `Optional<BusinessUnit> findExactByName(String name)` helper that streams `findByNameIgnoreCase(name)` and filters on `getName().equals(name)`. That also removes the current `findByName(...).filter(equals)` idiom, which is only safe by accident.

backend/src/main/java/com/skapp/community/common/service/impl/BusinessUnitServiceImpl.java
ThinuwanW, line 43: 🤖 **Claude** · 🟡 **Suggestion**

Business unit is now the only named configuration entity in the codebase with case-sensitive uniqueness. Every sibling enforces case-INsensitive uniqueness at the DAO level: `WorkLocationDao.existsByNameIgnoreCaseAndIsDeletedFalse` (used in WorkLocationServiceImpl:63/91/288), `PolicyLeaveTypeDao.existsByNameIgnoreCase[AndIdNot]` (PolicyLeaveTypeServiceImpl:77/113), `LeavePolicyDao.existsByNameIgnoreCaseAndLeaveType_Id...` (LeavePolicyServiceImpl:86/109), `CrmDealStageDao.existsByNameIgnoreCaseAndIsDeletedFalse[AndIdNot]`, `CrmCompanyDao.existsByNameIgnoreCaseAndIsDeletedFalse`, and enterprise `TemplateEnvelopeDao.findByNameIgnoreCase`. After this change an admin can create 'Sales' and 'SALES' side by side; they render as two near-identical rows in the org-configuration list and in every business-unit dropdown, with no way for a user to tell them apart. If the intent really is case-sensitive names, that is a product decision worth stating in the PR description; if the intent was only to stop a *rename* being blocked by its own case variant, the narrower fix is to keep `IgnoreCase` uniqueness and just exclude the entity's own id. Please confirm which, and align the other entities or document the exception.

backend/src/main/java/com/skapp/community/common/service/impl/BusinessUnitServiceImpl.java
```diff
@@ -185,6 +185,16 @@ private void validateBusinessUnitRequest(BusinessUnitRequestDto businessUnitRequ
 		}
 	}
 
+	private boolean existsByNameCaseSensitive(String name, Long excludeId) {
+		for (BusinessUnit businessUnit : businessUnitDao.findByNameIgnoreCase(name)) {
```
ThinuwanW, line 189: 🤖 **Claude** · 🟡 **Suggestion**

The chosen approach cannot be hardened at the DB layer as-is, because `com_business_unit.name` is declared `text NOT NULL` (people-ddl-script-v1-create-table-com-business-unit.sql:7). Two consequences worth calling out: (1) `findByNameIgnoreCase` compiles to `upper(name) = upper(?)` over an unindexable TEXT column, so every create/update does a full scan and materialises entities — acceptable at current table size, but it is a scan by construction, not just by missing index; (2) a `text` column cannot take a unique index in MySQL without a prefix length, so the natural belt-and-braces fix (a unique index enforcing the new case-sensitive semantics) requires a Liquibase changeset that first converts `name` to `varchar(100)` (matching `CommonConstants.BUSINESS_UNIT_NAME_MAX_LENGTH`) with an explicit binary/`_bin` or `_as_cs` collation, then adds `UNIQUE KEY uk_com_business_unit_name (name)`. Doing that would also let the duplicate check go back to a plain derived query and delete this in-memory helper entirely. Please add the changeset under `community/db/changelog/` rather than leaving the invariant enforced only in Java.

backend/src/main/java/com/skapp/community/common/service/impl/BusinessUnitServiceImpl.java
ThinuwanW, line 188: 🤖 **Claude** · 🟡 **Suggestion**

Naming/signature makes the helper easy to misread. `existsByNameCaseSensitive` copies Spring Data's derived-query naming convention for something that is not a repository method, so at the call sites it reads as a DAO call; and the `Long excludeId` parameter is invoked as a bare `null` in `createBusinessUnit` (line 43), which gives no hint at the call site what the second argument means. Rename to something intent-revealing and non-derived-query-shaped, e.g. `isDuplicateBusinessUnitName(String name, Long excludeBusinessUnitId)`, or split into two overloads (`isDuplicateName(name)` delegating to `isDuplicateName(name, null)`) so the create path never passes a literal null. Also consider a short comment explaining WHY the in-Java `equals` is needed (MySQL's case-insensitive default collation) — without it the next reader will 'simplify' this back to a derived `existsByName` and silently reintroduce the bug.

backend/src/main/java/com/skapp/community/common/service/impl/BusinessUnitServiceImpl.java
ThinuwanW, line 43: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Validation is entirely hand-rolled in the service while the DTO carries no bean-validation annotations and the controller does not trigger them. `BusinessUnitRequestDto` has plain `String name` / `String description` with no `@NotBlank` / `@Size`, and `BusinessUnitController.createBusinessUnit` / `updateBusinessUnit` take `@RequestBody BusinessUnitRequestDto` with no `@Valid`. As a result `validateBusinessUnitRequest` (lines 168-186) and the duplicated inline checks in `updateBusinessUnit` (lines 67-79) re-implement what `@NotBlank` and `@Size(max = 100)` would catch — the exact 'custom validation in the service layer that duplicates bean validation' pattern the guidelines call out. This is pre-existing rather than introduced here, but this PR is editing that validation block, so it is the right moment: add `@NotBlank` + `@Size(max = BUSINESS_UNIT_NAME_MAX_LENGTH)` on `name` and `@Size(max = BUSINESS_UNIT_DESCRIPTION_MAX_LENGTH)` on `description`, add `@Valid` on both `@RequestBody` params, and keep only the duplicate-name check in the service. Note PATCH semantics: `updateBusinessUnit` deliberately treats a null name as 'no change', so if you annotate the shared DTO with `@NotBlank` you will need a separate update DTO or a validation group rather than reusing this one.

backend/src/test/java/com/skapp/community/common/controller/v1/BusinessUnitControllerIntegrationTest.java
```diff
@@ -150,6 +150,22 @@ void createBusinessUnit_withDuplicateName_ReturnsBadRequest() throws Exception {
 						messageUtil.getMessage(CommonMessageConstant.COMMON_ERROR_BUSINESS_UNIT_NAME_ALREADY_EXISTS)));
 		}
 
+		@Test
+		@DisplayName("Create business unit with name differing only by casing - Returns Created")
+		void createBusinessUnit_withNameDifferingByCasing_ReturnsCreated() throws Exception {
+			seedBusinessUnit("Sales");
```
ThinuwanW, line 156: 🤖 **Claude** · 🟣 **Nit**

The new test asserts only the HTTP response and never verifies the resulting DB state, so it cannot distinguish 'a second unit was created' from 'the seeded unit was renamed/overwritten' — both would produce a 201 with name 'SALES'. Add an assertion on the repository after the request, e.g. `assertThat(businessUnitDao.findByNameIgnoreCase("sales")).hasSize(2)` plus a check that the original 'Sales' row still exists by id. That also gives the new `findByNameIgnoreCase` DAO method its only direct test coverage — right now nothing asserts that it returns a multi-element list, which is the whole reason its return type changed from `boolean`/`Optional` to `List`.

backend/src/main/java/com/skapp/community/common/service/impl/BusinessUnitServiceImpl.java
shakila-kamalasena: No need

shakila-kamalasena: No need

shakila-kamalasena: No need

shakila-kamalasena: We do not use them

backend/src/main/java/com/skapp/community/common/service/impl/BusinessUnitServiceImpl.java
shakila-kamalasena: We do not stream

shakila-kamalasena: Not an issue

backend/src/main/java/com/skapp/community/common/service/impl/BusinessUnitServiceImpl.java
shakila-kamalasena: No need

backend/src/test/java/com/skapp/community/common/controller/v1/BusinessUnitControllerIntegrationTest.java
shakila-kamalasena: Not an issue

backend/src/test/java/com/skapp/community/common/controller/v1/BusinessUnitControllerIntegrationTest.java
shakila-kamalasena: Not an issue

backend/src/test/java/com/skapp/community/common/controller/v1/BusinessUnitControllerIntegrationTest.java
shakila-kamalasena: No need

PR #1800, opened 2026-08-19
fix/leave request table height

Deshan1100:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR is a two-line presentational tweak to the My Leave Requests table: it pins a minimum height of 305px on the TableV2 container and lowers the skeleton row count from 5 to 4. I verified the `minHeight` prop exists on `TableViewProps` (typed `string`, default `"min-h-70"`, forwarded as TableV2's `className`), that Tailwind 4 auto-source-detection will pick up the arbitrary `min-h-[305px]` class, and that `skeletonRows={4}` now correctly matches the page size (`leaveRequestParams.size = "4"` in `leaveRequestFilterSlice.tsx:21`) — so the skeleton change is a genuine fix for a loading-state row-count mismatch. Both render contexts (`pages/community/leave/my-requests.tsx` and `LeaveAllocationSummary` on the dashboard) are full-width `ContentLayout` bodies, so the fixed min-height is safe in both. No correctness, type-safety, i18n, or security issues; remaining comments are about magic numbers and consistency with the sibling table.

Found **11** new issue(s): 🔴 1 important, 🟡 6 suggestion(s), 🟣 4 nit(s)

frontend/src/community/leave/components/molecules/LeaveRequests/LeaveRequests.tsx
```diff
@@ -180,7 +180,8 @@ const LeaveRequests: FC = () => {
       headers={tableHeaders}
       rows={transformToTableRows()}
       isLoading={isLoading}
-      skeletonRows={5}
+      minHeight="min-h-[305px]"
+      skeletonRows={4}
```
ThinuwanW, line 193: 🤖 **Claude** · 🟡 **Suggestion**

`skeletonRows={4}` is a second, independent copy of the page size, which actually lives in the Zustand slice (`leaveRequestFilterSlice.tsx:21` and again at line 81: `size: "4"`). The value 5 was wrong precisely because someone changed `size` without touching this literal, and hardcoding 4 leaves the same trap for the next change. Derive it from the store instead, which is the pattern already used elsewhere (`InvoiceTable.tsx:280` does `skeletonRows={itemsPerPage}`): read the params you already have access to via `useLeaveStore` and pass `skeletonRows={Number(leaveRequestParams.size)}` (with a sensible fallback if `size` can be empty). That removes the drift permanently.

frontend/src/community/leave/components/molecules/LeaveRequests/LeaveRequests.tsx
```diff
@@ -180,7 +180,8 @@ const LeaveRequests: FC = () => {
       headers={tableHeaders}
       rows={transformToTableRows()}
       isLoading={isLoading}
-      skeletonRows={5}
+      minHeight="min-h-[305px]"
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`min-h-[305px]` is an unexplained magic number that encodes "header + 4 rows + pagination" implicitly, and it is coupled to the `skeletonRows` value on the very next line — change the page size and this silently becomes wrong (either dead whitespace or a reappearing layout jump), with nothing in the code linking the two. It also sits off the Tailwind spacing scale (the scale is the v4 default of 0.25rem since `styles/global.css` does not override `--spacing`, so `min-h-76` = 304px is the nearest token), and it diverges from `TableView`'s own default of `min-h-70` (280px) that every other table in the app uses. Two options, either is fine: (a) use the scale token `min-h-76` and add a short comment stating the height corresponds to 4 rows, or (b) if 305px is really the correct stable height for a 4-row table, fix the `minHeight` default in `TableView.tsx:25` so all tables agree rather than special-casing this one call site.

ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

The identical defect remains one table over. `ManagerLeaveRequests/ManagerLeaveRequest.tsx:276` passes `skeletonRows={5}` while reading the *same* `leaveRequestParams` slice (`size: "4"`), and it passes no `minHeight`, so it keeps the 280px default. After this PR the two leave-request tables render different skeleton counts and different container heights from the same underlying page size. Please apply the same fix there (and to `CustomLeaveAllocationsTable.tsx:283` / `LeaveTypesTable.tsx:114` if their page sizes also disagree with their hardcoded 5 and 6) so the loading state is consistent across the leave module.

ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

This is the first `TableView` consumer of `minHeight`, and the prop's contract is a footgun worth documenting now that it is in use: `minHeight` is typed as a bare `string` (`TableView/types.ts:78`) but is forwarded as TableV2's `className` (`TableView.tsx:99`), so it must be a Tailwind class. A CSS length silently does nothing — and the same prop name in this repo does take a CSS length (`InvoiceCreateForm.tsx:1488` passes `minHeight="200px"` to `RichTextEditor`), so the confusion is not hypothetical. Suggest renaming to `minHeightClassName` or adding a JSDoc comment on the prop in `types.ts` (the file already documents `pagination` and `infiniteScroll` this way), and consider that overloading TableV2's `className` for min-height alone prevents callers from passing any other class to the table.

ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`min-height` only establishes a floor, so it removes the loading-to-loaded jump only if four real rows are never *taller* than 305px. Rows here render `LeaveRequestDates` plus pill-styled type/status chips (`styles.ts` `iconStyles` adds 0.5rem vertical padding and a border radius), and the duration cell can wrap on narrow viewports, all of which can exceed the skeleton row height. Worth eyeballing this against a full page of four multi-day requests at a narrow breakpoint; if the loaded table still overshoots, a fixed `height` (or matching the skeleton row height to the real row height in the UI package) is the more reliable fix.

frontend/src/community/leave/components/molecules/LeaveRequests/LeaveRequests.tsx
ThinuwanW, line 193: 🤖 **Claude** · 🔴 **Important**

Both new constants assume the table always renders 4 rows, but the page size is shared mutable global state that other pages overwrite. `useGetEmployeeLeaveRequests` (MyRequestApi.ts:154) sends the whole `leaveRequestParams` slice as query params, and this component never sets `size` — it relies on the slice default of "4" (leaveRequestFilterSlice.tsx:21). Meanwhile `ManagerLeaveRequest.tsx:225` and `UserLeaveHistory.tsx:471` both run `setLeaveRequestParams("size", "6")` on mount against that same slice, and `useLeaveStore` is a module-level store with no route-based reset. So for a People Manager / Super Admin: visit /community/leave/leave-requests (or a person's leave tab), then client-side navigate to /community/leave/my-requests — the request now goes out with size=6, the table renders 6 rows, the skeleton renders 4, and the 305px floor is sized for 4. That reintroduces exactly the loading-to-loaded jump this PR is removing, and makes it worse than the previous `skeletonRows={5}`. Fix: make this component own its page size instead of inheriting whatever ran last — add a mount effect `setLeaveRequestParams("size", MY_REQUESTS_PAGE_SIZE)` mirroring ManagerLeaveRequest, and derive `skeletonRows={Number(leaveRequestParams.size)}` from the store so the two can never disagree.

frontend/src/community/leave/components/molecules/LeaveRequests/LeaveRequests.tsx
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`TableView` already ships a min-height default — `minHeight = "min-h-70"` at TableView.tsx:25 (17.5rem / 280px), which every other TableView consumer currently relies on. This PR does not fix that default; it forks it for one call site, creating a second source of truth for table height. If the design system changes TableV2's row height, `min-h-70` gets updated and this table silently drifts (or vice versa). Prefer either (a) correcting the shared default in TableView if 280px is simply wrong, or (b) giving TableView a `reservedRows`/`rowCount` prop that computes the floor from the row height it already knows about, so consumers pass a row count rather than a raw Tailwind class string.

ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Missing paired change — the layout jump on this page is only half fixed. On both pages that render this component (pages/community/leave/my-requests.tsx:78-79 and, via LeaveAllocationSummary, pages/community/dashboard/index.tsx:258), `LeaveAllocation` sits directly above this table and has no reserved height at all: LeaveAllocation.tsx:130 appends `<LeaveAllocationSkeleton />` as an extra Grid child alongside the real cards rather than replacing them, and the pagination footer at LeaveAllocation.tsx:134 only mounts once `entitlement` arrives. Since that block is above the table in normal flow, its height change pushes the table down regardless of the table's own min-height, so the user still sees a jump. Either reserve height on the LeaveAllocation grid too, or note in the PR that this is a partial fix so it doesn't read as resolved.

ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Unit inconsistency: `min-h-[305px]` is an arbitrary px value, while the default it overrides (`min-h-70`) is a Tailwind spacing-scale token and the rest of the leave module uses rem for arbitrary heights (e.g. TeamAvailabilityModal.tsx:32 uses `min-h-[3.75rem]`, and the module's styles.ts files use rem throughout). A px floor stops tracking the root font size while the rem-based row typography inside it keeps growing. `min-h-76` (304px on Tailwind 4's 4px scale) or `min-h-[19rem]` keeps this on-scale and one character shorter.

frontend/src/community/leave/components/molecules/LeaveRequests/LeaveRequests.tsx
ThinuwanW, line 193: 🤖 **Claude** · 🟣 **Nit**

`skeletonRows={4}` is now identical to TableView's own default (`skeletonRows = 4` at TableView.tsx:19), so as written the prop is a no-op — the diff's only real effect is `minHeight`. That's fine if it's intentional documentation, but it means the value is now stated in three places (TableView default, this call site, the slice's `size: "4"`). Deriving it from the store as suggested above collapses all three.

frontend/src/community/leave/components/molecules/LeaveRequests/LeaveRequests.tsx
```diff
@@ -180,7 +180,8 @@ const LeaveRequests: FC = () => {
       headers={tableHeaders}
       rows={transformToTableRows()}
       isLoading={isLoading}
```
ThinuwanW, line 191: 🤖 **Claude** · 🟡 **Suggestion**

The query driving `isLoading` (and therefore whether the skeleton/min-height path is taken at all) is untyped and has an unguarded index. `useGetEmployeeLeaveRequests` (MyRequestApi.ts:153-167) declares no `UseQueryResult<T>` and its `select` does `response?.data?.results[0]` — no `?.` before `[0]`. Under strict mode that makes `leaveRequests` implicitly `any`, so `leaveRequests?.items` (line 90) and `leaveRequests?.totalPages` (line 191) are unchecked, and a response without a `results` array throws inside `select`. The sibling hook in the same file (`useGetMyRequests`, line 103) already does this correctly with `results?.[0].items ?? []` and an explicit `UseQueryResult<MyLeaveRequestPayloadType[]>`. Mirror that: add `?.[0]` and an explicit paginated return type so the row/pagination shape is checked.

Deshan1100: changed the skeleton with this No loading issues with this

frontend/src/community/leave/components/molecules/LeaveRequests/LeaveRequests.tsx
Deshan1100: check with the table height need this

Deshan1100: need size: "4" for match to the new 4 rows in the table view

Deshan1100: need table height

Deshan1100: need table height

Deshan1100: updated new height

Deshan1100: need this PR for the update the fix table height

Deshan1100: need this height

frontend/src/community/leave/components/molecules/LeaveRequests/LeaveRequests.tsx
Deshan1100: with new 4 rows update need this skeletonRows=4

Deshan1100: need this skeletonRows height

Deshan1100: need to update this size

shakila-kamalasena (approved):

akilarootcode (approved):

ThinuwanW (approved):

PR #1801, opened 2026-08-19
feat: assign leave policy in free tier

thusala:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/frontend/src/community/common/assets/languages/english/leaveModule.json
+++ b/frontend/src/community/common/assets/languages/english/leaveModule.json
@@ -522,7 +522,7 @@
       "accrualPreviewTitle": "Accrual Preview",
       "accrualPreviewTooltip": "This is a projection based on the selected policy and effective date. Actual balances may vary.",
       "flexibleInfoLabel": "This is a flexible policy. Leave taken will be recorded but is not deducted from any balance.",
-      "conflictWarning": "{{employeeName}} already has a {{leaveType}} policy. Assigning this policy will replace their existing {{leaveType}} policy.",
+      "conflictWarning": "{{employeeName}} has an existing {{leaveType}} policy. Assigning this policy will replace it.",
       "joinDateMissingLabel": "{{employeeName}} doesn't have a join date yet. Set one to accrue leave from their join date, or choose a specific date instead.",
       "colDate": "DATE",
       "colDays": "DAYS",
--- a/frontend/src/community/leave/components/molecules/IndividualEmployeeLeaveReportSection/IndividualEmployeeLeaveReportSection.tsx
+++ b/frontend/src/community/leave/components/molecules/IndividualEmployeeLeaveReportSection/IndividualEmployeeLeaveReportSection.tsx
@@ -107,7 +107,7 @@ const IndividualEmployeeLeaveReportSection: FC<Props> = ({
               </>
             )}
 
-            {leaveTypesList.length > 0 && (
+            {leaveTypesList?.length > 0 && (
               <UserLeaveUtilization
                 employeeId={selectedUser}
                 leaveTypesList={leaveTypesList}
--- a/frontend/src/community/leave/components/molecules/IndividualEmployeeLeaveReportSection/IndividualEmployeeLeaveReportSection.tsx
+++ b/frontend/src/community/leave/components/molecules/IndividualEmployeeLeaveReportSection/IndividualEmployeeLeaveReportSection.tsx
@@ -7,8 +7,10 @@ import { useGetLeaveTypes } from "~community/leave/api/LeaveApi";
 import UserAssignedLeaveTypes from "~community/leave/components/molecules/UserAssignedLeaveTypes/UserAssignedLeaveTypes";
 import UserLeaveHistory from "~community/leave/components/molecules/UserLeaveHistory/UserLeaveHistory";
 import UserLeavePolicies from "~community/leave/components/molecules/UserLeavePolicies/UserLeavePolicies";
+import UserLeavePoliciesSkeleton from "~community/leave/components/molecules/UserLeavePolicies/UserLeavePoliciesSkeleton";
 import UserLeaveUtilization from "~community/leave/components/molecules/UserLeaveUtilization/UserLeaveUtilization";
 import { USER_ASSIGNED_LEAVE_TYPES_PAGE_SIZE } from "~community/leave/constants/leavePolicyConstants";
+import useCanViewLeavePolicies from "~community/leave/hooks/useCanViewLeavePolicies";
 import useLeavePoliciesEnabled from "~community/leave/hooks/useLeavePoliciesEnabled";
 import { useLeaveStore } from "~community/leave/store/store";
 import { LeaveType } from "~community/leave/types/CustomLeaveAllocationTypes";
@@ -38,8 +40,16 @@ const IndividualEmployeeLeaveReportSection: FC<Props> = ({
 
   const { isAtLeastCoreTier } = useTier();
 
+  const canViewLeavePolicies = useCanViewLeavePolicies();
+
   const { isLeavePoliciesEnabled, isLoading: isLeavePolicyConfigLoading } =
     useLeavePoliciesEnabled();
+
+  const isPolicySectionLoading =
+    canViewLeavePolicies && isLeavePolicyConfigLoading;
+  const showLeavePolicies = canViewLeavePolicies && isLeavePoliciesEnabled;
+  const showAssignedLeaveTypes =
+    !isLeavePolicyConfigLoading && !isLeavePoliciesEnabled;
 
   const employeeName = [employeeFirstName, employeeLastName]
     .filter(Boolean)
@@ -76,7 +86,9 @@ const IndividualEmployeeLeaveReportSection: FC<Props> = ({
       pageHead={translateText(["pageHead"])}
     >
       <Stack sx={classes.sectionsWrapper}>
-        {!isLeavePolicyConfigLoading && isLeavePoliciesEnabled && (
+        {isPolicySectionLoading && <UserLeavePoliciesSkeleton />}
+
+        {showLeavePolicies && (
           <UserLeavePolicies
             employeeId={selectedUser}
             employeeName={employeeName}
@@ -85,7 +97,7 @@ const IndividualEmployeeLeaveReportSection: FC<Props> = ({
 
         <UpgradeOverlay customContainerStyles={classes.customContainerStyles}>
           <>
-            {!isLeavePolicyConfigLoading && !isLeavePoliciesEnabled && (
+            {showAssignedLeaveTypes && (
               <>
                 <h2 className="h2 text-black">{translateText(["pageHead"])}</h2>
                 <UserAssignedLeaveTypes
@@ -95,7 +107,7 @@ const IndividualEmployeeLeaveReportSection: FC<Props> = ({
               </>
             )}
 
-            {leaveTypesList?.length > 0 && (
+            {leaveTypesList.length > 0 && (
               <UserLeaveUtilization
                 employeeId={selectedUser}
                 leaveTypesList={leaveTypesList}
--- a/frontend/src/community/leave/components/molecules/IndividualEmployeeLeaveReportSection/IndividualEmployeeLeaveReportSection.tsx
+++ b/frontend/src/community/leave/components/molecules/IndividualEmployeeLeaveReportSection/IndividualEmployeeLeaveReportSection.tsx
@@ -1,3 +1,4 @@
+import { Stack } from "@mui/material";
 import { FC, useEffect, useMemo, useState } from "react";
 
 import PeopleLayout from "~community/common/components/templates/PeopleLayout/PeopleLayout";
@@ -74,15 +75,17 @@ const IndividualEmployeeLeaveReportSection: FC<Props> = ({
       containerStyles={classes.container}
       pageHead={translateText(["pageHead"])}
     >
-      <UpgradeOverlay customContainerStyles={classes.customContainerStyles}>
-        <>
-          {!isLeavePolicyConfigLoading &&
-            (isLeavePoliciesEnabled ? (
-              <UserLeavePolicies
-                employeeId={selectedUser}
-                employeeName={employeeName}
-              />
-            ) : (
+      <Stack sx={classes.sectionsWrapper}>
+        {!isLeavePolicyConfigLoading && isLeavePoliciesEnabled && (
+          <UserLeavePolicies
+            employeeId={selectedUser}
+            employeeName={employeeName}
+          />
+        )}
+
+        <UpgradeOverlay customContainerStyles={classes.customContainerStyles}>
+          <>
+            {!isLeavePolicyConfigLoading && !isLeavePoliciesEnabled && (
               <>
                 <h2 className="h2 text-black">{translateText(["pageHead"])}</h2>
                 <UserAssignedLeaveTypes
@@ -90,23 +93,24 @@ const IndividualEmployeeLeaveReportSection: FC<Props> = ({
                   pageSize={USER_ASSIGNED_LEAVE_TYPES_PAGE_SIZE}
                 />
               </>
-            ))}
+            )}
 
-          {leaveTypesList?.length > 0 && (
-            <UserLeaveUtilization
+            {leaveTypesList?.length > 0 && (
+              <UserLeaveUtilization
+                employeeId={selectedUser}
+                leaveTypesList={leaveTypesList}
+              />
+            )}
+
+            <UserLeaveHistory
               employeeId={selectedUser}
               leaveTypesList={leaveTypesList}
+              employeeLastName={employeeLastName}
+              employeeFirstName={employeeFirstName}
             />
-          )}
-
-          <UserLeaveHistory
-            employeeId={selectedUser}
-            leaveTypesList={leaveTypesList}
-            employeeLastName={employeeLastName}
-            employeeFirstName={employeeFirstName}
-          />
-        </>
-      </UpgradeOverlay>
+          </>
+        </UpgradeOverlay>
+      </Stack>
     </PeopleLayout>
   );
 };
--- a/frontend/src/community/leave/components/molecules/IndividualEmployeeLeaveReportSection/styles.ts
+++ b/frontend/src/community/leave/components/molecules/IndividualEmployeeLeaveReportSection/styles.ts
@@ -6,9 +6,13 @@ const styles = (): StyleProps => ({
     margin: "0rem auto",
     height: "auto"
   },
-  customContainerStyles: {
+  sectionsWrapper: {
     gap: "1.5rem",
     padding: "0.25rem"
+  },
+  customContainerStyles: {
+    gap: "1.5rem",
+    padding: "0rem"
   }
 });
 
--- a/frontend/src/community/leave/components/molecules/IndividualEmployeeLeaveReportSection/styles.ts
+++ b/frontend/src/community/leave/components/molecules/IndividualEmployeeLeaveReportSection/styles.ts
@@ -1,5 +1,9 @@
 import { StyleProps } from "~community/common/types/CommonTypes";
 
+// The outer wrapper and the overlay each space their own sections; both must use
+// the same gap for the whole tab to read as one evenly spaced list.
+const SECTION_GAP = "1.5rem";
+
 const styles = (): StyleProps => ({
   container: {
     padding: "0rem",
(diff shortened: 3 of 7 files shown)


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR lifts `UserLeavePolicies` out of the `UpgradeOverlay` wrapper so leave-policy assignment stays usable on the free tier, wraps both blocks in a new `Stack` that owns the outer spacing, and rewords the assign-modal conflict warning. The refactor preserves the previous conditional-render semantics exactly (the `!isLeavePolicyConfigLoading` guard, the `leaveTypesList?.length > 0` gate, and the props passed to each child are unchanged), the i18n key still receives both `employeeName` and `leaveType` interpolations from `AssignLeavePolicyModal.tsx:155`, and no other locale file or test references the old copy. Remaining comments are about spacing/permission gating and the still-missing loading state around the newly split branches — no correctness regressions found.

Found **10** new issue(s): 🔴 1 important, 🟡 5 suggestion(s), 🟣 4 nit(s)

frontend/src/community/leave/components/molecules/IndividualEmployeeLeaveReportSection/IndividualEmployeeLeaveReportSection.tsx
```diff
@@ -74,39 +75,42 @@ const IndividualEmployeeLeaveReportSection: FC<Props> = ({
...
-        <>
-          {!isLeavePolicyConfigLoading &&
-            (isLeavePoliciesEnabled ? (
-              <UserLeavePolicies
-                employeeId={selectedUser}
-                employeeName={employeeName}
-              />
-            ) : (
+      <Stack sx={classes.sectionsWrapper}>
+        {!isLeavePolicyConfigLoading && isLeavePoliciesEnabled && (
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The restructure splits the config-dependent render into two sibling branches, but neither renders while `isLeavePolicyConfigLoading` is true — the top of the section is simply blank until `useGetLeavePolicyConfig` resolves, then content pops in and pushes the overlay block down (layout shift). The established pattern elsewhere renders a skeleton for this exact state (see `MyLeaveAllocationSection.tsx:16-22`, which returns `LeaveAllocationSkeleton` while `isLoading`). Consider rendering `UserLeavePoliciesSkeleton` (already exists at `UserLeavePolicies/UserLeavePoliciesSkeleton.tsx`) while the config is loading. Also note the `!isLeavePolicyConfigLoading &&` in this first condition is redundant: `useLeavePoliciesEnabled` returns `Boolean(data?.isEnabled)`, which is already `false` during loading — only the second branch (`!isLeavePoliciesEnabled`) actually needs the loading guard.

frontend/src/community/leave/components/molecules/IndividualEmployeeLeaveReportSection/IndividualEmployeeLeaveReportSection.tsx
```diff
@@ -74,39 +75,42 @@ const IndividualEmployeeLeaveReportSection: FC<Props> = ({
...
+          <UserLeavePolicies
```
ThinuwanW, line 92: 🤖 **Claude** · 🟡 **Suggestion**

`UserLeavePolicies` is now rendered outside `UpgradeOverlay`, so on the free tier it is fully visible and interactive rather than sitting behind the blur. Everything that can reach the leave tab now renders it — `PeopleIndividualSection.tsx:69-76` routes here for leave managers and attendance managers, not just admins. The codebase already has a dedicated view gate for this data, `useCanViewLeavePolicies` (super admin / leave admin / people admin), used for exactly this purpose in `Drawer.tsx:131-134`; only the assign/unassign buttons inside `UserLeavePolicies` are gated (via `useCanManageLeavePolicies`), the policy list and its `useGetEmployeeLeavePolicies` fetch are not. For a non-admin manager the request will likely 403 and the component will render its `errorStateTitle` EmptyDataView ("Couldn't load leave policies") instead of simply hiding. Suggest gating with `useCanViewLeavePolicies()` alongside `isLeavePoliciesEnabled`, and passing that flag into `useLeavePoliciesEnabled(canView)` so the config query is not fired for users who cannot see the section.

frontend/src/community/leave/components/molecules/IndividualEmployeeLeaveReportSection/styles.ts
```diff
@@ -6,9 +6,13 @@ const styles = (): StyleProps => ({
...
     height: "auto"
   },
-  customContainerStyles: {
+  sectionsWrapper: {
     gap: "1.5rem",
     padding: "0.25rem"
+  },
+  customContainerStyles: {
+    gap: "1.5rem",
+    padding: "0rem"
```
ThinuwanW, line 15: 🤖 **Claude** · 🟡 **Suggestion**

`customContainerStyles` now sets `padding: "0rem"` unconditionally. `UpgradeOverlay` merges its sx as `mergeSx([classes.container, { padding: isFreeTier ? "1.25rem" : "0rem" }, customContainerStyles])` and `mergeSx` appends later entries last, so this key always wins and the deliberate 1.25rem free-tier inset is discarded. Combined with the wrapper's `borderRadius: "0.75rem"` + `overflow: "hidden"`, the blurred content and upgrade CTA now sit flush against the clipped rounded edge on the free tier. Since the new `sectionsWrapper` Stack already supplies the outer 0.25rem padding, drop the `padding` key from `customContainerStyles` entirely (keep only `gap`) so `UpgradeOverlay`'s own tier-aware padding applies.

frontend/src/community/leave/components/molecules/IndividualEmployeeLeaveReportSection/styles.ts
```diff
@@ -6,9 +6,13 @@ const styles = (): StyleProps => ({
     margin: "0rem auto",
     height: "auto"
   },
-  customContainerStyles: {
+  sectionsWrapper: {
     gap: "1.5rem",
     padding: "0.25rem"
+  },
+  customContainerStyles: {
+    gap: "1.5rem",
```
ThinuwanW, line 14: 🤖 **Claude** · 🟣 **Nit**

`gap: "1.5rem"` is now duplicated between `sectionsWrapper` and `customContainerStyles` — the two wrappers must stay in sync for the outer and inner sections to look like one evenly spaced list, but nothing enforces that. Consider extracting the value to a single local constant (or MUI `theme.spacing`) and referencing it from both keys.

frontend/src/community/leave/components/molecules/IndividualEmployeeLeaveReportSection/IndividualEmployeeLeaveReportSection.tsx
```diff
@@ -1,3 +1,4 @@
+import { Stack } from "@mui/material";
```
ThinuwanW, line 1: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

There is no `IndividualEmployeeLeaveReportSection.test.tsx`, and this PR changes the component's branching contract (policies block outside the overlay, assigned-leave-types block inside it, guarded by `isLeavePolicyConfigLoading` / `isLeavePoliciesEnabled`). A small `@testing-library/react` test with the `MockTheme` wrapper asserting the three states — loading, policies-enabled, policies-disabled — would lock in the free-tier placement so a future edit doesn't silently move `UserLeavePolicies` back inside `UpgradeOverlay`. Note the Jest config maps `~enterprise` to `src/fallback`, whose `UpgradeOverlay` stub renders `<></>` and drops its children, so a test would naturally verify that `UserLeavePolicies` renders independently of that wrapper.

frontend/src/community/leave/components/molecules/IndividualEmployeeLeaveReportSection/IndividualEmployeeLeaveReportSection.tsx
```diff
@@ -74,39 +75,42 @@ const IndividualEmployeeLeaveReportSection: FC<Props> = ({
...
-            ) : (
+      <Stack sx={classes.sectionsWrapper}>
+        {!isLeavePolicyConfigLoading && isLeavePoliciesEnabled && (
+          <UserLeavePolicies
+            employeeId={selectedUser}
+            employeeName={employeeName}
+          />
+        )}
+
+        <UpgradeOverlay customContainerStyles={classes.customContainerStyles}>
```
ThinuwanW, line 98: 🤖 **Claude** · 🔴 **Important**

The `~enterprise` UpgradeOverlay import resolves to `src/fallback/common/components/molecules/UpgradeOverlay/UpgradeOverlay.tsx` in community builds (tsconfig paths list `./src/fallback/*` as the second candidate) and in *every* Jest run (jest.config.js maps `^~enterprise(.*)$` straight to `<rootDir>/src/fallback$1`). That stub is declared with no props and returns an empty fragment, so it discards both `children` and `customContainerStyles`. This PR makes that asymmetry load-bearing: previously the whole block vanished together in a community build, now `UserLeavePolicies` is hoisted outside the overlay and renders, while the assigned-leave-types heading, `UserLeaveUtilization` and `UserLeaveHistory` are still silently dropped — the page looks functional but has lost the leave history entirely. Fix the fallback stub to accept and render its children (and accept an optional `customContainerStyles` it can ignore), the way the `AnnouncementProvider` fallback already does. This also blocks writing the component test that the missing-test finding asks for, since under Jest nothing inside the overlay can ever be asserted.

frontend/src/community/common/assets/languages/english/leaveModule.json
```diff
@@ -522,7 +522,7 @@
       "accrualPreviewTitle": "Accrual Preview",
       "accrualPreviewTooltip": "This is a projection based on the selected policy and effective date. Actual balances may vary.",
       "flexibleInfoLabel": "This is a flexible policy. Leave taken will be recorded but is not deducted from any balance.",
-      "conflictWarning": "{{employeeName}} already has a {{leaveType}} policy. Assigning this policy will replace their existing {{leaveType}} policy.",
+      "conflictWarning": "{{employeeName}} has an existing {{leaveType}} policy. Assigning this policy will replace it.",
```
ThinuwanW, line 525: 🤖 **Claude** · 🟡 **Suggestion**

The reworded copy now makes a firm promise ("Assigning this policy will replace it"), but the helper that decides when to show it, `findSupersededAssignment` in leavePolicyAssignmentUtils.ts, deliberately excludes assignments whose `policyId` equals the selected policy's id. The backend `assignPolicyToEmployee` ends the current active assignment whenever the policy id matches but the effective date or effective-date type differs, then creates a fresh one — the same destructive outcome the unassign modal explicitly warns about ("Any unused leave balance for this policy will be lost"). So the one case that reassigns the *same* policy on a new effective date replaces the assignment with no warning at all. Either drop the `policyId !== selectedPolicy.id` exclusion so the same-policy-different-date case also warns, or add a distinct message for it; while you are there, note the check only inspects the first fetched page of existing assignments.

frontend/src/community/leave/components/molecules/IndividualEmployeeLeaveReportSection/IndividualEmployeeLeaveReportSection.tsx
```diff
@@ -74,39 +75,42 @@ const IndividualEmployeeLeaveReportSection: FC<Props> = ({
...
+          <>
+            {!isLeavePolicyConfigLoading && !isLeavePoliciesEnabled && (
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The single ternary has been split into two sibling conditions that both re-derive the same guard (`!isLeavePolicyConfigLoading` combined with the enabled flag and its negation). They are only mutually exclusive by convention now, so a future edit to one branch can leave the other stale or make both render at once. Extract derived booleans once (e.g. `const isPolicyConfigResolved = !isLeavePolicyConfigLoading;` then `showPolicySection` / `showLegacyLeaveTypes`) and use those in the JSX. Related: `useLeavePoliciesEnabled` also returns `isError`, which this component discards — when the `/leave/policies/config` request fails, `isLeavePoliciesEnabled` is `false` and `isLoading` is `false`, so the legacy assigned-leave-types branch renders as though policies were switched off rather than surfacing any error. Consider consuming `isError` (retry/error view) or at least documenting that the legacy branch is the intended failure mode.

frontend/src/community/leave/components/molecules/IndividualEmployeeLeaveReportSection/IndividualEmployeeLeaveReportSection.tsx
```diff
@@ -74,39 +75,42 @@ const IndividualEmployeeLeaveReportSection: FC<Props> = ({
...
                   employeeId={selectedUser}
                   pageSize={USER_ASSIGNED_LEAVE_TYPES_PAGE_SIZE}
                 />
               </>
-            ))}
+            )}
 
-          {leaveTypesList?.length > 0 && (
-            <UserLeaveUtilization
+            {leaveTypesList?.length > 0 && (
```
ThinuwanW, line 110: 🤖 **Claude** · 🟣 **Nit**

`leaveTypesList` is `useState<LeaveType[]>([])` and is only ever set from a non-null value, so the optional chain before `.length` is dead defensive code and is inconsistent with the direct array access used elsewhere in the file. Drop the `?.` so the guard reads as a plain length check — otherwise it reads as if the value could be undefined, which under strict TS would make the `> 0` comparison an error rather than something to paper over.

frontend/src/community/leave/components/molecules/IndividualEmployeeLeaveReportSection/IndividualEmployeeLeaveReportSection.tsx
```diff
@@ -74,39 +75,42 @@ const IndividualEmployeeLeaveReportSection: FC<Props> = ({
...
-      <UpgradeOverlay customContainerStyles={classes.customContainerStyles}>
-        <>
-          {!isLeavePolicyConfigLoading &&
-            (isLeavePoliciesEnabled ? (
-              <UserLeavePolicies
-                employeeId={selectedUser}
-                employeeName={employeeName}
-              />
-            ) : (
+      <Stack sx={classes.sectionsWrapper}>
```
ThinuwanW, line 88: 🤖 **Claude** · 🟣 **Nit**

UpgradeOverlay's own container style sets `height: "100%"` together with `overflow: "hidden"`. It used to be the sole child of PeopleLayout's plain block Box; it is now a flex item in a column MUI Stack. Today the ancestor chain is auto-height so the percentage resolves to auto and nothing clips, but the combination is fragile — the moment any ancestor gains a definite height (or the wrapper gains `flex: 1` / a height), the overlay will clip the leave-history table instead of growing. Worth verifying the free-tier and paid-tier renders visually, and consider passing `height: "auto"` through `customContainerStyles` (which already overrides padding) to make the intent explicit.

frontend/src/community/leave/components/molecules/IndividualEmployeeLeaveReportSection/IndividualEmployeeLeaveReportSection.tsx
thusala: Done in 003128a74 — the fallback stub now takes `children` (plus an ignored `customContainerStyles`) and renders them. Worth noting the same stub also wrapped `PeopleTimeline` and `IndividualEmployeeTimeReportBody`, so community builds were dropping those two bodies entirely before this.

frontend/src/community/leave/components/molecules/IndividualEmployeeLeaveReportSection/IndividualEmployeeLeaveReportSection.tsx
thusala: Done in 003128a74 — the policy section is now gated on `useCanViewLeavePolicies()`. I kept `useLeavePoliciesEnabled()` unconditional though: it decides which of the two views renders, so gating the config query would flip a leave manager on a policies-enabled tenant onto the legacy assigned-leave-types view. With the render gate they correctly see neither block, just utilization and history.

frontend/src/community/leave/components/molecules/IndividualEmployeeLeaveReportSection/IndividualEmployeeLeaveReportSection.tsx
thusala: Done in 003128a74 — `UserLeavePoliciesSkeleton` now renders while the config resolves, and the redundant `!isLeavePolicyConfigLoading` is dropped from the enabled branch (`isLeavePoliciesEnabled` is already false while loading, as you noted).

frontend/src/community/leave/components/molecules/IndividualEmployeeLeaveReportSection/IndividualEmployeeLeaveReportSection.tsx
thusala: Guards extracted in 003128a74 as `isPolicySectionLoading` / `showLeavePolicies` / `showAssignedLeaveTypes`. Leaving `isError` alone: the component never consumed it and the legacy branch has always been the config-failure path, so changing that is a behaviour change outside this PR. Note `Drawer.tsx` treats an error as "policies on" instead, so the two disagree — worth aligning separately.

frontend/src/community/leave/components/molecules/IndividualEmployeeLeaveReportSection/IndividualEmployeeLeaveReportSection.tsx
thusala: Done in 003128a74 — dropped the `?.`.

frontend/src/community/leave/components/molecules/IndividualEmployeeLeaveReportSection/styles.ts
thusala: Not changing this one — the override predates the PR: `customContainerStyles` already set `padding: "0.25rem"`, so the tier-aware 1.25rem was discarded before as well. Total inset is unchanged (0.25rem, now supplied by `sectionsWrapper`). Dropping the key would give the free tier 1.5rem, which is a deliberate spacing change to the existing layout rather than part of this fix.

frontend/src/community/common/assets/languages/english/leaveModule.json
thusala: Out of scope for this PR — the copy was reworded, but the `policyId !== selectedPolicy.id` exclusion in `findSupersededAssignment` is pre-existing and unchanged here. You are right that the BE ends the active assignment when only the effective date differs, so that case replaces silently; it needs its own message rather than reusing this one, and the page-1-only check needs fixing too. Raising it as a follow-up.

frontend/src/community/leave/components/molecules/IndividualEmployeeLeaveReportSection/IndividualEmployeeLeaveReportSection.tsx
thusala: Out of scope — there are zero `*.test.tsx` files across all 84 molecules in `community/leave`, so a test here would not match the module convention. The `~enterprise`→`src/fallback` stub is fixed in 003128a74, so the blocker you flagged is gone if we do add tests for this module.

frontend/src/community/leave/components/molecules/IndividualEmployeeLeaveReportSection/IndividualEmployeeLeaveReportSection.tsx
thusala: Leaving as is — as noted, the ancestor chain is auto-height so nothing clips today, and `customContainerStyles` no longer needs to override anything beyond gap/padding. Adding a defensive `height: "auto"` for a hypothetical future ancestor is the kind of speculative guard we normally keep out; happy to add it if you would rather pin the intent.

frontend/src/community/leave/components/molecules/IndividualEmployeeLeaveReportSection/styles.ts
thusala: Keeping the literal `1.5rem` in both keys — we prefer plain values in these `styles.ts` files over extracted constants, so reverted in 4ce45261b.

frontend/src/community/leave/components/molecules/IndividualEmployeeLeaveReportSection/IndividualEmployeeLeaveReportSection.tsx
thusala: Reverted in 91685ad3b — keeping this out of scope. The stub dropping its children is pre-existing and not specific to this PR: it also wraps `PeopleTimeline` and `IndividualEmployeeTimeReportBody`, so community builds already lose those two bodies entirely. Fixing it belongs in its own change rather than riding along here. Raising it as a follow-up.

hasalarootcode (dismissed):

akilarootcode (dismissed):

ThinuwanW (approved):

PR #1802, opened 2026-08-19
Fix/flexible unlimited icon issue in policy balance card

hasalarootcode:
…yCard and PolicyLeaveBalanceCard

### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/frontend/src/community/common/assets/languages/english/leaveModule.json
+++ b/frontend/src/community/common/assets/languages/english/leaveModule.json
@@ -263,7 +263,7 @@
     "title": "Leave Policies",
     "pageHead": "Leave Policies | Skapp",
     "breadcrumbLeave": "Leave",
-    "breadcrumbLeavePolicies": "Leave policies",
+    "breadcrumbLeavePolicies": "Leave Policies",
     "createPolicyBtnTxt": "Create policy",
     "bulkAssignBtnTxt": "Bulk upload",
     "bulkAssignModal": {
--- a/frontend/src/community/leave/components/molecules/LeavePolicyCard/LeavePolicyCard.tsx
+++ b/frontend/src/community/leave/components/molecules/LeavePolicyCard/LeavePolicyCard.tsx
@@ -1,6 +1,6 @@
 import { Box, Stack, Tooltip, Typography } from "@mui/material";
 import { type Theme, useTheme } from "@mui/material/styles";
-import { ArrowRightIcon, ButtonV2 } from "@rootcodelabs/skapp-ui";
+import { InfinityIcon, ArrowRightIcon, ButtonV2 } from "@rootcodelabs/skapp-ui";
 import { KeyboardEvent, MouseEvent, forwardRef, useState } from "react";
 
 import { ToastType } from "~community/common/enums/ComponentEnums";
@@ -124,13 +124,17 @@ const LeavePolicyCard = forwardRef<HTMLDivElement, Props>(
           </Typography>
           <Stack>
             <Stack sx={classes.amount}>
-              <Typography
-                sx={isUnlimited ? classes.unlimited : classes.heading}
-              >
-                {balanceLabel}
-              </Typography>
-              {!isUnlimited && (
-                <Typography variant="body2">/ {totalDaysAllocated}</Typography>
+              {isUnlimited ? (
+                <Box component="span" role="img" aria-label={balanceLabel}>
+                  <InfinityIcon width="32" height="32" />
+                </Box>
+              ) : (
+                <>
+                  <Typography sx={classes.heading}>{balanceLabel}</Typography>
+                  <Typography variant="body2">
+                    / {totalDaysAllocated}
+                  </Typography>
+                </>
               )}
             </Stack>
             <Typography component="div" variant="caption">
--- a/frontend/src/community/leave/components/molecules/PolicyLeaveBalanceCard/PolicyLeaveBalanceCard.tsx
+++ b/frontend/src/community/leave/components/molecules/PolicyLeaveBalanceCard/PolicyLeaveBalanceCard.tsx
@@ -1,3 +1,4 @@
+import { InfinityIcon } from "@rootcodelabs/skapp-ui";
 import { useState } from "react";
 
 import Icon from "~community/common/components/atoms/Icon/Icon";
@@ -47,11 +48,17 @@ const PolicyLeaveBalanceCard = ({ policyBalance }: Props) => {
             <p className="body2">
               {translateText(["policyBalanceCard", "available"])}
             </p>
-            <p className="body2">
-              {policyBalance.isUnlimited || !policyBalance.isBalanceAvailable
-                ? balanceLabel
-                : `${balanceLabel} / ${policyBalance.totalDaysAllocated}`}
-            </p>
+            {policyBalance.isUnlimited ? (
+              <span role="img" aria-label={balanceLabel}>
+                <InfinityIcon width="32" height="32" />
+              </span>
+            ) : (
+              <p className="body2">
+                {!policyBalance.isBalanceAvailable
+                  ? balanceLabel
+                  : `${balanceLabel} / ${policyBalance.totalDaysAllocated}`}
+              </p>
+            )}
           </div>
           <div className={ROW_CLASSES}>
             <p className="body2">
--- a/frontend/src/community/leave/components/molecules/LeavePolicyCard/styles.ts
+++ b/frontend/src/community/leave/components/molecules/LeavePolicyCard/styles.ts
@@ -43,11 +43,6 @@ const styles = (theme: Theme): StyleProps => ({
     fontSize: "2.5rem",
     lineHeight: "2.5rem"
   },
-  unlimited: {
-    fontWeight: 500,
-    fontSize: "1.5rem",
-    lineHeight: "2.5rem"
-  },
   rightContent: {
     width: "max-content"
   }


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR replaces the textual "Unlimited" balance label with an `InfinityIcon` in two leave-policy components and title-cases the leave-policies breadcrumb. The breadcrumb change is correct and matches the app-wide Title Case breadcrumb convention (`peopleModule.json` breadcrumbs, plus `title`/`pageHead` on the same page), and removing the now-dead `unlimited` style from `styles.ts` is verified safe (no remaining `classes.unlimited` references). However, the new `isUnlimited` branch drops the previous handling of the `isUnlimited && !isBalanceAvailable` state, so an unavailable balance can now render an infinity glyph with an accessible name of "—"; the icon markup also diverges from the pre-existing equivalent in `UserLeavePolicies/LeavePolicyCard.tsx` (no color token, no `shrink-0`, oversized inside a compact tooltip) and is now duplicated in three places.

Found **11** new issue(s): 🔴 2 important, 🟡 6 suggestion(s), 🟣 3 nit(s)

frontend/src/community/leave/components/molecules/LeavePolicyCard/LeavePolicyCard.tsx
```diff
@@ -124,13 +124,17 @@ const LeavePolicyCard = forwardRef<HTMLDivElement, Props>(
...
           <Stack>
             <Stack sx={classes.amount}>
-              <Typography
-                sx={isUnlimited ? classes.unlimited : classes.heading}
-              >
-                {balanceLabel}
-              </Typography>
-              {!isUnlimited && (
-                <Typography variant="body2">/ {totalDaysAllocated}</Typography>
+              {isUnlimited ? (
```
ThinuwanW, line 127: 🤖 **Claude** · 🔴 **Important**

Behavioral regression: the new branch keys only on `isUnlimited` and no longer accounts for `!isBalanceAvailable`. `getPolicyBalanceLabel` (policyLeaveUtils.ts:120) returns the em-dash `BALANCE_UNAVAILABLE_LABEL` whenever `!isBalanceAvailable`, *regardless of* `isUnlimited`. Previously this state rendered that "—" as text. Now, when `isUnlimited && !isBalanceAvailable`, the card renders an infinity glyph — i.e. "you have unlimited balance" — while simultaneously being non-actionable (`isActionable = !isDisabled && isBalanceAvailable`) and wrapped in the `balanceUnavailableTooltip` ("Balance unavailable, please refresh") at line 164-172. Two directly contradictory signals on the same card. Worse, `aria-label={balanceLabel}` then sets the accessible name of the `role="img"` element to the literal string "—", which is meaningless to a screen reader.

The identical regression exists in `PolicyLeaveBalanceCard.tsx:51`, where the old code explicitly wrote `policyBalance.isUnlimited || !policyBalance.isBalanceAvailable ? balanceLabel : ...` — the `||` shows the original author deliberately handled this pair, and the refactor dropped it.

Reachability caveat: today the backend always sets `isDerived = true` (every construction path in `PolicyLeaveServiceImpl.calculateBalance`/`emptyBalance`/`unlimitedBalance` calls `setDerived(true)`), and `isBalanceAvailable` is mapped from it at line 541, so the combination is not currently produced. But `isBalanceAvailable` is a required `boolean` on `EmployeePolicyBalanceType` that the frontend defends against everywhere else, and a missing/omitted field deserializes to `undefined` (falsy) — so this silently mis-renders rather than degrading safely.

Fix: gate the icon on `isUnlimited && isBalanceAvailable` in both files (falling through to the `balanceLabel` text otherwise), and stop deriving the icon's accessible name from `balanceLabel` — pass the dedicated `translateText(["unlimited"])` value instead, so the label can never become "—".

frontend/src/community/leave/components/molecules/LeavePolicyCard/LeavePolicyCard.tsx
```diff
@@ -124,13 +124,17 @@ const LeavePolicyCard = forwardRef<HTMLDivElement, Props>(
...
+                <Box component="span" role="img" aria-label={balanceLabel}>
```
ThinuwanW, line 128: 🤖 **Claude** · 🟡 **Suggestion**

The screen-reader experience for unlimited policies is now worse than the visual one. The outer `Stack` at line 99 has `role="button"` with its own `aria-label` built from `cardAriaLabel`, whose English string is `"{{policyName}}, {{leaveType}}, {{balance}} of {{allocated}} days available"` (leaveModule.json:163). For a FLEXIBLE policy the backend returns `totalDaysAllocated = 0` (`unlimitedBalance` -> `zeroedBalance` sets it to zero), so that label resolves to "…, Unlimited of 0 days available". Because `role="button"` is a name-from-author widget, assistive tech generally exposes it as a leaf and does not announce descendant content — so the new inner `aria-label={balanceLabel}` on the icon is very likely never announced, and the only thing conveyed is the misleading "Unlimited of 0 days available".

Fix: add a dedicated key (e.g. `cardAriaLabelUnlimited`: "{{policyName}}, {{leaveType}}, unlimited balance available") to the `leavePolicyAllocation` namespace and select it when `isUnlimited` is true, instead of interpolating `allocated: totalDaysAllocated` into the generic string.

frontend/src/community/leave/components/molecules/PolicyLeaveBalanceCard/PolicyLeaveBalanceCard.tsx
```diff
@@ -47,11 +48,17 @@ const PolicyLeaveBalanceCard = ({ policyBalance }: Props) => {
...
               {translateText(["policyBalanceCard", "available"])}
             </p>
-            <p className="body2">
-              {policyBalance.isUnlimited || !policyBalance.isBalanceAvailable
-                ? balanceLabel
-                : `${balanceLabel} / ${policyBalance.totalDaysAllocated}`}
-            </p>
+            {policyBalance.isUnlimited ? (
+              <span role="img" aria-label={balanceLabel}>
+                <InfinityIcon width="32" height="32" />
```
ThinuwanW, line 53: 🤖 **Claude** · 🟡 **Suggestion**

A 32x32 icon is oversized for this container. This element sits inside the tooltip `title` content, whose rows are `flex flex-row justify-between gap-4` with `body2` text and only `gap-1` between rows. Swapping the `<p className="body2">` for a 32px SVG will inflate the "Available Balance" row height and leave it visually misaligned against the "Effective From Date" and "Expiry Date" rows immediately below (lines 63-78), which are still `body2`. The 32px size was copied from `UserLeavePolicies/LeavePolicyCard.tsx:36`, where it sits next to `h1b`-scale text in a 96px-tall card — a very different context.

Fix: size the tooltip variant to match the surrounding text (roughly `width="20" height="20"`, i.e. the same `1.25rem` used for the trigger's `spanStyles` on line 44) and keep 32 only for the large allocation card. Also prefer rem-based values over raw pixel strings — every other dimension in this module (`styles.ts`, `spanStyles`, Tailwind `size-5`) is rem-based, so hardcoded px will not scale with the user's font-size setting.

frontend/src/community/leave/components/molecules/LeavePolicyCard/LeavePolicyCard.tsx
ThinuwanW, line 128: 🤖 **Claude** · 🟡 **Suggestion**

Both new icon wrappers diverge from the pre-existing implementation of this exact same indicator. `UserLeavePolicies/LeavePolicyCard.tsx:31-36` renders `<span className="shrink-0 text-secondary-text" role="img" aria-label={...}>`. The two new call sites omit both classes, so (a) the SVG falls back to ambient `currentColor` rather than the secondary-text theme token, and (b) without `shrink-0` the wrapper can be compressed by its flex parent — relevant here since `LeavePolicyAllocation.tsx:135` renders these cards in `Grid size={{ xs: 6, md: 4 }}`, i.e. two-up on small screens next to a potentially long `leaveType.name`. Add the same `shrink-0` and color token to both.

Separately, please visually verify the unchanged `classes.amount` rule in `styles.ts` (`alignItems: "baseline"`, lines 37-40). That was tuned for text with an explicit `2.5rem` line-height; an inline SVG aligns its bottom edge to the text baseline and leaves descender whitespace below, so the infinity glyph may no longer be centred relative to the "Available" caption at line 140. The removed `unlimited` style also used a smaller `1.5rem` font, so the unlimited variant's optical weight changes here in a way that is easy to miss.

frontend/src/community/leave/components/molecules/LeavePolicyCard/LeavePolicyCard.tsx
```diff
@@ -1,6 +1,6 @@
 import { Box, Stack, Tooltip, Typography } from "@mui/material";
 import { type Theme, useTheme } from "@mui/material/styles";
-import { ArrowRightIcon, ButtonV2 } from "@rootcodelabs/skapp-ui";
+import { InfinityIcon, ArrowRightIcon, ButtonV2 } from "@rootcodelabs/skapp-ui";
```
ThinuwanW, line 3: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

This PR brings the "unlimited balance -> infinity icon" markup to three copies: `UserLeavePolicies/LeavePolicyCard.tsx:30-37` (pre-existing), `LeavePolicyCard/LeavePolicyCard.tsx:127-130`, and `PolicyLeaveBalanceCard.tsx:51-54`. All three are structurally identical (`role="img"` wrapper + `<InfinityIcon width="32" height="32"/>`), yet each sources its accessible name from a different place — `leavePolicyAssignment.balanceNotTrackedLabel` in one and `leavePolicyAllocation.unlimited` (via `getPolicyBalanceLabel`) in the other two — and each hardcodes its own size. Per the review guidelines ("2+ components with near-identical markup -> extract shared component"), extract an atom such as `UnlimitedBalanceIndicator` taking a `size` prop and owning the single canonical translation key. That keeps the a11y label, color token and sizing from drifting across the three surfaces the way the labels already have.

frontend/src/community/leave/components/molecules/LeavePolicyCard/LeavePolicyCard.tsx
```diff
@@ -124,13 +124,17 @@ const LeavePolicyCard = forwardRef<HTMLDivElement, Props>(
...
-              <Typography
-                sx={isUnlimited ? classes.unlimited : classes.heading}
-              >
-                {balanceLabel}
-              </Typography>
-              {!isUnlimited && (
-                <Typography variant="body2">/ {totalDaysAllocated}</Typography>
+              {isUnlimited ? (
+                <Box component="span" role="img" aria-label={balanceLabel}>
+                  <InfinityIcon width="32" height="32" />
```
ThinuwanW, line 129: 🤖 **Claude** · 🔴 **Important**

Disabled-state styling no longer applies to the balance indicator. `styles.ts` greys out disabled cards with `disabledCard["& .MuiTypography-root"] = { color: theme.palette.grey[700] }` — that selector matched the old `<Typography>` balance, but the new `Box component="span"` + `<InfinityIcon />` is not a MuiTypography, so the icon keeps its full-strength inherited color while the surrounding text (`leaveType.name`, the `Available` caption) drops to grey[700]. `filter: saturate(0)` only removes saturation, not luminance, so a dark icon still reads as "enabled" on a disabled card. This state is reachable: backend `PolicyLeaveServiceImpl.resolveDisabledReason` returns POLICY_INACTIVE, ALLOCATION_PERIOD_EXPIRED and NO_SUPERVISOR_ASSIGNED independently of `isUnlimited` (only FULLY_UTILIZED is excluded by `hasBalance`). Fix by extending the disabled override to cover the icon, e.g. add `"& svg": { color: theme.palette.grey[700] }` to `disabledCard` in `styles.ts` (and ensure the icon inherits `currentColor`), rather than relying on the Typography-only selector.

frontend/src/community/leave/components/molecules/LeavePolicyCard/LeavePolicyCard.tsx
ThinuwanW, line 127: 🤖 **Claude** · 🟡 **Suggestion**

The card's accessible name is now the only textual carrier of the balance for unlimited policies, and its template is wrong for that case. `cardAriaLabel` is "{{policyName}}, {{leaveType}}, {{balance}} of {{allocated}} days available", and for unlimited policies the backend builds the balance via `zeroedBalance()`/`unlimitedBalance()`, which sets `totalDaysAllocated = 0f`. So screen readers announce "<policy>, <leaveType>, Unlimited of 0 days available". Since this PR promotes "unlimited" to a distinct visual state, add a dedicated key (e.g. `leavePolicyAllocation.cardUnlimitedAriaLabel`: "{{policyName}}, {{leaveType}}, unlimited balance available") and select it when `isUnlimited` is true, so the `{{allocated}}` interpolation is not used for a value that is meaningless in that state.

frontend/src/community/leave/components/molecules/LeavePolicyCard/LeavePolicyCard.tsx
ThinuwanW, line 128: 🤖 **Claude** · 🟡 **Suggestion**

Vertical rhythm regression between unlimited and normal cards. The deleted `classes.unlimited` set `lineHeight: "2.5rem"`, matching `classes.heading`'s 2.5rem, so both variants produced a 2.5rem-tall balance row. The new wrapper contains only a 32px (2rem) icon, so the row is ~8px shorter and — because `activeCard` has a fixed `minHeight: 7.5rem` and `leftContent` uses `gap: 1.125rem` — the "Available" caption of unlimited cards sits higher than that of the limited cards rendered beside them. Give the wrapper a matching box, e.g. `sx={{ display: "flex", alignItems: "center", minHeight: "2.5rem" }}` on the `Box`, so both branches occupy the same height.

frontend/src/community/leave/components/molecules/LeavePolicyCard/LeavePolicyCard.tsx
```diff
@@ -124,13 +124,17 @@ const LeavePolicyCard = forwardRef<HTMLDivElement, Props>(
...
+                </Box>
+              ) : (
+                <>
+                  <Typography sx={classes.heading}>{balanceLabel}</Typography>
+                  <Typography variant="body2">
+                    / {totalDaysAllocated}
```
ThinuwanW, line 135: 🤖 **Claude** · 🟣 **Nit**

Number formatting is inconsistent with the sibling card that renders the same fields. This branch prints `balanceInDays.toString()` (from `getPolicyBalanceLabel`) and the raw `{totalDaysAllocated}`, while `UserLeavePolicies/LeavePolicyCard.tsx:47,50` renders the same two values through `formatDays()` from `~community/common/utils/commonUtil` (`Number.isInteger(v) ? String(v) : v.toFixed(1)`). Accrual is rounded to half-days, but `applyAccrualCap`/`capCarryover` use admin-configured `Float` caps (`accrualCapDays`, `maxCarryoverDays`), so `totalDaysAllocated`/`balanceInDays` can arrive with more than one decimal and would render un-truncated at 2.5rem here but truncated on the other card. Since these lines were rewritten anyway, wrap both in `formatDays(...)` (and consider using it inside `getPolicyBalanceLabel` so all callers agree).

frontend/src/community/leave/components/molecules/LeavePolicyCard/LeavePolicyCard.tsx
ThinuwanW, line 3: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

No tests accompany the rendering-logic change. The repo has 119 `*.test.tsx` files, but none of the three cards touched/related here (`LeavePolicyCard`, `PolicyLeaveBalanceCard`, `UserLeavePolicies/LeavePolicyCard`) has one, and this PR changes the conditional-rendering branches in two of them. Add a `LeavePolicyCard.test.tsx` (wrapped in the `MockTheme` provider used by the other tests) asserting the three states — unlimited (icon branch, `getByRole("img")` with the expected accessible name), limited (`balance / total` text), and `isBalanceAvailable === false` — which would also have caught the branch collapse flagged in the first pass.

frontend/src/community/common/assets/languages/english/leaveModule.json
```diff
@@ -263,7 +263,7 @@
     "title": "Leave Policies",
     "pageHead": "Leave Policies | Skapp",
     "breadcrumbLeave": "Leave",
-    "breadcrumbLeavePolicies": "Leave policies",
+    "breadcrumbLeavePolicies": "Leave Policies",
```
ThinuwanW, line 266: 🤖 **Claude** · 🟣 **Nit**

After this change `leavePolicies.title` (line 263) and `leavePolicies.breadcrumbLeavePolicies` (line 266) hold the identical string "Leave Policies", and `pages/community/leave/policies/index.tsx` renders both on the same screen (`title={translateText(["title"])}` and the last breadcrumb). The casing fix itself matches the app convention (see `peopleModule.json` breadcrumbs, which are Title Case), but keeping two keys with the same value invites them to drift apart again. Consider dropping `breadcrumbLeavePolicies` and reusing `translateText(["title"])` for the trailing breadcrumb.

akilarootcode (approved):

frontend/src/community/leave/components/molecules/LeavePolicyCard/LeavePolicyCard.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/molecules/LeavePolicyCard/LeavePolicyCard.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/molecules/PolicyLeaveBalanceCard/PolicyLeaveBalanceCard.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/molecules/LeavePolicyCard/LeavePolicyCard.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/molecules/LeavePolicyCard/LeavePolicyCard.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/molecules/LeavePolicyCard/LeavePolicyCard.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/molecules/LeavePolicyCard/LeavePolicyCard.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/molecules/LeavePolicyCard/LeavePolicyCard.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/molecules/LeavePolicyCard/LeavePolicyCard.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/molecules/LeavePolicyCard/LeavePolicyCard.tsx
hasalarootcode: Not needed

frontend/src/community/common/assets/languages/english/leaveModule.json
hasalarootcode: Not needed

akilarootcode (approved):

ThinuwanW (approved):

PR #1803, opened 2026-08-19
fix: fetch original-resolution image for birthday modal profile picture

SanithuGunasena07:
Fixes blurred profile pictures in the birthday celebration modal by fetching the original-resolution S3 image instead of the default thumbnail. The change is scoped to BirthdayModalShell.tsx only (passing isOriginalImage = true to useGetImageUrl), so all other avatars across the app (task rows, deal cards, owner pickers, etc.) are unaffected.

### DIFF
--- a/frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
+++ b/frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
@@ -31,7 +31,7 @@ const BirthdayModalShell: FC<Props> = ({
   showConfetti = false
 }) => {
   const translateAria = useTranslator("peopleAria", "birthdayNotifications");
-  const imageUrl = useGetImageUrl(employee.authPic ?? "");
+  const imageUrl = useGetImageUrl(employee.authPic ?? "", true);
   const [isConfettiVisible, setIsConfettiVisible] = useState(showConfetti);
 
   useEffect(() => {


ThinuwanW (commented):
## 🤖 Claude Code Review

Single-line change passing `isOriginalImage: true` to `useGetImageUrl` so the birthday modal avatar renders at full resolution instead of the 128x128 thumbnail. The intent is sound and the enterprise S3 original path is valid (uploadImage.ts uploads original and thumbnail under the same filename), but the flag does not actually reach the community-mode fetch, and in enterprise mode the shared `s3FileUrls` cache is keyed by filePath alone so the original/thumbnail variants collide with the AppBar avatar. As written the fix is likely a no-op in both modes.

Found **9** new issue(s): 🔴 3 important, 🟡 4 suggestion(s), 🟣 2 nit(s)

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
```diff
@@ -31,7 +31,7 @@ const BirthdayModalShell: FC<Props> = ({
   showConfetti = false
 }) => {
   const translateAria = useTranslator("peopleAria", "birthdayNotifications");
-  const imageUrl = useGetImageUrl(employee.authPic ?? "");
+  const imageUrl = useGetImageUrl(employee.authPic ?? "", true);
```
ThinuwanW, line 34: 🤖 **Claude** · 🔴 **Important**

In community mode this flag has no effect — the modal will still render the 128x128 thumbnail. `useGetImageUrl` (frontend/src/community/common/hooks/useGetImageUrl.ts:20-25) hardcodes the third argument of `useGetUploadedImage` to `true`, which is the `isThumbnail` param that ends up in the query string via `fileUploadEndpoints.DOWNLOAD_IMAGES(type, file, isThumbnail)`. `isOriginalImage` is only consumed by the enterprise `downloadS3File` branch (line 42-50). Since `NEXT_PUBLIC_MODE=community` is the default OSS build, the stated fix silently does nothing there. Fix in the hook: change the call to `useGetUploadedImage(FileTypes.USER_IMAGE, src, !isOriginalImage, environment !== appModes.ENTERPRISE)`. This is safe for React Query caching because `isThumbnail` is already part of the query key `["download-file", type, file, isThumbnail]`, so thumbnail and original responses stay in separate cache entries. Note the same latent gap exists in `~community/common/components/molecules/Avatar/Avatar.tsx:73-78` (used by `EditInfoCard` and `GeneralDetailsSection` with `isOriginalImage={true}`), so fixing the hook consistently is worth doing in both places.

ThinuwanW, line 34: 🤖 **Claude** · 🔴 **Important**

In enterprise mode this collides with the AppBar avatar because the S3 URL cache is keyed only by file path. `useCommonStore.s3FileUrls` (templateSlice.ts:20-23) merges entries under the bare `filePath`, with no distinction between the `profilePictures/original` and `profilePictures/thumbnail` signed URLs. Concrete failure for the SELF birthday modal: `AppBar.tsx:155-164` renders `<Avatar src={employee.authPic}>` with `isOriginalImage` defaulting to false and caches the thumbnail signed URL under `s3FileUrls[authPic]`. The modal mounts later (only after `useGetTodaysBirthdayNotifications` resolves and the queue is seeded in useBirthdayNotifications.ts:110), so by then `downloadS3File` hits the early return at useS3Download.ts:53-59 (`isProfilePic && s3FileUrls[filePath] && !includes('blob')`) and the original is never requested — the modal still shows the thumbnail. If the ordering ever flips (or on a page where the modal wins the race), the reverse happens: the AppBar avatar ends up pointing at the full-resolution original, a bandwidth regression for a 2.5rem avatar. Fix by making the cache key variant-aware, e.g. store under `${filePath}:${isOriginalImage ? 'original' : 'thumbnail'}` in `downloadS3File`/`forceRefetch` and read with the matching key in `useGetImageUrl`/`Avatar`; alternatively have `useGetImageUrl` skip the early-return path when `isOriginalImage` is true and a thumbnail URL is already cached.

ThinuwanW, line 34: 🤖 **Claude** · 🟣 **Nit**

Bare positional boolean at the call site is hard to read — `useGetImageUrl(employee.authPic ?? "", true)` gives no hint what `true` means without opening the hook. Consider either a named local (`const FETCH_ORIGINAL_IMAGE = true;`) or, better, changing the hook signature to an options object (`useGetImageUrl(src, { isOriginalImage: true })`). Low priority, but this hook is called from 8 files and the flag is easy to mis-order.

ThinuwanW, line 34: 🤖 **Claude** · 🔴 **Important**

Latent operator bug in the effect this change now depends on. In `frontend/src/community/common/hooks/useGetImageUrl.ts` the download effect guards with `if (src || !s3FileUrls[src])`. Because of the `||`, the guard is true whenever `src` is truthy (and also true when `src` is `""`, since `!s3FileUrls[""]` is true), so the cache check is effectively dead code and `downloadS3File` is invoked on every mount — including with `filePath: ""` for employees with no photo (very common here, since line 34 passes `employee.authPic ?? ""` and `authPic` is typed `string | null`). It should be `if (src && !s3FileUrls[src])`. Additionally the dependency array is `[src, isOriginalImage]` while the effect closes over `s3FileUrls` and `downloadS3File`, so it reads a stale cache snapshot. Fix: change `||` to `&&` and add the missing dependencies (wrapping `downloadS3File` in `useCallback` in `useS3Download` so it is referentially stable). This matters more now that the modal is asking for a different variant of the same key than everything else in the app.

ThinuwanW, line 34: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

The fix is applied to only one of two duplicated implementations. `frontend/src/community/common/hooks/useGetImageUrl.ts` (lines 14-52) and `frontend/src/community/common/components/molecules/Avatar/Avatar.tsx` (lines 67-103) contain byte-identical logic: the same `useS3Download()` destructure, the same `useGetUploadedImage(FileTypes.USER_IMAGE, src, true, environment !== appModes.ENTERPRISE)` call, the same community/enterprise `useEffect` with `notificationDefaultImage` special-casing, and the same `isOriginalImage` plumbing. This violates the '2+ hooks with same logic -> extract custom hook' guideline and means every other large avatar in the app resolves images through the `Avatar` copy while the birthday modal now goes through the hook copy. Fix: delete the duplicated state/effects from `Avatar.tsx` and have it call `useGetImageUrl(src, isOriginalImage)`, so future resolution fixes apply everywhere at once.

ThinuwanW, line 34: 🤖 **Claude** · 🟡 **Suggestion**

Unbounded payload with no loading state. The profile-picture dropzone in `frontend/src/community/people/components/organisms/PersonDetailsSection/SubSections/GeneralDetailsSections.tsx` (lines 162-172) sets `accept` but no `maxSize`, and `uploadImage.ts` uploads `authPic[0]` raw to `profilePictures/original`, so the 'original' can be a multi-megabyte phone photo. Requesting it here replaces a ~128x128 thumbnail (single-digit KB) with a potentially several-MB download for one modal avatar, and `BirthdayModalShell` renders no skeleton/blur-up while it loads — `image` is briefly set to the raw S3 key (`useGetImageUrl` line 36: `setImage(s3FileUrls[src] ?? src)`), which resolves as a relative URL and 404s until the signed URL arrives. Consider generating a mid-size (e.g. 512px) derivative at upload alongside the 128px thumbnail and requesting that instead, or at minimum add a `maxSize` to the dropzone and a loading placeholder to the modal avatar.

ThinuwanW, line 34: 🤖 **Claude** · 🟡 **Suggestion**

Aspect-ratio behavioral drift specific to this modal. `frontend/src/community/people/utils/image/thumbnailGenerator.ts` deliberately center-crops uploads to a 128x128 square before storing them under `profilePictures/thumbnail`, whereas the file under `profilePictures/original` retains the user's arbitrary aspect ratio (portrait, 16:9, etc.). Every other avatar in the app renders the pre-cropped square; this modal will now render an uncropped image, so correct framing depends entirely on the `@rootcodelabs/skapp-ui` `Avatar` applying `object-fit: cover` to its ``. Please verify that component's styling for `size="2xl"`; if it does not cover-crop, non-square uploads will letterbox or distort here only. If unsure, add `sx`/class-level `object-fit: cover` at this call site.

ThinuwanW, line 34: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

No test coverage for the changed behavior. There is no `BirthdayModalShell.test.tsx` (nor any test for `ColleagueBirthdayModal`, `SelfBirthdayModal`, or `useGetImageUrl`) — the only Birthday tests are `birthdayNotificationUtils.test.ts` and `birthdayNotificationCacheUtils.test.ts`, which cover pure utils. Given this PR changes which image variant a user actually sees, add a `BirthdayModalShell.test.tsx` using `@testing-library/react` with the `MockTheme` wrapper that mocks `~community/common/hooks/useGetImageUrl` and asserts it is called with `(authPic, true)`, plus a case where `employee.authPic` is `null` asserting the initials fallback renders (`screen.getByRole("img", { name: ... })`). Without this, a future refactor silently reverts to the thumbnail with no signal.

ThinuwanW, line 34: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

The community fallback stub cannot carry this flag, and its signature does not type-check against the caller. `frontend/src/fallback/common/hooks/useS3Download.ts` declares `const downloadS3File = (key: string) => {}`, but `useGetImageUrl` calls it with an object (`{ filePath, isProfilePic, isOriginalImage }`). Since `tsconfig.json` maps `~enterprise/*` to `["./src/enterprise/*", "./src/fallback/*"]` and `jest.config.js` maps `^~enterprise(.*)$` to `<rootDir>/src/fallback$1` unconditionally, a community-only checkout (enterprise submodule absent) resolves to the stub and `tsc` should reject the argument, and any Jest test necessarily exercises the stub. Fix: align the fallback stub with the enterprise signature — `const downloadS3File = (_params: { filePath: string; isProfilePic?: boolean; isOriginalImage?: boolean }) => {}` — and export `forceRefetch`/`downloadESignatureFiles` stubs too, so the two modes stay contract-compatible and the new flag is at least assertable in tests.

SanithuGunasena07: not required - rechecked

SanithuGunasena07: not required

SanithuGunasena07: Pre existing out of scope for PR

SanithuGunasena07: not required

SanithuGunasena07: not required

SanithuGunasena07: Pre existing out of scope for PR

SanithuGunasena07: requirement needs original

SanithuGunasena07: not required

SanithuGunasena07: not required

hasalarootcode (approved):

akilarootcode (approved):

ThinuwanW (approved):

PR #1804, opened 2026-08-19
Fix/leave policy balance issue

hasalarootcode:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


shakila-kamalasena (dismissed):

shakila-kamalasena (dismissed):

akilarootcode (dismissed):

ThinuwanW (commented):
## 🤖 Claude Code Review

This PR changes policy-leave accrual from "full cycle" to "as-of" so cards can show accrued-to-date, and removes the now-redundant policy/leave year sync on the my-requests page. The backend change is only half applied: `totalDaysAllocated` is now measured as of a point in time while `totalDaysUsed` is still summed over the entire cycle, so any employee with committed future-dated leave gets an understated (often negative) balance and spurious INSUFFICIENT_BALANCE / FULLY_UTILIZED outcomes; the apply/availability path also uses a different as-of date than the balance card, so validation and display disagree. The frontend cleanup is consistent but leaves `usePolicyLeaveStore.setSelectedYear` dead and permanently freezes the policy year to the browser's calendar year.

Found **15** new issue(s): 🔴 5 important, 🟡 6 suggestion(s), 🟣 4 nit(s)

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
```diff
@@ -781,8 +791,9 @@ private PolicyLeaveBalanceDto calculateBalance(EmployeeLeavePolicy assignment, P
 		}
 
 		PolicyLeaveUsageLookup usageLookup = usageLookupSupplier.get();
-		float accruedDays = accruedInCycle(assignment, cycle);
-		float accrualAllocation = PolicyLeaveAccrualUtil.accrualAllocationInCycle(policy, effectiveFrom, cycle);
+		float accruedDays = accruedInCycle(assignment, cycle, accrualAsOf);
+		float accrualAllocation = PolicyLeaveAccrualUtil.accrualAllocationAsOf(policy, effectiveFrom, cycle,
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Allocation and usage are now measured over different windows. `accrualAllocation` is computed as of `accrualAsOf` (today for the current cycle), but two lines below `totalDaysUsed = usageLookup.usedBetween(cycle.getStartDate(), cycle.getEndDate())` still sums the WHOLE cycle, including approved/pending leave dated after today. `balanceInDays = totalDaysAllocated - totalDaysUsed` therefore mixes a partial numerator with a full-cycle denominator.

Concrete failure: monthly accrual of 1 day (12/yr), today = 20 Feb, employee has an APPROVED 5-day leave in July. Before: 12 - 5 = 7. Now: 2 - 5 = -3. The card renders `-3 / 2` (getPolicyBalanceLabel returns `balanceInDays.toString()` verbatim), resolveDisabledReason returns FULLY_UTILIZED, and any new request fails with LEAVE_ERROR_POLICY_LEAVE_INSUFFICIENT_BALANCE quoting a negative day count. The same understated balance flows to the admin view via EmployeeLeavePolicyServiceImpl.getEmployeeLeavePolicies.

Suggested fix: keep `totalDaysAllocated`/`balanceInDays` on the full-cycle basis (accrualAllocationInCycle) so allocation and usage share the same window, and surface the accrued-to-date figure only through the existing `accruedDays` field (already on PolicyLeaveBalanceDto/EmployeePolicyBalanceResponseDto and currently unused by the UI). If the product intent really is to gate on accrued-to-date, then usage must be split into "used up to asOf" vs "committed after asOf" and the balance clamped at >= 0, and the UI denominator changed accordingly.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
```diff
@@ -740,20 +741,28 @@ private EmployeeLeavePolicy resolveActiveAssignment(Employee employee, Long poli
...
-		return calculateBalance(assignment, cycle, cycleAnchor, clampToCycle(today, cycle));
+		return calculateBalance(assignment, cycle, cycleAnchor, clampToCycle(today, cycle),
+				resolveAccrualAsOf(today, cycle));
 	}
 
 	private PolicyLeaveBalanceDto calculateBalanceForDate(EmployeeLeavePolicy assignment, LocalDate date) {
 		MonthDay cycleAnchor = resolveCycleAnchor();
 		return calculateBalance(assignment, PolicyLeaveAccrualUtil.resolveCycleContaining(date, cycleAnchor),
-				cycleAnchor, date);
+				cycleAnchor, date, date);
```
ThinuwanW, line 755: 🤖 **Claude** · 🔴 **Important**

`calculateBalanceForDate` passes the request's `date` as `accrualAsOf`, while every display path uses `resolveAccrualAsOf(today, cycle)`. That makes the balance used for validation (checkPolicyLeaveAvailability and applyPolicyLeaveRequest both call this) diverge from the balance shown on the card, in both directions:
- Future-dated request inside the current cycle: accrual is projected to the leave start date, so the request is validated against days the employee has not accrued yet — more permissive than the card, which is what disables the Apply button in the first place (LeavePolicyCard gates on `isDisabled`).
- Back-dated request inside the current cycle: accrual is truncated to the past start date while usage is still whole-cycle (see the finding on line 795), so a legitimate backdated request is rejected as INSUFFICIENT_BALANCE.
- Request in the next cycle: the balance card for that year projects the FULL cycle accrual (`resolveAccrualAsOf` returns `cycle.getEndDate()` when `today` is before cycle start), but this path returns accrual as of the request date, which for a PERIOD_END monthly policy is 0 on 5 Jan — card says days are available, apply says insufficient.

Fix: call `resolveAccrualAsOf(today, cycle)` here as well (passing `DateTimeUtils.getCurrentUtcDate()`), so the validation balance is exactly the balance the user was shown. If projecting to the leave date is deliberate, apply the same rule in `resolveAccrualAsOf` and in the card gating so the two agree.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
```diff
@@ -740,20 +741,28 @@ private EmployeeLeavePolicy resolveActiveAssignment(Employee employee, Long poli
...
 	}
 
 	private MonthDay resolveCycleAnchor() {
 		LeaveCycleDetailsDto leaveCycle = leaveCycleService.getLeaveCycleConfigs();
 		return MonthDay.of(leaveCycle.getStartMonth(), leaveCycle.getStartDate());
 	}
 
+	private LocalDate resolveAccrualAsOf(LocalDate today, PolicyLeaveDateWindowDto cycle) {
```
ThinuwanW, line 763: 🤖 **Claude** · 🔴 **Important**

New zero-accrual state is reported to users as "fully utilized". With as-of accrual, a policy that credits at PERIOD_END (or has a waiting period, or was just assigned) accrues 0 days at the start of a cycle, so `balanceInDays == 0` -> `hasBalance()` false -> `resolveDisabledReason` (line 558) returns FULLY_UTILIZED -> the card is disabled and LeavePolicyCard shows the `fullyUtilizedError` toast even though the employee has used nothing. Add a distinct reason (e.g. `NOT_YET_ACCRUED`) to PolicyBalanceDisabledReason, map it in `getDisabledReasonToastKeys`/`policyLeaveUtils` with its own copy, and pick it when `totalDaysUsed == 0 && totalDaysAllocated == 0`. Also please add a short javadoc on `resolveAccrualAsOf` explaining why a future cycle projects to `cycle.getEndDate()` while the current cycle stops at today — the two branches look like a bug otherwise, and the method duplicates `clampToCycle` for everything except that one case.

backend/src/main/java/com/skapp/community/leaveplanner/util/PolicyLeaveAccrualUtil.java
```diff
@@ -162,9 +162,13 @@ private static float unusedAtCycleEnd(LeavePolicy policy, LocalDate effectiveFro
 
 	public static float accrualAllocationInCycle(LeavePolicy policy, LocalDate effectiveFrom,
 			PolicyLeaveDateWindowDto cycle) {
+		return accrualAllocationAsOf(policy, effectiveFrom, cycle, cycle.getEndDate());
+	}
+
+	public static float accrualAllocationAsOf(LeavePolicy policy, LocalDate effectiveFrom,
+			PolicyLeaveDateWindowDto cycle, LocalDate asOf) {
```
ThinuwanW, line 169: 🤖 **Claude** · 🟡 **Suggestion**

This accrual math change ships with no test coverage: there are no unit tests for `PolicyLeaveAccrualUtil` or `PolicyLeaveServiceImpl` anywhere under backend/src/test, and the existing policy integration tests (LeavePolicyControllerIntegrationTest, EmployeeLeavePolicyControllerIntegrationTest) do not assert accrued/allocated day counts. Given the change silently alters every balance figure and the leave-application gate, please add unit tests for `accrualAllocationAsOf` covering: mid-cycle as-of with MONTHLY PERIOD_END vs PERIOD_START, as-of before `resolveAccrualStartDate` (waiting period) expecting 0, as-of after cycle end (should equal the full-cycle value), a future cycle via `resolveAccrualAsOf`, and the accrual-cap interaction. Add a service-level test asserting the balance for an employee holding an approved future-dated leave (the negative-balance case above).

frontend/pages/community/leave/my-requests.tsx
```diff
@@ -37,14 +36,9 @@ const MyRequests: NextPage = () => {
...
 
-  const { data: nextYearPolicyBalances } = useGetMyPolicyBalances(
-    nextYear.toString(),
-    isLeavePoliciesEnabled
-  );
-
-  const isNextYearAvailable = isLeavePoliciesEnabled
-    ? (nextYearPolicyBalances?.length ?? 0) > 0
-    : (isEntitlementAvailableNextYear?.length ?? 0) > 0;
+  const isNextYearAvailable =
```
ThinuwanW, line 39: 🤖 **Claude** · 🟡 **Suggestion**

With `isNextYearAvailable` now hard-false in policy mode, the year dropdown never renders for policy users, and since `usePolicyLeaveYearSync` is deleted nothing ever calls `usePolicyLeaveStore.setSelectedYear`. Two consequences worth cleaning up:
1. Dead code: `setSelectedYear` in `frontend/src/community/leave/store/policyLeaveStore.ts` (and its declaration in the `PolicyLeaveStore` interface) is now unreachable — remove it, or add the policy-mode year selector that would call it.
2. `selectedYear` is now permanently the store's initial value, `DateTime.now().year.toString()` — the browser's local calendar year, evaluated once at module load. The backend resolves cycles from the org leave-cycle anchor (`resolveCycleAnchor`/`resolveCycleContaining`), so for an org whose cycle starts in e.g. April, the frontend sends 2026 in Feb 2026 while the current cycle is 2025, and the user is shown the wrong cycle with no way to change it (previously they could at least switch via the dropdown). Consider deriving the year from the leave-cycle config returned by the backend instead of `DateTime.now().year`.

frontend/src/community/leave/components/molecules/LeavePolicyAllocation/LeavePolicyAllocation.tsx
```diff
@@ -70,10 +70,6 @@ const LeavePolicyAllocation: FC = () => {
     setCurrentPage(1);
```
ThinuwanW, line 70: 🤖 **Claude** · 🟣 **Nit**

Removing the `setCurrentPage(1)` effect keyed on `selectedYear` is only safe because `selectedYear` can no longer change (nothing calls `setSelectedYear` any more). It is a latent bug if a policy year selector is reintroduced: switching year would refetch a shorter balance list while `currentPage` stays on page 2+, and `currentAllocations` would slice past the end and render an empty grid. Safer alternative that does not depend on that assumption: clamp in the memo/derived state, e.g. reset when `currentPage > totalPages` (or `useMemo` a `safePage = Math.min(currentPage, Math.max(totalPages, 1))`) rather than relying on an effect per input.

backend/src/main/java/com/skapp/community/leaveplanner/util/PolicyLeaveAccrualUtil.java
```diff
@@ -162,9 +162,13 @@ private static float unusedAtCycleEnd(LeavePolicy policy, LocalDate effectiveFro
 
 	public static float accrualAllocationInCycle(LeavePolicy policy, LocalDate effectiveFrom,
 			PolicyLeaveDateWindowDto cycle) {
+		return accrualAllocationAsOf(policy, effectiveFrom, cycle, cycle.getEndDate());
```
ThinuwanW, line 165: 🤖 **Claude** · 🟣 **Nit**

`accrualAllocationInCycle` is now only referenced from `unusedAtCycleEnd` inside this same utility class (the service call site moved to `accrualAllocationAsOf`). Narrow it to `private static` so the public surface reflects the single supported entry point, or drop it and inline `accrualAllocationAsOf(policy, effectiveFrom, cycle, cycle.getEndDate())` at line 156 with a comment noting that prior cycles are always evaluated at cycle end.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
```diff
@@ -423,12 +423,13 @@ public Map<Long, PolicyLeaveBalanceDto> calculateBalancesForYear(Long employeeId
...
 				cycleAnchor);
 		LocalDate asOf = clampToCycle(today, cycle);
+		LocalDate accrualAsOf = resolveAccrualAsOf(today, cycle);
 
 		Map<Long, PolicyLeaveUsageLookup> usageLookups = buildUsageLookups(employeeId, assignments, cycle, cycleAnchor);
 
 		Map<Long, PolicyLeaveBalanceDto> balancesByAssignment = new LinkedHashMap<>();
-		assignments.forEach(assignment -> balancesByAssignment.put(assignment.getId(),
-				calculateBalance(assignment, cycle, cycleAnchor, asOf, () -> usageLookups.get(assignment.getId()))));
+		assignments.forEach(assignment -> balancesByAssignment.put(assignment.getId(), calculateBalance(assignment,
```
ThinuwanW, line 433: 🤖 **Claude** · 🔴 **Important**

`calculateBalancesForYear` has a second consumer that this PR does not appear to have considered: `EmployeeLeavePolicyServiceImpl.getEmployeeLeavePolicies` calls it with `year = null` and copies `balance.getTotalDaysAllocated()` into `EmployeeLeavePolicyResponseDto`. That value is rendered on the employee-profile / admin "assigned leave policies" card as `balanceInDays / totalDaysAllocated` (frontend `UserLeavePolicies/LeavePolicyCard.tsx` lines 43 and 50, i18n key `leaveModule.leavePolicyAssignment.leavesRemainingLabel`). With `accrualAsOf` now resolving to today, a leave admin opening an employee's profile in February sees "3.5 / 3.5" for a monthly-accrual policy that allocates 21 days a year, which reads as a mis-configured or exhausted assignment rather than "accrued so far". Either keep the full-cycle allocation on that path (pass `cycle.getEndDate()` as `accrualAsOf` from the admin call site, or expose both the accrued-to-date and full-cycle figures on the DTO), or ship the paired frontend/i18n change so the denominator is labelled as accrued-to-date. Right now one code change silently redefines two unrelated screens.

frontend/pages/community/leave/my-requests.tsx
ThinuwanW, line 39: 🤖 **Claude** · 🔴 **Important**

After this PR the policy-mode year can never change: `isNextYearAvailable` is hard-false so the dropdown never renders, and deleting `usePolicyLeaveYearSync` leaves `usePolicyLeaveStore.setSelectedYear` with zero callers anywhere in the repo — so `selectedYear` is frozen at `DateTime.now().year.toString()`, evaluated once at store-module load, in the browser's local timezone. `ApiEndpoints.GET_MY_POLICY_BALANCES` always appends `?year=<that value>`, and the backend's own OpenAPI description on `PolicyLeaveController` says: "The year identifies the leave cycle that starts in that year, which is not the calendar year when the organization's leave cycle does not start on January 1. Omitting it returns the cycle containing today." So for an org with, say, an April 1 anchor, from January to March the client asks for the cycle starting in the *current calendar year* — a cycle that has not begun. `resolveAccrualAsOf` then hits its new `today.isBefore(cycle.getStartDate())` branch and returns `cycle.getEndDate()`, so the card shows the full next-cycle allocation, `resolveDisabledReason` leaves it enabled, and any apply attempt for today fails `OUTSIDE_POLICY_PERIOD` because `usableFrom` is April 1. The same wrong window is used for the My Requests table (`getCurrentUserPolicyLeaveRequests` resolves the cycle from the same year value). Fix: stop sending `year` for the default view — the param is `@RequestParam(required = false)` and the service already resolves the cycle containing today — using a fixed sentinel in the `MY_POLICY_BALANCES` query key; or derive the cycle year from the leave-cycle config instead of `DateTime.now().year`. Also delete the now-unreachable `setSelectedYear` from `policyLeaveStore` (or restore a year control) so the dead setter does not read as a supported capability.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
```diff
@@ -806,12 +817,12 @@ private PolicyLeaveBalanceDto calculateBalance(EmployeeLeavePolicy assignment, P
 		return balance;
 	}
 
-	private float accruedInCycle(EmployeeLeavePolicy assignment, PolicyLeaveDateWindowDto cycle) {
+	private float accruedInCycle(EmployeeLeavePolicy assignment, PolicyLeaveDateWindowDto cycle, LocalDate asOf) {
```
ThinuwanW, line 823: 🤖 **Claude** · 🟡 **Suggestion**

`accruedInCycle` is now byte-for-byte the same computation as `PolicyLeaveAccrualUtil.accrualAllocationAsOf` minus the cap: both call `resolveAccrualStartDate(policy, assignment.getEffectiveFrom())` and then `roundToHalfDay(accruedWithinCycle(policy, accrualStartDate, cycle, accrualAsOf))` with the identical `accrualAsOf` that this PR threads through. So `calculateBalance` computes the same thing twice per assignment, and each `accruedWithinCycle` internally makes two `accruedUpTo` calls that walk accrual periods forward from the employee's accrual start date — for a DAILY policy on a long-tenured employee that is thousands of loop iterations, doubled for no benefit, and repeated per assignment on the balances endpoint. Collapse to: `float accruedDays = accruedInCycle(assignment, cycle, accrualAsOf); float accrualAllocation = PolicyLeaveAccrualUtil.applyAccrualCap(policy, accruedDays);` and drop the redundant util call. That also removes the class of bug where a future edit changes one expression and not the other.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
ThinuwanW, line 763: 🤖 **Claude** · 🟡 **Suggestion**

`today` here is always `DateTimeUtils.getCurrentUtcDate()` (i.e. `LocalDate.now(ZoneOffset.UTC)`), but this same class already reads `organizationService.getOrganizationTimeZone()` for working-day math in `calculateRequestedDays`. Before this PR `today` only affected coarse clamping and carryover expiry; now it directly determines the allocation number rendered on every card *and* the balance that `canAccommodate` validates against on the apply path. For an org in UTC+13, on the first day of a new accrual period the card understates the allocation and a legitimate request can be rejected as `INSUFFICIENT_BALANCE` for the first ~11-13 hours of local business time; for UTC-negative zones the accrual appears a day early. Resolve "today" in the organization's timezone (e.g. `LocalDate.now(DateTimeUtils.resolveZoneId(organizationService.getOrganizationTimeZone()))`) at the `getCurrentUserPolicyBalances` / `calculateBalancesForYear` entry points so the as-of date matches the calendar the employee is actually looking at.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
```diff
@@ -762,13 +771,14 @@ private LocalDate clampToCycle(LocalDate date, PolicyLeaveDateWindowDto cycle) {
...
-			MonthDay cycleAnchor, LocalDate asOf) {
-		return calculateBalance(assignment, cycle, cycleAnchor, asOf,
+			MonthDay cycleAnchor, LocalDate asOf, LocalDate accrualAsOf) {
+		return calculateBalance(assignment, cycle, cycleAnchor, asOf, accrualAsOf,
 				() -> buildUsageLookup(assignment, cycle, cycleAnchor));
 	}
 
 	private PolicyLeaveBalanceDto calculateBalance(EmployeeLeavePolicy assignment, PolicyLeaveDateWindowDto cycle,
-			MonthDay cycleAnchor, LocalDate asOf, Supplier<PolicyLeaveUsageLookup> usageLookupSupplier) {
+			MonthDay cycleAnchor, LocalDate asOf, LocalDate accrualAsOf,
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`calculateBalance` now takes two adjacent `LocalDate` parameters named `asOf` and `accrualAsOf` whose values deliberately diverge — for a not-yet-started cycle `asOf` is `cycle.getStartDate()` (via `clampToCycle`) while `accrualAsOf` is `cycle.getEndDate()` (via the new `resolveAccrualAsOf`) — and nothing documents that. `asOf` feeds only `usableCarryoverDays`, `accrualAsOf` feeds only the accrual math; swapping them at any of the three call sites compiles cleanly and silently changes carryover-expiry and allocation results. Please add a javadoc on `calculateBalance` and `resolveAccrualAsOf` stating what each date means and why the future-cycle branch returns the cycle end, and consider renaming `asOf` to something intention-revealing (e.g. `carryoverAsOf`) so the two are not distinguishable only by a prefix.

frontend/src/community/leave/components/molecules/LeavePolicyAllocation/LeavePolicyAllocation.tsx
ThinuwanW, line 70: 🤖 **Claude** · 🟡 **Suggestion**

Separate from the year-change concern: nothing clamps `currentPage` when the length of `policyBalances` changes. `useApplyPolicyLeave` invalidates `MY_POLICY_BALANCES(year)` on success, and policies can also be unassigned, so the list can shrink while the user sits on page 2. `totalPages` then drops below `currentPage`, `currentAllocations` slices past the end and returns an empty array, and the grid renders blank — `handleNextPage`/`handlePreviousPage` both no-op or are disabled at that point, so the only recovery is a reload. Add a guard keyed on the data rather than on the year, e.g. `useEffect(() => { setCurrentPage((p) => Math.min(p, Math.max(1, totalPages))); }, [totalPages]);` or clamp inside the `currentAllocations` memo.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
```diff
@@ -781,8 +791,9 @@ private PolicyLeaveBalanceDto calculateBalance(EmployeeLeavePolicy assignment, P
 		}
 
 		PolicyLeaveUsageLookup usageLookup = usageLookupSupplier.get();
-		float accruedDays = accruedInCycle(assignment, cycle);
-		float accrualAllocation = PolicyLeaveAccrualUtil.accrualAllocationInCycle(policy, effectiveFrom, cycle);
+		float accruedDays = accruedInCycle(assignment, cycle, accrualAsOf);
```
ThinuwanW, line 798: 🤖 **Claude** · 🟣 **Nit**

Threading `accrualAsOf` into `accruedInCycle` changes `balance.accruedDays`, which is mapped to `EmployeePolicyBalanceResponseDto.accruedDays` — but no frontend consumer reads it: `accruedDays` appears only as a field declaration in `PolicyLeaveTypes.ts` and is never rendered. So half of this change is invisible, while the user-visible number is `totalDaysAllocated`. Note also that `accruedDays` is uncapped whereas `totalDaysAllocated` is `applyAccrualCap`'d, so for any policy with `accrualCapDays` set the response now advertises `accruedDays > totalDaysAllocated`. Either surface `accruedDays` in the UI (it is the figure that actually explains the new denominator to the user) or drop it from the response DTO rather than shipping an inconsistent, unread field.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
```diff
@@ -423,12 +423,13 @@ public Map<Long, PolicyLeaveBalanceDto> calculateBalancesForYear(Long employeeId
 		PolicyLeaveDateWindowDto cycle = PolicyLeaveAccrualUtil.resolveCycle(resolveCycleYear(year, today, cycleAnchor),
```
ThinuwanW, line 425: 🤖 **Claude** · 🟣 **Nit**

`calculateBalancesForYear` is the only public `@Override` service method in this class without the convention entry log — every sibling (`getCurrentUserPolicyBalances`, `checkPolicyLeaveAvailability`, `applyPolicyLeaveRequest`, …) opens with `log.info("<methodName>: execution started")`. Since this PR is changing what this method computes and it is now the shared entry point for both the employee balances endpoint and the admin `getEmployeeLeavePolicies` view, add `log.info("calculateBalancesForYear: execution started")` (and the matching ended line) so the two callers are distinguishable in logs when a balance figure is disputed.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
hasalarootcode: Keeping the allocation on the full cycle is the bug this PR fixes. A 1-day-a-month policy was showing 12 days available in January, so the accrued-to-date denominator is intentional and I do not want to revert it.

The negative balance from a future-dated approved leave is real though. The fix I would make is to split usage into used-up-to-asOf and committed-after-asOf, and clamp balanceInDays at 0. Here or a follow-up?

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
hasalarootcode: Projecting accrual to the leave start date is deliberate: a July request should be checked against what will have accrued by July, not against February. The next-cycle PERIOD_END case where the card and the apply path disagree is real.

Do you want validation pinned to today accrual via resolveAccrualAsOf instead, so the two always agree, at the cost of rejecting future-dated requests inside the current cycle?

backend/src/main/java/com/skapp/community/leaveplanner/util/PolicyLeaveAccrualUtil.java
hasalarootcode: Unit tests on accrualAllocationAsOf fit the convention here since it is a util, so that is a new PolicyLeaveAccrualUtilUnitTest with the five cases you listed. The service-level assertion for the approved-future-leave case would have to go in the policy integration tests rather than a service unit test. Add both in this PR, or a follow-up?

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
hasalarootcode: Accrued-to-date on that card is the intent too. A monthly policy showing 21/21 in February is the thing being fixed. The balanceInDays / totalDaysAllocated label does read wrong without the paired copy change though. I would rather relabel it on the frontend than special-case the admin call site back to full cycle. Should that FE and i18n change ride in this PR?

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
hasalarootcode: resolveCycleYear and the request-list path both run off getCurrentUtcDate as well, so moving only the balance as-of to the org timezone would leave the cycle resolution and the accrual disagreeing inside one request. Worth doing as one pass over the class. Separate PR?

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
hasalarootcode: accruedDays exceeding totalDaysAllocated when a cap is set is real. Surfacing accruedDays on the card is the better answer since it is the number that explains the new denominator, but that is an FE plus copy change. Drop the field for now and add it back with the UI, or ship the UI here?

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
hasalarootcode: Fixed.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
hasalarootcode: Added.

backend/src/main/java/com/skapp/community/leaveplanner/util/PolicyLeaveAccrualUtil.java
hasalarootcode: Fixed. accrualAllocationAsOf went private with it, since collapsing the duplicate computation in calculateBalance removed its only external caller.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
hasalarootcode: Renamed asOf to carryoverAsOf across the three call sites and added the javadoc on resolveAccrualAsOf explaining the future-cycle branch. The per-parameter javadoc on calculateBalance itself is not in yet.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
hasalarootcode: Added the javadoc on resolveAccrualAsOf. NOT_YET_ACCRUED is a fair distinction but it needs the enum value plus the FE mapping and its own copy in policyLeaveUtils, so it is more than a backend change. In this PR or a follow-up?

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
hasalarootcode: Not needed

frontend/src/community/leave/components/molecules/LeavePolicyAllocation/LeavePolicyAllocation.tsx
hasalarootcode: Not needed

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
hasalarootcode: Not needed

frontend/pages/community/leave/my-requests.tsx
hasalarootcode: Not needed

backend/src/main/java/com/skapp/community/leaveplanner/util/PolicyLeaveAccrualUtil.java
hasalarootcode: Not needed

frontend/src/community/leave/components/molecules/LeavePolicyAllocation/LeavePolicyAllocation.tsx
hasalarootcode: Not needed

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/PolicyLeaveServiceImpl.java
hasalarootcode: Not needed

frontend/pages/community/leave/my-requests.tsx
hasalarootcode: Not needed

shakila-kamalasena (dismissed):

akilarootcode (approved):

ThinuwanW (approved):

PR #1805, opened 2026-08-19
fix: use tertiary background and 14px font for request table chips

Deshan1100:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/frontend/src/community/attendance/components/molecules/ManagerTimesheetRequestTable/ManagerTimesheetRequestTable.tsx
+++ b/frontend/src/community/attendance/components/molecules/ManagerTimesheetRequestTable/ManagerTimesheetRequestTable.tsx
@@ -176,7 +176,9 @@ const ManagerTimesheetRequestTable: FC<Props> = ({
           avatarUrl={timesheetRequest?.employee?.authPic as string}
           isResponsiveLayout={true}
           chipStyles={{
-            maxWidth: "15.625rem"
+            width: "fit-content",
+            maxWidth: "15.625rem",
+            backgroundColor: "var(--color-tertiary-background)"
           }}
         />
       ),
--- a/frontend/src/community/leave/components/molecules/LeaveRequestRow/RequestDates.tsx
+++ b/frontend/src/community/leave/components/molecules/LeaveRequestRow/RequestDates.tsx
@@ -1,4 +1,4 @@
-import { Box, Theme, Typography, useTheme } from "@mui/material";
+import { Box, Typography } from "@mui/material";
 import { FC } from "react";
 
 import { useTranslator } from "~community/common/hooks/useTranslator";
@@ -8,7 +8,6 @@ interface Props {
   days: string;
 }
 const RequestDates: FC<Props> = ({ dates, days }) => {
-  const theme: Theme = useTheme();
   const translateAria = useTranslator("leaveAria", "allLeaveRequests");
 
   return (
@@ -40,8 +39,9 @@ const RequestDates: FC<Props> = ({ dates, days }) => {
         {dates}
       </Typography>
       <div
+        className="body2"
         style={{
-          backgroundColor: theme.palette.common.white,
+          backgroundColor: "var(--color-tertiary-background)",
           borderRadius: "9.375rem",
           padding: "0.5rem 1rem"
         }}
--- a/frontend/src/community/leave/components/molecules/ManagerLeaveRequests/ManagerLeaveRequest.tsx
+++ b/frontend/src/community/leave/components/molecules/ManagerLeaveRequests/ManagerLeaveRequest.tsx
@@ -32,7 +32,7 @@ import { requestTypeSelector } from "~community/leave/utils/LeaveRequestFilterAc
 import { generateManagerLeaveRequestAriaLabel } from "~community/leave/utils/accessibilityUtils";
 
 const chipClassName =
-  "inline-flex w-fit items-center gap-2 rounded-full bg-tertiary-background px-4 py-2";
+  "body2 inline-flex w-fit items-center gap-2 rounded-full bg-tertiary-background px-4 py-2";
 
 interface Props {
   employeeLeaveRequests: LeaveRequestItemsType[];


ThinuwanW (commented):
## 🤖 Claude Code Review

Small styling-only PR that switches request-table chips to the tertiary background token and applies the global `body2` (14px) typography class. The changes are consistent with the existing `var(--color-tertiary-background)` / `.body2` conventions and correctly drop the now-unused `useTheme`/`Theme` imports in RequestDates. Two cross-consumer issues matter: the shared `RequestDates` pill is also rendered in the legacy `Table`-based Team Leave Analytics history table, whose row/hover backgrounds are #FAFAFA/#F4F4F5, so the pill becomes nearly (and on hover exactly) the same color as the row; and two byte-identical copies of the `chipClassName` constant were left un-updated, so the sibling policy leave-request table keeps the old font size.

Found **11** new issue(s): 🔴 5 important, 🟡 2 suggestion(s), 🟣 4 nit(s)

frontend/src/community/leave/components/molecules/LeaveRequestRow/RequestDates.tsx
```diff
@@ -40,8 +39,9 @@ const RequestDates: FC<Props> = ({ dates, days }) => {
         {dates}
       </Typography>
       <div
+        className="body2"
         style={{
-          backgroundColor: theme.palette.common.white,
+          backgroundColor: "var(--color-tertiary-background)",
```
ThinuwanW, line 44: 🤖 **Claude** · 🔴 **Important**

This shared component is not only used by the new TableView-based tables — it is also rendered by TeamAnalyticsLeaveHistoryTable (frontend/src/community/leave/components/molecules/TeamAnalyticsLeaveHistoryTable/TeamAnalyticsLeaveHistoryTable.tsx:104), which uses the legacy `Table` molecule. In that table, body rows are `theme.palette.grey[50]` (#FAFAFA) and the always-applied `row.active` hover style is `theme.palette.grey[100]` (#F4F4F5) — see frontend/src/community/common/components/molecules/Table/styles.ts:170 and :200. `--color-tertiary-background` is also #F4F4F5, so the day-count pill goes from clearly visible white to near-invisible at rest and completely invisible on row hover. Same applies to PolicyManagerLeaveRequests only if it ever moves to the legacy table (it currently uses TableView, so it is fine). Fix options: (a) add an optional `backgroundColor`/`variant` prop to RequestDates and keep white for the analytics table, or (b) use a token that contrasts with grey[50]/grey[100] (e.g. `--color-white` for the pill or `--color-secondary-accent` #E4E4E7), or (c) update the legacy Table row/hover backgrounds so the pill still reads. Please verify visually on the Team Leave Analytics page, including row hover.

frontend/src/community/leave/components/molecules/ManagerLeaveRequests/ManagerLeaveRequest.tsx
```diff
@@ -32,7 +32,7 @@ import { requestTypeSelector } from "~community/leave/utils/LeaveRequestFilterAc
 import { generateManagerLeaveRequestAriaLabel } from "~community/leave/utils/accessibilityUtils";
 
 const chipClassName =
-  "inline-flex w-fit items-center gap-2 rounded-full bg-tertiary-background px-4 py-2";
+  "body2 inline-flex w-fit items-center gap-2 rounded-full bg-tertiary-background px-4 py-2";
```
ThinuwanW, line 35: 🤖 **Claude** · 🔴 **Important**

There are three byte-identical copies of this `chipClassName` constant and only this one was updated: PolicyManagerLeaveRequests.tsx:40 and LeaveTypesTable.tsx:21 still read "inline-flex w-fit items-center gap-2 rounded-full bg-tertiary-background px-4 py-2". PolicyManagerLeaveRequests is the same kind of request table with the same type/status chips, so after this PR the two tables render the same chips at different font sizes (inherited body size vs 14px). Extract the constant once (e.g. `LEAVE_REQUEST_CHIP_CLASS` in ~community/leave/constants or a common styles module) and import it in all three call sites, so this class can never drift again. At minimum, add `body2` to PolicyManagerLeaveRequests.tsx:40 in this PR.

frontend/src/community/attendance/components/molecules/ManagerTimesheetRequestTable/ManagerTimesheetRequestTable.tsx
```diff
@@ -176,7 +176,9 @@ const ManagerTimesheetRequestTable: FC<Props> = ({
           avatarUrl={timesheetRequest?.employee?.authPic as string}
           isResponsiveLayout={true}
           chipStyles={{
-            maxWidth: "15.625rem"
+            width: "fit-content",
+            maxWidth: "15.625rem",
+            backgroundColor: "var(--color-tertiary-background)"
```
ThinuwanW, line 181: 🤖 **Claude** · 🟡 **Suggestion**

This `chipStyles` object is now the third near-duplicate of the same override (see ManagerLeaveRequest.tsx:142-147 and PolicyManagerLeaveRequests.tsx:132-137), each repeating `width: fit-content`, `maxWidth: 15.625rem` and the tertiary background. AvatarChip's own base style hardcodes `backgroundColor: "common.white"` (AvatarChip/styles.ts:13), which is what every caller now has to override. Consider either changing the AvatarChip default / adding a `variant` prop, or exporting a shared `TABLE_AVATAR_CHIP_STYLES` constant, and pass it from all three tables. Also note this copy omits `justifyContent: "flex-start"` that the two leave tables include — if that is intentional, fine; if not, the chips will align differently across tables. Additionally, this file already has a colocated `styles.ts` (`classes.*`); per the styling guidelines this inline sx object belongs there rather than being rebuilt inside `transformToTableRows` on every render.

frontend/src/community/attendance/components/molecules/ManagerTimesheetRequestTable/ManagerTimesheetRequestTable.tsx
```diff
@@ -176,7 +176,9 @@ const ManagerTimesheetRequestTable: FC<Props> = ({
           avatarUrl={timesheetRequest?.employee?.authPic as string}
```
ThinuwanW, line 176: 🤖 **Claude** · 🟣 **Nit**

While touching this chip: `avatarUrl={timesheetRequest?.employee?.authPic as string}` uses an unnecessary `as` assertion to silence a possible `undefined`. AvatarChip's prop type is already `string | undefined`, and the leave tables handle it properly with `?? ""` (ManagerLeaveRequest.tsx:140). Replace the assertion with `timesheetRequest?.employee?.authPic ?? ""` to stay consistent and keep strict-mode type safety meaningful.

frontend/src/community/leave/components/molecules/LeaveRequestRow/RequestDates.tsx
```diff
@@ -40,8 +39,9 @@ const RequestDates: FC<Props> = ({ dates, days }) => {
         {dates}
       </Typography>
       <div
+        className="body2"
```
ThinuwanW, line 42: 🤖 **Claude** · 🟣 **Nit**

Mixed styling approaches in one element: the global `.body2` utility class plus an inline `style` object, sitting right next to a MUI `Typography variant="body2"` that resolves to the same 14px/400 from the theme. For consistency with the sibling element (and to keep the pill's visuals in one place), consider `<Typography variant="body2" component="div" sx={{ backgroundColor: "var(--color-tertiary-background)", borderRadius: "9.375rem", padding: "0.5rem 1rem" }} aria-hidden="true">`. Behaviorally equivalent today, so this is purely a consistency point. Note the near-duplicate LeaveRequestDates/LeaveRequestDates.tsx:37-42 renders the same pill and was not given `body2`, so "my requests" and "all requests" now show different font sizes for the same chip — worth aligning.

frontend/src/community/attendance/components/molecules/ManagerTimesheetRequestTable/ManagerTimesheetRequestTable.tsx
```diff
@@ -176,7 +176,9 @@ const ManagerTimesheetRequestTable: FC<Props> = ({
...
           isResponsiveLayout={true}
           chipStyles={{
-            maxWidth: "15.625rem"
+            width: "fit-content",
```
ThinuwanW, line 179: 🤖 **Claude** · 🔴 **Important**

📁 **File-level observation**

Unreachable branch / dead code in the file this PR modifies (pre-existing, but it silently disables a user-facing action). The `status` cell is a ternary on `timesheetRequest?.status === TimeSheetRequestStates.PENDING` (line 265). Inside the FALSE branch (lines 315-343) there is a nested guard `{timesheetRequest?.status === TimeSheetRequestStates.PENDING && (...)}` at line 326 wrapping the `KebabMenu`. That condition can never be true — it is the negation of the branch it lives in — so the cancel kebab menu never renders.

Everything reachable only through it is therefore dead: `getKebabMenuOptions` (line 119), `cancelMutate` / `useCancelTimeRequest` (line 134), the `onSuccess` (line 87) and `onError` (line 102) toast handlers, and the `KebabMenu` import (line 25).

Fix: decide the intent. If a manager should be able to cancel a pending request, move the `KebabMenu` into the TRUE branch alongside the approve/decline IconButtons (lines 266-314) and delete the redundant inner condition. If cancel is intentionally manager-unavailable, delete lines 326-341 together with `getKebabMenuOptions`, the `useCancelTimeRequest` call, both toast handlers, and the now-unused `KebabMenu` import.

frontend/src/community/leave/components/molecules/LeaveRequestRow/RequestDates.tsx
```diff
@@ -1,4 +1,4 @@
-import { Box, Theme, Typography, useTheme } from "@mui/material";
+import { Box, Typography } from "@mui/material";
```
ThinuwanW, line 1: 🤖 **Claude** · 🔴 **Important**

📁 **File-level observation**

Missing paired change — the employee-side twins of the very tables being fixed were not updated, so this PR ships the inconsistency it is trying to remove. These are different files from the ones the first pass listed (PolicyManagerLeaveRequests.tsx:40 / LeaveTypesTable.tsx:21):

1. `LeaveRequestDates/LeaveRequestDates.tsx:37-42` is a near-byte-identical twin of this component's days pill (same `var(--color-tertiary-background)`, same `borderRadius: "9.375rem"`, same `padding: "0.5rem 1rem"`) and did NOT get `className="body2"`. It is rendered by `LeaveRequests.tsx` (my leave requests) and `PolicyLeaveRequests.tsx`, both of which are TableView-based request tables — so the days pill will be 14px in the manager tables and inherited-16px in the employee tables.

2. `PolicyLeaveRequests.tsx:24` holds a FOURTH copy of the chip class string under a different name (`CHIP_CLASSES` vs `chipClassName`) and a different radius token (`rounded-[9.375rem]` instead of `rounded-full`). No `body2` added.

3. `LeaveRequests/styles.ts:2-10` (`iconStyles`) is a FIFTH copy of the same pill, expressed as a style object.

Fix: extract one shared pill primitive (the codebase already has the right home for it — `LeaveRequests/styles.ts` proves the pattern) and have all five call sites use it, so the font-size/radius/background live in exactly one place. At minimum, add the same typography normalization to `LeaveRequestDates.tsx`, `PolicyLeaveRequests.tsx:24`, and `LeaveRequests/styles.ts` in this PR, and align `rounded-[9.375rem]` to `rounded-full` and `CHIP_CLASSES` to `chipClassName`.

frontend/src/community/leave/components/molecules/LeaveRequestRow/RequestDates.tsx
ThinuwanW, line 42: 🤖 **Claude** · 🟡 **Suggestion**

The global `.body2` class is not equivalent to `Typography variant="body2"`, so this will not fully align the pill with the sibling date text at line 31-40.

- `styles/global.css:131` defines `.body2` as `font-size: 14px; font-weight: 400` with no `line-height`. The MUI theme (`src/community/common/theme/theme.ts:494`) defines `body2` as `fontSize: "0.875rem"`, and MUI's default `body2` also carries `lineHeight: 1.43`, which the theme override does not remove. The plain div therefore inherits `line-height: 1.5` from the `body` rule at `global.css:146` and renders slightly taller than the adjacent Typography — a visible pill-height difference.
- The hardcoded `14px` also ignores the user's browser base font size, while the rem-based theme value scales with it. So the two elements diverge for anyone who bumps their default font size — an accessibility regression relative to using the theme.

Fix: render the pill as `<Typography variant="body2" component="div" sx={{ backgroundColor: "var(--color-tertiary-background)", borderRadius: "9.375rem", padding: "0.5rem 1rem" }}>` and drop the `className`/`style` pair. That gets the exact same typography as the sibling from a single source.

Worth knowing for the `chipClassName` change too: `.body2` is declared unlayered in global.css while Tailwind utilities land in `@layer utilities`, and unlayered CSS beats every layer regardless of specificity or order. Nothing conflicts today, but anyone who later adds `text-xs` or `font-medium` to `chipClassName` will find it silently overridden by `.body2`.

frontend/src/community/attendance/components/molecules/ManagerTimesheetRequestTable/ManagerTimesheetRequestTable.tsx
ThinuwanW, line 179: 🤖 **Claude** · 🟣 **Nit**

This `chipStyles` object copies three of the four properties the two leave tables use but omits `justifyContent: "flex-start"` (see `ManagerLeaveRequest.tsx:142-147` and `PolicyManagerLeaveRequests.tsx:132-137`, which set `width` / `maxWidth` / `justifyContent` / `backgroundColor`). That omission interacts with something specific to this table: every header here is `align: "center"` (line 166), whereas the leave tables use default alignment.

Before this change the chip had no `width`, and the base style sets `display: "flex"` (`AvatarChip/styles.ts:9`), so it stretched to fill the cell. Adding `width: "fit-content"` shrinks it, and where it now sits depends on how `TableV2` implements `align` — if the cell uses `text-align: center` a shrunk flex block will stay left-aligned and no longer line up under the centered "NAME" header; if the cell is itself a flex container with `justify-content: center` it will center. I could not verify this because `TableV2` lives in the external `@rootcodelabs/skapp-ui` package (node_modules is not installed here), hence flagging as a nit rather than a bug.

Please eyeball the name column at >1445px and in the 1150-1445px responsive band, and either add `justifyContent: "flex-start"` to match the leave tables or drop `align: "center"` from the name header so header and cell agree.

frontend/src/community/leave/components/molecules/LeaveRequestRow/RequestDates.tsx
ThinuwanW, line 1: 🤖 **Claude** · 🟣 **Nit**

Module structure / naming, worth cleaning up while this shared component is being touched. `RequestDates.tsx` is now the only file in `molecules/LeaveRequestRow/` — the `LeaveRequestRow` component the folder is named after no longer exists, so the directory name matches nothing. Per the `ComponentName/ComponentName.tsx` + barrel `index.tsx` convention, this should move to `molecules/RequestDates/RequestDates.tsx` with an `index.tsx` re-export.

Related: its three consumers disagree on import style. `ManagerLeaveRequest.tsx:23` and `PolicyManagerLeaveRequests.tsx:26` use the `~community/*` alias, but `TeamAnalyticsLeaveHistoryTable.tsx:34` uses a deep relative path (`../LeaveRequestRow/RequestDates`), which the architecture guidelines call out against. Same pattern in the other file this PR touches: `ManagerTimesheetRequestTable.tsx:44` imports `TimesheetRequestFilterBody` via `../TimesheetRequestFilterBody/TimesheetRequestFilterBody`. If you do move the folder you have to update all three imports anyway, so converting the relative ones to the alias is nearly free. (Note this is not the same as the ESLint-autofixable import-order rule — relative-vs-alias is not auto-fixed.)

frontend/src/community/leave/components/molecules/LeaveRequestRow/RequestDates.tsx
Deshan1100: need this background color as per the design

frontend/src/community/leave/components/molecules/ManagerLeaveRequests/ManagerLeaveRequest.tsx
Deshan1100: need to change the font size

frontend/src/community/attendance/components/molecules/ManagerTimesheetRequestTable/ManagerTimesheetRequestTable.tsx
Deshan1100: chip bg color updated

frontend/src/community/attendance/components/molecules/ManagerTimesheetRequestTable/ManagerTimesheetRequestTable.tsx
Deshan1100: kept as previous

frontend/src/community/leave/components/molecules/LeaveRequestRow/RequestDates.tsx
Deshan1100: need this change

frontend/src/community/attendance/components/molecules/ManagerTimesheetRequestTable/ManagerTimesheetRequestTable.tsx
Deshan1100: need this change to fix the chip size bg color

frontend/src/community/leave/components/molecules/LeaveRequestRow/RequestDates.tsx
Deshan1100: No need

frontend/src/community/leave/components/molecules/LeaveRequestRow/RequestDates.tsx
Deshan1100: need to update this font size

frontend/src/community/attendance/components/molecules/ManagerTimesheetRequestTable/ManagerTimesheetRequestTable.tsx
Deshan1100: no need

frontend/src/community/leave/components/molecules/LeaveRequestRow/RequestDates.tsx
Deshan1100: No need

shakila-kamalasena (approved):

hasalarootcode (approved):

akilarootcode (approved):

akilarootcode (approved):

PR #1806, opened 2026-08-19
[Fix] Restrict employee leave policies section to super admin or people+leave admin

thusala:
## Summary

The **Leave policies** section on an employee's Leave tab (People → employee → Leave) was rendered for any leave admin, people admin or super admin. That section is a people-module surface and its data comes from the employee leave-policy assignment endpoint, so a user who only holds one of the two admin roles should not see it at all. Visibility is now restricted to a super admin, or a user holding **both** People Admin and Leave Admin.

### DIFF
--- a/frontend/src/community/leave/components/molecules/IndividualEmployeeLeaveReportSection/IndividualEmployeeLeaveReportSection.tsx
+++ b/frontend/src/community/leave/components/molecules/IndividualEmployeeLeaveReportSection/IndividualEmployeeLeaveReportSection.tsx
@@ -2,6 +2,7 @@ import { Stack } from "@mui/material";
 import { FC, useEffect, useMemo, useState } from "react";
 
 import PeopleLayout from "~community/common/components/templates/PeopleLayout/PeopleLayout";
+import useSessionData from "~community/common/hooks/useSessionData";
 import { useTranslator } from "~community/common/hooks/useTranslator";
 import { useGetLeaveTypes } from "~community/leave/api/LeaveApi";
 import UserAssignedLeaveTypes from "~community/leave/components/molecules/UserAssignedLeaveTypes/UserAssignedLeaveTypes";
@@ -10,7 +11,6 @@ import UserLeavePolicies from "~community/leave/components/molecules/UserLeavePo
 import UserLeavePoliciesSkeleton from "~community/leave/components/molecules/UserLeavePolicies/UserLeavePoliciesSkeleton";
 import UserLeaveUtilization from "~community/leave/components/molecules/UserLeaveUtilization/UserLeaveUtilization";
 import { USER_ASSIGNED_LEAVE_TYPES_PAGE_SIZE } from "~community/leave/constants/leavePolicyConstants";
-import useCanViewLeavePolicies from "~community/leave/hooks/useCanViewLeavePolicies";
 import useLeavePoliciesEnabled from "~community/leave/hooks/useLeavePoliciesEnabled";
 import { useLeaveStore } from "~community/leave/store/store";
 import { LeaveType } from "~community/leave/types/CustomLeaveAllocationTypes";
@@ -40,7 +40,11 @@ const IndividualEmployeeLeaveReportSection: FC<Props> = ({
 
   const { isAtLeastCoreTier } = useTier();
 
-  const canViewLeavePolicies = useCanViewLeavePolicies();
+  const { isSuperAdmin, isPeopleAdmin, isLeaveAdmin } = useSessionData();
+
+  const canViewLeavePolicies = Boolean(
+    isSuperAdmin || (isPeopleAdmin && isLeaveAdmin)
+  );
 
   const { isLeavePoliciesEnabled, isLoading: isLeavePolicyConfigLoading } =
     useLeavePoliciesEnabled();


akilarootcode (approved):

ThinuwanW (commented):
## 🤖 Claude Code Review

This PR replaces the shared `useCanViewLeavePolicies` hook (OR: super admin, leave admin, or people admin) with an inline, stricter AND condition (super admin, or people admin AND leave admin) for showing the employee leave policies section. The new rule contradicts the backend `@PreAuthorize` on the endpoint the section calls, the drawer route's `requiredAuthLevel`, and the sibling `useCanManageLeavePolicies` hook — a LEAVE_ADMIN-only user is the only role the backend allows to assign/unassign policies, yet can no longer see the section that exposes that action. The change also introduces a second, divergent definition of `canViewLeavePolicies` while the original hook remains in use elsewhere.

Found **10** new issue(s): 🔴 2 important, 🟡 5 suggestion(s), 🟣 3 nit(s)

frontend/src/community/leave/components/molecules/IndividualEmployeeLeaveReportSection/IndividualEmployeeLeaveReportSection.tsx
```diff
@@ -40,7 +40,11 @@ const IndividualEmployeeLeaveReportSection: FC<Props> = ({
 
   const { isAtLeastCoreTier } = useTier();
 
-  const canViewLeavePolicies = useCanViewLeavePolicies();
+  const { isSuperAdmin, isPeopleAdmin, isLeaveAdmin } = useSessionData();
+
+  const canViewLeavePolicies = Boolean(
+    isSuperAdmin || (isPeopleAdmin && isLeaveAdmin)
```
ThinuwanW, line 46: 🤖 **Claude** · 🔴 **Important**

The new gate requires people admin AND leave admin, which is stricter than — and inconsistent with — every other authority definition for this feature:

- Backend `EmployeeLeavePolicyController.getEmployeeLeavePolicies` (GET /v1/leave/policy-assignments/employee/{employeeId}) is annotated `@PreAuthorize("hasAnyRole('ROLE_LEAVE_ADMIN', 'ROLE_PEOPLE_ADMIN')")` — an OR, not an AND.
- Assign/bulk-assign/unassign on that same controller are `ROLE_LEAVE_ADMIN` only.
- The `Leave Policies` drawer route (`src/community/common/utils/data/routes.ts`) uses `requiredAuthLevel: [SUPER_ADMIN, LEAVE_ADMIN, PEOPLE_ADMIN]` — also an OR.
- `UserLeavePolicies` gates its own "Assign policy" button with `useCanManageLeavePolicies()` = `isSuperAdmin || isLeaveAdmin`.

Net effect: a user holding LEAVE_ADMIN but not PEOPLE_ADMIN — precisely the role the backend authorizes to assign and unassign policies — no longer renders `UserLeavePolicies` at all, so the Assign/Unassign actions inside it become unreachable from the employee profile. A PEOPLE_ADMIN-only user likewise loses read access the backend explicitly grants them.

Either the UI condition should stay an OR (`isSuperAdmin || isPeopleAdmin || isLeaveAdmin`, i.e. keep using the existing hook), or — if the tightening is genuinely intended — the backend `@PreAuthorize` on all four `EmployeeLeavePolicyController` methods, the `2E` drawer route's `requiredAuthLevel`, and `useCanManageLeavePolicies` must be updated in the same change so the front end and the API agree on who may view and manage employee policy assignments.

frontend/src/community/leave/components/molecules/IndividualEmployeeLeaveReportSection/IndividualEmployeeLeaveReportSection.tsx
```diff
@@ -40,7 +40,11 @@ const IndividualEmployeeLeaveReportSection: FC<Props> = ({
 
   const { isAtLeastCoreTier } = useTier();
 
-  const canViewLeavePolicies = useCanViewLeavePolicies();
+  const { isSuperAdmin, isPeopleAdmin, isLeaveAdmin } = useSessionData();
+
+  const canViewLeavePolicies = Boolean(
```
ThinuwanW, line 45: 🤖 **Claude** · 🟡 **Suggestion**

`useCanViewLeavePolicies` was not removed — it is still imported and used by `src/community/common/components/organisms/Drawer/Drawer.tsx:131` (to decide whether to run the leave policy config query). After this change there are two things named `canViewLeavePolicies` with different meanings: the hook's OR logic in the drawer, and this component's AND logic. That is exactly the kind of drift that causes the sidebar and the page body to disagree about permissions.

If the stricter rule is intended, express it as a named hook alongside the existing pair (e.g. `useCanViewEmployeeLeavePolicies` in `~community/leave/hooks/`, matching `useCanManageLeavePolicies`) rather than inlining the role expression in a component, and give the local variable a name that reflects its narrower scope. That keeps the module's permission rules in one place and makes the difference from the drawer's rule explicit rather than accidental.

frontend/src/community/leave/components/molecules/IndividualEmployeeLeaveReportSection/IndividualEmployeeLeaveReportSection.tsx
```diff
@@ -40,7 +40,11 @@ const IndividualEmployeeLeaveReportSection: FC<Props> = ({
 
   const { isAtLeastCoreTier } = useTier();
 
-  const canViewLeavePolicies = useCanViewLeavePolicies();
+  const { isSuperAdmin, isPeopleAdmin, isLeaveAdmin } = useSessionData();
```
ThinuwanW, line 43: 🤖 **Claude** · 🟣 **Nit**

For the newly excluded roles the section degrades to nothing rather than to a fallback. `showAssignedLeaveTypes` is `!isLeavePolicyConfigLoading && !isLeavePoliciesEnabled`, so when the leave-policies feature is enabled a LEAVE_ADMIN-only (or PEOPLE_ADMIN-only) user sees neither `UserLeavePolicies` nor `UserAssignedLeaveTypes` — just utilization and history. `DirectorySteppers.tsx:63` still makes the Leave tab visible specifically for `isLeaveAdmin`, so that user is routed to a tab whose policy content silently disappears with no empty state or explanation. Worth confirming the intended UX: either keep the tab content meaningful for these roles, or render an explicit restricted/empty state instead of omitting the block.

frontend/src/community/leave/components/molecules/IndividualEmployeeLeaveReportSection/IndividualEmployeeLeaveReportSection.tsx
ThinuwanW, line 46: 🤖 **Claude** · 🔴 **Important**

Missing paired change / dead capability path: the assign-and-unassign feature is now unreachable for the exact role the backend designates as its owner. Backend `EmployeeLeavePolicyController` authorizes POST (assign), POST /bulk and DELETE (unassign) with `hasAnyRole('ROLE_LEAVE_ADMIN')` only, and the GET list endpoint with `hasAnyRole('ROLE_LEAVE_ADMIN', 'ROLE_PEOPLE_ADMIN')` — i.e. an OR. On the frontend, `useCanManageLeavePolicies` (super admin OR leave admin) still gates the "Assign policy" button, but that button lives inside `UserLeavePolicies`, which this component now renders only for super admin or people+leave admin. Net effect: a LEAVE_ADMIN-only user keeps full backend write permission and passes `useCanManageLeavePolicies`, but has no UI entry point to reach it — the leave-admin branch of `useCanManageLeavePolicies` becomes dead code, and `AssignLeavePolicyModal` / `UnassignLeavePolicyModal` are unreachable for that role. Fix: make the view gate match the backend contract (`isSuperAdmin || isLeaveAdmin || isPeopleAdmin`, i.e. keep the existing hook). If the intent really is to narrow access, the change must be paired: tighten the backend `@PreAuthorize` on the assignment endpoints and `useCanManageLeavePolicies` in the same PR, otherwise view is narrower than manage and the two permission models contradict each other.

frontend/src/community/leave/components/molecules/IndividualEmployeeLeaveReportSection/IndividualEmployeeLeaveReportSection.tsx
```diff
@@ -40,7 +40,11 @@ const IndividualEmployeeLeaveReportSection: FC<Props> = ({
...
+
+  const canViewLeavePolicies = Boolean(
+    isSuperAdmin || (isPeopleAdmin && isLeaveAdmin)
+  );
 
   const { isLeavePoliciesEnabled, isLoading: isLeavePolicyConfigLoading } =
```
ThinuwanW, line 49: 🤖 **Claude** · 🟡 **Suggestion**

Error state is silently swallowed on the query this permission flag now controls. `useLeavePoliciesEnabled` returns `isError`, and `Drawer.tsx` deliberately consumes it (`isLeavePoliciesConfigError`) and passes it into `getDrawerRoutes`, which keeps the "Leave Policies" nav item visible on failure (fail-open). Here `isError` is discarded, so if GET /leave/policies/config fails, `isLeavePoliciesEnabled` is false and `isLoading` is false, which makes `showAssignedLeaveTypes` true: an eligible admin on a policies-enabled tenant silently sees the legacy entitlement-card UI with no error feedback, while the drawer simultaneously still offers the Leave Policies nav item. Fix: destructure `isError` here too and either render an error state (the codebase has `EmptyDataView`, used for exactly this in `UserLeavePolicies.tsx:68`) or fail open to the policies branch, so this component and `getDrawerRoutes` agree on the failure behaviour.

frontend/src/community/leave/components/molecules/IndividualEmployeeLeaveReportSection/IndividualEmployeeLeaveReportSection.tsx
```diff
@@ -40,7 +40,11 @@ const IndividualEmployeeLeaveReportSection: FC<Props> = ({
...
     useLeavePoliciesEnabled();
```
ThinuwanW, line 50: 🤖 **Claude** · 🟡 **Suggestion**

Missing loading state for the roles the narrower gate now excludes. `isPolicySectionLoading` is `canViewLeavePolicies && isLeavePolicyConfigLoading`, so the skeleton only renders for privileged users, while the fallback branch (`showAssignedLeaveTypes`) also waits on `!isLeavePolicyConfigLoading` and has no skeleton of its own. Narrowing the gate moves people-admin-only and leave-admin-only users into that window, so they now get a blank section for the duration of the config request and then a sudden content pop-in. Fix: drive the skeleton off `isLeavePolicyConfigLoading` alone (independent of the permission flag), or add a loading placeholder to the assigned-leave-types branch.

frontend/src/community/leave/components/molecules/IndividualEmployeeLeaveReportSection/IndividualEmployeeLeaveReportSection.tsx
ThinuwanW, line 45: 🤖 **Claude** · 🟡 **Suggestion**

Authority rule inlined in a molecule instead of a hook, creating a third divergent definition and a name collision. `src/community/leave/hooks/` already holds the established per-capability hooks (`useCanViewLeavePolicies` = super admin OR leave admin OR people admin; `useCanManageLeavePolicies` = super admin OR leave admin). Adding a third, different rule inline means no other consumer can reuse it, and the local variable is named `canViewLeavePolicies` — identical to the still-live hook that `Drawer.tsx:131` uses with different truth conditions, so two same-named booleans in the codebase now disagree about who can view leave policies. Fix: express the rule once as a hook in `src/community/leave/hooks/` with a name that states what it gates (e.g. `useCanViewEmployeeLeavePolicies`) and consume it here, so the rule has a single source of truth and cannot drift further.

frontend/src/community/leave/components/molecules/IndividualEmployeeLeaveReportSection/IndividualEmployeeLeaveReportSection.tsx
```diff
@@ -2,6 +2,7 @@ import { Stack } from "@mui/material";
 import { FC, useEffect, useMemo, useState } from "react";
 
 import PeopleLayout from "~community/common/components/templates/PeopleLayout/PeopleLayout";
+import useSessionData from "~community/common/hooks/useSessionData";
```
ThinuwanW, line 5: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

No test coverage for a permission/visibility change. `IndividualEmployeeLeaveReportSection` has no `.test.tsx` (the directory contains only the component and `styles.ts`), and neither `useCanViewLeavePolicies` nor `useCanManageLeavePolicies` has a test either — so a rule that controls who can see employee leave policy assignments ships entirely unverified, and a future refactor of the inline boolean would be caught by nothing. Fix: add `IndividualEmployeeLeaveReportSection.test.tsx` (wrapped in `MockTheme`, mocking `useSessionData` and `useLeavePoliciesEnabled`) asserting the rendered branch for each relevant role combination — super admin, people+leave admin, leave-admin-only, people-admin-only, plain employee — plus the policies-disabled and config-error cases, using `screen.getByRole` for the section heading.

frontend/src/community/leave/components/molecules/IndividualEmployeeLeaveReportSection/IndividualEmployeeLeaveReportSection.tsx
ThinuwanW, line 45: 🤖 **Claude** · 🟣 **Nit**

Missing explicit type annotation on the derived flag. The guidelines require an explicit type for every variable, and the hooks this replaces annotate their contract (`useCanViewLeavePolicies = (): boolean`). Write `const canViewLeavePolicies: boolean = Boolean(...)` so the intent is stated rather than inferred (`useSessionData` returns `boolean | undefined` for each role flag, which is why the `Boolean()` wrapper is needed in the first place).

frontend/src/community/leave/components/molecules/IndividualEmployeeLeaveReportSection/IndividualEmployeeLeaveReportSection.tsx
ThinuwanW, line 46: 🤖 **Claude** · 🟣 **Nit**

Worth knowing when settling on the final rule: the `isSuperAdmin` disjunct is very nearly redundant. `RolesServiceImpl.saveSuperAdminRoles` and the super-admin branch of the role-update path force `PEOPLE_ADMIN` and `LEAVE_ADMIN` onto every super admin, so `isPeopleAdmin && isLeaveAdmin` is already true for them. The only case where the disjunct changes the outcome is an enterprise tenant with the leave module disabled, where `EpAuthorityServiceImpl` strips the `ROLE_LEAVE_*` authorities — and in that case this leave section arguably should not render at all. So in practice the new expression's only real effect is to exclude single-role admins, which is the strictness concern already raised; the super-admin term adds no coverage.

PR #1808, opened 2026-08-20
[Refactor] Use skapp-ui avatar components in work location employee selector

thusala:


hasalarootcode (dismissed):

ThinuwanW (commented):
## 🤖 Claude Code Review

This PR migrates the WorkLocationEmployeeSelector from the legacy MUI-based community AvatarChip/AvatarGroup to the @rootcodelabs/skapp-ui design-system components, and removes the now-unused `title`/`renderSurplus` escape hatch from the community AvatarGroup. The AvatarGroup.tsx cleanup is safe — I grepped every `<AvatarGroup` call site in the repo and none passed `title`, and the `Avatar` import was correctly dropped. The main concerns in the selector are that the overlapping-avatar group is hand-rolled with Tailwind instead of using the design system's own `AvatarGroup` (already used in ProjectsTable), duplicate DOM ids for chips that render in both the trigger and the popper list, and an `EmployeeAvatarChip` that is a byte-for-byte copy of one already living in SupervisorReassignmentModalSection.

Found **16** new issue(s): 🔴 3 important, 🟡 6 suggestion(s), 🟣 7 nit(s)

frontend/src/community/configurations/components/molecules/WorkLocationEmployeeSelector/WorkLocationEmployeeSelector.tsx
```diff
@@ -267,35 +321,35 @@ const WorkLocationEmployeeSelector = ({
...
-      <AvatarGroup
-        avatars={selectedEmployees.map((emp) => ({
-          firstName: emp.firstName ?? "",
-          lastName: (emp.lastName ?? "").trim(),
-          image: emp.authPic || null
-        }))}
-        max={4}
-        title={remainingTitle || undefined}
-      />
+      <div
```
ThinuwanW, line 332: 🤖 **Claude** · 🔴 **Important**

The overlapping avatar stack is hand-rolled (`flex -space-x-3` + a manually rendered surplus `Avatar`), but `@rootcodelabs/skapp-ui` already exports an `AvatarGroup` for exactly this. See `src/enterprise/projects/components/organisms/ProjectsTable/ProjectsTable.tsx:192`, which uses `<AvatarGroup avatars={[{id, firstName, lastName, src, size}]} maxVisible={5} ariaLabel={...} />`. Reimplementing it here means: (a) the overlap spacing, ring/border treatment and surplus-chip styling will drift from every other avatar group in the app; (b) the stacking order is inverted relative to the MUI group this replaces — with `-space-x-3` each later avatar paints on top of the previous one and the `+N` surplus lands on top of the last avatar, whereas the old MUI `AvatarGroup` painted the first avatar on top and the surplus behind; (c) the `ariaLabel` here is a raw joined name list, while the DS `AvatarGroup` convention in ProjectsTable is a translated aria label (`translateAria(["...membersGroupAriaLabel"], { count })`), so this also skips i18n for the group's accessible name. Suggest replacing the whole block with the DS `AvatarGroup`, passing `maxVisible={MAX_VISIBLE_AVATARS}` and a translated `ariaLabel`, and dropping the local `EmployeeGroupAvatar` wrapper (resolve the image URLs in a small per-item component or memoized map before building the `avatars` array).

frontend/src/community/configurations/components/molecules/WorkLocationEmployeeSelector/WorkLocationEmployeeSelector.tsx
```diff
@@ -243,14 +289,22 @@ const WorkLocationEmployeeSelector = ({
...
-    />
-  );
+  const renderAllEmployeesChip = () => {
+    const allEmployeesLabel = translateText(["form.allEmployees"]);
+
+    return (
+      <div className="w-fit max-w-full">
+        <AvatarChip
+          avatarProps={{
+            id: "all-employees",
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Duplicate DOM ids. `renderAllEmployeesChip()` hardcodes `id: "all-employees"` and is rendered in two places at once: as the trigger content (line 319, when `isAllSelected`) and as the popper list option (line 464, when the search box is empty). With the popper open and All Employees selected, two elements carry `id="all-employees"` simultaneously. The same problem applies to `EmployeeAvatarChip` (`id: String(employee.employeeId)`, line 57) and `EmployeeGroupAvatar` (`id: String(employee.employeeId)`, line 77): selected employees are rendered both in the trigger (lines 326 / 342) and in the "already selected" section of the list (line 441), so each selected employee's id appears twice — and for 3+ selections the group avatar and the list chip collide on the same id. Duplicate ids break `getElementById`, `label[for]`, and any `aria-labelledby`/`aria-describedby` the DS component derives from the id. Fix by namespacing per render context, matching the existing convention in ProjectsTable.tsx:207 (`id: \`avatar-${member.id}\``): e.g. accept an `idPrefix` prop on `EmployeeAvatarChip`/`EmployeeGroupAvatar` and pass `trigger-` vs `option-`, and use `trigger-all-employees` vs `option-all-employees` for the All Employees chip.

frontend/src/community/configurations/components/molecules/WorkLocationEmployeeSelector/WorkLocationEmployeeSelector.tsx
```diff
@@ -34,11 +35,56 @@ import {
...
+
 interface Props {
   formik: FormikProps<WorkLocationFormValues>;
   preloadedEmployees?: WorkLocationEmployee[];
 }
 
+const getEmployeeName = (employee: AllEmployeeDataType): string =>
+  concatStrings([employee.firstName, employee.lastName]).trim();
+
+const EmployeeAvatarChip: FC<{ employee: AllEmployeeDataType }> = ({
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`EmployeeAvatarChip` is a near-exact duplicate of the component of the same name already defined at `src/community/people/components/molecules/SupervisorReassignmentModalSection/SupervisorReassignmentModalSection.tsx:16-33` — same `AllEmployeeDataType` input, same `useGetImageUrl(employee.authPic ?? "")`, same `avatarProps` shape, same `concatStrings([firstName, lastName]).trim()` label. The only difference is the `w-fit min-w-0 max-w-full` wrapper div added here. Per the duplication guideline, extract this into a shared component (e.g. `~community/common/components/atoms/EmployeeAvatarChip/EmployeeAvatarChip.tsx`) with an optional `className`/`idPrefix` prop, and have both call sites consume it. Otherwise the next design-system prop change has to be applied in two places.

frontend/src/community/configurations/components/molecules/WorkLocationEmployeeSelector/WorkLocationEmployeeSelector.tsx
```diff
@@ -34,11 +35,56 @@ import {
...
+  employee
+}) => {
+  const imageUrl = useGetImageUrl(employee.authPic ?? "");
+
+  return (
+    <div className="w-fit min-w-0 max-w-full">
+      <AvatarChip
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The old chips passed `isTooltipEnabled` to the community `AvatarChip`, which wrapped the chip in an MUI `Tooltip` showing the full `first last` name on hover — important because the trigger box is width-constrained (`w-full`, `h-12`) and the new wrapper adds `min-w-0 max-w-full`, so long names will be clipped/truncated. The migrated `AvatarChip` here is given no tooltip or `title` affordance at all, so a truncated name becomes unreadable with no way to reveal it. Please verify whether the DS `AvatarChip` renders its own tooltip/`title` on truncation; if it does not, pass a `title` (or wrap in the existing `Tooltip` atom) so the behaviour is preserved. Note the new `EmployeeGroupAvatar` and the surplus `Avatar` do set `title`, which makes the missing tooltip on the chips look like an oversight rather than a deliberate drop.

frontend/src/community/configurations/components/molecules/WorkLocationEmployeeSelector/WorkLocationEmployeeSelector.tsx
```diff
@@ -267,35 +321,35 @@ const WorkLocationEmployeeSelector = ({
...
+      <div
+        className="flex -space-x-3"
+        role="group"
+        aria-label={selectedEmployees.map(getEmployeeName).join(", ")}
+      >
+        {visibleEmployees.map((emp) => (
+          <EmployeeGroupAvatar key={emp.employeeId} employee={emp} />
+        ))}
+        {remainingEmployees.length > 0 && (
+          <Avatar
```
ThinuwanW, line 341: 🤖 **Claude** · 🟣 **Nit**

The surplus `Avatar` uses `count` and `title`, and `EmployeeGroupAvatar` (line 82) uses `title` — no other call site in this repo passes either prop to the skapp-ui `Avatar` (every existing usage sticks to `id`/`size`/`src`/`firstName`/`lastName`/`alt`, e.g. TaskRowMeta.tsx:26, BirthdayModalShell.tsx:74, ProjectsTable.tsx:205). I could not verify these props exist on `AvatarProps` in 1.1.1-beta.13 because node_modules isn't installed here. Please confirm `count` renders the `+N` surplus and `title` is forwarded to the DOM; if `title` is not a declared prop it will either fail the strict type check or be silently dropped, which would also silently drop the hover-name affordance you're relying on for the hidden employees. Switching to the DS `AvatarGroup` (see the other finding) would sidestep this entirely since it owns the surplus rendering.

frontend/src/community/configurations/components/molecules/WorkLocationEmployeeSelector/WorkLocationEmployeeSelector.tsx
```diff
@@ -243,14 +289,22 @@ const WorkLocationEmployeeSelector = ({
...
-  const renderAllEmployeesChip = () => (
-    <AvatarChip
-      firstName={translateText(["form.allEmployees"]).trim()}
-      lastName=""
-      avatarUrl={undefined}
-      isTooltipEnabled
-    />
-  );
+  const renderAllEmployeesChip = () => {
+    const allEmployeesLabel = translateText(["form.allEmployees"]);
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

The previous implementation called `translateText(["form.allEmployees"]).trim()`; the new `allEmployeesLabel` drops the `.trim()`. If a translator ever leaves trailing whitespace in the `allEmployees` value it now flows into both the `label` and the avatar's `firstName` (which the DS uses to derive the initial), so the initial could be computed from a space. Cheap to restore: `translateText(["form.allEmployees"]).trim()`.

frontend/src/community/configurations/components/molecules/WorkLocationEmployeeSelector/WorkLocationEmployeeSelector.tsx
```diff
@@ -1,7 +1,8 @@
 import { CircularProgress } from "@mui/material";
-import { Checkbox } from "@rootcodelabs/skapp-ui";
+import { Avatar, AvatarChip, Checkbox } from "@rootcodelabs/skapp-ui";
```
ThinuwanW, line 2: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

No `.test.tsx` exists for this component (the directory contains only `WorkLocationEmployeeSelector.tsx`), and this PR adds two new sub-components plus a changed avatar-overflow rule (`MAX_VISIBLE_AVATARS = 4` now shows 4 avatars + surplus, where the old `max={4}` MUI group showed 3 + surplus). A small test with the `MockTheme` wrapper covering the three trigger states — 0 selected (placeholder), <=2 selected (chips), >2 selected (avatar group with the correct surplus count) — would lock in the new overflow threshold and catch regressions from future design-system upgrades.

ThinuwanW, line 2: 🤖 **Claude** · 🔴 **Important**

📁 **File-level observation**

This PR migrates the selector to skapp-ui primitives (`Avatar`, `AvatarChip`) but keeps hand-rolling the whole multi-select. `@rootcodelabs/skapp-ui` already exports `AvatarChipsInput` (+ `AvatarChipsInputResult`) for exactly this pattern — search box, result list, selected avatar chips, chip removal, live-region announcements — and it is already used in four places (`AddCustomerModal`, `AddProjectsModal`, `SendInvoiceEmailModal`, `AddProjectToGuest`). More telling: `configurations.workLocation.form.employeeMultiSelect` in `src/community/common/assets/languages/english/configurations.json` (lines 164-181) is a dead translation block whose keys map 1:1 onto `AvatarChipsInput`'s props (`ariaLabelClearButton`, `ariaLabelSearch`, `ariaLabelSearchResults`, `chipRemovedText`, `chipSelectedText`, `instructionText`, `itemSelectedText`, `noResultsText`, `regionAriaLabel`, `resultCountText`, `searchClearedText`, `selectedSearchPlaceholder`). A grep for `employeeMultiSelect` returns zero code references — those strings were clearly authored for this migration and were never wired up, while this file instead uses a second, duplicate 'no results' string (`form.noSearchResults`) and derives aria labels ad hoc. Either finish the intended migration to `AvatarChipsInput` (dropping the bespoke scroll listener, search filter, and role="listbox"/role="option" markup, and consuming the existing keys), or delete the dead `employeeMultiSelect` block in this PR so it does not rot further.

frontend/src/community/configurations/components/molecules/WorkLocationEmployeeSelector/WorkLocationEmployeeSelector.tsx
```diff
@@ -243,14 +289,22 @@ const WorkLocationEmployeeSelector = ({
...
+            size: "sm"
+          }}
+          label={allEmployeesLabel}
+        />
+      </div>
+    );
+  };
 
   const renderTriggerContent = () => {
     if (selectedCount === 0) {
```
ThinuwanW, line 309: 🤖 **Claude** · 🟡 **Suggestion**

Branch ordering bug in the rewritten `renderTriggerContent`: the `selectedCount === 0` guard runs before the `isAllSelected` guard, and `selectedCount` is `isAllSelected ? allEmployees.length : selectedIds.length` where `allEmployees` only contains the infinite-query pages fetched so far. So when editing a work location that has `isAllEmployees: true`, the trigger shows the "Assigned employees" placeholder instead of the "All Employees" chip until the first employee page resolves — and shows it permanently if the tenant has no ACTIVE/PENDING employees, making the field look empty while the form will still submit `isAllEmployees: true`. Fix by moving `if (isAllSelected) return renderAllEmployeesChip();` above the `selectedCount === 0` check (the `isAllSelected` case does not depend on `allEmployees` at all).

frontend/src/community/configurations/components/molecules/WorkLocationEmployeeSelector/WorkLocationEmployeeSelector.tsx
```diff
@@ -433,22 +486,16 @@ const WorkLocationEmployeeSelector = ({
...
                 );
               })}
 
-          {employeeSearchText.length > 0 &&
-            displayEmployees.length === 0 && (
-              <p className="text-center text-secondary-text body2 py-4">
-                {translateText(["form.noSearchResults"])}
-              </p>
-            )}
+          {employeeSearchText.length > 0 && displayEmployees.length === 0 && (
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

No loading state for the search request, so this "no results" line flashes on every keystroke. `useGetSearchedEmployees` debounces 500 ms and is `enabled: debouncedSearchTerm.length > 0`, so for at least half a second after typing `searchResults` is `undefined`, `displayEmployees` collapses to `[]`, and `employeeSearchText.length > 0 && displayEmployees.length === 0` is true — the user sees "No employees found for your search" before the query has even fired. Worse, the selected-employees block above filters locally, so a matching selected employee's chip can render at the same time as "no results", which is contradictory. Destructure `isLoading`/`isFetching` from `useGetSearchedEmployees` and render the existing `CircularProgress` block while a search is pending, gating the empty-state on `!isFetching`.

frontend/src/community/configurations/components/molecules/WorkLocationEmployeeSelector/WorkLocationEmployeeSelector.tsx
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`AvatarChip` is rendered without `showActionButton` or `backgroundColor`, unlike every analogous call site in the repo. `OwnerOptionItem` and `OwnerTriggerContent` (the same shape as this component: a popper trigger plus option rows) both pass `showActionButton={false}`, and `SelectedOwnerField` passes it explicitly alongside `onActionClick`. If the skapp-ui default is `true`, each of the ~N option rows and the trigger will render a remove/X button with no `onActionClick` handler — a dead control that steals the click from the row's own `onClick`/`toggleEmployee`. Separately, `OwnerTriggerContent` passes `backgroundColor="bg-secondary-background"` precisely because its trigger sits on `bg-tertiary-background`, which is this component's trigger background too (line 368) — without it the chip will likely be invisible against the trigger. Please set `showActionButton={false}` on both usages and an explicit `backgroundColor` for the trigger chip.

frontend/src/community/configurations/components/molecules/WorkLocationEmployeeSelector/WorkLocationEmployeeSelector.tsx
```diff
@@ -34,11 +35,56 @@ import {
...
 } from "~community/people/types/PeopleTypes";
 
+const MAX_VISIBLE_AVATARS = 4;
+
 interface Props {
   formik: FormikProps<WorkLocationFormValues>;
   preloadedEmployees?: WorkLocationEmployee[];
 }
 
+const getEmployeeName = (employee: AllEmployeeDataType): string =>
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Prop typing on the two new sub-components: `FC<{ employee: AllEmployeeDataType }>` uses an inline object type rather than a named props interface, which diverges from the repo convention for exactly this kind of atom (`OwnerAvatarChipProps` in `src/community/crm/components/atoms/OwnerAvatarChip/OwnerAvatarChip.tsx`). Both components read only `employeeId`, `firstName`, `lastName`, and `authPic`, so binding them to the full people-module `AllEmployeeDataType` couples the configurations module to a DTO it does not need and is part of why this file needs the `as AllEmployeeDataType[]` casts at lines 148, 221 and 246. Declare a narrow local type (e.g. `interface SelectableEmployee { employeeId: number; firstName: string; lastName: string; authPic?: string }`) and a named `EmployeeAvatarChipProps` / `EmployeeGroupAvatarProps` interface; `WorkLocationEmployee` and `AllEmployeeDataType` both structurally satisfy it, so the casts can go away too.

frontend/src/community/configurations/components/molecules/WorkLocationEmployeeSelector/WorkLocationEmployeeSelector.tsx
```diff
@@ -267,35 +321,35 @@ const WorkLocationEmployeeSelector = ({
 
     if (selectedCount <= 2) {
       return (
-        <div className="flex gap-2">
+        <div className="flex min-w-0 gap-2">
```
ThinuwanW, line 319: 🤖 **Claude** · 🟣 **Nit**

Overflow containment is applied inconsistently across the three trigger branches. `min-w-0` was added here and `w-fit min-w-0 max-w-full` on the chip wrapper (line 54), but the trigger container itself (line 368, `bg-tertiary-background h-12 rounded-lg flex items-center w-full ... px-3`) has neither `min-w-0` nor `overflow-hidden`, and the avatar-stack branch (line 336) has neither `min-w-0` nor `max-w-full`. `max-w-full` only caps the box at the parent width — it does not clip the label — so two long names (or five stacked `size="sm"` avatars) will visibly spill outside the fixed-height rounded trigger. The previous community `AvatarChip` handled this via its own `text-overflow: ellipsis` + `maxWidth` chip styles, which are gone now. Add `min-w-0 overflow-hidden` to the trigger container and apply the same `min-w-0 max-w-full` treatment to the avatar-stack wrapper.

frontend/src/community/configurations/components/molecules/WorkLocationEmployeeSelector/WorkLocationEmployeeSelector.tsx
```diff
@@ -267,35 +321,35 @@ const WorkLocationEmployeeSelector = ({
...
           ))}
         </div>
       );
     }
 
-    const remainingEmployees = selectedEmployees.slice(3);
-    const remainingTitle = remainingEmployees
-      .map((emp) => `${emp.firstName ?? ""} ${(emp.lastName ?? "").trim()}`.trim())
-      .join(", ");
+    const visibleEmployees = selectedEmployees.slice(0, MAX_VISIBLE_AVATARS);
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Behavioural drift in how many avatars are visible, and a leftover magic number. MUI's `AvatarGroup` with `max={4}` renders `max - 1` avatars plus the surplus badge, so the old code showed 3 avatars + "+N" (and the old `slice(3)` matched that). `MAX_VISIBLE_AVATARS = 4` with `slice(0, 4)` now shows 4 avatars + "+N", and with exactly 4 selected the "+N" badge disappears entirely where it previously read "3 avatars +1". Please confirm 4 visible is the intended design (it widens the stack by one circle inside the `h-12` trigger); if the intent was parity, use `slice(0, MAX_VISIBLE_AVATARS - 1)`. Also, the `selectedCount <= 2` threshold at line 322 is still a bare literal sitting next to the newly-named constant — extract it (e.g. `MAX_INLINE_CHIPS = 2`) so the two thresholds are visibly related.

frontend/src/community/configurations/components/molecules/WorkLocationEmployeeSelector/WorkLocationEmployeeSelector.tsx
```diff
@@ -267,35 +321,35 @@ const WorkLocationEmployeeSelector = ({
...
-          lastName: (emp.lastName ?? "").trim(),
-          image: emp.authPic || null
-        }))}
-        max={4}
-        title={remainingTitle || undefined}
-      />
+      <div
+        className="flex -space-x-3"
+        role="group"
+        aria-label={selectedEmployees.map(getEmployeeName).join(", ")}
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`selectedEmployees.map(getEmployeeName).join(", ")` (aria-label) and `remainingEmployees.map(getEmployeeName).join(", ")` (surplus title) are rebuilt on every render, as are `visibleEmployees` and `remainingEmployees`, because they live inside the `renderTriggerContent` function body rather than a `useMemo`. This component re-renders on every keystroke in the search box (`employeeSearchText` state) and on every `preloadedEmployees` identity change (the `= []` default plus the effect at line 240 makes that every render), so for a large selection you are allocating two long comma-joined strings per keystroke. Hoist `visibleEmployees`, `remainingEmployees`, and both joined strings into `useMemo`s keyed on `selectedEmployees`.

frontend/src/community/configurations/components/molecules/WorkLocationEmployeeSelector/WorkLocationEmployeeSelector.tsx
```diff
@@ -34,11 +35,56 @@ import {
...
+  const imageUrl = useGetImageUrl(employee.authPic ?? "");
+
+  return (
+    <div className="w-fit min-w-0 max-w-full">
+      <AvatarChip
+        avatarProps={{
+          id: String(employee.employeeId),
+          firstName: employee.firstName,
+          lastName: employee.lastName,
+          src: imageUrl ?? "",
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`avatarProps` hygiene in the two new sub-components. (1) `src: imageUrl ?? ""` here and at line 80: `useGetImageUrl` returns `null` until it resolves, and an empty-string `src` makes a rendered `` resolve to the current document URL in several browsers (a wasted request plus an error event) instead of falling through to the initials avatar. The closest reference implementation, `OwnerAvatarChip`, deliberately uses `src: imageUrl ?? undefined` / `lastName: owner.lastName ?? undefined` — match that. (2) `EmployeeGroupAvatar` passes `alt={employeeName}` but `EmployeeAvatarChip` passes no `alt` at all, so the chip's `` has no alternative text; `ProjectsTable.tsx:206` passes `alt` inside `avatarProps` for exactly this case. Add `alt: getEmployeeName(employee)` to the chip's `avatarProps`.

frontend/src/community/configurations/components/molecules/WorkLocationEmployeeSelector/WorkLocationEmployeeSelector.tsx
thusala: Done in 4692d00c1 — `.trim()` restored.

frontend/src/community/configurations/components/molecules/WorkLocationEmployeeSelector/WorkLocationEmployeeSelector.tsx
thusala: Done in 4692d00c1 — `isAllSelected` is now the first branch, so the All Employees chip no longer waits on the first employee page.

frontend/src/community/configurations/components/molecules/WorkLocationEmployeeSelector/WorkLocationEmployeeSelector.tsx
thusala: Done in 4692d00c1 — trigger container gets `min-w-0 overflow-hidden` and the avatar-stack wrapper `min-w-0 max-w-full`.

frontend/src/community/configurations/components/molecules/WorkLocationEmployeeSelector/WorkLocationEmployeeSelector.tsx
thusala: Done in 4692d00c1 — the visible/remaining slices and both joined labels are `useMemo`d on `selectedEmployees`.

frontend/src/community/configurations/components/molecules/WorkLocationEmployeeSelector/WorkLocationEmployeeSelector.tsx
thusala: Partly done in 4692d00c1 — extracted to `~community/common/components/atoms/EmployeeAvatarChip` with `idPrefix`/`className` props, and this file now consumes it. Left `SupervisorReassignmentModalSection` untouched since that file is outside this PR; it can be swapped over to the same atom separately.

frontend/src/community/configurations/components/molecules/WorkLocationEmployeeSelector/WorkLocationEmployeeSelector.tsx
thusala: Done in 4692d00c1 — named `EmployeeAvatarChipProps` / `EmployeeGroupAvatarProps` interfaces, and both now take the narrow `EmployeeAvatarData` (employeeId/firstName/lastName/authPic) instead of `AllEmployeeDataType`. The casts at 148/221/246 stay: they come from `useGetSearchedEmployees`, whose preprocessor returns `EmployeeDataType[]`, not from these props.

frontend/src/community/configurations/components/molecules/WorkLocationEmployeeSelector/WorkLocationEmployeeSelector.tsx
thusala: Done in 4692d00c1 — the empty state is now gated on a pending search (debounce window + `isFetching`) and shows the loader instead, and it also no longer renders while a selected employee still matches locally.

frontend/src/community/configurations/components/molecules/WorkLocationEmployeeSelector/WorkLocationEmployeeSelector.tsx
thusala: Not applicable here. The DS `AvatarGroup` takes a flat `avatars: AvatarProps[]` and renders no children, so the `src` has to be resolved before the array is built — but profile pictures go through `useGetImageUrl` per employee (community `useGetUploadedImage` + enterprise S3), and a hook cannot be called inside a `.map()`. ProjectsTable can use it because it resolves URLs from a single `useS3Download` map, which is enterprise-only. Rendering matches the DS group exactly otherwise: `AvatarGroup` itself renders `flex -space-x-3` plus a surplus `<Avatar count title>`, so spacing, stacking order and surplus styling are identical — no drift. The joined-name `aria-label` is parity with the MUI group this replaces, which built its label the same way.

frontend/src/community/configurations/components/molecules/WorkLocationEmployeeSelector/WorkLocationEmployeeSelector.tsx
thusala: Not an issue — the DS `AvatarChip` sets `title={displayLabel}` on its own truncating label div (`min-w-0 flex-1 truncate ... title=`), so the full name is revealed on hover without any extra wrapper.

frontend/src/community/configurations/components/molecules/WorkLocationEmployeeSelector/WorkLocationEmployeeSelector.tsx
thusala: Verified against `1.1.1-beta.13`: both are declared on `AvatarProps` and used — `count` renders the `+N` variant and `title` is spread onto the avatar div. Same props the DS `AvatarGroup` itself passes to its surplus avatar.

frontend/src/community/configurations/components/molecules/WorkLocationEmployeeSelector/WorkLocationEmployeeSelector.tsx
thusala: `showActionButton` already defaults to `false` in the DS `AvatarChip`, so there is no dead remove button to steal the click. Left `backgroundColor` at the DS default (`bg-white`), which is what the other chip call sites use and reads fine on the tertiary trigger.

frontend/src/community/configurations/components/molecules/WorkLocationEmployeeSelector/WorkLocationEmployeeSelector.tsx
thusala: Half done in 4692d00c1 — `alt` added to the chip's `avatarProps`. Kept `src: imageUrl ?? ""`: the DS `Avatar` guards with `if (src)` before rendering an ``, so an empty string falls through to the initials avatar and no empty-src request is made.

frontend/src/community/configurations/components/molecules/WorkLocationEmployeeSelector/WorkLocationEmployeeSelector.tsx
thusala: 4 visible is intentional — the DS group defaults to `maxVisible: 5`, and 4 `sm` avatars still fit the `h-12` trigger. Magic number extracted in 4692d00c1 as `MAX_INLINE_CHIPS = 2` next to `MAX_VISIBLE_AVATARS`.

frontend/src/community/configurations/components/molecules/WorkLocationEmployeeSelector/WorkLocationEmployeeSelector.tsx
thusala: Skipping in this PR — the whole `configurations` module has no `.test.tsx` yet, so this would be a new test setup rather than a change to the PR. Worth doing as its own task.

thusala: Out of scope for this PR. The dead `employeeMultiSelect` block predates this branch (it is not in the diff), and swapping the bespoke multi-select for `AvatarChipsInput` is a behaviour rewrite, not the avatar-component migration this PR does. Happy to raise both as a follow-up.

frontend/src/community/configurations/components/molecules/WorkLocationEmployeeSelector/WorkLocationEmployeeSelector.tsx
thusala: Done in 816e9387e — `EmployeeAvatarChip` now derives its avatar id from `useId()`, so every instance is unique no matter how many times the same employee is rendered. The All Employees chip takes a `trigger`/`option` prefix since it is rendered from a plain function, not a component.

frontend/src/community/common/components/atoms/EmployeeAvatarChip/EmployeeAvatarChip.tsx
```diff
@@ -0,0 +1,46 @@
...
+  lastName: string;
+  authPic?: string;
+}
+
+export interface EmployeeAvatarChipProps {
+  employee: EmployeeAvatarData;
+  className?: string;
+}
+
+export const getEmployeeAvatarName = (employee: EmployeeAvatarData): string =>
```
hasalarootcode: move this util out

hasalarootcode (changes requested):

frontend/src/community/common/components/atoms/EmployeeAvatarChip/EmployeeAvatarChip.tsx
thusala: Done in c114bfb75 — `getEmployeeAvatarName` moved to `~community/common/utils/commonUtil` next to `concatStrings`, and its `EmployeeAvatarData` type to `~community/common/types/CommonTypes`.

hasalarootcode (approved):

akilarootcode (approved):

PR #1810, opened 2026-08-20
fix: trap focus in birthday notification modal

SanithuGunasena07:
Marks the dashboard (drawer, app bar, and main content) as inert while the birthday notification modal is open, so keyboard navigation and screen readers stay confined to the modal instead of leaking into the Dashboard behind it. Restores automatically when the modal closes.


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR isolates the background from screen readers while a birthday modal is open by tagging the drawer layout root with a DOM id and imperatively applying `inert` from `BirthdayModalShell`. The mechanism is sound for the main content region (the modal is rendered as a sibling of `ContentWithDrawer` in `BaseLayout`, so it is not itself inerted, and `BirthdayNotificationProvider` only marks the modal eligible on drawer routes so the target element exists). However, the inert region does not cover several focusable siblings — most notably `SkipToContentPopup` — so the screen-reader/keyboard trap is incomplete and skip links now silently target inert content; the hardcoded, cross-module DOM id and the mount-time node capture are also fragile.

Found **12** new issue(s): 🔴 3 important, 🟡 5 suggestion(s), 🟣 4 nit(s)

frontend/src/community/common/components/organisms/ContentWithDrawer/ContentWithDrawer.tsx
```diff
@@ -31,7 +31,7 @@ const ContentWithDrawer = ({ children }: Props) => {
 
   return (
     <>
-      <Stack sx={classes.protectedWrapper}>
+      <Stack id="content-with-drawer-root" sx={classes.protectedWrapper}>
```
ThinuwanW, line 34: 🤖 **Claude** · 🔴 **Important**

The inert region is scoped to this Stack only, but several focusable elements live OUTSIDE it and remain reachable while the birthday modal is open, so the screen-reader/focus trap this PR is meant to fix is still leaky:

1. `SkipToContentPopup` is rendered in `BaseLayout` as a sibling of `<ContentWithDrawer>` (BaseLayout.tsx lines 77-81), and its Popper is portaled to `document.body`. A keyboard/SR user can still tab to `#skip-to-content-button`, open the popper, and activate a link. Worse, those links now point into the inert subtree (`#content-with-drawer-main-content`, `#side-bar`): `SkipToContentPopup` does `setAttribute("tabindex", "0")` then `focus({preventScroll:true})` — on an inert element `focus()` is a no-op, so focus silently drops to `<body>`, the one-time `blur` listener never fires, and the injected `tabindex="0"` is left behind on `#side-bar`. Net effect for a keyboard user: the skip menu appears to work but does nothing.
2. `ToastMessage`, `TimeWidgetPopupController`, `MyRequestModalController`, `QuickSetupModalController` (lines 80-95 of this file) and `CommonModalController` (BaseLayout) are all outside this Stack and stay interactive/announced.

Fix options: (a) move the `inert` target up so it covers everything behind the modal — e.g. wrap the whole `renderComponent` + controllers in a single `#app-root` and inert that, or iterate `document.body.children` skipping the modal portal; or (b) at minimum also inert/unmount `SkipToContentPopup` while a modal is open. Whatever is chosen, `SkipToContentPopup` should not offer navigation targets that are inert.

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
```diff
@@ -48,6 +48,16 @@ const BirthdayModalShell: FC<Props> = ({
 
     return () => clearTimeout(timer);
   }, [showConfetti]);
+  
+  useEffect(() => {
+    const mainContent = document.getElementById("content-with-drawer-root");
```

The DOM id `"content-with-drawer-root"` is a magic string duplicated in two modules (`~community/people` reaching into a `~community/common` organism's DOM). If someone renames the id on the Stack in `ContentWithDrawer.tsx` there is no compile error — the a11y fix just silently stops working. The codebase already centralises layout element ids in `~community/common/constants/commonConstants.ts` (`signedInUserSkipToContentList` / `unsignedInUserSkipToContentList` reference `#content-with-drawer-main-content` etc.). Export a constant such as `export const CONTENT_WITH_DRAWER_ROOT_ID = "content-with-drawer-root";` from that file and use it in both `ContentWithDrawer.tsx` (`id={CONTENT_WITH_DRAWER_ROOT_ID}`) and here.

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
```diff
@@ -48,6 +48,16 @@ const BirthdayModalShell: FC<Props> = ({
...
+
+    mainContent?.setAttribute("inert", "");
+
+    return () => {
+      mainContent?.removeAttribute("inert");
```

The effect captures the DOM node once at mount (`[]` deps) and holds that reference for the whole lifetime of the modal. If `ContentWithDrawer` unmounts and remounts while the modal is still open, the captured node becomes detached: the new layout root never gets `inert` (background becomes interactive again behind the modal) and the cleanup mutates a garbage node. This is reachable in `BaseLayout` — `renderComponent` swaps to `<FullScreenLoader/>` whenever `isEnterprise && isGlobalLoginMethodLoading` flips, and to `ContentWithoutDrawer` if `isProtectedRouteWithDrawer` becomes false (e.g. a programmatic `router.push` from a notification handler) — while `BirthdayModalController` lives outside `renderComponent` and keeps the modal mounted. Two ways to harden: re-query in the cleanup (`document.getElementById(...)?.removeAttribute("inert")`, matching the existing pattern in `ConsentModalWithOverlay.tsx`), and/or make `inert` React-owned by driving it from shared state (a context/Zustand flag like `isBlockingModalOpen`) so `ContentWithDrawer` renders the attribute itself and it survives remounts.

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
```diff
@@ -48,6 +48,16 @@ const BirthdayModalShell: FC<Props> = ({
 
     return () => clearTimeout(timer);
   }, [showConfetti]);
+  
+  useEffect(() => {
```

This is the second ad-hoc implementation of "hide the background from assistive tech while an overlay is open" in the codebase — `~enterprise/sign/.../ConsentModalWithOverlay.tsx` (lines 24-54) does the same thing against `#content-without-drawer-main-content` but with `aria-hidden="true"` plus `document.body.style.overflow` locking. Two different mechanisms for one concern will drift. Extract a shared hook (e.g. `~community/common/hooks/useInertBackground(elementId)`) that encapsulates set/remove + cleanup, and reuse it in both places (`inert` is the stronger option since it blocks focus and pointer events, not just the accessibility tree). Also worth deciding here whether background scroll should be locked while the birthday modal is open, as the sign flow does — `inert` alone does not prevent wheel/scroll behind the modal.

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
```diff
@@ -48,6 +48,16 @@ const BirthdayModalShell: FC<Props> = ({
...
+    const mainContent = document.getElementById("content-with-drawer-root");
+
+    mainContent?.setAttribute("inert", "");
+
+    return () => {
```

Possible race with focus restoration. `useBirthdayNotifications.finishSequence()` restores focus inside a `requestAnimationFrame` callback to `restoreFocusRef.current`, which is the element that was focused before the modal opened — i.e. almost always an element inside the now-inert `#content-with-drawer-root`. That restore only works if this cleanup (`removeAttribute("inert")`) has already run. React 18 flushes passive-effect destroy functions on a scheduler (MessageChannel) task, which in practice lands before the next rAF, so it usually works — but the ordering is not guaranteed by React, and if it loses the race `focus()` is a silent no-op on an inert element and the screen-reader user is dumped back to `<body>`. Consider making the restore robust rather than relying on timing: remove `inert` synchronously in the dismiss path (or in a `useLayoutEffect` cleanup instead of `useEffect`), or have `finishSequence` verify `document.activeElement === elementToRestore` and retry/fall back.

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
```diff
@@ -48,6 +48,16 @@ const BirthdayModalShell: FC<Props> = ({
 
     return () => clearTimeout(timer);
   }, [showConfetti]);
+
```
ThinuwanW, line 52: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

No test covers this behaviour, and there is currently no `BirthdayModalShell.test.tsx` (the only birthday tests are util-level: `birthdayNotificationUtils.test.ts`, `birthdayNotificationCacheUtils.test.ts`). Since this is a pure a11y side effect with no visible UI, it is exactly the kind of change that silently regresses — e.g. if the id on the `Stack` changes. A small RTL test with the `MockTheme` wrapper that renders a container with `id="content-with-drawer-root"`, mounts the shell, asserts `toHaveAttribute("inert")`, then unmounts and asserts the attribute is gone, would lock the contract in.

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx

Making `#content-with-drawer-root` inert silently breaks the app's skip-to-content feature and leaves permanent DOM damage behind. `SkipToContentPopup` is rendered in `BaseLayout` OUTSIDE `ContentWithDrawer`, so its trigger button stays focusable and is the first focusable element in the document — a keyboard user tabbing out of the birthday modal lands on it and can open the popup. But every one of its targets lives INSIDE the newly-inert subtree: `#content-with-drawer-main-content` (ContentWithDrawer.tsx:63), `#side-bar` (Drawer.tsx:239), and in enterprise mode the extra items from `useGetEnterpriseSkipToContentItems` (e.g. the quick-setup button in the drawer). `HTMLElement.focus()` on a node inside an inert subtree is a no-op, so the popup's effect (SkipToContentPopup.tsx:61-88) does the following: sets `tabindex="0"` on the target, calls `focus()` which does nothing (focus drops to `<body>`, stranding the keyboard user), and registers a `{ once: true }` blur listener that can never fire because the element never received focus — so `tabindex="0"` is left on `#side-bar` forever, adding a stray tab stop to the drawer even after the modal is dismissed. Fix: in SkipToContentPopup's focus effect, bail out before mutating anything when the target is unreachable, e.g. `if (!focusedElement || focusedElement.closest("[inert]")) { setFocusedItem(null); return; }`; and/or suppress rendering the skip popup while a blocking overlay is open. If you also want the skip button itself neutralized, the inert boundary needs to sit above it rather than on `content-with-drawer-root`.

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
ThinuwanW, line 52: 🤖 **Claude** · 🔴 **Important**

📁 **File-level observation**

This community component imports `ImageSkeleton` from `~enterprise/common/components/atoms/ImageSkeleton/ImageSkeleton`, but no fallback stub exists — `src/fallback/common/components/atoms/` contains only `Badge` and `driverJsPopover`. Two concrete consequences: (1) `jest.config.js` maps `^~enterprise(.*)$` to `<rootDir>/src/fallback$1` unconditionally, so ANY test that imports `BirthdayModalShell` (directly, or transitively via `BirthdayModalController` / `SelfBirthdayModal` / `ColleagueBirthdayModal`) fails with a module-resolution error — which is why the accessibility behaviour added here cannot be covered by a test at all, blocking the missing-test issue raised in the first pass. (2) `tsconfig.json` resolves `~enterprise/*` to `["./src/enterprise/*", "./src/fallback/*"]`, and `frontend/src/enterprise` is a git submodule (`.gitmodules`), so in a community-only checkout where the submodule is absent the import falls through to the non-existent fallback path and the build fails. Fix: add `src/fallback/common/components/atoms/ImageSkeleton/ImageSkeleton.tsx` following the shape of the existing `src/fallback/common/components/atoms/Badge` stub, or (better, since a plain skeleton has no enterprise-specific behaviour) move `ImageSkeleton` into `~community/common/components/atoms/`.

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
```diff
@@ -48,6 +48,16 @@ const BirthdayModalShell: FC<Props> = ({
 
     return () => clearTimeout(timer);
   }, [showConfetti]);
+  
+  useEffect(() => {
+    const mainContent = document.getElementById("content-with-drawer-root");
+
+    mainContent?.setAttribute("inert", "");
```

`inert` is applied with no `aria-hidden` fallback, so on browsers without `inert` support (Firefox < 112, Safari < 15.5, Chromium < 102) this line sets an attribute that does literally nothing and the screen-reader escape this PR is fixing silently persists — with no visible signal that the fix is inert-in-name-only. There is no `wicg-inert` polyfill in `frontend/package.json`, and the existing precedent in this codebase (`~enterprise/sign/.../ConsentModalWithOverlay.tsx:30`) uses `aria-hidden="true"` instead. Since screen-reader users are disproportionately on older browser/AT combinations, set both and clear both in the cleanup: `setAttribute("inert", "")` plus `setAttribute("aria-hidden", "true")`. (`aria-hidden` alone would leave the background keyboard-focusable, `inert` alone leaves older browsers unfixed — the pair covers both.)

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx

No background scroll lock. `inert` blocks focus, clicks and text selection inside the subtree, but it does not prevent scrolling — mouse-wheel/trackpad scrolling of the page behind the modal still works. The equivalent overlay in this codebase pairs the attribute mutation with `document.body.style.overflow = "hidden"` and restores it in the cleanup (`ConsentModalWithOverlay.tsx:33,42,52`). Please confirm whether `LargeModal` from `@rootcodelabs/skapp-ui` already locks body scroll; if it does not, add the lock/restore to this same effect so the two overlay implementations behave consistently.

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx

The cleanup unconditionally removes `inert` rather than only removing what this effect added, so there is no protection against clobbering another overlay's state. `ContentWithDrawer` already renders three sibling overlay controllers (`TimeWidgetPopupController`, `MyRequestModalController`, `QuickSetupModalController`) plus `ToastMessage`; the moment any of them adopts the same pattern on the same root, whichever unmounts first re-enables the background for the other. Cheap guard: capture whether the attribute was already present before setting it (`const wasInert = mainContent?.hasAttribute("inert")`) and only `removeAttribute` in the cleanup when `!wasInert`. Better still, if this ends up extracted into the shared hook the first pass suggested, give it a module-level reference count.

frontend/src/community/common/components/organisms/ContentWithDrawer/ContentWithDrawer.tsx
ThinuwanW, line 34: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

Dead code in a file this PR touches: `initialState` is imported from `~community/common/providers/ToastProvider` (line 13) but never referenced anywhere in the component — only `useToast` is used. Drop it from the import. The identical unused import exists in `ContentWithoutDrawer.tsx:8`, so both can be cleaned up together.

SanithuGunasena07: not required

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
SanithuGunasena07: not required as already following the practice used in codebase

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
SanithuGunasena07: out of scope for PR

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
SanithuGunasena07: not required

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
SanithuGunasena07: not required

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
SanithuGunasena07: not required

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
SanithuGunasena07: safe to skip no crashes, no impact on anything else in the app.

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
SanithuGunasena07: Pre existing out of scope for PR

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
SanithuGunasena07: Added the aria-hidden="true" fallback alongside inert in BirthdayModalShell.tsx both are set on open and removed together on close, so older browsers without inert support still get the screen-reader fix via aria-hidden

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
SanithuGunasena07: Out of scope for PR

frontend/src/community/people/components/molecules/BirthdayModals/BirthdayModalShell.tsx
SanithuGunasena07: not required  nothing else touches inert on that element

frontend/src/community/common/components/organisms/ContentWithDrawer/ContentWithDrawer.tsx
SanithuGunasena07: out of scope for PR as its pre exisitng

hasalarootcode (approved):

frontend/src/community/common/hooks/useInertBackground.ts
```diff
@@ -0,0 +1,17 @@
+import { useEffect } from "react";
+
+const useInertBackground = (elementId: string): void => {
```
akilarootcode: check the actual fix

PR #1812, opened 2026-08-20
Fix unauthorized breadcrumbs

erandi-rootcode:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

The change correctly fixes stale breadcrumbs on `/unauthorized`: that route is in `middleware.ts` `config.matcher`, so `IsAProtectedUrlWithDrawer` returns true and `BaseLayout` renders `ContentWithDrawer` → `AppBar` → `Breadcrumb items={breadcrumbs}`, while the page itself does not render `ContentLayout` (the only component that calls `setBreadcrumbs`). The zustand selector, the stable `setBreadcrumbs` dependency, and the empty-array reset are all correct, and `items={[]}` is an already-exercised path (e.g. `pages/community/notifications.tsx` uses `ContentLayout` with no `breadcrumbs` prop). No correctness, type-safety, or i18n issues found; my only concern is that the fix is per-page rather than addressing the same defect class globally.

Found **6** new issue(s): 🟡 4 suggestion(s), 🟣 2 nit(s)

frontend/pages/community/unauthorized.tsx
```diff
@@ -1,12 +1,20 @@
...
 import AccessDeniedCard from "~community/common/components/atoms/AcessDeniedCard/AccessDeniedCard";
 import { useTranslator } from "~community/common/hooks/useTranslator";
+import { useCommonStore } from "~community/common/stores/commonStore";
 
 const Unauthorized: NextPage = () => {
   const translateText = useTranslator("unauthorized");
 
+  const setBreadcrumbs = useCommonStore((state) => state.setBreadcrumbs);
+
+  useEffect(() => {
```
ThinuwanW, line 14: 🤖 **Claude** · 🟡 **Suggestion**

This fixes the symptom on one page, but the same stale-breadcrumb bug remains on every drawer-rendered route that doesn't render `ContentLayout`. Breadcrumbs are only ever written by `ContentLayout` (`src/community/common/components/templates/ContentLayout/ContentLayout.tsx:243-245`, a mount-only effect), so any page rendered inside `ContentWithDrawer` without it keeps the previous page's breadcrumbs. Concretely, `pages/404.tsx` (and `pages/_error.tsx`) hit this: for a path like `/people/does-not-exist`, middleware allows the request, `IsAProtectedUrlWithDrawer` matches the `/people/:path*` entry in `config.matcher`, so the 404 renders with the drawer and AppBar still showing the breadcrumbs of the page you came from. Consider centralizing the reset instead of adding this boilerplate per page — e.g. clear breadcrumbs in `AppBar`/`BaseLayout` on `router.events.on("routeChangeComplete")`, or have `ContentLayout` clear on unmount, or expose a tiny `useResetBreadcrumbs()` hook and use it here and in `404.tsx`. An alternative that also fits existing conventions is wrapping this page in `ContentLayout` (with `isTitleHidden` and no `breadcrumbs` prop), which already resets to `[]` via its default.

ThinuwanW, line 14: 🤖 **Claude** · 🟡 **Suggestion**

The reset lands too late to prevent the stale trail from being painted. `ContentWithDrawer` renders `<AppBar />` *outside* the `loading ? <FullScreenLoader/> : <main>{children}</main>` ternary (ContentWithDrawer.tsx:35-77), and `useRouteLoading` sets `loading` true on `routeChangeStart` and false on `routeChangeComplete`. So during the entire navigation into /unauthorized the page component is unmounted, the AppBar remains mounted, and it keeps rendering the *previous* route's breadcrumbs. Only after `routeChangeComplete` does this page mount, render once (still with the old crumbs in the store), and then this post-paint `useEffect` blanks them. The user sees the old trail for the whole transition and then watches it disappear. A `routeChangeStart`-scoped reset would avoid the flash entirely — e.g. in `AppBar` or `ContentWithDrawer`, subscribe `router.events.on("routeChangeStart", () => setBreadcrumbs([]))` so the trail is cleared before the loader appears and each page then sets its own. That also removes the need for pages to opt in individually.

frontend/pages/community/unauthorized.tsx
```diff
@@ -1,12 +1,20 @@
 import { NextPage } from "next";
 import Head from "next/head";
+import { useEffect } from "react";
```
ThinuwanW, line 3: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

This page compensates for a mount-only effect in the shared layout that is worth fixing at the source. `ContentLayout.tsx:243-245` is `useEffect(() => { setBreadcrumbs(breadcrumbs); }, [])` — an empty dependency array, so the `breadcrumbs` prop is captured once at mount and the store is never re-synced while the component stays mounted. This violates `react-hooks/exhaustive-deps` and means the breadcrumb store can never be corrected in place; it only happens to work today because `ContentWithDrawer` unmounts `children` on every route change. Note the naive fix is wrong: every call site passes a fresh array literal (e.g. `configurations/user-roles/[module].tsx:50-56`, `leave/types/[slug].tsx:116-130`), so adding `breadcrumbs` to the deps would re-run the effect on every parent render. Either memoize at the call sites, or guard inside the effect by comparing a stable derived key (e.g. `breadcrumbs.map(b => b.label).join("|")`) and use that as the dependency. At minimum add a comment explaining why the array is intentionally empty, so the next reader does not "fix" it into a render loop.

frontend/pages/community/unauthorized.tsx
```diff
@@ -1,12 +1,20 @@
...
 
 import AccessDeniedCard from "~community/common/components/atoms/AcessDeniedCard/AccessDeniedCard";
 import { useTranslator } from "~community/common/hooks/useTranslator";
+import { useCommonStore } from "~community/common/stores/commonStore";
 
 const Unauthorized: NextPage = () => {
   const translateText = useTranslator("unauthorized");
 
+  const setBreadcrumbs = useCommonStore((state) => state.setBreadcrumbs);
```
ThinuwanW, line 12: 🤖 **Claude** · 🟡 **Suggestion**

There are now two independent, hand-rolled writers into the breadcrumb slice — this page and `ContentLayout.tsx:223/243-245` — each repeating the same select-setter-then-effect boilerplate with no shared abstraction. Per the duplication guideline (2+ components with the same logic → extract a custom hook), pull this into e.g. `~community/common/hooks/useSetBreadcrumbs.ts` exporting `useSetBreadcrumbs(items: BreadcrumbItem[]): void`, and have both `ContentLayout` and this page call it. That gives one place to fix the dependency-array/staleness semantics described above instead of two, and makes the "a drawer route must declare its breadcrumbs" contract discoverable for the next page that needs it.

frontend/pages/community/unauthorized.tsx
```diff
@@ -1,12 +1,20 @@
...
+
+  useEffect(() => {
+    setBreadcrumbs([]);
```
ThinuwanW, line 15: 🤖 **Claude** · 🟣 **Nit**

This reset re-renders the whole AppBar even when it is a no-op. `breadcrumbSlice.ts:14-18` unconditionally does `set((state) => ({ ...state, breadcrumbs: items }))` — it spreads the entire state (unnecessary, Zustand already shallow-merges) and always stores a new array reference, so the state object identity always changes. `AppBar.tsx:46-48` subscribes with `useCommonStore((state) => state)`, i.e. the whole store with default `Object.is` equality, so any commonStore write re-renders the entire AppBar subtree — `ClockWidget`, the avatar, the notification badge. Arriving on /unauthorized when breadcrumbs are already `[]` therefore costs a full AppBar re-render for no visual change. Two cheap improvements: early-return in the slice when the incoming items are reference- or shallow-equal to current (`set((state) => state.breadcrumbs === items || (state.breadcrumbs.length === 0 && items.length === 0) ? state : { breadcrumbs: items })`), and narrow the AppBar selector to `useCommonStore(useShallow((state) => ({ notifyData: state.notifyData, setNotifyData: state.setNotifyData, breadcrumbs: state.breadcrumbs })))`.

frontend/pages/community/unauthorized.tsx
```diff
@@ -1,12 +1,20 @@
 import { NextPage } from "next";
 import Head from "next/head";
+import { useEffect } from "react";
 
 import AccessDeniedCard from "~community/common/components/atoms/AcessDeniedCard/AccessDeniedCard";
```
ThinuwanW, line 5: 🤖 **Claude** · 🟣 **Nit**

Naming violation on the import path this file depends on: the directory is spelled `AcessDeniedCard` (single `c`) while the component and file inside it are correctly `AccessDeniedCard`. Since this PR is already touching the import block of the only page that renders it, it is a good moment to rename `src/community/common/components/atoms/AcessDeniedCard/` → `AccessDeniedCard/` and update this import. Worth confirming it is the sole reference first — a grep shows the component is only used here.

frontend/pages/community/unauthorized.tsx
erandi-rootcode: Not related to this PR

frontend/pages/community/unauthorized.tsx
erandi-rootcode: Not Valid

These two call sites have different semantics — ContentLayout writes a prop-driven trail, this page performs an explicit reset — and each is three lines. Extracting a shared hook now would be a premature abstraction. It's also likely moot: if we centralize the reset as suggested in the other comments, this page's breadcrumb code goes away entirely.

frontend/pages/community/unauthorized.tsx
erandi-rootcode: Not valid for this PR

Both the state spread in breadcrumbSlice and the whole-store selector in AppBar are pre-existing shared code, untouched by this PR. The measurable cost here is one extra AppBar render per /unauthorized visit. Happy to raise it as a separate perf ticket rather than widen a bugfix diff.

frontend/pages/community/unauthorized.tsx
erandi-rootcode: Not valid for this PR

renaming AcessDeniedCard/ is an unrelated directory move. Keeping it out so this bugfix diff stays reviewable — raising it separately.

frontend/pages/community/unauthorized.tsx
erandi-rootcode: Out of scope for this PR

frontend/pages/community/unauthorized.tsx
erandi-rootcode: Out of scope for this PR

frontend/pages/community/unauthorized.tsx
```diff
@@ -1,12 +1,20 @@
...
+import { useCommonStore } from "~community/common/stores/commonStore";
 
 const Unauthorized: NextPage = () => {
   const translateText = useTranslator("unauthorized");
 
+  const setBreadcrumbs = useCommonStore((state) => state.setBreadcrumbs);
+
+  useEffect(() => {
+    setBreadcrumbs([]);
+  }, [setBreadcrumbs]);
```
shakila-kamalasena: Remove the setter from dependency array

shakila-kamalasena (changes requested):

frontend/pages/community/unauthorized.tsx
erandi-rootcode: removed

shakila-kamalasena (approved):

akilarootcode (approved):

PR #1815, opened 2026-08-20
Fix/3010: Leave Policy - Duplicate policy name error shows as a toast instead of an inline field error

hasalarootcode:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/backend/src/main/java/com/skapp/community/leaveplanner/controller/v1/LeavePolicyController.java
+++ b/backend/src/main/java/com/skapp/community/leaveplanner/controller/v1/LeavePolicyController.java
@@ -18,6 +18,7 @@ import org.springframework.web.bind.annotation.PostMapping;
 import org.springframework.web.bind.annotation.PutMapping;
 import org.springframework.web.bind.annotation.RequestBody;
 import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestParam;
 import org.springframework.web.bind.annotation.RestController;
 
 @RequiredArgsConstructor
@@ -34,6 +35,17 @@ public class LeavePolicyController {
 	@PreAuthorize("hasAnyRole('ROLE_LEAVE_ADMIN', 'ROLE_PEOPLE_ADMIN')")
 	public ResponseEntity<ResponseEntityDto> getAllLeavePolicies(LeavePolicyFilterDto leavePolicyFilterDto) {
 		ResponseEntityDto response = leavePolicyService.getAllLeavePolicies(leavePolicyFilterDto);
+		return new ResponseEntity<>(response, HttpStatus.OK);
+	}
+
+	@Operation(summary = "Check leave policy name availability",
+			description = "Returns whether the given policy name is still available for the given leave type. "
+					+ "Names are compared case-insensitively across both active and inactive policies")
+	@GetMapping("/name-availability")
+	@PreAuthorize("hasAnyRole('ROLE_LEAVE_ADMIN')")
+	public ResponseEntity<ResponseEntityDto> checkLeavePolicyNameAvailability(
+			@RequestParam(required = false) String name, @RequestParam(required = false) Long leaveTypeId) {
+		ResponseEntityDto response = leavePolicyService.checkLeavePolicyNameAvailability(name, leaveTypeId);
 		return new ResponseEntity<>(response, HttpStatus.OK);
 	}
 
--- a/backend/src/main/java/com/skapp/community/leaveplanner/payload/response/LeavePolicyNameAvailabilityResponseDto.java
+++ b/backend/src/main/java/com/skapp/community/leaveplanner/payload/response/LeavePolicyNameAvailabilityResponseDto.java
@@ -1,0 +1,16 @@
+package com.skapp.community.leaveplanner.payload.response;
+
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+
+@Getter
+@Setter
+@NoArgsConstructor
+@AllArgsConstructor
+public class LeavePolicyNameAvailabilityResponseDto {
+
+	private Boolean isAvailable;
+
+}
--- a/backend/src/main/java/com/skapp/community/leaveplanner/service/impl/LeavePolicyServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/leaveplanner/service/impl/LeavePolicyServiceImpl.java
@@ -21,6 +21,7 @@ import com.skapp.community.leaveplanner.payload.request.LeavePolicyFilterDto;
 import com.skapp.community.leaveplanner.payload.request.LeavePolicyRequestDto;
 import com.skapp.community.leaveplanner.payload.request.LeavePolicyUpdateRequestDto;
 import com.skapp.community.leaveplanner.payload.response.LeavePolicyConfigResponseDto;
+import com.skapp.community.leaveplanner.payload.response.LeavePolicyNameAvailabilityResponseDto;
 import com.skapp.community.leaveplanner.payload.response.LeavePolicyResponseDto;
 import com.skapp.community.leaveplanner.payload.response.LeavePolicyStatusResponseDto;
 import com.skapp.community.leaveplanner.repository.LeaveEntitlementDao;
@@ -179,6 +180,23 @@ public class LeavePolicyServiceImpl implements LeavePolicyService {
 
 		log.info("getAllLeavePolicies: execution ended");
 		return new ResponseEntityDto(false, pageDto);
+	}
+
+	@Override
+	@Transactional(readOnly = true)
+	public ResponseEntityDto checkLeavePolicyNameAvailability(String name, Long leaveTypeId) {
+		log.info("checkLeavePolicyNameAvailability: execution started");
+
+		if (leaveTypeId == null) {
+			throw new ModuleException(LeaveMessageConstant.LEAVE_ERROR_LEAVE_POLICY_LEAVE_TYPE_REQUIRED);
+		}
+
+		LeavePolicyValidationUtil.validateName(name);
+
+		boolean isAvailable = !leavePolicyDao.existsByNameIgnoreCaseAndLeaveType_Id(name, leaveTypeId);
+
+		log.info("checkLeavePolicyNameAvailability: execution ended");
+		return new ResponseEntityDto(false, new LeavePolicyNameAvailabilityResponseDto(isAvailable));
 	}
 
 	@Override
--- a/backend/src/test/java/com/skapp/community/leaveplanner/controller/v1/LeavePolicyControllerIntegrationTest.java
+++ b/backend/src/test/java/com/skapp/community/leaveplanner/controller/v1/LeavePolicyControllerIntegrationTest.java
@@ -46,6 +46,9 @@ class LeavePolicyControllerIntegrationTest {
 	private static final String SEED_LEAVE_TYPE = "INSERT INTO lv_leave_type (id, name, emoji_code, color_code, min_duration, is_attachment, is_attachment_must, is_comment_must, is_auto_approval, is_active) "
 			+ "VALUES (100, 'PolicyAnnual', 'U+1F3D6', '#FFC107', 'FULL_DAY', false, false, false, false, true)";
 
+	private static final String SEED_SECOND_LEAVE_TYPE = "INSERT INTO lv_leave_type (id, name, emoji_code, color_code, min_duration, is_attachment, is_attachment_must, is_comment_must, is_auto_approval, is_active) "
+			+ "VALUES (101, 'PolicyCasual', 'U+1F3D6', '#FFC107', 'FULL_DAY', false, false, false, false, true)";
+
 	private static final String SEED_POLICY = "INSERT INTO lv_leave_policy (id, name, leave_type_id, policy_type, status, is_carryover_enabled) "
 			+ "VALUES (500, 'Existing Policy', 100, 'ACCRUAL', 'ACTIVE', false)";
 
@@ -112,6 +115,13 @@ class LeavePolicyControllerIntegrationTest {
 	private ResultActions performGetAll(String authToken) throws Exception {
 		return mvc
 			.perform(get(ENDPOINT).accept(MediaType.APPLICATION_JSON).with(SecurityTestUtils.bearerToken(authToken)));
+	}
+
+	private ResultActions performNameAvailability(String authToken, String name, String leaveTypeId) throws Exception {
+		return mvc.perform(get(ENDPOINT + "/name-availability").param("name", name)
+			.param("leaveTypeId", leaveTypeId)
+			.accept(MediaType.APPLICATION_JSON)
+			.with(SecurityTestUtils.bearerToken(authToken)));
 	}
 
 	private ResultActions performEnable(String authToken) throws Exception {
@@ -340,6 +350,93 @@ class LeavePolicyControllerIntegrationTest {
 				.with(SecurityTestUtils.bearerToken(leaveAdminToken())))
 				.andDo(print())
 				.andExpect(status().isNotFound());
+		}
+
+	}
+
+	@Nested
+	@DisplayName("Leave Policy Name Availability")
+	class LeavePolicyNameAvailabilityTests {
+
+		@Test
+		@DisplayName("Returns available when no policy holds the name for the leave type")
+		@Sql(statements = { SEED_LEAVE_TYPE, SEED_POLICY })
+		void checkNameAvailability_UnusedName_ReturnsAvailable() throws Exception {
+			performNameAvailability(leaveAdminToken(), "Brand New Policy", "100").andDo(print())
+				.andExpect(status().isOk())
+				.andExpect(jsonPath(STATUS_PATH).value(STATUS_SUCCESSFUL))
+				.andExpect(jsonPath("$.results[0].isAvailable").value(true));
+		}
+
+		@Test
+		@DisplayName("Returns unavailable when the name matches an existing policy ignoring case")
+		@Sql(statements = { SEED_LEAVE_TYPE, SEED_POLICY })
+		void checkNameAvailability_DuplicateNameDifferentCase_ReturnsUnavailable() throws Exception {
+			performNameAvailability(leaveAdminToken(), "existing policy", "100").andDo(print())
+				.andExpect(status().isOk())
+				.andExpect(jsonPath("$.results[0].isAvailable").value(false));
+		}
+
+		@Test
+		@DisplayName("Returns bad request when the name is missing")
+		@Sql(statements = { SEED_LEAVE_TYPE })
+		void checkNameAvailability_MissingName_ReturnsBadRequest() throws Exception {
+			mvc.perform(get(ENDPOINT + "/name-availability").param("leaveTypeId", "100")
+				.accept(MediaType.APPLICATION_JSON)
+				.with(SecurityTestUtils.bearerToken(leaveAdminToken())))
+				.andDo(print())
+				.andExpect(status().isBadRequest());
+		}
+
+		@Test
+		@DisplayName("Returns unavailable when the name is held by an inactive policy")
+		@Sql(statements = { SEED_LEAVE_TYPE, SEED_INACTIVE_POLICY })
+		void checkNameAvailability_InactivePolicyName_ReturnsUnavailable() throws Exception {
+			performNameAvailability(leaveAdminToken(), "Inactive Policy", "100").andDo(print())
+				.andExpect(status().isOk())
+				.andExpect(jsonPath("$.results[0].isAvailable").value(false));
+		}
+
+		@Test
+		@DisplayName("Returns available when the same name is used under a different leave type")
+		@Sql(statements = { SEED_LEAVE_TYPE, SEED_SECOND_LEAVE_TYPE, SEED_POLICY })
+		void checkNameAvailability_SameNameOtherLeaveType_ReturnsAvailable() throws Exception {
+			performNameAvailability(leaveAdminToken(), "Existing Policy", "101").andDo(print())
+				.andExpect(status().isOk())
+				.andExpect(jsonPath("$.results[0].isAvailable").value(true));
+		}
+
+		@Test
+		@DisplayName("Returns bad request when the name is blank")
+		@Sql(statements = { SEED_LEAVE_TYPE })
+		void checkNameAvailability_BlankName_ReturnsBadRequest() throws Exception {
+			performNameAvailability(leaveAdminToken(), "   ", "100").andDo(print()).andExpect(status().isBadRequest());
+		}
+
+		@Test
+		@DisplayName("Returns bad request when the leave type is missing")
+		void checkNameAvailability_MissingLeaveTypeId_ReturnsBadRequest() throws Exception {
+			mvc.perform(get(ENDPOINT + "/name-availability").param("name", "Brand New Policy")
+				.accept(MediaType.APPLICATION_JSON)
+				.with(SecurityTestUtils.bearerToken(leaveAdminToken())))
+				.andDo(print())
+				.andExpect(status().isBadRequest());
+		}
+
+		@Test
+		@DisplayName("Non-admin user cannot check policy name availability")
+		@Sql(statements = { SEED_LEAVE_TYPE, DOWNGRADE_USER2_TO_EMPLOYEE })
+		void checkNameAvailability_LeaveEmployee_ReturnsForbidden() throws Exception {
+			performNameAvailability(user2Token(), "Brand New Policy", "100").andDo(print())
+				.andExpect(status().isForbidden());
+		}
+
+		@Test
+		@DisplayName("Returns 401 when no authentication token is provided")
+		void checkNameAvailability_NoAuth_ReturnsUnauthorized() throws Exception {
+			mvc.perform(get(ENDPOINT + "/name-availability").param("name", "Brand New Policy")
+				.param("leaveTypeId", "100")
+				.accept(MediaType.APPLICATION_JSON)).andDo(print()).andExpect(status().isUnauthorized());
 		}
 
 	}
--- a/frontend/src/community/common/assets/languages/english/leaveModule.json
+++ b/frontend/src/community/common/assets/languages/english/leaveModule.json
@@ -476,7 +476,8 @@
         "accrualCapStepInvalid": "Maximum balance must be in multiples of 0.5.",
         "maxCarryOverDaysRequired": "Please enter the maximum number of carry over days.",
         "maxCarryOverDaysInvalid": "Maximum carry over days must be greater than 0 and at most 365.",
-        "maxCarryOverDaysStepInvalid": "Maximum carry over days must be in multiples of 0.5."
+        "maxCarryOverDaysStepInvalid": "Maximum carry over days must be in multiples of 0.5.",
+        "policyNameDuplicate": "A policy with this name already exists for the selected leave type. Please use a different name."
       },
       "successToastTitle": "Policy Created",
       "successToastDescription": "Policy has been created successfully and is now active.",
--- a/frontend/src/community/leave/api/LeavePolicyApi.ts
+++ b/frontend/src/community/leave/api/LeavePolicyApi.ts
@@ -13,12 +13,15 @@ import { leavePolicyEndPoints } from "~community/leave/api/utils/ApiEndpoints";
 import { leavePolicyQueryKeys } from "~community/leave/api/utils/QueryKeys";
 import {
   AddLeavePolicyPayload,
+  CheckLeavePolicyNameAvailabilityParams,
   GetLeavePoliciesInfiniteArgs,
   GetLeavePoliciesParams,
   LeavePoliciesResponse,
   LeavePolicyConfigResponse,
   LeavePolicyConfigResult,
   LeavePolicyMutationResponse,
+  LeavePolicyNameAvailabilityResponse,
+  LeavePolicyNameAvailabilityResult,
   UpdateLeavePolicyVariables
 } from "~community/leave/types/LeavePolicyTypes";
 
@@ -64,6 +67,29 @@ export const useGetLeavePoliciesInfinite = ({
     refetchOnWindowFocus: false
   });
 };
+
+const checkLeavePolicyNameAvailability = async (
+  params: CheckLeavePolicyNameAvailabilityParams
+): Promise<LeavePolicyNameAvailabilityResult> => {
+  const response = await authFetch.get<LeavePolicyNameAvailabilityResponse>(
+    leavePolicyEndPoints.CHECK_LEAVE_POLICY_NAME_AVAILABILITY,
+    { params }
+  );
+  return response.data.results[0];
+};
+
+export const useCheckLeavePolicyNameAvailability = (): UseMutationResult<
+  LeavePolicyNameAvailabilityResult,
+  AxiosError,
+  CheckLeavePolicyNameAvailabilityParams
+> =>
+  useMutation<
+    LeavePolicyNameAvailabilityResult,
+    AxiosError,
+    CheckLeavePolicyNameAvailabilityParams
+  >({
+    mutationFn: checkLeavePolicyNameAvailability
+  });
 
 const addLeavePolicy = (
   leavePolicy: AddLeavePolicyPayload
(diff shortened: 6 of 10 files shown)


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds a `GET /v1/leave/policies/name-availability` endpoint (LEAVE_ADMIN only) plus a frontend pre-check in the leave policy wizard so duplicate names surface as an inline error before the user walks through the whole wizard. The backend slice is clean and well covered by nine new integration tests (role gating, case-insensitivity, inactive policies, cross-leave-type isolation, blank/missing params), the service method has the required entry/exit logging and `@Transactional(readOnly = true)`, ModuleException correctly maps to 400, and the new message keys already exist in `leave-messages.properties`. No blocking defects found; the issues below are validation asymmetry with the create path, a redundant round-trip and a silent dead-end in the wizard, and a duplicated i18n string.

Found **18** new issue(s): 🟡 10 suggestion(s), 🟣 8 nit(s)

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/LeavePolicyServiceImpl.java
```diff
@@ -181,6 +182,23 @@ public ResponseEntityDto getAllLeavePolicies(LeavePolicyFilterDto leavePolicyFil
 		return new ResponseEntityDto(false, pageDto);
 	}
 
+	@Override
+	@Transactional(readOnly = true)
+	public ResponseEntityDto checkLeavePolicyNameAvailability(String name, Long leaveTypeId) {
+		log.info("checkLeavePolicyNameAvailability: execution started");
+
+		if (leaveTypeId == null) {
```
ThinuwanW, line 190: 🤖 **Claude** · 🟡 **Suggestion**

Asymmetric validation vs. the create path this endpoint mirrors. `addLeavePolicy` resolves the leave type with `policyLeaveTypeDao.findByIdAndIsActiveTrue(...)` and throws `EntityNotFoundException(LEAVE_ERROR_POLICY_LEAVE_TYPE_NOT_FOUND)` (404) when it is missing or deactivated. The availability check only rejects a null id, so `?leaveTypeId=99999` (or an id whose leave type was just deactivated) returns `200 {"isAvailable": true}` — the UI happily advances and the subsequent POST then fails with a 404 the wizard has no inline handling for. Suggest resolving the leave type the same way here (`policyLeaveTypeDao.findByIdAndIsActiveTrue(leaveTypeId).orElseThrow(() -> new EntityNotFoundException(LeaveMessageConstant.LEAVE_ERROR_POLICY_LEAVE_TYPE_NOT_FOUND))`) before running the exists check, so both endpoints agree on what a valid leave type is.

frontend/src/community/leave/components/organisms/LeavePolicyWizard/LeavePolicyWizard.tsx
```diff
@@ -170,6 +217,13 @@ const LeavePolicyWizard: FC<Props> = ({ policyType }) => {
       return;
     }
 
+    if (
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

For FLEXIBLE policies the pre-check is a pure extra round-trip. `isLastStep` is `!isAccrual || activeStep === SUMMARY`, so for a FLEXIBLE policy the wizard is a single BASIC_INFO step that is simultaneously the last step. Clicking "Create Policy" therefore fires the availability GET and then immediately the create POST — but the POST already returns `LEAVE_ERROR_LEAVE_POLICY_ALREADY_EXISTS`, which `handleError` maps to the exact same inline error via `showDuplicatePolicyNameError()`. The pre-check buys nothing here and just slows creation down (and widens the TOCTOU window). Suggest gating it on `!isLastStep` as well: `if (!isLastStep && activeStep === LeavePolicyWizardSteps.BASIC_INFO && !(await isPolicyNameAvailable())) return;`.

frontend/src/community/leave/components/organisms/LeavePolicyWizard/LeavePolicyWizard.tsx
```diff
@@ -153,7 +166,41 @@ const LeavePolicyWizard: FC<Props> = ({ policyType }) => {
...
+        name: checkedName,
+        leaveTypeId: checkedLeaveType
+      });
+
+      const currentValues = formikRef.current?.values;
+      const isStale =
+        currentValues?.policyName.trim() !== checkedName ||
+        currentValues?.leaveType !== checkedLeaveType;
+
+      if (isStale) {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The stale branch produces a silent dead-end. The Next/Back buttons are disabled while `isCheckingPolicyName` is true, but the policy-name input and leave-type dropdown are not, so a user can edit the name mid-flight. When that happens this returns `false`, `advanceWizard` returns early, and nothing at all is rendered — no advance, no inline error, no toast. It is worse than it looks because the *next* Next click starts with `await formik.validateForm()`, which overwrites `errors` and wipes any duplicate error that was previously shown, so the user just sees a button that does nothing. Suggest either re-running the check against the current values before giving up, or falling through to `return true` (letting the create call be the authority, matching the `catch` branch's stated rationale) instead of returning `false` with no feedback.

frontend/src/community/leave/components/organisms/LeavePolicyWizard/LeavePolicyWizard.tsx
```diff
@@ -232,7 +300,7 @@ const LeavePolicyWizard: FC<Props> = ({ policyType }) => {
           icon={!isFirstStep ? <ArrowLeftIcon /> : undefined}
           iconPosition="start"
           onClick={handleBack}
-          disabled={isPending}
+          disabled={isPending || isCheckingPolicyName}
```
ThinuwanW, line 292: 🤖 **Claude** · 🟡 **Suggestion**

Disabling the tertiary button on `isCheckingPolicyName` also disables Cancel. On BASIC_INFO this button is "Cancel" (`isFirstStep` is true), so while the optional availability check is in flight the user cannot back out of the wizard. `authFetch` has no `timeout` configured, so a hung request leaves both buttons permanently disabled with no escape. The name check is explicitly best-effort (the catch block falls through to the create call), so it should not block cancelling — keep this button as `disabled={isPending}` and leave the `isCheckingPolicyName` guard only on the primary Next button at line 313.

frontend/src/community/common/assets/languages/english/leaveModule.json
```diff
@@ -476,7 +476,8 @@
         "accrualCapStepInvalid": "Maximum balance must be in multiples of 0.5.",
         "maxCarryOverDaysRequired": "Please enter the maximum number of carry over days.",
         "maxCarryOverDaysInvalid": "Maximum carry over days must be greater than 0 and at most 365.",
-        "maxCarryOverDaysStepInvalid": "Maximum carry over days must be in multiples of 0.5."
+        "maxCarryOverDaysStepInvalid": "Maximum carry over days must be in multiples of 0.5.",
+        "policyNameDuplicate": "A policy with this name already exists for the selected leave type. Please use a different name."
```
ThinuwanW, line 480: 🤖 **Claude** · 🟡 **Suggestion**

`policyNameDuplicate` is a verbatim duplicate of `duplicateToastDescription` at line 487 in the same `leavePolicies.createPolicy` namespace ("A policy with this name already exists for the selected leave type. Please use a different name."). Two copies of the same sentence will drift the moment someone edits one. Either keep `translateText(["duplicateToastDescription"])` in `showDuplicatePolicyNameError` and drop the new key, or keep the new key and remove the now-unreachable `createPolicy.duplicateToastTitle`/`duplicateToastDescription` pair — `handleError` short-circuits on `isDuplicatePolicyNameError` before ever reaching `getLeavePolicyErrorToastKeys`, so the create-scoped duplicate toast can never fire (the other scopes at lines 329/353 used by EditLeavePolicyModal and LeavePoliciesTable still need theirs).

frontend/src/community/leave/api/LeavePolicyApi.ts
```diff
@@ -65,6 +68,29 @@ export const useGetLeavePoliciesInfinite = ({
...
+  params: CheckLeavePolicyNameAvailabilityParams
+): Promise<LeavePolicyNameAvailabilityResult> => {
+  const response = await authFetch.get<LeavePolicyNameAvailabilityResponse>(
+    leavePolicyEndPoints.CHECK_LEAVE_POLICY_NAME_AVAILABILITY,
+    { params }
+  );
+  return response.data.results[0];
+};
+
+export const useCheckLeavePolicyNameAvailability = (): UseMutationResult<
```
ThinuwanW, line 81: 🤖 **Claude** · 🟣 **Nit**

Using `useMutation` for a GET diverges from the existing name-availability precedent in this codebase: `useCheckWorkLocationNameExists` (frontend/src/community/configurations/api/WorkLocationApi.ts:153) uses `useQuery` with a `workLocationQueryKeys.CHECK_WORK_LOCATION_NAME_EXISTS(name)` key factory. The mutation form is defensible here because the check is imperative (fired on a button click, not on render) and caching an availability answer would be actively wrong, but it does mean there is no entry in `leavePolicyQueryKeys` and no in-flight dedup. If you want the query-key convention without stale answers, `queryClient.fetchQuery({ queryKey: leavePolicyQueryKeys.NAME_AVAILABILITY(name, leaveTypeId), queryFn, staleTime: 0, gcTime: 0 })` gives the same imperative semantics. Also note the mutation has no `onError` — intentional (the caller's `catch` swallows it so the backend stays the authority), but worth a one-line comment at the hook so a future reader does not "fix" it by adding a toast.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/LeavePolicyServiceImpl.java
```diff
@@ -181,6 +182,23 @@ public ResponseEntityDto getAllLeavePolicies(LeavePolicyFilterDto leavePolicyFil
...
+			throw new ModuleException(LeaveMessageConstant.LEAVE_ERROR_LEAVE_POLICY_LEAVE_TYPE_REQUIRED);
+		}
+
+		LeavePolicyValidationUtil.validateName(name);
+
+		boolean isAvailable = !leavePolicyDao.existsByNameIgnoreCaseAndLeaveType_Id(name, leaveTypeId);
```
ThinuwanW, line 196: 🤖 **Claude** · 🟣 **Nit**

There is no DB-level uniqueness backing this check. `lv_leave_policy` (community/db/changelog/ddl-script/leave-ddl-script-v1-create-leave-policy-tables.sql) has only a PK and an FK — no unique index on `(name, leave_type_id)`. So both this endpoint and `addLeavePolicy` do check-then-insert with no constraint to lose the race against, and two concurrent creates can persist duplicate names. Pre-existing, but this PR builds a whole UX on the uniqueness invariant, so it is worth closing: a Liquibase changeset narrowing `name` from `text` to `varchar(100)` (matching `LeavePolicyConstant.MAX_NAME_LENGTH`) plus a unique index on `(name, leave_type_id)` would make the invariant real; a `DataIntegrityViolationException` catch mapping to `LEAVE_ERROR_LEAVE_POLICY_ALREADY_EXISTS` in `addLeavePolicy` would keep the client-facing error identical.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/LeavePolicyServiceImpl.java
```diff
@@ -181,6 +182,23 @@ public ResponseEntityDto getAllLeavePolicies(LeavePolicyFilterDto leavePolicyFil
 		return new ResponseEntityDto(false, pageDto);
 	}
 
+	@Override
+	@Transactional(readOnly = true)
+	public ResponseEntityDto checkLeavePolicyNameAvailability(String name, Long leaveTypeId) {
+		log.info("checkLeavePolicyNameAvailability: execution started");
+
+		if (leaveTypeId == null) {
+			throw new ModuleException(LeaveMessageConstant.LEAVE_ERROR_LEAVE_POLICY_LEAVE_TYPE_REQUIRED);
```
ThinuwanW, line 191: 🤖 **Claude** · 🟣 **Nit**

The null-leaveTypeId guard is inlined in the service while every other leave-policy validation lives in `LeavePolicyValidationUtil` — it duplicates the first branch of `LeavePolicyValidationUtil.validateRequiredFields` (same `LEAVE_ERROR_LEAVE_POLICY_LEAVE_TYPE_REQUIRED` constant), which cannot be reused here because it takes a `LeavePolicyRequestDto`. Suggest extracting `public static void validateLeaveTypeId(Long leaveTypeId)` into the util and having both `validateRequiredFields` and this method call it, so the two paths cannot drift.

backend/src/main/java/com/skapp/community/leaveplanner/controller/v1/LeavePolicyController.java
```diff
@@ -37,6 +38,17 @@ public ResponseEntity<ResponseEntityDto> getAllLeavePolicies(LeavePolicyFilterDt
 		return new ResponseEntity<>(response, HttpStatus.OK);
 	}
 
+	@Operation(summary = "Check leave policy name availability",
+			description = "Returns whether the given policy name is still available for the given leave type. "
+					+ "Names are compared case-insensitively across both active and inactive policies")
+	@GetMapping("/name-availability")
+	@PreAuthorize("hasAnyRole('ROLE_LEAVE_ADMIN')")
+	public ResponseEntity<ResponseEntityDto> checkLeavePolicyNameAvailability(
+			@RequestParam(required = false) String name, @RequestParam(required = false) Long leaveTypeId) {
```
ThinuwanW, line 47: 🤖 **Claude** · 🟣 **Nit**

`@RequestParam(required = false) Long leaveTypeId` returns a 500, not a 400, on a non-numeric value. `MethodArgumentTypeMismatchException` extends `TypeMismatchException` (not `BindException`), and `GlobalExceptionHandler` has no handler for it, so `?leaveTypeId=abc` falls through to `@ExceptionHandler(Exception.class)` → `INTERNAL_SERVER_ERROR`. The `required = false` + explicit `ModuleException` choice is right (making the params required would throw `MissingServletRequestParameterException`, a `ServletException`, which the handler maps to 500 — worse), so the remaining gap is just the type coercion. This is a codebase-wide pattern rather than something this PR invented, so the durable fix is adding `@ExceptionHandler(MethodArgumentTypeMismatchException.class)` → 400 to `GlobalExceptionHandler`; a local alternative is taking the param as `String` and parsing it inside a try/catch that throws `LEAVE_ERROR_LEAVE_POLICY_LEAVE_TYPE_REQUIRED`.

backend/src/test/java/com/skapp/community/leaveplanner/controller/v1/LeavePolicyControllerIntegrationTest.java
```diff
@@ -344,6 +354,93 @@ void deactivateLeavePolicy_UnknownPolicy_ReturnsNotFound() throws Exception {
 
 	}
 
+	@Nested
```
ThinuwanW, line 357: 🤖 **Claude** · 🟣 **Nit**

Good coverage overall (case-insensitivity, inactive policies, cross-leave-type isolation, blank/missing params, 403, 401). Three gaps worth closing: (1) no test for a name longer than `LeavePolicyConstant.MAX_NAME_LENGTH` (101 chars) asserting 400 — `validateName` has that branch and nothing exercises it here; (2) no test for a non-existent `leaveTypeId`, which currently returns 200/available and would pin down whichever behaviour you settle on for the leave-type validation asymmetry noted above; (3) no test that a name with surrounding whitespace behaves identically on the check and the create path — the frontend trims before both calls, but neither backend method trims, so a test would lock that contract in.

frontend/src/community/leave/components/organisms/LeavePolicyWizard/LeavePolicyWizard.tsx
```diff
@@ -153,7 +166,41 @@ const LeavePolicyWizard: FC<Props> = ({ policyType }) => {
     }
   };
 
-  const handleNext = async (): Promise<void> => {
+  const isPolicyNameAvailable = async (): Promise<boolean> => {
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

The backend side of this feature got nine integration tests; the frontend side got none. The new logic here is not trivial — the stale-values comparison, the swallow-and-continue catch, the `isAdvancingRef` re-entrancy guard, and the BASIC_INFO-only branch each have a distinct failure mode, and three of the four are invisible in the UI when they misbehave. There is no existing test file in this directory so this is not a regression, but a `LeavePolicyWizard.test.tsx` with a mocked `useCheckLeavePolicyNameAvailability` covering available / unavailable / rejected-promise / edited-mid-flight would be cheap insurance. Per the guidelines, wrap in `MockTheme` and prefer `screen.getByRole` over `getByTestId`.

frontend/src/community/leave/components/organisms/LeavePolicyWizard/LeavePolicyWizard.tsx
```diff
@@ -153,7 +166,41 @@ const LeavePolicyWizard: FC<Props> = ({ policyType }) => {
...
+        // flagging - or advancing - the wrong name.
+        return false;
+      }
+
+      if (!isAvailable) {
+        showDuplicatePolicyNameError();
+      }
+
+      return isAvailable;
+    } catch {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The bare `catch {}` swallows every failure mode of the availability check with zero user-visible feedback, and its justifying comment is only half true. (1) The comment says "let the create request run so the backend stays the authority" — that is accurate for FLEXIBLE policies (where BASIC_INFO is also the last step), but for ACCRUAL policies returning `true` here does NOT run the create request, it silently advances the wizard to ENTITLEMENT_SETUP. Please reword so the next reader isn't misled. (2) The catch is not limited to network/5xx faults. A 403 (e.g. a PEOPLE_ADMIN who can list policies but is not LEAVE_ADMIN, since this endpoint is `ROLE_LEAVE_ADMIN` only), a 400 from `validateName`, or an expired session all resolve to "name is available" with nothing shown to the user. The review guidelines require every mutation to surface user feedback on failure, and `useCheckLeavePolicyNameAvailability` has no `onError` either. (3) `checkLeavePolicyNameAvailability` does `return response.data.results[0]` with no guard, so an empty `results` array makes `const { isAvailable } = await ...` throw a TypeError that this same catch converts into "available" — the whole feature would silently no-op with no signal in dev or prod. Suggested fix: type the catch parameter as `AxiosError`, treat only network errors / 5xx as "proceed", and route 4xx through the existing `getLeavePolicyErrorToastKeys` + `setToastMessage` path (or add an `onError` to the hook); and make the API function fail loudly if `results[0]` is missing.

frontend/src/community/leave/components/organisms/LeavePolicyWizard/LeavePolicyWizard.tsx
```diff
@@ -242,7 +310,7 @@ const LeavePolicyWizard: FC<Props> = ({ policyType }) => {
           icon={!isLastStep ? <ArrowRightIcon /> : undefined}
           iconPosition="end"
           onClick={handleNext}
-          disabled={isPending}
+          disabled={isPending || isCheckingPolicyName}
```
ThinuwanW, line 302: 🤖 **Claude** · 🟡 **Suggestion**

No loading affordance while the name check is in flight. `isCheckingPolicyName` only disables both footer buttons; the primary button keeps its label ("Next", or "Create policy" for FLEXIBLE since `isPending` is still false), so from the user's point of view the wizard just freezes for the duration of the round-trip with no explanation. `ButtonV2` supports an `isLoading` prop and it is already used exactly this way elsewhere in the codebase (`EditLeavePolicyModal.tsx:166` -> `isLoading: isPending`, `OrganizationSettings.tsx:228`). Add `isLoading={isPending || isCheckingPolicyName}` to this primary button so the spinner state matches the disabled state.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/LeavePolicyServiceImpl.java
```diff
@@ -181,6 +182,23 @@ public ResponseEntityDto getAllLeavePolicies(LeavePolicyFilterDto leavePolicyFil
 		return new ResponseEntityDto(false, pageDto);
 	}
 
+	@Override
+	@Transactional(readOnly = true)
+	public ResponseEntityDto checkLeavePolicyNameAvailability(String name, Long leaveTypeId) {
```
ThinuwanW, line 187: 🤖 **Claude** · 🟡 **Suggestion**

The endpoint is shaped so it can only ever serve the create flow — there is no way to exclude the policy being edited. `updateLeavePolicy` (line 110) deliberately uses `existsByNameIgnoreCaseAndLeaveType_IdAndIdNot(name, leaveTypeId, id)`, but `checkLeavePolicyNameAvailability` hard-codes the `...AndLeaveType_Id` variant, so if `EditLeavePolicyModal` called it the policy's own current name would come back as unavailable. As a result the rename flow (frontend/src/community/leave/components/molecules/EditLeavePolicyModal/EditLeavePolicyModal.tsx) is left with the strictly worse UX this PR set out to fix: duplicates only surface as a post-submit toast via `getLeavePolicyErrorToastKeys`, never as an inline field error. Suggest adding an optional `excludeLeavePolicyId` (or `leavePolicyId`) parameter that delegates to `existsByNameIgnoreCaseAndLeaveType_IdAndIdNot` when present, and wiring the edit modal to it so both name-entry surfaces behave the same.

backend/src/main/java/com/skapp/community/leaveplanner/payload/response/LeavePolicyNameAvailabilityResponseDto.java
```diff
@@ -0,0 +1,16 @@
...
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+
+@Getter
+@Setter
+@NoArgsConstructor
+@AllArgsConstructor
+public class LeavePolicyNameAvailabilityResponseDto {
+
```
ThinuwanW, line 13: 🤖 **Claude** · 🟡 **Suggestion**

This is the fourth near-identical single-`Boolean` name-check response DTO in the codebase, and the only one with inverted polarity. Existing precedents: `community/common/payload/response/WorkLocationNameAvailabilityResponseDto` (field `isExists`), `crmplanner/payload/response/CrmExistsResponseDto` (field `isExists`), and `enterprise/common/payload/response/FolderNameExistsResponseDto` (field `isExists`). The endpoint naming follows the same convention (`GET /name-exists` + `checkWorkLocationNameExists`, `checkFolderNameExists`, `checkDealNameExists`). Introducing `isAvailable` (= `!exists`) plus `/name-availability` means every future reader and every frontend consumer has to remember which of the two polarities a given name check uses, and the guidelines call out "2+ new DTOs with identical fields -> consolidate". Prefer reusing/aligning with the `isExists` shape and the `check*NameExists` naming, or promote a single shared `NameExistsResponseDto` into `common/payload/response` and have all four use it.

backend/src/main/java/com/skapp/community/leaveplanner/controller/v1/LeavePolicyController.java
ThinuwanW, line 47: 🤖 **Claude** · 🟣 **Nit**

`@RequestParam(required = false) String name` misdocuments the contract. `name` is in fact mandatory — `LeavePolicyValidationUtil.validateName` rejects null immediately — but marking it optional makes the generated OpenAPI/Swagger spec advertise it as optional, so API consumers reading the docs will think a bare `GET /v1/leave/policies/name-availability` is valid. It also moves a check Spring already performs (`MissingServletRequestParameterException` -> 400) into the service. The direct precedent in this repo does the opposite: `WorkLocationController.checkWorkLocationNameExists` declares `@RequestParam String name` and keeps only the blank/length checks in the service. Suggest dropping `required = false` from `name` (`validateName` still covers blank and max-length).

frontend/src/community/leave/components/organisms/LeavePolicyWizard/LeavePolicyWizard.tsx
```diff
@@ -15,7 +15,10 @@ import ROUTES from "~community/common/constants/routes";
 import { ToastType } from "~community/common/enums/ComponentEnums";
 import { useTranslator } from "~community/common/hooks/useTranslator";
 import { useToast } from "~community/common/providers/ToastProvider";
-import { useAddLeavePolicy } from "~community/leave/api/LeavePolicyApi";
+import {
```
ThinuwanW, line 18: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

This component is now 331 lines and mixes step state, Formik wiring, two React Query mutations, async availability orchestration (`isPolicyNameAvailable`), a re-entrancy guard (`isAdvancingRef`), error-to-field mapping, and all of the JSX. The guidelines flag components over ~200 lines and "components doing data fetching AND rendering". The new logic added here is the natural seam: extract a `useLeavePolicyWizard` hook (module `hooks/` or colocated) owning `activeStep`, the formik instance, `advanceWizard`/`handleNext`/`handleBack`, `isPolicyNameAvailable`, and the `isPending`/`isCheckingPolicyName` flags, and leave this file as the presentational shell. That also makes the branch-heavy `advanceWizard`/`isPolicyNameAvailable` logic unit-testable without mounting MUI.

backend/src/test/java/com/skapp/community/leaveplanner/controller/v1/LeavePolicyControllerIntegrationTest.java
```diff
@@ -344,6 +354,93 @@ void deactivateLeavePolicy_UnknownPolicy_ReturnsNotFound() throws Exception {
...
+		@DisplayName("Returns bad request when the leave type is missing")
+		void checkNameAvailability_MissingLeaveTypeId_ReturnsBadRequest() throws Exception {
+			mvc.perform(get(ENDPOINT + "/name-availability").param("name", "Brand New Policy")
+				.accept(MediaType.APPLICATION_JSON)
+				.with(SecurityTestUtils.bearerToken(leaveAdminToken())))
+				.andDo(print())
+				.andExpect(status().isBadRequest());
+		}
+
+		@Test
```
ThinuwanW, line 426: 🤖 **Claude** · 🟣 **Nit**

The two new auth tests (`checkNameAvailability_LeaveEmployee_ReturnsForbidden` and `checkNameAvailability_NoAuth_ReturnsUnauthorized`) are placed inside `LeavePolicyNameAvailabilityTests`, but this file already has a dedicated `@Nested class RoleBasedAccessTests` (line 203) that collects every 401/403 case for create, list, update, activate and deactivate — including a `leavePolicies_PeopleAdmin_CanReadButNotCreate` test that pairs allowed and denied roles. Moving these two there keeps the auth matrix in one place (and makes it obvious that, unlike `GET /policies`, this endpoint excludes `ROLE_PEOPLE_ADMIN`, which is worth an explicit assertion given both are read-only operations on the same resource).

frontend/src/community/leave/api/LeavePolicyApi.ts
hasalarootcode: The check is imperative and fires on a click, and caching an availability answer would be actively wrong, so the mutation form is deliberate here. `fetchQuery` with `staleTime: 0` gets to the same behaviour with more moving parts.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/LeavePolicyServiceImpl.java
hasalarootcode: We don't add DB unique constraints in this codebase; uniqueness is enforced by the service-level `existsBy...` check. The check-then-insert race is accepted as part of that.

backend/src/main/java/com/skapp/community/leaveplanner/controller/v1/LeavePolicyController.java
hasalarootcode: Not needed

hasalarootcode: That is an existing pattern

frontend/src/community/leave/components/organisms/LeavePolicyWizard/LeavePolicyWizard.tsx
hasalarootcode: There's no test setup in this directory yet, so this is a first `LeavePolicyWizard.test.tsx` plus mocks for both mutations. Want it in this PR?

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/LeavePolicyServiceImpl.java
hasalarootcode: Will do it seperately if needed

backend/src/main/java/com/skapp/community/leaveplanner/payload/response/LeavePolicyNameAvailabilityResponseDto.java
hasalarootcode: Exisiting pattern

frontend/src/community/leave/components/organisms/LeavePolicyWizard/LeavePolicyWizard.tsx
hasalarootcode: Extracting `useLeavePolicyWizard` moves the formik instance, both mutations and all the step logic out of this file. Here, or a follow-up?

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/LeavePolicyServiceImpl.java
hasalarootcode: Not needed

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/LeavePolicyServiceImpl.java
hasalarootcode: Not needed

backend/src/test/java/com/skapp/community/leaveplanner/controller/v1/LeavePolicyControllerIntegrationTest.java
hasalarootcode: Not needed

backend/src/test/java/com/skapp/community/leaveplanner/controller/v1/LeavePolicyControllerIntegrationTest.java
hasalarootcode: Not needed

frontend/src/community/common/assets/languages/english/leaveModule.json
hasalarootcode: Not needed

frontend/src/community/leave/components/organisms/LeavePolicyWizard/LeavePolicyWizard.tsx
hasalarootcode: Not valid

frontend/src/community/leave/components/organisms/LeavePolicyWizard/LeavePolicyWizard.tsx
hasalarootcode: Not valid

frontend/src/community/leave/components/organisms/LeavePolicyWizard/LeavePolicyWizard.tsx
hasalarootcode: Not needed

frontend/src/community/leave/components/organisms/LeavePolicyWizard/LeavePolicyWizard.tsx
hasalarootcode: Not valid

frontend/src/community/leave/components/organisms/LeavePolicyWizard/LeavePolicyWizard.tsx
hasalarootcode: not needed

shakila-kamalasena (approved):

thusala (approved):

akilarootcode (approved):

PR #1816, opened 2026-08-20
fix: resolve accessibility (a11y) issues on Sign Configurations page

SanithuGunasena07:
Fixes accessibility violations flagged by axe DevTools on the Sign Configurations page. The collapsible sidebar menu button lost its accessible name when collapsed to icon-only mode, and the Date Format dropdown along with the expiration/reminder day inputs had no accessible name at all. Added aria-label/ariaLabel to each so screen readers can properly announce them.

### DIFF
--- a/frontend/src/community/common/components/organisms/Drawer/Drawer.tsx
+++ b/frontend/src/community/common/components/organisms/Drawer/Drawer.tsx
@@ -299,6 +299,7 @@ const Drawer = (): JSX.Element => {
                     }}
                     aria-expanded={isExpanded}
                     aria-controls={`sub-list-${routeId}`}
+                    aria-label={route?.name}
                   >
                     <ListItemIcon sx={classes.listItemIcon}>
                       {route?.icon && (
--- a/frontend/src/community/common/components/organisms/Drawer/Drawer.tsx
+++ b/frontend/src/community/common/components/organisms/Drawer/Drawer.tsx
@@ -299,7 +299,7 @@ const Drawer = (): JSX.Element => {
                     }}
                     aria-expanded={isExpanded}
                     aria-controls={`sub-list-${routeId}`}
-                    aria-label={route?.name}
+                    aria-label={!isDrawerExpanded ? route?.name : undefined}
                   >
                     <ListItemIcon sx={classes.listItemIcon}>
                       {route?.icon && (
--- a/frontend/src/community/common/components/organisms/Drawer/Drawer.tsx
+++ b/frontend/src/community/common/components/organisms/Drawer/Drawer.tsx
@@ -299,7 +299,7 @@ const Drawer = (): JSX.Element => {
                     }}
                     aria-expanded={isExpanded}
                     aria-controls={`sub-list-${routeId}`}
-                    aria-label={!isDrawerExpanded ? route?.name : undefined}
+                    aria-label={(!isDrawerExpanded && route?.name) || undefined}
                   >
                     <ListItemIcon sx={classes.listItemIcon}>
                       {route?.icon && (


ThinuwanW (commented):
## 🤖 Claude Code Review

This one-line PR adds `aria-label={route?.name}` to the main drawer `ListItemButton`, which correctly fixes a real gap: when the drawer is collapsed, `classes.listItemContent` sets `display: none`, removing `ListItemText` from the accessibility tree and leaving the icon-only button with no accessible name. However, `aria-label` overrides the content-derived name in *all* states, so in the expanded drawer the `NotificationBadge` count is no longer part of the button's accessible name — a regression for the `Inbox` route. I also note the label bypasses `translateAria` (used for every other aria string in this file) and that the adjacent `aria-expanded`/`aria-controls` on this same element are still applied to non-expandable routes.

Found **15** new issue(s): 🔴 2 important, 🟡 7 suggestion(s), 🟣 6 nit(s)

frontend/src/community/common/components/organisms/Drawer/Drawer.tsx
```diff
@@ -299,6 +299,7 @@ const Drawer = (): JSX.Element => {
                     }}
                     aria-expanded={isExpanded}
                     aria-controls={`sub-list-${routeId}`}
+                    aria-label={route?.name}
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`aria-label` overrides the accessible name computed from the element's contents, so this unconditionally discards text that was previously announced. Specifically, `NotificationBadge` (rendered at line 357 when `route?.notificationCount && !hasSubTree`) renders its count as plain text inside this `ListItemButton` with no `aria-hidden`. Before this change, the expanded drawer announced e.g. "Inbox 3"; now it announces only "Inbox", so the pending-signature count from `notificationSignCount` (see `getDrawerRoutes.ts` lines 310-318, the only main route that receives `notificationCount`) is silent to screen reader users. The label is only actually needed in the collapsed state, since `listItemContent` is `display: none` when `!isDrawerExpanded` (styles.ts line 122). Fix by scoping it: `aria-label={!isDrawerExpanded ? route?.name : undefined}` — when expanded the natural content-derived name (name + count) is restored. Alternatively keep it unconditional but compose the count into it, e.g. append a translated "{{count}} pending" suffix when `route?.notificationCount` is set.

ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

This aria-label is not internationalised. `route.name` originates from the hardcoded English literals in `~community/common/utils/data/routes.ts` ("Dashboard", "Timesheet", …) plus the inline literals in `getDrawerRoutes.ts` ("My Leave Requests", "Inbox"), and it never passes through i18next. Every other aria string in this file goes through `translateAria` (lines 230, 494-495), and the review guidelines require accessibility labels to be translated. Because the visible `ListItemText` uses the same untranslated string there is no label/name mismatch today, so this is not a correctness bug — but it does bake English into the accessibility layer. Preferably move the route names into the `commonComponents.drawer` namespace (with a `translationKey` field on each route entry) and use the translated value for both `primary` and `aria-label`.

frontend/src/community/common/components/organisms/Drawer/Drawer.tsx
```diff
@@ -299,6 +299,7 @@ const Drawer = (): JSX.Element => {
                     }}
                     aria-expanded={isExpanded}
                     aria-controls={`sub-list-${routeId}`}
```
ThinuwanW, line 301: 🤖 **Claude** · 🟡 **Suggestion**

Since this PR is fixing ARIA on this exact element, the two adjacent attributes are worth correcting in the same change: `aria-expanded` and `aria-controls` are applied to every main route, including those with `hasSubTree === false` (Dashboard id 0, Configurations id 8, and the collapsed single-route variants returned by `getDrawerRoutes`). The `Collapse`/`<List id={`sub-list-${routeId}`}>` is only rendered under `hasSubTree && (...)` at line 389, so for non-expandable routes `aria-controls` references an ID that never exists in the DOM (axe `aria-valid-attr-value`), and `aria-expanded="false"` makes a plain navigation link announce as a collapsed disclosure widget. Gate both: `aria-expanded={hasSubTree ? isExpanded : undefined}` and `aria-controls={hasSubTree ? `sub-list-${routeId}` : undefined}`.

frontend/src/community/common/components/organisms/Drawer/Drawer.tsx
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

No test accompanies this accessibility change, and there is no `Drawer.test.tsx` anywhere in the repo (the only related test is `src/fallback/common/utils/getEnterpriseDrawerRoutes.test.ts`), so nothing guards the behaviour from regressing. Consider adding a component test wrapped in `MockTheme` that renders the drawer collapsed and asserts `screen.getByRole("button", { name: "Dashboard" })` resolves, plus an expanded-state assertion that the accessible name still includes the notification count once the issue above is addressed. Role-based queries would also lock in the `aria-expanded`/`aria-controls` semantics.

ThinuwanW: 🤖 **Claude** · 🔴 **Important**

These nav buttons remain focusable and exposed to AT while the drawer is visually closed, and this change now gives them full accessible names. StyledDrawer.tsx sets the paper width to `0%` (below 600px) / `0rem` (below 1024px) when closed, the variant is `permanent`, and the inner Stack force-overrides `visibility: "visible"`. Width 0 + `overflowX: hidden` only clips — it does not remove elements from the tab order or the accessibility tree. Worse, `useDrawer` forces `isDrawerExpanded` to false on any viewport under 1024px, and `#side-bar` (this Stack's id) is a declared skip-to-content target in `signedInUserSkipToContentList`, so screen-reader users are actively directed into a collapsed drawer whose ~10 route buttons can now be read out and activated while invisible (WCAG 2.4.3 / 2.4.7). Before this change the route text was inside a `display: none` box so the buttons had no accessible name and were largely inert; the aria-label makes the hidden menu fully legible. Fix: when `isBelow1024 && !isDrawerExpanded`, apply `inert` (or `aria-hidden="true"` plus `tabIndex={-1}` on every ListItemButton) to the drawer content, or don't render the list at all in that state — and exclude `#side-bar` from the skip list while it is closed.

ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Missing the natural companion to this fix: `aria-current`. `isActiveRoute` is already computed on line 269-270 but is consumed only by `classes.listItemButton(...)` for a background colour, so the current page is signalled to sighted users by colour alone (WCAG 1.4.1) and not at all to AT — and now that the accessible name is a fixed label, there is no textual cue either. Add `aria-current={isActiveRoute ? "page" : undefined}` on this button (and equivalently on the sub-route buttons, which compute the same `router.pathname.includes(...)` check inline at line 436). Note `aria-current` appears nowhere in `frontend/src/`, so this is a gap across the nav, not just here.

frontend/src/community/common/components/organisms/Drawer/Drawer.tsx
```diff
@@ -299,6 +299,7 @@ const Drawer = (): JSX.Element => {
...
+                    aria-label={route?.name}
                   >
                     <ListItemIcon sx={classes.listItemIcon}>
```
ThinuwanW, line 304: 🤖 **Claude** · 🟡 **Suggestion**

Pending-item counts are conveyed visually only, and this is the state the new label was meant to serve. `NotificationDot` renders an empty coloured Box with no text and no `aria-label`/`role="status"`, and it is shown precisely when the drawer is collapsed or the group is closed (see the `show` conditions on lines 315-338) — i.e. the exact case where the aria-label is the sole accessible name. `NotificationBadge` does render its count as text, but only for leaf routes (`!hasSubTree`), and that text is now suppressed by the aria-label. Result: an AT user hears "Leave" whether or not 5 requests are pending. Fix: interpolate the count into the localized label (the aria JSON at `assets/languages/english/aria/common.json` already uses interpolation, e.g. the `modal` entries), or give `NotificationDot`/`NotificationBadge` visually-hidden text such as "3 pending requests".

frontend/src/community/common/components/organisms/Drawer/Drawer.tsx
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

The region referenced by `aria-controls` is not actually hidden when `aria-expanded` is false, which undermines the expand/collapse semantics being fixed here. Two causes, both around lines 389-438: (1) `collapsedSize="0rem"` — MUI v6 Collapse only applies `visibility: hidden` when `collapsedSize` is exactly the string `'0px'` (its default), so passing `"0rem"` keeps the collapsed sub-list in the accessibility tree, merely clipped by `height: 0; overflow: hidden`. Drop the prop or pass `"0px"`. (2) The sub-route button's `tabIndex={isExpanded ? 0 : -1}` ignores `isDrawerExpanded`, while the Collapse renders on `isExpanded && isDrawerExpanded`. `useDrawer`'s effect sets `isDrawerExpanded` false when the viewport drops below 1024px without clearing `expandedDrawerListItem` (only `handleDrawer` clears it), so a resize leaves a hidden sub-list whose buttons are still tab-focusable. Use `tabIndex={isExpanded && isDrawerExpanded ? 0 : -1}` (or gate on the same expression the Collapse uses).

frontend/src/community/common/components/organisms/Drawer/Drawer.tsx
```diff
@@ -299,6 +299,7 @@ const Drawer = (): JSX.Element => {
                     }}
```
ThinuwanW, line 299: 🤖 **Claude** · 🟡 **Suggestion**

This `onKeyDown` handler duplicates behaviour MUI already provides, and causes the click handler to run twice per Enter press. `ListItemButton` defaults to `component="div"`, so `ButtonBase.handleKeyDown` invokes the passed `onKeyDown` and then — for a non-native button on Enter — also calls `onClick`, without checking `defaultPrevented`. So `handleListItemButtonClick` fires twice: two `router.push` calls for leaf routes, two `setExpandedDrawerListItem` calls for groups. It is benign today only because both setters use the value captured in the render closure rather than a functional updater, so the second call writes the same value; converting either to `set(prev => ...)` would turn this into a toggle-then-untoggle bug where keyboard users cannot open a submenu. Also note Space is handled by ButtonBase's keyup path only, so the custom handler makes Enter and Space follow different code paths. Recommend removing the `onKeyDown` from this button and from the sub-route `ListItem` (line 419), which has the same double-invocation via keydown-bubbling plus ButtonBase's synthetic click.

frontend/src/community/common/components/organisms/Drawer/Drawer.tsx
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

`role="document"` on the Stack (line 240) is invalid here and matters because that Stack (`id="side-bar"`) is a registered skip-to-content destination. The `document` role is only meaningful inside an `application` role context; nesting it inside the `component="nav"` Drawer creates a landmark containing a document role, which makes some screen readers switch reading modes and can suppress the `nav` landmark. Since this PR is repairing ARIA on this component, remove the `role` (a plain container needs none) or use a real structural role. Same review pass: the `role="list"` / `role="listitem"` on MUI `List`/`ListItem` (lines 263, 276, 398, 410) are redundant — those already render `ul`/`li` — and the redundancy is harmless but should not be extended.

ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

Focus handling on navigation (lines 192-196) is a known a11y anti-pattern and is worth fixing in an a11y PR touching this element: it sets `tabindex="-1"` on `document.body` (and never removes it) and calls `document.body.focus()`. Focusing `body` leaves several screen readers silent about the route change and resets the tab position to the very top of the document. The codebase already has the machinery for a better target — `SkipToContentPopup` focuses `#content-with-drawer-main-content` and cleans its temporary tabindex up on blur. Move focus to that main-content container after `router.push` (reusing the same set-tabindex/focus/remove-on-blur pattern) and announce the new page via a live region.

ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

Pre-existing logic bug in the S3 logo effect (lines 181-185): the condition uses `organizationLogo || !s3FileUrls[organizationLogo]` where `&&` is clearly intended. As written, whenever `organizationLogo` is set the effect re-requests a signed URL even though it is already cached (the short-circuit means the cache check is never evaluated), and when it is undefined the second operand is true so `downloadS3File` is still invoked with `filePath: undefined` (harmless only because the hook guards `if (!filePath) return`). Also, the dependency array lists only `organizationLogo` while the body reads `s3FileUrls` and `downloadS3File`. Change to `if (organizationLogo && !s3FileUrls[organizationLogo])` and add the missing deps.

ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

Line 171 mutates the theme object during render (`theme.palette = updatedTheme.palette`). `useTheme()` returns the shared provider theme instance, so this is a render-phase side effect that reassigns the palette for every other component sharing that theme, is not undone on unmount, and is invisible to React's reconciliation (no re-render is triggered for consumers already rendered this pass). It also depends on `styles({ theme })` being evaluated lazily to pick up the new palette. Preferred fix: wrap the drawer subtree in a `ThemeProvider theme={updatedTheme}` or pass `updatedTheme` into `styles()` instead of mutating.

ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

Type-safety items from the checklist in this file: the store selector on line 98 is typed `(state: CommonStoreTypes | any)`, which collapses to `any` and silently disables checking for all five destructured members (`setIsDrawerExpanded`, `setOrgData`, ...) — use `CommonStoreTypes` alone and fix whatever it surfaces. Related: `useDrawer` needs a non-null assertion (`setExpandedDrawerListItem!`) for the same store, which is a symptom of the same weak typing. Line 402-405 also uses an `as` assertion to graft `notificationCount` onto the sub-route type; that field should be declared on the route type returned by `getDrawerRoutes` instead of asserted at the call site.

ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

Duplication and an inconsistent fallback in the colour lookup: `getSelectedDrawerItemColor(theme, router.pathname, ...)` is called four times per rendered route (lines 309, 346, 370, 377) with the same inputs, and the URL fallback differs — `?? null` at 309/346 versus `?? ""` at 373/380. Those are not equivalent: the helper returns black for `null`, but for `""` it evaluates `pathname.includes("")`, which is always true, so it returns the selected colour. No route currently ships a null URL so this is latent, but the two branches should not disagree. Hoist one `const itemColor = getSelectedDrawerItemColor(theme, router.pathname, route?.url ?? null)` per route and reuse it. Separately, the notification-dot conditions hardcode route ids `"1"`, `"2"`, `"5"` (lines 317, 325, 333) and `getDrawerRoutes` hardcodes sub-route ids `"2B"`, `"1B"`, `"5A"`, `"2C"`, `"2E"` — these should be named constants or an enum shared with `utils/data/routes.ts`, since a renumber in the data file silently breaks the badges.

SanithuGunasena07: scoped the aria-label to only apply when the drawer is collapsed:

frontend/src/community/common/components/organisms/Drawer/Drawer.tsx
SanithuGunasena07: Pre existing not code touched in this PR

frontend/src/community/common/components/organisms/Drawer/Drawer.tsx
SanithuGunasena07: Pre Existing out of scope for PR

SanithuGunasena07: not required

SanithuGunasena07: Pre Existing out of scope for PR

SanithuGunasena07: Pre Existing out of scope for PR

frontend/src/community/common/components/organisms/Drawer/Drawer.tsx
SanithuGunasena07: Pre Existing out of scope for PR

frontend/src/community/common/components/organisms/Drawer/Drawer.tsx
SanithuGunasena07: part 1 - Already fixed
part 2 - Pre Existing out of scope for PR

frontend/src/community/common/components/organisms/Drawer/Drawer.tsx
SanithuGunasena07: Pre Existing out of scope for PR

SanithuGunasena07: Pre Existing out of scope for PR

SanithuGunasena07: Pre Existing out of scope for PR

SanithuGunasena07: Pre Existing out of scope for PR

SanithuGunasena07: Pre Existing out of scope for PR

SanithuGunasena07: Pre Existing out of scope for PR

SanithuGunasena07: Pre Existing out of scope for PR

SanithuGunasena07: Pre Existing out of scope for PR

frontend/src/community/common/components/organisms/Drawer/Drawer.tsx
```diff
@@ -299,6 +299,7 @@ const Drawer = (): JSX.Element => {
...
                     aria-expanded={isExpanded}
                     aria-controls={`sub-list-${routeId}`}
+                    aria-label={!isDrawerExpanded ? route?.name : undefined}
```
hasalarootcode: use &&

hasalarootcode (dismissed):

frontend/src/community/common/components/organisms/Drawer/Drawer.tsx
SanithuGunasena07: updated to (!isDrawerExpanded && route?.name) || undefined

hasalarootcode (approved):

akilarootcode (approved):

PR #1818, opened 2026-08-20
fix/timesheet table issues

Deshan1100:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/frontend/src/community/attendance/components/molecules/EmployeeTimesheetRequestTable/EmployeeTimesheetRequestTable.tsx
+++ b/frontend/src/community/attendance/components/molecules/EmployeeTimesheetRequestTable/EmployeeTimesheetRequestTable.tsx
@@ -134,7 +134,7 @@ const EmployeeTimesheetRequestTable: FC<Props> = ({
   const tableHeaders: GridHeader[] = columns.map((col) => ({
     id: col.field,
     label: col.headerName,
-    align: "center"
+    align: "left"
   }));
   const transformToTableRows = (): GridRow[] => {
     return (requestData?.items ?? []).map(
--- a/frontend/src/community/attendance/components/molecules/EmployeeTimesheetRequestTable/styles.ts
+++ b/frontend/src/community/attendance/components/molecules/EmployeeTimesheetRequestTable/styles.ts
@@ -7,21 +7,21 @@ const styles = (theme: Theme) => ({
     display: "flex",
     flexDirection: "row",
     alignItems: "center",
-    justifyContent: "center",
+    justifyContent: "flex-start",
     flex: 0.7,
     maxWidth: "auto"
   },
   textDateStyles: {
     letterSpacing: "0.03em",
     whiteSpace: "nowrap",
-    textAlign: "center",
+    textAlign: "left",
     overflow: "hidden",
     textOverflow: "ellipsis"
   },
   outerBoxWrapper: {
     display: "flex",
     alignItems: "center",
-    justifyContent: "center",
+    justifyContent: "flex-start",
     flex: 1,
     maxWidth: "auto"
   },
@@ -29,7 +29,7 @@ const styles = (theme: Theme) => ({
     display: "inline-flex",
     flexDirection: "row",
     alignItems: "center",
-    justifyContent: "center",
+    justifyContent: "flex-start",
     gap: "0.5rem",
     [theme.breakpoints.down("md")]: {
       flexDirection: "column"
@@ -39,7 +39,7 @@ const styles = (theme: Theme) => ({
     letterSpacing: "0.03em",
     color: theme.palette.text.secondary,
     whiteSpace: "nowrap",
-    textAlign: "center",
+    textAlign: "left",
     overflow: "hidden",
     textOverflow: "ellipsis",
     textDecoration:
@@ -52,7 +52,7 @@ const styles = (theme: Theme) => ({
     letterSpacing: "0.03em",
     color: theme.palette.error.contrastText,
     whiteSpace: "nowrap",
-    textAlign: "center",
+    textAlign: "left",
     overflow: "hidden",
     textOverflow: "ellipsis"
   },
@@ -60,7 +60,7 @@ const styles = (theme: Theme) => ({
     letterSpacing: "0.03em",
     color: theme.palette.text.secondary,
     whiteSpace: "nowrap",
-    textAlign: "center",
+    textAlign: "left",
     overflow: "hidden",
     textOverflow: "ellipsis",
     textDecoration:
@@ -73,7 +73,7 @@ const styles = (theme: Theme) => ({
     display: "flex",
     flexDirection: "row",
     alignItems: "center",
-    justifyContent: "center",
+    justifyContent: "flex-start",
     flex: 1,
     maxWidth: "auto"
   },
@@ -81,11 +81,11 @@ const styles = (theme: Theme) => ({
     display: "flex",
     flexDirection: "row",
     alignItems: "center",
-    justifyContent: "center",
+    justifyContent: "flex-start",
     flex: 1,
     maxWidth: "auto",
     [theme.breakpoints.down("lg")]: {
-      justifyContent: "center"
+      justifyContent: "flex-start"
     },
     position: "relative"
   },
--- a/frontend/src/community/attendance/components/molecules/ManagerTimesheetRequestTable/ManagerTimesheetRequestTable.tsx
+++ b/frontend/src/community/attendance/components/molecules/ManagerTimesheetRequestTable/ManagerTimesheetRequestTable.tsx
@@ -163,7 +163,7 @@ const ManagerTimesheetRequestTable: FC<Props> = ({
   const tableHeaders: GridHeader[] = columns.map((col) => ({
     id: col.field,
     label: col.headerName,
-    align: "center"
+    align: "left"
   }));
 
   const transformToTableRows = (): GridRow[] => {
@@ -177,7 +177,7 @@ const ManagerTimesheetRequestTable: FC<Props> = ({
           isResponsiveLayout={true}
           chipStyles={{
             width: "fit-content",
-            maxWidth: "15.625rem",
+            maxWidth: "100%",
             backgroundColor: "var(--color-tertiary-background)"
           }}
         />
--- a/frontend/src/community/attendance/components/molecules/ManagerTimesheetRequestTable/styles.ts
+++ b/frontend/src/community/attendance/components/molecules/ManagerTimesheetRequestTable/styles.ts
@@ -7,21 +7,21 @@ const styles = (theme: Theme) => ({
     display: "flex",
     flexDirection: "row",
     alignItems: "center",
-    justifyContent: "center",
+    justifyContent: "flex-start",
     flex: 0.7,
     maxWidth: "auto"
   },
   textDateStyles: {
     letterSpacing: "0.03em",
     whiteSpace: "nowrap",
-    textAlign: "center",
+    textAlign: "left",
     overflow: "hidden",
     textOverflow: "ellipsis"
   },
   outerBoxWrapper: {
     display: "flex",
     alignItems: "center",
-    justifyContent: "center",
+    justifyContent: "flex-start",
     flex: 1,
     maxWidth: "auto"
   },
@@ -29,7 +29,7 @@ const styles = (theme: Theme) => ({
     display: "inline-flex",
     flexDirection: "row",
     alignItems: "center",
-    justifyContent: "center",
+    justifyContent: "flex-start",
     gap: "0.5rem",
     [theme.breakpoints.down("md")]: {
       flexDirection: "column"
@@ -39,7 +39,7 @@ const styles = (theme: Theme) => ({
     letterSpacing: "0.03em",
     color: theme.palette.text.secondary,
     whiteSpace: "nowrap",
-    textAlign: "center",
+    textAlign: "left",
     overflow: "hidden",
     textOverflow: "ellipsis",
     textDecoration:
@@ -52,7 +52,7 @@ const styles = (theme: Theme) => ({
     letterSpacing: "0.03em",
     color: theme.palette.error.contrastText,
     whiteSpace: "nowrap",
-    textAlign: "center",
+    textAlign: "left",
     overflow: "hidden",
     textOverflow: "ellipsis"
   },
@@ -60,7 +60,7 @@ const styles = (theme: Theme) => ({
     letterSpacing: "0.03em",
     color: theme.palette.text.secondary,
     whiteSpace: "nowrap",
-    textAlign: "center",
+    textAlign: "left",
     overflow: "hidden",
     textOverflow: "ellipsis",
     textDecoration:
@@ -73,7 +73,7 @@ const styles = (theme: Theme) => ({
     display: "flex",
     flexDirection: "row",
     alignItems: "center",
-    justifyContent: "center",
+    justifyContent: "flex-start",
     flex: 1,
     maxWidth: "auto"
   },
@@ -81,11 +81,11 @@ const styles = (theme: Theme) => ({
     display: "flex",
     flexDirection: "row",
     alignItems: "center",
-    justifyContent: "center",
+    justifyContent: "flex-start",
     flex: 1,
     maxWidth: "auto",
     [theme.breakpoints.down("lg")]: {
-      justifyContent: "center"
+      justifyContent: "flex-start"
     },
     position: "relative"
   },
--- a/frontend/src/community/leave/components/molecules/ManagerLeaveRequests/ManagerLeaveRequest.tsx
+++ b/frontend/src/community/leave/components/molecules/ManagerLeaveRequests/ManagerLeaveRequest.tsx
@@ -141,7 +141,7 @@ const ManagerLeaveRequest: FC<Props> = ({
               isResponsiveLayout
               chipStyles={{
                 width: "fit-content",
-                maxWidth: "15.625rem",
+                maxWidth: "100%",
                 justifyContent: "flex-start",
                 backgroundColor: "var(--color-tertiary-background)"
               }}
--- a/frontend/src/community/leave/components/molecules/PolicyManagerLeaveRequests/PolicyManagerLeaveRequests.tsx
+++ b/frontend/src/community/leave/components/molecules/PolicyManagerLeaveRequests/PolicyManagerLeaveRequests.tsx
@@ -131,7 +131,7 @@ const PolicyManagerLeaveRequests: FC = () => {
                 isResponsiveLayout
                 chipStyles={{
                   width: "fit-content",
-                  maxWidth: "15.625rem",
+                  maxWidth: "100%",
                   justifyContent: "flex-start",
                   backgroundColor: "var(--color-tertiary-background)"
                 }}


shakila-kamalasena (approved):

akilarootcode (approved):

ThinuwanW (commented):
## 🤖 Claude Code Review

This is a presentation-only PR: it switches the two timesheet request tables from centre- to left-alignment (header `align` plus the `justifyContent`/`textAlign` values in their colocated `styles.ts`) and relaxes the `AvatarChip` name-chip cap from `15.625rem` to `100%` in the manager timesheet table and the two manager leave-request tables. The alignment changes are mechanically consistent, but the `maxWidth: "100%"` change effectively removes the name-truncation cap entirely (MUI's Chip root already defaults to `max-width: 100%`), and none of these tables declare a column `width`, so a long employee name can now widen the name column and squeeze the rest of the row. Secondary issues: the manager timesheet action column header is now left-aligned while its cell content is still centred by auto margins, the two `styles.ts` files are byte-identical duplicates that had to receive the same ten-line edit twice, and a couple of the edited declarations are now dead or inert.

Found **15** new issue(s): 🔴 2 important, 🟡 6 suggestion(s), 🟣 7 nit(s)

frontend/src/community/attendance/components/molecules/ManagerTimesheetRequestTable/ManagerTimesheetRequestTable.tsx
```diff
@@ -177,7 +177,7 @@ const ManagerTimesheetRequestTable: FC<Props> = ({
           isResponsiveLayout={true}
           chipStyles={{
             width: "fit-content",
-            maxWidth: "15.625rem",
+            maxWidth: "100%",
```
ThinuwanW, line 180: 🤖 **Claude** · 🔴 **Important**

Changing `maxWidth` from `15.625rem` to `100%` removes the name truncation cap rather than relaxing it. MUI's `Chip` root already ships with `max-width: 100%`, so this override is a no-op and the only definite width constraint on the chip is gone. The ellipsis in `AvatarChip/styles.ts` (`.MuiChip-label` → `overflow: hidden; text-overflow: ellipsis; maxWidth: 100%`) is also percentage-based, and a percentage max-width does not participate in intrinsic (max-content) sizing, so with `width: fit-content` and `white-space: nowrap` the chip will now size to the full name. None of these tables set a `width` on the `name` header (see `tableHeaders` at line 163 here, and the equivalents in the two leave tables), so in an auto-layout table a long "firstName lastName" can widen the name column and squeeze `date`/`from`/`to`/`workedHours`/`status`. The same edit was made in `ManagerLeaveRequest.tsx:144` and `PolicyManagerLeaveRequests.tsx:134`, so all three tables are affected. If the goal was simply to let the chip use the available cell width instead of a hard 250px, use a bounded value — e.g. `maxWidth: "min(100%, 15.625rem)"` — or give the `name` header an explicit `width` (as `CompanyTable`/`DocumentsTable` do) so the column stays bounded. Whichever you pick, please verify with a long name (e.g. 30+ characters) at ~1280px viewport width before merging. Also note that `isTooltipEnabled` is not set on any of these three `AvatarChip` usages, so if the name does still get clipped the user has no way to read it — worth enabling while you are here.

frontend/src/community/attendance/components/molecules/ManagerTimesheetRequestTable/ManagerTimesheetRequestTable.tsx
```diff
@@ -163,7 +163,7 @@ const ManagerTimesheetRequestTable: FC<Props> = ({
   const tableHeaders: GridHeader[] = columns.map((col) => ({
     id: col.field,
     label: col.headerName,
-    align: "center"
+    align: "left"
```
ThinuwanW, line 166: 🤖 **Claude** · 🟡 **Suggestion**

The last column's header (`actionHeaderTxt`, field `status`) is now left-aligned along with every other header, but the cell content for PENDING rows is still explicitly centred: the two `IconButton`s at lines 267-313 use `margin: "0rem 0.75rem 0rem auto"` and `margin: "0rem auto 0rem 0rem"`, which is an auto-margin centring trick. The result is a left-aligned "Action" header sitting above a pair of centred buttons, while non-pending rows render a badge that follows normal flow. Either drop the `auto` margins on both `IconButton`s (e.g. `mr: "0.75rem"` on the decline button, no margin on the approve button) so the actions start at the left edge like the header, or keep `align: "center"` for just the `status` column by mapping it conditionally instead of applying `"left"` to every column.

frontend/src/community/attendance/components/molecules/ManagerTimesheetRequestTable/styles.ts
```diff
@@ -7,29 +7,29 @@ const styles = (theme: Theme) => ({
     display: "flex",
     flexDirection: "row",
     alignItems: "center",
-    justifyContent: "center",
+    justifyContent: "flex-start",
```
ThinuwanW, line 10: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

`ManagerTimesheetRequestTable/styles.ts` and `EmployeeTimesheetRequestTable/styles.ts` are byte-for-byte identical (same pre- and post-image blob in this diff), which is why this PR had to make the exact same ten-line edit twice. This is the "2+ components with near-identical code → extract" case from the review guidelines, and it is a live drift risk: the next alignment tweak will very likely land in only one copy. Extract into a single shared module (e.g. `~community/attendance/components/molecules/TimesheetRequestTable/styles.ts` or a `timesheetRequestTableStyles` helper under `attendance/utils`) and import it from both components. If the two tables genuinely need to diverge later, parameterise the shared factory rather than re-forking the file.

frontend/src/community/attendance/components/molecules/EmployeeTimesheetRequestTable/styles.ts
```diff
@@ -73,19 +73,19 @@ const styles = (theme: Theme) => ({
...
     display: "flex",
     flexDirection: "row",
     alignItems: "center",
-    justifyContent: "center",
+    justifyContent: "flex-start",
     flex: 1,
     maxWidth: "auto",
     [theme.breakpoints.down("lg")]: {
-      justifyContent: "center"
+      justifyContent: "flex-start"
```
ThinuwanW, line 88: 🤖 **Claude** · 🟣 **Nit**

After this change the `theme.breakpoints.down("lg")` block inside `statusOuterBoxStyles` sets `justifyContent: "flex-start"`, which is now identical to the base value on line 84 — the media query is dead code that only costs an extra emotion rule. Delete the breakpoint block entirely and keep just the base `justifyContent: "flex-start"`. Applies to both copies of this file (Employee and Manager).

frontend/src/community/attendance/components/molecules/ManagerTimesheetRequestTable/styles.ts
```diff
@@ -73,19 +73,19 @@ const styles = (theme: Theme) => ({
...
     flex: 1,
     maxWidth: "auto"
   },
   statusOuterBoxStyles: {
     display: "flex",
     flexDirection: "row",
     alignItems: "center",
-    justifyContent: "center",
+    justifyContent: "flex-start",
```
ThinuwanW, line 84: 🤖 **Claude** · 🟣 **Nit**

`statusOuterBoxStyles` is not referenced anywhere in `ManagerTimesheetRequestTable.tsx` — that component renders its status cell as a bare fragment (line 264 onward), not inside a `Box sx={classes.statusOuterBoxStyles}`. So this PR is editing a dead style key in this copy of the file. Related: `kebabMenuBoxStyle` (used at `ManagerTimesheetRequestTable.tsx:327`) relies on `position: relative` from `statusOuterBoxStyles`, which is never applied here, so its `position: absolute; right: 10%` resolves against some outer positioned ancestor; and that branch is unreachable anyway because it is guarded by `status === PENDING` inside the `else` arm of the same `status === PENDING` ternary. All pre-existing, but the shared-styles extraction suggested above is a good moment to prune it.

frontend/src/community/attendance/components/molecules/EmployeeTimesheetRequestTable/styles.ts
```diff
@@ -7,29 +7,29 @@ const styles = (theme: Theme) => ({
...
+    justifyContent: "flex-start",
     flex: 1,
     maxWidth: "auto"
   },
   timeBadgeContentStyles: {
     display: "inline-flex",
     flexDirection: "row",
     alignItems: "center",
-    justifyContent: "center",
+    justifyContent: "flex-start",
```
ThinuwanW, line 32: 🤖 **Claude** · 🟣 **Nit**

`timeBadgeContentStyles` is `display: "inline-flex"` with no width constraint, so it is shrink-to-fit in both the row and (below `md`) column directions — `justifyContent` has no observable effect either before or after this change. Same for `boxDateContainer`'s and `timeBadgeContentStyles`' `textAlign` siblings on nowrap single-line text. Harmless, but if the intent was to left-align the two time values you can drop these two declarations to keep the diff to the rules that actually do something. Applies to both copies of the file.

frontend/src/community/attendance/components/molecules/EmployeeTimesheetRequestTable/styles.ts
```diff
@@ -7,29 +7,29 @@ const styles = (theme: Theme) => ({
     display: "flex",
     flexDirection: "row",
     alignItems: "center",
-    justifyContent: "center",
+    justifyContent: "flex-start",
     flex: 0.7,
     maxWidth: "auto"
```
ThinuwanW, line 12: 🤖 **Claude** · 🟣 **Nit**

`maxWidth: "auto"` is not a valid `max-width` value (the property accepts `none`, a length, a percentage, or the intrinsic keywords — not `auto`), so the browser drops the declaration. It appears four times in this file (lines 12, 26, 78, 86) and four more times in the Manager copy. Pre-existing, but since this PR is already rewriting these exact style objects it is a cheap cleanup: remove them, or use `maxWidth: "none"` if the intent was to defeat an inherited cap.

frontend/src/community/attendance/components/molecules/EmployeeTimesheetRequestTable/styles.ts
```diff
@@ -73,19 +73,19 @@ const styles = (theme: Theme) => ({
...
+    justifyContent: "flex-start",
     flex: 1,
     maxWidth: "auto"
   },
   statusOuterBoxStyles: {
     display: "flex",
     flexDirection: "row",
     alignItems: "center",
-    justifyContent: "center",
+    justifyContent: "flex-start",
```
ThinuwanW, line 84: 🤖 **Claude** · 🔴 **Important**

`statusOuterBoxStyles` flipped to `justifyContent: "flex-start"` but the sibling `kebabMenuBoxStyle` (lines 92-104) was left as `position: "absolute"` with `right: "10%"` / 5% / 0% / 5% breakpoint offsets. In EmployeeTimesheetRequestTable those offsets were tuned for a *centred* badge; now the status badge sits hard against the left edge of the cell while the kebab trigger stays pinned near the right edge, so PENDING rows render a large dead gap between the status chip and the menu that grows with column width. It also means the kebab position no longer tracks the badge at all. Fix: drop the absolute positioning and make the kebab a normal flex sibling — e.g. remove `position: "relative"`/`position: "absolute"` and give `kebabMenuBoxStyle` `marginLeft: "auto"` (to keep it right-edge) or a plain `gap` on the parent (to keep it adjacent to the badge), then delete the now-pointless `right` breakpoint overrides. Note the identical `kebabMenuBoxStyle` exists in ManagerTimesheetRequestTable/styles.ts and needs the same treatment.

frontend/src/community/attendance/components/molecules/ManagerTimesheetRequestTable/ManagerTimesheetRequestTable.tsx
ThinuwanW, line 166: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Unreachable branch / dead code in the `status` cell. The cell is a ternary on `timesheetRequest?.status === TimeSheetRequestStates.PENDING`; inside the *else* arm it re-tests `timesheetRequest?.status === TimeSheetRequestStates.PENDING` (around line 326) before rendering the `KebabMenu`. That inner condition can never be true, so the entire manager-side cancel path is dead: `KebabMenu` (import line 25), `getKebabMenuOptions` (line 119), `useCancelTimeRequest`/`cancelMutate` (line 134), `onSuccess` (line 87), `onError` (line 102), and `classes.kebabMenuBoxStyle`. This is also the reason `statusOuterBoxStyles` ends up unused in this component. Since the PR is reworking this exact column, either delete the dead branch and its now-unused imports/handlers/styles, or fix the intended condition (e.g. render the kebab for non-PENDING cancellable states) if managers really are supposed to be able to cancel a request from this table.

frontend/src/community/attendance/components/molecules/EmployeeTimesheetRequestTable/styles.ts
ThinuwanW, line 32: 🤖 **Claude** · 🟡 **Suggestion**

`timeBadgeContentStyles` sets `justifyContent: "flex-start"`, but the same rule flips to `flexDirection: "column"` under `theme.breakpoints.down("md")` (lines 34-36). Once the container is a column, `justifyContent` governs the *vertical* axis and the horizontal placement of the stacked from/to times is governed by `alignItems`, which is still `"center"` (line 31). So below `md` — precisely the responsive case this table cares about — the two time values remain horizontally centred relative to each other, contradicting the left-alignment the PR is introducing. Add `alignItems: "flex-start"` inside the `breakpoints.down("md")` block (in both this file and the ManagerTimesheetRequestTable copy).

frontend/src/community/attendance/components/molecules/EmployeeTimesheetRequestTable/EmployeeTimesheetRequestTable.tsx
```diff
@@ -134,7 +134,7 @@ const EmployeeTimesheetRequestTable: FC<Props> = ({
   const tableHeaders: GridHeader[] = columns.map((col) => ({
     id: col.field,
     label: col.headerName,
-    align: "center"
+    align: "left"
```
ThinuwanW, line 137: 🤖 **Claude** · 🟡 **Suggestion**

Two concerns with `align: "left"` (same at ManagerTimesheetRequestTable.tsx:166). (1) It restates the default: every other table in the repo omits `align` and renders left — ManagerLeaveRequest.tsx:119-122 and PolicyManagerLeaveRequests.tsx:105-108 (both touched by this PR) map to `{ id, label }` only, as do LeaveTypesTable and CustomLeaveAllocationsTable, which supply `align` solely for `"right"` columns. Dropping the property entirely is the idiomatic change and makes these two `columns.map` callbacks identical to the leave tables'. (2) `align` on `GridHeader` appears to drive *cell* alignment as well, not just the header — LeaveTypesTable.tsx:51 sets `align: "right"` on its actions column and the cell (a bare `IconButton`, line 90-98) carries no right-alignment styling of its own. If that holds for TableV2 here, the header change alone already left-aligns the cells and the ten `justifyContent`/`textAlign` flips in styles.ts are a redundant second mechanism that will drift out of sync. Please confirm which layer actually aligns the cell and keep only one.

frontend/src/community/attendance/components/molecules/ManagerTimesheetRequestTable/ManagerTimesheetRequestTable.tsx
ThinuwanW, line 180: 🤖 **Claude** · 🟡 **Suggestion**

This `chipStyles` object omits `justifyContent: "flex-start"`, which both leave tables pass alongside the identical `width`/`maxWidth`/`backgroundColor` trio (ManagerLeaveRequest.tsx:145, PolicyManagerLeaveRequests.tsx:135). MUI's `Chip` root defaults to `justify-content: center`, so this name chip centres its avatar+label while the equivalent leave chips left-align theirs — an inconsistency that is more visible now that `maxWidth` is `100%` and the chip can span the full column. Add `justifyContent: "flex-start"` here so the four call sites match, and consider hoisting this repeated four-property object into a shared constant since it is now duplicated verbatim in three files.

frontend/src/community/attendance/components/molecules/EmployeeTimesheetRequestTable/styles.ts
```diff
@@ -7,29 +7,29 @@ const styles = (theme: Theme) => ({
...
     flex: 0.7,
     maxWidth: "auto"
   },
   textDateStyles: {
     letterSpacing: "0.03em",
     whiteSpace: "nowrap",
-    textAlign: "center",
+    textAlign: "left",
```
ThinuwanW, line 17: 🤖 **Claude** · 🟣 **Nit**

The five `textAlign: "center" -> "left"` flips in this file (lines 17, 42, 55, 63, and the Manager copy) are inert. Each applies to a `Typography` that is a shrink-to-fit flex item inside a `display: flex` / `inline-flex` parent and carries `whiteSpace: "nowrap"`, so the element's box always hugs its own text and `text-align` never has leftover inline space to distribute. (The paired `overflow: "hidden"` / `textOverflow: "ellipsis"` on those same rules are dead for the same reason — a nowrap shrink-to-fit box cannot overflow itself.) The alignment actually being changed comes from the parents' `justifyContent`. Consider dropping the `textAlign` lines rather than flipping them, so the diff reflects only the declarations that have an effect.

frontend/src/community/leave/components/molecules/ManagerLeaveRequests/ManagerLeaveRequest.tsx
```diff
@@ -141,7 +141,7 @@ const ManagerLeaveRequest: FC<Props> = ({
               isResponsiveLayout
               chipStyles={{
                 width: "fit-content",
-                maxWidth: "15.625rem",
+                maxWidth: "100%",
```
ThinuwanW, line 144: 🤖 **Claude** · 🟣 **Nit**

This PR applies the same `maxWidth` edit to ManagerLeaveRequest and PolicyManagerLeaveRequests, confirming the two are kept in lockstep, but their `chipClassName` constants have already drifted: this file (line 34-35) is `"body2 inline-flex w-fit ..."` while PolicyManagerLeaveRequests.tsx:39-40 is the same string minus `body2`. Since `.body2` is a real utility in styles/global.css (14px/400), the new policy table renders its type and status chips at the inherited font size instead. Worth aligning the two strings — or, better, hoisting the shared constant so the legacy and policy variants cannot diverge further.

frontend/src/community/attendance/components/molecules/EmployeeTimesheetRequestTable/EmployeeTimesheetRequestTable.tsx
ThinuwanW, line 137: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

Accessibility inconsistency in the status cell (around line 228): this file renders the decorative status icon as a bare `<span>{requestTypeSelector(...)}</span>`, whereas the equivalent cell in ManagerTimesheetRequestTable.tsx:321 wraps it as `<span role="img" aria-hidden="true">`. The icon duplicates the adjacent `pascalCaseFormatter(status)` text, so screen readers get redundant output here. Add `role="img" aria-hidden="true"` to match the manager table.

PR #1819, opened 2026-08-20
Fix/3070 late arrival trend invalid time blocks

shakila-kamalasena:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR replaces hard failures on malformed `timeBlocks` JSON with silent fallbacks to a half-of-`totalHours` default, and consolidates the create/update time-config paths behind a shared `resolveTimeBlocks` helper. The consolidation is a genuine improvement, but the analytics fallback is incomplete: a structurally valid array that simply lacks a `MORNING_HOURS` entry (or carries a non-numeric `hours` value) still reaches `Double.parseDouble` and produces a 500, which is the exact failure mode the change set out to remove. Additionally, moving the `totalHours / 2` computation into `resolveTimeBlocks` puts an unboxing NPE on the create path where none existed, since `totalHours` is a nullable `Float` and the endpoint has neither `@Valid` nor `@NotNull`.

Found **18** new issue(s): 🔴 6 important, 🟡 9 suggestion(s), 🟣 3 nit(s)

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeAnalyticsServiceImpl.java
```diff
@@ -413,7 +414,7 @@ private boolean isLateArrivalBasedOnLeave(LeaveRequest leaveRequest, LocalTime r
 			if (leaveRequest.getLeaveState() == LeaveState.FULLDAY)
 				return false;
 			if (leaveRequest.getLeaveState() == LeaveState.HALFDAY_MORNING) {
-				TimeBlockDto timeBlockDto = processTimeBlocks(timeConfig.getTimeBlocks());
+				TimeBlockDto timeBlockDto = processTimeBlocks(timeConfig.getTimeBlocks(), timeConfig.getTotalHours());
```
ThinuwanW, line 418: 🤖 **Claude** · 🔴 **Important**

The new fallback is incomplete, so the NPE/500 this PR aims to fix is still reachable. `processTimeBlocks` only falls back to defaults when the node is null/non-array/empty, or when a block is missing the `timeBlock`/`hours` keys. It does NOT cover a structurally valid array that contains neither a `MORNING_HOURS` block nor a parseable value:

1) Missing morning block — `TimeBlocks` also has `EXTRA_HOURS`, and nothing validates which blocks a client sends. A stored value of `[{"timeBlock":"EXTRA_HOURS","hours":2}]` (or evening-only) passes every guard, the loop matches nothing, and the returned DTO has `morningHours == null`. Line 419 then calls `Double.parseDouble(null)` → NPE → opaque 500 on the late-arrival dashboard for any employee with HALFDAY_MORNING leave.

2) Unparseable value — `hours` is read with `asString()` and never validated. A value of `""`, `"4h"`, or a JSON object/array yields `NumberFormatException` at line 419, again unwrapped into a 500 (see checklist item 1: every `Double.parseDouble` must be guarded).

Fix both inside `processTimeBlocks` so the caller can rely on the contract: after the loop, if `morningHours`/`eveningHours` are still null, return `buildDefaultTimeBlocks(totalHours)`; and parse each `hours` value there (try/catch `NumberFormatException`, or check `block.get("hours").isNumber()`) falling back to the default on failure. Returning a `TimeBlockDto` with numeric fields typed as `Float` instead of `String` would make this class of bug structurally impossible.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeServiceImpl.java
```diff
@@ -1314,15 +1314,21 @@ private TimeConfig createTimeConfig(TimeConfigDto.DayCapacity timeConfig) {
...
 
-	private void updateTimeConfig(TimeConfig currentConfig, TimeConfigDto.DayCapacity timeConfig) {
-		currentConfig.setTotalHours(timeConfig.totalHours());
+	private JsonNode resolveTimeBlocks(TimeConfigDto.DayCapacity timeConfig) {
+		if (timeConfig.timeBlocks() != null && !timeConfig.timeBlocks().isEmpty()) {
+			return mapper.valueToTree(timeConfig.timeBlocks());
+		}
 
 		Map<String, Float> hoursMap = new HashMap<>();
 		hoursMap.put(TimeBlocks.MORNING_HOURS.name(), timeConfig.totalHours() / 2);
```
ThinuwanW, line 1325: 🤖 **Claude** · 🔴 **Important**

`resolveTimeBlocks` introduces an unboxing NPE on the create path. `TimeConfigDto.DayCapacity.totalHours()` is a boxed `Float`, `TimeController.updateTimeConfig` (line 49) has no `@Valid` on the `@RequestBody`, and `TimeConfigDto`/`DayCapacity` carry no bean-validation annotations — so `totalHours: null` is fully reachable from `PATCH /v1/time/config`.

Previously `createTimeConfig` never divided `totalHours` (it called `mapper.valueToTree(timeBlocks)` unconditionally), so a payload with `totalHours: null` and absent `timeBlocks` did not touch this arithmetic. Now `timeConfig.totalHours() / 2` unboxes null → NPE → 500 with no useful error body.

Fix: add `@Valid` to the `@RequestBody` in `TimeController.updateTimeConfig` and `@NotNull` on `totalHours` (plus `@Valid` on the nested `Set<TimeBlock>` / `@NotNull` on `day` and `time()`, which is dereferenced unguarded at lines 1309 and 1332) so this returns a 400 through `GlobalExceptionHandler`. Alternatively, guard in `resolveTimeBlocks` and throw `new ModuleException(TimeMessageConstant.TIME_ERROR_INVALID_TIME_BLOCKS)` — note the entity column is `nullable = false`, so null was never a valid persisted value anyway.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeAnalyticsServiceImpl.java
```diff
@@ -413,7 +414,7 @@ private boolean isLateArrivalBasedOnLeave(LeaveRequest leaveRequest, LocalTime r
...
 				LocalTime adjustedLateThreshold = lateThreshold
 					.plusHours((long) Double.parseDouble(timeBlockDto.getMorningHours()));
```
ThinuwanW, line 420: 🤖 **Claude** · 🟡 **Suggestion**

Fractional default hours are silently truncated. `(long) Double.parseDouble(...)` drops the fraction, and the new default path makes half-hour values routine: `totalHours = 9` now yields `morningHours = "4.5"`, and the late threshold advances only 4 hours instead of 4h30m — employees are flagged late 30 minutes early on morning half-day leave. The same truncation exists in `TimeUtil.isCurrentTimeInMorning`/`isCurrentTimeInEvening` (`(long) morningHours`), so the two code paths disagree on the same config.

Fix: use minutes rather than whole hours, e.g. `lateThreshold.plusMinutes(Math.round(Double.parseDouble(timeBlockDto.getMorningHours()) * 60))`, and consider applying the same treatment in `TimeUtil` for consistency.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeAnalyticsServiceImpl.java
```diff
@@ -540,35 +541,44 @@ private String formatWeekRange(LocalDate start, LocalDate end) {
...
 			}
 		}
-		else {
-			throw new ModuleException(TimeMessageConstant.TIME_ERROR_INVALID_TIME_BLOCKS);
-		}
 
 		return timeBlockDto;
 	}
 
+	private TimeBlockDto buildDefaultTimeBlocks(Float totalHours) {
```
ThinuwanW, line 577: 🤖 **Claude** · 🟡 **Suggestion**

The "half of totalHours" default is now encoded in three places with two different semantics, and this PR adds the third. `buildDefaultTimeBlocks` here and `resolveTimeBlocks` in `TimeServiceImpl` (lines 1322-1325) both split `totalHours / 2` across MORNING/EVENING, while `TimeUtil.extractHours` — the other consumer of the very same `TimeConfig.timeBlocks` JSON — defaults missing blocks to `0.0f`. So identical stored data is interpreted as "half a day" by the analytics path and "zero hours" by the active-slot path (checklist items 9 and 10).

Also note `TimeUtil.extractHours` still NPEs outright on a null `timeBlocks` node (the for-each at line 28) — legacy rows written before `resolveTimeBlocks` existed can hit this, so the robustness added here does not protect that path.

Suggest extracting a single `TimeUtil` helper (e.g. `resolveTimeBlockHours(TimeConfig)` returning `Map<String, Float>`) used by the analytics path, `TimeServiceImpl`, and `extractHours`, so the default and the null/malformed handling are defined exactly once.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeAnalyticsServiceImpl.java
```diff
@@ -540,35 +541,44 @@ private String formatWeekRange(LocalDate start, LocalDate end) {
 		}
 	}
 
-	private TimeBlockDto processTimeBlocks(JsonNode timeBlocks) {
+	private TimeBlockDto processTimeBlocks(JsonNode timeBlocks, Float totalHours) {
+		if (timeBlocks == null || !timeBlocks.isArray() || timeBlocks.isEmpty()) {
+			return buildDefaultTimeBlocks(totalHours);
```
ThinuwanW, line 547: 🤖 **Claude** · 🟡 **Suggestion**

The fallback is completely silent — invalid or missing `timeBlocks` config now produces plausible-looking but fabricated numbers with no trace. Since these are the only two signals that a `time_config` row is corrupt (the `ModuleException` that used to surface it is gone), operators will have no way to find bad rows.

Add a `log.warn` before each fallback return, including the day/`TimeConfig` identity, e.g. `log.warn("processTimeBlocks: invalid or missing time blocks, falling back to defaults derived from totalHours: {}", totalHours)`. This class already has `@Slf4j`.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeAnalyticsServiceImpl.java
```diff
@@ -540,35 +541,44 @@ private String formatWeekRange(LocalDate start, LocalDate end) {
...
+		}
+
 		TimeBlockDto timeBlockDto = new TimeBlockDto();
+		for (JsonNode block : timeBlocks) {
+			if (!block.has("timeBlock") || !block.has("hours")) {
+				return buildDefaultTimeBlocks(totalHours);
```
ThinuwanW, line 554: 🤖 **Claude** · 🟡 **Suggestion**

`TimeMessageConstant.TIME_ERROR_INVALID_TIME_BLOCKS` is now dead code — this was its only usage in the codebase. Either remove the enum constant and the `api.error.time.invalid-time-blocks` key from `community/messages/time-messages.properties`, or reuse it for the genuinely unrecoverable case (see the `totalHours` null finding in `TimeServiceImpl`) rather than leaving an orphaned message constant.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeAnalyticsServiceImpl.java
```diff
@@ -540,35 +541,44 @@ private String formatWeekRange(LocalDate start, LocalDate end) {
...
 
-	private TimeBlockDto processTimeBlocks(JsonNode timeBlocks) {
+	private TimeBlockDto processTimeBlocks(JsonNode timeBlocks, Float totalHours) {
+		if (timeBlocks == null || !timeBlocks.isArray() || timeBlocks.isEmpty()) {
+			return buildDefaultTimeBlocks(totalHours);
+		}
+
 		TimeBlockDto timeBlockDto = new TimeBlockDto();
+		for (JsonNode block : timeBlocks) {
+			if (!block.has("timeBlock") || !block.has("hours")) {
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`block.has(...)` returns true for an explicit JSON null, so `{"timeBlock": null, "hours": null}` passes this guard and then hits `asString()` on a `NullNode` at lines 555-556 — which either yields the literal string "null" (silently producing a DTO with `morningHours` unset, feeding the NPE described above) or throws, depending on Jackson 3 coercion rules. Use `hasNonNull("timeBlock") && hasNonNull("hours")` instead so explicit nulls route to the default path along with missing keys.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeAnalyticsServiceImpl.java
```diff
@@ -540,35 +541,44 @@ private String formatWeekRange(LocalDate start, LocalDate end) {
...
-					timeBlockDto.setMorningHours(hours);
-				}
-				else if ("EVENING_HOURS".equals(timeBlock)) {
-					timeBlockDto.setEveningTimeBlock(timeBlock);
-					timeBlockDto.setEveningHours(hours);
-				}
+			String timeBlock = block.get("timeBlock").asString();
+			String hours = block.get("hours").asString();
+
+			if ("MORNING_HOURS".equals(timeBlock)) {
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Inconsistent use of constants within the same change. `buildDefaultTimeBlocks` correctly uses the newly imported `TimeBlocks.MORNING_HOURS.name()`/`TimeBlocks.EVENING_HOURS.name()`, but `processTimeBlocks` still compares against the string literals `"MORNING_HOURS"`/`"EVENING_HOURS"` and reads the literal keys `"timeBlock"`/`"hours"` — even though `TimeConfigFieldName.TIME_BLOCK`/`HOURS` exist for exactly this and are what `TimeServiceImpl.createTimeConfigJsonNode` writes. Since the import is already present, switch the literals to the enums so a rename can't silently desynchronize the writer and the reader.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeAnalyticsServiceImpl.java
```diff
@@ -540,35 +541,44 @@ private String formatWeekRange(LocalDate start, LocalDate end) {
 		}
 	}
 
-	private TimeBlockDto processTimeBlocks(JsonNode timeBlocks) {
+	private TimeBlockDto processTimeBlocks(JsonNode timeBlocks, Float totalHours) {
```
ThinuwanW, line 545: 🤖 **Claude** · 🟡 **Suggestion**

No test coverage for the new behavior. `TimeAnalyticsServiceImplUnitTest` exists but contains no case touching `timeBlocks` or `isLateArrival`, and this PR changes an exception-throwing contract into a silent-default one without adding any. Add cases for: null/empty/non-array `timeBlocks`, a block missing `hours`, an evening-only or `EXTRA_HOURS`-only array, and a non-numeric `hours` value — the last two currently fail (see the first finding), so the tests would catch the remaining gap.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeServiceImpl.java
```diff
@@ -1314,15 +1314,21 @@ private TimeConfig createTimeConfig(TimeConfigDto.DayCapacity timeConfig) {
 		return newTimeConfig;
 	}
 
-	private void updateTimeConfig(TimeConfig currentConfig, TimeConfigDto.DayCapacity timeConfig) {
-		currentConfig.setTotalHours(timeConfig.totalHours());
+	private JsonNode resolveTimeBlocks(TimeConfigDto.DayCapacity timeConfig) {
```
ThinuwanW, line 1315: 🤖 **Claude** · 🔴 **Important**

Missing paired change: the other reader of `TimeConfig.timeBlocks` was not hardened. This PR accepts that the column can be null (`time_blocks json DEFAULT NULL` in common-ddl-script-v1-create-tables.sql:107, and the new `timeBlocks == null` guard in TimeAnalyticsServiceImpl) and fixes the analytics reader. But `TimeUtil.extractHours` (TimeUtil.java:27-33) still does a bare enhanced-for over `currentDayConfig.getTimeBlocks()` and then `block.get("timeBlock").asString()` / `block.get("hours").asFloat()` with no null or shape guard. It is called from `TimeServiceImpl` line 1200 on the `getActiveTimeSlot` / clock-in path, so a config row with null or malformed `time_blocks` still throws an unhandled NPE and returns a 500 — for the interactive clock-in flow rather than a dashboard widget, which is arguably the higher-impact of the two. Add the same guard there (return the MORNING/EVENING defaults from `totalHours` when the node is null or not an array). Note also that `extractHours` defaults a missing block to `0.0f` while the new `buildDefaultTimeBlocks` defaults it to `totalHours / 2`; after this PR the two readers of the same column disagree on what a missing block means, so pick one default and share it.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeServiceImpl.java
```diff
@@ -1304,7 +1304,7 @@ private ResponseEntityDto getAllActiveSlots(Long employeeWorkLocationId, LocalDa
 	private TimeConfig createTimeConfig(TimeConfigDto.DayCapacity timeConfig) {
 		TimeConfig newTimeConfig = new TimeConfig();
 		newTimeConfig.setDay(timeConfig.day());
-		newTimeConfig.setTimeBlocks(mapper.valueToTree(timeConfig.timeBlocks()));
+		newTimeConfig.setTimeBlocks(resolveTimeBlocks(timeConfig));
```
ThinuwanW, line 1305: 🤖 **Claude** · 🔴 **Important**

📁 **File-level observation**

The root cause of the malformed `time_blocks` rows is an entirely unvalidated request payload, and this PR patches the symptom downstream instead. `TimeController.updateTimeConfig` (TimeController.java:49) takes `@RequestBody TimeConfigDto` with no `@Valid`, and TimeConfigDto carries zero bean-validation annotations: `DayCapacity(DayOfWeek day, Set<TimeBlock> timeBlocks, Float totalHours, boolean isWeekStartDay, LocalTime time)` and `TimeBlock(TimeBlocks timeBlock, Float hours)`. Consequences on this exact code path: (a) `hours` may be null, and `resolveTimeBlocks` serialises that straight into the JSON column via `mapper.valueToTree`, producing precisely the `{"timeBlock":"MORNING_HOURS","hours":null}` rows the reader then chokes on; (b) `timeConfig.time()` is dereferenced unguarded at lines 1309 and 1332, so a payload omitting `time` is an NPE/500; (c) `totalHours` may be null or negative. Fix at the boundary: add `@Valid` to the `@RequestBody`, `@NotNull @Positive` on `totalHours`, `@NotNull` on `day` and `time`, `@NotNull` on both `TimeBlock` components, and `@Valid` on the nested `Set<TimeBlock>` so element constraints actually run.

ThinuwanW, line 1305: 🤖 **Claude** · 🔴 **Important**

📁 **File-level observation**

`updateTimeConfigs` (line 211) — the public entry point for both helpers this PR rewrites — has no `@Transactional`, and it is far from a simple write. It does `timeConfigDao.findAll()`, mutates/creates configs, then `setRemovingTimeConfigs` (line 2066) cancels or revokes future leave requests, calls `updateLeaveEntitlement`, calls `handleCalendarEventsDeletion`, and sends emails/notifications — and only then calls `saveAll`. Any exception between those steps leaves leave requests cancelled, entitlements adjusted and calendar events deleted while the config changes are silently rolled back to nothing. This matters more after this PR because `resolveTimeBlocks` is now invoked per day capacity inside the `forEach`, so a bad later element (e.g. null `totalHours`) can throw after earlier elements have already been processed. Add `@Transactional` to the method. Separately, the emails/notifications should be moved to after-commit rather than sent mid-write.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeAnalyticsServiceImpl.java
```diff
@@ -540,35 +541,44 @@ private String formatWeekRange(LocalDate start, LocalDate end) {
...
 		}
-		else {
-			throw new ModuleException(TimeMessageConstant.TIME_ERROR_INVALID_TIME_BLOCKS);
-		}
 
 		return timeBlockDto;
 	}
 
+	private TimeBlockDto buildDefaultTimeBlocks(Float totalHours) {
+		float halfDayHours = (totalHours != null ? totalHours : 0f) / 2;
```
ThinuwanW, line 578: 🤖 **Claude** · 🔴 **Important**

The `totalHours != null ? totalHours : 0f` branch is unreachable dead code, and the value it would produce is wrong. `total_hours` is `NOT NULL` in the schema (common-ddl-script-v1-create-tables.sql:108) and `nullable = false` on the entity (TimeConfig.java:42), and the only caller chain feeds `processTimeBlocks` a `TimeConfig` loaded via `timeConfigDao.findByDay(...)` (line 393), so the null branch can never be taken. If it ever were, `halfDayHours` becomes 0.0, `getMorningHours()` returns "0.0", and line 419's `lateThreshold.plusHours(0)` silently cancels the half-day-morning adjustment entirely — every half-day-morning employee would be flagged as a late arrival with no error. Take `float totalHours` as a primitive parameter (and have `processTimeBlocks` do the same) so the impossible branch disappears instead of encoding a silently-wrong answer.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeServiceImpl.java
```diff
@@ -1314,15 +1314,21 @@ private TimeConfig createTimeConfig(TimeConfigDto.DayCapacity timeConfig) {
 		return newTimeConfig;
 	}
 
-	private void updateTimeConfig(TimeConfig currentConfig, TimeConfigDto.DayCapacity timeConfig) {
-		currentConfig.setTotalHours(timeConfig.totalHours());
+	private JsonNode resolveTimeBlocks(TimeConfigDto.DayCapacity timeConfig) {
+		if (timeConfig.timeBlocks() != null && !timeConfig.timeBlocks().isEmpty()) {
```
ThinuwanW, line 1316: 🤖 **Claude** · 🟡 **Suggestion**

`resolveTimeBlocks` treats "at least one block supplied" as "fully and correctly supplied", so partial or nonsensical configs are persisted verbatim. A payload containing only `MORNING_HOURS`, or only `EXTRA_HOURS` (the `TimeBlocks` enum has three values, not two), passes the `!isEmpty()` check and is written straight to the column. Nothing checks that the supplied set covers MORNING and EVENING, or that the hours sum to `totalHours`. Downstream, `TimeUtil.extractHours` silently returns 0.0f for the absent block, which collapses the evening window in `isCurrentTimeInEvening` and makes leave/holiday half-day detection wrong. Either validate the set here and reject with `throw new ModuleException(TimeMessageConstant.TIME_ERROR_INVALID_TIME_BLOCKS)` — which would also give that now-orphaned constant a purpose — or make the fallback per-block: start from the `totalHours / 2` defaults and overwrite only the blocks the caller actually supplied, instead of the current all-or-nothing branch.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeAnalyticsServiceImpl.java
ThinuwanW, line 577: 🤖 **Claude** · 🟡 **Suggestion**

Three of the four fields this method populates are never read by anyone. Grepping the whole codebase for `getMorningTimeBlock`, `getEveningTimeBlock` and `getEveningHours` returns zero hits; the only consumer of `TimeBlockDto` is line 419's `getMorningHours()`. So `buildDefaultTimeBlocks` adds three more dead setter calls to the three `processTimeBlocks` already had, and the class as a whole exists only to carry one string between two private methods in the same file. Since this code is being reworked anyway, either have both methods return the morning hours as a plain `float` and delete `TimeBlockDto`, or at minimum drop the unused fields. Note also that `TimeBlockDto` sits in `payload/request/` despite never being a request body — if it is kept, it belongs outside the payload package.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeServiceImpl.java
```diff
@@ -1314,15 +1314,21 @@ private TimeConfig createTimeConfig(TimeConfigDto.DayCapacity timeConfig) {
...
+			return mapper.valueToTree(timeConfig.timeBlocks());
+		}
 
 		Map<String, Float> hoursMap = new HashMap<>();
 		hoursMap.put(TimeBlocks.MORNING_HOURS.name(), timeConfig.totalHours() / 2);
 		hoursMap.put(TimeBlocks.EVENING_HOURS.name(), timeConfig.totalHours() / 2);
+		return createTimeConfigJsonNode(hoursMap);
```
ThinuwanW, line 1327: 🤖 **Claude** · 🟡 **Suggestion**

`createTimeConfigJsonNode` should stop being `public static` and stop allocating its own mapper. After this PR it has exactly one caller in the entire codebase — this line, in the same class — because `updateTimeConfig` no longer calls it directly; the `public static` surface is now unjustified API. It also does `new ObjectMapper()` on every invocation (line 197) while the class already injects `JsonMapper mapper` (line 152), which `resolveTimeBlocks` uses six lines above at 1319. Since `updateTimeConfigs` iterates day capacities, a request configuring all seven days on the default path allocates seven throwaway mappers. Make it a private instance method that builds the ArrayNode from the injected `mapper`.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeServiceImpl.java
```diff
@@ -1314,15 +1314,21 @@ private TimeConfig createTimeConfig(TimeConfigDto.DayCapacity timeConfig) {
...
 	}
 
-	private void updateTimeConfig(TimeConfig currentConfig, TimeConfigDto.DayCapacity timeConfig) {
-		currentConfig.setTotalHours(timeConfig.totalHours());
+	private JsonNode resolveTimeBlocks(TimeConfigDto.DayCapacity timeConfig) {
+		if (timeConfig.timeBlocks() != null && !timeConfig.timeBlocks().isEmpty()) {
+			return mapper.valueToTree(timeConfig.timeBlocks());
+		}
 
 		Map<String, Float> hoursMap = new HashMap<>();
```
ThinuwanW, line 1324: 🤖 **Claude** · 🟡 **Suggestion**

Two consistency problems in what gets persisted. (1) Non-deterministic ordering: `createTimeConfigJsonNode` iterates this `HashMap`, and the other branch's `mapper.valueToTree(timeConfig.timeBlocks())` iterates a `HashSet` (the record field is `Set<TimeBlock>`), so the stored `time_blocks` array order varies between runs. Readers index by name so it is not a correctness bug, but `updateTimeConfigs` returns the raw entities in its response, so clients see an unstable array order, and stored JSON diffs become noisy. Use a `LinkedHashMap` here and a `LinkedHashSet`/sorted ordering for the supplied set. (2) Value-type drift: this path writes `hours` as a JSON number (`block.put(HOURS, Float)` at line 203), whereas `OrganizationServiceImpl.getDefaultTimeConfigs` writes it as a JSON string (`CommonConstants.DEFAULT_TIME_CONFIG_VALUE_HOUR = "4.0"`, used at OrganizationServiceImpl.java:248). Every reader therefore has to tolerate both types — `processTimeBlocks`'s `asString()` happens to, but `TimeUtil.extractHours`'s `asFloat()` on a text node is coercion-dependent. Change the seed constant to a numeric type so all three writers emit the same shape.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeServiceImpl.java
```diff
@@ -1314,15 +1314,21 @@ private TimeConfig createTimeConfig(TimeConfigDto.DayCapacity timeConfig) {
...
 		hoursMap.put(TimeBlocks.EVENING_HOURS.name(), timeConfig.totalHours() / 2);
+		return createTimeConfigJsonNode(hoursMap);
+	}
+
+	private void updateTimeConfig(TimeConfig currentConfig, TimeConfigDto.DayCapacity timeConfig) {
+		currentConfig.setTotalHours(timeConfig.totalHours());
 
-		currentConfig.setTimeBlocks((timeConfig.timeBlocks() != null) ? (mapper.valueToTree(timeConfig.timeBlocks()))
-				: (createTimeConfigJsonNode(hoursMap)));
+		currentConfig.setTimeBlocks(resolveTimeBlocks(timeConfig));
```
ThinuwanW, line 1333: 🤖 **Claude** · 🟣 **Nit**

Unflagged behavioural change on the update path. The old `updateTimeConfig` branched on `timeConfig.timeBlocks() != null` only; `resolveTimeBlocks` adds `&& !isEmpty()`. So an admin who explicitly submits `"timeBlocks": []` previously had an empty array persisted and now silently gets `totalHours / 2` MORNING+EVENING blocks written instead. That is probably the intent, but it is a real change to an existing endpoint's semantics that no test covers and the diff does not call out — worth confirming the frontend never sends an empty array to mean "clear the blocks", and worth a `log.debug` recording that defaults were substituted so an operator can tell a defaulted config from an explicitly-configured one.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeAnalyticsServiceImpl.java
shakila-kamalasena: Fixed — after the loop, `processTimeBlocks` now falls back to `buildDefaultTimeBlocks` when no `MORNING_HOURS` block was matched (`getMorningHours() == null`), so evening-only / `EXTRA_HOURS`-only arrays no longer NPE at line 419.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeServiceImpl.java
shakila-kamalasena: Fixed with a service-layer guard in `resolveTimeBlocks`: null `totalHours()` now throws `ModuleException(TIME_ERROR_INVALID_TIME_BLOCKS)` → handled by `GlobalExceptionHandler`. Not adding `@Valid`/`@NotNull` — per our conventions we keep `@Valid` off controllers and bean-validation off DTOs; validation stays in the service layer.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeAnalyticsServiceImpl.java
shakila-kamalasena: Leaving `plusHours((long) …)` as is — the whole-hour threshold behavior is intentional; not switching to minutes.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeAnalyticsServiceImpl.java
shakila-kamalasena: Out of scope — a shared `TimeUtil` helper plus the `extractHours` null-node fix live in `TimeUtil.java`, which this PR doesn't touch. Better as a dedicated follow-up.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeAnalyticsServiceImpl.java
shakila-kamalasena: Skipping the log — keeping `buildDefaultTimeBlocks` unchanged in this PR.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeAnalyticsServiceImpl.java
shakila-kamalasena: Reused rather than removed — `TIME_ERROR_INVALID_TIME_BLOCKS` now backs the null-`totalHours` case in `resolveTimeBlocks`, so it's no longer dead code.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeAnalyticsServiceImpl.java
shakila-kamalasena: Fixed — switched to `hasNonNull(...)` so explicit JSON nulls route to the default path along with missing keys.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeAnalyticsServiceImpl.java
shakila-kamalasena: Fixed — `processTimeBlocks` now uses `TimeBlocks.MORNING_HOURS.name()`/`EVENING_HOURS.name()` and `TimeConfigFieldName.TIME_BLOCK`/`HOURS` instead of string literals, matching the writer.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeAnalyticsServiceImpl.java
shakila-kamalasena: Not adding unit tests in this PR.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeServiceImpl.java
shakila-kamalasena: Valid, but out of scope — `TimeUtil.extractHours` isn't part of this PR's two-file change. Tracking the clock-in NPE as a follow-up rather than widening this PR.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeServiceImpl.java
shakila-kamalasena: Not adding `@Valid`/`@NotNull`/`@Positive` — it conflicts with our conventions (no `@Valid` on controllers, no bean-validation on DTOs/entities; validation is service-layer). The null-`totalHours` case is now guarded in `resolveTimeBlocks`.

shakila-kamalasena: Real concern, but adding `@Transactional` to `updateTimeConfigs` and moving emails/notifications to after-commit is a pre-existing, broader change beyond this fix. Better as its own PR.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeAnalyticsServiceImpl.java
shakila-kamalasena: Keeping `totalHours` as `Float` and the existing null-safe default; not switching to a primitive.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeServiceImpl.java
shakila-kamalasena: Leaving the all-or-nothing behavior for this PR — per-block write-path validation is a separate change; the immediate crash is now handled on the read side via `buildDefaultTimeBlocks`.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeAnalyticsServiceImpl.java
shakila-kamalasena: Out of scope — `TimeBlockDto` itself isn't modified by this PR (only its callers). Collapsing it to a `float`/deleting it, or moving it out of `payload/request`, is a separate refactor.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeServiceImpl.java
shakila-kamalasena: Fixed — `createTimeConfigJsonNode` is now a `private` instance method that builds the ArrayNode from the injected `mapper` (removed `public static` and the per-call `new ObjectMapper()`).

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeServiceImpl.java
shakila-kamalasena: (1) Keeping `HashMap` — readers index blocks by name, so the stored order isn't relied on. (2) Out of scope — the `OrganizationServiceImpl` seed constant is outside this PR's files; aligning the value type across writers is a separate change.

backend/src/main/java/com/skapp/community/timeplanner/service/impl/TimeServiceImpl.java
shakila-kamalasena: Leaving as is — not adding a debug log; the `[]` → defaults behavior is intended.

erandi-rootcode (approved):

akilarootcode (approved):

PR #1820, opened 2026-08-20
[Fix] Probation end reminder recipients: super admin fallback, pending users, full name

akilarootcode:
## Summary

`Employee.getFullName()` concatenates `firstName + " " + lastName`, but `last_name` is nullable. Java string concatenation renders a null as the literal text `null`, so the method returns values like `"testoneguest null"`. This surfaced while fixing the probation end reminder email, where the value is used both in the subject line and in the greeting.

- `Employee.getFullName()` now joins only the name parts that are present, so a missing last name yields `"testoneguest"` rather than `"testoneguest null"`. A fully populated name is unchanged.

## Why it is worth fixing rather than working around

The probation change originally carried a private null-safe helper in the enterprise service. That is a near-duplicate of an existing shared method, so the helper was dropped in favour of correcting the method itself — which also fixes the other call sites.

Measured on the dev tenant:

| | count |
|---|---|
| employees with a null `last_name` | 22 |
| of those, still `ACTIVE` or `PENDING` | 11 |

| employee | before | after |
|---|---|---|
| 16 | `testoneguest null` | `testoneguest` |
| 19 | `akila.silvayop null` | `akila.silvayop` |

## Call sites affected

12 in total. Besides the probation email, the visible ones are the leave-policy bulk assign, the AI nudge reporter name, the dashboard super-admin name, the e-signature audit trail `actionVerifiedByName`, and the invoice/project admin and time-log descriptions. In every one of them the previous output for a missing last name was a literal `null` in user-facing text, so the change is a strict improvement rather than a behaviour trade-off.

## Architecture / flow

```mermaid
flowchart LR
    A[Employee.firstName] --> C{part present?}
    B[Employee.lastName] --> C
    C -- yes --> D[join with a single space]
    C -- no --> E[omit the part]
    D --> F[full name]
    E --> F
```

- **`EmployeeRepositoryImpl.findActiveEmployeesByExactNames`** matches on the SQL expression `lower(concat(concat(first_name, ' '), last_name))`, which does not drop blank parts, so the Java and SQL notions of a full name are no longer identical. No reachable failure was found — a blank `last_name` makes the SQL expression either `NULL` or a trailing-space string that `normalizeName` cannot produce on the request side — so this is latent rather than live. Aligning the two is a separate change with its own testing.
- **No escaping change.** `getFullName()` is interpolated unescaped into email subjects, and escaping it here would render an apostrophe as `&#39;` in the subject line. That is a pre-existing concern in the email layer and needs separate handling for the body and the subject.

- **Submodule hash / pointer bumps** are intentionally excluded. Only `Employee.java` is staged in this commit; gitlink updates are done at **deployment time**, not in feature PRs.

## Testing

| Check | Ran? | Result |
|-------|------|--------|
| Unit tests | see note | full suite was green at 831 tests before this change; the run after it could not be trusted, see below |
| Formatter (BE only) | yes | pass — `spring-javaformat:validate` and `checkstyle:check` both green |
| tsc + lint (BE only) | n/a | Java repo |
| e2e (Playwright) | yes | pass before this refactor — 3 tests, `emailReminderJobsQueued=6`, `sent: 6, failed: 0` |

**Note on the test run.** `mvn -o package -DskipTests` succeeds, so this compiles cleanly. A full `mvn test` immediately after could not be completed reliably on the authoring machine: the IDE Java language server holds `target/generated-sources` and respawns on kill, leaving a partial `target` and producing `NoClassDefFoundError` and Mockito mocking failures across unrelated tests — classpath corruption, not assertion failures. The last uncorrupted full run was **831 tests, 0 failures** immediately before this refactor. CI should be treated as the authority here; if it reports anything red, it is worth checking against this note before assuming a regression.

### DIFF
--- a/backend/src/main/java/com/skapp/community/peopleplanner/model/Employee.java
+++ b/backend/src/main/java/com/skapp/community/peopleplanner/model/Employee.java
@@ -35,6 +35,8 @@ import java.time.LocalDate;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 @Entity
 @Getter
@@ -186,7 +188,9 @@ public class Employee extends Auditable<String> {
 	private BusinessUnit businessUnit;
 
 	public String getFullName() {
-		return firstName + " " + lastName;
+		return Stream.of(firstName, lastName)
+			.filter(namePart -> namePart != null && !namePart.isBlank())
+			.collect(Collectors.joining(" "));
 	}
 
 }


shakila-kamalasena (approved):

ThinuwanW (commented):
## 🤖 Claude Code Review

The change to Employee.getFullName() is a genuine improvement — it stops rendering "John null" / "null null" in emails, notifications and DTOs when a name part is missing. However the fix is only half-applied: the SQL-side full-name construction that the primary consumer (bulk leave-policy assign) matches against still uses NULL-propagating CONCAT, and ~80 call sites across the codebase (including a private helper that is now a stale duplicate) still build the name with the old broken concatenation. No test was added for the new behavior.

Found **10** new issue(s): 🔴 1 important, 🟡 7 suggestion(s), 🟣 2 nit(s)

backend/src/main/java/com/skapp/community/peopleplanner/model/Employee.java
```diff
@@ -186,7 +188,9 @@ public class Employee extends Auditable<String> {
 	private BusinessUnit businessUnit;
 
 	public String getFullName() {
-		return firstName + " " + lastName;
+		return Stream.of(firstName, lastName)
```
ThinuwanW, line 191: 🤖 **Claude** · 🔴 **Important**

The fix is not carried through to the DB-side counterpart, so the one code path that actually *compares* getFullName() output still cannot match employees with a missing name part. EmployeeLeavePolicyServiceImpl.bulkAssignLeavePolicies (line 125-127) fetches candidates via EmployeeRepositoryImpl.findActiveEmployeesByExactNames, which builds the match key as `lower(concat(concat(first_name, ' '), last_name))` (EmployeeRepositoryImpl:1258-1262) and then re-groups the returned rows by `normalizeName(employee.getFullName())`. In MySQL, CONCAT returns NULL if any argument is NULL, so an employee with a NULL last_name (both columns are `DEFAULT NULL` per common-ddl-script-v1-create-tables.sql:128/133) produces a NULL expression, `NULL IN (:names)` is never true, and that employee is silently excluded from the result set — the caller then reports LEAVE_ERROR_BULK_EMPLOYEE_NOT_FOUND. The new Java-side semantics ('John' instead of 'John null') therefore never get exercised for exactly the rows this PR is trying to fix. Fix: make the criteria expression null-tolerant and whitespace-trimmed so it mirrors the new getFullName(), e.g. wrap each part in `criteriaBuilder.coalesce(root.get(Employee_.FIRST_NAME), "")` and apply `criteriaBuilder.trim(...)` around the concatenation (or build it with a nullif/case so a blank part doesn't leave a stray separator). The same NULL-propagating pattern is used in the employee search predicates at EmployeeRepositoryImpl:1147-1148, 1231-1232 and 1522 and in EpExternalSyncStagingRepositoryImpl:136, LeaveEntitlementRepositoryImpl:253-255/554/701/836, LeaveRequestRepositoryImpl:166/1193, PolicyLeaveRequestRepositoryImpl:259, BillableRateRepositoryImpl:84, CrmDealRepositoryImpl:180/231 and CrmContactOwnerRepositoryImpl:92-94 — those employees are invisible to full-name search for the same reason and should be aligned in the same pass or a tracked follow-up.

backend/src/main/java/com/skapp/community/peopleplanner/model/Employee.java
```diff
@@ -35,6 +35,8 @@
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
+import java.util.stream.Collectors;
```
ThinuwanW, line 38: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Fixing only the entity method leaves the codebase with two divergent definitions of an employee's full name. Roughly 80 call sites still inline `getFirstName() + " " + getLastName()` and will keep emitting 'John null' / 'null null'. The clearest case is PolicyLeaveServiceImpl:1097-1099, a private helper literally named `fullName(Employee)` whose body is the exact pre-PR implementation — it should simply delegate to `employee.getFullName()` and ideally be deleted. Other high-traffic clusters that feed user-visible email and notification bodies: TimeEmailServiceImpl (lines 40-336), AttendanceNotificationServiceImpl (52-269), PeopleEmailServiceImpl (47-278), PeopleNotificationServiceImpl (117-181), LeaveEmailServiceImpl (64-184), LeaveNotificationServiceImpl (151-391), EpCommonEmailEmailServiceImpl (23/34/63), EpEmployeeTimelineServiceImpl (494/525) and ReleaseServiceImpl:231. Replacing these with `getFullName()` is a mechanical change and is the only way the fix actually reaches most of the places users see the broken name.

backend/src/main/java/com/skapp/community/peopleplanner/model/Employee.java
```diff
@@ -186,7 +188,9 @@ public class Employee extends Auditable<String> {
 	private BusinessUnit businessUnit;
 
 	public String getFullName() {
-		return firstName + " " + lastName;
+		return Stream.of(firstName, lastName)
+			.filter(namePart -> namePart != null && !namePart.isBlank())
```
ThinuwanW, line 192: 🤖 **Claude** · 🟡 **Suggestion**

The predicate `namePart != null && !namePart.isBlank()` re-implements the existing shared utility `com.skapp.community.common.util.StringUtils.isNullOrBlank(String)` (StringUtils.java:24-26), which is used elsewhere in this flow (e.g. EmployeeLeavePolicyServiceImpl:281). Prefer `.filter(namePart -> !StringUtils.isNullOrBlank(namePart))` so the null/blank definition stays in one place. Note this getter is called inside per-row loops and mappers, so if you'd rather avoid the Stream/Collectors allocation entirely, `String.join(" ", ...)` over a filtered two-element check reads just as clearly and lets you drop the two new imports.

backend/src/main/java/com/skapp/community/peopleplanner/model/Employee.java
```diff
@@ -186,7 +188,9 @@ public class Employee extends Auditable<String> {
 	private BusinessUnit businessUnit;
 
 	public String getFullName() {
```
ThinuwanW, line 190: 🤖 **Claude** · 🟡 **Suggestion**

No test covers the new behavior — grepping backend/src/test for `getFullName`/`fullName` returns no matches, so nothing pins the four cases this method now distinguishes. Add a small unit test asserting: firstName='John', lastName=null → 'John'; firstName=null, lastName='Doe' → 'Doe'; both blank/null → ''; both present → 'John Doe'. Without it, a future refactor can silently reintroduce the 'John null' output this PR removed.

backend/src/main/java/com/skapp/community/peopleplanner/model/Employee.java
```diff
@@ -186,7 +188,9 @@ public class Employee extends Auditable<String> {
...
-		return firstName + " " + lastName;
+		return Stream.of(firstName, lastName)
+			.filter(namePart -> namePart != null && !namePart.isBlank())
+			.collect(Collectors.joining(" "));
```
ThinuwanW, line 193: 🤖 **Claude** · 🟣 **Nit**

The method can now return an empty string where it previously always returned something non-empty. Callers that interpolate it directly into user-facing text will render a blank instead of an obviously-wrong placeholder: EpProbationInternalServiceImpl:141/162/165 (reminder email recipient and employee names), ProjectServiceImpl:460 (invoice line description) and 513 (project admin name), DashboardServiceImpl:60 (superAdminName). This is still better than 'null null', but an email reading 'Hi ,' is also broken and harder to notice in monitoring. AiNudgeServiceImpl:55 already models the right pattern — it falls back to `reporter.getUser().getEmail()` for guest employees. Consider applying a similar fallback (work email, or a message-constant placeholder) at the display sites where a blank name would be visible, rather than pushing the fallback into the entity.

backend/src/main/java/com/skapp/community/peopleplanner/model/Employee.java
ThinuwanW, line 192: 🤖 **Claude** · 🟡 **Suggestion**

The filter drops parts that are entirely blank but does not trim parts that are merely padded. `firstName = "John "` (the `first_name` column is a plain `varchar(50)` with no trimming on write, and `Validations.validateName` does not strip) with `lastName = "Smith"` now yields `"John  Smith"` with a double space, and `firstName = " John"` yields a leading space. Since the point of this change is to produce a clean canonical name, normalise the parts as well as filter them: insert `.map(String::strip)` before the `.filter(...)` (or filter on `StringUtils.isNullOrBlank` after stripping). Note that `StringUtils.normalizeName` already collapses whitespace on the Java side, so the untrimmed output is currently masked in the bulk-assign matcher but is visible verbatim in every email/notification consumer.

backend/src/main/java/com/skapp/community/peopleplanner/model/Employee.java
ThinuwanW, line 38: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Two MapStruct mappers derive public API response fields from this method, so the change silently alters API output and introduces an inconsistent representation of "no name". `CrmMapper` maps `@Mapping(target = "ownerName", source = "owner.fullName")` on `crmTaskToCrmTaskResponseDto`, which resolves to `getFullName()`; `EpPeopleMapper` maps `recordedBy` via `employeeTimeline.getRecordedBy() != null ? ...getFullName() : null`. After this change the same field can be `null` (no recorder) or `""` (recorder with blank names), which clients must now handle as two distinct "unknown" encodings. Pick one convention — either have `getFullName()` return `null` when nothing is left to join, or have both mappers coalesce to `""` — and add a mapper/API-level assertion for the blank-name case so the response contract is pinned, since neither mapper reads as a consumer of this method at a glance.

backend/src/main/java/com/skapp/community/peopleplanner/model/Employee.java
ThinuwanW, line 190: 🤖 **Claude** · 🟡 **Suggestion**

The method is being formalised as the canonical full-name builder but still ignores `middleName`, which is persisted on this same entity (`@Column(name = "middle_name")`) and is surfaced separately by `PeopleReadServiceImpl`, `EmployeeExportMapperService` and `PeopleServiceImplV2`. As written the method is `getFirstAndLastName()`, not `getFullName()`, and the stream shape invites a future contributor to add `middleName` to `Stream.of(...)` and change every email, CRM `ownerName` and timeline `recordedBy` value at once. Decide the contract explicitly and record it: add a Javadoc (matching the style of `StringUtils.isNullOrBlank`/`removeNullCharacters`, which document their null behaviour) stating that middle name is deliberately excluded and that the method returns an empty string — never null — when no name part is present.

backend/src/main/java/com/skapp/community/peopleplanner/model/Employee.java
ThinuwanW, line 38: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Null-safety gap at a call site this change is meant to harden: `DashboardServiceImpl.getDashboardEmailOrganizationDetails` does `superAdmin.get().getEmployee().getFullName()` inside an `isPresent()` guard that only proves the `User` exists. `User.employee` is the inverse side of the `@MapsId` one-to-one and is not guaranteed non-null, so a user row without an employee record throws an NPE that reaches `GlobalExceptionHandler` as an opaque 500 on the dashboard email path. Making the name itself null-safe does not help if the dereference before it can fail. Rewrite as `superAdmin.map(User::getEmployee).map(Employee::getFullName).orElse("")`, which also removes the mutable `superAdminName` local.

backend/src/main/java/com/skapp/community/peopleplanner/model/Employee.java
ThinuwanW, line 191: 🤖 **Claude** · 🟣 **Nit**

A `Stream` plus a `Collector` (and two new imports) is heavy machinery for joining two fields, and this method is called per row in hot paths — `ProjectMapper.memberBillableRateListToProjectMembersResponseDto` maps it for every project member, `EmployeeLeavePolicyServiceImpl` calls it inside a `Collectors.groupingBy` classifier over every matched employee, and `EpProbationInternalServiceImpl.buildReminderJobs` calls it once per recipient in a nested loop. Each invocation allocates a stream pipeline, a `StringJoiner` and a lambda capture for a two-element join. A plain `Stream.of(...).filter(...).toList()` fed to `String.join(" ", parts)`, or a straightforward conditional, reads more directly and allocates less. Minor, but worth taking while the method is already being touched.

PR #1821, opened 2026-08-21
Fix/998 Contacts Accessibility Issues

umeshaliyanage:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds an `aria-label` to the react-phone-input-2 country dropdown listbox and wires the new `commonAria.components.inputPhoneNumber.countryList` translation key into the existing DOM-mutation effect. The change is small, type-safe, correctly i18n'd (key path resolves, JSON is valid, English is the only locale), and the added `useEffect` dependency is correct. My findings are about the surrounding accessibility machinery the change plugs into — the label value duplicates the combobox announcement, the DOM queries are document-scoped, and the listbox still isn't programmatically associated with the input that carries `aria-activedescendant`.

Found **12** new issue(s): 🔴 1 important, 🟡 9 suggestion(s), 🟣 2 nit(s)

frontend/src/community/common/assets/languages/english/aria/common.json
```diff
@@ -74,7 +74,8 @@
     "inputPhoneNumber": {
       "countryCode": "country code",
       "phoneNumber": "phone number",
-      "countryCodeDescription": "Country code +{{countryCode}}"
+      "countryCodeDescription": "Country code +{{countryCode}}",
+      "countryList": "Country code"
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The new `countryList` key resolves to "Country code", which is essentially the same string as the sibling `countryCode` key ("country code") already used for the combobox `aria-label` at InputPhoneNumber.tsx:159/166. A screen reader user will hear "<label> country code, combobox" and then "Country code, listbox" — the popup announcement adds no information and reads as a repeat. The element being labelled is a list of countries (name + dial code), not a code field. Suggest a value that describes the list, e.g. "Country list" or "Select a country", which also makes the key name and value agree. Compare the convention used by sibling keys in this file: `itemSelector.popperLabel: "Item Selector"`, `multiSelectManagerSearch.managerOptions: "Other supervisors options"`.

frontend/src/community/common/components/molecules/InputPhoneNumber/InputPhoneNumber.tsx
```diff
@@ -87,6 +89,7 @@ const InputPhoneNumber: FC<Props> = ({
 
       if (list) {
         list.setAttribute("role", "listbox");
+        list.setAttribute("aria-label", countryListLabel);
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The new `aria-label` is applied via `document.querySelector(".country-list")` (line 87), which is document-wide rather than scoped to this component instance. Multiple `InputPhoneNumber` instances are mounted on the same page in practice — `PrimaryContactDetailsSection` and `SecondaryContactDetailsSection` both render `EmergencyContactDetailsSection` → `ContactDetailsFormSection` → `InputPhoneNumber` on the emergency-details page. Every mounted instance runs its own 300ms interval against the same first-matching node, so instance B's interval also rewrites the attributes on instance A's dropdown. The `aria-label` itself happens to be identical across instances so it is not corrupted, but `aria-selected` (line 107) is derived from each instance's own `countryCodeValue` and therefore does get cross-written. Fix: attach a `ref` to a wrapper element around `<PhoneInput>` (or use `phoneInputRef.current` container) and run the queries as `container.querySelector(...)` / `container.querySelectorAll(...)` instead of off `document`.

frontend/src/community/common/components/molecules/InputPhoneNumber/InputPhoneNumber.tsx
```diff
@@ -87,6 +89,7 @@ const InputPhoneNumber: FC<Props> = ({
 
       if (list) {
         list.setAttribute("role", "listbox");
```
ThinuwanW, line 112: 🤖 **Claude** · 🟡 **Suggestion**

Naming the listbox is only half of the combobox pattern — nothing associates the popup with the controls that reference it. The `.country-list` element gets `role="listbox"` and now an `aria-label`, but it is never given an `id`, so (a) the combobox input (`inputProps` at lines 167-171) has `aria-haspopup="listbox"`/`aria-expanded` but no `aria-controls`/`aria-owns` pointing at it, and (b) the `aria-activedescendant` set on `.flag-dropdown input` at line 117 references option ids on an element that input neither owns nor controls, so assistive tech may ignore it entirely. Consider assigning a stable id in the same block (e.g. `list.setAttribute("id", listId)` from a `useId()`), then adding `aria-controls: listId` to `inputProps` and `input.setAttribute("aria-controls", listId)` alongside the existing `aria-activedescendant` call. Separately, note that `enableSearch` (line 205) makes react-phone-input-2 render a `<li class="search">` as the first child of `.country-list`; a `listbox` whose children include a non-`option` search field is invalid ARIA and some screen readers will not expose that search input. Applying `role="listbox"` to only the options container (or using `aria-owns` listing the `country-option-*` ids) avoids that.

frontend/src/community/common/components/molecules/InputPhoneNumber/InputPhoneNumber.tsx
```diff
@@ -8,14 +8,14 @@ import Tooltip from "~community/common/components/atoms/Tooltip/Tooltip";
 import { ZIndexEnums } from "~community/common/enums/CommonEnums";
 import { useTranslator } from "~community/common/hooks/useTranslator";
 import { phoneNumberPattern } from "~community/common/regex/regexPatterns";
+import { getPhoneNumberMaxLength } from "~community/common/utils/commonUtil";
```
ThinuwanW, line 18: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

`aria-expanded` (lines 168-170) is computed from `phoneInputRef.current?.state.open`, a ref read during render. Opening the dropdown mutates react-phone-input-2's internal state but never triggers a re-render of `InputPhoneNumber`, and `PhoneInput` re-renders with the same already-captured `inputProps` object, so `aria-expanded` stays `"false"` for the entire lifetime of the component. This directly undercuts the accessibility improvement in this PR: a screen reader user is told the combobox is collapsed even while the newly-labelled listbox is open. Since this file already mutates DOM attributes on an interval, the simplest fix is to set it there too — inside `handleDropdownAccessibility`, do `input.setAttribute("aria-expanded", list ? "true" : "false")` on the combobox input — or hold the open state in `useState` and update it from `handleCountryKeyDown` plus the dropdown's own open/close events.

frontend/src/community/common/components/molecules/InputPhoneNumber/InputPhoneNumber.tsx
```diff
@@ -117,7 +120,7 @@ const InputPhoneNumber: FC<Props> = ({
 
     const interval = setInterval(handleDropdownAccessibility, 300);
     return () => clearInterval(interval);
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

This PR adds a dependency to an effect that runs `setInterval(..., 300)` for the whole lifetime of the component, performing two `querySelectorAll`/`querySelector` sweeps of the entire document plus a `forEach` over ~240 country nodes, every 300ms, even when the dropdown is closed or the field is disabled. With several phone inputs on a page (see the emergency-details case above) that multiplies. Since you're touching this effect anyway, consider replacing the polling with a `MutationObserver` on the `.flag-dropdown` container (react-phone-input-2 mounts/unmounts the list on open) or driving the attribute application from the dropdown open state, and bail out early when `list` is null.

frontend/src/community/common/components/molecules/InputPhoneNumber/InputPhoneNumber.tsx
ThinuwanW, line 18: 🤖 **Claude** · 🔴 **Important**

📁 **File-level observation**

`aria-activedescendant` is written to the wrong element (lines 111-118). `document.querySelector(".flag-dropdown input")` resolves to react-phone-input-2's *search box* (rendered inside `.country-list`, itself inside `.flag-dropdown`, and only present while the dropdown is open because `enableSearch` is set) — not to the readonly country-code input that this PR's sibling code marks with `role="combobox"` / `aria-haspopup="listbox"` (lines 167-171). That input renders as `.form-control`, a sibling of `.flag-dropdown`, so it never receives `aria-activedescendant`. Net effect: the combobox that now points at a freshly-named listbox still never announces which option is active, and when the dropdown is closed the selector matches nothing so the branch silently no-ops. Fix: capture the combobox input via a ref (or scope with a container ref and query `.form-control`) and set `aria-activedescendant` there, clearing it when the dropdown closes.

frontend/src/community/common/components/molecules/InputPhoneNumber/InputPhoneNumber.tsx
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The element now being given an accessible name is not a valid listbox. Because `enableSearch` is enabled (line 205), react-phone-input-2 renders the search textbox as a child `<li class="search">` of the same `.country-list` UL that gets `role="listbox"` + the new `aria-label`. A `listbox` may only own `option`/`group` children, so naming it makes screen readers announce a listbox whose first child is an unowned textbox — the search field can be skipped or mis-reported in browse mode. Fix: either mark the search row as non-content (`li.search` → `role="presentation"` and leave the input outside the listbox), or move `role="listbox"`/`aria-label` onto a wrapper that contains only the `.country` items rather than the whole UL.

frontend/src/community/common/components/molecules/InputPhoneNumber/InputPhoneNumber.tsx
```diff
@@ -117,7 +120,7 @@ const InputPhoneNumber: FC<Props> = ({
...
-  }, [countryCodeValue]);
+  }, [countryCodeValue, countryListLabel]);
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The new label is applied on a polling tick, so there is a race on open: the dropdown UL is created the moment the user activates the flag button, but `role`/`aria-label` are not written until the next interval fire — up to 300ms later. Screen readers announce the popup as soon as it appears in the accessibility tree, so in practice the listbox will frequently be announced *unnamed*, which is exactly what this PR is trying to fix. Fix: apply the attributes synchronously at the point the dropdown is opened (in `handleCountryKeyDown` after `setOpen(true)`, and on the button's click handler), or replace the interval with a `MutationObserver` scoped to a container ref that reacts the instant `.country-list` is inserted.

frontend/src/community/common/components/molecules/InputPhoneNumber/InputPhoneNumber.tsx
```diff
@@ -63,6 +63,8 @@ const InputPhoneNumber: FC<Props> = ({
   const theme: Theme = useTheme();
   const phoneInputRef = useRef<any>(null);
 
+  const countryListLabel = translateText(["countryList"]);
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

This PR hoists one translation into a named const but leaves the two other accessible names duplicated inline. `${ariaLabel ? ariaLabel : label} ${translateText(["countryCode"])}` is built twice — once for the `role="group"` wrapper (line 159) and once for the combobox `inputProps` (line 166) — so the group and its own child expose the *identical* accessible name and a screen reader announces the same phrase twice on entry. Fix: hoist it alongside `countryListLabel` (e.g. `const countryCodeLabel = ...`), and give the group a distinguishing name (or drop `role="group"`/its label entirely, since a two-control group rarely needs one) rather than repeating the field name verbatim.

frontend/src/community/common/components/molecules/InputPhoneNumber/InputPhoneNumber.tsx
```diff
@@ -63,6 +63,8 @@ const InputPhoneNumber: FC<Props> = ({
   const theme: Theme = useTheme();
   const phoneInputRef = useRef<any>(null);
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The effect this PR modifies is built entirely on `any`, which is what lets the new code type-check without any guarantees. `useRef<any>(null)` (line 64) is what makes `phoneInputRef.current?.state.open` and `.setOpen()` compile, and `options.forEach((el: any, index) => ...)` (line 95) discards the `Element` type so the `el?.querySelector(...)` / `el.setAttribute(...)` calls next to the new label are unchecked. Repo guidelines forbid `any` under strict mode. Fix: type the ref against the library instance (`useRef<PhoneInput | null>(null)` or a minimal local interface `{ state: { open: boolean }; setOpen: (open: boolean) => void }`) and type the iteration as `(el: HTMLElement, index: number)` since `querySelectorAll` already yields elements.

frontend/src/community/common/components/molecules/InputPhoneNumber/InputPhoneNumber.tsx
ThinuwanW, line 18: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

No test coverage accompanies this change. Sibling molecules in the same directory all ship a colocated test (`InputField.test.tsx`, `Dropdown.test.tsx`, `DropDownSearch.test.tsx`, …), but `InputPhoneNumber` has no `.test.tsx` at all — so this imperatively-applied ARIA wiring, which depends on undocumented internal class names of a third-party library (`.country-list`, `.country`, `.dial-code`, `.flag-dropdown input`), will break silently on any `react-phone-input-2` upgrade. Fix: add `InputPhoneNumber.test.tsx` with the `MockTheme` wrapper, open the dropdown, advance timers past the 300ms interval, and assert `screen.getByRole("listbox", { name: ... })` plus the option roles/`aria-selected` state.

frontend/src/community/common/assets/languages/english/aria/common.json
```diff
@@ -74,7 +74,8 @@
     "inputPhoneNumber": {
       "countryCode": "country code",
       "phoneNumber": "phone number",
-      "countryCodeDescription": "Country code +{{countryCode}}"
+      "countryCodeDescription": "Country code +{{countryCode}}",
```
ThinuwanW, line 77: 🤖 **Claude** · 🟣 **Nit**

The `inputPhoneNumber` aria block is accumulating dead keys. A codebase-wide grep shows neither `phoneNumber` nor `countryCodeDescription` (the line this PR edits to append a comma) is referenced anywhere — only `countryCode` and the newly added `countryList` are consumed. `countryCodeDescription` ("Country code +{{countryCode}}") in particular looks like it was intended for exactly this kind of dropdown/selection labelling, which suggests the new key may be re-adding something that already exists unused. Fix: delete the two unreferenced keys in the same PR, or wire `countryCodeDescription` up instead of introducing a third near-identical string.

umeshaliyanage: both countryCodeDescription and phoneNumber are existing code and don't belong to my changeset

frontend/src/community/common/assets/languages/english/aria/common.json
umeshaliyanage: renamed to "Country List"

frontend/src/community/common/components/molecules/InputPhoneNumber/InputPhoneNumber.tsx
umeshaliyanage: doesn't belong to my changeset

frontend/src/community/common/components/molecules/InputPhoneNumber/InputPhoneNumber.tsx
umeshaliyanage: put  a const countryCodeLabel = `${ariaLabel ? ariaLabel : label} ${translateText(["countryCode"])}`;

frontend/src/community/common/components/molecules/InputPhoneNumber/InputPhoneNumber.tsx
umeshaliyanage: No test wriiten for ui

umeshaliyanage: gated in input  const input = container.querySelector(".form-control");
 if (input) {
input.setAttribute("aria-expanded", list ? "true" : "false");
 150 is inside handleDropdownAccessibility, 211 is in inputProps

umeshaliyanage: Fixed. Now queries .form-control (the input actually carrying role="combobox") via a container ref, and removes the attribute when the dropdown closes

const selected = list ? container.querySelector<HTMLElement>(".country.highlight") : null;
 const selectedIndex = selected ? Array.from(options).indexOf(selected) : -1;
 if (selectedIndex >= 0) {
input.setAttribute("aria-activedescendant", `${listboxId}-option-${selectedIndex}`);

frontend/src/community/common/components/molecules/InputPhoneNumber/InputPhoneNumber.tsx
umeshaliyanage: Switched to a MutationObserver scoped to the container ref

frontend/src/community/common/components/molecules/InputPhoneNumber/InputPhoneNumber.tsx
umeshaliyanage: Done id set on the listbox from a useId(), and "aria-controls": listboxId added to inputProps 
Also moved aria-activedescendant to .form-control 
Deferring enableSearch: that li is in the library's React tree, so reparenting risks a removeChild crash. role="group" meanwhile the only listbox-permitted role that's honest about a wrapped text input.

umeshaliyanage: Done added containerRef on the <Stack> wrapping <PhoneInput>; every document.querySelector/querySelectorAll in the effect is now container.*, with an early return if the ref is null
removed interval

frontend/src/community/common/components/molecules/InputPhoneNumber/InputPhoneNumber.tsx
umeshaliyanage: Done — id set on the listbox from a useId(), and "aria-controls": listboxId added to inputProps (static, so no imperative set needed).
Also moved aria-activedescendant to .form-control — .flag-dropdown input was the library's search box, not the combobox — and namespaced option ids as `${listboxId}-option-${i}`, since two instances mount here and it resolves document-wide

frontend/src/community/common/components/molecules/InputPhoneNumber/InputPhoneNumber.tsx
```diff
@@ -61,7 +73,13 @@ const InputPhoneNumber: FC<Props> = ({
     "inputPhoneNumber"
   );
   const theme: Theme = useTheme();
-  const phoneInputRef = useRef<any>(null);
+  const phoneInputRef = useRef<PhoneInputInstance | null>(null);
+  const containerRef = useRef<HTMLDivElement>(null);
+  const listboxId = useId();
+
+  const countryListLabel = translateText(["countryList"]);
```
hasalarootcode: dont create seperate variables for translations, use it directly

frontend/src/community/common/components/molecules/InputPhoneNumber/InputPhoneNumber.tsx
```diff
@@ -61,7 +73,13 @@ const InputPhoneNumber: FC<Props> = ({
...
+  const countrySearchLabel = translateText(["countrySearch"]);
+  const countryCodeLabel = `${ariaLabel ? ariaLabel : label} ${translateText(["countryCode"])}`;
```
hasalarootcode: Fix the above sonar issue

hasalarootcode (changes requested):

frontend/src/community/common/components/molecules/InputPhoneNumber/InputPhoneNumber.tsx
umeshaliyanage: used directly

frontend/src/community/common/components/molecules/InputPhoneNumber/InputPhoneNumber.tsx
umeshaliyanage: fixed

hasalarootcode (approved):

frontend/src/community/common/components/molecules/InputPhoneNumber/InputPhoneNumber.tsx
umeshaliyanage: Both fixed. The ref is typed against a minimal local interface rather than the library's export

hasalarootcode (approved):

ThinuwanW (approved):

PR #1822, opened 2026-08-21
Fix/1004 inbox accessibility issues

umeshaliyanage:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR improves DropdownList accessibility by adding `aria-invalid`/`aria-describedby` to both Select branches, naming the popup listbox via `MenuListProps`, and hiding the decorative required asterisk from assistive tech. The direction is right, but two issues will prevent it from working as intended in production: MUI's `Select` does not forward arbitrary `aria-*` keys from `inputProps` to the `role="combobox"` element (they land on the `aria-hidden` native input), and the new `errorId` is derived from `inputName`, which is demonstrably non-unique on several existing screens. No tests were added to cover any of the new attributes.

Found **16** new issue(s): 🔴 3 important, 🟡 7 suggestion(s), 🟣 6 nit(s)

frontend/src/community/common/components/molecules/DropdownList/DropdownList.tsx
```diff
@@ -176,16 +189,19 @@ const DropdownList: FC<Props> = ({
...
             sx={{
               ...classes.selectStyle(theme, isDisabled, readOnly as boolean),
               ...selectStyles
             }}
             fullWidth
             inputProps={{
-              "aria-label": ariaLabel || label,
-              "aria-required": required
+              "aria-label": accessibleName,
+              "aria-invalid": !!error,
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`aria-invalid` (and the pre-existing `aria-required`) passed through `inputProps` almost certainly never reach the element a screen reader interacts with. MUI v6's `SelectInput` explicitly destructures only `'aria-label'` and `'aria-describedby'` and applies them to the visible `role="combobox"` div; every other unrecognised prop falls into `...other`, which is spread onto the *hidden* `<input aria-hidden className={classes.nativeInput}>`. Attributes on an `aria-hidden` node are ignored by AT, so this change is a no-op for `aria-invalid`/`aria-required` while `aria-describedby` does work — an inconsistent half-fix.

Fix: either pass MUI's own props on `<Select>` (`error={!!error}` and `required={required}`, which SelectInput maps onto the combobox), or set them explicitly on the display element:
`SelectDisplayProps={{ "aria-invalid": !!error, "aria-required": required ? "true" : undefined }}`
Apply the same change to the second `<Select>` (line 322). Please verify with a DOM assertion, e.g. `expect(screen.getByRole("combobox")).toHaveAttribute("aria-invalid", "true")` — that assertion is what distinguishes the working fix from the current one.

frontend/src/community/common/components/molecules/DropdownList/DropdownList.tsx
```diff
@@ -104,6 +104,14 @@ const DropdownList: FC<Props> = ({
   const theme: Theme = useTheme();
   const classes = styles(theme);
 
+  const accessibleName = ariaLabel || label;
+
+  const errorId = error ? `${inputName}-error` : undefined;
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`errorId = `${inputName}-error`` is not unique. `inputName` is a form-field name, not a DOM-unique identifier, and it repeats on real screens:
- `EmergencyDetailsForm.tsx` renders `PrimaryDetailsSection` and `SecondaryDetailsSection` side by side, both with `<DropdownList inputName="relationship" error={errors.relationship} />` (AddNewResourceFlow), and the edit flow does the same via `PrimaryContactDetailsSection` + `SecondaryContactDetailsSection` -> `ContactDetailsFormSection` (`inputName="relationship"`).
- `ProjectMembersTable.tsx:177` renders `<DropdownList inputName="billableFrequency">` once per row inside a `.map()`.

With duplicate ids, `aria-describedby="relationship-error"` on the secondary contact's combobox resolves to the *primary* contact's error node, so the wrong message is announced (and the HTML is invalid).

Fix: seed the id from React's `useId()` so it is unique per instance:
`const reactId = useId();`
`const errorId = error ? `${id ?? inputName}-${reactId}-error` : undefined;`

frontend/src/community/common/components/molecules/DropdownList/DropdownList.tsx
```diff
@@ -128,7 +136,12 @@ const DropdownList: FC<Props> = ({
           lineHeight={1.5}
           sx={{ ...classes.labelStyle(isDisabled, !!error), ...labelStyles }}
         >
-          {label} {required && <span style={{ color: "red" }}>*</span>}
+          {label}{" "}
+          {required && (
+            <span aria-hidden="true" style={{ color: "red" }}>
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Hiding the asterisk with `aria-hidden="true"` is only correct if the required state is conveyed some other way — and per the `aria-required` finding above it currently is not. Combined, this PR makes required-ness *less* discoverable than before: the combobox's accessible name is `aria-label={accessibleName}` (which never contains the asterisk), and a browse-mode user reading the `<label>` text will no longer hear the `*` either. Please land the `SelectDisplayProps`/`required` fix in the same change, or keep the asterisk readable until it does.

frontend/src/community/common/components/molecules/DropdownList/DropdownList.tsx
```diff
@@ -104,6 +104,14 @@ const DropdownList: FC<Props> = ({
   const theme: Theme = useTheme();
   const classes = styles(theme);
 
+  const accessibleName = ariaLabel || label;
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

No test coverage for any of the new behaviour, even though `DropdownList.test.tsx` already exists next to the component. Please add assertions for: (1) `getByRole("combobox")` has `aria-describedby` equal to the rendered error node's `id` when `error` is set, and no `aria-describedby` when it isn't; (2) `aria-invalid`/`aria-required` are present on the combobox (this is the assertion that would have caught the `inputProps` forwarding problem); (3) opening the menu yields a listbox with the expected accessible name (`getByRole("listbox", { name: ... })`). Note also that the existing `expect(screen.getByText("*")).toBeInTheDocument()` at line 42 still passes despite the new `aria-hidden`, because `getByText` does not filter hidden nodes — so the current suite gives no signal about this change at all.

frontend/src/community/common/components/molecules/DropdownList/DropdownList.tsx
```diff
@@ -128,7 +136,12 @@ const DropdownList: FC<Props> = ({
           lineHeight={1.5}
           sx={{ ...classes.labelStyle(isDisabled, !!error), ...labelStyles }}
         >
-          {label} {required && <span style={{ color: "red" }}>*</span>}
+          {label}{" "}
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

While wiring up accessibility here, the `<Typography component="label">` is still an orphan `<label>`: it has no `htmlFor` and no `id`, so clicking the label does not focus/open the dropdown and the visible text is not programmatically associated with the control. Consider giving the label an id (derived from the same `useId()` suggested above) and adding `aria-labelledby` on the Select via `SelectDisplayProps`, or dropping to a plain `Typography` if a real `<label>` association is not intended — an empty `<label>` element is misleading markup.

frontend/src/community/common/components/molecules/DropdownList/DropdownList.tsx
```diff
@@ -104,6 +104,14 @@ const DropdownList: FC<Props> = ({
   const theme: Theme = useTheme();
   const classes = styles(theme);
 
+  const accessibleName = ariaLabel || label;
+
+  const errorId = error ? `${inputName}-error` : undefined;
+
+  const menuListProps = accessibleName
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

The `menuListProps` ternary is unnecessary and weakens typing: it infers the union `{ "aria-label": string } | {}`, which is implicit rather than explicit (the guidelines require explicit types). React/MUI already omit attributes whose value is `undefined`, so this collapses to a single explicitly-typed object:
`const menuListProps: Partial<MenuListProps> = { "aria-label": accessibleName };`
(importing `MenuListProps` from `@mui/material`).

frontend/src/community/common/components/molecules/DropdownList/DropdownList.tsx
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

This exact error block (`<Typography role="alert" aria-live="assertive">{error}</Typography>`) is duplicated across at least a dozen sibling molecules — `MultivalueDropdownList`, `DropDownSearch`, `DropdownAutocomplete`, `InputDate`, `DurationSelector`, `ColorInputField`, `EmojiPicker`, `PeopleSearch`, `Search`, etc. — none of which get the new `id`/`aria-describedby` wiring. Fixing only `DropdownList` guarantees drift. Consider extracting a shared `FieldErrorText` molecule that owns the id generation and returns it for `aria-describedby`, then adopting it here first and in the siblings as a follow-up.

frontend/src/community/common/components/molecules/DropdownList/DropdownList.tsx
```diff
@@ -292,6 +308,7 @@ const DropdownList: FC<Props> = ({
             name={inputName}
             disabled={isDisabled}
             multiple={isMultiValue}
+            MenuProps={{ MenuListProps: menuListProps }}
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Adding `MenuListProps` here names a listbox that contains no options — this branch renders either a `CircularProgress` inside a `Box` or a bare `Typography` (`noOptionsText`) as direct children of the `role="listbox"` `<ul>`, which is invalid listbox content. Naming it makes it easier to reach but still announces as an empty listbox. Consider `aria-busy="true"` on the loading branch and rendering `noOptionsText` as a `<MenuItem disabled>` so it is exposed as content of the listbox rather than as stray markup.

frontend/src/community/common/components/molecules/DropdownList/DropdownList.tsx
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`accessibleName` has no fallback and can legitimately be `undefined`. Both `label` and `ariaLabel` are optional props, and many call sites pass `ariaLabel={translateText([...]) ?? ""}` — which yields an empty string when the key is missing, so `||` falls through to `label`, which may itself be undefined. When that happens this PR produces the worst case: `aria-label` is dropped from the combobox, AND `menuListProps` collapses to `{}` so the popup gets `role="listbox"` with no accessible name either. The only remaining visible text is the placeholder Typography, which is explicitly `aria-hidden={true}` (line 231), and MUI's notched-outline legend (fed by `label={placeholder}` on line 170) sits inside an `aria-hidden` fieldset. Net result: a completely unnamed combobox and an unnamed listbox. Fix: chain a real fallback, e.g. `const accessibleName = ariaLabel || label || placeholder;`, and consider making the prop contract explicit by typing Props as a union that requires at least one of `label` / `ariaLabel`. At minimum add a dev-only warning when none of the three is present, so the gap is visible during development rather than only in an audit.

frontend/src/community/common/components/molecules/DropdownList/DropdownList.tsx
```diff
@@ -320,6 +339,7 @@ const DropdownList: FC<Props> = ({
 
       {!!error && (
         <Typography
+          id={errorId}
```
ThinuwanW, line 350: 🤖 **Claude** · 🟡 **Suggestion**

The error node is now referenced by `aria-describedby` while simultaneously being an assertive live region (`role="alert"`, which already implies `aria-live="assertive"` plus `aria-atomic="true"`, and the explicit `aria-live="assertive"` is therefore redundant). Combining the two means NVDA/JAWS interrupt to announce the message the moment it renders, and then announce the identical text a second time as the field's description on focus / on every re-focus. Pick one channel: either keep `id` + `aria-describedby` and drop `role="alert"`/`aria-live` (the description model, which is what Formik-style on-blur validation usually wants), or keep the live region and drop the `aria-describedby` link. If you keep both intentionally, at least remove the redundant `aria-live="assertive"` and downgrade to `role="status"` so the announcement is polite rather than interrupting.

frontend/src/community/common/components/molecules/DropdownList/DropdownList.tsx
```diff
@@ -176,16 +189,19 @@ const DropdownList: FC<Props> = ({
...
             }}
             sx={{
               ...classes.selectStyle(theme, isDisabled, readOnly as boolean),
               ...selectStyles
             }}
             fullWidth
             inputProps={{
-              "aria-label": ariaLabel || label,
-              "aria-required": required
+              "aria-label": accessibleName,
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The four-key `inputProps` object is now duplicated verbatim in both `Select` branches, and this PR had to edit the same two new attributes in two places — exactly the pattern that produces drift. The PR already demonstrates the right fix by hoisting `accessibleName`, `errorId` and `menuListProps`; do the same here: define a single `const selectInputProps = { "aria-label": accessibleName, "aria-invalid": !!error, "aria-required": required, "aria-describedby": errorId };` next to `menuListProps` (typed explicitly, e.g. `React.InputHTMLAttributes<HTMLInputElement>` or a small local `AriaSelectProps` interface) and pass `inputProps={selectInputProps}` in both branches. Longer term the two `Select` blocks are ~90% identical and differ only in children and a couple of `sx` values — extracting a single `Select` with a computed children expression would remove the duplication class entirely rather than one instance of it.

frontend/src/community/common/components/molecules/DropdownList/DropdownList.tsx
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

This branch previously had no `MenuProps` at all; now that the PR introduces one, the asymmetry with the populated branch becomes a deliberate-looking omission. The populated branch sets `MenuProps.style` with `maxHeight: 300` and `zIndex: ZIndexEnums.NEWMODAL`, and `NEWMODAL` is the established modal-layer escape hatch used throughout the codebase (TransferMembersModal, ReassignMemberRow, TimeInput, AddEditTeamModal all do the same). Because this branch omits it, a DropdownList rendered inside a modal while its `itemList` is still loading will pop its menu at the default z-index and can render behind the modal surface — and it also loses the 300px max-height cap. Since `menuListProps` is already hoisted, hoist the menu style too and share it: build one `menuProps` object (`{ style: { maxHeight: 300, zIndex: ZIndexEnums.NEWMODAL, ...(enableTextWrapping ? { width: "max-content" } : {}) }, MenuListProps: menuListProps }`) and pass the same object to both `Select`s.

frontend/src/community/common/components/molecules/DropdownList/DropdownList.tsx
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

This JSX was rewritten by the PR, so the hardcoded `style={{ color: "red" }}` should be cleaned up while you are here. It violates two of the styling guidelines: a raw colour literal instead of a theme token, and an inline `style` instead of `sx` / a colocated `styles.ts` entry. The rest of this component already uses `theme.palette.error.contrastText` for error state (see `labelStyle` and `errorTextStyle` in the sibling `styles.ts`), so the asterisk is currently the only element rendering a non-theme red — it will not follow a theme change and its contrast is unverified against the label background. Replace with `<Box component="span" aria-hidden="true" sx={classes.requiredAsteriskStyle}>*</Box>` and add `requiredAsteriskStyle: { color: theme.palette.error.contrastText }` to `styles.ts`. (Same class of issue on the `CircularProgress` `style={{ color: "black" }}` in the branch this PR also touches — use `theme.palette.common.black` or the primary token.)

frontend/src/community/common/components/molecules/DropdownList/DropdownList.tsx
```diff
@@ -292,6 +308,7 @@ const DropdownList: FC<Props> = ({
             name={inputName}
```
ThinuwanW, line 314: 🤖 **Claude** · 🟣 **Nit**

Null-safety inconsistency between the two branches that this PR makes more visible by touching both. The populated branch normalises with `value={value?.toString() ?? ""}`, but this empty-list branch (line 294) uses `value={value?.toString()}` with no `?? ""`. When the optional `value` prop is undefined — which is the common case during initial load, i.e. exactly when this branch renders — MUI receives `value={undefined}` and treats the Select as uncontrolled, then logs "MUI: A component is changing an uncontrolled Select to be controlled" as soon as `itemList` populates and a value arrives. Add the same `?? ""` fallback here so both branches behave identically.

frontend/src/community/common/components/molecules/DropdownList/DropdownList.tsx
```diff
@@ -176,16 +189,19 @@ const DropdownList: FC<Props> = ({
...
             inputProps={{
-              "aria-label": ariaLabel || label,
-              "aria-required": required
+              "aria-label": accessibleName,
+              "aria-invalid": !!error,
+              "aria-required": required,
+              "aria-describedby": errorId
             }}
             displayEmpty={!!placeholder?.length}
             renderValue={
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`renderValue` branches on the outer `value` prop rather than the value MUI hands it, and the comparison is `value !== ""`. When the optional `value` prop is `undefined` (no selection yet), `undefined !== ""` is true, so the "selected" renderer runs, `itemList.find(item => item.value === undefined)` returns undefined, and an empty Typography is rendered — the placeholder never appears even though `displayEmpty` is on. This matters for this PR specifically: the placeholder is the natural accessible-name fallback suggested above, and it is currently unreachable for undefined values. Fix by using the callback argument and a nullish check, e.g. `renderValue={(selected) => (selected === undefined || selected === "" ? placeholderNode : selectedNode)}`. Note the existing test "displays placeholder when no value is selected" does not catch this — it passes because `label={placeholder}` makes MUI render the placeholder text inside the `aria-hidden` notched-outline `<legend>`, which `getByText` still matches. Worth asserting on the actual placeholder element (or using `getByRole("combobox")` text content) when the missing tests from the first pass are added.

frontend/src/community/common/components/molecules/DropdownList/DropdownList.tsx
```diff
@@ -176,16 +189,19 @@ const DropdownList: FC<Props> = ({
                 maxHeight: 300,
                 zIndex: ZIndexEnums.NEWMODAL,
                 ...(enableTextWrapping ? { width: "max-content" } : {})
-              }
+              },
+              MenuListProps: menuListProps
             }}
             sx={{
               ...classes.selectStyle(theme, isDisabled, readOnly as boolean),
```
ThinuwanW, line 211: 🤖 **Claude** · 🟣 **Nit**

Type-assertion cleanups in the code this accessibility pass touches (strict mode is on, and the guidelines call out `as` used to silence optionality). Three cases: (1) line 196 `readOnly as boolean` — `readOnly` is `boolean | undefined`; give the prop a default of `false` in the destructure, as is already done for `isDisabled`, `errorFocusOutlineNeeded` and `enableTextWrapping`, and drop the cast. (2) line 152 `ariaDescription={tooltip as string}` — `tooltip` is typed `string | JSX.Element`, and `Tooltip`'s `ariaDescription` is `string`; when a caller passes JSX, the assertion is a lie and the element ends up with `aria-description="[object Object]"`. Guard it: `ariaDescription={typeof tooltip === "string" ? tooltip : undefined}`. (3) the `emoji as string` casts in `renderValue` and the MenuItem map have the same shape — `getEmoji` should accept `string | undefined` instead.

frontend/src/community/common/components/molecules/DropdownList/DropdownList.tsx
umeshaliyanage: no test needed unless its a util

frontend/src/community/common/components/molecules/DropdownList/DropdownList.tsx
umeshaliyanage: inputProps only forwards aria-label/aria-describedby to the combobox in MUI v6 — aria-invalid/aria-required were no-ops. Moved both to SelectDisplayProps on both <Select> blocks.

frontend/src/community/common/components/molecules/DropdownList/DropdownList.tsx
umeshaliyanage: EmergencyDetailsForm 126-130 renders both sections with inputName="relationship" and an error, so the secondary combobox's aria-describedby did resolve to the primary's error node. errorId now seeds from useId(), which fixes all 5 relationship sites and the billableFrequency row case in one place.

frontend/src/community/common/components/molecules/DropdownList/DropdownList.tsx
umeshaliyanage: aria-required moved from inputProps (where MUI drops it) to SelectDisplayProps, so the combobox now exposes aria-required="true" and the hidden asterisk is fine.

frontend/src/community/common/components/molecules/DropdownList/DropdownList.tsx
umeshaliyanage: the label gets an id from the same useId() seed, wired via SelectDisplayProps, Guarded as label && !ariaLabel so call sites passing an explicit ariaLabel keep their override. This also replaces MUI's default self-referencing aria-labelledby, which pointed at the combobox's own id. Accessible name verified as "Relationship", with the aria-hidden asterisk correctly excluded.

frontend/src/community/common/components/molecules/DropdownList/DropdownList.tsx
umeshaliyanage: collapsed to const menuListProps: Partial<MenuListProps> = { "aria-label": accessibleName }; with MenuListProps imported from @mui/material

frontend/src/community/common/components/molecules/DropdownList/DropdownList.tsx
umeshaliyanage: out of scope

frontend/src/community/common/components/molecules/DropdownList/DropdownList.tsx
umeshaliyanage: <MenuItem disabled value=""> instead of a bare Box/Typography, so the listbox contains real options. Added aria-busy on the loading branch's listbox only, and aria-hidden on the spinner so it isn't announced as an unnamed progressbar
 
 The MenuItem also needs &.Mui-disabled { opacity: 1 }, otherwise the text renders at 0.38 opacity and fails contrast

frontend/src/community/common/components/molecules/DropdownList/DropdownList.tsx
umeshaliyanage: Added || placeholder to the chain.

frontend/src/community/common/components/molecules/DropdownList/DropdownList.tsx
umeshaliyanage: Dropped the redundant aria-live="assertive" and switched role="alert" → role="status", keeping id + aria-describedby.
Kept both channels since validation is on blur (validateOnChange: false, validateOnBlur: true), so description-only would announce nothing when the error appears.

frontend/src/community/common/components/molecules/DropdownList/DropdownList.tsx
umeshaliyanage: Dropped the redundant aria-live="assertive" and switched role="alert" → role="status", keeping id + aria-describedby.
Kept both channels since validation is on blur (validateOnChange: false, validateOnBlur: true), so description-only would announce nothing when the error appears.

frontend/src/community/common/components/molecules/DropdownList/DropdownList.tsx
umeshaliyanage: ARIA state moved to SelectDisplayProps earlier in this PR, so the duplication was across two objects and I hoisted both: selectInputProps: InputHTMLAttributes<HTMLInputElement> and selectDisplayProps: HTMLAttributes<HTMLDivElement>

frontend/src/community/common/components/molecules/DropdownList/DropdownList.tsx
umeshaliyanage: Done  added requiredAsteriskStyle and spinnerStyle to styles.ts and switched both to sx; no inline style or raw colour literals left in the component.

frontend/src/community/common/components/molecules/DropdownList/DropdownList.tsx
umeshaliyanage: ignored

frontend/src/community/common/components/molecules/DropdownList/DropdownList.tsx
umeshaliyanage: Confirmed all three points — combobox.textContent was "" for an undefined value, and the placeholder text was only present in the aria-hidden <legend>. renderValue now takes the callback argument and nullish-checks it.

frontend/src/community/common/components/molecules/DropdownList/DropdownList.tsx
umeshaliyanage: ignored adding fallbacks
(1) got readOnly = false in the destructure, matching the other defaulted prop

frontend/src/community/common/components/molecules/DropdownList/DropdownList.tsx
```diff
@@ -300,28 +349,41 @@ const DropdownList: FC<Props> = ({
...
-              "aria-required": required
-            }}
+            inputProps={selectInputProps}
+            SelectDisplayProps={selectDisplayProps}
           >
             {showSpinnerWhenNoData ? (
-              <Box display={"flex"} justifyContent={"center"}>
-                <CircularProgress size={20} style={{ color: "black" }} />
-              </Box>
+              <MenuItem
```
Deshan1100: check this again

frontend/src/community/common/components/molecules/DropdownList/DropdownList.tsx
```diff
@@ -67,6 +76,9 @@ interface Props {
   noOptionsText?: string;
 }
 
+const getMenuContainer = (): HTMLElement | null =>
+  typeof document === "undefined" ? null : document.querySelector("main");
```
Deshan1100: set the type first if need only set null

frontend/src/community/common/components/molecules/DropdownList/DropdownList.tsx
```diff
@@ -104,6 +116,31 @@ const DropdownList: FC<Props> = ({
   const theme: Theme = useTheme();
   const classes = styles(theme);
 
+  const accessibleName = ariaLabel || label || placeholder;
```
Deshan1100: check what happens here

frontend/src/community/common/components/molecules/DropdownList/DropdownList.tsx
```diff
@@ -125,18 +162,28 @@ const DropdownList: FC<Props> = ({
...
         <Typography
           component="label"
+          id={labelId}
           lineHeight={1.5}
           sx={{ ...classes.labelStyle(isDisabled, !!error), ...labelStyles }}
         >
-          {label} {required && <span style={{ color: "red" }}>*</span>}
+          {label}{" "}
+          {required && (
+            <Box
```
Deshan1100: revert this

frontend/src/community/common/components/molecules/DropdownList/DropdownList.tsx
```diff
@@ -125,18 +162,28 @@ const DropdownList: FC<Props> = ({
...
+          )}
         </Typography>
         {tooltip && (
           <Tooltip
             title={tooltip}
             maxWidth={toolTipWidth}
             id={toolTipId}
             isDisabled={isDisabled}
-            ariaDescription={tooltip as string}
+            ariaDescription={typeof tooltip === "string" ? tooltip : undefined}
```
Deshan1100: revert and check

Deshan1100 (changes requested):

frontend/src/community/common/components/molecules/DropdownList/DropdownList.tsx
umeshaliyanage: reverted

frontend/src/community/common/components/molecules/DropdownList/DropdownList.tsx
umeshaliyanage: reverted

frontend/src/community/common/components/molecules/DropdownList/DropdownList.tsx
umeshaliyanage: const getMenuContainer = () => document.querySelector<HTMLElement>("main");

frontend/src/community/common/components/molecules/DropdownList/DropdownList.tsx
umeshaliyanage: changes to accessibleName = ariaLabel || label  has been used in existing dropdowns incase one doesnt exist other wise it misses some use cases

frontend/src/community/common/components/molecules/DropdownList/DropdownList.tsx
umeshaliyanage: reverted back to box

Deshan1100 (approved):

ThinuwanW (approved):

PR #1823, opened 2026-08-24
Refactor/crm company tab section

dulsara-manakal:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds a large v2 CRM Companies surface (API layer, Zustand-normalized store usage, company side panel, company/task modals, task rows) behind a hardcoded `isCrmCompaniesV2 = false` flag, so none of it ships enabled — but several parts are wired against the wrong backend contract and would break the moment the flag is flipped. The most serious issues are that `useGetTasksInfinite` targets `GET /v2/crm/task`, which returns a non-paginated `{tasks: []}` payload rather than `{items, currentPage, totalPages}` (the side-panel effect will throw), and that the v2 task API returns nested `type`/`owner`/`contact`/`company`/`deal` objects while every new consumer reads flat `typeId`/`ownerId`/`contactId` ids. Also flagged: a broken optimistic-toggle rollback, a task owner default dropped from v1 that makes task creation impossible for sales reps, an accidental tightening of the website regex, and a set of unguarded record lookups.

Found **43** new issue(s): 🔴 17 important, 🟡 18 suggestion(s), 🟣 8 nit(s)

frontend/src/community/crm/v2/api/TaskApi.ts
```diff
@@ -0,0 +1,124 @@
...
+
+interface UpdateTaskVariables {
+  id: number;
+  payload: CrmTaskEntity;
+}
+
+const fetchTasks = async (
+  params: CrmTaskCompletedFilterRequest
+): Promise<CrmTaskCompletedListResponse> => {
+  const response = await authFetchV2.get(crmTaskEndpoints.GET_TASKS, {
```
ThinuwanW, line 28: 🤖 **Claude** · 🔴 **Important**

`GET_TASKS` points at `/crm/task` on authFetchV2, i.e. `GET /v2/crm/task`. That endpoint (CrmTaskControllerV2.getTasks → CrmTaskFilterDto) is NOT paginated: it ignores `page`/`size` (they are not fields on CrmTaskFilterDto), returns only open tasks, and responds with `CrmTaskListResponseDtoV2` = `{ tasks: [...] }` — not `{ items, currentPage, totalItems, totalPages }` as `CrmTaskCompletedListResponse` declares. Consequences: `getNextPageParam` reads `undefined + 1` (NaN) so pagination is silently dead, and in CompanySidePanel.tsx:166 `fetchedTasks.pages.flatMap((page) => page.items)` produces `[undefined]`, which makes `normalizeTasks` throw on `task.id` — the company side panel will crash as soon as the v2 flag is on. Use the paginated endpoints instead: `/v2/crm/task/completed` (CrmTaskCompletedFilterDto supports companyId + page/size) and/or `/v2/crm/task/related`, or type this hook against the non-paginated `{tasks}` shape and drop useInfiniteQuery.

frontend/src/community/crm/v2/api/TaskApi.ts
```diff
@@ -0,0 +1,124 @@
...
+): UseMutationResult<CrmTaskEntity, AxiosError, CrmTaskEntity> => {
+  const queryClient = useQueryClient();
+
+  return useMutation({
+    mutationFn: createTask,
+    onSuccess: (createdTask) => {
+      queryClient.invalidateQueries({
+        queryKey: crmLimitationQueryKeys.GET_CRM_LIMITATION
+      });
+      if (createdTask.companyId !== undefined) {
```
ThinuwanW, line 77: 🤖 **Claude** · 🔴 **Important**

The v2 task API returns `CrmTaskResponseDtoV2`, which nests `type`, `owner`, `company`, `contact` and `deal` as objects — it has no `typeId`, `ownerId`, `companyId`, `contactId` or `dealId` scalars. `CrmTaskEntity` and every new consumer assume the flat shape, so: (a) these `createdTask.companyId`/`contactId` guards are always false and the metrics/list invalidations never run; (b) `linkTaskToRelatedEntities` never links a created task to anything; (c) TaskRowMeta never renders an owner avatar (`task.ownerId` undefined) and TaskRowContent always falls back to the default type icon (`task.typeId` undefined). Either map the response to the flat entity in the API layer (e.g. `owner.employeeId → ownerId`, `type.id → typeId`) or change `CrmTaskEntity` and its consumers to the nested contract. Note the create request DTO also has no `companyId`, so a task can only reach a company through its contact/deal.

frontend/src/community/crm/v2/components/molecules/AddTaskModalContent/AddTaskModalContent.tsx
```diff
@@ -0,0 +1,108 @@
...
+  });
+
+  const { setSubmitting } = formik;
+
+  const handleSuccess = (createdTask: CrmTaskEntity) => {
+    if (createdTask.id !== undefined) {
+      setTasks({ ...tasks, [createdTask.id]: createdTask });
+    }
+
+    const linked = linkTaskToRelatedEntities(createdTask, {
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

A task created from the company side panel never shows up in the list. Nothing invalidates `crmTaskQueryKeys` after create/update (useCreateTask only invalidates the CRM limitation and contact/company metrics), and the store path relies on `linkTaskToRelatedEntities`, which can only append to `companies[companyId].taskIds` — but the form never sets `companyId` (the backend create DTO has no such field) and the response carries no flat ids either. So `company.taskIds` is unchanged and no refetch happens. Fix by invalidating the task list query key in `useCreateTask`/`useUpdateTask`, and/or by appending the created task id to the currently open company's `taskIds` explicitly.

frontend/src/community/crm/v2/components/molecules/TaskRow/TaskRow.tsx
```diff
@@ -0,0 +1,92 @@
...
+      open: true,
+      toastType: ToastType.ERROR,
+      title: translateText(["toggleErrorTitle"]),
+      description: translateText(["toggleErrorDescription"])
+    });
+  };
+
+  const { mutate: updateCompletion } = useUpdateTask(
+    (updatedTask) => setTasks(updateTask(tasks, taskId, updatedTask)),
+    () => handleToggleError(isCompleted)
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The optimistic-toggle rollback is a no-op. `handleToggleChange` writes the new value into the store first; React re-renders, so `isCompleted` becomes the new value; TanStack Query re-applies the latest options object on every render, so when the request fails this `onError` closure calls `handleToggleError(isCompleted)` with the already-flipped value and restores the wrong state (the user sees an error toast but the checkbox stays toggled). Follow the v1 pattern: capture `const wasCompleted = task.isCompleted` before the optimistic write and pass a per-call handler via `updateCompletion(vars, { onError: () => handleToggleError(wasCompleted) })`.

frontend/src/community/crm/v2/components/molecules/TaskModalForm/TaskModalForm.tsx
```diff
@@ -0,0 +1,462 @@
...
+    contactFilters.companyId !== undefined;
+
+  const isDealSearchEnabled =
+    debouncedDealSearchText.length > 0 ||
+    hasSelectedContact ||
+    dealFilters.companyId !== undefined;
+
+  const { data: ownerLookupData } = useGetOwnerLookup(
+    ownerFilters,
+    Boolean(isCrmSalesManager) && debouncedOwnerSearchText.length > 0
```
ThinuwanW, line 159: 🤖 **Claude** · 🔴 **Important**

A CRM sales representative can never create a task. The owner lookup is gated on `isCrmSalesManager` (correct — `/v1/crm/contact/owners` requires ROLE_CRM_SALES_MANAGER), so for a rep `ownerDropdownItems` is always empty; meanwhile `getTaskFormInitialValues` (taskUtil.ts:107) sets no `ownerId` and `getTaskValidationSchema` marks `ownerId` required — submit is permanently blocked with no way to fill the field. v1's AddTaskModalContent defaulted the owner to the current user via `useGetUserPersonalDetails`. Restore that default (and keep the remove button manager-only, as `showRemoveButton` already does).

frontend/src/community/crm/v2/regex/crmRegexPatterns.ts
```diff
@@ -0,0 +1,3 @@
+export function isValidCompanyWebsiteUrl(): RegExp {
+  return /^https:\/\/(www\.)?[a-z0-9-]{1,63}(\.[a-z0-9-]{1,63}){0,9}\.[a-z]{2,63}(\/[^\s?#]*)?$/i;
```
ThinuwanW, line 2: 🤖 **Claude** · 🔴 **Important**

This is a copy of `~community/crm/regex/crmRegexPatterns.isValidCompanyWebsiteUrl` with one difference: the v1 pattern starts with `^(https:\/\/)?` (optional protocol) while this one requires a literal `https://`. That silently tightens validation — `www.acme.com` and `acme.com` were accepted in v1 and are now rejected, and existing stored websites without a protocol will fail on edit. Restore the optional group, or, if requiring the protocol is intentional, apply it in v1 too and normalise legacy values (SidePanelCompanyHeader calls `openInNewTab(website)` directly, which navigates relative to the app for protocol-less values). Also consider importing the existing v1 helper instead of duplicating the regex.

frontend/src/community/crm/v2/utils/dealUtil.ts
```diff
@@ -55,3 +60,14 @@ export const resolveDeals = (
   dealIds
     .map((id) => deals[id])
     .filter((deal): deal is CrmDealEntity => Boolean(deal));
+
+export const getDealNameById = (deals: CrmDealRecord, dealId?: number) => {
+  if (dealId !== undefined) {
+    return deals[dealId].name;
```
ThinuwanW, line 66: 🤖 **Claude** · 🔴 **Important**

Unguarded normalized-record lookups will throw whenever an id list and its record get out of sync (a page of ids loaded while the record was reset/replaced, an entity deleted elsewhere, or task types not yet initialized). Sites: `deals[dealId].name` here, `taskTypes[typeId].name` in taskUtil.ts:83, `tasks[taskId]` then `task.isCompleted` in TaskRow.tsx:37, `deals[dealId]` passed to DealAccordionItemHeader in SidePanelDealSection.tsx:99 (`deal.name`), `contacts[contactId]` then `contact.metrics` in SidePanelContactsSection.tsx:56, `companies[id]` then `company.name` in CompanyTable.tsx:111, and `owners[task.ownerId]` handed to OwnerAvatar in TaskRowMeta (OwnerAvatar dereferences `owner.authPic`/`owner.firstName`). Add optional chaining and skip/render a fallback when the entity is missing — `resolveDeals` in this same file already shows the safe pattern.

frontend/src/community/crm/v2/api/TaskApi.ts
```diff
@@ -0,0 +1,124 @@
...
+): UseInfiniteQueryResult<
+  InfiniteData<CrmTaskCompletedListResponse>,
+  AxiosError
+> =>
+  useInfiniteQuery({
+    queryKey: crmTaskQueryKeys.LIST(params),
+    queryFn: ({ pageParam }) => fetchTasks({ ...params, page: pageParam }),
+    initialPageParam: 0,
+    getNextPageParam: (lastPage) => {
+      const nextPage = lastPage.currentPage + 1;
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`getNextPageParam` dereferences `lastPage.currentPage`/`lastPage.totalPages` without a guard, and the page is derived from `response?.data?.results?.[0]`, which is optional-chained precisely because it can be undefined. The pre-existing `useGetDealsInfinite` guards this (`lastPage?.currentPage !== undefined && ...`). Same pattern in CompanyApi.ts:51 and ContactApi.ts. Please mirror the guarded version so an empty/error-shaped response cannot throw inside React Query.

frontend/src/community/crm/v2/api/CompanyApi.ts
```diff
@@ -1,10 +1,179 @@
...
+  payload
+}: EditCompanyVariables): Promise<CrmCompanyEntity> => {
+  const response = await authFetch.patch(
+    crmCompanyEndpoints.EDIT_COMPANY(id),
+    payload
+  );
+  return response?.data?.results?.[0];
+};
+
+export const useEditCompany = (
```
ThinuwanW, line 147: 🤖 **Claude** · 🟡 **Suggestion**

Create/edit/delete company only invalidate `crmLimitationQueryKeys`; none of them touch `crmCompanyQueryKeys.LIST`, `DETAIL(id)` or `METRICS(id)`. State consistency depends entirely on the Zustand writes in the modal contents, so any component that later reads the React Query cache (the side panel re-opening on a cached `DETAIL`, a list refetch) can serve stale or deleted rows. Add `queryClient.invalidateQueries({ queryKey: crmCompanyQueryKeys.ALL })` (which is currently declared but unused) plus `DETAIL`/`METRICS` for edit, and the same for delete.

frontend/src/community/crm/v2/components/molecules/SidePanelAddDeal/SidePanelAddDeal.tsx
```diff
@@ -0,0 +1,207 @@
...
+
+  const isFormDisabled = isPending || isUserLoading || isCheckingCrmLimit;
+
+  const handleSubmit = (values: InlineDealFormValues) => {
+    guardCrmCreate(CrmLimitResource.DEALS, () => {
+      createDeal({
+        name: values.name.trim(),
+        stageId: initialStageId,
+        contactId: Number(values.contactId),
+        ownerId: Number(currentUser?.employeeId),
```
ThinuwanW, line 118: 🤖 **Claude** · 🟡 **Suggestion**

`ownerId: Number(currentUser?.employeeId)` becomes `NaN` if the personal-details query fails (only `isUserLoading` is guarded, not the error case), and `stageId: initialStageId` is `undefined` when the stage store has not been hydrated. Both produce a malformed create payload. Guard the submit on `currentUser?.employeeId` and `initialStageId` being present (disable the save action and/or bail with a toast) instead of coercing.

frontend/src/community/crm/v2/utils/commonUtil.ts
```diff
@@ -26,14 +62,14 @@ export const toOwnersRecord = (owners: CrmOwnerEntity[]): CrmOwnerRecord => {
   return ownerRecord;
 };
 
-export const toStagesRecord = (stages: CrmStageEntity[]): CrmStageRecord => {
-  const stageRecord: CrmStageRecord = {};
-  for (const stage of stages) {
-    if (stage.id != null) {
-      stageRecord[stage.id] = stage;
+export const toDealsRecord = (deals: CrmDealEntity[]): CrmDealRecord => {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Duplicated helpers introduced inside v2: `toDealsRecord` now exists here and in `v2/utils/dealUtil.ts` with identical behaviour (the only consumer, TaskModalForm, imports the dealUtil one — so this copy is dead), and `getPriorityConfig` now exists in both `v2/utils/taskUtil.ts` (new) and `v2/utils/priorityUtil.ts` (existing, non-null-safe variant still used by v2 PriorityLabel). Two same-named exports in the same utils folder is an easy import mistake; delete this `toDealsRecord` and consolidate the two `getPriorityConfig` implementations on the safer new one.

frontend/src/community/crm/v2/components/molecules/SidePanelTasksSection/SidePanelTasksSection.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const handleAddTask = () => {
+    guardCrmCreate(CrmLimitResource.TASKS, () => {
+      setTaskModalType(CrmModalTypes.ADD_TASK_MODAL);
+      setIsTaskModalOpen(true);
+    });
+  };
+
+  if (taskIds !== undefined && taskIds.length > 0) {
+    return (
+      <div ref={loadingRef}>
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`loadingRef` is attached to the wrapper around the whole task list rather than to a zero-height sentinel after it. `useInfiniteScroll` uses `threshold: 1`, so the observer only fires while the entire list is inside the viewport — once the list grows past one screen, pagination stops. SidePanelDealSection (added in this same PR) does it correctly with a trailing `<div ref={loadingRef} />`. This mirrors the v1 component, so it is a ported issue rather than a new one, but worth fixing while the code is being rewritten.

frontend/pages/community/crm/companies.tsx
```diff
@@ -11,10 +12,21 @@ import { CompanyTable } from "~community/crm/components/organisms/CompanyTable/C
...
+import TaskModalControllerV2 from "~community/crm/v2/components/organisms/TaskModalController/TaskModalController";
+import { useInitializeCrmData } from "~community/crm/v2/hooks/useInitializeCrmData";
+import { useCrmStoreV2 } from "~community/crm/v2/store/store";
+import { CrmModalTypes as CrmModalTypesV2 } from "~community/crm/v2/types/CrmTypes";
 import useCrmLimitGuard from "~enterprise/crm/hooks/useCrmLimitGuard";
 import { CrmLimitResource } from "~enterprise/crm/types/CrmLimitTypes";
 
-const Companies: NextPage = () => {
+// Flip to true to serve the CRM Companies page from the normalized v2 store surface.
+const isCrmCompaniesV2 = false;
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The entire v2 surface is gated behind a module-level `const isCrmCompaniesV2 = false`, so ~4k lines of new code ship dead, are never exercised in any environment, and TypeScript narrows the ternary at line 128 to the V1 branch. That is also why the contract mismatches flagged above (task endpoint shape, nested task DTO) would not be caught by manual QA on this PR. Prefer a real toggle (env var / config flag consistent with `NEXT_PUBLIC_MODE` handling) so the branch can actually be enabled in a test environment, and land the v2 page only once the task endpoints are verified end to end.

frontend/src/community/crm/v2/components/molecules/SidePanelAddDeal/AddDealContactSearch.tsx
```diff
@@ -0,0 +1,106 @@
+import { ButtonV2, CloseIcon, InputField } from "@rootcodelabs/skapp-ui";
```
ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

This component and `SelectableSearchField` (also added in this PR) are the same widget: read-only InputField with a tertiary clear button when a value is selected, otherwise a SearchableDropdown. The only real differences are the `variant` and that this one owns its search text state. Collapse them into one — `SelectableSearchField` with an optional `variant` and `errorMessage`/`state` passthrough — so contact selection behaves identically in the task modal and the inline add-deal row.

frontend/src/community/crm/v2/components/molecules/TaskModalForm/TaskModalForm.tsx
```diff
@@ -0,0 +1,462 @@
+import {
```
ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

TaskModalForm is 462 lines and CompanySidePanel 344, well past the ~200-line decomposition guideline: this file owns three debounced searches, three lookup queries, two store-sync effects, three memoized dropdown item lists, ten handlers and the full JSX. Extract the lookup/search wiring into hooks (e.g. `useTaskOwnerSearch`, `useTaskContactSearch`, `useTaskDealSearch`) and split the JSX into the field groups. Relatedly, this PR adds unit tests for the new utils only — none of the new components (CompanyTable, CompanySidePanel, TaskRow, the modal contents) have a `.test.tsx`, which the review checklist requires for new components.

frontend/src/community/crm/v2/api/utils/ApiEndpoints.ts
```diff
@@ -14,18 +32,21 @@ export const crmDealEndpoints = {
   CREATE_DEAL_STAGE: `${moduleAPIPath.CRM}/deal/stage`,
   UPDATE_DEAL_STAGE: (id: number) => `${moduleAPIPath.CRM}/deal/stage/${id}`,
   REORDER_DEAL_STAGES: `${moduleAPIPath.CRM}/deal/stage/reorder`,
-  DELETE_DEAL_STAGE: (id: number) => `${moduleAPIPath.CRM}/deal/stage/${id}`
-};
-
-export const crmCompanyEndpoints = {
-  GET_COMPANIES_BY_IDS: `${moduleAPIPath.CRM}/company/ids`
+  DELETE_DEAL_STAGE: (id: number) => `${moduleAPIPath.CRM}/deal/stage/${id}`,
+  GET_DEALS: `${moduleAPIPath.CRM}/deal`
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Dead additions: `crmDealEndpoints.GET_DEALS` duplicates `crmDealEndpointsV2.GET_DEALS` and has no consumer (v2 DealApi uses the V2 constant); `crmCompanyQueryKeys.ALL` and `crmTaskQueryKeys.ALL` in QueryKeys.ts are declared but never used; `TASK_TYPE_ICON_SIZE` in taskConstants.ts is exported while `getTaskTypeIcon` hardcodes `size = 20`. Also note `crmCompanyQueryKeys.COMPANIES_BY_IDS` and `crmDealQueryKeys.GET_DEALS` still use the old `crm-*-v2` string prefixes while the new keys use the `CRM_COMPANIES`/`CRM_DEALS` constants, so a prefix invalidation on `ALL` would not reach them.

frontend/src/community/crm/v2/components/templates/SidePanelWrapper/SidePanelWrapper.tsx
```diff
@@ -0,0 +1,27 @@
...
+    if (!isOpen) return;
+
+    const originalOverflow = document.body.style.overflow;
+    document.body.style.overflow = "hidden";
+    return () => {
+      document.body.style.overflow = originalOverflow;
+    };
+  }, [isOpen]);
+
+  return <div style={{ zIndex: ZIndexEnums.CRM_SIDE_PANEL }}>{children}</div>;
```
ThinuwanW, line 24: 🤖 **Claude** · 🟣 **Nit**

`zIndex` has no effect on a statically positioned element, so this wrapper does not actually stack the side panel above anything — either add a positioning context (`relative`/`fixed`) or drop the inline style and let the SidePanel component own its stacking. The inline style also conflicts with the colocated-styles convention; a Tailwind class or `styles.ts` entry would be more consistent.

frontend/src/community/crm/v2/utils/commonUtil.ts
```diff
@@ -7,6 +9,40 @@ import {
   CrmTaskTypeRecord
 } from "~community/crm/v2/types/CrmCommonTypes";
 
+const isEmptyValue = (value?: string | number) =>
+  value === undefined || Number(value) === 0;
+
+export const formatTableValue = (value?: string | number, prefix = "") =>
+  isEmptyValue(value) ? "-" : `${prefix}${value}`;
+
+export const formatMonetaryValue = (value?: string) => {
```
ThinuwanW, line 16: 🤖 **Claude** · 🟣 **Nit**

There are now four overlapping formatters in this file — `formatTableValue`, `formatMonetaryValue`, `formatMonetaryValueWithDecimals` and the pre-existing `formatCurrency` — producing inconsistent output for the same data (`$14700000` vs `$14700000.00`, none with thousand separators). `isEmptyValue` also treats a legitimate zero count/amount as missing (`formatTableValue(0)` → `-`) and `formatMonetaryValue("0.50")` renders `$0`, which the accompanying test documents but reads as a rounding bug to a user. Consider one currency helper with a `decimals` option built on `Intl.NumberFormat`, and make the zero-as-placeholder rule explicit at the call sites that want it.

frontend/src/community/crm/v2/components/organisms/CompanyTable/CompanyTable.tsx
```diff
@@ -0,0 +1,208 @@
...
+    },
+    {
+      id: "accountValue",
+      label: translateText(["table", "columns", "accountValueHeader"]),
+      width: "20%",
+      align: "right"
+    }
+  ];
+
+  const transformToTableRows = (): GridRow[] =>
```
ThinuwanW, line 111: 🤖 **Claude** · 🟣 **Nit**

`transformToTableRows()` rebuilds every row's JSX on every render (including on each keystroke in the search box, before debouncing settles) and is not memoized; wrap it in `useMemo` keyed on `companyIds`/`companies`/`translateText`. Also note the sync effect replaces `companyIds` wholesale from the query pages, so the optimistic prepend done in AddCompanyModalContent is discarded the next time `data` changes — worth confirming that ordering is intended.

frontend/src/community/crm/v2/components/molecules/AddTaskModalContent/AddTaskModalContent.tsx
```diff
@@ -0,0 +1,108 @@
...
+      setTasks: store.setTasks,
+      setCompanies: store.setCompanies,
+      setContacts: store.setContacts,
+      setDeals: store.setDeals,
+      setIsTaskModalOpen: store.setIsTaskModalOpen
+    }))
+  );
+
+  const formik = useFormik<CrmTaskEntity>({
+    initialValues: getTaskFormInitialValues(selectedContactId),
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

A task created from the Company side panel is never linked to the company, so it never appears in that company's Tasks tab. Three facts combine: (1) the backend `CrmTaskCreateRequestDto` has no `companyId` field at all — `CrmTaskServiceImpl.persistNewTask` derives the company purely from `contactId` (`company = contact.getCompany()`); (2) `getTaskFormInitialValues(selectedContactId)` is the only place a contact is pre-filled, and on `/community/crm/companies` `selectedContactId` is always `null` (only contact flows set it) — `CompanySidePanel` never plumbs the company or one of its contacts into the modal; (3) `getTaskValidationSchema` requires only `name`, `typeId`, `ownerId` and `dueAt`, so contact is optional. Repro: open a company panel → Tasks tab → Add task → fill name/type/owner/due date → Save. The task is persisted with `company = null` and is unreachable from the company. Fix: either add `companyId` to `CrmTaskCreateRequestDto` + `persistNewTask` and pass `selectedCompanyId` through `getTaskFormInitialValues`, or make `contactId` required (with a Yup rule) whenever the modal is opened from the company panel and default it from the panel's company-scoped contact list.

frontend/src/community/crm/v2/components/organisms/CompanySidePanel/CompanySidePanel.tsx
```diff
@@ -0,0 +1,344 @@
...
+      !fetchedCompany ||
+      !fetchedMetrics ||
+      !fetchedTasks ||
+      !fetchedDeals ||
+      !fetchedContacts
+    ) {
+      return;
+    }
+
+    const taskItems = fetchedTasks.pages.flatMap((page) => page.items);
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The hydration effect assumes every infinite-query page carries an `items` array, and the task endpoint does not. `GET /v2/crm/task` binds `CrmTaskFilterDto` (no page/size) and `CrmTaskServiceImplV2.getTasks` returns `CrmTaskListResponseDtoV2` = `{ tasks: [...] }`. So `fetchedTasks.pages.flatMap((page) => page.items)` evaluates to `[undefined]`, and `normalizeTasks([undefined])` then throws `TypeError: Cannot read properties of undefined (reading 'id')` on the very first render of the panel — the whole side panel dies, not just the Tasks tab. The same pattern on lines 169-170 (`mergeDeals` / `toContactsRecord`) would throw identically if a deal/contact page ever came back without `items`. Two fixes needed: guard the reads (`page.items ?? []`) and, more importantly, point the task list at a paginated, company-scoped endpoint — none exists today (`/v2/crm/task/completed` filters by company but returns only completed tasks, and `/v2/crm/task/related` accepts only `contactId`/`dealId`), so this needs a paired backend change.

frontend/src/community/crm/v2/components/molecules/TaskModalForm/TaskModalForm.tsx
```diff
@@ -0,0 +1,462 @@
...
+    isContactSearchEnabled
+  );
+  const { data: dealLookupData } = useGetDealLookup(
+    dealFilters,
+    isDealSearchEnabled
+  );
+
+  useEffect(() => {
+    if (!contactLookupData) return;
+    setContacts({ ...contacts, ...toContactsRecord(contactLookupData.items) });
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

These two lookup-sync effects use replace-not-merge semantics and silently destroy already-hydrated store data. `toContactsRecord` / `toDealsRecord` assign `record[id] = entity` wholesale, and the enclosing top-level spread only merges at the id level — so a lookup row overwrites the full entity. The contact lookup returns `CrmContactLookupResponseDtoV2` = `{id, name, companyId}` only. Concrete repro: open a company panel → Contacts tab renders `contact.metrics.openTasksCount` / `closedDealValue` correctly → click Add task (the contact lookup is company-scoped so it returns those same contacts) → this effect replaces each contact with `{id,name,companyId}` → close the modal → every Contacts row now shows `-`. Line 178 does the same to deals, dropping `taskIds`, which then makes `linkTaskToRelatedEntities` skip the deal it should have updated. Note the module already has the correct helpers: `mergeCompanies` and `mergeDeals` do `merged[id] = {...merged[id], ...incoming}`. Add a `mergeContacts` with the same shape and use `mergeDeals` here instead of `toDealsRecord`.

frontend/src/community/crm/v2/components/molecules/SidePanelAddDeal/SidePanelAddDeal.tsx
```diff
@@ -0,0 +1,207 @@
...
+
+  const { guardCrmCreate, isCheckingCrmLimit } = useCrmLimitGuard();
+
+  const isFormDisabled = isPending || isUserLoading || isCheckingCrmLimit;
+
+  const handleSubmit = (values: InlineDealFormValues) => {
+    guardCrmCreate(CrmLimitResource.DEALS, () => {
+      createDeal({
+        name: values.name.trim(),
+        stageId: initialStageId,
```
ThinuwanW, line 116: 🤖 **Claude** · 🔴 **Important**

`stageId` can be sent as `undefined`. `initialStageId = getInitialStageId(stages)` reads the Zustand `stages` record, which is populated asynchronously by `useInitializeCrmData` (board init-data) — a query this component does not observe. The v1 twin at `~community/crm/components/molecules/SidePanelAddDeal/SidePanelAddDeal.tsx` guarded exactly this: it used `useGetMappedDealStages()` and included `isStagesLoading` in `isFormDisabled` (line 111 here), then asserted `stageId: initialStageId!`. That gate was dropped in the copy. Open a company panel and click Add deal before board init-data resolves and the payload omits `stageId`; `CrmValidations.validateDealStageId` rejects it server-side, so the user just gets the generic 'Failed to add deal' toast. Fix: add the stages-loaded condition to `isFormDisabled` (e.g. `initialStageId === undefined`) so the inline form is disabled until a stage id exists.

frontend/src/community/crm/v2/components/molecules/SidePanelAddDeal/SidePanelAddDeal.tsx
```diff
@@ -0,0 +1,207 @@
...
+    });
+  };
+
+  const formik = useFormik<InlineDealFormValues>({
+    initialValues: {
+      name: "",
+      contactId: defaultContact ? String(defaultContact.id) : ""
+    },
+    validationSchema: inlineAddDealValidations(translateText),
+    validateOnChange: false,
```
ThinuwanW, line 130: 🤖 **Claude** · 🔴 **Important**

Two more behaviours regressed relative to the v1 `SidePanelAddDeal` this file was copied from. (1) Duplicate-name handling was dropped: v1's `handleCreateDealError(error)` read `error.response.data.results[0].messageKey` and, on `CrmErrorMessageKeyEnum.DEAL_EXISTS`, called `formik.setFieldError('name', translateText(['inlineAddDeal','validations','dealNameExists']))`. The v2 version takes no argument and only fires the generic toast, so `CRM_ERROR_DEAL_EXISTS` (thrown by `CrmDealServiceImpl` via `existsByNameAndContact_IdAndIsDeletedFalse`) now reads as an unexplained failure, and the `inlineAddDeal.validations.dealNameExists` string is left unused. (2) The validation modes were flipped from `validateOnChange: true` / `validateOnBlur: false` to the opposite. The deal name is driven by `SubTaskInput`'s `onValueChange` → `setFieldValue('name', value)` with no formik blur wiring, so with `validateOnChange: false` a name error never clears while the user fixes it. Compounding this, `isInvalid={Boolean(formik.errors.contactId)}` (line 160) dropped v1's `formik.touched.contactId &&` check, so after one failed submit the contact field stays red until the next submit. Restore the v1 error mapping and validation modes.

frontend/src/community/crm/v2/components/molecules/TaskRow/TaskRowMeta.tsx
```diff
@@ -0,0 +1,41 @@
...
+        <PriorityIcon
+          icon={priorityConfig.icon}
+          bgColor={priorityConfig.bgColor}
+        />
+      )}
+
+      {task.ownerId !== undefined && (
+        <OwnerAvatar
+          id={`task-owner-${task.id}`}
+          owner={owners[task.ownerId]}
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Normalized-record reads in the new components are unguarded, and unlike the deal-store case these crash on a timing window that is reachable on first load. `owners[task.ownerId]` is typed non-optional and handed straight to `OwnerAvatar`, which does `useGetImageUrl(owner.authPic ?? "")` and `owner.firstName` — if the owner is not in the store this is a `TypeError`. The `owners` and `taskTypes` records come from `useInitializeCrmData`, a query completely independent of the five queries `CompanySidePanel` gates its skeleton on, and the page discards that hook's loading state (see `pages/community/crm/companies.tsx:89`), so the panel can render task rows before owners/task types exist. The same unguarded shape appears in `taskUtil.getTaskTypeName` (`taskTypes[typeId].name`, called unconditionally from `TaskRow`), `TaskRow` (`tasks[taskId].isCompleted`), `SidePanelContactsSection` (`contacts[contactId]` then `contact.metrics`), `SidePanelDealSection` (`deals[dealId]` passed to `DealAccordionItemHeader`), and `CompanyTable` (`companies[id].metrics`). Fix uniformly: have the lookup helpers return `T | undefined`, and either skip or render a placeholder row when the referenced entity is absent.

frontend/pages/community/crm/companies.tsx
```diff
@@ -61,4 +73,58 @@ const Companies: NextPage = () => {
...
+  const { setIsCompanyModalOpen, setCompanyModalType, selectedCompanyId } =
+    useCrmStoreV2(
+      useShallow((store) => ({
+        setIsCompanyModalOpen: store.setIsCompanyModalOpen,
+        setCompanyModalType: store.setCompanyModalType,
+        selectedCompanyId: store.selectedCompanyId
+      }))
+    );
+
+  useInitializeCrmData();
```
ThinuwanW, line 89: 🤖 **Claude** · 🔴 **Important**

`useInitializeCrmData()`'s return value is discarded, so the page has no gate on the board init-data query. That hook returns `{ isCrmInitialDataLoading, isCrmInitialDataError }` and is the only source of `stages`, `owners` and `taskTypes` in the v2 store. With it ungated: `useGetTaskTypeOptions` returns `[]` so the Add-task Type dropdown is empty and unselectable while `typeId` is a required Yup field; `getInitialStageId` returns `undefined` (see the SidePanelAddDeal finding); and `owners[...]`/`taskTypes[...]` lookups crash. The error branch only raises a toast inside the hook — the page renders as if the data were present. Fix: destructure the result and hold the table/side panel behind a loading state, and render an error state (not just a toast) when `isCrmInitialDataError`.

frontend/src/community/crm/v2/components/organisms/CompanySidePanel/CompanySidePanel.tsx
```diff
@@ -0,0 +1,344 @@
...
+  } = useGetDealsInfinite(dealFilters);
+  const {
+    data: fetchedContacts,
+    isLoading: isContactsLoading,
+    fetchNextPage: fetchNextContactsPage,
+    hasNextPage: hasNextContactsPage,
+    isFetchingNextPage: isFetchingNextContactsPage
+  } = useGetContactsInfinite(contactFilters);
+
+  const isLoading =
```
ThinuwanW, line 156: 🤖 **Claude** · 🔴 **Important**

The side panel has no error state, so a failed request leaves a permanent skeleton. `isLoading` ORs the `isLoading` of five queries, and the body renders `isLoading || !company ? <CompanySidePanelSkeleton /> : ...`. If `useGetCompanyById` errors (404 on a deleted company, 403, network), `isLoading` goes false but `company` stays `undefined`, so the skeleton renders forever with an empty `<h2>` in the header and no way for the user to learn what happened. None of the five queries' `isError`/`error` values are read anywhere in this file. Fix: destructure `isError` from the detail query (at minimum) and render an error view with a retry, mirroring the `deals.sidePanel.errors.dealNotFound*` copy that already exists in `crmModule.json`.

frontend/src/community/crm/v2/components/molecules/SidePanelHeaderInfoItem/SidePanelHeaderInfoItem.tsx
```diff
@@ -0,0 +1,51 @@
+import { FC, ReactElement } from "react";
```
ThinuwanW, line 1: 🤖 **Claude** · 🔴 **Important**

📁 **File-level observation**

Large-scale duplication of v1 presentational code into v2 rather than sharing it. I diffed these against their `~community/crm/...` counterparts and they are byte-identical apart from renaming `interface Props` to `interface XProps`: `SidePanelHeaderInfoItem`, `SkeletonShape`, `SidePanelHeaderSkeleton`, `SidePanelHeaderActionsSkeleton`, `SidePanelTabsSkeleton`, `SidePanelWrapper`, `CrmDeleteModalContent` (plus two optional aria props), `SelectableSearchField`, `SidePanelCompanyHeader`, `companyValidations`, `useGetPriorityOptions`. Near-identical (type swap only): `SelectedOwnerField`, `AddDealContactSearch`, `TaskRow`/`TaskRowCheckbox`/`TaskRowContent`/`TaskRowMeta`/`TaskRowSubtitle`, `SidePanelMetricCards`, `CompanyModalForm`, and `taskUtil`'s `getDueDateStatus`/`getTaskTypeIcon`/`getPriorityConfig` (v1 `TASK_TYPE_ICON_MAP` is now also `TASK_TYPE_ICONS` in v2 `taskConstants`). Two secondary problems come with it: (a) atomic-design categories drifted — `SidePanelWrapper` moved from v1 `atoms/` to v2 `templates/` and `SidePanelHeaderInfoItem` from `atoms/` to `molecules/`, for identical markup; (b) the new 3-line `v2/regex/crmRegexPatterns.ts` re-exports the name `isValidCompanyWebsiteUrl` while `v2/utils/dealValidations.ts:8` still imports `isDealNameValid` from the v1 `~community/crm/regex/crmRegexPatterns`, so v2 code now pulls regexes from two same-named modules. The purely presentational pieces (skeletons, `SkeletonShape`, `SidePanelWrapper`, `CrmDeleteModalContent`, `SidePanelHeaderInfoItem`, `SelectableSearchField`) carry no v1/v2 type coupling at all and should be imported from the existing v1 paths (or promoted to `~community/crm/components/`) instead of forked — otherwise every future fix has to land twice.

frontend/src/community/crm/v2/components/molecules/TaskRow/TaskRow.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+  const handleToggleChange = (nextIsCompleted: boolean) => {
+    setTasks(updateTask(tasks, taskId, { isCompleted: nextIsCompleted }));
+
+    updateCompletion({ id: taskId, payload: { isCompleted: nextIsCompleted } });
+  };
+
+  return (
+    <div
+      role="button"
```
ThinuwanW, line 70: 🤖 **Claude** · 🔴 **Important**

The task row advertises itself as an interactive control that does nothing. It renders `role="button"`, `tabIndex={0}`, `cursor-pointer`, a hover background and `aria-label={translateText(['openTaskDetails'], { name: task.name })}` ("Open task details panel for X"), but `onRowClick` is optional and its only caller, `SidePanelTasksList`, renders `<TaskRow key={taskId} taskId={taskId} />` with no handler — and the v2 `TaskModalController` only handles `ADD_TASK_MODAL`, so there is no detail surface to open. Every task in the company panel is therefore a keyboard-focusable button that a screen reader announces as actionable and that silently no-ops. Separately, the `onKeyDown` handler matches `e.key === " "` without calling `e.preventDefault()`, so pressing Space on a focused row scrolls the panel. Either wire the task detail panel or drop `role`/`tabIndex`/`aria-label`/`cursor-pointer` until it exists.

frontend/src/community/crm/v2/components/organisms/CompanySidePanel/CompanySidePanel.tsx
```diff
@@ -0,0 +1,344 @@
...
+    isFetchingNextPage: isFetchingNextTasksPage
+  } = useGetTasksInfinite(taskFilters);
+  const {
+    data: fetchedDeals,
+    isLoading: isDealsLoading,
+    fetchNextPage: fetchNextDealsPage,
+    hasNextPage: hasNextDealsPage,
+    isFetchingNextPage: isFetchingNextDealsPage
+  } = useGetDealsInfinite(dealFilters);
+  const {
```
ThinuwanW, line 148: 🤖 **Claude** · 🟡 **Suggestion**

All three tabs' data is fetched eagerly and the whole panel blocks on the slowest query. `useGetTasksInfinite`, `useGetDealsInfinite` and `useGetContactsInfinite` are called unconditionally with no `enabled` flag, even though only one of the three sections is mounted at a time (`activeTab`), and `isLoading` ORs all five queries — so the default Tasks tab cannot paint until the deals and contacts requests also resolve. Fix: pass `enabled: activeTab === CrmSidePanelTabEnum.DEALS` (etc.) so each tab fetches on first activation, and compute the skeleton gate from the company/metrics queries plus only the active tab's query.

frontend/src/community/crm/v2/components/organisms/CompanyTable/CompanyTable.tsx
```diff
@@ -0,0 +1,208 @@
...
+
+  const hasSearchTerm = debouncedSearch !== "";
+
+  useEffect(() => {
+    if (!data) return;
+
+    const items = data.pages.flatMap((page) => page.items);
+    const normalized = normalizeCompanies(items);
+
+    setCompanies({ ...companies, ...normalized.companies });
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

This write replaces company objects wholesale instead of merging them, discarding relations hydrated elsewhere. `normalizeCompanies` builds a fresh record (`companies[id] = company`) and the top-level spread only dedupes by id, so each refetch swaps the whole entity — dropping `taskIds`, `dealIds`, `contactIds` and the detail-endpoint `metrics` that `CompanySidePanel` wrote, including the id `handleDealCreated` just appended. The same file's `mergeCompanies(existing, incoming)` in `companyUtil.ts:143` already implements the correct per-entity merge and is used by three other v2 components. Use `setCompanies(mergeCompanies(companies, items))` here and delete `normalizeCompanies` (keep only the id-list extraction, which `toContactIds`/`toDealIds` already model).

frontend/src/community/crm/v2/components/molecules/SidePanelTasksSection/SidePanelTasksSection.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const { guardCrmCreate } = useCrmLimitGuard();
+
+  const { setIsTaskModalOpen, setTaskModalType } = useCrmStoreV2(
+    useShallow((store) => ({
+      setIsTaskModalOpen: store.setIsTaskModalOpen,
+      setTaskModalType: store.setTaskModalType
+    }))
+  );
+
+  const translateText = useTranslator(
```
ThinuwanW, line 38: 🤖 **Claude** · 🟡 **Suggestion**

i18n namespace violation: the Company side panel pulls its task copy out of the Contacts namespace. This component, `SidePanelTasksList` and `TaskRowSubtitle` all call `useTranslator("crmModule", "contacts", "contactDetailsPanel", "tasks")`, while the panel's own copy lives under `crmModule.companies.sidePanel` — so the empty-state *title* comes from `contacts.contactDetailsPanel.tasks.emptyTitle` and the *description* from `companies.sidePanel.tasks.emptyDescription` (passed in by `CompanySidePanel`) in the same view. Editing the contact panel's wording will silently change the company panel. Move the shared task strings (`emptyTitle`, `addTaskButtonEmptyView`, `dueDateOverdue`/`dueDateToday`/`dueDateDueOn`) to a neutral namespace such as `crmModule.tasks.*` (which already exists and holds `openTaskDetails`/`toggleError*`) and read them from there.

frontend/src/community/crm/v2/components/molecules/SidePanelTasksSection/SidePanelTasksSection.tsx
```diff
@@ -0,0 +1,86 @@
...
+}
+
+const SidePanelTasksSection: FC<SidePanelTasksSectionProps> = ({
+  taskIds,
+  emptyDescription,
+  hasNextPage = false,
+  isFetchingNextPage = false,
+  onFetchNextPage
+}) => {
+  const { guardCrmCreate } = useCrmLimitGuard();
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`isCheckingCrmLimit` is discarded here while the sibling `SidePanelDealSection` uses it. `useCrmLimitGuard.guardCrmCreate` computes `isLimited = limitationData?.[resource]?.limitReached`, so while the enterprise limitation query is still pending it is `undefined` and the guard calls `onAllowed()`. With no disabled/loading state on either Add-task button, clicking early opens the modal even when the task limit is reached; the user fills the form and only then hits the server-side `validateTaskCreationLimit` rejection. Destructure `isCheckingCrmLimit` and apply `disabled`/`isLoading` to both the `SidePanelTasksList` button and the `EmptyDataView` button, matching `SidePanelDealSection`.

frontend/src/community/crm/v2/components/molecules/SidePanelCompanyHeader/SidePanelCompanyHeader.tsx
```diff
@@ -0,0 +1,86 @@
...
+          icon={
+            <BrowserIcon
+              width="20"
+              height="20"
+              fill="var(--color-secondary-icon)"
+            />
+          }
+          value={website}
+          endIcon={IconName.POP_OUT_ICON}
+          onClick={() => openInNewTab(website)}
```
ThinuwanW, line 43: 🤖 **Claude** · 🟡 **Suggestion**

`openInNewTab(website)` passes the stored value straight to `window.open`, which resolves scheme-less strings against the current origin. This PR tightens the frontend regex to require `^https://`, but every company created before it passed the v1 pattern `^(https:\/\/)?...` where the scheme was optional, so existing rows can hold `acme.com`. For those, clicking the website link opens `/community/crm/acme.com` in a new tab rather than the site. Normalize before opening, e.g. `openInNewTab(/^https?:\/\//i.test(website) ? website : \`https://${website}\`)`, and consider a one-off backfill for rows without a scheme.

frontend/src/community/crm/v2/utils/contactUtil.ts
```diff
@@ -26,6 +36,15 @@ export const getContactDisplayName = (
   return [contact.firstName, contact.lastName].filter(Boolean).join(" ");
 };
 
+export const getContactNameById = (
+  contacts: CrmContactRecord,
+  contactId?: number
+) => {
+  if (contactId !== undefined) {
+    return getContactDisplayName(contacts[contactId]);
```
ThinuwanW, line 57: 🤖 **Claude** · 🟡 **Suggestion**

`getContactNameById` returns `""` (not `undefined`) for a contact that is not in the store, and every consumer reads that as "nothing selected". In `TaskModalForm`, `handleDealSelect` sets `contactId` from `deals[dealId]?.contactId`, but the deal lookup effect only hydrates the *deals* record — so that contact is typically absent, `selectedContactName` is `""`, and `SelectableSearchField` falls back to the search input even though `values.contactId` is set. The user sees an empty contact box with no clear button while the deal filter is silently scoped to a contact they cannot see or unset (it only self-corrects after the next contact-lookup round trip). Make the helper return `undefined` when `contacts[contactId]` is missing, and have `TaskModalForm` distinguish "selected but not yet resolved" (render a disabled/loading field) from "nothing selected".

frontend/src/community/crm/v2/components/molecules/SidePanelDealSection/DealAccordionItemBadge.tsx
```diff
@@ -0,0 +1,44 @@
...
+  );
+
+  const stages = useCrmStoreV2((store) => store.stages);
+
+  if (deal.stageId !== undefined) {
+    const stage = stages[deal.stageId];
+
+    if (stage?.name !== undefined) {
+      const stageName =
+        stage.name in DefaultStageNameEnum
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

This reimplements `useStageNameMapper`, which already exists and is used by two other v2 components (`DealsTableV2`, `DealPropertiesSidebar`). It duplicates both the `useTranslator("crmModule", "deals", "defaultStageNames")` call and the fallback logic, and it uses a subtly different test: `stage.name in DefaultStageNameEnum` checks enum *keys*, whereas the hook does `Object.values(DefaultStageNameEnum).includes(...)` and checks *values*. They agree today only because the enum is 1:1 key↔value; rename a single member and the two will disagree silently. Replace the local logic with `const { getStageByName } = useStageNameMapper();`.

frontend/src/community/crm/v2/constants/taskConstants.ts
```diff
@@ -1,12 +1,28 @@
...
+  MediumPriorityIcon,
+  MeetingFilledIcon,
+  PhoneFilledIcon
 } from "@rootcodelabs/skapp-ui";
-import { FC } from "react";
+import { ComponentType, FC, SVGProps } from "react";
 
 import { CrmPriorityEnum } from "~community/crm/v2/enums/common";
 
+export const TASK_TYPE_ICON_SIZE = 20;
```
ThinuwanW, line 14: 🤖 **Claude** · 🟡 **Suggestion**

Several symbols are added or kept but never read. `TASK_TYPE_ICON_SIZE = 20` has no consumer — `taskUtil.getTaskTypeIcon(typeName = "", size = 20)` hardcodes the same magic number instead. `PRIORITY_OPTIONS[].key` is now dead in v2: the copied `useGetPriorityOptions` switched from `option.key` to `option.value`/`option.value.toLowerCase()`. In `api/utils/QueryKeys.ts`, `crmTaskQueryKeys.ALL` and `crmCompanyQueryKeys.ALL` are declared and never used or invalidated. In `api/ContactApi.ts`, `useGetContactMetrics` (plus `crmContactEndpoints.GET_CONTACT_METRICS` and `crmContactQueryKeys.METRICS`) has no caller anywhere in the app — yet `TaskApi.useCreateTask` and `DealApi.useCreateDeal` both invalidate `crmContactQueryKeys.METRICS(id)`, i.e. a key nothing subscribes to. Finally, `SidePanelWrapper`'s `isOpen` prop and `TaskRow`'s `onRowClick`/`isCheckTaskVisible` props are never passed by any caller. Delete what is unused and wire `TASK_TYPE_ICON_SIZE` into `getTaskTypeIcon`'s default.

frontend/src/community/crm/v2/components/molecules/SidePanelMetricCards/SidePanelMetricCards.tsx
```diff
@@ -0,0 +1,33 @@
...
+        <Card
+          key={metric.id}
+          className="flex flex-col gap-1 flex-1 min-w-0 p-3 overflow-hidden border border-secondary-accent rounded-lg bg-white"
+        >
+          <p className="body2 text-secondary-text">{metric.title}</p>
+          <div className="flex items-center gap-2">
+            <p className="subtitle2">
+              {metric.isCurrency
+                ? formatMonetaryValueWithDecimals(metric.amount)
+                : metric.amount}
```
ThinuwanW, line 24: 🤖 **Claude** · 🟡 **Suggestion**

Non-currency metrics bypass formatting, so a missing count renders as an empty card. `getCompanyMetricItems` supplies `openDealsCount`/`closedDealsCount` with `isCurrency` unset, and this branch renders `{metric.amount}` raw: `undefined` (metrics not yet hydrated, or a partial response) renders nothing at all, whereas the `isCurrency` branch would show `-`. The result is two adjacent cards with an em-dash and two that are blank. Route the non-currency branch through `formatTableValue(metric.amount)` so the placeholder is consistent. (Note the v2 copy also dropped v1's `MetricChipLabel`, but `CompanySidePanelSkeleton` still renders a rounded chip placeholder in the first card — the skeleton no longer matches what loads in.)

frontend/src/community/crm/v2/utils/__test__/sidePanelSectionUtil.test.ts
```diff
@@ -0,0 +1,19 @@
+import { CrmDealStageEnum } from "~community/crm/v2/enums/common";
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Test coverage and layout problems in the new tests. (1) Roughly 25 new components ship with no tests: `CompanyTable`, `CompanySidePanel`, `CompanyModalForm`, `TaskModalForm`, all five `TaskRow*` files, all four modal-content components, `CompanyModalController`, `TaskModalController`, and every `SidePanel*Section`. Only four pure-util test files were added, so none of the store wiring, mutation callbacks or role gating described in the other findings is covered — per the guidelines these need `.test.tsx` files with the `MockTheme` wrapper and `screen.getByRole` queries. (2) This file is named `sidePanelSectionUtil.test.ts` but tests `getInitialStageId` from `dealUtil` — there is no `sidePanelSectionUtil` module; rename it `dealUtil.test.ts` or fold it into the existing deal tests. (3) The four new files live in `utils/__test__/` while the module's existing v2 tests (`utils/boardUtil.test.ts`, `utils/dealIngest.test.ts`, `hooks/useInitializeCrmData.test.ts`) sit beside their sources; pick one convention.

frontend/src/community/crm/v2/api/TaskApi.ts
dulsara-manakal: This will fixed in this [PR](https://github.com/SkappHQ/skapp/pull/1794)

frontend/src/community/crm/v2/api/TaskApi.ts
dulsara-manakal: This will fixed in this [PR](https://github.com/SkappHQ/skapp/pull/1794)

frontend/src/community/crm/v2/components/molecules/AddTaskModalContent/AddTaskModalContent.tsx
dulsara-manakal: No need to invalidate this scenario, it is handled in store.

frontend/src/community/crm/v2/components/molecules/TaskRow/TaskRow.tsx
dulsara-manakal: Fixed it allign to  V1

frontend/src/community/crm/v2/regex/crmRegexPatterns.ts
dulsara-manakal: Actually there is an issue in V1 regex, so I fixed it in V2. The website link always must contain "https"

frontend/src/community/crm/v2/components/molecules/TaskModalForm/TaskModalForm.tsx
dulsara-manakal: Fixed this. Sales reps actually cannot change the owner their owner will be preselected as them.

frontend/src/community/crm/v2/utils/dealUtil.ts
dulsara-manakal: This handle is at the page level because we added a skeleton to the pages.

frontend/src/community/crm/v2/utils/commonUtil.ts
dulsara-manakal: Fixed this.

frontend/src/community/crm/v2/api/TaskApi.ts
dulsara-manakal: Added a guard following the same pattern in other places in Skapp.

frontend/src/community/crm/v2/api/CompanyApi.ts
dulsara-manakal: we dont want to invalidate we handle it with responce replacing the store

frontend/src/community/crm/v2/components/templates/SidePanelWrapper/SidePanelWrapper.tsx
dulsara-manakal: This did have an impact to the content layout and side bar of MUI

frontend/src/community/crm/v2/utils/commonUtil.ts
dulsara-manakal: The requirements differ from place to place, that's the reason.

frontend/src/community/crm/v2/components/molecules/SidePanelAddDeal/SidePanelAddDeal.tsx
dulsara-manakal: This will not reach the guard on initial data retrivals.

frontend/src/community/crm/v2/components/molecules/SidePanelTasksSection/SidePanelTasksSection.tsx
dulsara-manakal: Fixed this.

frontend/pages/community/crm/companies.tsx
dulsara-manakal: This is the decision we took for this CRM refactoring.

frontend/src/community/crm/v2/components/molecules/SidePanelAddDeal/AddDealContactSearch.tsx
dulsara-manakal: They're not the same widget. Prop surfaces barely overlap.

frontend/src/community/crm/v2/components/molecules/TaskModalForm/TaskModalForm.tsx
dulsara-manakal: Added tests only for utils and necessary hooks.

frontend/src/community/crm/v2/utils/__test__/sidePanelSectionUtil.test.ts
dulsara-manakal: The test is added only for utils and necessary hooks, no need for a UI component.

frontend/src/community/crm/v2/components/molecules/SidePanelTasksSection/SidePanelTasksSection.tsx
dulsara-manakal: this translation key issue exists everywhere, so  we have seperate task to fix this.

frontend/src/community/crm/v2/components/organisms/CompanyTable/CompanyTable.tsx
dulsara-manakal: Use useMemo for this, not the consistent pattern in Skapp, and we actually need to render the table every time with fresh data.

frontend/src/community/crm/v2/components/organisms/CompanySidePanel/CompanySidePanel.tsx
dulsara-manakal: This is the requirement.

frontend/src/community/crm/v2/components/molecules/SidePanelDealSection/DealAccordionItemBadge.tsx
dulsara-manakal: used the useStageNameMapper

frontend/src/community/crm/v2/components/molecules/SidePanelMetricCards/SidePanelMetricCards.tsx
dulsara-manakal: This will not reach the guard with isLoading on the top of the side panel.

frontend/src/community/crm/v2/components/molecules/SidePanelCompanyHeader/SidePanelCompanyHeader.tsx
dulsara-manakal: That's fine with the requirement.

frontend/src/community/crm/v2/components/molecules/SidePanelTasksSection/SidePanelTasksSection.tsx
dulsara-manakal: Fixed.

frontend/src/community/crm/v2/components/organisms/CompanyTable/CompanyTable.tsx
dulsara-manakal: Added the merging part.

frontend/src/community/crm/v2/utils/contactUtil.ts
dulsara-manakal: in the dropdown, we  dont want to  hydrate the whole object id is enough.

frontend/src/community/crm/v2/api/utils/ApiEndpoints.ts
dulsara-manakal: Fixed this.

frontend/src/community/crm/v2/constants/taskConstants.ts
dulsara-manakal: Fixed this.

frontend/src/community/crm/v2/components/molecules/AddTaskModalContent/AddTaskModalContent.tsx
dulsara-manakal: Known and accepted. The BE derives the company from contactId only, we don't get the companyId on task creation. Tracking separately.

frontend/src/community/crm/v2/components/organisms/CompanySidePanel/CompanySidePanel.tsx
dulsara-manakal: Correct against development. This targets refactor/crm-task-api-refactoring, which makes GET /v2/crm/task paginated (CrmTaskFilterDtoV2 + toPageDto).

frontend/src/community/crm/v2/components/molecules/TaskModalForm/TaskModalForm.tsx
dulsara-manakal: Valid, fixed. Added mergeContacts and switched both effects to mergeContacts/mergeDeals.

frontend/src/community/crm/v2/components/molecules/SidePanelAddDeal/SidePanelAddDeal.tsx
dulsara-manakal: Guarded in practice by the new page-level init-data loader.

frontend/src/community/crm/v2/components/molecules/SidePanelAddDeal/SidePanelAddDeal.tsx
dulsara-manakal: v1 was changed to validateOnChange: false / validateOnBlur: true, and the touched gate removed in CRM-501; this matches current v1.

frontend/src/community/crm/v2/components/molecules/TaskRow/TaskRowMeta.tsx
dulsara-manakal: Split. The init-data-backed reads (owners, taskTypes) were a real race - fixed at source with a page-level loader.

frontend/pages/community/crm/companies.tsx
dulsara-manakal: Valid, fixed. Return value destructured and the page body held behind FullScreenLoader until init-data resolves.

frontend/src/community/crm/v2/components/organisms/CompanySidePanel/CompanySidePanel.tsx
dulsara-manakal: Fixed.

frontend/src/community/crm/v2/components/molecules/SidePanelHeaderInfoItem/SidePanelHeaderInfoItem.tsx
dulsara-manakal: v2 is a shadow build that replaces v1 at cutover; importing v1 components would block deleting it.

frontend/src/community/crm/v2/components/molecules/TaskRow/TaskRow.tsx
dulsara-manakal: That will used in task page.

frontend/src/community/crm/v2/constants/taskConstants.ts
dulsara-manakal: Keep onRowClick since we need it in the task page.

PR #1826, opened 2026-08-24
[Feat] Add job family and job title to birthday notifications API

akilarootcode:
## Summary

Adds an employee's job family and job title to the "today's birthday notifications" API response, so consumers can display an employee's role alongside their birthday without a follow-up lookup.

- Added `EmployeeBirthdayResponseDto` (`employeeId`, `firstName`, `lastName`, `middleName`, `authPic`, `jobFamily`, `jobTitle`) as the element type of `BirthdayNotificationResponseDto.employeeBirthdays`, replacing the shared `EmployeeBasicDetailsResponseDto` for this one response so the ~20 other consumers of that shared DTO are unaffected.
- Added `PeopleMapper.employeeToEmployeeBirthdayResponseDto`, mapping `jobFamily.name` / `jobTitle.name` the same way `employeeToEmployeeSignInResponseDto` already does.
- `PeopleServiceImpl.getTodayBirthdayNotifications` now maps through the new method.
- `EmployeeRepositoryImpl.findEmployeeBirthdaysOnByViewerAndScope` fetch-joins `Employee_.jobFamily` and `Employee_.jobTitle` (`LEFT`) so both associations come back in the existing query instead of triggering per-employee lazy loads.
- Extended `PeopleBirthdayNotificationControllerIntegrationTest` to assert `jobFamily`/`jobTitle` on the existing SELF-scope case, and added a case covering an employee with neither association set (both fields serialize as `null`).

## Architecture / flow

```mermaid
flowchart LR
    A[GET /v1/people/birthday-notifications/today] --> B[PeopleController]
    B --> C[PeopleServiceImpl.getTodayBirthdayNotifications]
    C --> D[EmployeeRepositoryImpl\nCriteria query + LEFT JOIN FETCH\njobFamily, jobTitle]
    D --> E[PeopleMapper.employeeToEmployeeBirthdayResponseDto]
    E --> F[EmployeeBirthdayResponseDto\n+ jobFamily, jobTitle]
```

- The frontend `EmployeeBirthdayType` and birthday modal components are unchanged. They ignore unknown response fields today, so nothing breaks, but rendering job family/title in the birthday modals is a separate UI decision this PR doesn't make.
- `EmployeeBasicDetailsResponseDto` (the shared DTO used by ~20 other responses) is untouched — a new, narrower DTO was added instead of widening a shared one.

- **Submodule hash / pointer bumps** (`backend/src/main/java/com/skapp/enterprise`, `backend/src/main/resources/config`) are intentionally excluded. Gitlink updates are done at **deployment time**, not in feature PRs.

## Testing

| Check | Ran? | Result |
|-------|------|--------|
| Unit / integration tests (`PeopleBirthdayNotificationControllerIntegrationTest`) | yes | pass (15/15) |
| Formatter (spring-javaformat validate) | yes | pass |
| Checkstyle | yes | pass |
| e2e (Playwright) | no | No existing birthday e2e spec and no frontend change in this PR; the additive API field is fully covered by the integration tests above |

### DIFF
--- a/backend/src/main/java/com/skapp/community/peopleplanner/mapper/PeopleMapper.java
+++ b/backend/src/main/java/com/skapp/community/peopleplanner/mapper/PeopleMapper.java
@@ -46,6 +46,7 @@ import com.skapp.community.peopleplanner.payload.request.employee.personal.Emplo
 import com.skapp.community.peopleplanner.payload.request.employee.personal.EmployeePersonalFamilyDetailsDto;
 import com.skapp.community.peopleplanner.payload.response.CreateEmployeeResponseDto;
 import com.skapp.community.peopleplanner.payload.response.EmployeeAllDataExportResponseDto;
+import com.skapp.community.peopleplanner.payload.response.EmployeeBirthdayResponseDto;
 import com.skapp.community.peopleplanner.payload.response.EmployeeDataExportResponseDto;
 import com.skapp.community.peopleplanner.payload.response.EmployeeDetailedResponseDto;
 import com.skapp.community.peopleplanner.payload.response.EmployeeJobFamilyDto;
@@ -194,6 +195,10 @@ public interface PeopleMapper {
 	EmployeeSignInResponseDto employeeToEmployeeSignInResponseDto(Employee employee);
 
 	EmployeeBasicDetailsResponseDto employeeToEmployeeBasicDetailsResponseDto(Employee employee);
+
+	@Mapping(target = "jobFamily", source = "jobFamily.name")
+	@Mapping(target = "jobTitle", source = "jobTitle.name")
+	EmployeeBirthdayResponseDto employeeToEmployeeBirthdayResponseDto(Employee employee);
 
 	List<HolidayBasicDetailsResponseDto> holidaysToHolidayBasicDetailsResponseDtos(List<Holiday> holidays);
 
--- a/backend/src/main/java/com/skapp/community/peopleplanner/mapper/PeopleMapper.java
+++ b/backend/src/main/java/com/skapp/community/peopleplanner/mapper/PeopleMapper.java
@@ -200,6 +200,8 @@ public interface PeopleMapper {
 	@Mapping(target = "jobTitle", source = "jobTitle.name")
 	EmployeeBirthdayResponseDto employeeToEmployeeBirthdayResponseDto(Employee employee);
 
+	List<EmployeeBirthdayResponseDto> employeesToEmployeeBirthdayResponseDtos(List<Employee> employees);
+
 	List<HolidayBasicDetailsResponseDto> holidaysToHolidayBasicDetailsResponseDtos(List<Holiday> holidays);
 
 	JobTitleDto jobTitleToJobTitleDto(JobTitle jobTitle);
--- a/backend/src/main/java/com/skapp/community/peopleplanner/payload/response/BirthdayNotificationResponseDto.java
+++ b/backend/src/main/java/com/skapp/community/peopleplanner/payload/response/BirthdayNotificationResponseDto.java
@@ -1,6 +1,5 @@
 package com.skapp.community.peopleplanner.payload.response;
 
-import com.skapp.community.peopleplanner.payload.request.EmployeeBasicDetailsResponseDto;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
 import lombok.Setter;
@@ -15,6 +14,6 @@ public class BirthdayNotificationResponseDto {
 
 	private LocalDate lastViewedDate;
 
-	private List<EmployeeBasicDetailsResponseDto> employeeBirthdays;
+	private List<EmployeeBirthdayResponseDto> employeeBirthdays;
 
 }
--- a/backend/src/main/java/com/skapp/community/peopleplanner/payload/response/EmployeeBirthdayResponseDto.java
+++ b/backend/src/main/java/com/skapp/community/peopleplanner/payload/response/EmployeeBirthdayResponseDto.java
@@ -1,21 +1,12 @@
 package com.skapp.community.peopleplanner.payload.response;
 
+import com.skapp.community.peopleplanner.payload.request.EmployeeBasicDetailsResponseDto;
 import lombok.Getter;
 import lombok.Setter;
 
 @Getter
 @Setter
-public class EmployeeBirthdayResponseDto {
-
-	private Long employeeId;
-
-	private String firstName;
-
-	private String lastName;
-
-	private String middleName;
-
-	private String authPic;
+public class EmployeeBirthdayResponseDto extends EmployeeBasicDetailsResponseDto {
 
 	private String jobFamily;
 
--- a/backend/src/main/java/com/skapp/community/peopleplanner/payload/response/EmployeeBirthdayResponseDto.java
+++ b/backend/src/main/java/com/skapp/community/peopleplanner/payload/response/EmployeeBirthdayResponseDto.java
@@ -1,0 +1,24 @@
+package com.skapp.community.peopleplanner.payload.response;
+
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+public class EmployeeBirthdayResponseDto {
+
+	private Long employeeId;
+
+	private String firstName;
+
+	private String lastName;
+
+	private String middleName;
+
+	private String authPic;
+
+	private String jobFamily;
+
+	private String jobTitle;
+
+}
--- a/backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/EmployeeRepositoryImpl.java
+++ b/backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/EmployeeRepositoryImpl.java
@@ -1674,6 +1674,8 @@ public class EmployeeRepositoryImpl implements EmployeeRepository {
 		root.fetch(Employee_.personalInfo);
 		root.fetch(Employee_.user);
 		root.fetch(Employee_.employeeRole, JoinType.LEFT);
+		root.fetch(Employee_.jobFamily, JoinType.LEFT);
+		root.fetch(Employee_.jobTitle, JoinType.LEFT);
 
 		Join<Employee, EmployeePersonalInfo> personalInfoJoin = root.join(Employee_.personalInfo);
 		Join<Employee, User> userJoin = root.join(Employee_.user);
--- a/backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
@@ -1613,9 +1613,8 @@ public class PeopleServiceImpl implements PeopleService {
 			return new ResponseEntityDto(false, new BirthdayNotificationResponseDto(lastViewedDate, List.of()));
 		}
 
-		List<EmployeeBirthdayResponseDto> response = employeesWithBirthdays.stream()
-			.map(peopleMapper::employeeToEmployeeBirthdayResponseDto)
-			.toList();
+		List<EmployeeBirthdayResponseDto> response = peopleMapper
+			.employeesToEmployeeBirthdayResponseDtos(employeesWithBirthdays);
 
 		log.info("getTodayBirthdayNotifications: execution ended");
 		return new ResponseEntityDto(false, new BirthdayNotificationResponseDto(lastViewedDate, response));
--- a/backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
+++ b/backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
@@ -93,6 +93,7 @@ import com.skapp.community.peopleplanner.payload.response.BirthdayNotificationRe
 import com.skapp.community.peopleplanner.payload.response.BirthdayNotificationViewedResponseDto;
 import com.skapp.community.peopleplanner.payload.response.CreateEmployeeResponseDto;
 import com.skapp.community.peopleplanner.payload.response.EmployeeAllDataExportResponseDto;
+import com.skapp.community.peopleplanner.payload.response.EmployeeBirthdayResponseDto;
 import com.skapp.community.peopleplanner.payload.response.EmployeeBulkErrorResponseDto;
 import com.skapp.community.peopleplanner.payload.response.EmployeeBulkResponseDto;
 import com.skapp.community.peopleplanner.payload.response.EmployeeCountDto;
@@ -1612,8 +1613,8 @@ public class PeopleServiceImpl implements PeopleService {
 			return new ResponseEntityDto(false, new BirthdayNotificationResponseDto(lastViewedDate, List.of()));
 		}
 
-		List<EmployeeBasicDetailsResponseDto> response = employeesWithBirthdays.stream()
-			.map(peopleMapper::employeeToEmployeeBasicDetailsResponseDto)
+		List<EmployeeBirthdayResponseDto> response = employeesWithBirthdays.stream()
+			.map(peopleMapper::employeeToEmployeeBirthdayResponseDto)
 			.toList();
 
 		log.info("getTodayBirthdayNotifications: execution ended");
--- a/backend/src/test/java/com/skapp/community/peopleplanner/controller/v1/PeopleBirthdayNotificationControllerIntegrationTest.java
+++ b/backend/src/test/java/com/skapp/community/peopleplanner/controller/v1/PeopleBirthdayNotificationControllerIntegrationTest.java
@@ -77,6 +77,10 @@ class PeopleBirthdayNotificationControllerIntegrationTest {
 
 	private static final String CURRENT_USER_EMAIL = "user1@gmail.com";
 
+	private static final String CURRENT_EMPLOYEE_JOB_FAMILY = "Software Engineer";
+
+	private static final String CURRENT_EMPLOYEE_JOB_TITLE = "Associate";
+
 	private static final Long CURRENT_EMPLOYEE_ID = 1L;
 
 	private static final Long TEAMMATE_EMPLOYEE_ID = 2L;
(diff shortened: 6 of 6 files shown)


ThinuwanW (commented):
## 🤖 Claude Code Review

Clean, well-scoped additive change: two new LEFT fetch joins prevent N+1 when the new `jobFamily`/`jobTitle` fields are mapped, the MapStruct nested-source mapping mirrors the existing `employeeToEmployeeSignInResponseDto` pattern and is null-safe, and the integration test covers both the populated case (seed data for employee 1 is `job_family_id=1` → 'Software Engineer', `job_title_id=1` → 'Associate', so the new constants are correct) and the null case. Security, transactional boundaries, logging, naming, and package placement all follow the established conventions, and no enterprise override or other caller of the changed mapper/DTO exists. Only minor, non-blocking observations below.

Found **8** new issue(s): 🟡 3 suggestion(s), 🟣 5 nit(s)

backend/src/main/java/com/skapp/community/peopleplanner/payload/response/EmployeeBirthdayResponseDto.java
```diff
@@ -0,0 +1,24 @@
+package com.skapp.community.peopleplanner.payload.response;
```
ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

This DTO restates all five fields of `EmployeeBasicDetailsResponseDto` (employeeId, firstName, lastName, middleName, authPic) verbatim and only adds `jobFamily`/`jobTitle`. Creating a dedicated DTO rather than widening the shared one is the right call here (widening would force the extra fetch joins on every leave/time/PM endpoint that reuses it), but the field duplication itself is avoidable: `public class EmployeeBirthdayResponseDto extends EmployeeBasicDetailsResponseDto` with only the two new fields would keep the shape in sync automatically. The repo already uses this pattern for response DTOs (e.g. `AllLeaveRequestsResponseDto extends LeaveRequestResponseDto`, `LeaveRequestByIdResponseDto extends LeaveRequestResponseDto`, `ManagerTeamResponseDto extends ManagerSummarizedTeamResponseDto`). Note that if you do this, MapStruct still generates the base-field mappings, and the `@Mapping` annotations on the mapper method stay unchanged. If you keep the standalone class, consider a brief comment noting the intentional divergence so a future field added to `EmployeeBasicDetailsResponseDto` isn't silently missed here.

backend/src/main/java/com/skapp/community/peopleplanner/payload/response/BirthdayNotificationResponseDto.java
```diff
@@ -15,6 +14,6 @@ public class BirthdayNotificationResponseDto {
 
 	private LocalDate lastViewedDate;
 
-	private List<EmployeeBasicDetailsResponseDto> employeeBirthdays;
+	private List<EmployeeBirthdayResponseDto> employeeBirthdays;
```
ThinuwanW, line 17: 🤖 **Claude** · 🟣 **Nit**

This changes the wire contract for `GET /v1/people/birthday-notifications/today`, but the TypeScript consumer of that contract in the same repo was not updated: `frontend/src/community/people/types/BirthdayNotificationTypes.ts` still declares `EmployeeBirthdayType` as `{ employeeId, firstName, lastName, authPic, email }`, so `jobFamily`/`jobTitle` are unreachable from `useBirthdayNotifications`, `birthdayNotificationUtils`, and the birthday modals — the new backend fields are effectively dead until that follow-up lands. Worth confirming the frontend change is a tracked companion PR rather than an oversight. While there, note the same type declares `email: string`, which neither the old `EmployeeBasicDetailsResponseDto` nor the new `EmployeeBirthdayResponseDto` ever returns — pre-existing contract drift that this PR is a natural opportunity to correct (add `jobFamily: string | null; jobTitle: string | null;` and drop or make optional `email`).

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/EmployeeRepositoryImpl.java
```diff
@@ -1674,6 +1674,8 @@ public List<Employee> findEmployeeBirthdaysOnByViewerAndScope(LocalDate date, Lo
 		root.fetch(Employee_.personalInfo);
```
ThinuwanW, line 1674: 🤖 **Claude** · 🟣 **Nit**

Pre-existing, but this PR edits exactly these lines so it's worth noting. `personalInfo`, `user`, and `employeeRole` are each both fetched (lines 1674-1676) and joined again (lines 1680-1682). Hibernate does not de-duplicate a `Fetch` node against a separately-created `Join` node, so the generated SQL contains two joins per association — one for the fetch and one for the predicate — meaning three redundant joins on this query. The two newly added lines (jobFamily/jobTitle) correctly avoid the problem because they are fetch-only. The fix is to derive the join from the fetch instead of creating a second one, e.g. `Join<Employee, EmployeePersonalInfo> personalInfoJoin = (Join<Employee, EmployeePersonalInfo>) root.fetch(Employee_.personalInfo);` (Hibernate's `SqmSingularJoin` implements both `Fetch` and `Join`, so the cast is safe) and likewise for `userJoin` and `roleJoin` — remembering to preserve `JoinType.LEFT` on the `employeeRole` one so guest-role filtering semantics don't change. Low impact given the result-set size, so fine to defer, but it will keep growing as more fetches are added to this method.

backend/src/main/java/com/skapp/community/peopleplanner/payload/response/EmployeeBirthdayResponseDto.java
ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Contract gap in the opposite direction from the one already noted. The TypeScript type for this payload (`frontend/src/community/people/types/BirthdayNotificationTypes.ts`, `EmployeeBirthdayType`) declares `email: string` as a **required** field, and the frontend test fixture in `birthdayNotificationUtils.test.ts` supplies it — but neither this new DTO nor the `EmployeeBasicDetailsResponseDto` it replaces has ever contained an `email` field, so `employee.email` is `undefined` at runtime for every entry. It happens to be unused today (nothing in `BirthdayModalShell`/`ColleagueBirthdayModal`/`SelfBirthdayModal` reads it), which is why it has gone unnoticed, but this PR is the moment the payload is being redefined, so it should be resolved rather than carried forward. Fix either by adding `private String email;` to this DTO plus `@Mapping(target = "email", source = "user.email")` on the new mapper method (the `user` association is already fetch-joined in `findEmployeeBirthdaysOnByViewerAndScope`, so this costs nothing), or by deleting `email` from `EmployeeBirthdayType` and the test fixture. Adding it is the safer option if any consumer is expected to key off email.

backend/src/main/java/com/skapp/community/peopleplanner/mapper/PeopleMapper.java
```diff
@@ -195,6 +196,10 @@ List<ManagerSummarizedTeamResponseDto> managerTeamsToManagerTeamCountTeamRespons
 
 	EmployeeBasicDetailsResponseDto employeeToEmployeeBasicDetailsResponseDto(Employee employee);
 
+	@Mapping(target = "jobFamily", source = "jobFamily.name")
```
ThinuwanW, line 199: 🤖 **Claude** · 🟡 **Suggestion**

This PR correctly expresses the `jobFamily.name` / `jobTitle.name` projection declaratively via MapStruct, but the exact same null-safe copy is still hand-rolled in two other places in the same module: `PeopleServiceImplV2.mapToEmployeeDirectoryResponseDto` (`dto.setJobFamily(employee.getJobFamily() != null ? employee.getJobFamily().getName() : null);` and the matching `setJobTitle`, lines 59-60) and `PeopleReadServiceImpl.mapCommonDetails` (line 369). The review guidelines call out manual field-by-field mapping where MapStruct should be used. Since you are already touching this mapper, consider adding the equivalent `@Mapping(target = "jobFamily", source = "jobFamily.name")` pairs to mapper methods for `EmployeeDirectoryResponseDto` and `EmployeeCommonDetailsDto` and deleting the hand-written ternaries, so there is one place that defines how a job family/title is projected onto a response.

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/EmployeeRepositoryImpl.java
```diff
@@ -1674,6 +1674,8 @@ public List<Employee> findEmployeeBirthdaysOnByViewerAndScope(LocalDate date, Lo
...
 		root.fetch(Employee_.user);
 		root.fetch(Employee_.employeeRole, JoinType.LEFT);
+		root.fetch(Employee_.jobFamily, JoinType.LEFT);
```
ThinuwanW, line 1677: 🤖 **Claude** · 🟣 **Nit**

The two added fetch joins are the entire performance point of this change (without them the mapper would issue two extra selects per returned employee), but no test can detect their removal. `PeopleBirthdayNotificationControllerIntegrationTest` is annotated `@Transactional` at class level, so the MockMvc request joins the test-managed transaction and the Hibernate session is still open when `employeeToEmployeeBirthdayResponseDto` dereferences `getJobFamily().getName()`. Delete lines 1677-1678 and every test still passes — the lazy proxies just resolve with extra queries. If you want the joins protected against a future regression, assert the statement count around the request (e.g. `SessionFactory.unwrap(SessionFactoryImplementor.class).getStatistics().getPrepareStatementCount()` with `hibernate.generate_statistics` enabled for that test), or at minimum leave a comment on these two lines stating they exist to avoid N+1 so they are not "cleaned up" later.

backend/src/test/java/com/skapp/community/peopleplanner/controller/v1/PeopleBirthdayNotificationControllerIntegrationTest.java
```diff
@@ -77,6 +77,10 @@ class PeopleBirthdayNotificationControllerIntegrationTest {
 
 	private static final String CURRENT_USER_EMAIL = "user1@gmail.com";
 
+	private static final String CURRENT_EMPLOYEE_JOB_FAMILY = "Software Engineer";
```
ThinuwanW, line 80: 🤖 **Claude** · 🟣 **Nit**

These two new expectations are coupled to `src/test/resources/data.sql` auto-increment ordering rather than seeded by the test itself. Every other value asserted in `getTodayBirthdayNotifications_WithSelfScope_ReturnsOnlySelf` that this PR sits next to is either explicitly seeded (`mutateEmployee(CURRENT_EMPLOYEE_ID, e -> e.setAuthPic("auth-pic-one.png"))`) or a plain name column; `jobFamily`/`jobTitle` instead depend on `employee.job_family_id = 1` / `job_title_id = 1` resolving to the first `job_family` and first `job_title` rows inserted with `VALUES (default, ...)`. Note `data.sql` inserts two `job_title` rows both named 'Associate' (ids 1 and 4), so the assertion is satisfied by accident for more than one id and inserting a row earlier in the file would silently shift what is being asserted. Prefer seeding the association in the test — e.g. `mutateEmployee(CURRENT_EMPLOYEE_ID, e -> { e.setJobFamily(jobFamilyDao.getReferenceById(...)); e.setJobTitle(jobTitleDao.getReferenceById(...)); })` — or at least look the expected names up from the DAOs instead of hardcoding string literals.

backend/src/test/java/com/skapp/community/peopleplanner/controller/v1/PeopleBirthdayNotificationControllerIntegrationTest.java
```diff
@@ -368,6 +374,24 @@ void getTodayBirthdayNotifications_OnLastDayOfShortFebruary_IncludesLeapDayBirth
 			}
 		}
 
+		@Test
+		@DisplayName("Get today's birthdays when the matching employee has no job family or job title - Returns nulls for both")
+		void getTodayBirthdayNotifications_WithEmployeeWithoutJobFamilyAndJobTitle_ReturnsNullsForBoth()
+				throws Exception {
+			seedConfig(true, false, false);
+			giveBirthdayOn(CURRENT_EMPLOYEE_ID, today);
+			mutateEmployee(CURRENT_EMPLOYEE_ID, employee -> {
```
ThinuwanW, line 383: 🤖 **Claude** · 🟣 **Nit**

Coverage of the two new fields is all-or-nothing: the SELF-scope test covers both populated and this test covers both null, but the mixed state is untested even though it is reachable in production. `employee.job_family_id` and `job_title_id` are independently nullable in `common-ddl-script-v1-create-tables.sql`, MapStruct emits two independent null-safe accessors, and `JobServiceImpl.transferJobTitle` (line 363) writes only `employee.setJobTitle(...)` without touching `jobFamily`. Consider parameterising this test (`@ParameterizedTest` over jobFamily-only-null / jobTitle-only-null / both-null) so a future change to the mapping — for example switching to a single nested-object mapping or a `qualifiedByName` expression — cannot regress one field while the other still passes.

backend/src/main/java/com/skapp/community/peopleplanner/payload/response/EmployeeBirthdayResponseDto.java
akilarootcode: Fixed in 096e36266 — EmployeeBirthdayResponseDto now extends EmployeeBasicDetailsResponseDto, keeping only jobFamily/jobTitle. MapStruct output is identical (checked the generated PeopleMapperImpl); 15/15 tests and checkstyle still pass.

backend/src/main/java/com/skapp/community/peopleplanner/payload/response/BirthdayNotificationResponseDto.java
akilarootcode: Confirmed — EmployeeBirthdayType declares neither field today. Not wiring the FE in this PR; it's scoped backend-only (see PR's Exclusions section). Will pick this up if/when the birthday modals get a job-title/family display, along with the separate pre-existing `email` drift you flagged.

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/EmployeeRepositoryImpl.java
akilarootcode: Confirmed — this fetch+join duplication predates this PR (unchanged from before this commit touched the file). Not fixing it here since these are the only lines this PR should be touching; worth its own cleanup PR.

backend/src/main/java/com/skapp/community/peopleplanner/payload/response/EmployeeBirthdayResponseDto.java
akilarootcode: Confirmed — neither the old nor new DTO has ever returned `email`, and it's read nowhere in the modals. Pre-existing FE/BE contract drift unrelated to job title/family; fixing it means changing the FE contract, which is out of scope here. Tracking as separate follow-up.

backend/src/main/java/com/skapp/community/peopleplanner/mapper/PeopleMapper.java
akilarootcode: Confirmed at both locations (PeopleServiceImplV2 L59-60, PeopleReadServiceImpl L369). Real duplication, but consolidating it touches two services/endpoints unrelated to the birthday API, outside this PR's scope. Leaving for a dedicated mapper-consolidation refactor.

backend/src/main/java/com/skapp/community/peopleplanner/repository/impl/EmployeeRepositoryImpl.java
akilarootcode: Confirmed the class is @Transactional, so lazy-loading would mask a regression here. Not adding either suggested guard: a comment violates CLAUDE.md's no-comments rule, and statement-count assertions aren't a pattern used anywhere else in this suite — not worth the inconsistency for one query.

backend/src/test/java/com/skapp/community/peopleplanner/controller/v1/PeopleBirthdayNotificationControllerIntegrationTest.java
akilarootcode: Confirmed — data.sql does have two 'Associate' job_title rows, and yes it resolves by coincidence here. Leaving as-is: this test already hardcodes firstName/lastName the same way straight from seed data with no DAO lookup, so this isn't a deviation, it's the file's existing convention.

backend/src/test/java/com/skapp/community/peopleplanner/controller/v1/PeopleBirthdayNotificationControllerIntegrationTest.java
akilarootcode: Confirmed the mixed state is reachable, but employeeJobFamilyName/employeeJobTitleName in the generated mapper are fully independent, unconditional null checks — a mixed-state test can't catch anything the existing both-null/both-populated cases don't already cover. Not adding it per CLAUDE.md's "do not write unnecessary tests."

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
```diff
@@ -1612,8 +1613,8 @@ public ResponseEntityDto getTodayBirthdayNotifications() {
 			return new ResponseEntityDto(false, new BirthdayNotificationResponseDto(lastViewedDate, List.of()));
 		}
 
-		List<EmployeeBasicDetailsResponseDto> response = employeesWithBirthdays.stream()
-			.map(peopleMapper::employeeToEmployeeBasicDetailsResponseDto)
+		List<EmployeeBirthdayResponseDto> response = employeesWithBirthdays.stream()
```
erandi-rootcode: we can create a mapper as below and use it rather than stream

List<EmployeeBirthdayResponseDto> employeesToEmployeeBirthdayResponseDtos(
            List<Employee> employees);

since we already have the singular mapper and the mappings are available? even null are handled

erandi-rootcode (changes requested):

backend/src/main/java/com/skapp/community/peopleplanner/service/impl/PeopleServiceImpl.java
akilarootcode: done.

shakila-kamalasena (approved):

erandi-rootcode (approved):

PR #1828, opened 2026-08-24
Fix/2605: Early signout issue

hasalarootcode:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR moves the access token from JS-written cookies to HttpOnly cookies written by a new `/api/auth/access-token` route, adds edge-side refresh in middleware, and caches the token in Zustand. The direction is a real improvement, but three issues are production-blocking: the `isPasswordChangedForTheFirstTime` cookie now expires with the 1-hour access token while the backend refresh response never returns that field, silently disabling the forced first-password-reset gate; the new POST route accepts a caller-supplied token with no origin/content-type check, enabling login-CSRF and cookie-attribute injection; and in community builds `~enterprise` resolves to `src/fallback` where `authenticationEndpoints` is `{}`, so the edge refresh URL is `${apiUrl}undefined` and every user is force-logged-out hourly. Several smaller correctness, resilience and typing issues are listed below.

Found **32** new issue(s): 🔴 13 important, 🟡 12 suggestion(s), 🟣 7 nit(s)

frontend/pages/api/auth/access-token.ts
```diff
@@ -0,0 +1,73 @@
...
+    setCookieHeaders.push(
+      `accessToken=${accessToken}; Path=/; Max-Age=${maxAge}; Secure; HttpOnly; SameSite=Lax`
+    );
+  }
+
+  if (isPasswordChangedForTheFirstTime !== undefined) {
+    setCookieHeaders.push(
+      `isPasswordChangedForTheFirstTime=${String(
+        isPasswordChangedForTheFirstTime
+      )}; Path=/; Max-Age=${maxAge}; Secure; HttpOnly; SameSite=Lax`
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The `isPasswordChangedForTheFirstTime` cookie is given the access token's Max-Age. Backend config (`application-ep-prd.yml`) sets `jwt.access-token.expiration-time: 3600000`, so this drops from the previous 31 days (`COOKIE_EXPIRY_DAYS`) to 1 hour. Worse, the middleware can never restore it: `AuthServiceImpl.refreshAccessToken()` returns `AccessTokenResponseDto`, which contains only `accessToken` — no `isPasswordChangedForTheFirstTime`. So `refreshSessionAtEdge` always yields `undefined` for that field and `applyRefreshedSession` skips re-setting the cookie. Net effect: one hour after sign-in the cookie is gone, `resolveRouteAccess` receives `undefined`, and the `=== "false"` branch at middleware.ts:253 never fires — a user who has not yet changed their initial password can browse the entire app and never be forced to `/reset-password`. The flag is also not in the JWT claims (`JwtServiceImpl.createAccessTokenClaims` / `EpJwtServiceImpl`), so it cannot be recovered from the token either. Fix: give this cookie its own long Max-Age (independent of the token), and/or have the backend include `isPasswordChangedForTheFirstTime` in the refresh-token response so `applyRefreshedSession` can re-issue it.

frontend/pages/api/auth/access-token.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  if (maxAge <= 0) {
+    return res.status(400).json({ message: "Access token has expired" });
+  }
+
+  const setCookieHeaders: string[] = [];
+
+  if (typeof accessToken === "string" && accessToken.length > 0) {
+    setCookieHeaders.push(
+      `accessToken=${accessToken}; Path=/; Max-Age=${maxAge}; Secure; HttpOnly; SameSite=Lax`
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The POST branch writes an arbitrary caller-supplied string into the session cookie with no origin or content-type validation. Next's pages-router body parser accepts `application/x-www-form-urlencoded`, which is a CORS 'simple request' (no preflight), so a cross-site auto-submitting form can POST `accessToken=<attacker JWT>` and the browser will honour the resulting `Set-Cookie` (SameSite=Lax restricts *sending*, not *setting*). That is classic login-CSRF: the victim ends up operating inside the attacker's account. Second problem: the value is string-interpolated into the header unescaped. `getTokenMaxAgeSeconds` only reads `split(".")[1]`, so `header.PAYLOAD.sig; Domain=evil.example` still parses a valid `exp`, letting an attacker append cookie attributes. Fix: reject any request whose `Content-Type` is not `application/json`; validate `Origin`/`Sec-Fetch-Site` is same-origin; and build the header with the `cookie` package's `serialize()` (or reject values containing `;`, `,`, whitespace) instead of template interpolation.

frontend/pages/api/auth/access-token.ts
```diff
@@ -0,0 +1,73 @@
...
+  req: NextApiRequest,
+  res: NextApiResponse<ResponseData>
+) {
+  if (req.method === "GET") {
+    const storedToken = req.cookies.accessToken ?? null;
+    const storedPasswordFlag = req.cookies.isPasswordChangedForTheFirstTime;
+
+    return res.status(200).json({
+      message: storedToken ? "Session cookie found" : "No session cookie",
+      accessToken: storedToken,
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The GET branch reads the HttpOnly `accessToken` cookie and returns it in a JSON body to any same-origin script, which negates the HttpOnly flag this PR just introduced — any XSS can exfiltrate the token with a single `fetch`. If the goal is a client-side bearer token, a token readable by JS is unavoidable, but then the round trip buys nothing; if the goal is HttpOnly hardening, the client should not need the raw token and requests should go out with `credentials: 'include'` instead. At minimum: add `Cache-Control: no-store` (a bearer token on a cacheable GET is a shared-cache hazard behind a CDN/proxy) and restrict the response to what the caller actually needs. Consider returning only `{ isAuthenticated, exp, roles }` and letting the middleware/backend hold the token.

frontend/src/community/auth/utils/authUtils.ts
```diff
@@ -156,53 +203,40 @@ export const clearCookies = async (): Promise<void> => {
...
+    return cachedAccessToken;
   }
 
-  return currentAccessToken;
-};
-
-export const isTokenExpired = (token: string): boolean => {
-  try {
-    const claims = extractClaimsFromToken(token);
+  if (!cachedAccessToken) {
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`if (!cachedAccessToken)` means the persisted cookie is only consulted when the in-memory store is empty — never when the cached token is merely expired. On a client-side navigation the middleware may have just refreshed the cookie server-side, but the store still holds the old expired token, so this falls straight through to `getNewAccessToken()` and burns an extra refresh round trip on every expiry, every tab. Drop the guard so the stored token is re-read whenever the cached one is missing *or* expired: `if (!cachedAccessToken || isTokenExpired(cachedAccessToken)) { const storedToken = await retrieveStoredAccessToken(); ... }`.

frontend/src/community/auth/utils/edgeSessionUtils.ts
```diff
@@ -0,0 +1,139 @@
...
+  };
+
+  if (process.env.NEXT_PUBLIC_MODE === appModes.ENTERPRISE) {
+    const tenantId = getTenantIdFromRequest(request);
+    if (tenantId) headers[TENANT_HEADER_NAME] = tenantId;
+  }
+
+  try {
+    const response = await fetch(
+      `${apiUrl}${authenticationEndpoints.REFRESH_TOKEN}`,
```
ThinuwanW, line 93: 🤖 **Claude** · 🔴 **Important**

`authenticationEndpoints` is imported from `~enterprise/common/api/utils/ApiEndpoints`, which in community builds (and always under Jest — see `jest.config` `^~enterprise(.*)$` → `src/fallback$1`) resolves to `src/fallback/common/api/utils/ApiEndpoints.ts`, whose entire content is `export const authenticationEndpoints = {}`. `REFRESH_TOKEN` is therefore `undefined` and the edge fetch targets `${apiUrl}undefined`. Because `next.config.mjs` sets `typescript.ignoreBuildErrors: true`, this ships silently. The consequence is a hard regression in community mode: every request after the 1-hour access token expires hits `refreshSessionAtEdge` → non-ok → `didAttemptRefresh && !refreshedSession` → `clearSessionCookies` → redirect to `/signin`. Community users get force-logged-out hourly, where previously the 31-day cookie kept them routed. The community backend does expose this route (`AuthController` `/v1/auth/session/refresh-token`), so add `REFRESH_TOKEN` to `~community/common/api/utils/ApiEndpoints.ts` and import it from there in this community-side file.

frontend/middleware.ts
```diff
@@ -376,6 +376,39 @@ export function middleware(request: NextRequest) {
...
+      ? String(refreshedSession.isPasswordChangedForTheFirstTime)
+      : request.cookies.get("isPasswordChangedForTheFirstTime")?.value;
+
+  const response = resolveRouteAccess(
+    request,
+    token,
+    isPasswordChangedForTheFirstTime
+  );
+
+  if (didAttemptRefresh && !refreshedSession) {
```

`refreshSessionAtEdge` returns `null` for every failure mode — network error, backend 5xx, malformed JSON — and this branch treats all of them as 'session is dead' and wipes the cookies. A single flaky backend response or edge network blip logs the user out mid-session. Distinguish definitive rejection from transient failure: have `refreshSessionAtEdge` return a discriminated result (e.g. `{ status: 'unauthorized' } | { status: 'error' } | RefreshedSession`) and only call `clearSessionCookies` on `unauthorized` (401/403 from the backend). On transient errors, leave the cookies alone and let the next request retry.

frontend/src/community/auth/providers/AuthProvider.tsx
```diff
@@ -97,14 +97,9 @@ export const AuthProvider: React.FC<AuthProviderProps> = ({ children }) => {
...
           if (params.redirect) {
-            // Verify that cookies are set before redirecting
-            const accessToken = getCookieValue("accessToken");
-            const isPasswordChangedForTheFirstTime = getCookieValue(
-              "isPasswordChangedForTheFirstTime"
-            );
-
-            // Ensure critical cookies are available before redirect
-            if (accessToken && isPasswordChangedForTheFirstTime !== null) {
+            const userData = await checkUserAuthentication();
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

This replaces a check that the actual cookies existed with `checkUserAuthentication()`, which resolves from the in-memory Zustand store that `setAccessToken` just populated — so it now succeeds even when the cookie was never persisted (`setAccessToken` ignores non-2xx responses from `/api/auth/access-token`, e.g. the 400 'Access token has expired' path, or a >4KB JWT the browser rejects). The user is then redirected to `/dashboard`, middleware finds no `accessToken` cookie, and bounces them back to `/signin`. Since the guard's whole purpose is verifying cookie persistence, verify it directly: have `setAccessToken` check `response.ok` and surface failure, or `GET /api/auth/access-token` here and assert the server actually sees the cookie. Separately, `checkAuth()` on line 97 already fetched the same user — have `checkAuth` return the `userData` it computed and reuse it instead of calling `checkUserAuthentication()` a second time.

frontend/src/community/auth/utils/authUtils.ts
```diff
@@ -125,23 +158,37 @@ export const getNewAccessToken = async (): Promise<string | null> => {
...
-    const expiryDate = new Date(
-      Date.now() + COOKIE_EXPIRY_DAYS * unitConversion.MILLISECONDS_PER_DAY
-    );
+export const setAccessToken = async (token: string): Promise<void> => {
+  useCommonStore.getState().setAccessToken(token);
 
-    document.cookie = `accessToken=${token}; path=/; expires=${expiryDate.toUTCString()}; Secure; SameSite=Lax`;
+  if (typeof window !== "undefined") {
+    try {
+      await fetch(internalApiEndpoints.ACCESS_TOKEN, {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Neither `setAccessToken` (here) nor `setIsPasswordChangedForTheFirstTime` (line 183) checks `response.ok`. `fetch` only rejects on network failure, so every 400 the new route can return ('Nothing to set', 'Access token has expired') and every 405 is swallowed silently — the `catch` never runs and nothing is logged. Check `if (!response.ok)` and log/propagate, otherwise a cookie that was never written looks like a success to every caller (including `handleAuthResponse` and the enterprise `handleAuthResponse` in `src/enterprise/auth/utils/authUtils.ts`).

frontend/middleware.ts
```diff
@@ -376,6 +376,39 @@ export function middleware(request: NextRequest) {
...
+export async function middleware(request: NextRequest) {
+  if (isUnguardedPath(request.nextUrl.pathname)) {
+    return NextResponse.next();
+  }
+
+  let token = request.cookies.get("accessToken")?.value;
+  let refreshedSession: RefreshedSession | null = null;
+  let didAttemptRefresh = false;
+
+  if (!token || isTokenExpired(token)) {
```

This fires a blocking outbound request to the backend on every matched request whenever the token is missing or expired — including for fully logged-out visitors, as long as *any* cookie is present (`refreshSessionAtEdge` only bails when the whole cookie header is absent, and in enterprise mode the `tenant` cookie alone satisfies that). Since the access-token cookie's Max-Age now equals the token TTL, this also runs on Next `<Link>` prefetches. Two mitigations: (a) in `refreshSessionAtEdge`, bail early unless a refresh-token cookie is actually present in the header rather than checking for any cookie at all; (b) skip the refresh for prefetch requests (`request.headers.get('next-router-prefetch')` / `purpose: prefetch`).

frontend/src/community/auth/utils/edgeSessionUtils.ts
```diff
@@ -0,0 +1,139 @@
...
+    cookie: cookieHeader
+  };
+
+  if (process.env.NEXT_PUBLIC_MODE === appModes.ENTERPRISE) {
+    const tenantId = getTenantIdFromRequest(request);
+    if (tenantId) headers[TENANT_HEADER_NAME] = tenantId;
+  }
+
+  try {
+    const response = await fetch(
```
ThinuwanW, line 92: 🤖 **Claude** · 🟡 **Suggestion**

This `fetch` has no timeout. Middleware runs in the request path, so an unresponsive backend stalls every page navigation until the platform's own timeout. Add `signal: AbortSignal.timeout(3000)` (or similar) and treat the abort as a transient failure per the `clearSessionCookies` finding.

frontend/src/community/auth/utils/edgeSessionUtils.ts
```diff
@@ -0,0 +1,139 @@
...
+  request: NextRequest
+): Promise<RefreshedSession | null> => {
+  const apiUrl = process.env.NEXT_PUBLIC_API_URL;
+  const cookieHeader = request.headers.get("cookie");
+
+  if (!apiUrl || !cookieHeader) return null;
+
+  const headers: Record<string, string> = {
+    "Content-Type": "application/json",
+    cookie: cookieHeader
```
ThinuwanW, line 84: 🤖 **Claude** · 🟡 **Suggestion**

The entire browser cookie header for the frontend origin is forwarded verbatim to `NEXT_PUBLIC_API_URL`. When the API is on a separate host, this hands the backend cookies the browser would never have sent it (analytics, feature flags, the frontend `accessToken` itself). Parse the header and forward only the refresh-token cookie the backend actually reads (`CookieUtil.REFRESH_TOKEN_COOKIE_NAME`, or `<tenantId>` + `REFRESH_TOKEN_COOKIE_SUFFIX` per `EpCookieUtil.getRefreshTokenCookieName`).

frontend/src/community/auth/utils/edgeSessionUtils.ts
```diff
@@ -0,0 +1,139 @@
+import type { NextRequest, NextResponse } from "next/server";
+
+import { appModes } from "~community/common/constants/configs";
+import { LOCALHOST } from "~community/common/constants/stringConstants";
+import { authenticationEndpoints } from "~enterprise/common/api/utils/ApiEndpoints";
+import {
```
ThinuwanW, line 19: 🤖 **Claude** · 🟡 **Suggestion**

This is a `~community` file importing `TENANT_COOKIE_NAME`, `TENANT_HEADER_NAME`, `TENANT_QUERY_PARAM` and `TENANT_SELECTION_SUBDOMAINS` from `~enterprise/common/constants/stringConstants`. In community builds and under Jest that resolves to `src/fallback/common/constants/stringConstants.ts`, which exports only `APP` and `LOCALHOST` — all four are `undefined` there, and `TENANT_SELECTION_SUBDOMAINS.includes(...)` would throw a TypeError if `getTenantIdFromRequest` were ever reached (today it is saved only by the `NEXT_PUBLIC_MODE === appModes.ENTERPRISE` guard, and any test of this file would break). Add stub exports for these four to `src/fallback/common/constants/stringConstants.ts`, consistent with how `APP`/`LOCALHOST` are stubbed.

frontend/src/community/auth/utils/edgeSessionUtils.ts
```diff
@@ -0,0 +1,139 @@
...
+  subdomain !== LOCALHOST &&
+  !TENANT_SELECTION_SUBDOMAINS.includes(subdomain);
+
+export interface RefreshedSession {
+  accessToken: string;
+  isPasswordChangedForTheFirstTime?: boolean;
+  backendSetCookies: string[];
+}
+
+const getTenantIdFromRequest = (request: NextRequest): string => {
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`getTenantIdFromRequest` re-implements `getTenantId()` from `~enterprise/common/utils/tenantUtil.ts` with slightly different rules — the edge version adds the `hostname.split('.').length > 2`, IPv4 and `localhost` checks, and drops `normalizeTenantId` on the subdomain branch. Two tenant-resolution implementations that can disagree will eventually send `X-Tenant-ID` to the wrong schema. Extract the shared predicate into one function used by both (the client version can pass `window.location.hostname`, the edge version `request.nextUrl.hostname`). Also, `process.env.NEXT_PUBLIC_API_URL` on line 46 should go through `getApiUrl()` from `~community/common/utils/getConstants`, which is what the rest of the codebase uses.

frontend/middleware.ts
```diff
@@ -200,22 +206,20 @@ const allowedRoutes: Record<
...
 
-export function middleware(request: NextRequest) {
-  // Get accessToken from cookies
-  const token = request.cookies.get("accessToken")?.value;
+const isUnguardedPath = (currentPath: string): boolean =>
+  currentPath === ROUTES.SIGN.DOCUMENT_ACCESS ||
+  currentPath.startsWith(ROUTES.SIGN.SIGN) ||
+  currentPath.startsWith(ROUTES.SIGN.INFO);
 
+function resolveRouteAccess(
```

Missing explicit return types on new/refactored code, which the strict-mode guideline requires. `resolveRouteAccess` has an inferred return type (annotate `NextResponse`); `decodeJWTToken` in `tokenUtils.ts:3` returns implicit `any`; `const { accessToken, isPasswordChangedForTheFirstTime } = req.body || {}` in `access-token.ts:32` destructures `any`, which is what lets a non-string `accessToken` reach `String(...)` unchecked. Give the API route an explicit request-body interface and validate the fields against it.

frontend/src/community/auth/utils/authUtils.ts
```diff
@@ -125,23 +158,37 @@ export const getNewAccessToken = async (): Promise<string | null> => {
   return refreshPromise;
 };
 
-export const setAccessToken = (token: string) => {
-  if (typeof window !== "undefined") {
-    const expiryDate = new Date(
-      Date.now() + COOKIE_EXPIRY_DAYS * unitConversion.MILLISECONDS_PER_DAY
-    );
+export const setAccessToken = async (token: string): Promise<void> => {
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`COOKIE_EXPIRY_DAYS` in `src/community/auth/constants/authConstants.ts` was the only consumer of the old expiry logic and is now unreferenced anywhere in the repo. Delete it (and the now-unused `unitConversion` import path it relied on) so the constant does not get resurrected against the new cookie-lifetime scheme.

frontend/src/community/common/stores/commonStore.ts
```diff
@@ -20,7 +21,8 @@ export const useCommonStore = create<
       ...commonModalSlice(set),
       ...notificationsSlice(set),
       ...orgDetailsSlice(set),
-      ...breadcrumbSlice(set)
+      ...breadcrumbSlice(set),
+      ...authTokenSlice(set)
```
ThinuwanW, line 25: 🤖 **Claude** · 🟣 **Nit**

`authTokenSlice` puts the raw access token into `commonStore`, which is wrapped in `devtools({ name: 'commonStore' })`. With the Redux DevTools extension installed the bearer token and every `setAccessToken` action payload become visible in the extension's state/action log and time-travel history. Consider a dedicated store without `devtools`, or pass a `serialize`/action sanitizer to redact `accessToken`.

frontend/src/community/common/stores/slices/authTokenSlice.ts
```diff
@@ -0,0 +1,23 @@
+import { SetType } from "~community/common/types/storeTypes";
```
ThinuwanW, line 1: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

No tests accompany the four new modules (`tokenUtils.ts`, `edgeSessionUtils.ts`, `pages/api/auth/access-token.ts`, this slice), and there is no existing test covering `middleware.ts`. Given this PR reworks the session/auth path, `tokenUtils` in particular is cheap and high-value to cover: `getTokenMaxAgeSeconds` with a malformed token, an expired token and a valid one, plus `isTokenExpired` on a token with no `exp` claim. Note that any test of `edgeSessionUtils` will hit the `~enterprise` → `src/fallback` mapping described above, so the fallback stubs need to land first.

frontend/src/community/auth/utils/authUtils.ts
```diff
@@ -156,53 +203,40 @@ export const clearCookies = async (): Promise<void> => {
...
 
-export const extractClaimsFromToken = (token: string): Record<string, any> => {
-  try {
-    const claims = decodeJWTToken(token);
-    return claims || {};
-  } catch (error) {
-    console.error("Failed to parse token:", error);
-    return {};
-  }
+  const newToken = await getNewAccessToken();
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Behavioral drift that breaks anonymous/public pages. The old `getAccessToken()` returned `null` immediately when no `accessToken` cookie existed — zero network calls. The new version falls through to `getNewAccessToken()` in every case where the store is empty and no persisted token is found. Trace what that does for a visitor with no session: `AuthProvider` (mounted in `_app.tsx` around *every* page) calls `checkAuth()` -> `checkUserAuthentication()` -> `getAccessToken()` -> `getNewAccessToken()` -> `authAxios.post(/v1/auth/session/refresh-token)`. `AuthServiceImpl.refreshAccessTokenFromCookie` throws `COMMON_ERROR_INVALID_REFRESH_TOKEN` when the refresh cookie is absent, and that key is in `SIGN_OUT_ERROR_KEYS` in `authInterceptor.ts:13`, so the response interceptor calls `await signOut()` — which clears cookies and sets `window.location.href` to `/signin`. Net effect: anonymous visitors on `/signup`, `/auth/signin`, `/verify/*`, `/forgot-password`, `/maintenance` and especially the external-signer route `/sign/document/access` (which `isUnguardedPath` in middleware.ts deliberately exempts from auth) get bounced to the login page, behind a `FullScreenLoader`, after 4 wasted round trips. Fix: only escalate to `getNewAccessToken()` when there is evidence of a session — e.g. return `null` when both the store and `retrieveStoredAccessToken()` came back empty, and let the axios interceptor's 401 path drive the refresh instead.

frontend/src/community/auth/utils/authUtils.ts
```diff
@@ -338,6 +372,9 @@ export const signOut = async (redirect: boolean = true): Promise<void> => {
 
   if (typeof window !== "undefined") {
     const currentPath = window.location.pathname;
+
+    if (currentPath === ROUTES.AUTH.SIGNIN) return;
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The new `if (currentPath === ROUTES.AUTH.SIGNIN) return;` guard is a band-aid over the redirect loop described above, and it only covers one path. `ROUTES.AUTH.SIGNIN` is `/signin`, but in enterprise mode the login page users actually land on is `/auth/signin` (next.config.mjs rewrites it to `pages/enterprise/auth/signin.tsx`), and signup is `/signup`. Those pages still get `window.location.href = '/signin?callback=/auth/signin'`, so an enterprise user cannot stay on the enterprise login page. Fix: compare against a shared list of public routes (SIGNIN, ENTERPRISE_SIGNIN, `/auth/signin`, SIGNUP, FORGOT_PASSWORD, VERIFY_*, SIGN.DOCUMENT_ACCESS, MAINTENANCE) rather than a single constant — ideally the same list used to decide whether a refresh should be attempted at all.

frontend/src/community/auth/utils/edgeSessionUtils.ts
```diff
@@ -0,0 +1,139 @@
...
+  return response;
+};
+
+export const applyRefreshedSession = (
+  response: NextResponse,
+  refreshedSession: RefreshedSession | null
+): NextResponse => {
+  if (!refreshedSession) return response;
+
+  const maxAge = getTokenMaxAgeSeconds(refreshedSession.accessToken);
```
ThinuwanW, line 153: 🤖 **Claude** · 🔴 **Important**

`applyRefreshedSession` can delete the cookie it is trying to set. `getTokenMaxAgeSeconds` returns `0` whenever the token has no `exp` claim or `exp` is already in the past (clock skew between the edge runtime and the backend is enough), and `response.cookies.set({ maxAge: 0 })` serializes to `Max-Age=0`, which instructs the browser to drop the cookie. The user then arrives at the next page with no `accessToken` cookie, the middleware refreshes again, and every single navigation pays a blocking backend round trip while never persisting a session. Note `pages/api/auth/access-token.ts:47` already guards this with `if (maxAge <= 0) return 400` — the edge path is missing the equivalent. Fix: `if (maxAge <= 0) return response;` (treat it as a failed refresh, or omit `maxAge` to write a session cookie) before writing either cookie.

frontend/src/community/auth/providers/AuthProvider.tsx
```diff
@@ -85,7 +85,7 @@ export const AuthProvider: React.FC<AuthProviderProps> = ({ children }) => {
     async (params: EnterpriseSignInParams): Promise<AuthResponseType> => {
       try {
         // Clear all existing cookies before signing in
-        await fetch("/api/clear-cookies", {
+        await fetch(internalApiEndpoints.CLEAR_COOKIES, {
```
ThinuwanW, line 111: 🤖 **Claude** · 🔴 **Important**

`signIn` clears the browser cookies via `/api/clear-cookies` but never clears the in-memory Zustand token, which this PR makes the *primary* source of truth for `getAccessToken()` (authUtils.ts:223). If `handleSignIn` then fails or returns a response without an `accessToken` (`handleAuthResponse` returns FAILURE without touching the store), the store still holds the previous user's token. `getAccessToken()`/`checkUserAuthentication()` will happily return it, so the SPA renders the previous user's session while the cookies are gone — and the next call to `setAccessToken` is the only thing that would have overwritten it. Fix: call `useCommonStore.getState().clearAccessToken()` alongside the `CLEAR_COOKIES` POST (or just reuse `clearCookies()` from authUtils, which already does both).

frontend/pages/api/auth/access-token.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  const setCookieHeaders: string[] = [];
+
+  if (typeof accessToken === "string" && accessToken.length > 0) {
+    setCookieHeaders.push(
+      `accessToken=${accessToken}; Path=/; Max-Age=${maxAge}; Secure; HttpOnly; SameSite=Lax`
+    );
+  }
+
+  if (isPasswordChangedForTheFirstTime !== undefined) {
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`isPasswordChangedForTheFirstTime` is accepted from the request body with no type validation and interpolated raw via `String(...)` straight into a `Set-Cookie` header, and the whole handler has no try/catch. Three concrete failures: (1) a value containing `;` lets the caller append or override cookie attributes on that Set-Cookie line (`Max-Age`, `Domain`, `SameSite`); (2) a value containing CR/LF makes `res.setHeader` throw `ERR_INVALID_CHAR`, which — with no try/catch — surfaces as an opaque 500 rather than a 400; (3) any non-boolean such as `0` or `"maybe"` produces a cookie value that is neither `"true"` nor `"false"`, and `middleware.ts:253-263` compares against exactly those two strings, so the forced-password-reset gate is silently skipped for that user. Fix: reject unless `typeof isPasswordChangedForTheFirstTime === "boolean"`; validate `accessToken` against a JWT shape (`/^[\w-]+\.[\w-]+\.[\w-]+$/`) with a length cap; and wrap the handler body in try/catch returning 400/500 JSON.

frontend/src/community/auth/utils/edgeSessionUtils.ts
```diff
@@ -0,0 +1,139 @@
...
+    name: "accessToken",
+    value: refreshedSession.accessToken,
+    path: "/",
+    maxAge,
+    secure: true,
+    httpOnly: true,
+    sameSite: "lax"
+  });
+
+  if (refreshedSession.isPasswordChangedForTheFirstTime !== undefined) {
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Two fields of `RefreshedSession` are dead against the real backend contract, and one of them hides a security-gate regression. `REFRESH_TOKEN` resolves to `/v1/auth/session/refresh-token`, handled by `AuthController.refreshAccessTokenFromCookie`, which returns `AccessTokenResponseDto` — a DTO with a single `accessToken` field and no `isPasswordChangedForTheFirstTime` — and whose signature takes only `HttpServletRequest`, so it cannot emit any `Set-Cookie`. Therefore `refreshedSession.isPasswordChangedForTheFirstTime` is always `undefined` (this branch never runs, and the ternary at middleware.ts:394-397 always takes the cookie fallback) and `backendSetCookies` is always `[]` (line 134 is a no-op). The consequence of the first: the flag cookie is never re-issued on refresh, so once it lapses the middleware sees `undefined` and *both* reset-password gates stop applying — a user who never changed their first-time password gains full app access after the first token refresh. Fix: either delete both fields, or have the backend return the flag on refresh and re-issue the cookie here. Also note that if `backendSetCookies` ever becomes non-empty, forwarding an API-host `Set-Cookie` verbatim onto the frontend origin is wrong — its `Domain`/`Path` will not match and the browser will drop it silently.

frontend/src/community/auth/utils/tokenUtils.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    return Date.now() > exp * unitConversion.MILLISECONDS_PER_SECOND;
+  } catch (error) {
+    console.error("Failed to parse token:", error);
+    return true;
+  }
+};
+
+export const getTokenMaxAgeSeconds = (token: string): number => {
+  const exp = extractClaimsFromToken(token)?.exp as number | undefined;
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`exp` is coerced with an `as number | undefined` assertion instead of a runtime check, so a non-numeric `exp` propagates `NaN` through the whole pipeline: `isTokenExpired` computes `Date.now() > NaN` => `false`, so a garbage token is treated as *valid* and the middleware skips refresh; `getTokenMaxAgeSeconds` returns `Math.max(NaN, 0)` => `NaN`, which passes the `maxAge <= 0` guard in access-token.ts:47 and emits a literal `Max-Age=NaN`, and in `applyRefreshedSession` gets handed to `NextResponse.cookies.set`, whose `cookie` serializer throws `TypeError: option maxAge is invalid` — a 500 from middleware on every request. This is reachable because the POST endpoint accepts any string as `accessToken`. Fix: `const exp = extractClaimsFromToken(token).exp; if (typeof exp !== "number" || !Number.isFinite(exp)) return 0;` in both functions. Separately, `decodeJWTToken` has no explicit return type — `JSON.parse` widens it to `any`, so the newly added `return null` is invisible to TypeScript and no caller is forced to handle it. Annotate it `Record<string, unknown> | null` (strict mode requires explicit return types on exported functions).

frontend/pages/api/auth/access-token.ts
```diff
@@ -0,0 +1,73 @@
...
+
+export default function handler(
+  req: NextApiRequest,
+  res: NextApiResponse<ResponseData>
+) {
+  if (req.method === "GET") {
+    const storedToken = req.cookies.accessToken ?? null;
+    const storedPasswordFlag = req.cookies.isPasswordChangedForTheFirstTime;
+
+    return res.status(200).json({
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The GET response body contains a bearer token but sets no cache directives. Add `res.setHeader("Cache-Control", "no-store")` and `res.setHeader("Vary", "Cookie")` so no intermediary proxy, CDN, or the browser's own disk cache can retain one user's access token and hand it to another. Separately, `isPasswordChangedForTheFirstTime` in this response is dead weight: `retrieveStoredAccessToken` (authUtils.ts:112) reads only `data.accessToken`, and no other caller exists — drop the field rather than widening what this endpoint discloses.

frontend/src/community/auth/utils/authUtils.ts
```diff
@@ -125,23 +158,37 @@ export const getNewAccessToken = async (): Promise<string | null> => {
   return refreshPromise;
 };
 
-export const setAccessToken = (token: string) => {
-  if (typeof window !== "undefined") {
-    const expiryDate = new Date(
-      Date.now() + COOKIE_EXPIRY_DAYS * unitConversion.MILLISECONDS_PER_DAY
-    );
+export const setAccessToken = async (token: string): Promise<void> => {
+  useCommonStore.getState().setAccessToken(token);
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The Zustand write sits *outside* the `typeof window !== "undefined"` guard that wraps the fetch two lines below (and that `setIsPasswordChangedForTheFirstTime` applies consistently). `useCommonStore` is a module-level singleton, so on the server it is shared by every concurrent request handled by that Node process — any server-side invocation of `setAccessToken` (SSR path, `getInitialProps`, an API route that imports this module) writes one user's access token into state visible to other users' requests. `getNewAccessToken` has no window guard either, so the reachability is not obvious from here. Move the `useCommonStore.getState().setAccessToken(token)` call inside the existing `typeof window !== "undefined"` block.

frontend/src/community/auth/utils/edgeSessionUtils.ts
```diff
@@ -0,0 +1,139 @@
...
+        data?.results?.[0]?.isPasswordChangedForTheFirstTime,
+      backendSetCookies: responseHeaders.getSetCookie?.() ?? []
+    };
+  } catch {
+    return null;
+  }
+};
+
+export const clearSessionCookies = (response: NextResponse): NextResponse => {
+  ["accessToken", "isPasswordChangedForTheFirstTime"].forEach((name) =>
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The cookie names `"accessToken"` and `"isPasswordChangedForTheFirstTime"` are now hardcoded across four files: here and at lines 113/124, `middleware.ts:384` and `:397`, `pages/api/auth/access-token.ts:16/17/55/61`, and `pages/api/clear-cookies.ts`. The attribute sets have already drifted — the edge and the API route write `HttpOnly; Secure; SameSite=Lax`, while `clear-cookies` sprays seven attribute permutations per cookie. Extract the names (and ideally a single `buildSessionCookie(name, value, maxAge)` helper) into `~community/auth/constants/authConstants.ts` — which is the same file left holding the now-unreferenced `COOKIE_EXPIRY_DAYS` — so a rename or attribute change cannot half-land.

frontend/src/community/auth/providers/AuthProvider.tsx
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`checkUserAuthentication()` is invoked immediately after `await checkAuth()`, which already called exactly that function and stored the result via `setUser`. That is a second full token decode and, if the store happens to be empty, a second `GET /api/auth/access-token` plus a second refresh POST on the critical sign-in path. Have `checkAuth` return the `userData` it computed (`const userData = await checkAuth();`) and branch on that, so the redirect decision reuses the value the provider just committed to state instead of re-deriving it.

frontend/src/community/auth/utils/authUtils.ts
```diff
@@ -33,6 +32,13 @@ import {
   CommunitySignUpParams
 } from "../types/auth";
 import authAxios from "./authInterceptor";
+import {
+  decodeJWTToken,
+  extractClaimsFromToken,
+  isTokenExpired
+} from "./tokenUtils";
+
+export { decodeJWTToken, extractClaimsFromToken, isTokenExpired };
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Re-exporting `decodeJWTToken`, `extractClaimsFromToken` and `isTokenExpired` from `authUtils` keeps two live import paths for the same three functions, which defeats the point of splitting `tokenUtils.ts` out. `pages/enterprise/sign/sign.tsx:6` still imports `decodeJWTToken` from `~community/auth/utils/authUtils`, so that page's bundle still pulls in axios, `authInterceptor`, `useCommonStore` and `middleware.ts` just to base64-decode a token. Update the remaining callers to import from `~community/auth/utils/tokenUtils` and delete the re-export line.

frontend/middleware.ts
```diff
@@ -200,22 +206,20 @@ const allowedRoutes: Record<
   ...commonRoutes
 };
 
-export function middleware(request: NextRequest) {
-  // Get accessToken from cookies
-  const token = request.cookies.get("accessToken")?.value;
+const isUnguardedPath = (currentPath: string): boolean =>
+  currentPath === ROUTES.SIGN.DOCUMENT_ACCESS ||
+  currentPath.startsWith(ROUTES.SIGN.SIGN) ||
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`isUnguardedPath`'s `startsWith(ROUTES.SIGN.SIGN)` and `startsWith(ROUTES.SIGN.INFO)` branches can never fire: `config.matcher` (lines 438-446) registers `"/sign"` as an exact path plus specific subtrees (`/sign/contacts/:path*`, `/sign/inbox/:path*`, …) and never matches `/sign/sign` or `/sign/info`, so middleware does not run for those paths at all. The condition was dead before this PR too, but promoting it into a named `isUnguardedPath` helper makes it read as a deliberate, load-bearing exemption for the external-signer flow when it is not one. Either add the missing matcher entries (and then genuinely exempt them) or drop the two `startsWith` branches so the intent is honest.

frontend/src/community/common/stores/slices/authTokenSlice.ts
```diff
@@ -0,0 +1,23 @@
...
+export const authTokenSlice = (
+  set: SetType<AuthTokenSliceType>
+): AuthTokenSliceType => ({
+  accessToken: null,
+  setAccessToken: (token: string | null) =>
+    set((state) => ({
+      ...state,
+      accessToken: token
+    })),
+  clearAccessToken: () =>
```
ThinuwanW, line 18: 🤖 **Claude** · 🟣 **Nit**

`clearAccessToken` is behaviorally identical to `setAccessToken(null)`, and because this slice is spread into `commonStore` both names become part of the global store surface — two ways to express one transition invites the store and the cookie to be cleared through different paths. Either drop `clearAccessToken` and use `setAccessToken(null)`, or narrow the setter to `(token: string) => void` so `clearAccessToken` is the only way to null it (that also makes the `string | null` parameter type in `StoreTypes.ts:61` accurate to how `authUtils` actually calls it — always with a string).

frontend/src/community/auth/utils/authUtils.ts
```diff
@@ -90,6 +89,40 @@ export interface User {
 let isRefreshing = false;
 let refreshPromise: Promise<string | null> | null = null;
 
+let retrievePromise: Promise<string | null> | null = null;
```
ThinuwanW, line 124: 🤖 **Claude** · 🟣 **Nit**

`retrievePromise` is nulled in `finally`, so only *concurrently pending* calls are coalesced — a resolved negative result is never remembered. Since `axiosInterceptor.ts:23` calls `getAccessToken()` on every single request, a user without a persisted cookie re-issues `GET /api/auth/access-token` on each one. Consider caching the settled outcome (or a `hasCheckedStoredToken` flag) so the endpoint is consulted at most once per page load.

frontend/pages/api/auth/access-token.ts
hasalarootcode: Fixed. The flag cookie gets its own 31 day Max-Age now (`PASSWORD_CHANGED_COOKIE_MAX_AGE_SECONDS`), independent of the token's.

frontend/pages/api/auth/access-token.ts
hasalarootcode: Fixed. POST rejects anything that isn't `application/json` and anything cross-origin (`Sec-Fetch-Site`, then `Origin` against `Host`), and the token is validated against a JWT pattern with a length cap before it reaches the header.

frontend/pages/api/auth/access-token.ts
hasalarootcode: Added `Cache-Control: no-store` and `Vary: Cookie`, and dropped `isPasswordChangedForTheFirstTime` from the response. The token itself still comes back, since the axios interceptor needs a bearer for the API host - the cookie is there to survive reloads, not to keep the token away from JS.

frontend/src/community/auth/utils/authUtils.ts
hasalarootcode: Fixed. The stored cookie is read whenever the cached token is missing or expired.

frontend/src/community/auth/utils/edgeSessionUtils.ts
hasalarootcode: Fixed. Added `REFRESH_TOKEN` to `~community/common/api/utils/ApiEndpoints.ts` and imported it from there.

frontend/middleware.ts
hasalarootcode: Fixed. `refreshSessionAtEdge` returns `{ status: "success" | "unauthorized" | "error" }`, and only `unauthorized` (401/403, or no refresh cookie at all) clears the cookies.

frontend/src/community/auth/providers/AuthProvider.tsx
hasalarootcode: Fixed. `setAccessToken` checks `response.ok` and returns a boolean, `handleAuthResponse` fails the sign-in when the cookie wasn't written, and `checkAuth` returns the `userData` it computed so the redirect reuses it instead of calling `checkUserAuthentication` again.

frontend/src/community/auth/utils/authUtils.ts
hasalarootcode: Fixed. Both check `response.ok`, log the status and return a boolean.

frontend/middleware.ts
hasalarootcode: Fixed. `refreshSessionAtEdge` bails unless a `refreshToken` / `*_refreshToken` cookie is present, and the middleware skips the refresh on prefetch requests.

frontend/src/community/auth/utils/edgeSessionUtils.ts
hasalarootcode: Fixed. `AbortSignal.timeout(EDGE_REFRESH_TIMEOUT_MS)`, and an abort lands in the `error` status so the cookies are left alone.

frontend/src/community/auth/utils/edgeSessionUtils.ts
hasalarootcode: Fixed. Only `refreshToken` and `*_refreshToken` cookies are forwarded now.

frontend/src/community/auth/utils/edgeSessionUtils.ts
hasalarootcode: Fixed. Added the four stubs to `src/fallback/common/constants/stringConstants.ts`.

frontend/src/community/auth/utils/edgeSessionUtils.ts
hasalarootcode: Moved to `getApiUrl()`. The shared predicate would have to live in `~enterprise/common/utils/tenantUtil.ts`, which is a separate submodule repo - want that as a follow-up there?

frontend/middleware.ts
hasalarootcode: Fixed. Return types on `middleware`, `resolveRouteAccess` and `decodeJWTToken`, and the request body goes through `AccessTokenRequestBody` with a per-field check.

frontend/src/community/auth/utils/authUtils.ts
hasalarootcode: It's referenced again - `PASSWORD_CHANGED_COOKIE_MAX_AGE_SECONDS` is derived from it for the flag cookie's Max-Age.

frontend/src/community/common/stores/commonStore.ts
hasalarootcode: Fixed. Added a `stateSanitizer` to the devtools middleware that redacts `accessToken`.

frontend/src/community/common/stores/slices/authTokenSlice.ts
hasalarootcode: The fallback stubs are in now, so `tokenUtils` is testable. Want the tests in this PR or a follow-up?

frontend/src/community/auth/utils/authUtils.ts
hasalarootcode: Fixed. `getAccessToken` returns `null` when both the store and the stored cookie are empty, so anonymous visitors never reach the refresh call.

frontend/src/community/auth/utils/authUtils.ts
hasalarootcode: Fixed. Checks against `PUBLIC_ROUTES` in `authConstants.ts`.

frontend/src/community/auth/utils/edgeSessionUtils.ts
hasalarootcode: Fixed. `applyRefreshedSession` returns early when `maxAge <= 0`.

frontend/src/community/auth/providers/AuthProvider.tsx
hasalarootcode: Fixed. `signIn` calls `clearAccessToken()` before the `CLEAR_COOKIES` POST.

frontend/pages/api/auth/access-token.ts
hasalarootcode: Fixed. Strict `boolean` check on the flag, JWT pattern plus length cap on the token, and the handler body is wrapped in try/catch.

frontend/src/community/auth/utils/edgeSessionUtils.ts
hasalarootcode: Removed both fields. The flag cookie has its own 31 day Max-Age now, so it survives a refresh without the backend returning it.

frontend/src/community/auth/utils/tokenUtils.ts
hasalarootcode: Fixed. Both go through `getExpiryInSeconds`, which returns `null` unless `exp` is a finite number. `decodeJWTToken` is annotated `Record<string, unknown> | null`.

frontend/pages/api/auth/access-token.ts
hasalarootcode: Fixed. `no-store` and `Vary: Cookie` are set, and the flag is out of the response.

frontend/src/community/auth/utils/authUtils.ts
hasalarootcode: Fixed. The function returns early on the server, so the store write is behind the guard.

frontend/src/community/auth/utils/edgeSessionUtils.ts
hasalarootcode: Fixed. The names, the shared attributes and `buildSessionCookieHeader` are in `authConstants.ts`, used by the edge utils, the middleware and the API route.

frontend/src/community/auth/providers/AuthProvider.tsx
hasalarootcode: Fixed. `checkAuth` returns the `userData` and `signIn` branches on that.

frontend/src/community/auth/utils/authUtils.ts
hasalarootcode: Both remaining callers are in submodules - `pages/enterprise/sign/sign.tsx` and `src/enterprise/auth/utils/authUtils.ts` - so dropping the re-export needs matching PRs on those repos. Raise them as a follow-up?

frontend/middleware.ts
hasalarootcode: Dropped the two branches.

frontend/src/community/common/stores/slices/authTokenSlice.ts
hasalarootcode: Narrowed `setAccessToken` to `(token: string)` so `clearAccessToken` is the only way to null it. Updated `StoreTypes.ts` as well.

frontend/src/community/auth/utils/authUtils.ts
hasalarootcode: Fixed. A `hasCheckedStoredToken` flag remembers the negative result for the page load, reset on sign-in and sign-out.

frontend/pages/api/auth/access-token.ts
```diff
@@ -0,0 +1,122 @@
...
+    req.query[TENANT_QUERY_PARAM]?.toString(),
+    req.cookies[TENANT_COOKIE_NAME]
+  );
+
+  const result = await requestSessionRefresh(
+    buildRefreshCookieHeader(cookies),
+    tenantId
+  );
+
+  return result.status === "success" ? result.session.accessToken : null;
```
thusala: rename to successful and import from common place

frontend/pages/api/auth/access-token.ts
```diff
@@ -0,0 +1,122 @@
...
+};
+
+export default async function handler(
+  req: NextApiRequest,
+  res: NextApiResponse<ResponseData>
+) {
+  if (req.method === "GET") {
```
thusala: const

frontend/pages/api/auth/access-token.ts
```diff
@@ -0,0 +1,122 @@
...
+    res.setHeader("Cache-Control", "no-store");
+
+    const storedToken = req.cookies[ACCESS_TOKEN_COOKIE_NAME];
+
+    if (storedToken && !isTokenExpired(storedToken)) {
+      return res
+        .status(200)
+        .json({ message: "Session cookie read", accessToken: storedToken });
```
thusala: remove message return

frontend/pages/api/auth/access-token.ts
```diff
@@ -0,0 +1,122 @@
...
+      buildSessionCookieHeader(
+        IS_PASSWORD_CHANGED_COOKIE_NAME,
+        String(isPasswordChangedForTheFirstTime),
+        PASSWORD_CHANGED_COOKIE_MAX_AGE_SECONDS
+      )
+    );
+  }
+
+  if (cookies.length === 0) {
+    return res.status(400).json({ message: "Nothing to set" });
```
thusala: return a messageKey in here. shall we try to return in BE pattern?

frontend/src/community/auth/providers/AuthProvider.tsx
```diff
@@ -84,8 +87,10 @@ export const AuthProvider: React.FC<AuthProviderProps> = ({ children }) => {
   const signIn = useCallback(
     async (params: EnterpriseSignInParams): Promise<AuthResponseType> => {
       try {
-        // Clear all existing cookies before signing in
-        await fetch("/api/clear-cookies", {
+        // Clear all existing cookies and the in memory token before signing in
+        useCommonStore.getState().clearAccessToken();
```
thusala: use the store hook and destructure and use it

frontend/src/community/auth/providers/AuthProvider.tsx
```diff
@@ -94,32 +99,19 @@ export const AuthProvider: React.FC<AuthProviderProps> = ({ children }) => {
...
-              window.location.href = redirectPath;
+            if (userData) {
+              window.location.href = resolvePostSignInPath(
+                router.query.callback,
+                router.asPath.split("?")[0]
+              );
             } else {
-              console.error("Access token cookie not found after sign-in");
-              throw new Error("Authentication cookies not properly set");
+              console.error("Access token not available after sign-in");
```
thusala: remove console error

frontend/src/community/auth/utils/authUtils.ts
```diff
@@ -86,10 +111,62 @@ export interface User {
...
+        return null;
+      }
+
+      const data = await response.json();
+
+      return typeof data?.accessToken === "string" && data.accessToken
+        ? data.accessToken
+        : null;
+    } catch {
+      console.error("Failed to retrieve the stored access token");
```
thusala: let's remove console error

frontend/src/community/auth/utils/authUtils.ts
```diff
@@ -86,10 +111,62 @@ export interface User {
...
+  }
+
+  if (hasCheckedStoredToken) {
+    return null;
+  }
+
+  retrievePromise = (async () => {
+    try {
+      const response = await fetch(internalApiEndpoints.ACCESS_TOKEN, {
+        method: "GET",
```
thusala: const

frontend/src/community/auth/utils/authUtils.ts
```diff
@@ -125,23 +202,62 @@ export const getNewAccessToken = async (): Promise<string | null> => {
...
 
-export const setAccessToken = (token: string) => {
-  if (typeof window !== "undefined") {
-    const expiryDate = new Date(
-      Date.now() + COOKIE_EXPIRY_DAYS * unitConversion.MILLISECONDS_PER_DAY
-    );
+export const setAccessToken = async (token: string): Promise<boolean> => {
+  if (typeof window === "undefined") return false;
+
+  useCommonStore.getState().setAccessToken(token);
```
thusala: store

frontend/src/community/auth/utils/authUtils.ts
```diff
@@ -125,23 +202,62 @@ export const getNewAccessToken = async (): Promise<string | null> => {
...
+  resetStoredTokenCheck();
 
-    document.cookie = `accessToken=${token}; path=/; expires=${expiryDate.toUTCString()}; Secure; SameSite=Lax`;
+  try {
+    const response = await fetch(internalApiEndpoints.ACCESS_TOKEN, {
+      method: "POST",
```
thusala: const and remove unnessacary

frontend/src/community/auth/utils/authUtils.ts
```diff
@@ -125,23 +202,62 @@ export const getNewAccessToken = async (): Promise<string | null> => {
...
-    const expiryDate = new Date(
-      Date.now() + COOKIE_EXPIRY_DAYS * unitConversion.MILLISECONDS_PER_DAY
-    );
+export const persistSession = async ({
+  accessToken,
+  isPasswordChangedForTheFirstTime
+}: SessionPayload): Promise<boolean> => {
+  if (typeof window === "undefined") return false;
+
+  useCommonStore.getState().setAccessToken(accessToken);
```
thusala: import store

frontend/src/community/auth/utils/edgeSessionUtils.ts
```diff
@@ -0,0 +1,158 @@
...
+const hasTenantSubdomain = (hostname: string, subdomain: string): boolean =>
+  hostname.split(".").length > 2 &&
+  subdomain !== LOCALHOST &&
+  !TENANT_SELECTION_SUBDOMAINS.includes(subdomain);
+
+export interface RefreshedSession {
+  accessToken: string;
+}
+
+export type RefreshSessionResult =
```
thusala, line 37: make this typesafe

frontend/src/community/common/stores/commonStore.ts
```diff
@@ -2,13 +2,19 @@ import { create } from "zustand";
...
 import { commonModalSlice } from "./slices/commonModalSlice";
 import { notificationsSlice } from "./slices/notificationsSlice";
 import { orgDetailsSlice } from "./slices/orgDetailsSlice";
 import { settingsModalSlice } from "./slices/settingsModalSlice";
 import { templateSlice } from "./slices/templateSlice";
 
+const REDACTED_VALUE = "<redacted>";
+
+const redactAccessToken = (state: CommonStoreTypes): CommonStoreTypes =>
+  state?.accessToken ? { ...state, accessToken: REDACTED_VALUE } : state;
```
thusala: remove

frontend/src/community/common/stores/commonStore.ts
```diff
@@ -20,8 +26,9 @@ export const useCommonStore = create<
       ...commonModalSlice(set),
       ...notificationsSlice(set),
       ...orgDetailsSlice(set),
-      ...breadcrumbSlice(set)
+      ...breadcrumbSlice(set),
+      ...authTokenSlice(set)
     }),
-    { name: "commonStore" }
+    { name: "commonStore", stateSanitizer: redactAccessToken }
```
thusala: revert

frontend/middleware.ts
```diff
@@ -200,180 +210,263 @@ const allowedRoutes: Record<
...
-        tenantStatus === TenantStatusEnums.TRIAL_ENDED_USER_LIMIT_EXCEEDED
-      ) {
-        return NextResponse.next();
-      } else {
-        return NextResponse.redirect(
-          new URL(ROUTES.DASHBOARD.BASE, request.url)
-        );
-      }
-    }
+const RESTRICTED_ROUTE_RULES: { routes: string[]; requiredRole: string }[] = [
```
thusala: move to interface { routes: string[]; requiredRole: string }

thusala (changes requested):

frontend/src/community/auth/providers/AuthProvider.tsx
hasalarootcode: Fixed

frontend/src/community/auth/providers/AuthProvider.tsx
hasalarootcode: Fixed

frontend/src/community/auth/utils/authUtils.ts
hasalarootcode: Fixed

frontend/src/community/auth/utils/authUtils.ts
hasalarootcode: Fixed

frontend/src/community/auth/utils/authUtils.ts
hasalarootcode: Fixed

frontend/src/community/auth/utils/authUtils.ts
hasalarootcode: Fixed

frontend/src/community/auth/utils/authUtils.ts
hasalarootcode: Fixed

frontend/src/community/common/stores/commonStore.ts
hasalarootcode: Fixed

frontend/src/community/common/stores/commonStore.ts
hasalarootcode: Fixed

frontend/middleware.ts
hasalarootcode: Fixed

frontend/pages/api/auth/access-token.ts
hasalarootcode: Fixed

frontend/pages/api/auth/access-token.ts
hasalarootcode: Fixed

frontend/pages/api/auth/access-token.ts
hasalarootcode: Fixed

frontend/pages/api/auth/access-token.ts
hasalarootcode: Fixed

thusala (approved):

shakila-kamalasena (approved):

akilarootcode (approved):

ThinuwanW (approved):

PR #1830, opened 2026-08-24
CRM deal list-view schema: migrations, baseline DDL & JPA entities

anusham-rc:
## Summary
Migration-phase change for the CRM deal table-view feature: schema + its JPA entity mappings.

**Baseline DDL** (`common-ddl-script-v1-create-tables.sql`)
- `crm_deal_order_index` table — per-view (`board`/`list`) fractional row ordering, `utf8mb4_bin`, FK to `crm_deal` `ON DELETE CASCADE`.
- `user_settings.crm_deal_list_view` JSON column — per-user column order/visibility + active sort.

**JPA entities**
- `CrmDealOrderIndex` — maps `crm_deal_order_index` (`@MapsId` 1:1 with `CrmDeal`).
- `UserSettings.crmDealListView` — maps the new JSON column.

## Related
- Liquibase migrations: rootcodelabs/skapp-migrations#214
- Config API (uses `UserSettings.crmDealListView`): SkappHQ/skapp#1831
- Row ordering / reorder / sorting (uses `CrmDealOrderIndex`): separate PR
- The backend code PRs depend on the entities here, so this PR merges first.

### DIFF
--- a/backend/src/main/java/com/skapp/community/common/model/UserSettings.java
+++ b/backend/src/main/java/com/skapp/community/common/model/UserSettings.java
@@ -38,4 +38,8 @@ public class UserSettings {
 	@Convert(converter = JsonTypeConverter.class)
 	private JsonNode notifications;
 
+	@Column(name = "crm_deal_list_view", columnDefinition = "json")
+	@Convert(converter = JsonTypeConverter.class)
+	private JsonNode crmDealListView;
+
 }
--- a/backend/src/main/java/com/skapp/community/common/model/UserSettings.java
+++ b/backend/src/main/java/com/skapp/community/common/model/UserSettings.java
@@ -38,7 +38,7 @@ public class UserSettings {
 	@Convert(converter = JsonTypeConverter.class)
 	private JsonNode notifications;
 
-	@Column(name = "crm_deal_list_view", columnDefinition = "json")
+	@Column(name = "crm_deal_list_view")
 	@Convert(converter = JsonTypeConverter.class)
 	private JsonNode crmDealListView;
 
--- a/backend/src/main/java/com/skapp/community/crmplanner/model/CrmDealOrderIndex.java
+++ b/backend/src/main/java/com/skapp/community/crmplanner/model/CrmDealOrderIndex.java
@@ -1,0 +1,37 @@
+package com.skapp.community.crmplanner.model;
+
+import jakarta.persistence.Column;
+import jakarta.persistence.Entity;
+import jakarta.persistence.FetchType;
+import jakarta.persistence.Id;
+import jakarta.persistence.JoinColumn;
+import jakarta.persistence.MapsId;
+import jakarta.persistence.OneToOne;
+import jakarta.persistence.Table;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+
+@Entity
+@Getter
+@Setter
+@NoArgsConstructor
+@Table(name = "crm_deal_order_index")
+public class CrmDealOrderIndex {
+
+	@Id
+	@Column(name = "deal_id", nullable = false, updatable = false)
+	private Long dealId;
+
+	@OneToOne(fetch = FetchType.LAZY)
+	@MapsId
+	@JoinColumn(name = "deal_id")
+	private CrmDeal deal;
+
+	@Column(name = "board", nullable = false)
+	private String board;
+
+	@Column(name = "list", nullable = false)
+	private String list;
+
+}
--- a/backend/src/main/java/com/skapp/community/crmplanner/model/CrmDealOrderIndex.java
+++ b/backend/src/main/java/com/skapp/community/crmplanner/model/CrmDealOrderIndex.java
@@ -5,7 +5,6 @@ import jakarta.persistence.Entity;
 import jakarta.persistence.FetchType;
 import jakarta.persistence.Id;
 import jakarta.persistence.JoinColumn;
-import jakarta.persistence.MapsId;
 import jakarta.persistence.OneToOne;
 import jakarta.persistence.Table;
 import lombok.Getter;
@@ -20,11 +19,7 @@ import lombok.Setter;
 public class CrmDealOrderIndex {
 
 	@Id
-	@Column(name = "deal_id", nullable = false, updatable = false)
-	private Long dealId;
-
 	@OneToOne(fetch = FetchType.LAZY)
-	@MapsId
 	@JoinColumn(name = "deal_id")
 	private CrmDeal deal;
 
--- a/backend/src/main/resources/community/db/changelog/ddl-script/common-ddl-script-v1-create-tables.sql
+++ b/backend/src/main/resources/community/db/changelog/ddl-script/common-ddl-script-v1-create-tables.sql
@@ -528,8 +528,9 @@ CREATE TABLE IF NOT EXISTS `rule_property`
 
 CREATE TABLE IF NOT EXISTS `user_settings`
 (
-    `user_id`       bigint NOT NULL,
-    `notifications` json DEFAULT NULL,
+    `user_id`            bigint NOT NULL,
+    `notifications`      json DEFAULT NULL,
+    `crm_deal_list_view` json DEFAULT NULL,
     PRIMARY KEY (`user_id`),
     CONSTRAINT `FK_user_settings_user_user_id` FOREIGN KEY (`user_id`) REFERENCES `user` (`user_id`)
 ) ENGINE = InnoDB;
@@ -676,6 +677,15 @@ CREATE TABLE IF NOT EXISTS `crm_deal`
     CONSTRAINT `FK_crm_deal_employee_owner_id` FOREIGN KEY (`owner_id`) REFERENCES `employee` (`employee_id`)
 ) ENGINE = InnoDB;
 
+CREATE TABLE IF NOT EXISTS `crm_deal_order_index`
+(
+    `deal_id` bigint                                         NOT NULL,
+    `board`   text CHARACTER SET utf8mb4 COLLATE utf8mb4_bin NOT NULL,
+    `list`    text CHARACTER SET utf8mb4 COLLATE utf8mb4_bin NOT NULL,
+    PRIMARY KEY (`deal_id`),
+    CONSTRAINT `FK_crm_deal_order_index_crm_deal_deal_id` FOREIGN KEY (`deal_id`) REFERENCES `crm_deal` (`id`) ON DELETE CASCADE
+) ENGINE = InnoDB;
+
 CREATE TABLE IF NOT EXISTS `crm_task`
 (
     `id`                 bigint  NOT NULL AUTO_INCREMENT,
@@ -703,6 +713,7 @@ CREATE TABLE IF NOT EXISTS `crm_task`
 ) ENGINE = InnoDB;
 
 -- rollback drop table crm_task;
+-- rollback drop table crm_deal_order_index;
 -- rollback drop table crm_deal;
 -- rollback drop table crm_deal_stage;
 -- rollback drop table crm_contact;
--- a/backend/src/main/resources/community/db/changelog/ddl-script/common-ddl-script-v1-create-tables.sql
+++ b/backend/src/main/resources/community/db/changelog/ddl-script/common-ddl-script-v1-create-tables.sql
@@ -530,7 +530,7 @@ CREATE TABLE IF NOT EXISTS `user_settings`
 (
     `user_id`            bigint NOT NULL,
     `notifications`      json DEFAULT NULL,
-    `crm_deal_list_view` json DEFAULT NULL,
+    `crm_deal_list_view` json,
     PRIMARY KEY (`user_id`),
     CONSTRAINT `FK_user_settings_user_user_id` FOREIGN KEY (`user_id`) REFERENCES `user` (`user_id`)
 ) ENGINE = InnoDB;
--- a/backend/src/main/resources/community/db/changelog/ddl-script/common-ddl-script-v1-create-tables.sql
+++ b/backend/src/main/resources/community/db/changelog/ddl-script/common-ddl-script-v1-create-tables.sql
@@ -683,7 +683,7 @@ CREATE TABLE IF NOT EXISTS `crm_deal_order_index`
     `board`   text CHARACTER SET utf8mb4 COLLATE utf8mb4_bin NOT NULL,
     `list`    text CHARACTER SET utf8mb4 COLLATE utf8mb4_bin NOT NULL,
     PRIMARY KEY (`deal_id`),
-    CONSTRAINT `FK_crm_deal_order_index_crm_deal_deal_id` FOREIGN KEY (`deal_id`) REFERENCES `crm_deal` (`id`) ON DELETE CASCADE
+    CONSTRAINT `FK_crm_deal_order_index_crm_deal_deal_id` FOREIGN KEY (`deal_id`) REFERENCES `crm_deal` (`id`)
 ) ENGINE = InnoDB;
 
 CREATE TABLE IF NOT EXISTS `crm_task`


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds a `crm_deal_order_index` entity/table and a `crm_deal_list_view` JSON column on `user_settings`. The schema changes are the serious problem: both edits are made *inside already-applied Liquibase changesets* in the baseline `common-ddl-script-v1-create-tables.sql` rather than in new changelog files, which will fail checksum validation on every existing deployment and — since `CREATE TABLE IF NOT EXISTS` is a no-op — never actually create the table/column, tripping Hibernate's `ddl-auto: validate`. Secondary concerns: `board`/`list` are `NOT NULL` with no backfill for existing deals, and the new `board` key duplicates the existing `crm_deal.order_index` fractional index with no deprecation path.

Found **15** new issue(s): 🔴 4 important, 🟡 7 suggestion(s), 🟣 4 nit(s)

backend/src/main/resources/community/db/changelog/ddl-script/common-ddl-script-v1-create-tables.sql
```diff
@@ -528,8 +528,9 @@ CREATE TABLE IF NOT EXISTS `rule_property`
 
 CREATE TABLE IF NOT EXISTS `user_settings`
 (
-    `user_id`       bigint NOT NULL,
-    `notifications` json DEFAULT NULL,
+    `user_id`            bigint NOT NULL,
+    `notifications`      json DEFAULT NULL,
+    `crm_deal_list_view` json DEFAULT NULL,
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Both schema edits modify changesets that have ALREADY been applied to every existing database, which will break startup.

The `crm_deal_list_view` column is added inside changeset `AkilaSachintha:common-ddl-script-v1-create-tables` (declared at line 3), and the `crm_deal_order_index` table (lines 680-687) plus the rollback comment (line 716) are inside `AkilaSachintha:common-ddl-script-v1-create-crm-tables` (declared at line 593). Neither changeset is marked `runOnChange` and neither has a `validCheckSum`.

Two consequences:

1. Liquibase computes a checksum per changeset from its contents. Existing DBs already have DATABASECHANGELOG rows for both of these, so on next boot Liquibase raises `ValidationFailedException` ("the file ... was changed") and the application fails to start. `application-prd.yml:24-26` and `application-non-prd.yml:23-24` both run this changelog with `enabled: true`.

2. Even if the checksums were cleared, `CREATE TABLE IF NOT EXISTS` is a no-op on an existing schema and a re-run `CREATE TABLE user_settings` is likewise skipped — so `crm_deal_order_index` and `user_settings.crm_deal_list_view` would still never be created. Both `application-prd.yml:15` and `application-non-prd.yml:15` set `ddl-auto: validate`, so Hibernate would then fail schema validation with "missing table [crm_deal_order_index]" / "missing column [crm_deal_list_view]".

Fix: revert `common-ddl-script-v1-create-tables.sql` to its original state and follow the convention used by every other schema change in this repo — add two new files under `ddl-script/`:
- `common-ddl-script-v1-alter-table-user-settings-add-crm-deal-list-view.sql` containing `ALTER TABLE \`user_settings\` ADD COLUMN \`crm_deal_list_view\` json DEFAULT NULL;` plus a `-- rollback ALTER TABLE ... DROP COLUMN ...` line (see `time-ddl-script-v1-alter-table-time-record-add-source-columns.sql` for the exact format).
- `crm-ddl-script-v1-create-table-crm-deal-order-index.sql` containing the CREATE TABLE plus `-- rollback DROP TABLE IF EXISTS \`crm_deal_order_index\`;` (see `people-ddl-script-v1-create-table-ppl-custom-skill.sql`).

Each needs its own `-- liquibase formatted sql` header and a unique `-- changeset <author>:<id>` line, and both must be appended as `include` entries at the end of `community/db/changelog/db.changelog.yml`.

backend/src/main/resources/community/db/changelog/ddl-script/common-ddl-script-v1-create-tables.sql
```diff
@@ -676,6 +677,15 @@ CREATE TABLE IF NOT EXISTS `crm_deal`
     CONSTRAINT `FK_crm_deal_employee_owner_id` FOREIGN KEY (`owner_id`) REFERENCES `employee` (`employee_id`)
 ) ENGINE = InnoDB;
 
+CREATE TABLE IF NOT EXISTS `crm_deal_order_index`
+(
+    `deal_id` bigint                                         NOT NULL,
+    `board`   text CHARACTER SET utf8mb4 COLLATE utf8mb4_bin NOT NULL,
```
ThinuwanW, line 683: 🤖 **Claude** · 🔴 **Important**

`board` and `list` are `NOT NULL` with no default, but nothing populates them — for existing rows or new ones.

Existing deployments already have `crm_deal` rows, and this PR ships no DML changeset to seed `crm_deal_order_index`. It also does not touch `CrmDealServiceImpl`, so the deal-creation paths (`CrmDealServiceImpl.java:148-149` and the bulk path at `:483-484`, which today only call `deal.setOrderIndex(FractionalIndexUtil.generateKeyBetween(...))`) still won't insert an order-index row. The table will be permanently empty until follow-up code lands, and by then it will be out of sync with `crm_deal`.

The practical failure: whatever query reads this next will either INNER JOIN and silently drop every pre-existing deal from the board/list view, or LEFT JOIN and get NULL sort keys that collapse ordering.

Fix: add a DML changeset alongside the DDL that backfills from the existing fractional index so current board ordering is preserved — `INSERT INTO crm_deal_order_index (deal_id, board, list) SELECT id, order_index, order_index FROM crm_deal;` (copying `order_index` into `board` is what preserves today's ordering; seeding `list` from it gives a deterministic starting order). Follow the naming pattern of `common-dml-script-v1-update-employee-role-set-crm-role.sql`. Then make every deal-creation path write the row in the same transaction.

backend/src/main/java/com/skapp/community/crmplanner/model/CrmDealOrderIndex.java
```diff
@@ -0,0 +1,37 @@
...
+	@Column(name = "deal_id", nullable = false, updatable = false)
+	private Long dealId;
+
+	@OneToOne(fetch = FetchType.LAZY)
+	@MapsId
+	@JoinColumn(name = "deal_id")
+	private CrmDeal deal;
+
+	@Column(name = "board", nullable = false)
+	private String board;
```
ThinuwanW, line 27: 🤖 **Claude** · 🟡 **Suggestion**

`board` duplicates the existing `CrmDeal.orderIndex` with no deprecation path, creating two writable sources of truth for the same ordering.

`CrmDeal.orderIndex` (`CrmDeal.java:49-50`, column `crm_deal.order_index`, `text NOT NULL`) is today's board ordering key. It is written by `CrmDealServiceImpl.reorderDeal` (`:318-320`), `updateDealStage` (`:358-362`), and `createDeal` (`:148-149`), all via `FractionalIndexUtil.generateKeyBetween`, and read by `CrmDealRepository.findMaxOrderIndexByStageId` / `findMinOrderIndexByStageId`. The new `board` column is the same fractional-index concept in a new home.

Shipping both leaves two columns that can silently diverge once anything writes to either. This is much cheaper to settle now, before production data exists in the new table. Decide explicitly and note it in the PR description:
- If `crm_deal_order_index` supersedes it: plan the follow-up changeset that drops `crm_deal.order_index` and removes `CrmDeal.orderIndex` + the two repository methods, and reference it here.
- If not: drop `board` from this table and keep it to `list` only, so `crm_deal.order_index` remains the single board key.

backend/src/main/java/com/skapp/community/common/model/UserSettings.java
```diff
@@ -38,4 +38,8 @@ public class UserSettings {
 	@Convert(converter = JsonTypeConverter.class)
 	private JsonNode notifications;
 
+	@Column(name = "crm_deal_list_view", columnDefinition = "json")
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

This puts CRM-specific state in the module-agnostic `common` entity, which doesn't scale.

`UserSettings` lives in `com.skapp.community.common.model` and currently holds only the generic `notifications` blob. Adding `crmDealListView` means the `common` module now carries knowledge of the `crmplanner` domain, and the precedent is that every future module adds its own column here — `okrListView`, `invoiceListView`, and so on — each requiring a schema migration on a shared table.

Two cleaner options:
- Keep a single generic column (e.g. `view_preferences` json) keyed by module/view name, so new per-module view state needs no DDL at all.
- Or own it in the CRM module: a `crm_user_preference` table with a `CrmUserPreference` entity under `com.skapp.community.crmplanner.model`, matching the module-boundary convention the rest of the codebase follows.

The second also keeps CRM columns out of a table read on every auth/user load.

backend/src/main/java/com/skapp/community/common/model/UserSettings.java
```diff
@@ -38,4 +38,8 @@ public class UserSettings {
...
+	@Convert(converter = JsonTypeConverter.class)
```
ThinuwanW, line 42: 🤖 **Claude** · 🟡 **Suggestion**

This is a raw, unbounded, unvalidated JSON blob with no schema.

MySQL's `json` type accepts values up to ~1GB, and `spring.servlet.multipart.max-file-size: 5MB` (application-prd.yml:29-31) only bounds multipart uploads — it does not limit a JSON request body. Whatever endpoint eventually writes `crmDealListView` should validate before persisting: allow-list the permitted column/view keys, bound array sizes, and reject anything oversized, rather than storing whatever the client sends.

Worth noting alongside this: `JsonTypeConverter.convertToDatabaseColumn` uses `jsonNode.toPrettyString()`, so every stored value is inflated with indentation and newlines. For a per-user view-config blob that is read on user load, `toString()` would be meaningfully smaller with no loss.

backend/src/main/resources/community/db/changelog/ddl-script/common-ddl-script-v1-create-tables.sql
```diff
@@ -676,6 +677,15 @@ CREATE TABLE IF NOT EXISTS `crm_deal`
...
+    `list`    text CHARACTER SET utf8mb4 COLLATE utf8mb4_bin NOT NULL,
+    PRIMARY KEY (`deal_id`),
+    CONSTRAINT `FK_crm_deal_order_index_crm_deal_deal_id` FOREIGN KEY (`deal_id`) REFERENCES `crm_deal` (`id`) ON DELETE CASCADE
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`ON DELETE CASCADE` here is inconsistent with the CRM schema and is invisible to Hibernate.

Every other CRM foreign key omits cascade — including `FK_crm_task_crm_deal_deal_id` (line 711), which points at the same parent table. CRM entities also use soft deletes (`is_deleted boolean NOT NULL DEFAULT FALSE` on `crm_deal`, `crm_task`, `crm_contact`, `crm_company`), so a hard `DELETE FROM crm_deal` should essentially never happen and the cascade will never fire in normal operation. The team has also moved away from this pattern elsewhere — see `people-ddl-script-v1-alter-table-com-work-location-geofence-drop-work-location-cascade.sql`.

Separately, a DB-level cascade is not visible to Hibernate: if a deal is ever hard-deleted through JPA, the persistence context won't know the `crm_deal_order_index` row is gone and can serve a stale cached instance.

Either drop `ON DELETE CASCADE` to match the rest of the CRM schema, or — if the cascade is intentional — add `@OnDelete(action = OnDeleteAction.CASCADE)` on the `deal` association in `CrmDealOrderIndex` so Hibernate is aware of it.

backend/src/main/resources/community/db/changelog/ddl-script/common-ddl-script-v1-create-tables.sql
ThinuwanW, line 683: 🤖 **Claude** · 🟡 **Suggestion**

`board` and `list` are `text`, which MySQL cannot index without a prefix length — in a table whose entire purpose is sorting.

Every `ORDER BY o.board` / `ORDER BY o.list` against this table will be a filesort over the joined rows, with no index available. Fractional index keys from `FractionalIndexUtil.generateKeyBetween` are short strings, so `varchar(191) CHARACTER SET utf8mb4 COLLATE utf8mb4_bin NOT NULL` would hold them comfortably while remaining indexable; add `KEY idx_crm_deal_order_index_board (board)` and the equivalent for `list`.

(`crm_deal.order_index` has the same limitation, which is presumably where the `text` came from — but a purpose-built ordering table is the right place to fix it rather than inherit it.) If you change the type, also set `@Column(length = 191)` on the corresponding fields in `CrmDealOrderIndex` so the mapping stays explicit under `ddl-auto: validate`.

The `utf8mb4_bin` collation itself is correct and worth keeping — byte-ordered comparison is exactly what fractional index keys need.

backend/src/main/java/com/skapp/community/crmplanner/model/CrmDealOrderIndex.java
```diff
@@ -0,0 +1,37 @@
...
+
+	@Column(name = "list", nullable = false)
```
ThinuwanW, line 29: 🤖 **Claude** · 🟣 **Nit**

`list` is a fragile choice for a column and field name.

It is a MySQL keyword (non-reserved in 8.x, so it happens to work unquoted today) and Hibernate emits it unquoted in generated SQL — including against H2 in the test suite, which builds the schema from entities (`ddl-auto: create`, `backend/src/test/resources/application.yml:16`). It also reads poorly next to the existing `crm_deal.order_index`.

Renaming to `list_order` / `listOrder` (and correspondingly `board_order` / `boardOrder`) removes the keyword risk entirely and makes both columns self-describing as ordering keys. Worth doing now, while the table has no data and no dependent code.

backend/src/main/java/com/skapp/community/crmplanner/model/CrmDealOrderIndex.java
```diff
@@ -0,0 +1,37 @@
...
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+
+@Entity
+@Getter
+@Setter
+@NoArgsConstructor
+@Table(name = "crm_deal_order_index")
+public class CrmDealOrderIndex {
```
ThinuwanW, line 19: 🤖 **Claude** · 🟣 **Nit**

This entity does not extend `Auditable<String>`, unlike its CRM siblings.

`CrmDeal`, `CrmDealStage`, `CrmContact`, `CrmCompany`, and `CrmTask` all extend it, and their tables carry `created_by` / `created_date` / `last_modified_by` / `last_modified_date`. `board` and `list` are user-mutable through drag-and-drop reordering, so `last_modified_by` / `last_modified_date` would be genuinely useful when debugging "who moved this deal" reports.

The one CRM exception is `CrmTaskType`, which is static reference data — not really comparable to a hot user-write table. If audit columns are being deliberately skipped here to keep reorder writes cheap, that's a defensible trade-off; please state it in the PR description so it doesn't read as an oversight. Note that adding them requires the four columns in the DDL as well.

backend/src/main/java/com/skapp/community/crmplanner/model/CrmDealOrderIndex.java
```diff
@@ -0,0 +1,37 @@
+package com.skapp.community.crmplanner.model;
```
ThinuwanW, line 1: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

The new entity is entirely unreferenced — a grep across `backend/src` finds zero mentions of `CrmDealOrderIndex` outside its own file. There is no `CrmDealOrderIndexDao`, and no service, repository, or mapper touches it.

That's fine if this is a deliberate schema-first PR, but it means the JPA mapping is completely unexercised. In particular, the `@MapsId` derived-identity mapping requires callers to set the `deal` association before persist — setting only `dealId` will throw at flush time — and nothing in the H2 test suite currently persists or reads a row to catch that.

Consider landing the `CrmDealOrderIndexDao` plus a small repository test in this PR so the mapping is verified before the consuming feature is built on top of it.

backend/src/main/resources/community/db/changelog/ddl-script/common-ddl-script-v1-create-tables.sql
```diff
@@ -676,6 +677,15 @@ CREATE TABLE IF NOT EXISTS `crm_deal`
     CONSTRAINT `FK_crm_deal_employee_owner_id` FOREIGN KEY (`owner_id`) REFERENCES `employee` (`employee_id`)
 ) ENGINE = InnoDB;
 
+CREATE TABLE IF NOT EXISTS `crm_deal_order_index`
```
ThinuwanW, line 680: 🤖 **Claude** · 🔴 **Important**

Nothing in CI verifies that this hand-written MySQL DDL agrees with the JPA entities, so the entire risk surface of this PR is untested. `backend/src/test/resources/application.yml` sets `spring.liquibase.enabled: false` and `spring.jpa.hibernate.ddl-auto: create` against an in-memory H2 datasource — meaning the test suite builds its schema from the annotations and never executes this changelog. Production runs the opposite way (`application-prd.yml` / `application-non-prd.yml` set `ddl-auto: validate` with Liquibase enabled), so any mismatch between the new `crm_deal_order_index` DDL and `CrmDealOrderIndex` — a misspelled column, a missing column, an incompatible type — passes green in CI and first surfaces as a `SchemaManagementException` that aborts application startup at deploy time. This is exactly how the first-pass migration issues would reach production undetected. Fix: add a Testcontainers MySQL integration test (or a dedicated `@SpringBootTest` profile) that runs the Liquibase changelog against a real MySQL image with `ddl-auto: validate`, so entity/DDL drift is a build failure rather than a deploy failure. At minimum, add this schema-vs-entity check for the CRM module before merging.

backend/src/main/java/com/skapp/community/common/model/UserSettings.java
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

This new JSON column is loaded and Jackson-parsed on essentially every authenticated request, not just on CRM screens. `User.settings` is declared as `@OneToOne(mappedBy = "user", cascade = CascadeType.ALL)` in `User.java:66` with no `fetch` attribute — the JPA default for `@OneToOne` is EAGER, and an inverse (`mappedBy`) one-to-one cannot be proxied lazily without bytecode enhancement anyway. `UserServiceImpl.getCurrentUser()` does a `userDao.findByEmail(...)` on every invocation (and it is invoked from most service methods, often more than once per request), so every one of those now additionally reads an unbounded blob off the wire and runs it through `JsonTypeConverter`. That converter also does `new ObjectMapper()` on every single `convertToEntityAttribute` call (`JsonTypeConverter.java:29`) — `ObjectMapper` is an expensive, thread-safe object explicitly designed to be constructed once and shared — so this change doubles the per-user-load mapper allocations from one to two. Two fixes, both worth doing: (1) hoist a `private static final ObjectMapper MAPPER` in `JsonTypeConverter` instead of allocating per read; (2) keep per-screen UI preferences out of the eagerly-fetched `UserSettings` — put them in a dedicated table/entity keyed by user id, fetched lazily and only by the CRM endpoints that need them.

backend/src/main/java/com/skapp/community/common/model/UserSettings.java
ThinuwanW, line 42: 🤖 **Claude** · 🟡 **Suggestion**

`JsonTypeConverter.convertToEntityAttribute` calls `readTree(...)` with no error handling (`JsonTypeConverter.java:24-32`). In Jackson 3 (`tools.jackson`) the parse failure type is unchecked, so it compiles fine and escapes entity hydration at runtime. Combined with the eager fetch of `User.settings` described above, an unparseable value in `crm_deal_list_view` does not degrade the CRM screen — it makes every `getCurrentUser()` call blow up, so the user is locked out of the whole application, and the exception falls through to the catch-all `@ExceptionHandler(Exception.class)` in `GlobalExceptionHandler.java:178` and surfaces as a generic 500 with the `COMMON_ERROR_MODULE_EXCEPTION` key rather than anything diagnosable. This is a pre-existing gap in shared code, but this PR is what routes a second, client-shaped payload through it. The repo already has the right idiom: `InternalServerException`'s own javadoc names "a stored value is corrupt" as its intended use case. Wrap the `readTree` call in a try/catch and rethrow `new InternalServerException(<new key>)`, adding the key to `CommonMessageConstant` and `common-messages.properties`. Note that the message-key/properties pairing is not build-verified in this repo, so add both in the same commit.

backend/src/main/resources/community/db/changelog/ddl-script/common-ddl-script-v1-create-tables.sql
```diff
@@ -676,6 +677,15 @@ CREATE TABLE IF NOT EXISTS `crm_deal`
...
+(
+    `deal_id` bigint                                         NOT NULL,
```
ThinuwanW, line 682: 🤖 **Claude** · 🟡 **Suggestion**

There is no reachable cleanup path for rows in this table, so it will grow monotonically and never shrink. Every CRM delete in this codebase is a soft delete — `CrmDealServiceImpl.deleteDeal` sets `isDeleted = true` and saves (line 535), `CrmContactServiceImpl` does the same for cascaded deals (line 247), and a grep for `crmDealDao.delete`/`deleteById` across `crmplanner` returns nothing. A `crm_deal` row is therefore never physically removed, which means the row-level FK action on this constraint can never actually trigger, and no application code deletes the index row either. Two consequences to decide on explicitly: (a) index rows for deleted deals accumulate forever, and (b) any future query that reads this table must join `crm_deal` and filter `is_deleted = false`, or deleted deals reappear in board/list ordering. Pick one and make it explicit — either delete the index row inside `deleteDeal` (and inside the contact-cascade path in `CrmContactServiceImpl`, which soft-deletes deals in bulk), or document that retention is intentional and make the soft-delete filter part of the repository contract for every read of this table.

backend/src/main/java/com/skapp/community/crmplanner/model/CrmDealOrderIndex.java
```diff
@@ -0,0 +1,37 @@
...
+@Entity
+@Getter
+@Setter
+@NoArgsConstructor
+@Table(name = "crm_deal_order_index")
+public class CrmDealOrderIndex {
+
+	@Id
+	@Column(name = "deal_id", nullable = false, updatable = false)
+	private Long dealId;
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

The class-level `@Setter` generates a public `setDealId(...)` that is a silent no-op for persistence. Under `@MapsId`, Hibernate derives the `deal_id` value from the `deal` association at flush time and treats the `@Id` basic attribute as non-insertable/non-updatable, so the most natural-looking usage — construct via the no-arg constructor, set `dealId`, set `board`/`list`, save — will not persist the id from the setter and will fail at flush with an identifier-generation error instead. The trap is easy to fall into here because the id field and the association are both named after `deal_id`, unlike the repo's other derived-identity entities (`UserSettings`, `EmployeeRole`, `EmployeePersonalInfo`) where the two names differ. Since the entity has no callers yet, close the hole now: annotate the id field with `@Setter(AccessLevel.NONE)` and add an explicit constructor taking `(CrmDeal deal, String board, String list)` so the association is always the single source of the identifier.

shakila-kamalasena (dismissed):

backend/src/main/resources/community/db/changelog/ddl-script/common-ddl-script-v1-create-tables.sql
anusham-rc: This baseline’s changesets are edited in place by established team convention — e.g. adding first_name/last_name to `crm_contact`, `order_index` to `crm_deal`, and the utf8mb4 collation change all landed the same way. Real tenant migrations go through the skapp-migrations repo, which does use new changelog files. Keeping the repo convention here.

backend/src/main/resources/community/db/changelog/ddl-script/common-ddl-script-v1-create-tables.sql
anusham-rc: Backfill is intentionally deferred to a follow-up (an admin migration endpoint that seeds `board` from `crm_deal.order_index` and generates `list` keys). Read/reorder paths tolerate missing rows until then.

backend/src/main/java/com/skapp/community/crmplanner/model/CrmDealOrderIndex.java
anusham-rc: Intentional coexistence: `board` mirrors `crm_deal.order_index` during the transition; dropping the old column is a deferred follow-up once the app fully reads/writes the new table.

backend/src/main/java/com/skapp/community/common/model/UserSettings.java
anusham-rc: Deliberate — mirrors how `user_settings` already stores per-concern config (`notifications`). One column per view; the JSON contract lives with the backend.

backend/src/main/java/com/skapp/community/common/model/UserSettings.java
anusham-rc: By design — an opaque, frontend-owned view-config blob. The backend stores/returns it as-is and treats NULL as absent config, falling back to defaults.

backend/src/main/resources/community/db/changelog/ddl-script/common-ddl-script-v1-create-tables.sql
anusham-rc: Acknowledged, harmless: `crm_deal` is soft-deleted so the cascade rarely fires, and board/list reads filter on `is_deleted`, so ordering rows for deleted deals are never surfaced.

backend/src/main/resources/community/db/changelog/ddl-script/common-ddl-script-v1-create-tables.sql
anusham-rc: Consistent with the existing `crm_deal.order_index` (also `TEXT utf8mb4_bin`, migration 00121). Neighbour lookups are keyed by `deal_id` (PK). Can move to `varchar(255)` + index if profiling shows a need.

backend/src/main/java/com/skapp/community/crmplanner/model/CrmDealOrderIndex.java
anusham-rc: `list` is a non-reserved keyword and is backticked in the DDL / quoted by Hibernate. Working as intended.

backend/src/main/java/com/skapp/community/crmplanner/model/CrmDealOrderIndex.java
anusham-rc: Intentional — this is a pure link/order table, matching the PM `item_order_index` precedent (also non-audited), as noted.

backend/src/main/java/com/skapp/community/crmplanner/model/CrmDealOrderIndex.java
anusham-rc: False positive from the PR split — this PR is the schema+entity layer; the DAO/service that reference `CrmDealOrderIndex` are in the sibling ordering PR.

backend/src/main/resources/community/db/changelog/ddl-script/common-ddl-script-v1-create-tables.sql
anusham-rc: Fair forward-looking suggestion, but out of scope for this schema PR. Worth tracking separately.

backend/src/main/java/com/skapp/community/common/model/UserSettings.java
anusham-rc: The added column is one small blob, and the `new ObjectMapper()` per read lives in the existing `JsonTypeConverter` (unchanged by this PR), so it is not a regression introduced here.

backend/src/main/java/com/skapp/community/common/model/UserSettings.java
anusham-rc: Pre-existing converter behaviour, not modified in this PR.

backend/src/main/resources/community/db/changelog/ddl-script/common-ddl-script-v1-create-tables.sql
anusham-rc: Same soft-delete point — rows for soft-deleted deals are filtered on read; acceptable for an ordering table.

backend/src/main/java/com/skapp/community/crmplanner/model/CrmDealOrderIndex.java
anusham-rc: Nit acknowledged — `@MapsId` derives the id from the association, so the setter is intentionally unused.

backend/src/main/java/com/skapp/community/common/model/UserSettings.java
akilarootcode: remove columnDefinition = "json"

backend/src/main/java/com/skapp/community/crmplanner/model/CrmDealOrderIndex.java
```diff
@@ -0,0 +1,37 @@
...
+
+	@OneToOne(fetch = FetchType.LAZY)
+	@MapsId
```
akilarootcode: remove mapsId

backend/src/main/resources/community/db/changelog/ddl-script/common-ddl-script-v1-create-tables.sql
akilarootcode: remove  ON DELETE CASCADE

akilarootcode (changes requested):

backend/src/main/java/com/skapp/community/common/model/UserSettings.java
anusham-rc: removed

backend/src/main/resources/community/db/changelog/ddl-script/common-ddl-script-v1-create-tables.sql
anusham-rc: removed

backend/src/main/java/com/skapp/community/crmplanner/model/CrmDealOrderIndex.java
anusham-rc: removed

shakila-kamalasena (approved):

akilarootcode (approved):

ThinuwanW (approved):

PR #1831, opened 2026-08-24
Add per-user CRM deal list-view config API

anusham-rc:
## Summary
Per-user CRM deal **list (table) view** config API, added to the existing **`CrmDealController` (v1)** — mirroring how per-user settings (e.g. notification settings on `PeopleController`) are hosted on a domain controller rather than a standalone one.

- `GET /v1/crm/deal/list-view-config` — the user's saved config JSON, or a **default** when nothing is saved.
- `PUT /v1/crm/deal/list-view-config` — persists the posted JSON as-is to `user_settings.crm_deal_list_view`.

The config is an **opaque JSON blob owned by the frontend** — the backend stores/returns it, no server-side field DTOs. `DefaultCrmDealListViewConfig` supplies the first-load default. Logic sits in a dedicated `CrmDealListViewConfigService` injected into the controller (kept separate from `CrmDealService` because the enterprise `EpCrmDealServiceImpl` subclass constructor can't take new deps).

## Depends on
- **SkappHQ/skapp#1830** — the `UserSettings.crmDealListView` entity/column lives there; this PR references it, so **#1830 must merge first**.

## Related
- Migrations: rootcodelabs/skapp-migrations#214


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds GET/PUT `/v1/crm/deal/list-view-config` endpoints that read and persist a per-user CRM deal table config as an opaque JSON blob on `UserSettings`. The blocking problem is that the persistence side does not exist: `UserSettings` has no `crmDealListView` field (so `getCrmDealListView()`/`setCrmDealListView()` will not compile) and there is no Liquibase changeset adding the column to the `user_settings` table. Beyond that, the PUT endpoint accepts and stores completely unvalidated, unbounded JSON, the new service has no logging despite `@Slf4j`, there are no tests, and the hardcoded default config does not actually match the columns the deal table ships with.

Found **20** new issue(s): 🔴 5 important, 🟡 8 suggestion(s), 🟣 7 nit(s)

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealListViewConfigServiceImpl.java
```diff
@@ -0,0 +1,55 @@
...
+	private final UserDao userDao;
+
+	private final ObjectMapper objectMapper;
+
+	@Override
+	@Transactional(readOnly = true)
+	public ResponseEntityDto getListViewConfig() {
+		User user = userService.getCurrentUser();
+		UserSettings settings = user.getSettings();
+		JsonNode saved = settings != null ? settings.getCrmDealListView() : null;
```
ThinuwanW, line 39: 🤖 **Claude** · 🔴 **Important**

`UserSettings` has no `crmDealListView` property. I read `backend/src/main/java/com/skapp/community/common/model/UserSettings.java` in full — it only declares `id`, `user`, and `notifications` — and grepped the whole repo for `crmDealListView` / `crm_deal_list_view`: the only hits are in this PR's own files. So both `settings.getCrmDealListView()` (line 33) and `settings.setCrmDealListView(config)` (line 49) will fail to compile. The schema half is missing too: `user_settings` is created in `community/db/changelog/ddl-script/common-ddl-script-v1-create-tables.sql` with only `user_id` and `notifications`, and there is no new changeset adding a JSON column. To complete this: (1) add `@Column(name = "crm_deal_list_view", columnDefinition = "json") @Convert(converter = JsonTypeConverter.class) private JsonNode crmDealListView;` to `UserSettings` (mirroring the existing `notifications` field), and (2) add a new `common-ddl-script-v1-alter-table-user-settings-add-crm-deal-list-view.sql` under `community/db/changelog/ddl-script/` with `ALTER TABLE user_settings ADD COLUMN crm_deal_list_view json DEFAULT NULL;` plus a rollback comment, and register it with an `- include:` entry in `community/db/changelog/db.changelog.yml`. Never rely on `ddl-auto` for this.

backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmDealController.java
```diff
@@ -81,4 +86,22 @@ public ResponseEntity<ResponseEntityDto> deleteDeal(@PathVariable Long id) {
...
+
+	@Operation(summary = "Update the current user's deal list-view config",
+			description = "Persists the deal table config JSON sent by the client as-is; does not affect other users.")
+	@PutMapping("/list-view-config")
+	@PreAuthorize("hasAnyRole('ROLE_CRM_SALES_REPRESENTATIVE')")
+	public ResponseEntity<ResponseEntityDto> updateListViewConfig(@RequestBody JsonNode config) {
+		ResponseEntityDto response = crmDealListViewConfigService.updateListViewConfig(config);
+		return new ResponseEntity<>(response, HttpStatus.OK);
+	}
+
```
ThinuwanW, line 118: 🤖 **Claude** · 🔴 **Important**

The PUT body is an unvalidated, unbounded `JsonNode` persisted verbatim. Three concrete problems: (1) Any JSON is accepted — `"hello"`, `42`, `true`, `[]` are all valid `JsonNode`s and will be stored, then handed straight back to the frontend on the next GET, breaking the deal table for that user with no way to recover except another PUT. (2) There is no size bound — a client can PUT a multi-megabyte blob per user; `JsonTypeConverter.convertToDatabaseColumn` even calls `toPrettyString()`, inflating it further before it hits the `json` column. (3) Sending literal `null` stores `NullNode`, which the converter maps to SQL NULL, and the echoed response `results` array then contains a bare `null` — if reset-to-default is the intent it should be an explicit operation, not an accident of the type. Fix: introduce a proper `CrmDealListViewConfigRequestDto` (the codebase uses `*RequestDto` everywhere else, and it also gives Swagger a real schema), or at minimum validate in the service before saving — require `config != null && config.isObject()`, require a `fields` array whose entries have a known `field` value and a numeric `width`, and cap the serialized length / number of fields. Reject with `throw new ModuleException(CrmMessageConstant.<NEW_KEY>)` and add the matching key to `CrmMessageConstant` plus `crm-messages.properties`, so failures come back as a 400 through `GlobalExceptionHandler` instead of being silently persisted.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealListViewConfigServiceImpl.java
```diff
@@ -0,0 +1,55 @@
...
+
+	private final UserService userService;
+
+	private final UserDao userDao;
+
+	private final ObjectMapper objectMapper;
+
+	@Override
+	@Transactional(readOnly = true)
+	public ResponseEntityDto getListViewConfig() {
```
ThinuwanW, line 34: 🤖 **Claude** · 🟡 **Suggestion**

The class is annotated `@Slf4j` but neither method logs anything, so the annotation is currently dead. Every other service in this module follows the convention (see `CrmTaskTypeServiceImpl.getTaskTypes` and `PeopleServiceImpl.updateNotificationSettings`): `log.info("getListViewConfig: execution started")` at the top and `log.info("getListViewConfig: execution ended")` before the return, likewise for `updateListViewConfig`. Add these. Do not log the config payload itself, just the method boundaries.

backend/src/main/java/com/skapp/community/crmplanner/constant/DefaultCrmDealListViewConfig.java
```diff
@@ -0,0 +1,48 @@
...
+	}
+
+	public static JsonNode build(ObjectMapper objectMapper) {
+		ArrayNode fields = objectMapper.createArrayNode();
+		fields.add(field(objectMapper, "DEAL_NAME", 400));
+		fields.add(field(objectMapper, "VALUE", 140));
+		fields.add(field(objectMapper, "STAGE", 140));
+		fields.add(field(objectMapper, "COMPANY_NAME", 140));
+		fields.add(field(objectMapper, "CONTACT_NAME", 140));
+		fields.add(field(objectMapper, "PRIORITY", 140));
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The javadoc claims this default matches "the columns the deal table ships with", but it does not. `frontend/src/community/crm/components/organisms/DealsTable/DealsTable.tsx` and `DealsTableV2.tsx` both define exactly six columns — dealName, value, stage, companyName, contactName, dealOwner — with no PRIORITY column (grepping `priority` in `DealsTableV2.tsx` returns nothing). This default emits a seventh `PRIORITY` field the table cannot render. Separately, both tables hardcode `sortable: false` and `draggable: false` on every column, while this default sets `isSortable: true` and `isDraggable: true` for all seven. Either the frontend column set is being extended in a companion change (in which case say so and keep them in sync), or drop `PRIORITY` and set `isSortable`/`isDraggable` to `false` so a brand-new user's default config actually reflects what the UI supports.

backend/src/main/java/com/skapp/community/crmplanner/constant/DefaultCrmDealListViewConfig.java
```diff
@@ -0,0 +1,48 @@
...
+ * order is the default column order; {@code sort} is null (no active sort).
+ */
+public final class DefaultCrmDealListViewConfig {
+
+	private DefaultCrmDealListViewConfig() {
+	}
+
+	public static JsonNode build(ObjectMapper objectMapper) {
+		ArrayNode fields = objectMapper.createArrayNode();
+		fields.add(field(objectMapper, "DEAL_NAME", 400));
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The column identifiers `DEAL_NAME`, `VALUE`, `STAGE`, `COMPANY_NAME`, `CONTACT_NAME`, `PRIORITY`, `DEAL_OWNER` are bare string literals that mirror the frontend's `DEAL_TABLE_COLUMN_WIDTH_RATIO` keys in `frontend/src/community/crm/constants/dealConstants.ts` with nothing tying the two together. Add an enum in `com.skapp.community.crmplanner.type` (e.g. `CrmDealListViewField`) holding the identifier and its default width, and build the default from `CrmDealListViewField.values()`. That gives you a single source of truth and makes the request-body validation suggested on the controller trivial to implement (reject unknown `field` values). It also matches how the module already handles defaults — `DefaultCrmDealStageTemplate` builds from the `DefaultCrmDealStageValues` enum rather than literals.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealListViewConfigServiceImpl.java
```diff
@@ -0,0 +1,55 @@
+package com.skapp.community.crmplanner.service.impl;
```
ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Architectural concern worth a second look before this lands. This is the only class under `com.skapp.community.crmplanner` that imports a `com.skapp.community.common.repository.*` DAO (I grepped the whole crmplanner package — `UserDao` here is the sole hit); every other CRM service goes through its own `Crm*Dao`. It also mutates a shared common-module entity from a feature module. On top of that, putting a CRM-specific column on the shared `user_settings` table means each future module that wants a saved list view adds another column to the same table. Two cleaner options: (a) expose the read/write through a common `UserSettingsService` interface so the CRM module does not reach into `UserDao` directly, or (b) model this as a generic per-user view-preference row (user_id + view_key + config json) that other modules can reuse. If the shared-column approach is a deliberate, time-boxed call, it would help to note that in the class javadoc.

backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmDealController.java
```diff
@@ -81,4 +86,22 @@ public ResponseEntity<ResponseEntityDto> deleteDeal(@PathVariable Long id) {
 		return new ResponseEntity<>(response, HttpStatus.OK);
 	}
 
+	@Operation(summary = "Get the current user's deal list-view config",
+			description = "Returns the user's saved deal table config JSON, or the default config when nothing is saved.")
+	@GetMapping("/list-view-config")
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

No tests were added for either new endpoint. `backend/src/test/java/com/skapp/community/crmplanner/controller/v1/CrmDealControllerIntegrationTest.java` already exists with the `@SpringBootTest` + `@AutoConfigureMockMvc` harness and a `BASE_PATH = "/v1/crm/deal"` constant, so extending it is cheap. Worth covering at minimum: GET returns the default config when the user has no `UserSettings` row; GET returns the default when a `UserSettings` row exists but the column is null; PUT then GET round-trips the saved config; PUT by user A does not change user B's config; and an unauthorized role gets 403. The round-trip test in particular would have caught the missing entity field.

backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmDealController.java
```diff
@@ -81,4 +86,22 @@ public ResponseEntity<ResponseEntityDto> deleteDeal(@PathVariable Long id) {
...
+		return new ResponseEntity<>(response, HttpStatus.OK);
+	}
+
+	@Operation(summary = "Update the current user's deal list-view config",
+			description = "Persists the deal table config JSON sent by the client as-is; does not affect other users.")
+	@PutMapping("/list-view-config")
+	@PreAuthorize("hasAnyRole('ROLE_CRM_SALES_REPRESENTATIVE')")
+	public ResponseEntity<ResponseEntityDto> updateListViewConfig(@RequestBody JsonNode config) {
+		ResponseEntityDto response = crmDealListViewConfigService.updateListViewConfig(config);
+		return new ResponseEntity<>(response, HttpStatus.OK);
```
ThinuwanW, line 116: 🤖 **Claude** · 🟣 **Nit**

`GlobalExceptionHandler` does not extend `ResponseEntityExceptionHandler` and has no `@ExceptionHandler(HttpMessageNotReadableException.class)`, so a malformed or empty PUT body falls through to the catch-all `@ExceptionHandler(Exception.class)` and returns 500 rather than 400. This is pre-existing and affects every `@RequestBody` endpoint, but this endpoint is the one explicitly designed to receive free-form client JSON, so it is the most likely to hit it. Consider adding a `HttpMessageNotReadableException` handler returning 400 as part of this change.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealListViewConfigServiceImpl.java
```diff
@@ -0,0 +1,55 @@
...
+@Slf4j
+@Service
+@RequiredArgsConstructor
+public class CrmDealListViewConfigServiceImpl implements CrmDealListViewConfigService {
+
+	private final UserService userService;
+
+	private final UserDao userDao;
+
+	private final ObjectMapper objectMapper;
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Consistency: everywhere else in the codebase the Jackson mapper is injected as `tools.jackson.databind.json.JsonMapper` (`PeopleServiceImpl:214`, `EpUserServiceImpl:49`, `EpGuestUserCacheServiceImpl:32`, and all the integration tests), which is the type Spring Boot 4 auto-configures and the type `JacksonConfig`'s `JsonMapperBuilderCustomizer` beans apply to. Injecting the `ObjectMapper` supertype resolves to the same bean by assignability so it should work, but switching to `JsonMapper` matches the convention and removes any doubt. As a smaller point, the default config is rebuilt from scratch on every GET that has no saved config — since it is immutable in practice, you could build it once into a `static final JsonNode` (or return a deep copy of a cached node) instead of taking an `ObjectMapper` parameter.

backend/src/main/java/com/skapp/community/crmplanner/constant/DefaultCrmDealListViewConfig.java
```diff
@@ -0,0 +1,48 @@
...
+import tools.jackson.databind.node.ArrayNode;
+import tools.jackson.databind.node.ObjectNode;
+
+/**
+ * The default CRM deal list-view config returned when a user has not saved one. The
+ * config itself is an opaque JSON blob owned by the frontend; this default is the only
+ * shape the backend keeps, matching the columns the deal table ships with. {@code fields}
+ * order is the default column order; {@code sort} is null (no active sort).
+ */
+public final class DefaultCrmDealListViewConfig {
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

The sibling class in this package, `DefaultCrmDealStageTemplate`, uses Lombok's `@UtilityClass` rather than a hand-written `final class` + private constructor. Using `@UtilityClass` here too would keep the two consistent. Also, the class lives in `constant/` but is a builder rather than a constant holder — `DefaultCrmDealListViewConfigTemplate` (or moving it beside the enum suggested above) would read more accurately.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealListViewConfigServiceImpl.java
```diff
@@ -0,0 +1,55 @@
...
+	@Transactional
+	public ResponseEntityDto updateListViewConfig(JsonNode config) {
+		User user = userService.getCurrentUser();
+		UserSettings settings = user.getSettings();
+		if (settings == null) {
+			settings = new UserSettings();
+			settings.setUser(user);
+			user.setSettings(settings);
+		}
+		settings.setCrmDealListView(config);
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Missing paired schema change — this will break application startup, not just fail to compile. `user_settings` is created in `community/db/changelog/ddl-script/common-ddl-script-v1-create-tables.sql` with only `user_id` and `notifications json DEFAULT NULL`, and no changeset anywhere adds a column for this config (I grepped the whole `resources` tree for `crm_deal_list_view`). Both `application-non-prd.yml` and `application-prd.yml` set `spring.jpa.hibernate.ddl-auto: validate`, so once the `crmDealListView` field exists on the `UserSettings` entity, Hibernate schema validation will fail on boot with a missing-column error — the app will not start. Fix: add a new DDL script (follow the existing naming, e.g. `common-ddl-script-v1-alter-table-user-settings-add-crm-deal-list-view.sql`) doing `ALTER TABLE user_settings ADD COLUMN crm_deal_list_view json DEFAULT NULL;` with a rollback comment, and register it as a new `- include:` entry at the end of `community/db/changelog/db.changelog.yml`. Annotate the entity field the same way `notifications` is: `@Column(name = "crm_deal_list_view", columnDefinition = "json") @Convert(converter = JsonTypeConverter.class)`. Note also that the enterprise profiles (`config/application-ep-*.yml`) set `ddl-auto: none` and `liquibase.enabled: false`, so the enterprise/multi-tenant deployment path needs its own migration for tenant schemas — confirm how that is applied before release.

backend/src/main/java/com/skapp/community/crmplanner/constant/DefaultCrmDealListViewConfig.java
```diff
@@ -0,0 +1,48 @@
...
+
+		ObjectNode config = objectMapper.createObjectNode();
+		config.set("fields", fields);
+		config.putNull("sort");
+		return config;
+	}
+
+	private static ObjectNode field(ObjectMapper objectMapper, String field, int width) {
+		ObjectNode node = objectMapper.createObjectNode();
+		node.put("field", field);
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The default config's property vocabulary does not match the frontend column contract it is supposed to seed, so the frontend cannot consume it without a translation layer. The shipped column definitions in `frontend/src/community/crm/components/organisms/DealsTable/DealsTable.tsx:59-126` and `frontend/src/community/crm/v2/components/organisms/DealsTableV2/DealsTableV2.tsx:68-138` use `id`/`field` values in camelCase (`dealName`, `value`, `stage`, `companyName`, `contactName`, `dealOwner`) and un-prefixed boolean keys (`visible`, `sortable`, `draggable`, `resizable`). This default emits SCREAMING_SNAKE field values (`DEAL_NAME`, `COMPANY_NAME`, ...) and `is`-prefixed keys (`isVisible`, `isSortable`, `isDraggable`, `isResizable`), plus `fieldId` and `isGroupable` which have no counterpart in the `Column` type at all — I grepped `frontend/src` and found zero table-column usages of `isVisible`, `isSortable`, `isGroupable`, or `fieldId`. Nothing in the frontend reads `/list-view-config` today either, so this contract is entirely unverified. Fix: agree the wire shape with the frontend first and emit exactly those keys/casing (or make the frontend map them explicitly), and drop `fieldId`/`isGroupable` until something consumes them. A typed response DTO instead of a hand-built `JsonNode` would make this contract checkable at compile time and give Swagger a real schema.

backend/src/main/java/com/skapp/community/crmplanner/constant/DefaultCrmDealListViewConfig.java
```diff
@@ -0,0 +1,48 @@
...
+		node.putNull("fieldId");
+		node.put("width", width);
+		node.put("isVisible", true);
+		node.put("isSortable", true);
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`isSortable: true` and `isDraggable: true` are wrong for every column in this default. (a) Backend: `CrmDealSort` (crmplanner/type/CrmDealSort.java) only supports NAME, CREATED_DATE, CLOSING_AT, AMOUNT, STAGE_TYPE and STAGE_ORDER — there is no sort key for COMPANY_NAME, CONTACT_NAME, PRIORITY or DEAL_OWNER, and `CrmDealFilterDto.sortKey` is a `CrmDealSort` enum, so a client that trusts this default and requests a sort on those columns cannot be served (an unmapped `sortKey` value fails enum binding). (b) Frontend: both deal tables set `sortable: false` and `draggable: false` on all six columns, and pin DEAL_NAME deliberately. So this default advertises interactions the system does not implement. Fix: set `isSortable` per column from what `CrmDealSort` actually supports (name → NAME, value → AMOUNT, stage → STAGE_ORDER; false for the rest) and set `isDraggable` to match the shipped table behaviour (currently `false` everywhere) rather than blanket `true`.

backend/src/main/java/com/skapp/community/crmplanner/constant/DefaultCrmDealListViewConfig.java
```diff
@@ -0,0 +1,48 @@
...
+		config.set("fields", fields);
+		config.putNull("sort");
+		return config;
+	}
+
+	private static ObjectNode field(ObjectMapper objectMapper, String field, int width) {
+		ObjectNode node = objectMapper.createObjectNode();
+		node.put("field", field);
+		node.putNull("fieldId");
+		node.put("width", width);
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The `width` values (400 / 140) are in the wrong unit for the consumer. The frontend computes `width: tableWidth * DEAL_TABLE_COLUMN_WIDTH_RATIO.<COL>` from `frontend/src/community/crm/constants/dealConstants.ts` (ratios 0.2337 / 0.1532) and uses `minWidth: 400` for the deal-name column and `minWidth: 140` for the rest — i.e. 400/140 are the *minWidths*, and actual widths are container-relative. Seeding absolute pixel widths as `width` will either be ignored or will pin columns to 400/140px regardless of viewport, breaking the responsive layout on first load for a user with no saved config. Fix: either emit the ratios the frontend already uses (and name the key accordingly), or emit `minWidth` and let the frontend keep deriving `width` from container size. Whichever is chosen, drive the numbers from a single shared source rather than duplicating the frontend constants here.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealListViewConfigServiceImpl.java
```diff
@@ -0,0 +1,55 @@
...
+		JsonNode config = (saved == null || saved.isNull()) ? DefaultCrmDealListViewConfig.build(objectMapper) : saved;
+		return new ResponseEntityDto(false, config);
+	}
+
+	@Override
+	@Transactional
+	public ResponseEntityDto updateListViewConfig(JsonNode config) {
+		User user = userService.getCurrentUser();
+		UserSettings settings = user.getSettings();
+		if (settings == null) {
```
ThinuwanW, line 64: 🤖 **Claude** · 🟡 **Suggestion**

The get-or-create-`UserSettings` block is now the fifth near-identical copy of this logic (`PeopleServiceImpl:1761-1769`, `PeopleServiceImpl:2572-2575`, `AuthServiceImpl:598`, `EpOrganizationServiceImpl:358`), and it also drops a step the established pattern performs. `PeopleServiceImpl.updateNotificationSettings` re-fetches the user with `userDao.findById(currentUser.getUserId())` and throws `ModuleException(CommonMessageConstant.COMMON_ERROR_USER_NOT_FOUND)` if absent before mutating and saving; this method mutates whatever `userService.getCurrentUser()` returned and saves it directly. Today that is safe because `@Transactional` makes the entity managed, but it is a silent divergence from the module's convention and it is fragile if the transaction boundary ever changes. Fix: extract a `getOrCreateSettings(User user)` helper onto `UserService` (or a small `UserSettingsService`) and reuse it from all five call sites, so the create-if-absent behaviour and the not-found error path live in one place.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealListViewConfigServiceImpl.java
```diff
@@ -0,0 +1,55 @@
...
+			settings = new UserSettings();
+			settings.setUser(user);
+			user.setSettings(settings);
+		}
+		settings.setCrmDealListView(config);
+		userDao.save(user);
+
+		return new ResponseEntityDto(false, config);
```
ThinuwanW, line 74: 🤖 **Claude** · 🟡 **Suggestion**

PUT echoes the request body rather than the effective persisted state, so PUT and GET can disagree about the same server state (guidelines checklist item 10 — symmetric endpoints must return the same response structure for the same state). Concretely: a client PUTs a body of literal `null`; `JsonTypeConverter.convertToDatabaseColumn` maps `NullNode` to SQL NULL, so the row is cleared, but this method returns `null` in `results[0]`. A subsequent GET hits the `saved == null || saved.isNull()` branch and returns the full default config instead. A client that writes the PUT response into its cache is now out of sync with the server until it refetches. Fix: after persisting, resolve the effective config through the same code path GET uses (i.e. fall back to `DefaultCrmDealListViewConfig.build(...)` when the stored value is null/`NullNode`) and return that, so both endpoints always return the same shape for the same state. Extracting a private `resolveConfig(UserSettings settings)` used by both methods also removes the duplicated null/`isNull` check.

backend/src/main/java/com/skapp/community/crmplanner/constant/DefaultCrmDealListViewConfig.java
```diff
@@ -0,0 +1,48 @@
+package com.skapp.community.crmplanner.constant;
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

Naming and structure deviate from the two existing sibling defaults in this package. The established pattern is a `Default*Values` enum in `crmplanner/type/` holding the data plus a `Default*Template` class in `crmplanner/constant/` that only maps it (`DefaultCrmDealStageValues` + `DefaultCrmDealStageTemplate`, `DefaultCrmTaskTypeValues` + `DefaultCrmTaskTypeTemplate`). This new class is named `...Config` rather than `...Template` and inlines the data as literals in the builder. Fix: rename to `DefaultCrmDealListViewTemplate` and move the per-column data (field key, width, flags) into a `DefaultCrmDealListViewValues` enum under `crmplanner/type/`, leaving this class as the JSON mapper — which also removes the seven repeated `field(...)` calls and gives the column vocabulary a single typed home.

backend/src/main/java/com/skapp/community/crmplanner/constant/DefaultCrmDealListViewConfig.java
```diff
@@ -0,0 +1,48 @@
...
+		fields.add(field(objectMapper, "VALUE", 140));
+		fields.add(field(objectMapper, "STAGE", 140));
+		fields.add(field(objectMapper, "COMPANY_NAME", 140));
+		fields.add(field(objectMapper, "CONTACT_NAME", 140));
+		fields.add(field(objectMapper, "PRIORITY", 140));
+		fields.add(field(objectMapper, "DEAL_OWNER", 140));
+
+		ObjectNode config = objectMapper.createObjectNode();
+		config.set("fields", fields);
+		config.putNull("sort");
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`config.putNull("sort")` reports "no active sort", but the deals list API is never unsorted: `CrmDealFilterDto` defaults to `sortKey = CrmDealSort.STAGE_ORDER` with `sortOrder = Sort.Direction.ASC`. So a client rendering this default shows no sort indicator while the server is in fact sorting by stage order. Fix: seed `sort` with the API's real default (`{"field": <the STAGE key>, "direction": "ASC"}` in whatever shape the frontend expects) so the config reflects actual server behaviour, or leave it null but drop the javadoc claim and document that null means "server default".

backend/src/main/java/com/skapp/community/crmplanner/constant/DefaultCrmDealListViewConfig.java
```diff
@@ -0,0 +1,48 @@
...
+ * config itself is an opaque JSON blob owned by the frontend; this default is the only
+ * shape the backend keeps, matching the columns the deal table ships with. {@code fields}
+ * order is the default column order; {@code sort} is null (no active sort).
+ */
+public final class DefaultCrmDealListViewConfig {
+
+	private DefaultCrmDealListViewConfig() {
+	}
+
+	public static JsonNode build(ObjectMapper objectMapper) {
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

The `ObjectMapper` parameter is not needed and forces the caller to inject a mapper purely to allocate nodes. `createObjectNode()`/`createArrayNode()` only delegate to the node factory, so `tools.jackson.databind.node.JsonNodeFactory.instance.objectNode()` / `.arrayNode()` gives an identical result with no dependency — which would also let `CrmDealListViewConfigServiceImpl` drop its mapper field entirely. Separately, this rebuilds the same constant tree on every request that has no saved config; if you keep the per-call build, that is fine (the returned `ObjectNode` is mutable so it must not be cached and shared), but a `deepCopy()` of a private static template would be cheaper and makes the immutability intent explicit.

backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmDealController.java
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Placement worth a second look: the deal read/write surface has been migrated to v2 (`CrmDealControllerV2` at `/v2/crm/deal`, `CrmDealServiceV2`, and the consumer `DealsTableV2`), yet these two new endpoints were added to the v1 controller. The table that would consume this config is the v2 table, so the v2 UI will have to reach back through the v1 client for it. There is precedent for both (the v2 frontend API file still points `DELETE_DEAL` and `CHECK_DEAL_NAME_EXISTS` at v1 paths), so this may be deliberate — but if the deal surface is being consolidated on v2, adding these to `CrmDealControllerV2` now avoids another endpoint to migrate later. Note also that `/v1/crm/deal/stage` is a separate `CrmDealStageController`, so a dedicated controller for this sub-resource would match module precedent too.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealListViewConfigServiceImpl.java
anusham-rc: Invalid. `UserSettings.crmDealListView` and the `user_settings.crm_deal_list_view` column come from #1830, which merges before this PR (noted in the description). So the field and schema already exist on the merge base â€” this PR just adds the config API on top. No change needed.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealListViewConfigServiceImpl.java
anusham-rc: Invalid. Same as the line 33 comment: the entity field and column (with rollback) land in #1830, which merges first. Once it's on the base, `ddl-auto: validate` has the column it needs. The enterprise tenant-schema migration lives in the migrations repo (rootcodelabs/skapp-migrations#214), linked from #1830. No change needed.

backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmDealController.java
anusham-rc: Invalid, by design. The config is an opaque JSON blob owned by the frontend â€” the backend stores and returns it as-is and doesn't model its shape, so there's no `*RequestDto`; that's why it takes a raw `JsonNode`. The notification-settings DTO is different because those booleans are server-controlled. Only the logged-in sales rep can overwrite their own config, so a bad payload only affects that user and is fixed by the next PUT. Reset-to-default is a client concern (PUT the default). The one fair point â€” an upper size limit â€” is minor hardening we can add separately.

backend/src/main/java/com/skapp/community/crmplanner/constant/DefaultCrmDealListViewConfig.java
anusham-rc: Invalid. The config is owned by the frontend, and that frontend doesn't exist yet â€” nothing reads `/list-view-config` today. The current `DealsTable`/`DealsTableV2` use the `skapp-ui` `Column` type and aren't the consumer of this endpoint. The exact keys/casing (and whether `fieldId`/`isGroupable` stay) get finalised with the frontend later. So this isn't a mismatch to fix â€” it's a contract not yet agreed.

backend/src/main/java/com/skapp/community/crmplanner/constant/DefaultCrmDealListViewConfig.java
anusham-rc: Invalid. `isSortable`/`isDraggable` are frontend UI hints, not tied to the backend `CrmDealSort` enum. Nothing consumes the config yet, so no client is handed an unusable sort key â€” server-side sorting still goes through `CrmDealFilterDto.sortKey` independently. The comparison is against the old `skapp-ui` tables, which aren't this endpoint's consumer. Per-column defaults get agreed with the frontend later; nothing to fix here.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealListViewConfigServiceImpl.java
anusham-rc: Fixed. Added boundary logs to both methods, matching the module convention: `getListViewConfig: execution started`/`ended` and `updateListViewConfig: execution started`/`ended successfully`. No payload is logged, just the method boundaries.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealListViewConfigServiceImpl.java
anusham-rc: Fixed (part 1). Switched the injected field to `JsonMapper`, matching the codebase convention (e.g. `PeopleServiceImpl`). It resolves to the same auto-configured bean, so behaviour is unchanged.

backend/src/main/java/com/skapp/community/crmplanner/constant/DefaultCrmDealListViewConfig.java
anusham-rc: Fixed (javadoc). The class doc no longer claims the default matches the shipped deal-table columns; it now says this is just the initial shape the backend seeds, finalised with the frontend later. The PRIORITY column and the `isSortable`/`isDraggable` defaults stay as-is â€” nothing consumes the config yet, so they're part of the not-yet-agreed contract, not a mismatch. PRIORITY already exists elsewhere in the CRM and is seeded ahead of the consumer on purpose.

backend/src/main/java/com/skapp/community/crmplanner/constant/DefaultCrmDealListViewConfig.java
anusham-rc: Fixed. Added a `CrmDealListViewField` enum in `crmplanner/type/` as the single home for the column vocabulary, plus `DefaultCrmDealListViewValues` holding each field + default width in a static `DEFAULT_FIELDS` list â€” mirroring `DefaultCrmDealStageValues`/`CrmDealStageName`. The default is now built by iterating that list instead of string literals, so there's one source of truth and future validation can reject unknown `field` values.

backend/src/main/java/com/skapp/community/crmplanner/constant/DefaultCrmDealListViewConfig.java
anusham-rc: Fixed. The builder is now `DefaultCrmDealListViewTemplate` with `@UtilityClass` (matching `DefaultCrmDealStageTemplate`), and the per-column data moved into the enum-backed `DefaultCrmDealListViewValues` under `crmplanner/type/`. This class is now just the JSON mapper. Renamed from `...Config` to `...Template` to read accurately.

backend/src/main/java/com/skapp/community/crmplanner/constant/DefaultCrmDealListViewConfig.java
anusham-rc: Fixed. Restructured to match the sibling pattern: `DefaultCrmDealListViewValues` (+ `CrmDealListViewField` enum) in `crmplanner/type/` holds the per-column data, and `DefaultCrmDealListViewTemplate` (`@UtilityClass`) in `crmplanner/constant/` just maps it to JSON â€” same shape as `DefaultCrmDealStageValues` + `DefaultCrmDealStageTemplate`. This also removes the seven repeated `field(...)` calls.

backend/src/main/java/com/skapp/community/crmplanner/constant/DefaultCrmDealListViewConfig.java
anusham-rc: Fixed (idea 1). `build()` now allocates nodes via `JsonNodeFactory.instance` and takes no mapper argument, which also let `CrmDealListViewConfigServiceImpl` drop its `JsonMapper` field. Skipped idea 2 (static template + `deepCopy`): the returned `ObjectNode` is mutable so it can't be shared safely, and the tree is tiny and only built when a user has no saved config â€” caching would add a mutability risk for negligible gain.

backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmDealController.java
anusham-rc: Invalid (out of scope). This is pre-existing `GlobalExceptionHandler` behaviour affecting every `@RequestBody` endpoint, not something this PR introduces. Adding an `HttpMessageNotReadableException` handler is an app-wide change that should be its own PR with its own tests. No change here.

backend/src/main/java/com/skapp/community/crmplanner/constant/DefaultCrmDealListViewConfig.java
anusham-rc: Invalid (premature). Nothing consumes the config yet, so what `width` means is part of the not-yet-agreed contract â€” these numbers are just seed defaults, not a claim about pixels vs ratios vs minWidths. The current tables aren't this endpoint's consumer. The frontend will settle this when that side is built. No change now.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealListViewConfigServiceImpl.java
anusham-rc: Invalid (minor edge case). The asymmetry only happens for a literal-`null` PUT body, which isn't a real client flow â€” the frontend PUTs its actual config, so GET and PUT agree. Reset-to-default is a client concern (PUT the default), not a server path. Adding a resolve-through-GET step just for the null case is extra code for something that shouldn't occur; PUT keeps echoing the persisted body. No change.

backend/src/main/java/com/skapp/community/crmplanner/constant/DefaultCrmDealListViewConfig.java
anusham-rc: Invalid (premature). `sort` here is frontend list-view UI state, not a mirror of the server `CrmDealFilterDto` default â€” `null` means "no user-applied sort", and the frontend can fall back to the server default when rendering. Nothing consumes the config yet, so the exact `sort` shape is still to be agreed. The javadoc already documents `sort` is null (no active sort). No change.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealListViewConfigServiceImpl.java
anusham-rc: Invalid (intentional â€” follows the existing precedent). This mirrors how per-user notification settings already work: `PeopleServiceImpl` injects the common `UserDao` directly and stores its config as a JSON blob on the shared `user_settings` table (`UserSettings.notifications`). There's no `UserSettingsService` in the codebase today â€” feature modules reach `UserSettings` via `UserDao`, so this follows that convention rather than skipping a layer. The suggested alternatives (a common `UserSettingsService`, or a generic per-user preference table) would be a new pattern; if we want that, it should be a deliberate, module-wide decision rather than part of this feature PR. The `crm_deal_list_view` column sits alongside `notifications` for the same reason.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealListViewConfigServiceImpl.java
anusham-rc: Partially fixed. `updateListViewConfig` now re-fetches the user with `userDao.findById(...)` and throws `ModuleException(COMMON_ERROR_USER_NOT_FOUND)` before saving, matching `updateNotificationSettings`. Left the shared `getOrCreateSettings(User)` helper out of this PR â€” it touches four other call sites across the people/auth/enterprise modules, so that's a cross-cutting refactor for its own PR.

backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmDealController.java
anusham-rc: Invalid (deliberate placement). These endpoints sit on the v1 `CrmDealController` on purpose, mirroring how per-user settings live on their domain controller (e.g. notification settings on `PeopleController`). Mixing v2 UI with v1 endpoints is already the norm here â€” the v2 frontend API file already points `DELETE_DEAL` and `CHECK_DEAL_NAME_EXISTS` at v1 paths â€” so a v2 table using a v1 config endpoint is consistent. The logic lives in a separate `CrmDealListViewConfigService` (not `CrmDealService`) because the enterprise `EpCrmDealServiceImpl` subclass constructor can't take new deps. These can move if the deal surface fully consolidates on v2 later, but that's a broader migration. No change.

anusham-rc: Fixed. Added four integration tests to `CrmDealControllerIntegrationTest`, driving the endpoints over HTTP:
- `getListViewConfig_NoSavedConfig_ReturnsDefault` â€” GET returns the default (7 fields, `DEAL_NAME` first, `sort` null) when the user has no saved config.
- `updateThenGetListViewConfig_RoundTripsSavedConfig` â€” PUT then GET round-trips the saved config (and PUT echoes it back).
- `updateListViewConfig_DoesNotAffectAnotherUser` â€” user1 saving leaves user2 on the default.
- `getListViewConfig_WithoutCrmRole_ReturnsForbidden` â€” a non-CRM role gets 403.

These run green only once #1830 is on the base, since the persistence field (`UserSettings.crmDealListView`) lands there â€” same dependency as the service code.

erandi-rootcode: Akila mentioned to follow this ordering 

1. operation
2. preauthorize
3. postmapping

backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmDealController.java
```diff
@@ -81,4 +86,22 @@ public ResponseEntity<ResponseEntityDto> deleteDeal(@PathVariable Long id) {
...
+	@GetMapping("/list-view-config")
+	@PreAuthorize("hasAnyRole('ROLE_CRM_SALES_REPRESENTATIVE')")
+	public ResponseEntity<ResponseEntityDto> getListViewConfig() {
+		ResponseEntityDto response = crmDealListViewConfigService.getListViewConfig();
+		return new ResponseEntity<>(response, HttpStatus.OK);
+	}
+
+	@Operation(summary = "Update the current user's deal list-view config",
+			description = "Persists the deal table config JSON sent by the client as-is; does not affect other users.")
+	@PutMapping("/list-view-config")
```
erandi-rootcode: Here too

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealListViewConfigServiceImpl.java
```diff
@@ -0,0 +1,69 @@
...
+@Service
+@RequiredArgsConstructor
+public class CrmDealListViewConfigServiceImpl implements CrmDealListViewConfigService {
+
+	private final UserService userService;
+
+	private final UserDao userDao;
+
+	@Override
+	@Transactional(readOnly = true)
```
erandi-rootcode, line 33: Since this is a Get so it is not required. Verify before removal. we have a separate pool for read/ write database connections.

backend/src/main/java/com/skapp/community/crmplanner/type/DefaultCrmDealListViewValues.java
```diff
@@ -0,0 +1,25 @@
...
+@Getter
+@RequiredArgsConstructor
+public class DefaultCrmDealListViewValues {
+
+	private final CrmDealListViewField field;
+
+	private final int width;
+
+	public static final List<DefaultCrmDealListViewValues> DEFAULT_FIELDS = List.of(
+			new DefaultCrmDealListViewValues(CrmDealListViewField.DEAL_NAME, 400),
```
erandi-rootcode: should we move this values to constants?

erandi-rootcode (changes requested):

backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmDealController.java
anusham-rc: Fixed. Reordered to @Operation → @PreAuthorize → @Mapping.

backend/src/main/java/com/skapp/community/crmplanner/controller/v1/CrmDealController.java
anusham-rc: Fixed. Reordered to @Operation → @PreAuthorize → @Mapping.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealListViewConfigServiceImpl.java
anusham-rc: Verified — keeping it. `@Transactional(readOnly = true)` is the convention for every read method in the CRM module, and with the separate read/write pool the readOnly flag is what routes the query to the read connection. Removing it would diverge and lose that routing. No change.

backend/src/main/java/com/skapp/community/crmplanner/type/DefaultCrmDealListViewValues.java
anusham-rc: Invalid. This follows the sibling convention — `DefaultCrmDealStageValues` and `DefaultCrmTaskTypeValues` also live in `crmplanner/type/`, with only their `*Template` mappers in `constant/`. Moving it would break that pattern. No change.

erandi-rootcode (dismissed):

erandi-rootcode (dismissed):

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealListViewConfigServiceImpl.java
```diff
@@ -0,0 +1,77 @@
...
+	public ResponseEntityDto getListViewConfig() {
+		log.info("getListViewConfig: execution started");
+
+		User user = userService.getCurrentUser();
+		UserSettings settings = user.getSettings();
+		JsonNode saved = settings != null ? settings.getCrmDealListView() : null;
+
+		CrmDealListViewConfigDto config = (saved == null || saved.isNull()) ? DefaultCrmDealListViewTemplate.build()
```
akilarootcode, line 42: use instance variables for OBJECT_MAPPER

akilarootcode (changes requested):

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmDealListViewConfigServiceImpl.java
anusham-rc: handled separately

erandi-rootcode (approved):

akilarootcode (approved):

ThinuwanW (approved):

PR #1832, opened 2026-08-24
Refactor/crm contact section

dulsara-manakal:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

The backend half of this PR (flattening CrmContactResponseDtoV2/CrmContactMetricsResponseDtoV2 to scalar companyId/ownerId, with matching Criteria projections, mapper annotations and integration-test updates) is coherent and correct. The frontend half, however, is built against task/metric response shapes the backend does not actually produce: `GET /v2/crm/task` returns `{tasks: [...]}` rather than a page, so `useGetTasksInfinite` will crash the (already live) Company side panel; `CrmTaskResponseDtoV2` is still nested (`type`/`owner`/`company`/`contact`/`deal`), so task linking, cache invalidation, owner avatars and type icons all silently no-op; and `/v1/crm/contact/{id}/metrics` doesn't return three of the four fields the new contact metric cards read. There are also i18n key mismatches on the add/edit contact toasts, a company-unlink payload regression versus v1 (`undefined` vs `null` for a `JsonNullable` field), an owner field that sales representatives can never fill, and a sizeable deletion of existing companyUtil tests.

Found **36** new issue(s): 🔴 13 important, 🟡 15 suggestion(s), 🟣 8 nit(s)

frontend/src/community/crm/v2/api/TaskApi.ts
```diff
@@ -0,0 +1,124 @@
...
+
+interface UpdateTaskVariables {
+  id: number;
+  payload: CrmTaskEntity;
+}
+
+const fetchTasks = async (
+  params: CrmTaskCompletedFilterRequest
+): Promise<CrmTaskCompletedListResponse> => {
+  const response = await authFetchV2.get(crmTaskEndpoints.GET_TASKS, {
```
ThinuwanW, line 33: 🤖 **Claude** · 🔴 **Important**

`GET /v2/crm/task` is not paginated and does not return `items`. `CrmTaskControllerV2.getTasks` takes `CrmTaskFilterDto` (no page/size) and `CrmTaskServiceImplV2.getTasks` returns `CrmTaskListResponseDtoV2`, i.e. `{ tasks: [...] }`. So `fetchTasks` resolves to `{tasks: [...]}`, `lastPage.currentPage` is `undefined` (getNextPageParam computes `NaN`), and in CompanySidePanel/ContactSidePanel `fetchedTasks.pages.flatMap((page) => page.items)` yields `[undefined]`, which `normalizeTasks` then dereferences: `items.forEach((task) => { if (task.id !== undefined) ...})` throws `Cannot read properties of undefined (reading 'id')`. Since `companies.tsx` renders `CompanySidePanelV2` today, opening a company side panel will throw. Either point `GET_TASKS` at the paginated `/v2/crm/task/related` (contactId/dealId) and add a company-scoped paginated endpoint, or extend the backend `getTasks` to accept page/size and return a `PageDto`. As a defensive measure also make `normalizeTasks`/the flatMap tolerate a missing `items` array.

frontend/src/community/crm/v2/api/TaskApi.ts
```diff
@@ -0,0 +1,124 @@
...
+export const useUpdateTask = (
+  onSuccess: (task: CrmTaskEntity) => void,
+  onError: (error: AxiosError) => void
+): UseMutationResult<CrmTaskEntity, AxiosError, UpdateTaskVariables> => {
+  const queryClient = useQueryClient();
+
+  return useMutation({
+    mutationFn: updateTask,
+    onSuccess: (updatedTask) => {
+      if (updatedTask.companyId !== undefined) {
```
ThinuwanW, line 115: 🤖 **Claude** · 🔴 **Important**

The v2 task response is still nested while the frontend types assume it is flat. `CrmTaskResponseDtoV2` exposes `type`, `owner`, `company`, `contact`, `deal` as objects (see `CrmTaskRepositoryImpl.buildTaskProjectionSelection` and the DTO), but `CrmTaskEntity` declares `typeId`/`ownerId`/`companyId`/`contactId`/`dealId`. Consequences: (1) `updatedTask.companyId` and `createdTask.companyId/contactId` are always `undefined`, so neither `useCreateTask` nor `useUpdateTask` ever invalidates company/contact metrics; (2) `AddTaskModalContent`'s `linkTaskToRelatedEntities(createdTask, ...)` is a no-op, so a newly created task is never appended to `contact.taskIds`/`company.taskIds` and never shows up in the side-panel list; (3) `TaskRowMeta` never renders the owner avatar (`task.ownerId` undefined) and `TaskRowContent` always falls back to the default icon (`task.typeId` undefined). Either flatten `CrmTaskResponseDtoV2` the same way this PR flattened `CrmContactResponseDtoV2`, or read the nested fields (`createdTask.contact?.id`, `createdTask.owner?.employeeId`, `createdTask.type?.id`) in the frontend. Requests are already flat (`CrmTaskCreateRequestDto`), so only responses are affected.

frontend/src/community/crm/v2/components/molecules/AddContactModalContent/AddContactModalContent.tsx
```diff
@@ -0,0 +1,110 @@
...
+    if (createdContact.id !== undefined) {
+      setContacts({ ...contacts, [createdContact.id]: createdContact });
+      setContactIds([createdContact.id, ...contactIds]);
+    }
+
+    handleCloseModal();
+    setToastMessage({
+      open: true,
+      toastType: ToastType.SUCCESS,
+      title: translateText(["toastMessages", "successTitle"]),
```
ThinuwanW, line 76: 🤖 **Claude** · 🔴 **Important**

Translation key mismatch — these toasts will render raw key strings. The translator is scoped to `crmModule.contacts.addContactModal`, but `crmModule.json` defines the toast copy under `contactToastMessages`, not `toastMessages` (see crmModule.json lines 68-73 for add and 139-144 for edit; only `deleteContactModal` uses `toastMessages`). The same mismatch exists at EditContactModalContent.tsx:70/71 and :80/81, and at AddContactModalContent.tsx:86/87. Either rename the JSON keys to `toastMessages` (consistent with `deleteContactModal`) or change the four call sites to `translateText(["contactToastMessages", ...])`.

frontend/src/community/crm/v2/utils/contactUtil.ts
```diff
@@ -26,21 +57,208 @@ export const getContactDisplayName = (
...
+
+  if (currentValues.email !== initialValues.email) {
+    changedFields.email = currentValues.email;
+  }
+
+  if (currentValues.contactNumber !== initialValues.contactNumber) {
+    changedFields.contactNumber = currentValues.contactNumber;
+  }
+
+  if (currentValues.companyId !== initialValues.companyId) {
```
ThinuwanW, line 208: 🤖 **Claude** · 🔴 **Important**

Clearing a contact's company silently does not persist. `getChangedContactFields` sets `changedFields.companyId = currentValues.companyId`, which is `undefined` when the user clicks the clear button in ContactModalForm (`handleClearCompany` → `setFieldValue("companyId", undefined)`). `JSON.stringify` drops `undefined` keys, so the PATCH body omits `companyId` entirely, and `CrmContactEditRequestDto.companyId` is a `JsonNullable<Long>` defaulting to `JsonNullable.undefined()` — `applyContactEdit` therefore skips the `isPresent()` branch and leaves the company attached. Note `Object.keys(changedFields).length` is still 1, so the request is sent but is a no-op. The v1 implementation avoided this by using `null` (`companyId: selectedContact?.company?.id ?? null`). Fix: emit `changedFields.companyId = currentValues.companyId ?? null` (and initialise `getContactFormInitialValues` with `null` instead of `undefined`) so the explicit null reaches `JsonNullable`.

frontend/src/community/crm/v2/components/molecules/AddTaskModalContent/AddTaskModalContent.tsx
```diff
@@ -0,0 +1,108 @@
...
+      setTasks: store.setTasks,
+      setCompanies: store.setCompanies,
+      setContacts: store.setContacts,
+      setDeals: store.setDeals,
+      setIsTaskModalOpen: store.setIsTaskModalOpen
+    }))
+  );
+
+  const formik = useFormik<CrmTaskEntity>({
+    initialValues: getTaskFormInitialValues(selectedContactId),
```
ThinuwanW, line 51: 🤖 **Claude** · 🔴 **Important**

The add-task form never defaults `ownerId`, which makes it unusable for CRM sales representatives. `getTaskFormInitialValues` only sets `name`, `priority`, `notes` and (optionally) `contactId`, while `getTaskValidationSchema` marks `ownerId` as required. In TaskModalForm the owner dropdown is fed by `useGetOwnerLookup(ownerFilters, Boolean(isCrmSalesManager) && debouncedOwnerSearchText.length > 0)` — for a non-manager `isCrmSalesManager` is false, so the lookup is permanently disabled, `ownerDropdownItems` stays empty, `selectedOwner` is never resolved, and the form can never satisfy validation. The v1 `AddTaskModalContent` avoided this by pre-filling the owner from `useGetUserPersonalDetails()` (and used `enableReinitialize: true` so the value lands once the user loads). Do the same here: pass the current user's `employeeId` into `getTaskFormInitialValues` and add `enableReinitialize: true` (the sibling `AddContactModalContent` already does both).

frontend/src/community/crm/v2/components/molecules/TaskRow/TaskRow.tsx
```diff
@@ -0,0 +1,92 @@
...
+      open: true,
+      toastType: ToastType.ERROR,
+      title: translateText(["toggleErrorTitle"]),
+      description: translateText(["toggleErrorDescription"])
+    });
+  };
+
+  const { mutate: updateCompletion } = useUpdateTask(
+    (updatedTask) => setTasks(updateTask(tasks, taskId, updatedTask)),
+    () => handleToggleError(isCompleted)
```
ThinuwanW, line 55: 🤖 **Claude** · 🔴 **Important**

The optimistic-update rollback is a no-op because of a stale-value capture. `handleToggleChange` first writes the new value into the store, which re-renders the row so `isCompleted` (line 39) now equals the *new* value. TanStack Query v5 refreshes the observer's options on every render, so when the PATCH fails it invokes the latest `() => handleToggleError(isCompleted)` — which restores the already-optimistic value. The checkbox therefore stays in the failed state while the error toast claims the update failed. Pass the previous value explicitly instead, e.g. `updateCompletion({ id: taskId, payload: { isCompleted: nextIsCompleted } }, { onError: () => handleToggleError(!nextIsCompleted) })`, or capture it in `onMutate` context.

frontend/src/community/crm/v2/utils/contactUtil.ts
```diff
@@ -26,21 +57,208 @@ export const getContactDisplayName = (
...
+      variant: CrmMetricLabelThemeEnum.RED
+    };
+  }
+
+  return [
+    openTasks,
+    {
+      id: "activeDealsCount",
+      title: translateText(["metrics", "activeDeals"]),
+      amount: metrics?.activeDealsCount
```
ThinuwanW, line 120: 🤖 **Claude** · 🔴 **Important**

Three of the four contact metric cards will always render empty. `useGetContactMetrics` hits `GET /v1/crm/contact/{id}/metrics`, which returns `CrmContactMetrics` = `{closedDealValue, closedDealCount, openTasksCount, overdueTasksCount}` (see `CrmContactServiceImpl.getContactMetricsById` / `com.skapp.community.crmplanner.type.CrmContactMetrics`). `getContactMetricItems` reads `metrics?.activeDealsCount`, `metrics?.totalRevenue` and `metrics?.pipelineRevenue`, none of which exist in that payload — `SidePanelMetricCards` will show `-` for the two currency cards and a blank value for "No. of active deals". The frontend `CrmContactMetrics` interface declares every field optional, so TypeScript can't catch this. Either add these aggregates to the backend metrics type or build the cards from the fields the API actually returns.

frontend/src/community/crm/v2/components/molecules/ContactModalForm/ContactModalForm.tsx
```diff
@@ -0,0 +1,274 @@
...
+    content: option.isSuggested ? (
+      <SuggestedBadge label={translateText(["labels", "suggested"])}>
+        {option.name}
+      </SuggestedBadge>
+    ) : (
+      option.name
+    )
+  }));
+
+  const selectedCompanyName = getCompanyNameById(companies, values.companyId);
```
ThinuwanW, line 125: 🤖 **Claude** · 🔴 **Important**

Selecting a searched company leaves the field blank. `handleCompanySelect` only stores the id (`setFieldValue("companyId", Number(item.id))`), and the read-only display resolves the name via `getCompanyNameById(companies, values.companyId)` from the Zustand store — but this component never merges `companyLookupData.items` into the store. Only `ContactTable` does that, and only for the unfiltered first `DEFAULT_LOOKUP_PAGE_SIZE` page plus the companies referenced by the currently visible contacts. Any company found via the search box that isn't already cached will render an empty read-only input after selection. Add a `useEffect` that merges `companyLookupData?.items` (and `domainSearchData?.companies`) into the store via `mergeCompanies`, the way `ContactTable` does, or keep the selected `CrmCompanyEntity` in local state and use its `name` directly.

frontend/src/community/crm/v2/components/molecules/SidePanelTasksSection/SidePanelTasksSection.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const handleAddTask = () => {
+    guardCrmCreate(CrmLimitResource.TASKS, () => {
+      setTaskModalType(CrmModalTypes.ADD_TASK_MODAL);
+      setIsTaskModalOpen(true);
+    });
+  };
+
+  if (taskIds !== undefined && taskIds.length > 0) {
+    return (
+      <div ref={loadingRef}>
```
ThinuwanW, line 59: 🤖 **Claude** · 🟡 **Suggestion**

`loadingRef` is attached to the wrapper around the whole task list rather than to a zero-height sentinel after it. `useInfiniteScroll` uses `threshold: 1`, which requires the observed element to be *fully* visible: a list taller than the scroll container will never trigger a fetch, and a short list is fully visible immediately and will chain-fetch every page on mount. Follow the pattern in the sibling `SidePanelDealSection`, which renders the list normally and places `<div ref={loadingRef} />` after it.

frontend/src/community/crm/v2/utils/dealUtil.ts
```diff
@@ -55,3 +60,14 @@ export const resolveDeals = (
   dealIds
     .map((id) => deals[id])
     .filter((deal): deal is CrmDealEntity => Boolean(deal));
+
+export const getDealNameById = (deals: CrmDealRecord, dealId?: number) => {
+  if (dealId !== undefined) {
+    return deals[dealId].name;
```
ThinuwanW, line 66: 🤖 **Claude** · 🟡 **Suggestion**

Unguarded record lookups that will throw on a cache miss, inconsistent with their siblings. `getDealNameById` does `deals[dealId].name` and `getTaskTypeName` (taskUtil.ts:83) does `taskTypes[typeId].name`, while `getCompanyNameById` and `getContactNameById` correctly use `?.`/undefined-safe access. `taskTypes` is only ever populated by `useInitializeCrmData`, which is gated behind `isCrmDataInitialized` and never refetched, so a type added mid-session crashes `TaskRowContent`. Related: `TaskRowMeta.tsx:42` passes `owners[task.ownerId]` (typed as non-optional by `Record<number, T>` but potentially undefined at runtime) straight into `OwnerAvatar`, which dereferences `owner.authPic`/`owner.firstName`; and `SidePanelDealSection` passes `deals[dealId]` into `DealAccordionItemHeader` unchecked. Add `?.` and null-guard the render sites.

frontend/src/community/crm/v2/api/TaskApi.ts
```diff
@@ -0,0 +1,124 @@
...
+
+export const useUpdateTask = (
+  onSuccess: (task: CrmTaskEntity) => void,
+  onError: (error: AxiosError) => void
+): UseMutationResult<CrmTaskEntity, AxiosError, UpdateTaskVariables> => {
+  const queryClient = useQueryClient();
+
+  return useMutation({
+    mutationFn: updateTask,
+    onSuccess: (updatedTask) => {
```
ThinuwanW, line 114: 🤖 **Claude** · 🟡 **Suggestion**

Asymmetric cache invalidation across the new task/contact mutations. `useCreateTask` invalidates the CRM limitation key, company metrics, contact metrics and `crmContactQueryKeys.LISTS`, but `useUpdateTask` invalidates only company metrics — toggling a task complete changes the contact's `openTasksCount`/`overdueTasksCount`, so the contact side panel and contact table metrics go stale. Neither mutation invalidates `crmTaskQueryKeys.LIST`, so the task list itself is only kept in sync by manual store writes. Similarly `useEditContact` (ContactApi.ts) invalidates nothing at all, and `useDeleteContact` invalidates only the limitation key — neither touches `crmContactQueryKeys.DETAIL`/`LIST`/`METRICS`. Please make the invalidation sets symmetric with `useCreateTask`.

frontend/src/community/crm/v2/utils/__test__/companyUtil.test.ts
```diff
@@ -1,148 +1,10 @@
-import { CrmIndustryEnum } from "~community/crm/v2/enums/common";
-import {
-  CrmCompanyEntity,
-  CrmCompanyRecord
-} from "~community/crm/v2/types/CrmCommonTypes";
+import { CrmCompanyEntity } from "~community/crm/v2/types/CrmCommonTypes";
```
ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

This PR deletes ~140 lines of passing tests for functions it does not change: the entire `normalizeCompanies`, `updateCompany`, `getSelectedCompany`, `getCompanyFormInitialValues`, `getTrimmedCompanyValues` and `getCompanyMetricItems` suites, plus the immutability assertion from `removeCompany`. `commonUtil.test.ts` similarly drops the `formatTableValue(23) === "23"` and `formatTableValue(undefined, "+")` cases while merging others. Those helpers are still exported and still used (e.g. `updateCompany` is central to `CompanySidePanel`). If the deletion is intentional please explain it in the PR description; otherwise restore the suites.

frontend/src/community/crm/v2/components/organisms/ContactTable/ContactTable.tsx
```diff
@@ -0,0 +1,315 @@
+import {
```
ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

This PR adds roughly 25 new components (ContactTable, ContactSidePanel, ContactModalController, ContactModalForm, TaskModalForm, TaskRow + 4 sub-components, SidePanelDealSection + 4 sub-components, SidePanelTasksSection/List, SidePanelContactsSection, SelectableSearchField, SelectedOwnerField, EditableContactOwnerField, OwnerAvatar, OwnerDropdownItem, SuggestedBadge, …) with no `.test.tsx` files, and the only new util test (`taskUtil.test.ts`) covers `linkTaskToRelatedEntities` alone — `getDueDateStatus`, `normalizeTasks`, `getPriorityConfig`, `parseDueDate`, `getContactMetricItems`, `getChangedContactFields`, `getCompanyOptions` and `getEmailDomain` are all untested. Per the review guidelines new components need corresponding tests with the `MockTheme` wrapper and `@testing-library/react` role-based queries. Given the contract mismatches noted above, a couple of tests asserting the parsed API shape would have caught them.

backend/src/main/java/com/skapp/community/crmplanner/repository/impl/CrmContactRepositoryImpl.java
```diff
@@ -115,13 +113,8 @@ public Page<CrmContactMetricsResponseDtoV2> getContactMetricsV2(CrmContactMetric
...
 				contact.get(CrmContact_.name), contact.get(CrmContact_.email), contact.get(CrmContact_.contactNumber),
 				contact.get(CrmContact_.lastContactAt), contact.get(Auditable_.lastModifiedDate),
-				cb.construct(CrmCompanyResponseDto.class, company.get(CrmCompany_.id), company.get(CrmCompany_.name),
-						company.get(CrmCompany_.industry), company.get(CrmCompany_.website),
-						company.get(CrmCompany_.address), company.get(CrmCompany_.contactNumber)),
-				cb.construct(CrmOwnerResponseDto.class, owner.get(Employee_.employeeId), owner.get(Employee_.firstName),
-						owner.get(Employee_.lastName), owner.get(Employee_.authPic)),
-				cb.construct(CrmContactMetrics.class, closedValueSub.cast(String.class), closedCountSub, openTaskSub,
-						overdueTaskSub)));
+				company.get(CrmCompany_.id), owner.get(Employee_.employeeId), cb.construct(CrmContactMetrics.class,
```
ThinuwanW, line 117: 🤖 **Claude** · 🟡 **Suggestion**

Soft-deleted companies are no longer masked consistently in the v2 contact responses. This projection now emits `company.get(CrmCompany_.id)` unconditionally, whereas `findContactsForLookupV2` a few methods below (line ~269) explicitly nulls the id for deleted companies via `selectCase().when(isTrue(company.isDeleted), nullLiteral(Long.class))`, and `CrmTaskRepositoryImpl.buildTaskProjectionSelection` joins with `contactCompany.on(cb.isFalse(isDeleted))`. The same gap exists for `CrmMapperV2.crmContactToCrmContactResponseDtoV2` (used by getContactById/createContact/editContact), while the v1 equivalents all route through `CrmUtil.toContactDetailDto`/`toContactListItemDto`, which null the company. Now that the client resolves `companyId` → name from its own store this mostly renders as `-` rather than a leak, but the three v2 contact paths should mask deleted companies the same way for consistency.

frontend/src/community/crm/v2/api/utils/QueryKeys.ts
```diff
@@ -19,7 +45,13 @@ export const crmCompanyQueryKeys = {
...
+  NAME_EXISTS: (name: string) => [CRM_COMPANIES, "name-exists", name],
+  LOOKUP: (params: CrmCompanyFilterRequest) => [
+    CRM_COMPANIES,
+    "lookup",
+    params
+  ],
+  DOMAIN_SEARCH: (domain: string) => [CRM_COMPANIES, "domain-search", domain]
 };
 
 export const crmLookupQueryKeys = {
```
ThinuwanW, line 57: 🤖 **Claude** · 🟡 **Suggestion**

Several dead/duplicated symbols were introduced or left behind. `crmLookupQueryKeys` (this line) has no remaining references now that ContactApi migrated to `crmContactQueryKeys`. `crmTaskQueryKeys.ALL` is never used. `crmDealEndpoints.GET_DEALS` (ApiEndpoints.ts:28) duplicates `crmDealEndpointsV2.GET_DEALS` and is unreferenced. `toDealsRecord` now exists in both `commonUtil.ts:65` (new, unused) and `dealUtil.ts:8` (the one actually imported by TaskModalForm). `DEFAULT_LOOKUP_PAGE_SIZE` is duplicated between `~community/crm/constants/commonConstants` and the new `~community/crm/v2/constants/commonConstants`, and the two are imported inconsistently across the PR (OwnerPopupSearch/DealPropertiesSidebar use the v1 path, everything new uses the v2 path). `TASK_TYPE_ICON_SIZE` (taskConstants.ts:14) is exported but `getTaskTypeIcon` hardcodes `size = 20`. The unused types `CrmContactLookupResponse` and `CrmCompanyDomainSearchFilterRequest` can also go.

frontend/pages/community/crm/contacts.tsx
```diff
@@ -11,10 +12,21 @@ import { ContactTable } from "~community/crm/components/organisms/ContactTable/C
...
+import SidePanelWrapperV2 from "~community/crm/v2/components/templates/SidePanelWrapper/SidePanelWrapper";
+import { useInitializeCrmData } from "~community/crm/v2/hooks/useInitializeCrmData";
+import { useCrmStoreV2 } from "~community/crm/v2/store/store";
+import { CrmModalTypes as CrmModalTypesV2 } from "~community/crm/v2/types/CrmTypes";
 import useCrmLimitGuard from "~enterprise/crm/hooks/useCrmLimitGuard";
 import { CrmLimitResource } from "~enterprise/crm/types/CrmLimitTypes";
 
-const Contacts: NextPage = () => {
+// Flip to true to serve the CRM Contacts page from the normalized v2 store surface.
+const isCrmContactsV2 = false;
```
ThinuwanW, line 27: 🤖 **Claude** · 🟣 **Nit**

A hardcoded `const isCrmContactsV2 = false;` ships the entire ContactsV2 tree (table, side panel, modal controllers and their transitive imports) as unreachable code that neither TypeScript nor tree-shaking will remove, and it can only be toggled by editing source. Prefer the pattern used elsewhere for mode switching (an env var / feature-flag helper) so the branch can be flipped per-environment and the dead branch is at least intentional.

frontend/src/community/crm/v2/components/organisms/ContactSidePanel/ContactSidePanel.tsx
```diff
@@ -0,0 +1,302 @@
...
+                taskIds={contact.taskIds}
+                hasNextPage={hasNextTasksPage}
+                isFetchingNextPage={isFetchingNextTasksPage}
+                onFetchNextPage={fetchNextTasksPage}
+              />
+            )}
+
+            {activeTab === CrmSidePanelTabEnum.DEALS && (
+              <SidePanelDealSection
+                dealIds={contact.dealIds}
```
ThinuwanW, line 286: 🤖 **Claude** · 🟣 **Nit**

`SidePanelDealSection` is rendered without `emptyDescription`, so it falls back to `crmModule.deals.sidePanel.emptyDescription` = "No deals associated with this company" on a *contact* panel. The correct string already exists at `crmModule.contacts.contactDetailsPanel.deals.emptyDescription` ("No deals associated with this contact") but is currently unused — pass it through, mirroring how CompanySidePanel passes `emptyDescription` to the tasks section.

frontend/src/community/crm/v2/api/CompanyApi.ts
```diff
@@ -190,3 +192,44 @@ export const useGetCompaniesByIds = (
...
+  return response?.data?.results?.[0];
+};
+
+export const useGetCompanyLookup = (
+  params: CrmCompanyFilterRequest,
+  enabled?: boolean
+): UseQueryResult<CrmCompanyListResponse> =>
+  useQuery({
+    queryKey: crmCompanyQueryKeys.LOOKUP(params),
+    queryFn: () => fetchCompanyLookup(params),
```
ThinuwanW, line 211: 🤖 **Claude** · 🟣 **Nit**

Two small divergences from the v1 domain-search behaviour: (a) `limit` is set to `DEFAULT_LOOKUP_PAGE_SIZE` (50) whereas `src/community/crm/api/CompanyApi.ts:155` uses the purpose-built `DOMAIN_SEARCH_LIMIT` (10) — the suggestion list is only meant to be a short prefix of the dropdown; (b) in ContactModalForm the query is enabled on `emailDomain.length > 0`, i.e. as soon as the debounced text contains a single `@`, whereas the duplicate-email check next to it is correctly gated on `isValidEmail().test(...)`. That means partially typed addresses fire a request whose domain can fail `CrmValidations.validateDomain`. Consider gating the domain search on the same `isValidEmail()` predicate.

frontend/src/community/crm/v2/components/molecules/TaskRow/TaskRow.tsx
```diff
@@ -0,0 +1,92 @@
...
+    updateCompletion({ id: taskId, payload: { isCompleted: nextIsCompleted } });
+  };
+
+  return (
+    <div
+      role="button"
+      tabIndex={0}
+      aria-label={translateText(["openTaskDetails"], { name: task.name })}
+      className="relative flex items-center gap-4 p-3 min-w-0 min-h-[63px] bg-white hover:bg-secondary-background overflow-hidden cursor-pointer"
+      onClick={onRowClick}
```
ThinuwanW, line 70: 🤖 **Claude** · 🟣 **Nit**

`TaskRow` always renders `role="button"`, `tabIndex={0}`, `cursor-pointer` and `aria-label={translateText(["openTaskDetails"], ...)}` even though `onRowClick` is optional — and `SidePanelTasksList` (the only caller in this PR) never passes one. Screen-reader and keyboard users get a focusable "Open task details panel for X" control that does nothing. Either make `onRowClick` required, or drop the button role/tabIndex/aria-label when it is absent.

frontend/src/community/crm/v2/utils/__test__/sidePanelSectionUtil.test.ts
```diff
@@ -0,0 +1,19 @@
+import { CrmDealStageEnum } from "~community/crm/v2/enums/common";
```
ThinuwanW, line 1: 🤖 **Claude** · 🟣 **Nit**

The file is named `sidePanelSectionUtil.test.ts` but its only subject is `getInitialStageId` imported from `../dealUtil`, and there is no `sidePanelSectionUtil` module. Rename to `dealUtil.test.ts` (or fold it into the existing deal util tests) so the test file maps onto the module under test.

frontend/pages/community/crm/contacts.tsx
```diff
@@ -61,4 +73,58 @@ const Contacts: NextPage = () => {
...
+  const { setIsContactModalOpen, setContactModalType, selectedContactId } =
+    useCrmStoreV2(
+      useShallow((store) => ({
+        setIsContactModalOpen: store.setIsContactModalOpen,
+        setContactModalType: store.setContactModalType,
+        selectedContactId: store.selectedContactId
+      }))
+    );
+
+  useInitializeCrmData();
```
ThinuwanW, line 89: 🤖 **Claude** · 🔴 **Important**

`useInitializeCrmData()` and `ContactTable` now both own the same `contacts` slice, and the initializer wins destructively. `useInitializeCrmData` does `setContacts(toContactsRecord(data.contacts))` — a full replace, not a merge — and the board init-data payload (`CrmBoardContactResponseDto`) carries only `{id, name, company}`: no `email`, `contactNumber`, `companyId`, `ownerId`, `metrics`, `taskIds`, `dealIds`. Both queries are kicked off on mount of the same page, so if init-data resolves after the contact list, every row rendered from `contactIds` collapses to name-only: blank email/phone, `-` for deal value and open tasks, no company, no owner avatar. It is a pure ordering race, so it will reproduce intermittently. Fix by making the initializer merge field-wise (the same shape as `mergeCompanies` in companyUtil.ts) instead of replacing, e.g. `const store = useCrmStoreV2.getState(); store.setContacts(mergeContacts(store.contacts, data.contacts));`, or by keeping board-init lookup contacts in a separate slice from the table's contacts.

frontend/src/community/crm/v2/components/organisms/ContactSidePanel/ContactSidePanel.tsx
```diff
@@ -0,0 +1,302 @@
...
+            name={contact?.name}
+            lastModifiedDate={contact?.lastModifiedDate}
+          />
+        )
+      }
+      headerActions={
+        isLoading ? (
+          <SidePanelHeaderActionsSkeleton />
+        ) : (
+          <KebabMenu
```
ThinuwanW, line 239: 🤖 **Claude** · 🔴 **Important**

The kebab menu renders "Edit contact" and "Delete contact" unconditionally, but `useDeleteContact` calls `DELETE /v1/crm/contact/{id}`, which is annotated `@PreAuthorize("hasRole('ROLE_CRM_SALES_MANAGER')")` (CrmContactController.java:88). A CRM sales representative sees the Delete item, confirms the destructive modal, and gets a 403 surfaced as a generic "Failed to delete contact" toast. The sibling panel added in the same PR does gate this: `CompanySidePanel` renders `CompanySidePanelHeaderActions` which returns `null` when `!isCrmSalesManager`. Mirror that here — pull `isCrmSalesManager` from `useSessionData()` and either hide the whole KebabMenu or filter the delete entry out of `menuItems` for non-managers.

frontend/src/community/crm/v2/components/molecules/AddContactModalContent/AddContactModalContent.tsx
```diff
@@ -0,0 +1,110 @@
...
+      setIsContactModalOpen: store.setIsContactModalOpen
+    }))
+  );
+
+  const { data: currentUser } = useGetUserPersonalDetails();
+
+  const formik = useFormik<CrmContactEntity>({
+    initialValues: getContactFormInitialValues(
+      undefined,
+      Number(currentUser?.employeeId)
```
ThinuwanW, line 49: 🤖 **Claude** · 🔴 **Important**

`getContactFormInitialValues(undefined, Number(currentUser?.employeeId))` produces `ownerId: NaN` for every render before `useGetUserPersonalDetails` resolves (its `select` returns `undefined` until the response lands). Two consequences: (1) `Yup.number()` in `getContactValidationSchema` rejects `NaN` via its type check, so submitting emits Yup's default untranslated message ("ownerId must be a `number` type…") — and for a non-sales-manager `ContactModalForm` renders the owner field only when `selectedOwner` resolves, so that error has nowhere to display and Save appears to do nothing; (2) `enableReinitialize: true` means the form is reset the instant `ownerId` flips from `NaN` to the real id, discarding anything already typed into name/email. Fix: `currentUser?.employeeId ? Number(currentUser.employeeId) : undefined`, and disable Save (or defer mounting the form) while the current-user query is loading.

frontend/src/community/crm/v2/components/organisms/ContactTable/ContactTable.tsx
```diff
@@ -0,0 +1,315 @@
...
+    {
+      id: "owner",
+      label: translateText(["table", "columns", "contactOwnerHeader"]),
+      width: "10%"
+    }
+  ];
+
+  const transformToTableRows = (): GridRow[] =>
+    contactIds.map((id) => {
+      const contact = contacts[id];
```
ThinuwanW, line 197: 🤖 **Claude** · 🔴 **Important**

Several of the new components index a normalized record with an id taken from a separately-maintained id array and immediately dereference the result, so any id/record skew is a TypeError rather than a blank cell. Concretely: here `const contact = contacts[id]` is followed by `contact.metrics` and `contact.name`; `SidePanelContactsSection.tsx:56` does `contacts[contactId]` then `contact.metrics?.…`; `SidePanelDealSection.tsx:99` does `deals[dealId]` and hands it straight to `DealAccordionItemHeader`, which reads `deal.name`/`deal.ownerId`; `TaskRow.tsx:37` does `tasks[taskId]` then `Boolean(task.isCompleted)`; and `TaskRowMeta.tsx:33` passes `owners[task.ownerId]` to `OwnerAvatar`, which reads `owner.authPic`/`owner.firstName`. The ids and the records are written by different effects (and, for `contacts`, by two different owners — see the `useInitializeCrmData` finding), and `DeleteContactModalContent` removes from the record and the id array in two separate `set` calls. Add a guard (`if (!contact) return null` / `.filter(Boolean)` when building rows) so a stale id degrades instead of crashing the page.

frontend/src/community/crm/v2/components/organisms/CompanySidePanel/CompanySidePanel.tsx
```diff
@@ -47,43 +72,134 @@ const CompanySidePanel: FC<CompanySidePanelProps> = ({ companyId }) => {
...
+      !fetchedCompany ||
+      !fetchedMetrics ||
+      !fetchedTasks ||
+      !fetchedDeals ||
+      !fetchedContacts
+    ) {
+      return;
+    }
 
+    const taskItems = fetchedTasks.pages.flatMap((page) => page.items);
```
ThinuwanW, line 166: 🤖 **Claude** · 🔴 **Important**

`fetchedTasks.pages.flatMap((page) => page.items)` (and the same call at ContactSidePanel.tsx:135, plus the deal/contact variants on lines 169-170) assumes every page object has an `items` array. `useGetTasksInfinite` hits `GET /v2/crm/task`, which returns `CrmTaskListResponseDtoV2` = `{ tasks: [...] }` — there is no `items` key — so `page.items` is `undefined`. `flatMap` does not flatten a non-array, so the result is `[undefined]`, and `normalizeTasks` then evaluates `task.id` on `undefined` and throws, taking down the whole side panel rather than showing an empty task list. Independent of fixing the endpoint mismatch, these should be defensive: `page.items ?? []`. Worth doing for the deal and contact flatMaps too, since `response?.data?.results?.[0]` can legitimately be `undefined` on an unexpected payload.

frontend/src/community/crm/v2/utils/taskUtil.ts
```diff
@@ -0,0 +1,190 @@
...
+  }
+};
+
+export const getTaskTypeIcon = (typeName = "", size = 20): ReactElement =>
+  createElement(
+    TASK_TYPE_ICONS[typeName.toLowerCase()] ?? ChecklistVerificationFilledIcon,
+    { width: size, height: size }
+  );
+
+export const getPriorityConfig = (priority?: CrmPriorityEnum) => {
```
ThinuwanW, line 66: 🤖 **Claude** · 🟡 **Suggestion**

`getPriorityConfig` now exists twice with the same name and different semantics: this new one returns `undefined` when no `PRIORITY_OPTIONS` entry matches, while the pre-existing `~community/crm/v2/utils/priorityUtil.ts` defaults an unknown/absent priority to `CrmPriorityEnum.LOW` and non-null-asserts the result. So `TaskRowMeta` (new) renders no priority icon for a task with an unrecognised priority while `PriorityLabel` (existing) renders a LOW icon for the same data. Delete one — keep `priorityUtil.ts` (drop the `!` and return `undefined`) and import from there in `taskUtil`, or fold `priorityUtil` into `taskUtil` and update `PriorityLabel`.

frontend/src/community/crm/v2/components/molecules/ContactModalForm/ContactModalForm.tsx
```diff
@@ -0,0 +1,274 @@
...
+        placeholder={translateText(["placeholders", "email"])}
+        onChange={handleChange}
+        onBlur={handleBlur}
+        aria-label={translateText(["ariaLabels", "email"])}
+        maxLength={CONTACT_EMAIL_MAX_LENGTH}
+        required
+        fullWidth
+      />
+
+      {values.companyId === undefined ? (
```
ThinuwanW, line 174: 🤖 **Claude** · 🟡 **Suggestion**

This PR introduces `SelectableSearchField` — a molecule whose whole job is "render a readOnly InputField with a clear button when a value is selected, otherwise a SearchableDropdown" — but only `TaskModalForm` uses it. The identical pattern is hand-rolled twice more in the same PR: here (the `values.companyId === undefined ? <SearchableDropdown/> : <InputField readOnly rightIcon={<ButtonV2 icon={<CloseIcon/>}/>}/>` block, ~35 lines) and in `SidePanelAddDeal/AddDealContactSearch.tsx`. Route all three through `SelectableSearchField`; it already accepts `selectedValue`, `onClear`, `clearAriaLabel`, `fieldAriaLabel`, `items`, `onSelect` and `emptyMessage`, and only needs an optional `variant`/`state`/`errorMessage` passthrough to cover the AddDeal case.

frontend/src/community/crm/v2/api/utils/ApiEndpoints.ts
```diff
@@ -14,7 +25,8 @@ export const crmDealEndpoints = {
   CREATE_DEAL_STAGE: `${moduleAPIPath.CRM}/deal/stage`,
   UPDATE_DEAL_STAGE: (id: number) => `${moduleAPIPath.CRM}/deal/stage/${id}`,
   REORDER_DEAL_STAGES: `${moduleAPIPath.CRM}/deal/stage/reorder`,
-  DELETE_DEAL_STAGE: (id: number) => `${moduleAPIPath.CRM}/deal/stage/${id}`
+  DELETE_DEAL_STAGE: (id: number) => `${moduleAPIPath.CRM}/deal/stage/${id}`,
+  GET_DEALS: `${moduleAPIPath.CRM}/deal`
```
ThinuwanW, line 29: 🤖 **Claude** · 🟡 **Suggestion**

Dead/duplicated symbols added or left behind by this PR (separate from the QueryKeys cleanup already raised): `crmDealEndpoints.GET_DEALS` on this line is never referenced and duplicates `crmDealEndpointsV2.GET_DEALS` on line 16 — both resolve to `/crm/deal`, and the only two v2 callers (`fetchDeals` in DealApi.ts) use the V2 constant. `TASK_TYPE_ICON_SIZE` (constants/taskConstants.ts:14) has no reference; `getTaskTypeIcon` in taskUtil.ts hardcodes `size = 20` instead — either use the constant or drop it. In `types/CrmTypes.ts`, `CrmContactLookupResponse`, `CrmTaskListResponse`, `CrmTaskRelatedListResponse`, `CrmTaskRelatedFilterRequest`, `CrmTaskFilterRequest` and `CrmCompanyDomainSearchFilterRequest` now have zero references after the ContactApi/TaskApi rewrite (the domain search passes `{domain, limit}` inline instead of using its own request type).

frontend/src/community/crm/v2/api/utils/QueryKeys.ts
```diff
@@ -1,16 +1,42 @@
...
+  LIST: (params: CrmTaskCompletedFilterRequest) => [CRM_TASKS, "list", params]
+};
 
 export const crmDealQueryKeys = {
   GET_DEALS: (filters: CrmDealFilterRequest) => ["crm-deals-v2", filters],
   DEAL_BY_ID: (id: number) => ["crm-deal-v2", id],
   CHECK_DEAL_NAME_EXISTS: (name: string) => ["crm-deal-name-exists-v2", name],
-  DEAL_STAGES: ["crm-deal-stages-v2"]
+  DEAL_STAGES: ["crm-deal-stages-v2"],
+  LOOKUP: (params: CrmDealFilterRequest) => [CRM_DEALS, "lookup", params]
```
ThinuwanW, line 39: 🤖 **Claude** · 🟡 **Suggestion**

The new `crmDealQueryKeys.LOOKUP` is namespaced under `CRM_DEALS` (`"crm-deals"`), but its siblings in the same factory use different roots — `GET_DEALS` → `"crm-deals-v2"`, `DEAL_BY_ID` → `"crm-deal-v2"`, `CHECK_DEAL_NAME_EXISTS` → `"crm-deal-name-exists-v2"`. The factory therefore no longer has a single prefix, so a broad `invalidateQueries({ queryKey: [CRM_DEALS] })` (the obvious follow-up once `useCreateDeal`/`useEditDeal` start invalidating lists) silently misses the deal list and detail queries. Pick one root for the whole factory and add an `ALL` entry, as `crmCompanyQueryKeys` and `crmContactQueryKeys` do.

frontend/src/community/crm/v2/components/molecules/SidePanelTasksSection/SidePanelTasksSection.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const { guardCrmCreate } = useCrmLimitGuard();
+
+  const { setIsTaskModalOpen, setTaskModalType } = useCrmStoreV2(
+    useShallow((store) => ({
+      setIsTaskModalOpen: store.setIsTaskModalOpen,
+      setTaskModalType: store.setTaskModalType
+    }))
+  );
+
+  const translateText = useTranslator(
```
ThinuwanW, line 38: 🤖 **Claude** · 🟡 **Suggestion**

These are generic task molecules but their translator is hard-scoped to the contacts namespace: `useTranslator("crmModule", "contacts", "contactDetailsPanel")` here, and `useTranslator("crmModule", "contacts", "contactDetailsPanel", "tasks")` in `SidePanelTasksList.tsx` and `TaskRowSubtitle.tsx`. All three are rendered inside `CompanySidePanel` as well, so the company panel's "No tasks" title, "Add task" button and every due-date string are resolved out of `crmModule.contacts.*`. It works today only because the keys happen to be generic, and it violates the `module.section.key` convention. Move the shared task strings to `crmModule.tasks.*` (which already holds `openTaskDetails`, `checkTaskMark*`, `toggleError*` used by `TaskRow`/`TaskRowCheckbox`) and keep only contact-specific copy under `contactDetailsPanel`.

frontend/src/community/crm/v2/components/organisms/CompanySidePanel/CompanySidePanel.tsx
```diff
@@ -188,6 +304,36 @@ const CompanySidePanel: FC<CompanySidePanelProps> = ({ companyId }) => {
...
+                onDealCreated={handleDealCreated}
+                companyId={companyId}
+                hasNextPage={hasNextDealsPage}
+                isFetchingNextPage={isFetchingNextDealsPage}
+                onFetchNextPage={fetchNextDealsPage}
+              />
+            )}
+
+            {activeTab === CrmSidePanelTabEnum.CONTACTS && (
+              <SidePanelContactsSection
```
ThinuwanW, line 320: 🤖 **Claude** · 🟡 **Suggestion**

`SidePanelContactsSection` is rendered without `emptyDescription`, so an empty Contacts tab on a company panel falls back to `crmModule.contacts.table.emptyDataState` — "No contacts available." / "Create a contact to get started." — which reads as a global empty state, not "this company has no contacts". The sibling Tasks section on line 334 correctly passes `translateText(["tasks", "emptyDescription"])`, and `crmModule.companies.sidePanel` already has a `tasks.emptyDescription` key. Add a matching `companies.sidePanel.contacts.emptyDescription` and pass it here (and consider a `title` override, since the fallback title is also table-flavoured).

frontend/src/community/crm/v2/components/molecules/SidePanelAddDeal/SidePanelAddDeal.tsx
```diff
@@ -0,0 +1,207 @@
...
+
+  const { guardCrmCreate, isCheckingCrmLimit } = useCrmLimitGuard();
+
+  const isFormDisabled = isPending || isUserLoading || isCheckingCrmLimit;
+
+  const handleSubmit = (values: InlineDealFormValues) => {
+    guardCrmCreate(CrmLimitResource.DEALS, () => {
+      createDeal({
+        name: values.name.trim(),
+        stageId: initialStageId,
```
ThinuwanW, line 117: 🤖 **Claude** · 🟡 **Suggestion**

`stageId: initialStageId` can be `undefined`: `getInitialStageId(stages)` returns `undefined` until the board init-data query has populated the store (line 74), and `isFormDisabled` only guards `isPending || isUserLoading || isCheckingCrmLimit` — it does not wait for stages. A user who opens the panel and hits Save inside that window posts a deal without `stageId`, `CrmValidations.validateDealStageId` throws, and `handleCreateDealError` shows the generic "Failed to add deal / Please try again or contact support" toast with no indication of what went wrong. Include `initialStageId === undefined` in `isFormDisabled` (the current-user guard on the same line is the right precedent).

frontend/src/community/crm/v2/components/molecules/EditableContactOwnerField/EditableContactOwnerField.tsx
```diff
@@ -0,0 +1,111 @@
...
+  );
+
+  const owners = useCrmStoreV2((store) => store.owners);
+
+  const ownerFilters: CrmOwnerLookupFilterRequest = {
+    searchKeyword: debouncedOwnerSearch,
+    size: DEFAULT_LOOKUP_PAGE_SIZE
+  };
+
+  const { data: ownerLookupData, isFetching } = useGetOwnerLookup(ownerFilters);
```
ThinuwanW, line 45: 🤖 **Claude** · 🟡 **Suggestion**

Two issues with this lookup. (1) `useGetOwnerLookup(ownerFilters)` passes no `enabled`, so it defaults to `true` and fires `GET /v1/crm/contact/owners` on mount with an empty `searchKeyword` — and it still fires when an owner is already selected, because the `if (selectedOwner) return <SelectedOwnerField .../>` early return sits below the hook. Every other owner-lookup call site gates it: `TaskModalForm` uses `Boolean(isCrmSalesManager) && debouncedOwnerSearchText.length > 0`, `OwnerPopupSearch` uses `isCrmSalesManager ?? false`. Pass `debouncedOwnerSearch.length > 0 && !selectedOwner`. (2) `handleOwnerSelect` sets only `ownerId` on the form and never merges the picked owner into the store's `owners` record, while `selectedOwner` is resolved via `getOwnerById(owners, ownerId)` from that same store — so any lookup result not present in board init-data would set the value but leave the field looking unselected (the same failure mode as the company field in `ContactModalForm`).

frontend/src/community/crm/v2/utils/contactValidations.ts
```diff
@@ -0,0 +1,42 @@
...
+      })
+      .required(translator(["validations", "name"])),
+    email: Yup.string()
+      .trim()
+      .max(CONTACT_EMAIL_MAX_LENGTH, translator(["validations", "emailLength"]))
+      .matches(isValidEmail(), {
+        message: translator(["validations", "invalidEmail"])
+      })
+      .required(translator(["validations", "email"])),
+    contactNumber: Yup.string()
```
ThinuwanW, line 33: 🤖 **Claude** · 🟣 **Nit**

`CONTACT_NUMBER_MAX_LENGTH` (15) is enforced only through the `maxLength` attribute on the `InputField` in `ContactModalForm`, never in the Yup schema — `name` and `email` both get a `.max(...)` here, `contactNumber` does not. A pasted/programmatic value longer than 15 passes client validation and is rejected only by the backend. Add `.max(CONTACT_NUMBER_MAX_LENGTH, translator(["validations", "contactNumberLength"]))` (and the matching key in `crmModule.json` under both `addContactModal.validations` and `editContactModal.validations`).

frontend/src/community/crm/v2/components/molecules/SidePanelContactInfo/SidePanelContactInfo.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+  const companyName = getCompanyNameById(companies, contact.companyId);
+
+  return (
+    <div className="flex items-center justify-between max-w-[629px] w-full">
+      <SidePanelHeaderInfoItem
+        icon={
+          <EmailOutlineIcon style={{ color: "var(--color-secondary-icon)" }} />
+        }
+        value={contact.email}
```
ThinuwanW, line 28: 🤖 **Claude** · 🟣 **Nit**

The email row (line 28) and phone row (line 33) render unconditionally, which is why `SidePanelHeaderInfoItemProps.value` had to be widened to `value?: string` in this PR. `CrmContactEntity.email`/`contactNumber` are both optional and the backend leaves `contactNumber` null for contacts created without one, so those rows render a bare mail/phone icon with no text next to it. The company row directly below is already guarded with `{companyName && ...}` — apply the same guard to the other two rather than loosening the child's prop type.

frontend/src/community/crm/v2/components/organisms/CompanyTable/CompanyTable.tsx
```diff
@@ -183,9 +183,7 @@ export const CompanyTable: FC = () => {
         height: "34.5rem",
         hasMore: hasNextPage,
         isFetchingNextPage,
-        onLoadMore: () => {
-          void fetchNextPage();
-        }
+        onLoadMore: fetchNextPage
```
ThinuwanW, line 186: 🤖 **Claude** · 🟣 **Nit**

This drive-by change replaces `onLoadMore: () => { void fetchNextPage(); }` with `onLoadMore: fetchNextPage`, and the new `ContactTable.tsx:293` copies it. `TableViewInfiniteScroll.onLoadMore` is typed `() => void`, but `fetchNextPage` has signature `(options?: FetchNextPageOptions) => Promise<...>`. If the underlying table ever invokes the callback with an argument (an event, a page index), React Query receives it as an options object; and the returned promise is now unhandled, which the `void` was presumably there to make explicit. The wrapper form costs nothing — worth keeping in both places.

PR #1833, opened 2026-08-25
refactor: update deal stage components and API integration for v2 str…

dulsara-manakal:
…ucture

### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR ports the CRM configurations tab to a v2 module backed by the normalized `useCrmStoreV2` surface, gated behind a hardcoded `isCrmConfigurationsV2 = false` flag. The port is a close copy of the v1 components, but two behaviours were dropped in the move: default stage names are no longer mapped through `useStageNameMapper`/`getStageDisplayName` before rendering, and the deal-stage list state (`draftStages`) is never re-synced after create/edit/delete because the v2 `DealApi` mutations do not invalidate `crmDealQueryKeys.DEAL_STAGES` (v1 did). Both would be user-visible defects once the flag is flipped, plus several smaller correctness/typing/test-coverage gaps noted below.

Found **22** new issue(s): 🔴 5 important, 🟡 9 suggestion(s), 🟣 8 nit(s)

frontend/src/community/crm/v2/components/organisms/DealStagesSection/DealStagesSection.tsx
```diff
@@ -0,0 +1,205 @@
...
+  );
+
+  const { data: fetchedStages, isLoading } = useGetDealStages();
+
+  const orderedStages = getOrderedStages(stages);
+
+  const [draftStages, setDraftStages] = useState<CrmStageEntity[]>([]);
+  const [hasChanges, setHasChanges] = useState(false);
+
+  useEffect(() => {
```
ThinuwanW, line 64: 🤖 **Claude** · 🔴 **Important**

The stage list will not refresh after add / edit / delete. `draftStages` (what the list actually renders, line 171) is only ever written by this effect, which depends on `fetchedStages`. The v2 mutation hooks in `~community/crm/v2/api/DealApi.ts` (`useCreateDealStage`, `useUpdateDealStage`, `useDeleteDealStage`, `useReorderDealStages`) only invalidate `crmLimitationQueryKeys.GET_CRM_LIMITATION` — unlike their v1 counterparts in `~community/crm/api/crmDealApi.ts`, which all call `queryClient.invalidateQueries({ queryKey: crmDealQueryKeys.DEAL_STAGES })`. So `fetchedStages` keeps its identity, this effect never re-runs, and the modals' manual `setStages(...)` writes land only in the Zustand record, which this component never reads for rendering. Net effect: create a stage → it does not appear; delete a stage → it stays visible; rename/recolor → no change until a full page reload. Fix by adding `queryClient.invalidateQueries({ queryKey: crmDealQueryKeys.DEAL_STAGES })` to the four v2 stage mutations (restoring v1 parity), and/or by deriving `draftStages` from the store (add `orderedStages` to this effect's dependencies, or seed `draftStages` from `getOrderedStages(stages)` whenever there are no unsaved reorder changes).

frontend/src/community/crm/v2/components/molecules/DraggableDealStageCard/DraggableDealStageCard.tsx
```diff
@@ -0,0 +1,105 @@
...
+    listeners,
+    setNodeRef,
+    transform,
+    transition,
+    isDragging
+  } = useSortable({ id: stageId, disabled: !isDraggable });
+  const translateText = useTranslator("configurations", "crm");
+
+  const stageColor = stage.color ?? CrmDealStageColorsEnum.SKY;
+  const stageName = stage.name ?? "";
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Default stage names render as raw enum keys. `stage.name` is used verbatim for the card title (line 74) and for the `aria.editStage` / `aria.deleteStage` labels (lines 84-85, 94-95). The backend seeds default stages with `CrmDealStageName` values (`LEAD`, `QUALIFIED`, `DEMO_SCHEDULED`, `PROPOSAL_SENT`, `NEGOTIATION`, `WON`, `LOST`), so the tab will show "DEMO_SCHEDULED" instead of "Demo Scheduled". v1 avoided this by mapping through `useGetMappedDealStages` → `useStageNameMapper` before the data ever reached the card; the v2 path passes the raw API array straight through `DealStagesSection` → `DealStagesDraggableContent` → this card. Note that `DealStageModalForm` and `DeleteDealStageModalContent` in this same PR *do* translate via `getStageDisplayName`, so the list and the modals will disagree. Fix by resolving the display name here (e.g. `const stageName = getStageDisplayName(translateStageName, stage.name) ?? ""` using `useTranslator("crmModule", "deals", "defaultStageNames")`) and using it for both the title and the aria labels — or by mapping names once in `DealStagesSection` before setting `draftStages`.

frontend/src/community/crm/v2/components/molecules/DealStageModalForm/DealStageModalForm.tsx
```diff
@@ -0,0 +1,237 @@
...
+      color: values.color
+    });
+  };
+
+  const formik = useFormik({
+    initialValues,
+    onSubmit: isEdit ? handleEdit : handleCreate,
+    validationSchema: getStageValidationSchema(
+      translateText,
+      orderedStages,
```
ThinuwanW, line 159: 🤖 **Claude** · 🔴 **Important**

The client-side uniqueness check compares against untranslated names, so duplicate-looking stages can be created. `orderedStages` (line 58) comes from the Zustand record and holds raw names (`DEMO_SCHEDULED`), while `values.name` in the form holds the *display* name — `initialValues.name` is built with `getStageDisplayName(...)` at line 64. The Yup `is-deal-stage-name-unique` test in `stageValidations.ts:36` then lowercases and compares "demo scheduled" against "demo_scheduled" and passes. The backend's `existsByNameIgnoreCaseAndIsDeletedFalse` check compares the same two strings and also passes, so the user ends up with two stages that both render as "Demo Scheduled". v1 passed the *mapped* `dealStages` into `dealStageValidations`, so the check worked. Fix by passing display-mapped names into `getStageValidationSchema` (e.g. `orderedStages.map((s) => ({ ...s, name: getStageDisplayName(translateStageName, s.name) }))`), keeping the comparison basis consistent with what the input field contains.

frontend/src/community/crm/v2/utils/stageUtil.ts
```diff
@@ -0,0 +1,126 @@
...
+  CrmStageRecord
+} from "~community/crm/v2/types/CrmCommonTypes";
+import { CrmDealStageReorderItem } from "~community/crm/v2/types/CrmTypes";
+
+export const getStageDisplayName = (
+  translateText: TranslatorFunctionType,
+  stageName?: string
+) => {
+  if (stageName !== undefined) {
+    return stageName in DefaultStageNameEnum
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`stageName in DefaultStageNameEnum` uses the `in` operator, which walks the prototype chain, so `Object.prototype` keys are treated as default stage names. The stage-name regex (`/^(?=.*\p{L})[\p{L}\p{N} \-.']+$/u`) permits names like `constructor`, `toString`, `valueOf`, and `hasOwnProperty`, so a user-created stage with one of those names would be routed to `translateText(["toString"])`, which resolves to a missing key and renders the raw i18n key string in the UI. Use the same check v1's `useStageNameMapper` uses: `Object.values(DefaultStageNameEnum).includes(stageName as DefaultStageNameEnum)`. While here, give the function an explicit return type (`: string | undefined`) — same for `getSelectedStage` (`: CrmStageEntity | undefined`) — since strict mode is on and the guidelines ask for explicit return types.

frontend/src/community/crm/v2/components/organisms/DealStagesSection/DealStagesSection.tsx
```diff
@@ -0,0 +1,205 @@
...
+    if (!fetchedStages) return;
+
+    setStages(toStagesRecord(fetchedStages));
+    setDraftStages(fetchedStages);
+    setHasChanges(false);
+  }, [fetchedStages]);
+
+  const handleSuccess = () => {
+    setStages(applyStageOrder(stages, draftStages));
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The reorder success handler discards the server response and hand-rolls the cache update from closure state (`stages` and `draftStages`). `useReorderDealStages` types its `onSuccess` as `(reorderedStages: CrmStageEntity[]) => void` and the backend returns the persisted stages, so the authoritative order is available and ignored. `applyStageOrder` also restamps `orderIndex` as `index + 1` across *all* draft stages including WON/LOST, whereas `toStageReorderPayload` deliberately excludes terminal stages from the request — and the backend reassigns from the existing set of order indexes (`availableOrderIndexes`), which may be non-contiguous after deletions. The two can therefore drift apart. Prefer `setStages(toStagesRecord(reorderedStages))` from the mutation response (or invalidate `crmDealQueryKeys.DEAL_STAGES` and let the refetch be the source of truth) instead of the manual restamp.

frontend/src/community/configurations/utils/configurationTabsUtil.tsx
```diff
@@ -7,6 +7,10 @@ import OrganizationConfigurations from "~community/configurations/components/org
 import PeopleConfigurations from "~community/configurations/components/organisms/PeopleConfigurations/PeopleConfigurations";
 import TimeConfigurations from "~community/configurations/components/organisms/TimeConfigurations/TimeConfigurations";
 import { ConfigurationTab } from "~community/configurations/types/ConfigurationTabTypes";
+import CrmConfigurationsV2 from "~community/crm/v2/components/organisms/CrmConfigurations/CrmConfigurations";
+
+// Flip to true to serve the CRM configurations tab from the normalized v2 store surface.
+const isCrmConfigurationsV2 = false;
```
ThinuwanW, line 13: 🤖 **Claude** · 🟡 **Suggestion**

`const isCrmConfigurationsV2 = false` is a hardcoded module constant, so the entire ~900-line v2 tree added by this PR is unreachable dead code that still gets imported and bundled into the configurations chunk. It also means the v2 path ships with zero runtime exercise — the two behavioural bugs flagged above would not surface in any manual or automated pass. Consider driving this from the existing env/config surface (e.g. a `NEXT_PUBLIC_*` flag read through `~community/common/constants/configs`) so it can actually be toggled per environment, and rename it to read as a flag (`IS_CRM_CONFIGURATIONS_V2_ENABLED`). If it must stay a literal, please land it flipped on behind a short-lived branch or add a follow-up ticket, and confirm the bundler isn't shipping both trees to every user.

frontend/src/community/crm/v2/utils/stageUtil.ts
```diff
@@ -0,0 +1,126 @@
+import { TranslatorFunctionType } from "~community/common/types/CommonTypes";
```
ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

No tests accompany the new logic. `stageUtil.ts` (`toStageReorderPayload`, `applyStageOrder`, `getChangedStageFields`, `addStage`/`updateStage`/`removeStage`, `getStageDisplayName`) and `stageValidations.ts` are pure functions with real branching, and the v2 module already has an established convention for this (`utils/__test__/commonUtil.test.ts`, `utils/boardUtil.test.ts`, `hooks/useInitializeCrmData.test.ts`); v1's regex also has `crm/regex/crmRegexPatterns.test.ts`, but the new `isDealStageNameValid` in `crm/v2/regex/crmRegexPatterns.ts` has none. Please add unit tests for at least the reorder payload/order-restamping and the uniqueness + regex validation paths — the reorder payload in particular has a hard backend contract (payload must contain every non-terminal stage and the first entry must be the INITIAL stage, else `CRM_ERROR_DEAL_STAGE_REORDER_INVALID_REQUEST`).

frontend/src/community/crm/v2/components/organisms/DealStagesSection/DealStagesSection.tsx
```diff
@@ -0,0 +1,205 @@
...
+    useShallow((store) => ({
+      stages: store.stages,
+      setStages: store.setStages,
+      setIsDealStageModalOpen: store.setIsDealStageModalOpen,
+      setDealStageModalType: store.setDealStageModalType,
+      setSelectedDealStageId: store.setSelectedDealStageId
+    }))
+  );
+
+  const { data: fetchedStages, isLoading } = useGetDealStages();
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`useGetDealStages()` destructures only `data` and `isLoading`; `isError` is dropped. If the deal-stages fetch fails, the skeleton disappears and the section silently renders an empty list with an enabled "Add" button and no explanation. Every other async path in this PR shows a toast on failure. Pull `isError` off the query and surface a toast (or an inline error state) — the `configurations.crm` namespace already has an error-toast pattern you can mirror, e.g. the `dealsSection.toastMessages.reorderError*` keys.

frontend/src/community/crm/v2/components/molecules/DealStageModalForm/DealStageModalForm.tsx
```diff
@@ -0,0 +1,237 @@
...
+      return;
+    }
+
+    updateDealStage({ id: selectedDealStage.id, ...changedFields });
+  };
+
+  const handleCreate = (values: CrmStageEntity) => {
+    createDealStage({
+      name: values.name?.trim(),
+      description: values.description?.trim(),
```
ThinuwanW, line 149: 🤖 **Claude** · 🟣 **Nit**

Behavioural drift from v1 on empty descriptions. v1's `handleCreate` sent `description: values.description.trim() || null` and `getChangedDealStageFields` normalised `""` → `null` on both sides before diffing; v2 sends `values.description?.trim()`, i.e. an empty string. The backend's `validateDealStageDescription` short-circuits on blank and then calls `stage.setDescription("")`, so you persist `""` where v1 persisted `NULL`. Harmless for display (`description ?? ""`), but it makes the column inconsistent between rows created before and after this change, and any backend query using `IS NULL` would behave differently. Suggest normalising to `null` in both `handleCreate` and `getChangedStageFields` to match v1.

frontend/src/community/crm/v2/components/molecules/DeleteDealStageModalContent/DeleteDealStageModalContent.tsx
```diff
@@ -0,0 +1,142 @@
...
+    if (selectedDealStage?.id !== undefined) {
+      deleteStage(selectedDealStage.id);
+    }
+  };
+
+  return (
+    <div className="flex flex-col">
+      <div>
+        {translateText(["deleteDealStageModal", "description"], {
+          stageName: getStageDisplayName(
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`getStageDisplayName` returns `string | undefined` (it has no `else` branch when `stageName` is undefined), and the result is passed straight in as the `stageName` interpolation value. If `selectedDealStage` is ever undefined — e.g. the store record hasn't been populated yet, or the id was cleared — i18next will not substitute and the modal will literally read "Are you sure you want to delete the '{{stageName}}' stage?". Guard it (`?? ""`, as `DealStageModalForm` does at line 64) or, better, early-return `null` from the component when `selectedDealStage` is undefined — that would also stop the confirm button rendering for a stage that can't be deleted.

frontend/src/community/crm/v2/constants/stageConstants.ts
```diff
@@ -14,3 +16,12 @@ export const STAGE_COLOR_MAP: Record<CrmDealStageColorsEnum, string> = {
   ROSEWOOD: "#b55253",
   INDIGO: "#9366fd"
 };
+
```
ThinuwanW, line 19: 🤖 **Claude** · 🟣 **Nit**

`DEAL_STAGE_COLORS` is a byte-for-byte copy of `dealStageColors` in `~community/crm/utils/crmUtil.ts`, and `isDealStageNameValid` in `crm/v2/regex/crmRegexPatterns.ts` is identical to the v1 version — as are `DealStagesDraggableContent`, `DraggableDealStageCard`, `DraggableDealStageCardSkeleton`, `DealStageModalController`, and `getStageValidationSchema`/`dealStageValidations`. That's expected for a flagged v1→v2 migration, but with the flag hardcoded off the duplication has no expiry date and the two copies have already diverged (see the display-name and uniqueness findings). Please attach a follow-up to delete the v1 `configurations/components/**/DealStage*` tree and the v1 `crmDealApi` stage hooks once the flag flips, so the divergence doesn't become permanent. Separately, `Object.entries(STAGE_COLOR_MAP)` widens `value` to `string`, so `setFieldValue("color", color.value)` writes a plain `string` into a field typed `CrmDealStageColorsEnum` — typing the map as `Object.entries(STAGE_COLOR_MAP) as [CrmDealStageColorsEnum, string][]` would keep the enum contract honest.

frontend/src/community/crm/v2/components/molecules/DealStageModalForm/DealStageModalForm.tsx
```diff
@@ -0,0 +1,237 @@
...
+  const initialValues: CrmStageEntity = {
+    name: isEdit
+      ? (getStageDisplayName(translateStageName, selectedDealStage?.name) ?? "")
+      : "",
+    description: isEdit ? (selectedDealStage?.description ?? "") : "",
+    color: selectedDealStage?.color ?? CrmDealStageColorsEnum.SKY
+  };
+
+  const handleCreateSuccess = (createdStage: CrmStageEntity) => {
+    setStages(addStage(stages, createdStage));
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Hand-patching the store on create desynchronises `orderIndex` from the server. The backend (`CrmDealStageServiceImpl.resolveOrderIndexForNewStage`) inserts a new stage AT the first terminal stage's orderIndex and shifts WON/LOST up by one (`stage.setOrderIndex(stage.getOrderIndex() + 1)` for every stage with orderIndex >= the new one). `addStage` only inserts the returned stage, so the local record ends up with the new stage and WON sharing the same orderIndex, and LOST one lower than the server has it. `getOrderedStages`/`getBoardStageIds` sort by `orderIndex ?? 0`, so with a tie the numeric key order wins and the kanban board renders the new stage AFTER Won. Because `useInitializeCrmData` is gated on `isCrmDataInitialized`, the board-init query never refetches, so the wrong lane order persists for the whole session. Fix by re-deriving the record from a fresh fetch (invalidate `crmDealQueryKeys.DEAL_STAGES` in `useCreateDealStage` and let the query be the source of truth) rather than mutating the store, or at minimum replicate the shift: bump `orderIndex` on every existing stage whose orderIndex >= the created stage's before inserting.

frontend/src/community/crm/v2/utils/stageValidations.ts
```diff
@@ -0,0 +1,46 @@
...
+        "is-deal-stage-name-unique",
+        translator(["dealStageModal", "validations", "nameExists"]),
+        (value) =>
+          stages.every(
+            (stage) =>
+              stage.id === currentStageId ||
+              stage.name?.trim().toLowerCase() !== value.trim().toLowerCase()
+          )
+      ),
+    description: Yup.string()
```
ThinuwanW, line 35: 🤖 **Claude** · 🔴 **Important**

The description rule only enforces max length, but the backend also enforces a character whitelist: `CrmValidations.validateDealStageDescription` rejects anything not matching `^[\p{L}\p{N} \-.,']+$` (CrmConstants.DEAL_STAGE_DESCRIPTION_REGEX). A perfectly ordinary description such as "Deals we won (Q3)" or one containing ':' , '/' or a newline passes client validation, is submitted, and comes back as a ModuleException that the form surfaces as the generic "Failed to add the deal stage. Please try again" toast — the user has no way to know which character is at fault. Add a `.matches()` rule mirroring the backend regex (export it alongside `isDealStageNameValid` in `crm/v2/regex/crmRegexPatterns.ts`) with a dedicated `descriptionInvalid` translation key. Note the name rule was mirrored correctly (same regex and 2/50 bounds), so only description is out of sync.

frontend/src/community/crm/v2/components/molecules/DeleteDealStageModalContent/DeleteDealStageModalContent.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+  const selectedDealStage = getSelectedStage(stages, selectedDealStageId);
+
+  const handleCloseModal = () => {
+    setIsDealStageModalOpen(false);
+  };
+
+  const handleSuccess = () => {
+    if (selectedDealStage?.id !== undefined) {
+      setStages(removeStage(stages, selectedDealStage.id));
```
ThinuwanW, line 43: 🤖 **Claude** · 🟡 **Suggestion**

The delete failure path gives unactionable feedback. The backend rejects deletion deterministically in two cases — `CRM_ERROR_CANNOT_DELETE_TERMINAL_STAGE` and `CRM_ERROR_CANNOT_DELETE_STAGE_WITH_DEALS` (raised when `crmDealDao.existsByStageIdAndIsDeletedFalse` is true) — yet `handleError` always renders "Failed to delete the deal stage. Please try again", which will never succeed on retry. The `onError` callback already receives the `AxiosError`; read `error.response.data.results[0].messageKey` (the same shape `TanStackProvider` inspects) and branch to a specific message, e.g. "This stage still has deals assigned. Move them to another stage first." `CrmErrorMessageKeyEnum` in `crm/v2/enums/common.ts` is the natural place to add the two keys. The same applies to the create/update error toast when the server reports a duplicate name.

frontend/src/community/crm/v2/utils/stageUtil.ts
```diff
@@ -0,0 +1,126 @@
...
+/** Restamps orderIndex from the saved order so the record matches what the server now holds. */
+export const applyStageOrder = (
+  stages: CrmStageRecord,
+  orderedStages: CrmStageEntity[]
+): CrmStageRecord => {
+  const reordered: CrmStageRecord = { ...stages };
+
+  orderedStages.forEach((stage, index) => {
+    if (stage.id !== undefined) {
+      reordered[stage.id] = { ...reordered[stage.id], orderIndex: index + 1 };
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`applyStageOrder` resurrects unknown stages as ghost records. `reordered[stage.id] = { ...reordered[stage.id], orderIndex: index + 1 }` spreads `undefined` when the id is not in the record, producing an entry that has only `orderIndex` — no `id`, `name`, `color` or `stageType`. That is reachable whenever the draft list and the store diverge (e.g. a stage is deleted while an unsaved reorder is pending), and such an entry would flow into `getOrderedStages`, which is what the modal's uniqueness check and other consumers read. Guard it: skip ids that are not already present (`if (stage.id === undefined || !reordered[stage.id]) return;`).

frontend/src/community/crm/v2/components/molecules/DraggableDealStageCard/DraggableDealStageCardSkeleton.tsx
```diff
@@ -0,0 +1,17 @@
+const DraggableDealStageCardSkeleton = () => {
```
ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

This file is byte-for-byte identical to `~community/configurations/components/molecules/DealStageCard/DraggableDealStageCardSkeleton.tsx`, and it also bypasses the v2 module's own skeleton primitive. `crm/v2/components/atoms/SkeletonShape/SkeletonShape.tsx` exists and is what every other v2 skeleton uses (SidePanelHeaderSkeleton, SidePanelTabsSkeleton, CompanySidePanelSkeleton, DealsKanbanBoardSkeletonV2) — it already encapsulates `animate-pulse bg-secondary-accent` plus rounding. Rebuild the two placeholder blocks from `SkeletonShape` so the v2 tree has one skeleton idiom. Same reuse point for `toStageIds` in stageUtil.ts, which overlaps `getBoardStageIds` in `crm/v2/utils/commonUtil.ts` (both extract defined ids from stages, differing only in input shape).

frontend/src/community/crm/v2/utils/stageUtil.ts
```diff
@@ -0,0 +1,126 @@
...
+  CrmDealStageEnum,
+  DefaultStageNameEnum
+} from "~community/crm/v2/enums/common";
+import {
+  CrmStageEntity,
+  CrmStageRecord
+} from "~community/crm/v2/types/CrmCommonTypes";
+import { CrmDealStageReorderItem } from "~community/crm/v2/types/CrmTypes";
+
+export const getStageDisplayName = (
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Exported helpers here lack explicit return types, which the project's strict-typing rule calls for and which is how the implicit `undefined` returns slipped in: `getStageDisplayName` is inferred as `string | undefined` and `getSelectedStage` as `CrmStageEntity | undefined` (the sibling helpers `toStageIds`, `addStage`, `applyStageOrder` all annotate theirs). Annotate both, and reconsider `getChangedStageFields` returning `CrmStageEntity`: because every field on that interface is optional, an update payload is indistinguishable from a full entity, which is exactly why `DealApi.updateDealStage` has to reach for the `id!` non-null assertion. A dedicated `CrmStageUpdatePayload` with a required `id` (mirroring v1's `CrmDealStageUpdatePayload`) would remove the assertion and make an empty-payload PATCH a compile error.

frontend/src/community/crm/v2/components/organisms/DealStagesSection/DealStagesSection.tsx
```diff
@@ -0,0 +1,205 @@
...
+
+  const orderedStages = getOrderedStages(stages);
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Derived values in the new components are recomputed on every render, unlike the surrounding code. `getOrderedStages(stages)` sorts the whole record on each render here and again in DealStageModalForm (line 44), where `getStageValidationSchema(...)` also rebuilds the whole Yup schema per render; `currencyOptions` in the v2 CrmConfigurations (line 16) and the three `stagesData.filter(...)` passes in DealStagesDraggableContent are in the same category. The v2 module already memoizes equivalents (`useMemo` for `industryOptions` in CompanyModalForm, for `stageIds` in DealsKanbanBoardV2, and v1 wrapped this exact sort in `useMemo` inside `useGetMappedDealStages`). Wrap them in `useMemo` keyed on `stages`.

frontend/src/community/crm/v2/components/molecules/DeleteDealStageModalContent/DeleteDealStageModalContent.tsx
```diff
@@ -0,0 +1,142 @@
...
+      ])
+    });
+  };
+
+  const { mutate: deleteStage, isPending } = useDeleteDealStage(
+    handleSuccess,
+    handleError
+  );
+
+  const handleDeleteStage = () => {
```
ThinuwanW, line 85: 🤖 **Claude** · 🟣 **Nit**

Two small state-hygiene issues in the delete flow. (1) `handleDeleteStage` silently no-ops when `selectedDealStage` is undefined, so a stale/missing selection makes the Delete button appear dead with no feedback — either disable the button when `selectedDealStage?.id === undefined` or show the error toast. (2) Neither `handleSuccess` nor `handleCloseModal` resets `selectedDealStageId`, so the global store keeps pointing at an id that was just removed from `stages`; call `setSelectedDealStageId(null)` on close, the way `handleAddStage` in DealStagesSection has to do defensively.

frontend/src/community/crm/v2/components/molecules/DraggableDealStageCard/DraggableDealStageCard.tsx
```diff
@@ -0,0 +1,105 @@
...
+import Icon from "~community/common/components/atoms/Icon/Icon";
+import { useTranslator } from "~community/common/hooks/useTranslator";
+import { IconName } from "~community/common/types/IconTypes";
+import { STAGE_COLOR_MAP } from "~community/crm/v2/constants/stageConstants";
+import { CrmDealStageColorsEnum } from "~community/crm/v2/enums/common";
+import { CrmStageEntity } from "~community/crm/v2/types/CrmCommonTypes";
+
+interface DraggableDealStageCardProps {
+  stage: CrmStageEntity;
+  stageId: number;
```
ThinuwanW, line 15: 🤖 **Claude** · 🟣 **Nit**

The `stageId` prop duplicates `stage.id`, forcing the parent to pass the same value twice and to guard `stage.id !== undefined` inline in the JSX. Since the component already receives the whole entity, either drop `stageId` and narrow inside (the parent's guard already proves it is defined), or pass only the id plus the fields actually rendered. Related: `STAGE_COLOR_MAP[stageColor]` yields `undefined` for any color value the backend returns that is not in `CrmDealStageColorsEnum`; a `?? STAGE_COLOR_MAP[CrmDealStageColorsEnum.SKY]` fallback would match the defensive `stage.color ?? SKY` on line 41.

frontend/src/community/crm/v2/components/organisms/CrmConfigurations/CrmConfigurations.tsx
```diff
@@ -0,0 +1,72 @@
+import { Dropdown, InfoIcon } from "@rootcodelabs/skapp-ui";
```
ThinuwanW, line 1: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

Naming: all seven new components reuse the exact file and component names of their v1 counterparts (CrmConfigurations, DealStagesSection, DealStageModalController, DealStageModalForm, DeleteDealStageModalContent, DealStagesDraggableContent, DraggableDealStageCard), which is why `configurationTabsUtil.tsx` has to alias the import to `CrmConfigurationsV2`. The v2 module elsewhere uses a `V2` suffix for exactly this shadowing case (DealCardV2, DealStageLaneV2, DealsTableV2, DealsKanbanBoardV2, DealSidePanelV2, AddDealSidePanelV2). Suffixing these too would keep grep/imports unambiguous while both trees coexist behind the toggle.

frontend/src/community/crm/v2/constants/stageConstants.ts
ThinuwanW, line 19: 🤖 **Claude** · 🟣 **Nit**

`DEAL_STAGE_COLORS` uses the raw enum key for both `name` and `value`, so the colour swatches are labelled "SKY", "ROSEWOOD", "APRICOT" in whatever the ColorSelector renders for `name` (label/tooltip/aria text). `value` must stay the enum key — the backend maps it to `CrmDealStageColors` — but `name` is user-facing and should come from `t()` like every other option list in this module (cf. `industryOptions` in CompanyModalForm, which builds labels via `useTranslator("crmModule","companies","industryOptions")`). Since this constant is a module-level array it cannot call the hook; expose a `useDealStageColorOptions()` hook or build the options inside the form with the translator.

frontend/src/community/crm/v2/components/organisms/DealStagesSection/DealStagesSection.tsx
dulsara-manakal: We dont need the invalidation, we handle it via zustand store.

frontend/src/community/crm/v2/components/molecules/DraggableDealStageCard/DraggableDealStageCard.tsx
dulsara-manakal: Fixed this.

frontend/src/community/configurations/utils/configurationTabsUtil.tsx
dulsara-manakal: This is for testing purposes.

frontend/src/community/crm/v2/utils/stageUtil.ts
dulsara-manakal: Test added

frontend/src/community/crm/v2/components/molecules/DealStageModalForm/DealStageModalForm.tsx
dulsara-manakal: Fixed this

frontend/src/community/crm/v2/components/organisms/CrmConfigurations/CrmConfigurations.tsx
dulsara-manakal: The root directory is V2 anyway.

frontend/src/community/crm/v2/utils/stageUtil.ts
dulsara-manakal: Fixed this.

frontend/src/community/crm/v2/components/organisms/DealStagesSection/DealStagesSection.tsx
dulsara-manakal: Reordering depends on BE return values, not any local reordering.

frontend/src/community/crm/v2/components/organisms/DealStagesSection/DealStagesSection.tsx
dulsara-manakal: Keep it the same as V1 since the requirement doesn't get changed in refactoring.

frontend/src/community/crm/v2/components/molecules/DeleteDealStageModalContent/DeleteDealStageModalContent.tsx
dulsara-manakal: Not reachable.

frontend/src/community/crm/v2/components/molecules/DealStageModalForm/DealStageModalForm.tsx
dulsara-manakal: we change the types from V1 to V2 so  this wont happen.

frontend/src/community/crm/v2/constants/stageConstants.ts
dulsara-manakal: This is a known issue for this refactoring.

frontend/src/community/crm/v2/components/molecules/DealStageModalForm/DealStageModalForm.tsx
dulsara-manakal: Already fixed

frontend/src/community/crm/v2/utils/stageValidations.ts
dulsara-manakal: Not going to  fix this since V1 had same identical behavior.

frontend/src/community/crm/v2/components/molecules/DeleteDealStageModalContent/DeleteDealStageModalContent.tsx
dulsara-manakal: Not going to fix this since V1 had same identical behavior.

frontend/src/community/crm/v2/utils/stageUtil.ts
dulsara-manakal: This fixed. re ordering is done with BE responce.

frontend/src/community/crm/v2/components/molecules/DraggableDealStageCard/DraggableDealStageCardSkeleton.tsx
dulsara-manakal: Accept things for this refactor.

frontend/src/community/crm/v2/utils/stageUtil.ts
dulsara-manakal: Added

frontend/src/community/crm/v2/components/molecules/DeleteDealStageModalContent/DeleteDealStageModalContent.tsx
dulsara-manakal: The id is only read by modals that set it on open, so a stale value is unreachable.

frontend/src/community/crm/v2/components/organisms/DealStagesSection/DealStagesSection.tsx
dulsara-manakal: Add dependencies

frontend/src/community/crm/v2/components/molecules/DraggableDealStageCard/DraggableDealStageCard.tsx
dulsara-manakal: stageId isn't a duplicate by choice, `CrmStageEntity.id` is optional in the relaxed v2 types, and `useSortable` requires `UniqueIdentifier = string | number`.

frontend/src/community/crm/v2/constants/stageConstants.ts
dulsara-manakal: All twelve colour names are single words, so the only difference is capitalisation.

frontend/src/community/crm/v2/hooks/useStageNameMapper.ts
```diff
@@ -0,0 +1,22 @@
...
+import { DefaultStageNameEnum } from "~community/crm/v2/enums/common";
+
+const useStageNameMapper = () => {
+  const translateText = useTranslator(
+    "crmModule",
+    "deals",
+    "defaultStageNames"
+  );
+
+  const getStageDisplayName = (stageName?: string) => {
```
hasalarootcode: remove this hook and make this as a util then, take the translateText from params in the parent and also stageName should be mandatory for the util and handle the undefined in the parent

hasalarootcode (changes requested):

frontend/src/community/crm/v2/hooks/useStageNameMapper.ts
dulsara-manakal: Removed the hook and moved it to `stageUtil` as `getStageDisplayName`. translateText comes from the parent now, and `stageName` is mandatory.

hasalarootcode (approved):

akilarootcode (approved):

ThinuwanW (approved):

PR #1834, opened 2026-08-25
[Fix] Use placeholder avatar for the work location "All Employees" chip

thusala:
## Summary

The **All Employees** chip in the work location employee selector is not a person, but it was rendering through the same avatar path as real employees — so it showed an initial letter derived from its own label ("A") in the avatar circle. This switches that one chip to the skapp-ui `Avatar` placeholder treatment (`isPlaceholder: true`), which renders the design system's `AvatarIcon` silhouette instead of initials. Employee chips are untouched and keep their photos/initials.

### DIFF
--- a/frontend/src/community/configurations/components/molecules/WorkLocationEmployeeSelector/WorkLocationEmployeeSelector.tsx
+++ b/frontend/src/community/configurations/components/molecules/WorkLocationEmployeeSelector/WorkLocationEmployeeSelector.tsx
@@ -293,7 +293,8 @@ const WorkLocationEmployeeSelector = ({
           avatarProps={{
             id: `${idPrefix}-all-employees`,
             firstName: allEmployeesLabel,
-            size: "sm"
+            size: "sm",
+            isPlaceholder: true
           }}
           label={allEmployeesLabel}
         />


shakila-kamalasena (approved):

akilarootcode (approved):

ThinuwanW (commented):
## 🤖 Claude Code Review

This PR is a two-line, purely presentational change: it adds `isPlaceholder: true` to the `avatarProps` of the 'All Employees' AvatarChip in WorkLocationEmployeeSelector. I read the full component, verified the i18n key `workLocation.form.allEmployees` exists in src/community/common/assets/languages/english/configurations.json:163, confirmed the helper `renderAllEmployeesChip` is the single source for both the trigger and the dropdown option so trigger/option stay visually consistent, and grepped for other `isAllEmployees`/`avatarProps` usages — no other call site needs the same treatment. No correctness, state, React Query, form, security, or i18n issues are introduced; the only residual risk is that `isPlaceholder` cannot be verified against the pinned library version because `@rootcodelabs/skapp-ui` is a private GitHub-registry package and is not installed in this checkout.

Found **6** new issue(s): 🔴 1 important, 🟡 1 suggestion(s), 🟣 4 nit(s)

frontend/src/community/configurations/components/molecules/WorkLocationEmployeeSelector/WorkLocationEmployeeSelector.tsx
```diff
@@ -293,7 +293,8 @@ const WorkLocationEmployeeSelector = ({
           avatarProps={{
             id: `${idPrefix}-all-employees`,
             firstName: allEmployeesLabel,
-            size: "sm"
+            size: "sm",
+            isPlaceholder: true
```
ThinuwanW, line 297: 🤖 **Claude** · 🟣 **Nit**

`isPlaceholder` is the first and only usage of this prop anywhere in the repo (grep for `isPlaceholder` returns only this line plus an unrelated local `isPlaceholderAvailable()` helper in AvatarSearch.tsx), and `@rootcodelabs/skapp-ui` is pinned to a pre-release version (`1.1.2-beta.1`, package.json:32, resolved from npm.pkg.github.com). Since the package isn't installed in this checkout I couldn't confirm the prop exists on the Avatar prop type. Please double-check the CI frontend build (`build-fe.yml`) passed with the pinned version: if the prop is not part of `AvatarProps`, TypeScript's excess-property check on this inline `avatarProps` object literal will fail the build; if the type is loose (index signature / extends MUI AvatarProps), the prop would instead be silently dropped and the intended visual fix would never render. Also worth confirming the beta tag is immutable, or bumping to a stable release, so the prop can't disappear under the same version string.

frontend/src/community/configurations/components/molecules/WorkLocationEmployeeSelector/WorkLocationEmployeeSelector.tsx
```diff
@@ -293,7 +293,8 @@ const WorkLocationEmployeeSelector = ({
           avatarProps={{
             id: `${idPrefix}-all-employees`,
             firstName: allEmployeesLabel,
```
ThinuwanW, line 295: 🤖 **Claude** · 🟣 **Nit**

With `isPlaceholder: true` the avatar presumably renders a generic placeholder graphic instead of initials, which makes `firstName: allEmployeesLabel` on the line above likely dead — it was only there to derive the "AE" initials. Two things to confirm: (1) if skapp-ui's placeholder mode ignores `firstName`, drop it to avoid a misleading prop; if it still uses it for `alt`/`title`, keep it. (2) Unlike EmployeeGroupAvatar (EmployeeGroupAvatar.tsx:22-23, which sets both `alt` and `title`), this Avatar sets neither, so a placeholder rendered as an `img`/icon could end up as an unlabeled graphic — consider adding `alt: allEmployeesLabel`. This is low risk since the surrounding `AvatarChip` still renders the visible `label={allEmployeesLabel}` as accessible text.

frontend/src/community/configurations/components/molecules/WorkLocationEmployeeSelector/WorkLocationEmployeeSelector.tsx
ThinuwanW, line 297: 🤖 **Claude** · 🔴 **Important**

There is no build-time or test-time gate in this repo that can catch this prop if it does not exist on the design-system `Avatar` props, so a silent no-op would reach production with a fully green CI. Concretely: `frontend/next.config.mjs:466-468` sets `typescript: { ignoreBuildErrors: true }` and `eslint: { ignoreDuringBuilds: true }`; `.github/workflows/build-fe.yml` runs only `npm run build` (i.e. `next build`), so TS never runs; `jest.config.js` uses `next/jest` (SWC transpile-only, no typecheck); `lint-staged` in `package.json` runs only prettier; and there is no `.test.tsx` anywhere under `src/community/configurations`. Meanwhile `@rootcodelabs/skapp-ui` is pinned to the exact version `1.1.2-beta.1` (package.json line 32, and package-lock.json resolves it to that exact tarball from the private GitHub registry) and this PR does not bump it. If `isPlaceholder` was added to the design system after `1.1.2-beta.1`, the prop is either dropped silently (chip keeps rendering initials, PR fixes nothing) or spread onto the DOM, producing a `Received \`true\` for a non-boolean attribute \`isplaceholder\`` React warning on every render. Fix: (1) confirm `isPlaceholder` is on the exported `AvatarProps` in exactly `1.1.2-beta.1` — the package already exports `AvatarProps` and it is imported in `src/community/crm/components/atoms/OwnerAvatarChip/OwnerAvatarChip.tsx:1`, so you can verify by typing the literal, e.g. `const avatarProps: AvatarProps = { ... }`; (2) if the prop landed in a later design-system release, bump the dependency and the lockfile in this same PR, otherwise the fix does nothing in the deployed build; (3) separately, consider whether `typescript.ignoreBuildErrors` should stay on, since it means strict mode is documented as a review rule but never actually enforced by CI.

frontend/src/community/configurations/components/molecules/WorkLocationEmployeeSelector/WorkLocationEmployeeSelector.tsx
```diff
@@ -293,7 +293,8 @@ const WorkLocationEmployeeSelector = ({
           avatarProps={{
```
ThinuwanW, line 293: 🤖 **Claude** · 🟡 **Suggestion**

The `avatarProps` object passes neither `alt` nor `title`, which is inconsistent with every other avatar in this same file and module and is newly relevant because of this change. `EmployeeAvatarChip` (src/community/common/components/atoms/EmployeeAvatarChip/EmployeeAvatarChip.tsx:28) passes `alt: employeeName`, and `EmployeeGroupAvatar` (src/community/common/components/atoms/EmployeeGroupAvatar/EmployeeGroupAvatar.tsx:22-23) passes both `alt` and `title` — both are rendered by this component just a few lines below (lines 322-326 and 338-340). Before this PR the All Employees avatar rendered text initials, which are inherently readable; switching to a placeholder graphic means the avatar is now an image/icon with no accessible name of its own. This chip is rendered inside `role="option"` at line 453, where the accessible name is computed from the option's contents, so an `` with no `alt` attribute causes screen readers to fall back to announcing the src/filename inside the option name. Fix: decide explicitly and encode it — either `alt: allEmployeesLabel` for parity with the sibling wrappers, or `alt: ""` to mark it decorative given `label={allEmployeesLabel}` already supplies visible text next to it. Adding `title: allEmployeesLabel` would also match `EmployeeGroupAvatar`.

frontend/src/community/configurations/components/molecules/WorkLocationEmployeeSelector/WorkLocationEmployeeSelector.tsx
```diff
@@ -293,7 +293,8 @@ const WorkLocationEmployeeSelector = ({
...
             id: `${idPrefix}-all-employees`,
             firstName: allEmployeesLabel,
-            size: "sm"
+            size: "sm",
```
ThinuwanW, line 296: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

No test accompanies this change, and this is the only remaining safety net given that both the TypeScript and ESLint build gates are disabled (see the finding on line 297). The repo has 227 `.test.tsx`/`.test.ts` files, but `src/community/configurations` has zero — and note `jest.config.js` excludes `**/organisms/**` from coverage but not `**/molecules/**`, so this file is already counted in the coverage denominator at 0%. Suggest adding `WorkLocationEmployeeSelector.test.tsx` alongside the component (with the `MockTheme` wrapper the other component tests use, and `screen.getByRole` queries per the guidelines) covering at minimum: the All Employees chip renders when `formik.values.isAllEmployees` is true, and `toggleAllEmployees` sets `isAllEmployees` true while clearing `employeeIds` (lines 278-285). A test that asserts on the rendered avatar would also have caught the prop-name risk above.

frontend/src/community/configurations/components/molecules/WorkLocationEmployeeSelector/WorkLocationEmployeeSelector.tsx
```diff
@@ -293,7 +293,8 @@ const WorkLocationEmployeeSelector = ({
...
+            isPlaceholder: true
           }}
           label={allEmployeesLabel}
```
ThinuwanW, line 299: 🤖 **Claude** · 🟣 **Nit**

Dead code in the helper this PR is modifying: `allEmployeesLabel` is built as `translateText(["form.allEmployees"]).trim()` (line 288), but the source string in `src/community/common/assets/languages/english/configurations.json:163` is `"allEmployees": "All Employees"` with no leading or trailing whitespace, so the `.trim()` never changes anything. Unlike `concatStrings([firstName, lastName ?? ""]).trim()` in OwnerAvatarChip.tsx:32 — where the trim genuinely strips a trailing space from a missing last name — there is nothing here to trim. Drop the `.trim()`. Minor related point while you are in this function: `allEmployeesLabel` is recomputed on every invocation, and `renderAllEmployeesChip` is invoked up to twice per render (line 307 for the trigger and line 453 for the dropdown option), so hoisting it above `renderAllEmployeesChip` would avoid the duplicate `translateText` lookup and make it obvious both call sites render the same label.

frontend/src/community/configurations/components/molecules/WorkLocationEmployeeSelector/WorkLocationEmployeeSelector.tsx
thusala: no related to changeset

frontend/src/community/configurations/components/molecules/WorkLocationEmployeeSelector/WorkLocationEmployeeSelector.tsx
thusala: This is the requirement

frontend/src/community/configurations/components/molecules/WorkLocationEmployeeSelector/WorkLocationEmployeeSelector.tsx
thusala: no need

frontend/src/community/configurations/components/molecules/WorkLocationEmployeeSelector/WorkLocationEmployeeSelector.tsx
thusala: not reltated

thusala: not related

frontend/src/community/configurations/components/molecules/WorkLocationEmployeeSelector/WorkLocationEmployeeSelector.tsx
thusala: not related

PR #1835, opened 2026-08-25
[Fix]: 3005 Clear Business Unit form error on edit

shakila-kamalasena:
### Summary

- On the Create/Edit Business Unit form, a validation error (e.g. duplicate name) stayed on screen even after the user corrected the input, because `validateOnChange` is off and the error only re-evaluated on blur/submit.
- Added a `handleFieldChange` wrapper that clears a field's error via `setFieldError` as soon as the user edits it, then delegates to `formik.handleChange`. Wired it to both the name `InputField` and description `TextArea`.
- Keeps the existing `validateOnChange: false` / `validateOnBlur: true` behaviour so validation still runs on blur.

### How to test

1. Log in as Super Admin / People Admin and open Business Unit Configuration.
2. Click Create, enter an existing Business Unit name, and blur/submit to trigger the validation error.
3. Start editing the name to a valid value.
4. The error message should clear immediately as you type (previously it stayed until the next blur).

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/frontend/src/community/configurations/components/molecules/BusinessUnitFormModal/BusinessUnitFormModal.tsx
+++ b/frontend/src/community/configurations/components/molecules/BusinessUnitFormModal/BusinessUnitFormModal.tsx
@@ -6,7 +6,7 @@ import {
   TextArea
 } from "@rootcodelabs/skapp-ui";
 import { useFormik } from "formik";
-import { FC } from "react";
+import { ChangeEvent, FC } from "react";
 
 import {
   useCreateBusinessUnit,
@@ -113,6 +113,15 @@ const BusinessUnitFormModal: FC<Props> = ({
     onSubmit: handleSubmit
   });
 
+  const handleFieldChange = (
+    e: ChangeEvent<HTMLInputElement | HTMLTextAreaElement>
+  ) => {
+    if (formik.errors[e.target.name as keyof BusinessUnitFormValues]) {
+      formik.setFieldError(e.target.name, undefined);
+    }
+    formik.handleChange(e);
+  };
+
   const handleClose = () => {
     if (isPending) return;
     formik.resetForm();
@@ -135,7 +144,7 @@ const BusinessUnitFormModal: FC<Props> = ({
             label={translateText(["form", "nameLabel"])}
             name="name"
             value={formik.values.name}
-            onChange={formik.handleChange}
+            onChange={handleFieldChange}
             onBlur={formik.handleBlur}
             placeholder={translateText(["form", "namePlaceholder"])}
             errorMessage={formik.errors.name}
@@ -147,7 +156,7 @@ const BusinessUnitFormModal: FC<Props> = ({
             label={translateText(["form", "descriptionLabel"])}
             name="description"
             value={formik.values.description}
-            onChange={formik.handleChange}
+            onChange={handleFieldChange}
             onBlur={formik.handleBlur}
             placeholder={translateText(["form", "descriptionPlaceholder"])}
             errorMessage={formik.errors.description}


ThinuwanW (commented):
## 🤖 Claude Code Review

This is a small, well-targeted fix: a `handleFieldChange` wrapper clears the Formik error for a field as soon as the user types, which (combined with `validateOnChange: false`) unblocks the disabled Save button after a validation failure. The approach matches existing codebase precedent (signin.tsx, OrganizationDetails.tsx, DiscountModal.tsx), the `ChangeEvent<HTMLInputElement | HTMLTextAreaElement>` typing is consistent with how skapp-ui `InputField`/`TextArea` are used elsewhere, and no data-integrity risk exists because `formik.handleSubmit()` still re-validates before firing the mutation. I found no correctness bugs — the remaining points are UX/consistency and maintainability.

Found **12** new issue(s): 🟡 6 suggestion(s), 🟣 6 nit(s)

frontend/src/community/configurations/components/molecules/BusinessUnitFormModal/BusinessUnitFormModal.tsx
```diff
@@ -113,6 +113,15 @@ const BusinessUnitFormModal: FC<Props> = ({
     onSubmit: handleSubmit
   });
 
+  const handleFieldChange = (
+    e: ChangeEvent<HTMLInputElement | HTMLTextAreaElement>
+  ) => {
+    if (formik.errors[e.target.name as keyof BusinessUnitFormValues]) {
```
ThinuwanW, line 119: 🤖 **Claude** · 🟡 **Suggestion**

Clearing the error without re-validating makes `formik.isValid` optimistically `true`, so `isSaveDisabled` (line 131 = `isPending || !formik.dirty || !formik.isValid`) enables Save even while the value is still invalid — e.g. the name is still a duplicate, still empty after deleting all characters, or still over BUSINESS_UNIT_NAME_MAX_LENGTH. Clicking Save then re-validates, blocks the submit, re-sets the error and the button snaps back to disabled. No bad data is sent, but the button flickers enabled→disabled. Two cleaner options: (a) revalidate just that field instead of blanking it — `formik.handleChange(e); formik.validateField(e.target.name);` (or `setFieldValue(name, value, true)`), which keeps `isValid` truthful while still clearing the error as soon as the input becomes valid; or (b) drop `!formik.isValid` from `isSaveDisabled` and rely solely on submit-time validation, so the button state is stable. Also note this fix depends on Formik's `setIn` deleting the key when the value is `undefined` (passing `""` as the rest of the codebase does would leave the key present and keep `isValid` false) — that implementation detail is load-bearing here and worth a short comment so a future refactor to `""` doesn't silently re-break the Save button.

frontend/src/community/configurations/components/molecules/BusinessUnitFormModal/BusinessUnitFormModal.tsx
```diff
@@ -6,7 +6,7 @@ import {
   TextArea
 } from "@rootcodelabs/skapp-ui";
 import { useFormik } from "formik";
-import { FC } from "react";
+import { ChangeEvent, FC } from "react";
```
ThinuwanW, line 9: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

This exact clear-error-on-change wrapper is now open-coded in at least six places: pages/community/signin.tsx:156, pages/community/setup-organization.tsx:81, pages/enterprise/setup-organization.tsx:222, settings/.../OrganizationDetails.tsx:43, enterprise/.../DiscountModal.tsx:129 and enterprise/.../EditCustomerContactModal.tsx:192 — each with slightly different semantics (`""` vs `undefined`, before vs after `handleChange`), which is exactly why the `isValid` subtlety above is easy to get wrong. More importantly, the sibling modal in this same directory — DealStageModalForm.tsx:140-179 — has the identical configuration (`validateOnChange: false` + skapp-ui `InputField`/`TextArea` + raw `formik.handleChange`) and therefore still has the bug this PR fixes. Consider extracting a small shared helper/hook (e.g. `useClearFieldErrorOnChange(formik)` under `~community/common/utils` or `hooks`) and applying it to DealStageModalForm in this PR or a follow-up. It would also become unit-testable, matching the existing precedent in people/utils/jobFamilyUtils/jobFamilyUtils.test.ts which tests exactly this kind of error-clearing helper — there is currently no test covering the new behaviour.

frontend/src/community/configurations/components/molecules/BusinessUnitFormModal/BusinessUnitFormModal.tsx
```diff
@@ -113,6 +113,15 @@ const BusinessUnitFormModal: FC<Props> = ({
     onSubmit: handleSubmit
   });
 
+  const handleFieldChange = (
```
ThinuwanW, line 116: 🤖 **Claude** · 🟣 **Nit**

`handleFieldChange` is recreated on every render and passed down to `InputField`/`TextArea`, defeating any memoization in those skapp-ui components. The established pattern in this codebase wraps these handlers in `useCallback` (see OrganizationDetails.tsx:43-50 and signin.tsx:156-163). Suggest `const handleFieldChange = useCallback((e: ChangeEvent<HTMLInputElement | HTMLTextAreaElement>) => { ... }, [formik]);`.

frontend/src/community/configurations/components/molecules/BusinessUnitFormModal/BusinessUnitFormModal.tsx
ThinuwanW, line 119: 🤖 **Claude** · 🟣 **Nit**

`e.target.name as keyof BusinessUnitFormValues` is a type assertion used to index `formik.errors` with an arbitrary DOM string, so TypeScript can no longer catch a mismatch between the `name` props on the JSX (lines 145/157) and the keys of `BusinessUnitFormValues`. If a field name is ever renamed on only one side, the guard silently never matches and the error simply stops clearing — a bug with no compile error and no runtime warning. Safer: narrow instead of asserting, e.g. `const field = e.target.name; if (field in formik.errors) { formik.setFieldError(field, undefined); }` — or drop the guard entirely and call `setFieldError` unconditionally (the guard only saves a re-render, which the `useCallback` above would partly address anyway).

frontend/src/community/configurations/components/molecules/BusinessUnitFormModal/BusinessUnitFormModal.tsx
ThinuwanW, line 9: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Missing paired change: `src/community/configurations/components/molecules/DealStageModalForm/DealStageModalForm.tsx` is a near-identical form in the SAME module (same skapp-ui `InputField`/`TextArea`, same `validateOnChange: false` + `validateOnBlur: true`, same name/description fields, same uniqueness-check validation schema shape) and has the exact same stale-error defect this PR fixes — it still wires `onChange={handleChange}` directly at lines 167 and 179. After this PR the two sibling forms in the same settings screen behave differently: fixing an invalid name clears the error in Business Unit but not in Deal Stage. Either apply the same fix there, or (preferably, and it also addresses the first pass's duplication point) extract one shared `useClearErrorOnChange(formik)` hook under `~community/common/hooks/` and consume it from both.

frontend/src/community/configurations/components/molecules/BusinessUnitFormModal/BusinessUnitFormModal.tsx
```diff
@@ -135,7 +144,7 @@ const BusinessUnitFormModal: FC<Props> = ({
             label={translateText(["form", "nameLabel"])}
             name="name"
             value={formik.values.name}
-            onChange={formik.handleChange}
+            onChange={handleFieldChange}
             onBlur={formik.handleBlur}
             placeholder={translateText(["form", "namePlaceholder"])}
             errorMessage={formik.errors.name}
```
ThinuwanW, line 150: 🤖 **Claude** · 🟡 **Suggestion**

Root cause not addressed: both fields render `errorMessage`/`state` straight off `formik.errors` with no `formik.touched` gate (lines 150-151 for name, 162-163 for description). Because `validateOnBlur: true` runs the WHOLE schema, blurring the Description field surfaces the `nameRequired` error on the Name field the user has never touched. The new clear-on-change handler is a band-aid over this display policy. Gate the display instead: `errorMessage={formik.touched.name ? formik.errors.name : undefined}` and `state={formik.touched.name && formik.errors.name ? "error" : "default"}` (same for description). This is the standard Formik pattern and would let you drop most of the manual error juggling.

frontend/src/community/configurations/components/molecules/BusinessUnitFormModal/BusinessUnitFormModal.tsx
ThinuwanW, line 9: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

No test coverage for the new behavior. There is no `BusinessUnitFormModal.test.tsx` anywhere in the folder, so the newly added change handler ships untested — and it is precisely the kind of state-transition logic that is cheap to test and easy to regress. The codebase already tests this exact pattern elsewhere (`src/community/people/utils/jobFamilyUtils/jobFamilyUtils.test.ts` asserts that the change handler calls `setFieldError`). Add a `.test.tsx` with `MockTheme`, using `@testing-library/react` and `screen.getByRole("textbox", { name: /name/i })`, covering: (a) blur with an empty name renders the required message, (b) typing a valid name removes it, and (c) the Save button's enabled state across that transition.

frontend/src/community/configurations/components/molecules/BusinessUnitFormModal/BusinessUnitFormModal.tsx
```diff
@@ -113,6 +113,15 @@ const BusinessUnitFormModal: FC<Props> = ({
     onSubmit: handleSubmit
   });
 
+  const handleFieldChange = (
+    e: ChangeEvent<HTMLInputElement | HTMLTextAreaElement>
+  ) => {
+    if (formik.errors[e.target.name as keyof BusinessUnitFormValues]) {
+      formik.setFieldError(e.target.name, undefined);
```
ThinuwanW, line 120: 🤖 **Claude** · 🟣 **Nit**

Convention note, and a warning against the wrong fix. Every other clear-on-change site in this repo passes an empty string (`setFieldError(name, "")`) — see `pages/community/signin.tsx:159`, `pages/community/setup-organization.tsx:81`, `src/community/settings/.../OrganizationDetails.tsx:46`, `src/community/people/.../SocialMediaDetailsSection.tsx:68`, and ~10 more. This PR is the only one passing `undefined`. The two are NOT equivalent: Formik's internal `setIn` deletes the key when the value is `undefined` but keeps it when it is `""`. Please do not "fix" the first pass's `formik.isValid` finding by switching to `""` — that would leave the key present forever and permanently disable Save until the next blur/submit revalidation, trading one bug for another. The correct resolution is to keep `undefined` and stop gating `isSaveDisabled` on `formik.isValid`, or to re-run validation after clearing.

frontend/src/community/configurations/components/molecules/BusinessUnitFormModal/BusinessUnitFormModal.tsx
```diff
@@ -113,6 +113,15 @@ const BusinessUnitFormModal: FC<Props> = ({
...
+    }
+    formik.handleChange(e);
+  };
+
   const handleClose = () => {
     if (isPending) return;
     formik.resetForm();
```
ThinuwanW, line 127: 🤖 **Claude** · 🟣 **Nit**

Dead code that the PR's error-lifecycle work should have cleaned up. `formik.resetForm()` here (and the identical call in `handleSuccess`) never has an observable effect: `onClose` is `closeCommonModal`, which sets `commonModalType: CommonModalType.NONE` (`src/community/common/stores/slices/commonModalSlice.ts:25-30`), so `CommonModalController`'s switch falls through to the `default` branch and this component unmounts immediately — formik state, values and errors are all discarded anyway. Both calls just schedule a throwaway state update on an unmounting tree. Safe to delete; if they are being kept deliberately as defensive code, a one-line comment saying so would help, since their presence currently implies the modal survives close (it does not).

frontend/src/community/configurations/components/molecules/BusinessUnitFormModal/BusinessUnitFormModal.tsx
ThinuwanW, line 9: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Server-side duplicate-name errors bypass the form entirely. `handleError` (line 58) takes no parameters and discards the `AxiosError` that `useCreateBusinessUnit`/`useUpdateBusinessUnit` pass in, always rendering the generic `toasts.error` toast. The backend throws `COMMON_ERROR_BUSINESS_UNIT_NAME_ALREADY_EXISTS` from `BusinessUnitServiceImpl` (via `existsByName` / `existsByNameAndBusinessUnitIdNot`) — which is reachable whenever the client-side uniqueness check misses, e.g. a concurrent create by another admin. Given this PR is specifically about the form's error UX, consider accepting the error argument and, when the response maps to that message key, calling `formik.setFieldError("name", translateText(["validations", "nameExists"]))` so the user sees an actionable inline error instead of "something went wrong".

ThinuwanW, line 9: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

Pre-existing validation gap that the new error-clearing logic sits on top of. `useGetBusinessUnits()` (line 41) has no `enabled` flag or loading guard, and `businessUnitValidation` defaults `businessUnits` to `[]` when the data is still `undefined`. During that window the `name-unique` test vacuously passes, so a duplicate name can be submitted and will only fail server-side as a generic toast (see the related finding above). In practice the query is usually warm from `BusinessUnitsSection`, so the impact is small — but disabling Save while `isLoading` is true, or adding `isLoading` to `isSaveDisabled`, would close it.

frontend/src/community/configurations/components/molecules/BusinessUnitFormModal/BusinessUnitFormModal.tsx
```diff
@@ -113,6 +113,15 @@ const BusinessUnitFormModal: FC<Props> = ({
     onSubmit: handleSubmit
   });
 
+  const handleFieldChange = (
+    e: ChangeEvent<HTMLInputElement | HTMLTextAreaElement>
+  ) => {
```
ThinuwanW, line 118: 🤖 **Claude** · 🟣 **Nit**

Missing explicit return type on the new handler, which the project's type-safety rule asks for on every function. Annotate it as `): void => {`. Worth noting the codebase is inconsistent here (`handleClose` below also omits it, while `OrganizationDetails.tsx:52` and `DealStageModalForm.tsx:99` do annotate), so this is only a nit — but new code is the right place to follow the stricter convention.

frontend/src/community/configurations/components/molecules/BusinessUnitFormModal/BusinessUnitFormModal.tsx
shakila-kamalasena: Not needed. This follows the project's Formik convention (clear the field error on change with `setFieldError(name, undefined)`). The Save button re-enabling is cosmetic and submit re-validates before anything is sent, so no invalid data goes through. Reworking the `isSaveDisabled` gating or re-running validation is a behavior change beyond this fix.

frontend/src/community/configurations/components/molecules/BusinessUnitFormModal/BusinessUnitFormModal.tsx
shakila-kamalasena: Not needed. Out of scope. DealStageModalForm and the other sites are separate files this PR does not touch, and extracting a shared hook plus adding tests is a broader refactor. This PR only fixes the Business Unit form.

frontend/src/community/configurations/components/molecules/BusinessUnitFormModal/BusinessUnitFormModal.tsx
shakila-kamalasena: Not needed. The project's Formik convention for this pattern does not wrap the handler in useCallback, and this modal is lightweight.

frontend/src/community/configurations/components/molecules/BusinessUnitFormModal/BusinessUnitFormModal.tsx
shakila-kamalasena: Not needed. The `as keyof BusinessUnitFormValues` assertion is the project's convention for this handler, kept consistent with the other forms.

frontend/src/community/configurations/components/molecules/BusinessUnitFormModal/BusinessUnitFormModal.tsx
shakila-kamalasena: Not needed. Out of scope. DealStageModalForm is a separate file not touched by this PR, and a shared hook is a broader refactor.

frontend/src/community/configurations/components/molecules/BusinessUnitFormModal/BusinessUnitFormModal.tsx
shakila-kamalasena: Not needed. Out of scope. Gating display on `formik.touched` changes the error-display policy across both fields, beyond this PR's fix, which uses the project's clear-on-change convention.

frontend/src/community/configurations/components/molecules/BusinessUnitFormModal/BusinessUnitFormModal.tsx
shakila-kamalasena: Not needed. We do not add UI tests.

frontend/src/community/configurations/components/molecules/BusinessUnitFormModal/BusinessUnitFormModal.tsx
shakila-kamalasena: We are keeping `undefined`, which matches the project convention. No change needed here.

frontend/src/community/configurations/components/molecules/BusinessUnitFormModal/BusinessUnitFormModal.tsx
shakila-kamalasena: Not related to this PR. The `resetForm` calls are pre-existing and not part of this change.

frontend/src/community/configurations/components/molecules/BusinessUnitFormModal/BusinessUnitFormModal.tsx
shakila-kamalasena: Not needed. Out of scope. `handleError` is pre-existing, and mapping server-side duplicate-name errors to an inline field error is a separate feature beyond this fix.

shakila-kamalasena: Not related to this PR. The `useGetBusinessUnits` loading gap is pre-existing and not introduced by this change.

frontend/src/community/configurations/components/molecules/BusinessUnitFormModal/BusinessUnitFormModal.tsx
shakila-kamalasena: Not needed. The codebase is inconsistent here and `handleClose` omits it too, so this is not breaking anything.

hasalarootcode (approved):

akilarootcode (approved):

ThinuwanW (approved):

PR #1836, opened 2026-08-25
fix: re-invite-pending-employee-toast-title

SanithuGunasena07:
renamed to the miss spelled to reinvited propperly

### DIFF
--- a/frontend/src/community/common/assets/languages/english/peopleModule.json
+++ b/frontend/src/community/common/assets/languages/english/peopleModule.json
@@ -575,7 +575,7 @@
       "employeeAddErrorToastTitle": "Oops! Something went wrong",
       "employeeAddErrorToastDescription": "Failed to add the resource. Please try again.",
       "areYouSureModalTitle": "Are you sure?",
-      "reInvitationForOneEmployeeSuccessfulToastTitle": "Reinvitated Successfully",
+      "reInvitationForOneEmployeeSuccessfulToastTitle": "Reinvited Successfully",
       "reInvitationForOneEmployeeSuccessfulToastDescription": "Reinvitation sent for the employee successfully",
       "addFullProfile": "Add Full Profile",
       "save": "Save"


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR is a single-value typo fix in the English `peopleModule.json` locale file, changing the toast title "Reinvitated Successfully" to "Reinvited Successfully". I verified the key `reInvitationForOneEmployeeSuccessfulToastTitle` is consumed only by `useEditEmployee` in `frontend/src/community/people/api/PeopleApi.ts:901` (via `useTranslator("peopleModule", "addResource", "commonText")`, which correctly resolves to the `addResource.commonText` block containing this key), that no tests, snapshots, or other code reference the old string, that English is the only locale bundle in the repo so no other translation files need syncing, and that the new wording now matches the existing bulk-reinvite toast `reInvitationSuccessTitle` at line 481 of the same file. The JSON key name and structure are untouched, so there is no risk of a missing-key regression. The change is correct and safe to merge; I found no blocking issues.

Found **9** new issue(s): 🔴 1 important, 🟡 4 suggestion(s), 🟣 4 nit(s)

frontend/src/community/common/assets/languages/english/peopleModule.json
```diff
@@ -575,7 +575,7 @@
       "employeeAddErrorToastTitle": "Oops! Something went wrong",
       "employeeAddErrorToastDescription": "Failed to add the resource. Please try again.",
       "areYouSureModalTitle": "Are you sure?",
-      "reInvitationForOneEmployeeSuccessfulToastTitle": "Reinvitated Successfully",
+      "reInvitationForOneEmployeeSuccessfulToastTitle": "Reinvited Successfully",
```
ThinuwanW, line 578: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

Out of scope for this PR, but flagging while you are in typo-cleanup mode: the same user-facing wording bug still exists in the enterprise bundle at `frontend/src/enterprise/common/assets/languages/english/pmModule.json` under `reinviteGuestModal` — `successTitle: "Reinvited Succesfully"` and `successDescription: "...re-invited succesfully"` both misspell "successfully", and `errorTitle: "Ops! Something went wrong"` should be "Oops! Something went wrong" to match the convention used everywhere else in the codebase (e.g. `reInvitationErrorTitle` at line 485 of this file). That file lives in the enterprise submodule, so it would need a separate PR against that repo — leaving it to your judgement whether to bundle it into the same cleanup effort.

ThinuwanW, line 578: 🤖 **Claude** · 🔴 **Important**

This toast can assert a reinvitation that the backend never sent, and hide a silently-dropped email change. Backend `PeopleServiceImpl.updateEmployee` -> `createUserEntity` (PeopleServiceImpl.java:587-599) returns early for any existing user, and only calls `resendInvitationEmail` (which is the ONLY place `user.setEmail` runs on the update path) when `accountStatus == PENDING`. Meanwhile the frontend gate that decides whether work email is editable is inconsistent between the two EmploymentDetailsSection variants: `EmploymentFormSection/SubSections/EmploymentDetailsSection.tsx:275-281` correctly checks `employee?.common?.accountStatus !== AccountStatusTypes.PENDING`, but `AddNewResourceFlow/EmploymentDetailsSection/EmploymentDetailsSection.tsx:584-590` (the variant rendered by pages/community/people/directory/edit-all-information/[id].tsx:491) instead checks `!isPendingInvitationListOpen` — a global Zustand UI flag that is only reset on mount of directory/index.tsx and pending.tsx, and stays true while on the edit page. Consequence: reach an ACTIVE employee's edit page while that flag is still true, change the work email, confirm the 'a new invitation will be sent' modal, and the PATCH succeeds while the backend neither persists the new email (responses map work email from `user.email` via PeopleMapper) nor sends any invitation — yet the user sees 'Reinvited Successfully / Reinvitation sent for the employee successfully'. Conversely a genuinely PENDING employee opened from the normal directory tab has the field readOnly, so the reinvite flow is unreachable. Fix: change the AddNewResourceFlow variant to gate on `accountStatus === AccountStatusTypes.PENDING` (matching its sibling) rather than `isPendingInvitationListOpen`, and ideally have the backend return whether an invitation was actually sent so the toast is driven by the response instead of assumed.

ThinuwanW, line 578: 🤖 **Claude** · 🟡 **Suggestion**

Which of the two success toasts this string competes with is chosen from a leaky global flag, so this text can appear after operations that have nothing to do with reinvitation. `useEditEmployee` (PeopleApi.ts:894-915) branches on `isReinviteConfirmationModalOpen` to pick between this toast and `editToastTitle`/`editToastDescription`. That flag is overloaded to mean both 'the modal is visible' and 'the user already confirmed the reinvite', it lives in the shared people store (addNewEmployeeSlice.ts:168), and `resetEmployeeData()` (addNewEmployeeSlice.ts:361-379) does NOT reset it — nor does anything on route change/unmount; only the modal's own onCancel/onClick handlers clear it. So if PeopleTable's bulk-reinvite modal is opened (PeopleTable.tsx:647) and the user navigates away instead of confirming or cancelling, the flag stays true globally. Two knock-on effects: (a) `useEditEmployee` is also consumed by `SetJoinDateModal.tsx:29` (leave module) and `enterprise/sign/.../AddContactModal.tsx:205`, neither of which has anything to do with invitations — a stale-true flag makes saving a join date announce 'Reinvitation sent for the employee successfully', a factually false statement to the user; (b) the same flag doubles as the 'already confirmed' guard at [id].tsx:387 and EmploymentFormSection/EmploymentDetailsForm.tsx:123, so a stale-true value causes the email-change confirmation modal to be skipped entirely and a reinvitation to go out with no confirmation. Fix: stop deriving the toast from modal-visibility state — pass an explicit `isReinvite` flag (or a per-call onSuccess) into the `useEditEmployee` mutation from the callers that actually trigger a reinvite — and add `isReinviteConfirmationModalOpen: false` to `resetEmployeeData()` so the flag cannot survive a route change.

ThinuwanW, line 578: 🤖 **Claude** · 🟡 **Suggestion**

After this fix, this value is byte-identical to `peoples.reInvitationSuccessTitle` at line 481 ('Reinvited Successfully'), which is the bulk-reinvite equivalent consumed by PeopleTable.tsx:225. Two independent copies of the same sentence in two namespaces is exactly what let this typo survive in one copy while the other was already correct — the drift this PR is repairing. Consider collapsing to a single key (keep `peoples.reInvitationSuccessTitle`, or promote a shared key) and have `useEditEmployee` read it, differentiating only the description (singular vs. plural). At minimum add a comment or keep the two adjacent so a future wording change touches both.

ThinuwanW, line 578: 🤖 **Claude** · 🟡 **Suggestion**

Asymmetric error handling around this string pair: the single-employee reinvite path has a dedicated SUCCESS toast (this key and its description) but no dedicated ERROR toast. `useEditEmployee`'s onError (PeopleApi.ts:933-940) always falls back to `editErrorToastTitle`/`editErrorToastDescription` ('Failed to update the resource. Please try again.'), so when a reinvite-triggered email change fails the user is told a generic update failed and has no idea whether the invitation was sent to the old or new address. The bulk path already models this correctly with `reInvitationErrorTitle`/`reInvitationErrorDescription` (lines 485-486). Add a matching `reInvitationForOneEmployeeErrorToastTitle`/`...Description` pair and branch on the reinvite intent in onError the same way onSuccess does.

ThinuwanW, line 578: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Missing cache invalidation on the mutation that renders this toast. `useEditEmployee` (PeopleApi.ts:917-931) invalidates only `EMPLOYEE_COUNT()`, `EMPLOYEE_DATA_TABLE(params)` and `GET_ME`. It never invalidates `peopleQueryKeys.EMPLOYEE_BY_ID(id)` (`["employee-by-id", id]`, QueryKeys.ts:46-48), which is the key `useGetEmployee` reads from (PeopleApi.ts:859-868) — the three invalidated keys share no prefix with it, so nothing cascades. The edit-all-information page papers over this with its own `refetchEmployeeData()`, but `SetJoinDateModal` and the enterprise `AddContactModal` do not, so they show a success toast over a stale individual-employee cache. `PENDING_EMPLOYEE_COUNT` is likewise not invalidated even though a reinvite/email change affects the pending-invitations view. Add `EMPLOYEE_BY_ID(Number(employeeId))` and `PENDING_EMPLOYEE_COUNT` to the invalidation list in the mutation's onSuccess.

frontend/src/community/common/assets/languages/english/peopleModule.json
```diff
@@ -575,7 +575,7 @@
...
       "reInvitationForOneEmployeeSuccessfulToastDescription": "Reinvitation sent for the employee successfully",
```
ThinuwanW, line 579: 🤖 **Claude** · 🟣 **Nit**

Grammar issue in the paired description that the title fix leaves behind: an invitation is sent TO a person, not FOR them. 'Reinvitation sent for the employee successfully' should read 'Reinvitation sent to the employee successfully' (or, more naturally, 'A new invitation has been sent to the employee.'). The bulk sibling at line 482 has the identical preposition error ('Reinvitation sent for the selected employees' -> 'sent to the selected employees'). Also in the same reinvite string family, line 484 (`workEmailChangingReinvitationConfirmationModalDescription`) contains a double space between 'address.' and 'Are you sure' — collapse it to one. Since this PR is a copy fix for exactly these strings, worth sweeping them together.

frontend/src/community/common/assets/languages/english/peopleModule.json
```diff
@@ -575,7 +575,7 @@
       "employeeAddErrorToastTitle": "Oops! Something went wrong",
       "employeeAddErrorToastDescription": "Failed to add the resource. Please try again.",
```
ThinuwanW, line 576: 🤖 **Claude** · 🟣 **Nit**

Inconsistent user-facing product terminology inside the very block being edited: the success toasts on lines 573-574 speak of a 'User' ('User Added', 'User has been added and an invitation has been sent successfully.'), while the error toast here and its neighbours leak the internal term 'resource' ('Failed to add the resource. Please try again.', plus line 561 'Resource added successfully' and line 563 'Resource updated successfully.'). 'Resource' is internal jargon and reads oddly for a person. Standardise this block on 'user' (or 'employee', matching line 579) so a failed add and a successful add describe the same noun.

frontend/src/community/common/assets/languages/english/peopleModule.json
ThinuwanW, line 578: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

Misspelled i18n keys remain in the same reinvite feature this PR is tidying — flagging so the cleanup can be finished in one pass rather than drifting again. `reInvitenConfirmationModalTitle` and `reInvitenConfirmationModalDescription` (lines 487-488) should be `reInviteConfirmationModal*`; the only consumers are PeopleTable.tsx:746-747, so the rename is a two-line paired change. Separately, `divesityDetails` (line 839) should be `diversityDetails`, with consumers at AddNewResourceFlow/EmploymentDetailsSection/IdentificationDetailsSection.tsx:33 and EmploymentFormSection/SubSections/IdentificationDetailsSection.tsx:40. Note there is no build-time check that translation keys resolve, so any rename must update every consumer in the same commit or the raw key string renders in the UI.

SanithuGunasena07: Out of scope for this PR

SanithuGunasena07: pre existing out of scope for PR

SanithuGunasena07: pre existing out of scope for PR

SanithuGunasena07: pre existing out of scope for PR

SanithuGunasena07: pre existing out of scope for PR

SanithuGunasena07: pre existing out of scope for PR

frontend/src/community/common/assets/languages/english/peopleModule.json
SanithuGunasena07: pre existing out of scope for PR

frontend/src/community/common/assets/languages/english/peopleModule.json
SanithuGunasena07: pre existing out of scope for PR

frontend/src/community/common/assets/languages/english/peopleModule.json
SanithuGunasena07: pre existing out of scope for PR

hasalarootcode (approved):

akilarootcode (approved):

PR #1837, opened 2026-08-25
[Fix] Keep Assign Leave Policy calendar inside the viewport

thusala:
## Summary

The Assign Leave Policy modal could not be completed on a 1366x768 laptop at 100% zoom. The skapp-ui `DatePicker` Popper always draws its calendar *below* the trigger, and the "specific date" input sits low inside a vertically centred modal — so the calendar's last week rows fell past the bottom of the viewport and the Popper's own `max-height` clipped roughly 33px of them. Because the modal locks background scrolling, those rows were unreachable: a user could not pick a date in the final week of a six-row month.

This PR adds a `useCalendarPopperPlacement` hook that measures the trigger against the viewport when the calendar opens and places the calendar on whichever side it actually fits, so the full month stays reachable at any modal position or viewport height.


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds a `useCalendarPopperPlacement` hook that manually flips/pins a skapp-ui DatePicker popper inside a modal, and wires it into `AssignLeavePolicyForm`. The placement math is sound for the flip cases, but the documented scroll fallback is a no-op, the calendar height is a magic constant that will silently drift from the design system, and the library appears to already expose `isFlip` for this exact problem (used at `TaskModalForm.tsx:361`). The sibling `SetJoinDateModal` — opened from this same flow — has the identical bug and was not migrated.

Found **18** new issue(s): 🔴 3 important, 🟡 10 suggestion(s), 🟣 5 nit(s)

frontend/src/community/common/hooks/useCalendarPopperPlacement.ts
```diff
@@ -0,0 +1,98 @@
...
+
+    const onCalendarOpenChange = useCallback((isOpen: boolean): void => {
+      if (isOpen) {
+        setPlacement(getPlacement(triggerRef.current));
+      }
+      setIsCalendarOpen(isOpen);
+    }, []);
+
+    const calendarPopperProps = useMemo(
+      () => ({ ...placement, containerClassName: "overflow-y-auto" }),
```
ThinuwanW, line 55: 🤖 **Claude** · 🔴 **Important**

`containerClassName: "overflow-y-auto"` is a no-op as written. `overflow-y-auto` only produces a scrollbar when the element's height is constrained; here the popper container has no `max-height`, so it sizes to its content and never overflows. The doc comment above promises "lets the calendar scroll if even that is not enough room" — that fallback does not actually work, and it is precisely the branch that matters in the short-viewport case the pinned-offset path lands in (e.g. `window.innerHeight` 300 → pinned top 16 + 340px calendar = 356px, still 56px past the fold, with no scroll). Fix: constrain the height alongside the overflow, e.g. `containerClassName: "max-h-[calc(100vh-2rem)] overflow-y-auto"`, ideally derived from `CALENDAR_POPPER_VIEWPORT_MARGIN_PX` so the two stay in sync.

frontend/src/community/common/constants/commonConstants.ts
```diff
@@ -38,3 +38,7 @@ export const SEARCH_DEBOUNCE_DELAY = 500;
 export const BUSINESS_UNIT_TRANSFER_UNASSIGN_VALUE = "unassign";
 export const BUSINESS_UNIT_NAME_MAX_LENGTH = 100;
 export const BUSINESS_UNIT_DESCRIPTION_MAX_LENGTH = 250;
+
+export const CALENDAR_POPPER_HEIGHT_PX = 340;
```
ThinuwanW, line 42: 🤖 **Claude** · 🟡 **Suggestion**

`CALENDAR_POPPER_HEIGHT_PX = 340` is an unexplained magic number coupled to the rendered height of a third-party component (`@rootcodelabs/skapp-ui` DatePicker, pinned at `1.1.2-beta.1`). Nothing in the repo enforces it: a design-system padding change, a 6-week month grid vs a 5-week one, or the taller year/month picker views all break the assumption silently, and the calendar reappears past the fold — the exact bug this PR fixes. Prefer measuring the popper content after open (ref + `ResizeObserver`, or a `getBoundingClientRect()` on the popper container) so the placement tracks reality. If the constant stays, add a comment stating where 340 came from, that it must be re-verified on every skapp-ui bump, and that it assumes the widest (6-row) month layout so the estimate errs on the safe side.

frontend/src/community/common/hooks/useCalendarPopperPlacement.ts
```diff
@@ -0,0 +1,98 @@
...
+    CALENDAR_POPPER_VIEWPORT_MARGIN_PX,
+    window.innerHeight -
+      CALENDAR_POPPER_VIEWPORT_MARGIN_PX -
+      CALENDAR_POPPER_HEIGHT_PX
+  );
+
+  return { position: "bottom-start", offset: pinnedTop - bottom };
+};
+
+/**
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The library appears to already solve this: `TaskModalForm.tsx:361` renders a DatePicker inside a modal with `popperProps={{ position: "bottom-start", isFlip: true }}`, and `isFlip` is used on skapp-ui poppers elsewhere too (`YearSelector.tsx:61`, `MultiTeamSelector.tsx:189`, `DocumentsTable.tsx:158`). Before landing 98 lines of hand-rolled placement plus a hardcoded height constant, please confirm `isFlip` (combined with a `max-h` on the container) doesn't cover this case. If it genuinely doesn't — e.g. `isFlip` only flips and can't pin when neither side fits — say so explicitly in the JSDoc block so the next reader doesn't re-litigate it, and consider pushing the pin behaviour upstream into skapp-ui instead.

frontend/src/community/common/hooks/useCalendarPopperPlacement.ts
```diff
@@ -0,0 +1,98 @@
...
+  position: "bottom-start",
+  offset: CALENDAR_POPPER_OFFSET_PX
+};
+
+const ABOVE_TRIGGER: CalendarPopperPlacement = {
+  position: "top-start",
+  offset: CALENDAR_POPPER_OFFSET_PX
+};
+
+const getPlacement = (
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

This is now the third independent implementation of the same "flip the popper when there isn't room below" heuristic, each with a different hardcoded threshold: `InputDate.tsx:231-245` (`bottomSpace < 300`), `ChipAutocomplete.tsx:60-75` (`flipThreshold = 220`), and this one (340 + 8 + 16). Three different numbers for what is conceptually one rule means the behaviour is inconsistent across the app and each site has to be fixed separately. Consider generalising this hook to take the estimated content height (and optionally the ref element type) as parameters, then migrating `ChipAutocomplete` onto it — `InputDate` is on MUI Popper so it can stay out of scope.

frontend/src/community/common/hooks/useCalendarPopperPlacement.ts
```diff
@@ -0,0 +1,98 @@
...
+ * side fits, and lets the calendar scroll if even that is not enough room.
+ */
+export const useCalendarPopperPlacement =
+  (): UseCalendarPopperPlacementReturn => {
+    const triggerRef = useRef<HTMLDivElement>(null);
+    const [isCalendarOpen, setIsCalendarOpen] = useState<boolean>(false);
+    const [placement, setPlacement] =
+      useState<CalendarPopperPlacement>(BELOW_TRIGGER);
+
+    const onCalendarOpenChange = useCallback((isOpen: boolean): void => {
```
ThinuwanW, line 34: 🤖 **Claude** · 🟡 **Suggestion**

Placement is computed once, on open, and never recomputed. While the calendar is open, a window resize, a device rotation, or a mobile on-screen keyboard appearing changes `window.innerHeight` and the trigger rect, leaving the popper positioned against stale measurements — and in the pinned branch the stale value is an absolute pixel offset, so it can push the calendar completely off-screen rather than merely misaligning it. Add a `resize` listener (and, if the popper can reposition, a scroll listener) registered while `isCalendarOpen` is true that re-runs `getPlacement(triggerRef.current)`, cleaned up in the effect teardown.

frontend/src/community/common/hooks/useCalendarPopperPlacement.ts
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  const pinnedTop = Math.max(
+    CALENDAR_POPPER_VIEWPORT_MARGIN_PX,
+    window.innerHeight -
+      CALENDAR_POPPER_VIEWPORT_MARGIN_PX -
+      CALENDAR_POPPER_HEIGHT_PX
+  );
+
+  return { position: "bottom-start", offset: pinnedTop - bottom };
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

The pinned branch returns `offset: pinnedTop - bottom`, which is typically a large negative number (trigger bottom at 700px, pinned top at 16px → `offset: -684`). This assumes skapp-ui's Popper applies `offset` as an unclamped signed translation along the main axis, with no internal collision handling that would fight it. No other call site in this repo passes `offset` at all, so that semantic is unverified here. Please confirm against the skapp-ui Popper implementation (negative values accepted, not `Math.max(0, …)`-clamped, and not overridden when `isFlip`/collision logic is active) and note the finding in the JSDoc, since the whole third branch silently degrades to a plain `bottom-start` popper if the assumption is wrong.

frontend/src/community/leave/components/molecules/AssignLeavePolicyModal/AssignLeavePolicyForm.tsx
```diff
@@ -15,6 +15,7 @@ import type {
   GridHeader,
   GridRow
 } from "~community/common/components/organisms/TableView/types";
+import { useCalendarPopperPlacement } from "~community/common/hooks/useCalendarPopperPlacement";
```
ThinuwanW, line 18: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

`SetJoinDateModal.tsx:62-68` has the identical problem and was not migrated: a skapp-ui DatePicker inside a `SmallModal` with `popperProps={{ position: "bottom-start" }}` and a bare `<div>` trigger. It is not an unrelated screen — `AssignLeavePolicyModal.tsx:275-279` renders it directly, so a user hitting the join-date path from this very flow still gets the unreachable calendar this PR exists to fix. Apply `useCalendarPopperPlacement` there too (it is a ~5-line change mirroring this file), or state in the PR description why that modal is exempt.

frontend/src/community/leave/components/molecules/AssignLeavePolicyModal/AssignLeavePolicyForm.tsx
```diff
@@ -161,9 +169,12 @@ const AssignLeavePolicyForm: FC<Props> = ({
                 date ? (DateTime.fromJSDate(date).toISODate() ?? "") : ""
               )
             }
-            popperProps={{ position: "bottom-start" }}
+            triggerRef={triggerRef}
+            isOpen={isCalendarOpen}
```
ThinuwanW, line 173: 🤖 **Claude** · 🟣 **Nit**

The DatePicker is now controlled via `isOpen`, but it is conditionally rendered behind `effectiveDateType === EffectiveDateType.SPECIFIC` (line 159) while `isCalendarOpen` lives in the hook and outlives the unmount. If the calendar is open and the user switches to the Join Date radio — and the popper's outside-click handler does not fire `onOpenChange(false)` first, which many popper implementations swallow — `isCalendarOpen` stays `true`; switching back to Specific remounts the DatePicker already open, with a `placement` measured against the trigger's old position. Same applies if `SmallModal` keeps its `content` mounted when `isOpen` flips false (e.g. while `SetJoinDateModal` is showing) and `resetForm()` runs. Cheapest guard: `isOpen={isCalendarOpen && effectiveDateType === EffectiveDateType.SPECIFIC}`, or expose a `closeCalendar` from the hook and call it from `handleEffectiveDateTypeChange`.

frontend/src/community/common/hooks/useCalendarPopperPlacement.ts
```diff
@@ -0,0 +1,98 @@
+import { PopperPosition } from "@rootcodelabs/skapp-ui";
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

No test accompanies this hook, though the repo has an established colocated `renderHook` convention (25 hook test files, e.g. `src/community/people/hooks/useStepper.test.tsx`, `src/fallback/common/hooks/useProductTour.test.ts`). `getPlacement` is pure branch logic — below/above/pinned plus the null-trigger guard — and is trivially testable by stubbing `window.innerHeight` and `getBoundingClientRect`, which is exactly the kind of geometry math that regresses invisibly when the height constant drifts. Export `getPlacement` (or move it to a `utils` file) so it can be asserted directly, and add `useCalendarPopperPlacement.test.ts` covering the three branches and the `null` trigger.

frontend/src/community/common/hooks/useCalendarPopperPlacement.ts
ThinuwanW, line 55: 🤖 **Claude** · 🔴 **Important**

`containerClassName: "overflow-y-auto"` almost certainly *replaces* the DatePicker's default popper container classes rather than merging with them. Every other Popper consumer in this repo passes visual styling through this exact prop and treats it as the full class list — `TableView.tsx:131` and `FilterButton.tsx:63` use `"rounded-4 shadow-lg"`, `ChipAutocomplete.tsx:257` and `SearchableDropdown.tsx:223` use `"rounded-md border border-secondary-accent bg-white shadow-lg"`. If the skapp-ui DatePicker forwards `popperProps.containerClassName` straight to its internal Popper (which is what the `Omit<PopperProps, ...>` spread pattern in `TableView.tsx:47-55` implies), overriding it with only `overflow-y-auto` strips the calendar's background, border and shadow — leaving a transparent calendar rendered over the modal content. Fix: preserve the surface classes alongside the overflow, e.g. `containerClassName: "overflow-y-auto rounded-4 border border-secondary-accent bg-white shadow-lg"`, or better, accept a `containerClassName` argument on the hook and append to it so callers can keep the library default. Verify visually against a build before merging.

frontend/src/community/common/hooks/useCalendarPopperPlacement.ts
```diff
@@ -0,0 +1,98 @@
...
+import { RefObject, useCallback, useMemo, useRef, useState } from "react";
+
+import {
+  CALENDAR_POPPER_HEIGHT_PX,
+  CALENDAR_POPPER_OFFSET_PX,
+  CALENDAR_POPPER_VIEWPORT_MARGIN_PX
+} from "~community/common/constants/commonConstants";
+
+interface CalendarPopperPlacement {
+  position: PopperPosition;
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The popper prop shapes are hand-rolled instead of derived from the library's exported types, and this PR is the repo's *only* consumer of `position`+`offset` together, of `triggerRef`, and of `onOpenChange` (grep confirms zero other usages of `offset`, `triggerRef`, or `onOpenChange` anywhere under `src/`). The established pattern is `TableView.tsx:1-6,47-55`, which imports `type PopperProps` from `@rootcodelabs/skapp-ui` and types against it via `Omit<PopperProps, ...>`. Because `CalendarPopperPlacement` and the `& { containerClassName: string }` intersection are independent structural types, they'll still typecheck if skapp-ui renames or drops `offset`/`containerClassName` in a future version — the prop is then silently dropped at runtime and the whole placement fix stops working with no compile error. Fix: `import { type PopperProps } from "@rootcodelabs/skapp-ui"` and declare `type CalendarPopperPlacement = Required<Pick<PopperProps, "position" | "offset">>`, with the return type as `Pick<PopperProps, "position" | "offset" | "containerClassName">`.

frontend/src/community/leave/components/molecules/AssignLeavePolicyModal/AssignLeavePolicyForm.tsx
```diff
@@ -161,9 +169,12 @@ const AssignLeavePolicyForm: FC<Props> = ({
...
+            onOpenChange={onCalendarOpenChange}
+            popperProps={calendarPopperProps}
           >
-            <div>
+            <div ref={triggerRef}>
```
ThinuwanW, line 177: 🤖 **Claude** · 🟡 **Suggestion**

The same `triggerRef` object is wired to two consumers: as the DatePicker's `triggerRef` prop (line 172) and as `ref` on the child `<div>` (line 177). React attaches refs bottom-up during commit, so if skapp-ui's DatePicker also *writes* to the ref you hand it (i.e. `triggerRef` is meant to be attached by the DatePicker to its own trigger wrapper, not read from), the last writer wins and `triggerRef.current` will point at an element you didn't intend — silently measuring the wrong rect in `getPlacement`. Since the ref is already on the child element the popper anchors to, pick one: either drop `ref={triggerRef}` from the div (if the prop is write-side) or drop `triggerRef={triggerRef}` (if it's read-side). Confirm against the skapp-ui 1.1.2-beta.1 `DatePicker` prop docs which direction it is, since there is no other usage in the repo to copy from.

frontend/src/community/leave/components/molecules/AssignLeavePolicyModal/AssignLeavePolicyForm.tsx
```diff
@@ -161,9 +169,12 @@ const AssignLeavePolicyForm: FC<Props> = ({
                 date ? (DateTime.fromJSDate(date).toISODate() ?? "") : ""
               )
             }
```
ThinuwanW, line 171: 🤖 **Claude** · 🟡 **Suggestion**

Nothing closes the calendar when a date is picked. `onSelect` (lines 167-171) only calls `onSpecificDateChange` — it never calls `onCalendarOpenChange(false)`. Before this PR the DatePicker was uncontrolled and owned its own open state, so it could close itself on select; now `isOpen` is fully driven by the hook's `isCalendarOpen`, and the calendar will stay open after selection unless the library independently fires `onOpenChange(false)` on select. This is exactly the class of behavioral drift that controlled/uncontrolled migrations introduce. Fix: have the hook expose a `closeCalendar` callback and call it from `onSelect` after `onSpecificDateChange`, so the close is guaranteed regardless of library internals.

frontend/src/community/leave/components/molecules/AssignLeavePolicyModal/AssignLeavePolicyForm.tsx
ThinuwanW, line 177: 🤖 **Claude** · 🟡 **Suggestion**

The calendar's open state is now available in this component but is not surfaced to assistive tech. The trigger `<div>` (line 177) wrapping the read-only `InputField` has no `aria-haspopup` or `aria-expanded`, so screen-reader users get no indication that activating the field opens a calendar or whether it is currently open. The repo already does this correctly elsewhere: `ChipAutocomplete.tsx:230-231` sets `aria-expanded`/`aria-controls`, and `TableView.tsx:84-85` sets `aria-expanded`/`aria-haspopup="dialog"` on its filter trigger. Fix: add `aria-haspopup="dialog"` and `aria-expanded={isCalendarOpen}` (plus `aria-controls` pointing at the popper id if the DatePicker accepts one) to the trigger element or the `InputField`.

frontend/src/community/common/hooks/useCalendarPopperPlacement.ts
ThinuwanW, line 55: 🤖 **Claude** · 🟡 **Suggestion**

Applying `overflow-y-auto` to the popper container turns it into a scroll/clip container for *all* its descendants, not just the day grid. If the skapp-ui DatePicker renders its month/year navigation as an absolutely-positioned overlay inside the calendar surface (the usual pattern — `InputDate.tsx:448` shows the MUI equivalent supports `views={["year", "month", "day"]}`), that overlay will now be clipped or forced to scroll instead of overflowing the calendar. Note this class is applied unconditionally on every open, including the common case where the calendar fits fine below the trigger, so the risk is not limited to the cramped-viewport path. Fix: scope the class to the pinned branch only (return it from `getPlacement` alongside the placement), and pair it with an explicit `max-height` so it applies to the day grid rather than the whole surface.

frontend/src/community/common/hooks/useCalendarPopperPlacement.ts
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

The JSDoc and the code disagree on when scrolling kicks in. The comment says the hook "lets the calendar scroll if even that is not enough room", implying scroll is the last-resort branch, but `containerClassName: "overflow-y-auto"` is merged into `calendarPopperProps` unconditionally at line 88 — it applies identically in the `BELOW_TRIGGER` and `ABOVE_TRIGGER` cases where the calendar demonstrably fits. Either make the behavior conditional to match the doc, or reword the doc so the next reader doesn't assume a branch that isn't there.

frontend/src/community/common/constants/commonConstants.ts
ThinuwanW, line 42: 🤖 **Claude** · 🟣 **Nit**

These three constants are consumed by exactly one file (`useCalendarPopperPlacement.ts`) and are not "common" in any cross-module sense — unlike the surrounding entries such as `SEARCH_DEBOUNCE_DELAY` or `MODAL_HEADER_ICON_GAP`. Putting single-consumer values in a shared constants barrel grows it without benefit and makes it harder to see that the numbers are coupled to one hook's geometry. Fix: declare them as module-local `const`s at the top of `useCalendarPopperPlacement.ts` (where `BELOW_TRIGGER`/`ABOVE_TRIGGER` already live), and only promote them to `commonConstants.ts` if a second consumer appears.

frontend/src/community/leave/components/molecules/AssignLeavePolicyModal/AssignLeavePolicyForm.tsx
ThinuwanW, line 18: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

This file is now 243 lines, past the >200-line threshold in the review guidelines for decomposing into atoms/molecules, and this PR pushed it further. The specific-date block (lines 159-203) is a self-contained unit — it owns the DatePicker, the trigger, the error state, and now the popper-placement hook. Extracting it into a `SpecificDateField` molecule that calls `useCalendarPopperPlacement` internally would also remove the need to thread `triggerRef`/`isCalendarOpen`/`calendarPopperProps`/`onCalendarOpenChange` through this component's body, and would give the hook a natural, testable colocated consumer.

frontend/src/community/common/hooks/useCalendarPopperPlacement.ts
thusala: Not a no-op — the container has an explicit inline `max-height`.

index.esm.js:9435-9438 sets `maxHeight: document.documentElement.scrollHeight - popperPosition.top - viewportPadding` on the same div, plus a base `overflow-hidden`. `overflow-y-auto` overrides the `overflow-y` axis, turning that cap from clipping into scrolling. That clip is the ~33px the bug report describes.

The suggested `max-h-[calc(100vh-2rem)]` would be dead code: an inline style beats a Tailwind class, so the library's `maxHeight` wins regardless.

Real residual point: the cap is derived from document height, not viewport height. Under `SmallModal` the body is `position: fixed` (index.esm.js:1584-1586), so document height collapses to roughly the viewport and the two nearly coincide. Not fixable from the consumer side — it needs a `viewportPadding`/max-height change in skapp-ui.

JSDoc reworded in 9bdc750 to stop implying scroll is a last-resort branch.

frontend/src/community/common/constants/commonConstants.ts
thusala: Measuring the real height isn't available at the point the decision is made: placement is computed in `onOpenChange` before the popper mounts, so there is no node to measure. A `ResizeObserver` pass would mean placing once, measuring, then re-placing — a visible reposition on every open. Left as-is for a targeted fix.

frontend/src/community/common/hooks/useCalendarPopperPlacement.ts
thusala: Checked `isFlip` against the implementation — it does not cover this case.

index.esm.js:9313-9320:
```js
const documentHeight = document.documentElement.scrollHeight;
if (pos.top < 0) { pos.top = anchorRect.bottom + offset + scrollY; }
else if (pos.top + popperRect.height > documentHeight) {
  pos.top = anchorRect.top - popperRect.height - offset + scrollY;
}
```

Two gaps:
1. It measures against **document** height, not `window.innerHeight` — so it detects document overflow, not the viewport overflow this bug is about.
2. It flips once and never re-checks. The flipped `pos.top` can be negative (off the top) with no pinning — exactly the third branch this hook adds.

`viewportPadding` doesn't help either: it only clamps `left` (index.esm.js:9331-9340), never `top`. There is no vertical viewport containment in the Popper.

Agreed the durable fix belongs upstream in skapp-ui. Noting the reasoning here rather than in the JSDoc to keep the diff small.

frontend/src/community/common/hooks/useCalendarPopperPlacement.ts
thusala: Out of scope — generalising the hook and migrating `ChipAutocomplete` is a refactor across unrelated components, not part of this bug fix.

Worth noting the thresholds aren't really comparable: `InputDate` (300) and `ChipAutocomplete` (220) are MUI Popper sites estimating their own content heights, while this one is 340+8+16 against a skapp-ui Popper with different placement semantics. Consolidating is a separate task, ideally after the vertical containment lands upstream in skapp-ui.

frontend/src/community/common/hooks/useCalendarPopperPlacement.ts
thusala: Confirmed and fixed in d87d738 — and it's worse than "stale measurements".

`Popper` does re-run `calculatePosition()` on resize/scroll while open (index.esm.js:9358-9375, plus a `ResizeObserver` on the anchor), but it reuses the `offset` it was handed. In the pinned branch that offset is an absolute translation derived from the trigger's *old* rect, so the library actively re-applies it against a moved trigger — drifting further off-screen rather than just misaligning.

Added a `resize` listener while `isCalendarOpen` that re-runs `getPlacement`, cleaned up on teardown. Skipped a scroll listener: `SmallModal` locks background scroll (`body { position: fixed; overflow: hidden }`, index.esm.js:1584-1586), so there's no scroll to track here.

frontend/src/community/common/hooks/useCalendarPopperPlacement.ts
thusala: Verified — the offset is applied unclamped, so the pinned branch works.

index.esm.js:9259: `pos.top = anchorRect.bottom + offset + scrollY`

Plain signed addition, no `Math.max`. With `offset = pinnedTop - bottom` that resolves to `pinnedTop`, which is the intent. Nothing clamps it, and `isFlip` is off so no collision logic competes.

Fair that it reads as an abuse of a prop documented as a "distance"; it is the only lever the Popper exposes for absolute vertical placement.

frontend/src/community/leave/components/molecules/AssignLeavePolicyModal/AssignLeavePolicyForm.tsx
thusala: Out of scope for this PR — `SetJoinDateModal` is a separate component and changing it means re-testing the join-date flow, which this PR doesn't otherwise touch.

Agreed it's a real instance of the same bug, and correct that it's reachable from this flow. Tracking as a follow-up rather than widening this fix.

frontend/src/community/leave/components/molecules/AssignLeavePolicyModal/AssignLeavePolicyForm.tsx
thusala: Checked: the popper's click-away closes it first. `SmallModal` content stays mounted, and the radio is outside the popper, so `handleClose` fires and `onOpenChange(false)` resets `isCalendarOpen` before the DatePicker unmounts.

The stale-placement concern is also moot — placement is recomputed on every open (`onCalendarOpenChange` calls `getPlacement` whenever `isOpen` is true), so a remount measures fresh. No change.

frontend/src/community/common/hooks/useCalendarPopperPlacement.ts
thusala: No need

frontend/src/community/common/hooks/useCalendarPopperPlacement.ts
thusala: Not correct — `containerClassName` merges, and the surface classes aren't on it.

`Popper` (dist/esm/index.esm.js:9435): `` className: `flex flex-col overflow-hidden ${containerClassName}` `` — appended, not replaced.

The background/border/shadow live on **`wrapperClassName`**, which `DatePicker` sets itself (index.esm.js:9468): `` `min-w-max rounded-lg border border-secondary-accent bg-white shadow-lg p-2 ${popperProps.wrapperClassName || ""}` ``.

So `containerClassName: "overflow-y-auto"` cannot strip the surface — it's a different element. No change.

frontend/src/community/common/hooks/useCalendarPopperPlacement.ts
thusala: Agreed — fixed in 9bdc750.

```ts
import { type PopperProps } from "@rootcodelabs/skapp-ui";

type CalendarPopperPlacement = Required<Pick<PopperProps, "position" | "offset">>;
// return type:
Pick<PopperProps, "position" | "offset" | "containerClassName">
```

Matches the `TableView.tsx:3,48` precedent, and a rename/removal upstream now fails compilation instead of silently dropping the prop. `tsc --noEmit` and eslint clean.

frontend/src/community/leave/components/molecules/AssignLeavePolicyModal/AssignLeavePolicyForm.tsx
thusala: Both wirings are required — this is the documented contract, not a duplicate.

index.esm.js:9458: `...(externalTriggerRef ? {} : { ref: internalTriggerRef })`

When you pass `triggerRef`, `DatePicker` deliberately stops attaching its own ref ("the DatePicker won't manage its own internal ref" in the d.ts) and reads `triggerRef.current` for `anchorEl` (index.esm.js:9468). So the consumer must attach it.

Dropping either one breaks it: without `ref={triggerRef}` the ref stays null and `anchorEl: null` means the Popper renders nothing ("When null, popper will not render"). No change.

frontend/src/community/leave/components/molecules/AssignLeavePolicyModal/AssignLeavePolicyForm.tsx
thusala: The library already closes it. `DatePicker` wraps `onDayClick` (index.esm.js:9470-9474):

```js
onDayClick: (day, modifiers, e) => {
  if (dayPickerProps.mode === "single") handleClose();
  dayPickerProps.onDayClick?.(day, modifiers, e);
}
```

`handleClose` calls `internalOnOpenChange(false)`, which is our `onCalendarOpenChange`. We pass `mode="single"` (AssignLeavePolicyForm.tsx:161), so selection closes the calendar. No change.

frontend/src/community/leave/components/molecules/AssignLeavePolicyModal/AssignLeavePolicyForm.tsx
thusala: Already handled by the library. `DatePicker` clones the trigger child with these props (index.esm.js:9459-9465):

```js
"aria-expanded": isOpen,
"aria-haspopup": "dialog",
"aria-controls": isOpen ? calendarId : undefined,
```

They land on our `<div>` trigger automatically, wired to the same `isOpen` we drive. Adding them by hand would duplicate and could conflict with `aria-controls`. No change.

frontend/src/community/common/hooks/useCalendarPopperPlacement.ts
thusala: The month/year overlay risk doesn't apply here — `DatePicker` hardcodes `navLayout` and renders no year/month picker views (index.esm.js:9471-9491); the caption and chevrons are in normal flow, not an absolutely-positioned overlay. There's no `views` prop to opt into, unlike the MUI `InputDate`.

On scoping it to the pinned branch only: that would reintroduce the original clipping, since the library's document-height cap plus `overflow-hidden` applies in the below/above branches too. Explained in the reworded JSDoc (9bdc750). Keeping it unconditional.

frontend/src/community/common/hooks/useCalendarPopperPlacement.ts
thusala: Right, the doc oversold it. Reworded in 9bdc750 to state that `overflow-y-auto` is applied on every open and why, instead of implying a last-resort scroll branch.

Keeping it unconditional is deliberate: the Popper caps container height off document height on every open and clips with `overflow-hidden` (index.esm.js:9435-9438), so the below/above branches can clip too — that cap is the original ~33px symptom. Scoping the class to the pinned branch would reintroduce it.

frontend/src/community/common/constants/commonConstants.ts
thusala: Reasonable, but leaving as-is: `commonConstants.ts` was just touched in 1abe1ba and moving these back into the hook is churn on a bug fix. Happy to fold in if you'd rather.

frontend/src/community/leave/components/molecules/AssignLeavePolicyModal/AssignLeavePolicyForm.tsx
thusala: Out of scope — extracting a `SpecificDateField` molecule is a refactor of a file this PR only adds 12 lines to. Happy to do it separately.

frontend/src/community/common/hooks/useCalendarPopperPlacement.ts
thusala: Extracted in d00b3e9 — `getPlacement` is now `getCalendarPopperPlacement` in `~community/common/utils/calendarPopperUtils.ts`, with `CalendarPopperPlacement` moved to `CommonTypes` (utils here don't export types inline). The hook is down to state + the resize listener; the geometry is a pure function, directly assertable without `renderHook`.

frontend/src/community/common/hooks/useCalendarPopperPlacement.ts
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * Keeps a skapp-ui DatePicker calendar reachable inside a modal by measuring
+ * the trigger against the viewport on every open. See
+ * getCalendarPopperPlacement for the placement rules.
+ *
+ * The Popper container caps its own height off the document height, not the
+ * viewport, and clips with overflow-hidden - so overflow-y-auto is applied on
+ * every open, not just the pinned case, to make whatever it caps scrollable.
+ */
```
hasalarootcode: remove comments

frontend/src/community/common/hooks/useCalendarPopperPlacement.ts
```diff
@@ -0,0 +1,77 @@
...
+      useState<CalendarPopperPlacement>(BELOW_TRIGGER);
+
+    const onCalendarOpenChange = useCallback((isOpen: boolean): void => {
+      if (isOpen) {
+        setPlacement(getCalendarPopperPlacement(triggerRef.current));
+      }
+      setIsCalendarOpen(isOpen);
+    }, []);
+
+    // The Popper re-runs its own positioning on resize but reuses whatever
```
hasalarootcode: remove comments

frontend/src/community/common/utils/calendarPopperUtils.ts
```diff
@@ -0,0 +1,57 @@
...
+};
+
+/**
+ * Picks the side of the trigger a skapp-ui DatePicker calendar fits on.
+ *
+ * The Popper always draws the calendar below its trigger, so a trigger that
+ * sits low in a vertically centred modal pushes the last week rows past the
+ * bottom of the viewport - where nothing can scroll them into view, because
+ * the modal locks background scrolling. Prefers below, flips above when below
+ * would overflow, and pins the calendar inside the viewport when neither side
```
hasalarootcode: remove comments

hasalarootcode (changes requested):

frontend/src/community/common/hooks/useCalendarPopperPlacement.ts
thusala: done

frontend/src/community/common/hooks/useCalendarPopperPlacement.ts
thusala: done

frontend/src/community/common/utils/calendarPopperUtils.ts
thusala: done

hasalarootcode (approved):

PR #1838, opened 2026-08-25
fix: Block clock-in for pending leave requests when "Clock in on leave days" is disabled

SanithuGunasena07:
Fixes clock-in bypass for pending leave requests. The backend query behind the "Clock in on leave days" restriction only checked for APPROVED leave, so employees with a pending request could still clock in via the warning modal. Now it checks both PENDING and APPROVED, matching the leave-availability check already used for the warning modal, so the clock-in button is correctly disabled with a tooltip in both cases.

### DIFF
--- a/backend/src/main/java/com/skapp/community/leaveplanner/repository/LeaveRequestRepository.java
+++ b/backend/src/main/java/com/skapp/community/leaveplanner/repository/LeaveRequestRepository.java
@@ -5,6 +5,7 @@ import com.skapp.community.leaveplanner.model.LeaveRequest;
 import com.skapp.community.leaveplanner.payload.EmployeeLeaveHistoryFilterDto;
 import com.skapp.community.leaveplanner.payload.LeaveRequestFilterDto;
 import com.skapp.community.leaveplanner.payload.TeamLeaveHistoryFilterDto;
+import com.skapp.community.leaveplanner.type.LeaveRequestStatus;
 import com.skapp.community.leaveplanner.payload.request.EmployeesOnLeavePeriodFilterDto;
 import com.skapp.community.leaveplanner.payload.response.EmployeeLeaveRequestReportExportDto;
 import com.skapp.community.leaveplanner.payload.response.EmployeeLeaveRequestReportQueryDto;
@@ -32,7 +33,8 @@ public interface LeaveRequestRepository {
 
 	List<LeaveRequest> findAllLeaveRequestsByDateRange(LeaveRequestFilterDto leaveRequestFilterDto);
 
-	List<LeaveRequest> findLeaveRequestsForTodayByUser(LocalDate currentDate, Long employeeId);
+	List<LeaveRequest> findLeaveRequestsForTodayByUser(LocalDate currentDate, Long employeeId,
+			List<LeaveRequestStatus> statuses);
 
 	List<LeaveRequest> findLeaveRequestAvailabilityForGivenDate(LocalDate date, Long employeeId);
 
--- a/backend/src/main/java/com/skapp/community/leaveplanner/repository/LeaveRequestRepository.java
+++ b/backend/src/main/java/com/skapp/community/leaveplanner/repository/LeaveRequestRepository.java
@@ -5,7 +5,6 @@ import com.skapp.community.leaveplanner.model.LeaveRequest;
 import com.skapp.community.leaveplanner.payload.EmployeeLeaveHistoryFilterDto;
 import com.skapp.community.leaveplanner.payload.LeaveRequestFilterDto;
 import com.skapp.community.leaveplanner.payload.TeamLeaveHistoryFilterDto;
-import com.skapp.community.leaveplanner.type.LeaveRequestStatus;
 import com.skapp.community.leaveplanner.payload.request.EmployeesOnLeavePeriodFilterDto;
 import com.skapp.community.leaveplanner.payload.response.EmployeeLeaveRequestReportExportDto;
 import com.skapp.community.leaveplanner.payload.response.EmployeeLeaveRequestReportQueryDto;
@@ -33,8 +32,9 @@ public interface LeaveRequestRepository {
 
 	List<LeaveRequest> findAllLeaveRequestsByDateRange(LeaveRequestFilterDto leaveRequestFilterDto);
 
-	List<LeaveRequest> findLeaveRequestsForTodayByUser(LocalDate currentDate, Long employeeId,
-			List<LeaveRequestStatus> statuses);
+	List<LeaveRequest> findLeaveRequestsForTodayByUser(LocalDate currentDate, Long employeeId);
+
+	List<LeaveRequest> findPendingAndApprovedLeaveRequestsForTodayByUser(LocalDate currentDate, Long employeeId);
 
 	List<LeaveRequest> findLeaveRequestAvailabilityForGivenDate(LocalDate date, Long employeeId);
 
--- a/backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/LeaveRequestRepositoryImpl.java
+++ b/backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/LeaveRequestRepositoryImpl.java
@@ -334,7 +334,8 @@ public class LeaveRequestRepositoryImpl implements LeaveRequestRepository {
 	}
 
 	@Override
-	public List<LeaveRequest> findLeaveRequestsForTodayByUser(LocalDate currentDate, Long employeeId) {
+	public List<LeaveRequest> findLeaveRequestsForTodayByUser(LocalDate currentDate, Long employeeId,
+			List<LeaveRequestStatus> statuses) {
 		CriteriaBuilder criteriaBuilder = entityManager.getCriteriaBuilder();
 		CriteriaQuery<LeaveRequest> criteriaQuery = criteriaBuilder.createQuery(LeaveRequest.class);
 		Root<LeaveRequest> leaveRequest = criteriaQuery.from(LeaveRequest.class);
@@ -342,9 +343,7 @@ public class LeaveRequestRepositoryImpl implements LeaveRequestRepository {
 		Join<LeaveRequest, Employee> employee = leaveRequest.join(LeaveRequest_.employee);
 
 		Predicate employeePredicate = criteriaBuilder.equal(employee.get(Employee_.employeeId), employeeId);
-		Predicate statusPredicate = criteriaBuilder.or(
-				criteriaBuilder.equal(leaveRequest.get(LeaveRequest_.status), LeaveRequestStatus.PENDING),
-				criteriaBuilder.equal(leaveRequest.get(LeaveRequest_.status), LeaveRequestStatus.APPROVED));
+		Predicate statusPredicate = leaveRequest.get(LeaveRequest_.status).in(statuses);
 
 		Predicate datePredicate = criteriaBuilder.between(criteriaBuilder.literal(currentDate),
 				leaveRequest.get(LeaveRequest_.startDate), leaveRequest.get(LeaveRequest_.endDate));
--- a/backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/LeaveRequestRepositoryImpl.java
+++ b/backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/LeaveRequestRepositoryImpl.java
@@ -334,8 +334,7 @@ public class LeaveRequestRepositoryImpl implements LeaveRequestRepository {
 	}
 
 	@Override
-	public List<LeaveRequest> findLeaveRequestsForTodayByUser(LocalDate currentDate, Long employeeId,
-			List<LeaveRequestStatus> statuses) {
+	public List<LeaveRequest> findLeaveRequestsForTodayByUser(LocalDate currentDate, Long employeeId) {
 		CriteriaBuilder criteriaBuilder = entityManager.getCriteriaBuilder();
 		CriteriaQuery<LeaveRequest> criteriaQuery = criteriaBuilder.createQuery(LeaveRequest.class);
 		Root<LeaveRequest> leaveRequest = criteriaQuery.from(LeaveRequest.class);
@@ -343,7 +342,31 @@ public class LeaveRequestRepositoryImpl implements LeaveRequestRepository {
 		Join<LeaveRequest, Employee> employee = leaveRequest.join(LeaveRequest_.employee);
 
 		Predicate employeePredicate = criteriaBuilder.equal(employee.get(Employee_.employeeId), employeeId);
-		Predicate statusPredicate = leaveRequest.get(LeaveRequest_.status).in(statuses);
+		Predicate statusPredicate = criteriaBuilder.equal(leaveRequest.get(LeaveRequest_.status),
+				LeaveRequestStatus.APPROVED);
+
+		Predicate datePredicate = criteriaBuilder.between(criteriaBuilder.literal(currentDate),
+				leaveRequest.get(LeaveRequest_.startDate), leaveRequest.get(LeaveRequest_.endDate));
+
+		criteriaQuery.where(criteriaBuilder.and(employeePredicate, statusPredicate, datePredicate));
+
+		TypedQuery<LeaveRequest> query = entityManager.createQuery(criteriaQuery);
+
+		return query.getResultList();
+	}
+
+	@Override
+	public List<LeaveRequest> findPendingAndApprovedLeaveRequestsForTodayByUser(LocalDate currentDate,
+			Long employeeId) {
+		CriteriaBuilder criteriaBuilder = entityManager.getCriteriaBuilder();
+		CriteriaQuery<LeaveRequest> criteriaQuery = criteriaBuilder.createQuery(LeaveRequest.class);
+		Root<LeaveRequest> leaveRequest = criteriaQuery.from(LeaveRequest.class);
+
+		Join<LeaveRequest, Employee> employee = leaveRequest.join(LeaveRequest_.employee);
+
+		Predicate employeePredicate = criteriaBuilder.equal(employee.get(Employee_.employeeId), employeeId);
+		Predicate statusPredicate = leaveRequest.get(LeaveRequest_.status)
+			.in(LeaveRequestStatus.PENDING, LeaveRequestStatus.APPROVED);
 
 		Predicate datePredicate = criteriaBuilder.between(criteriaBuilder.literal(currentDate),
 				leaveRequest.get(LeaveRequest_.startDate), leaveRequest.get(LeaveRequest_.endDate));
--- a/backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/LeaveRequestRepositoryImpl.java
+++ b/backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/LeaveRequestRepositoryImpl.java
@@ -342,8 +342,9 @@ public class LeaveRequestRepositoryImpl implements LeaveRequestRepository {
 		Join<LeaveRequest, Employee> employee = leaveRequest.join(LeaveRequest_.employee);
 
 		Predicate employeePredicate = criteriaBuilder.equal(employee.get(Employee_.employeeId), employeeId);
-		Predicate statusPredicate = criteriaBuilder.equal(leaveRequest.get(LeaveRequest_.status),
-				LeaveRequestStatus.APPROVED);
+		Predicate statusPredicate = criteriaBuilder.or(
+				criteriaBuilder.equal(leaveRequest.get(LeaveRequest_.status), LeaveRequestStatus.PENDING),
+				criteriaBuilder.equal(leaveRequest.get(LeaveRequest_.status), LeaveRequestStatus.APPROVED));
 
 		Predicate datePredicate = criteriaBuilder.between(criteriaBuilder.literal(currentDate),
 				leaveRequest.get(LeaveRequest_.startDate), leaveRequest.get(LeaveRequest_.endDate));
--- a/backend/src/main/java/com/skapp/community/timeplanner/payload/response/ActiveTimeSlotResponseDto.java
+++ b/backend/src/main/java/com/skapp/community/timeplanner/payload/response/ActiveTimeSlotResponseDto.java
@@ -12,6 +12,8 @@ public class ActiveTimeSlotResponseDto {
 
 	private TimeRecordActionTypes periodType;
 
+	private boolean leavePending;
+
 	private LocalDateTime starTime;
 
 	private float workHours;
--- a/backend/src/main/java/com/skapp/community/timeplanner/payload/response/ActiveTimeSlotResponseDto.java
+++ b/backend/src/main/java/com/skapp/community/timeplanner/payload/response/ActiveTimeSlotResponseDto.java
@@ -12,7 +12,7 @@ public class ActiveTimeSlotResponseDto {
 
 	private TimeRecordActionTypes periodType;
 
-	private boolean leavePending;
+	private Boolean isLeavePending;
 
 	private LocalDateTime starTime;
 
--- a/frontend/src/community/attendance/store/slices/attendanceSlice.ts
+++ b/frontend/src/community/attendance/store/slices/attendanceSlice.ts
@@ -11,7 +11,7 @@ const attendanceParams: attendanceStatusTypes = {
   slotStartTime: null,
   breakHours: null,
   workHours: null,
-  leavePending: false
+  isLeavePending: false
 };
 
 const attendanceLeaveStatus: attendanceLeaveStatusTypes = {
--- a/frontend/src/community/attendance/store/slices/attendanceSlice.ts
+++ b/frontend/src/community/attendance/store/slices/attendanceSlice.ts
@@ -10,7 +10,8 @@ const attendanceParams: attendanceStatusTypes = {
   slotType: AttendanceSlotType.READY,
   slotStartTime: null,
   breakHours: null,
-  workHours: null
+  workHours: null,
+  leavePending: false
 };
 
 const attendanceLeaveStatus: attendanceLeaveStatusTypes = {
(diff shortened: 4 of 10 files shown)


ThinuwanW (commented):
## 🤖 Claude Code Review

This one-line change widens `findLeaveRequestsForTodayByUser` from APPROVED-only to PENDING+APPROVED so that a pending leave request blocks the clock-in UI. The predicate itself is correct and matches the existing idiom used by the sibling method `findLeaveRequestAvailabilityForGivenDate`. However, the method is shared by two unrelated consumers, and the second one — the admin/manager clock-in summary report in `TimeAnalyticsServiceImpl` — silently changes behavior as collateral damage. Separately, the stated goal ("block clock-in") is not actually enforced server-side: the only leave gating lives in the read-only `getActiveTimeSlot` endpoint, while the `addTimeRecord` write path has no leave check at all.

Found **13** new issue(s): 🔴 4 important, 🟡 6 suggestion(s), 🟣 3 nit(s)

backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/LeaveRequestRepositoryImpl.java
```diff
@@ -342,8 +342,9 @@ public List<LeaveRequest> findLeaveRequestsForTodayByUser(LocalDate currentDate,
 		Join<LeaveRequest, Employee> employee = leaveRequest.join(LeaveRequest_.employee);
 
 		Predicate employeePredicate = criteriaBuilder.equal(employee.get(Employee_.employeeId), employeeId);
-		Predicate statusPredicate = criteriaBuilder.equal(leaveRequest.get(LeaveRequest_.status),
-				LeaveRequestStatus.APPROVED);
+		Predicate statusPredicate = criteriaBuilder.or(
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

This repository method has two callers, and only one of them wants the new PENDING behavior.

Caller 1 — `TimeServiceImpl.checkLeaveOrHolidayOrNonWorkingDay()` (line 1202): this is the intended target. Fine.

Caller 2 — `TimeAnalyticsServiceImpl.buildClockInSummaryResponse()` (line 279): this is the admin/manager *Clock-In Summary* report. At line 285 it does `if (!clockInOnLeaveDaysStatus && !leaveRequestsList.isEmpty()) return Optional.empty();` — the employee is dropped from the report entirely. After this change, any employee who merely *submitted* a leave request for today (not yet approved, possibly about to be denied) silently disappears from the manager's clock-in/late-arrival/not-clocked-in report.

That is backwards for the report's purpose: an employee with an unapproved request is still expected at work, and is exactly the person a manager needs to see. It also means an employee can remove themselves from attendance oversight just by filing a leave request that never gets approved.

Fix: don't overload one method for two different policies. Either add a `List<LeaveRequestStatus> statuses` (or varargs) parameter to `findLeaveRequestsForTodayByUser` and have `TimeServiceImpl` pass `PENDING, APPROVED` while `TimeAnalyticsServiceImpl` keeps passing `APPROVED` only, or leave this method as-is and add a separate `findPendingAndApprovedLeaveRequestsForDate(...)` used solely by the clock-in gating path. If excluding pending-leave employees from the report *is* intended, please say so in the PR description, because it is not implied by the commit message.

ThinuwanW: 🤖 **Claude** · 🔴 **Important**

📁 **File-level observation**

The PR is titled "Block clock-in for pending leave requests", but nothing is actually blocked on the server.

The only place leave requests gate clock-in is `TimeServiceImpl.checkLeaveOrHolidayOrNonWorkingDay()`, which is called exclusively from `getActiveTimeSlot()` (line 243) — a `@Transactional(readOnly = true)` GET that just returns `periodType = LEAVE_DAY` so the frontend hides the clock-in button.

The actual write path, `TimeServiceImpl.addTimeRecord()` (line 519) → `recordClockInAndClockOut()` (line 1976), performs no leave-day, holiday, or non-working-day validation whatsoever. It only checks that the date is not in the future and that a record does not already exist. A client that POSTs directly to the add-time-record endpoint can still clock in on a pending *or* approved leave day, regardless of the `CLOCK_IN_ON_LEAVE_DAYS` config.

This gap predates the PR, but the PR's stated objective depends on it. Consider extracting the leave/holiday/non-working-day evaluation into a shared helper and invoking it from `recordClockInAndClockOut` for the `START` action, throwing a `ModuleException` with a new `TimeMessageConstant` key (and a matching entry in `time-messages.properties`) when clock-in is disallowed. Otherwise this fix is cosmetic and the server remains authoritative-in-name-only.

ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Pre-existing date inconsistency in `TimeAnalyticsServiceImpl.buildClockInSummaryResponse()` that this change makes materially worse.

Line 280 looks up leave with `DateTimeUtils.getCurrentUtcDate()`, while every other lookup in the same method uses the caller-supplied `filterDto.getDate()` (line 289 `findByDateAndEmployee`, line 325 `getHoliday`, line 326 `getLeaveRequest`). So when an admin pulls the clock-in summary for a *past* date, the leave-based exclusion at line 285 is evaluated against *today's* leave requests, not the requested date's.

Before this PR that mismatch only fired on approved leave. Now a brand-new pending request filed today can retroactively erase an employee from a historical attendance report. Change line 280 to pass `filterDto.getDate()` so the exclusion is consistent with the rest of the method.

ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

No test coverage accompanies this behavior change. Grepping the test sources for `findLeaveRequestsForTodayByUser`, `getAllActiveSlotsNoLeaveDay`, and `buildClockInSummaryResponse` returns zero matches — none of the three have any existing tests, so there is nothing asserting the old APPROVED-only contract and nothing that will assert the new one.

Given the method feeds two consumers with divergent requirements, please add at minimum: (a) a repository-level test confirming a PENDING request for the given date is returned and DENIED/CANCELLED/REVOKED are not, and (b) a service-level test on `getActiveTimeSlot()` asserting `periodType == LEAVE_DAY` when the only leave request is PENDING and `CLOCK_IN_ON_LEAVE_DAYS` is false. If you adopt the status-parameter split suggested above, a `clockInSummary` test pinning the APPROVED-only behavior would guard against future regressions.

ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

The method name `findLeaveRequestsForTodayByUser` no longer describes what it returns. It was already slightly off (the date is a parameter, not necessarily "today" — `TimeAnalyticsServiceImpl` and `TimeServiceImpl` both happen to pass the current date, but nothing enforces that), and now the returned set spans two statuses rather than just approved leave.

A reader of the two call sites sees `List<LeaveRequest> leaveRequestsList` and will reasonably assume these are confirmed leave days. Consider renaming to something explicit such as `findPendingAndApprovedLeaveRequestsByEmployeeAndDate(LocalDate date, Long employeeId)`, and renaming the local variables at `TimeServiceImpl:1202` and `TimeAnalyticsServiceImpl:279` accordingly. This is worth doing regardless of how the shared-method issue is resolved.

backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/LeaveRequestRepositoryImpl.java
```diff
@@ -342,8 +342,9 @@ public List<LeaveRequest> findLeaveRequestsForTodayByUser(LocalDate currentDate,
...
+				criteriaBuilder.equal(leaveRequest.get(LeaveRequest_.status), LeaveRequestStatus.PENDING),
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Minor consistency point: `findByEmployeeAndDate` in this same file (line 842) expresses the identical condition more concisely as `leaveRequestRoot.get(LeaveRequest_.status).in(LeaveRequestStatus.APPROVED, LeaveRequestStatus.PENDING)`. Using `.in(...)` here instead of `criteriaBuilder.or(equal, equal)` would match that and read a little better.

Note this is genuinely a toss-up — the sibling method directly below, `findLeaveRequestAvailabilityForGivenDate` (lines 370-372), uses the same `criteriaBuilder.or(...)` form you've written, so the file is already inconsistent with itself. Either form is defensible; take it or leave it.

backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/LeaveRequestRepositoryImpl.java
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Mid-shift clock-out lockout, self-triggerable by the employee. Trace the consumer: `TimeServiceImpl.getActiveTimeSlot()` calls `checkLeaveOrHolidayOrNonWorkingDay()` at line 243 and returns its result immediately (line 244-246) — *before* the existing `timeRecordDao.findByEmployeeAndDate(...)` lookup at line 249. Inside that helper, this query's result feeds `getAllActiveSlotsNoLeaveDay(...)`, which returns periodType=LEAVE_DAY. Now that PENDING counts, the sequence is: employee clocks in at 09:00 -> at 13:00 applies for leave covering today (`LeaveServiceImpl.applyLeaveRequest` sets status PENDING at line 193 and `validateLeaveRequest` at line 775 only checks date-range ordering, entitlement and overlap — nothing forbids today or a past date) -> the very next poll of the active-slot endpoint returns LEAVE_DAY. On the client, `ClockWidget.tsx` (isDisabled at lines 63-73, showTimer at lines 91-97) hides the running timer and disables the button, and `PlayButton.tsx` (getDisableStatus, lines 64-75) disables pause/resume. There is no END path for LEAVE_DAY, so the employee cannot clock out; the record stays open and later surfaces via `getIncompleteClockOuts`. Note the privilege inversion: previously only a manager approving leave could produce LEAVE_DAY; now any employee can put themselves into it. A half-day-evening pending request has the same effect from the evening boundary onward via `TimeUtil.isCurrentTimeInEvening`. Fix: make the leave short-circuit conditional on there being no open time record for today — e.g. move the `timeRecordDao.findByEmployeeAndDate` lookup above the `checkLeaveOrHolidayOrNonWorkingDay()` call and skip the LEAVE_DAY branch when a record exists with a null clockOutTime — or restrict the PENDING half of this predicate to requests created before the clock-in time. Either way, add a service-layer guard so an in-progress shift can always be closed.

ThinuwanW: 🤖 **Claude** · 🔴 **Important**

📁 **File-level observation**

This change makes the method an exact functional duplicate of one that already exists in the same class. Compare with `findLeaveRequestAvailabilityForGivenDate` (impl at lines 360-380): identical employee-id equality on the joined Employee, identical inclusive containment of the parameter date between startDate and endDate (expressed as lessThanOrEqualTo/greaterThanOrEqualTo instead of between-on-a-literal, semantically the same), and now an identical PENDING-or-APPROVED status disjunction — even written with the same or-of-two-equals shape. The two are declared on adjacent lines of `LeaveRequestRepository` (lines 35 and 37), so the duplication is hard to miss once the status sets converge. Two identically-behaving queries with different names will drift the next time only one of them is edited. Fix: delete `findLeaveRequestsForTodayByUser` from the interface and impl and point both callers (`TimeServiceImpl:1202`, `TimeAnalyticsServiceImpl:280`) at `findLeaveRequestAvailabilityForGivenDate`, which already has three other callers and a name that does not falsely imply 'today'.

ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Missing paired frontend change — the user-facing copy is now factually wrong. The backend collapses PENDING and APPROVED into the same `TimeRecordActionTypes.LEAVE_DAY` periodType, and the client renders a single fixed string for it: `ClockWidget.tsx:85` resolves `notAllowedToClockInOnLeaveDaysTooltip`, defined in `frontend/src/community/common/assets/languages/english/attendanceModule.json:245` as 'You are not allowed to clock in on leave days'. An employee whose request is still awaiting manager approval will be told they are on a leave day, with no hint that the block will lift if the request is denied — a predictable support-ticket generator for a change whose entire purpose is this new state. Because the response carries no status information, the UI cannot distinguish the two cases even if it wanted to. Fix: either add a distinct periodType for the pending case, or surface the leave status on `ActiveTimeSlotResponseDto` so the widget can select between the existing string and a new pending-specific one (e.g. 'Clock-in is blocked while your leave request for today is awaiting approval'). Also add the corresponding key to the other language bundles under the same directory.

ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

The two consumers of this query interpret its result inconsistently, and widening the result set amplifies the weaker one. `TimeServiceImpl.getAllActiveSlotsNoLeaveDay` (lines 1257-1271) is LeaveState-aware: it only reports LEAVE_DAY for FULLDAY, or for HALFDAY_MORNING/HALFDAY_EVENING when the current time actually falls in that half. `TimeAnalyticsServiceImpl.buildClockInSummaryResponse` (line 285) just tests `!leaveRequestsList.isEmpty()` and returns `Optional.empty()`, dropping the employee from the clock-in summary entirely. So a half-day request — and now a half-day request that no manager has approved — erases an employee who is expected to work the other half of the day from the admin's attendance list, hiding both their real clock-in time and any late arrival. Note also that the list can now legitimately contain two rows for one date (e.g. an approved morning half-day plus a pending evening half-day), which makes the isEmpty() test even blunter. Fix: extract the LeaveState/time-of-day predicate used in `getAllActiveSlotsNoLeaveDay` into a shared helper and apply it in the analytics path too, or at minimum only suppress the row when a FULLDAY request is present.

ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

N+1 query pattern that this change makes more expensive. `TimeAnalyticsServiceImpl.clockInSummary` (lines 209-213) maps over every employee returned by `employeeTeamDao.getEmployeesByTeamIds` and calls `buildClockInSummaryResponse`, which executes this query once per employee (line 279-280) — plus a time-record lookup, plus `getLeaveRequest`/`isLateArrival`, each of which issues its own leave query. For an org-wide summary (teams defaulted to -1L at line 201, i.e. all employees) that is hundreds of round-trips, and each now matches a strictly larger row set. This class already has the right shape to fix it: `findLeaveRequestsByDateRangeAndEmployees` (line 383) takes a `List<Long> employeeIds`. Add an equivalent single-date/multi-employee variant, call it once in `clockInSummary`, and group the result by employeeId before the stream.

backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/LeaveRequestRepositoryImpl.java
```diff
@@ -342,8 +342,9 @@ public List<LeaveRequest> findLeaveRequestsForTodayByUser(LocalDate currentDate,
...
+				criteriaBuilder.equal(leaveRequest.get(LeaveRequest_.status), LeaveRequestStatus.APPROVED));
 
 		Predicate datePredicate = criteriaBuilder.between(criteriaBuilder.literal(currentDate),
```
ThinuwanW, line 371: 🤖 **Claude** · 🟣 **Nit**

Date/timezone inconsistency in the gating this change now leans on more heavily. `TimeServiceImpl.checkLeaveOrHolidayOrNonWorkingDay` passes `DateTimeUtils.getCurrentUtcDate()` as the date matched against startDate/endDate here (line 1174, 1202), but the surrounding logic mixes zones: the working-day check at line 1177 is explicitly passed `organizationService.getOrganizationTimeZone()`, while the half-day boundary checks in `TimeUtil.isCurrentTimeInMorning`/`isCurrentTimeInEvening` (TimeUtil.java:41, 48) use bare `LocalTime.now()` — the JVM default zone, neither UTC nor the org timezone. For an organization far from UTC, near midnight the leave lookup can be performed for the wrong calendar day while the half-day window is evaluated against a third clock. This is pre-existing, but including PENDING roughly doubles how often the branch is reached, so the wrong-day cases become visible. Consider resolving both the date and the time-of-day from the organization timezone consistently.

backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/LeaveRequestRepositoryImpl.java
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

This adds a 17th hand-written copy of the 'PENDING or APPROVED' status set inside this one file (also at lines 371-372, 397, 423, 463, 585, 747, 842, 1073, 1112, 1231, 1287, 1551, 1653, 1748, 1836, 1885), with two different spellings for the same idea. `LeaveRequestStatus` currently has five values, so any future addition that should count as an active request (a cancellation-pending state, for example) would need all 17 sites updated and would silently misbehave at whichever ones are missed. Consider a single `public static final Set<LeaveRequestStatus> ACTIVE_LEAVE_STATUSES = EnumSet.of(PENDING, APPROVED)` (or a constant in the `type`/`constant` package) and using `status.in(ACTIVE_LEAVE_STATUSES)` everywhere, so 'what counts as an active leave request' is defined in exactly one place.

backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/LeaveRequestRepositoryImpl.java
SanithuGunasena07: not required

SanithuGunasena07: Added a List statuses param to findLeaveRequestsForTodayByUser (option 1). TimeServiceImpl now passes PENDING, APPROVED for the clock-in block, and TimeAnalyticsServiceImpl passes APPROVED only, so the report keeps showing employees with unapproved leave requests.

SanithuGunasena07: Pre existing out of scope for PR

SanithuGunasena07: Pre existing behaviour recordClockInAndClockOut() never validated leave/holiday/non-working-day, only the GET status check did. This PR didn't touch the write path, so nothing new is introduced here. Hence out of scope for the PR

SanithuGunasena07: not renaming already existing method names

backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/LeaveRequestRepositoryImpl.java
SanithuGunasena07: Already resolved, switched to .in(statuses) in the fix for the shared-method issue above.

backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/LeaveRequestRepositoryImpl.java
SanithuGunasena07: Pre existing as getActiveTimeSlot() already checked leave/holiday before checking for an open clock-in, before this PR.

SanithuGunasena07: not required

SanithuGunasena07: added isLeavePending to ActiveTimeSlotResponseDto, set it from the matched leave request's status in the backend leave-day check, and threaded it through the store/API on the frontend. ClockWidget.tsx now shows a distinct tooltip (notAllowedToClockInOnPendingLeaveDaysTooltip) when the block is due to a pending request, vs. the existing "leave days" message for approved ones.

SanithuGunasena07: Pre-existing behaviour the manager's report was already just checking "any leave request today?" without looking at whether it's a full day or half day. That check was already there before this PR.

SanithuGunasena07: Pre-existing as the N+1 query pattern in clockInSummary() already existed before this PR

backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/LeaveRequestRepositoryImpl.java
SanithuGunasena07: Pre existing behaviour not something new introduced in this PR

backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/LeaveRequestRepositoryImpl.java
SanithuGunasena07: pre existing

backend/src/main/java/com/skapp/community/leaveplanner/repository/LeaveRequestRepository.java
```diff
@@ -32,7 +33,8 @@ public interface LeaveRequestRepository {
 
 	List<LeaveRequest> findAllLeaveRequestsByDateRange(LeaveRequestFilterDto leaveRequestFilterDto);
 
-	List<LeaveRequest> findLeaveRequestsForTodayByUser(LocalDate currentDate, Long employeeId);
+	List<LeaveRequest> findLeaveRequestsForTodayByUser(LocalDate currentDate, Long employeeId,
+			List<LeaveRequestStatus> statuses);
```
erandi-rootcode: without passing the List<LeaveRequestStatus> statuses, add and filter the statuses in the criteria query itself. if it's used for any other purposes with other stauses, create a separate crtieria query for your requirement keeping the existing one as it is

erandi-rootcode (approved):

erandi-rootcode (changes requested):
check criteria query

backend/src/main/java/com/skapp/community/leaveplanner/repository/LeaveRequestRepository.java
SanithuGunasena07: reverted findLeaveRequestsForTodayByUser to its original APPROVED-only query with no statuses param, and added a separate findPendingAndApprovedLeaveRequestsForTodayByUser for the clock-in check's PENDING+APPROVED need.

erandi-rootcode (dismissed):
Approved BE changes

Deshan1100 (dismissed):
Approved FE

backend/src/main/java/com/skapp/community/timeplanner/payload/response/ActiveTimeSlotResponseDto.java
```diff
@@ -12,6 +12,8 @@ public class ActiveTimeSlotResponseDto {
 
 	private TimeRecordActionTypes periodType;
 
+	private boolean leavePending;
```
hasalarootcode: use Boolean and rename isLeavePending

frontend/src/community/attendance/store/slices/attendanceSlice.ts
```diff
@@ -10,7 +10,8 @@ const attendanceParams: attendanceStatusTypes = {
   slotType: AttendanceSlotType.READY,
   slotStartTime: null,
   breakHours: null,
-  workHours: null
+  workHours: null,
+  leavePending: false
```
hasalarootcode: isLeavePending

frontend/src/community/attendance/types/attendanceTypes.ts
```diff
@@ -29,6 +29,7 @@ export interface attendanceStatusTypes {
   slotStartTime: string | null;
   breakHours: number | null;
   workHours: number | null;
+  leavePending?: boolean;
```
hasalarootcode: here also

hasalarootcode (changes requested):

frontend/src/community/attendance/types/attendanceTypes.ts
SanithuGunasena07: renamed

frontend/src/community/attendance/store/slices/attendanceSlice.ts
SanithuGunasena07: renamed

backend/src/main/java/com/skapp/community/timeplanner/payload/response/ActiveTimeSlotResponseDto.java
SanithuGunasena07: renamed

hasalarootcode (approved):

Deshan1100 (approved):

erandi-rootcode (approved):

ThinuwanW (approved):

PR #1839, opened 2026-08-27
Fix/3029: Leave policies table missing assigned employees column

hasalarootcode:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds an `assignedEmployeeCount` to the leave-policy list/update responses, backed by a new grouped-count Criteria query, and surfaces it as an ASSIGNED EMPLOYEES column in the policies table. The backend query and DTO wiring are structurally sound (batched single query, no N+1, correct null-safe defaults), but the count does not filter out terminated/deleted employees — diverging from every other employee-count query in this codebase — and the single-assign/unassign React Query mutations never invalidate the policy list, so the new column goes stale. Test coverage for the new query only exercises one policy, so a per-policy key-mapping regression would go undetected.

Found **14** new issue(s): 🔴 4 important, 🟡 6 suggestion(s), 🟣 4 nit(s)

backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/EmployeeLeavePolicyRepositoryImpl.java
```diff
@@ -76,6 +81,30 @@ public Page<EmployeeLeavePolicy> findByEmployeeIdAndStatusOrderByEffectiveFromDe
...
+
+		CriteriaBuilder cb = entityManager.getCriteriaBuilder();
+		CriteriaQuery<Tuple> query = cb.createTupleQuery();
+		Root<EmployeeLeavePolicy> root = query.from(EmployeeLeavePolicy.class);
+		Path<Long> policyIdPath = root.get(EmployeeLeavePolicy_.policy).get(LeavePolicy_.id);
+
+		query
+			.multiselect(policyIdPath.alias(LeavePolicyConstant.POLICY_ID_ALIAS),
+					cb.count(root).alias(LeavePolicyConstant.ASSIGNED_COUNT_ALIAS))
+			.where(cb.and(policyIdPath.in(policyIds), cb.equal(root.get(EmployeeLeavePolicy_.status), status)))
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The count filters only on `EmployeeLeavePolicy.status = ACTIVE` and never checks the employee's account status. Terminating or soft-deleting an employee does not end their leave-policy assignments — `PeopleServiceImpl.updateUserStatus`/`terminateUser` only flips `Employee.accountStatus`, and nothing in the people module touches `EmployeeLeavePolicyDao` — so the assignment row stays ACTIVE forever and the new ASSIGNED EMPLOYEES column permanently over-reports after any termination. This contradicts the rest of the codebase: every other leave-module aggregate excludes them (e.g. `LeaveEntitlementRepositoryImpl` line 316, `LeaveRequestRepositoryImpl` lines 457/1003/1041, `PolicyLeaveRequestRepositoryImpl` line 223 all add `not(employee.get(Employee_.ACCOUNT_STATUS).in(AccountStatus.TERMINATED, AccountStatus.DELETED))`), and the directly analogous field `BusinessUnitSummaryResponseDto.assignedEmployeeCount` is computed with `countByBusinessUnitBusinessUnitIdAndAccountStatusIn(id, Set.of(ACTIVE, PENDING))` (`BusinessUnitServiceImpl` line 111). It is also inconsistent within this feature: bulk assign only resolves employees via `employeeDao.findActiveEmployeesByExactNames(...)`, so terminated employees can never be *added* to the count but are never removed from it either. Fix: add an inner join to `Employee` in the tuple query and a predicate excluding `AccountStatus.TERMINATED` and `AccountStatus.DELETED` (matching the existing helper predicates), e.g. `Join<EmployeeLeavePolicy, Employee> employeeJoin = root.join(EmployeeLeavePolicy_.employee); ... cb.not(employeeJoin.get(Employee_.ACCOUNT_STATUS).in(AccountStatus.TERMINATED, AccountStatus.DELETED))`. Note this forces a real join (the current `root.get(policy).get(id)` path is FK-only and joins nothing), which is fine but should be paired with a regression test.

frontend/src/community/leave/components/molecules/LeavePoliciesTable/LeavePoliciesTable.tsx
```diff
@@ -177,6 +181,11 @@ const LeavePoliciesTable: FC<Props> = ({
               }
             />
           ),
+          assignedEmployees: (
```
ThinuwanW, line 187: 🤖 **Claude** · 🔴 **Important**

This new column reads server data that the single-assign and unassign mutations never invalidate, so it renders a stale count. In `~community/leave/api/LeavePolicyAssignmentApi.ts`, `useAssignLeavePolicy.onSuccess` (line 80) and `useUnassignLeavePolicy.onSuccess` (line 150) invalidate only `leavePolicyAssignmentQueryKeys.ALL` and `leaveAnalyticsQueryKeys.EMPLOYEE_LEAVE_ENTITLEMENTS_FOR_ANALYTICS_ALL` — not `leavePolicyQueryKeys.ALL`, which is what backs `useGetLeavePoliciesInfinite` here. `useBulkAssignLeavePolicies` (line 117) already does invalidate `leavePolicyQueryKeys.ALL`, so the three sibling mutations are now inconsistent. Repro: open the leave policies page (list cached), assign or unassign a policy for an employee from the people module, navigate back — the ASSIGNED EMPLOYEES value is unchanged until a hard refresh or cache GC. Fix: add `queryClient.invalidateQueries({ queryKey: leavePolicyQueryKeys.ALL })` to the `onSuccess` of both `useAssignLeavePolicy` and `useUnassignLeavePolicy` (import `leavePolicyQueryKeys` — it is already imported in that file).

backend/src/test/java/com/skapp/community/leaveplanner/controller/v1/LeavePolicyControllerIntegrationTest.java
```diff
@@ -263,9 +266,21 @@ void getAllLeavePolicies_LeaveAdmin_ReturnsSeededPolicy() throws Exception {
 				.andExpect(jsonPath(STATUS_PATH).value(STATUS_SUCCESSFUL))
 				.andExpect(jsonPath("$.results[0].items", hasSize(1)))
 				.andExpect(jsonPath("$.results[0].items[0].name").value("Existing Policy"))
+				.andExpect(jsonPath("$.results[0].items[0].assignedEmployeeCount").value(0))
 				.andExpect(jsonPath("$.results[0].totalItems").value(1));
 		}
 
+		@Test
```
ThinuwanW, line 337: 🤖 **Claude** · 🟡 **Suggestion**

The only test for the new query seeds a single policy (500) with two assignments, so it verifies the ACTIVE/ENDED filter but not the per-policy grouping. A regression in the tuple-alias extraction — e.g. reading the count into the key slot, or collapsing all policies onto one entry — would still pass. Add a case that seeds a second policy (e.g. 502) with a different number of ACTIVE assignments plus one policy with zero, and assert distinct `assignedEmployeeCount` values per item; that is the behaviour `countByPolicyIdsAndStatus` actually exists for. Also worth covering: the PR now populates the count on the PUT `/v1/leave/policies/{id}` response (`LeavePolicyServiceImpl` line 124), but `updateLeavePolicy_LeaveAdmin_UpdatesName` (line 287) never asserts `assignedEmployeeCount`, so that new code path is untested. Finally, once the terminated-employee filter is added, seed a TERMINATED employee assignment and assert it is excluded.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/LeavePolicyServiceImpl.java
```diff
@@ -116,7 +120,10 @@ public ResponseEntityDto updateLeavePolicy(Long id, LeavePolicyUpdateRequestDto
 
 		log.info("updateLeavePolicy: policy updated successfully");
 
-		return new ResponseEntityDto(false, leaveMapper.leavePolicyToLeavePolicyResponseDto(leavePolicy));
+		LeavePolicyResponseDto leavePolicyResponseDto = leaveMapper.leavePolicyToLeavePolicyResponseDto(leavePolicy);
+		populateAssignedEmployeeCounts(List.of(leavePolicyResponseDto));
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`updateLeavePolicy` only mutates `name` (line 118), which cannot change the assignment count, yet it now issues an extra grouped-count query on every rename inside a write transaction. Meanwhile `addLeavePolicy` (line 101) returns the DTO without calling `populateAssignedEmployeeCounts` at all, relying on the `= 0L` field initializer. The two write paths are therefore asymmetric for no stated reason, and the frontend does not consume either response body — `useUpdateLeavePolicy` just calls `invalidateQueries(leavePolicyQueryKeys.ALL)` and refetches the list. Either drop this call and let the DTO default stand (cheapest, and consistent with `addLeavePolicy`), or populate it in `addLeavePolicy` too and document that the mutation responses are contract-complete. Note the same value is now defaulted in three places — the DTO initializer, `getOrDefault(dto.getId(), 0L)` at line 203, and `?? 0` in the table component — which is the kind of redundancy that hides a real null later.

backend/src/main/java/com/skapp/community/leaveplanner/constant/LeavePolicyConstant.java
```diff
@@ -17,4 +17,8 @@ public class LeavePolicyConstant {
 
 	public static final int MAX_BULK_ASSIGN_ROWS = 1000;
 
+	public static final String POLICY_ID_ALIAS = "policyId";
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`LeavePolicyConstant` currently holds domain/validation constants (`MAX_NAME_LENGTH`, `MIN_DAYS`, `MAX_BULK_ASSIGN_ROWS`); adding Criteria-API tuple aliases here mixes a persistence-layer detail into a domain constants class, and the constants are referenced only from a single repository method. The rest of the codebase does not alias tuple selections at all — `CrmDealRepositoryImpl.countDealsByStageIds` (line 291-298) and `PolicyLeaveRequestRepositoryImpl.findCommittedUsageForPoliciesInWindow` (line 158-166) both do the exact same grouped-count-into-`Map<Long, Long>` shape using positional `t.get(0, Long.class)` / `t.get(1, Long.class)`. Consider matching that existing pattern and dropping both constants, or if you prefer named aliases, keep them as `private static final` fields on `EmployeeLeavePolicyRepositoryImpl` where they are used. Secondary concern with the current form: `policyIdPath` is aliased and then reused in `where` and `groupBy`; Hibernate handles this today, but reusing a single aliased `SqmPath` across three clauses is more fragile across upgrades than the positional form.

frontend/src/community/leave/types/LeavePolicyTypes.ts
```diff
@@ -92,6 +92,7 @@ export interface LeavePolicyType {
   leaveTypeEmoji: string | null;
   policyType: PolicyType;
   status: LeavePolicyStatus;
+  assignedEmployeeCount?: number;
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`assignedEmployeeCount?: number` is typed optional, but the backend contract makes it non-nullable: `LeavePolicyResponseDto` initializes the field to `0L` and `populateAssignedEmployeeCounts` always sets it via `getOrDefault(..., 0L)`, so the list endpoint always serializes a number. Typing it optional forces the defensive `?? 0` at LeavePoliciesTable.tsx:186 and weakens the contract. Consider `assignedEmployeeCount: number` and dropping the fallback — this also mirrors the sibling non-optional fields above it (`policyType`, `status`). If `LeavePolicyType` is reused for partially-populated objects anywhere that would break, leave it optional but add a short comment like the existing one on line 96 explaining why.

ThinuwanW: 🤖 **Claude** · 🔴 **Important**

📁 **File-level observation**

Missing paired change: this file declares TWO near-identical interfaces mirroring the backend `LeavePolicyResponseDto` — `LeavePolicyType` (used by the list/infinite query) and `LeavePolicyResponseDto` (used by `LeavePolicyMutationResponse`, i.e. the return type of `useAddLeavePolicy`/`useUpdateLeavePolicy`). This PR added `assignedEmployeeCount` only to `LeavePolicyType`. But the backend change to `updateLeavePolicy` now populates and returns `assignedEmployeeCount` on the PUT response, so the mutation-response type is out of sync with the contract this PR itself created — any consumer that wants to read the count off the mutation result gets a type error, and there is no compile-time signal that the field exists. Fix: add the field to `LeavePolicyResponseDto` as well. Better still, collapse the duplication — `LeavePolicyResponseDto` and `LeavePolicyType` differ only in which fields are optional, so define one interface (or `type LeavePolicyType = LeavePolicyResponseDto`) so the next backend field addition cannot drift between them again.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/LeavePolicyServiceImpl.java
```diff
@@ -171,6 +178,8 @@ public ResponseEntityDto getAllLeavePolicies(LeavePolicyFilterDto leavePolicyFil
 		List<LeavePolicyResponseDto> leavePolicyResponseDtos = leaveMapper
 			.leavePolicyListToLeavePolicyResponseDtoList(leavePolicyPage.getContent());
 
+		populateAssignedEmployeeCounts(leavePolicyResponseDtos);
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`populateAssignedEmployeeCounts` runs unconditionally for every caller of `getAllLeavePolicies`, including the unpaginated one. `getAllLeavePolicies` maps `size < 0` to `Pageable.unpaged()` (line 174), and `AssignLeavePolicyModal` on the frontend calls `useGetLeavePoliciesInfinite` with `size: UNPAGINATED_SIZE` (`-1`) purely to populate a policy dropdown. On that path `policyIds` becomes every policy id in the tenant, producing an unbounded `IN (...)` list plus a grouped count scan over the whole `lv_employee_leave_policy` table — for a response whose consumer never reads `assignedEmployeeCount`. Fix options: (a) make count population opt-in via a flag on `LeavePolicyFilterDto` so the dropdown path skips it; or (b) at minimum chunk the id list before the `IN` predicate in `countByPolicyIdsAndStatus`. Note the frontend also shares one query-key namespace between the table and the modal, so the heavy variant is exercised whenever the modal opens.

backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/EmployeeLeavePolicyRepositoryImpl.java
```diff
@@ -76,6 +81,30 @@ public Page<EmployeeLeavePolicy> findByEmployeeIdAndStatusOrderByEffectiveFromDe
...
+		}
+
+		CriteriaBuilder cb = entityManager.getCriteriaBuilder();
+		CriteriaQuery<Tuple> query = cb.createTupleQuery();
+		Root<EmployeeLeavePolicy> root = query.from(EmployeeLeavePolicy.class);
+		Path<Long> policyIdPath = root.get(EmployeeLeavePolicy_.policy).get(LeavePolicy_.id);
+
+		query
+			.multiselect(policyIdPath.alias(LeavePolicyConstant.POLICY_ID_ALIAS),
+					cb.count(root).alias(LeavePolicyConstant.ASSIGNED_COUNT_ALIAS))
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`cb.count(root)` counts assignment ROWS, but the value is surfaced under a column literally labelled "ASSIGNED EMPLOYEES". The at-most-one-ACTIVE-row-per-(employee, leaveType) invariant that makes rows == employees is enforced only in application code (`EmployeeLeavePolicyServiceImpl.assignPolicyToEmployee`); `lv_employee_leave_policy` has no unique constraint on (employee_id, policy_id, status) — see `leave-ddl-script-v1-create-leave-policy-tables.sql`, which declares only the two FKs. `bulkAssignLeavePolicies` also commits row-by-row in separate `TransactionTemplate` transactions, so two concurrent bulk uploads touching the same employee can each insert an ACTIVE row and silently inflate the displayed headcount. Use `cb.countDistinct(root.get(EmployeeLeavePolicy_.employee))` so the number is truthful regardless of duplicate rows; it costs nothing here and makes the query match the label.

backend/src/main/java/com/skapp/community/leaveplanner/payload/response/LeavePolicyResponseDto.java
```diff
@@ -44,4 +44,6 @@ public class LeavePolicyResponseDto {
 
 	private AccrualTiming accrualTiming;
 
+	private Long assignedEmployeeCount = 0L;
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

Two problems with the `= 0L` field initializer. (1) It makes "count was never computed" indistinguishable from "genuinely zero assignments". `addLeavePolicy` (LeavePolicyServiceImpl:101) returns this DTO without calling `populateAssignedEmployeeCounts`, so its `0` is a default, not a measured value — whereas `updateLeavePolicy` (line 124) measures it. That asymmetry is currently harmless only because a brand-new policy really has zero assignments; the day creation supports assigning members, the response starts lying with no compile-time or test signal. (2) Defaults in response DTOs are unusual in this package — every other field here is left null and set explicitly. Prefer dropping the initializer and populating the field on all three response paths (add/update/list), so a missing value surfaces as `null` rather than a plausible-looking `0`.

backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/EmployeeLeavePolicyRepositoryImpl.java
```diff
@@ -76,6 +81,30 @@ public Page<EmployeeLeavePolicy> findByEmployeeIdAndStatusOrderByEffectiveFromDe
...
+			return Map.of();
+		}
+
+		CriteriaBuilder cb = entityManager.getCriteriaBuilder();
+		CriteriaQuery<Tuple> query = cb.createTupleQuery();
+		Root<EmployeeLeavePolicy> root = query.from(EmployeeLeavePolicy.class);
+		Path<Long> policyIdPath = root.get(EmployeeLeavePolicy_.policy).get(LeavePolicy_.id);
+
+		query
+			.multiselect(policyIdPath.alias(LeavePolicyConstant.POLICY_ID_ALIAS),
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The two new alias constants are unnecessary — this exact query shape already has a convention in the same module. `PolicyLeaveRequestRepositoryImpl.findCommittedUsageForPoliciesInWindow` (lines 148-169) builds the same `Path<Long> policyId = root.get(...policy).get(LeavePolicy_.id)` + `createTupleQuery()` + `multiselect` + `Map<Long, ...>` result, and reads results positionally (`result.get(0, Long.class)`, `result.get(1, ...)`) with no string aliases at all. Following that pattern here — `tuple.get(0, Long.class)` / `tuple.get(1, Long.class)` — keeps the two grouped-count queries in the module consistent and lets `POLICY_ID_ALIAS`/`ASSIGNED_COUNT_ALIAS` be deleted entirely, which also removes the reason they were added to `LeavePolicyConstant`. Alternatively use `cb.construct` into a small projection record, as `EpExternalSyncStagingRepositoryImpl` and `AdmsAttendanceLogRepositoryImpl` do for aggregates.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/LeavePolicyServiceImpl.java
```diff
@@ -116,7 +120,10 @@ public ResponseEntityDto updateLeavePolicy(Long id, LeavePolicyUpdateRequestDto
 
 		log.info("updateLeavePolicy: policy updated successfully");
 
-		return new ResponseEntityDto(false, leaveMapper.leavePolicyToLeavePolicyResponseDto(leavePolicy));
+		LeavePolicyResponseDto leavePolicyResponseDto = leaveMapper.leavePolicyToLeavePolicyResponseDto(leavePolicy);
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`assignedEmployeeCount` has no corresponding property on the `LeavePolicy` entity, so `leavePolicyToLeavePolicyResponseDto` (and the list variant that reuses it) will now emit a MapStruct "Unmapped target property: assignedEmployeeCount" warning on every build. `LeaveMapper` is a plain `@Mapper(componentModel = "spring")` with no `unmappedTargetPolicy`, so this is a warning rather than a build failure — but it is new build noise that hides the next real unmapped-property mistake. Add `@Mapping(target = "assignedEmployeeCount", ignore = true)` alongside the existing `@Mapping` annotations on `LeaveMapper:164-167` to state the intent that this field is populated post-mapping.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/LeavePolicyServiceImpl.java
```diff
@@ -181,6 +190,19 @@ public ResponseEntityDto getAllLeavePolicies(LeavePolicyFilterDto leavePolicyFil
 		return new ResponseEntityDto(false, pageDto);
 	}
 
+	private void populateAssignedEmployeeCounts(List<LeavePolicyResponseDto> leavePolicyResponseDtos) {
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

`populateAssignedEmployeeCounts` is a private helper wedged between two public `@Override` methods (`getAllLeavePolicies` and `enableLeavePolicies`). Every other private helper in this class — `removeExistingLeaveAllocations`, `voidLeaveRequests`, `releaseEntitlementUsage`, `buildLeavePolicy`, `applyAccrualDetail`, etc. — is grouped after all public methods, from line 274 onward. Move it down with the others so the public API of the class stays readable top-to-bottom.

backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/EmployeeLeavePolicyRepositoryImpl.java
hasalarootcode: Fixed. Added an inner join to `Employee` and excluded `TERMINATED`/`DELETED`, with a regression test seeding a terminated assignment.

backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/EmployeeLeavePolicyRepositoryImpl.java
hasalarootcode: Fixed. `cb.countDistinct` on the employee join now.

frontend/src/community/leave/components/molecules/LeavePoliciesTable/LeavePoliciesTable.tsx
hasalarootcode: Fixed. Added `leavePolicyQueryKeys.ALL` to the `onSuccess` of both `useAssignLeavePolicy` and `useUnassignLeavePolicy`.

backend/src/test/java/com/skapp/community/leaveplanner/controller/v1/LeavePolicyControllerIntegrationTest.java
hasalarootcode: Added `getAllLeavePolicies_LeaveAdmin_CountsAssignmentsPerPolicy`, which seeds three policies with 2/0/1 active assignments and asserts each count, plus `getAllLeavePolicies_TerminatedEmployee_ExcludedFromCount`. The PUT no longer returns the count, so there is nothing to assert there.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/LeavePolicyServiceImpl.java
hasalarootcode: Dropped the call from `updateLeavePolicy`. Both write paths now return the DTO without the count and only the list populates it.

backend/src/main/java/com/skapp/community/leaveplanner/payload/response/LeavePolicyResponseDto.java
hasalarootcode: Dropped the initializer. Only the list path populates the field now, so add and update return null instead of a plausible 0. I did not populate it on the two write paths since nothing reads it there.

frontend/src/community/leave/types/LeavePolicyTypes.ts
hasalarootcode: Fixed. `assignedEmployeeCount: number` and the `?? 0` is gone.

hasalarootcode: `updateLeavePolicy` no longer populates the count, so the mutation response does not carry it and the type stays correct. Collapsing the two interfaces is wider than this PR, happy to raise it separately.

backend/src/main/java/com/skapp/community/leaveplanner/constant/LeavePolicyConstant.java
hasalarootcode: Both constants are already gone, `LeavePolicyConstant` only holds the domain values now.

backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/EmployeeLeavePolicyRepositoryImpl.java
hasalarootcode: Already done, the query reads positionally and both alias constants are removed.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/LeavePolicyServiceImpl.java
hasalarootcode: Leaving the mapper as is. The field is populated after mapping, so the warning is expected.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/LeavePolicyServiceImpl.java
hasalarootcode: Moved down with the other private helpers.

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/LeavePolicyServiceImpl.java
hasalarootcode: The dropdown is the only unpaginated caller, so the cheapest fix is skipping population when `pageable.isUnpaged()`, which changes no contract. The flag on `LeavePolicyFilterDto` needs an FE change too.

backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/EmployeeLeavePolicyRepositoryImpl.java
```diff
@@ -76,6 +83,25 @@ public Page<EmployeeLeavePolicy> findByEmployeeIdAndStatusOrderByEffectiveFromDe
...
+		Path<Long> policyId = root.get(EmployeeLeavePolicy_.policy).get(LeavePolicy_.id);
+
+		query.multiselect(policyId, cb.countDistinct(employee))
+			.where(policyId.in(policyIds), cb.equal(root.get(EmployeeLeavePolicy_.status), status),
+					cb.not(employee.get(Employee_.ACCOUNT_STATUS).in(AccountStatus.TERMINATED, AccountStatus.DELETED)))
+			.groupBy(policyId);
+
+		return entityManager.createQuery(query)
+			.getResultList()
+			.stream()
```
shakila-kamalasena: Remove stream

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/LeavePolicyServiceImpl.java
```diff
@@ -134,7 +143,13 @@ public ResponseEntityDto deactivateLeavePolicy(Long id) {
 		leavePolicy.setStatus(LeavePolicyStatus.INACTIVE);
 		leavePolicy = leavePolicyDao.save(leavePolicy);
 
-		log.info("deactivateLeavePolicy: policy deactivated successfully");
+		int cancelledRequests = cancelPendingPolicyLeaveRequests(leavePolicy.getId());
+		int revokedRequests = revokeFutureApprovedPolicyLeaveRequests(leavePolicy.getId());
+		int endedAssignments = endActivePolicyAssignments(leavePolicy.getId());
+
+		log.info(
+				"deactivateLeavePolicy: policy deactivated successfully, cancelledRequests: {}, revokedRequests: {}, endedAssignments: {}",
```
shakila-kamalasena: Remove the log variables

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/LeavePolicyServiceImpl.java
```diff
@@ -267,6 +284,57 @@ public void setDefaultLeavePolicyConfig() {
 		log.info("setDefaultLeavePolicyConfig: execution ended");
 	}
 
+	private void populateAssignedEmployeeCounts(List<LeavePolicyResponseDto> leavePolicyResponseDtos) {
+		if (leavePolicyResponseDtos.isEmpty()) {
+			return;
+		}
+
+		List<Long> policyIds = leavePolicyResponseDtos.stream().map(LeavePolicyResponseDto::getId).toList();
```
shakila-kamalasena: Remove stream

shakila-kamalasena (approved):

shakila-kamalasena (changes requested):
comments

backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/EmployeeLeavePolicyRepositoryImpl.java
hasalarootcode: Fixed

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/LeavePolicyServiceImpl.java
hasalarootcode: Fixed

shakila-kamalasena (dismissed):

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/LeavePolicyServiceImpl.java
hasalarootcode: Fixed

shakila-kamalasena (dismissed):

backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/EmployeeLeavePolicyRepositoryImpl.java
```diff
@@ -76,6 +83,27 @@ public Page<EmployeeLeavePolicy> findByEmployeeIdAndStatusOrderByEffectiveFromDe
...
+	@Override
+	public Map<Long, Long> countByPolicyIdsAndStatus(List<Long> policyIds, EmployeeLeavePolicyStatus status) {
+		CriteriaBuilder cb = entityManager.getCriteriaBuilder();
+		CriteriaQuery<Tuple> query = cb.createTupleQuery();
+		Root<EmployeeLeavePolicy> root = query.from(EmployeeLeavePolicy.class);
+		Join<EmployeeLeavePolicy, Employee> employee = root.join(EmployeeLeavePolicy_.employee);
+		Path<Long> policyId = root.get(EmployeeLeavePolicy_.policy).get(LeavePolicy_.id);
+
+		query.multiselect(policyId, cb.countDistinct(employee))
+			.where(policyId.in(policyIds), cb.equal(root.get(EmployeeLeavePolicy_.status), status),
```
akilarootcode: same like AccountStatus.TERMINATED, AccountStatus.DELETED, remove status parameter

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/LeavePolicyServiceImpl.java
```diff
@@ -172,6 +184,8 @@ public ResponseEntityDto getAllLeavePolicies(LeavePolicyFilterDto leavePolicyFil
 		List<LeavePolicyResponseDto> leavePolicyResponseDtos = leaveMapper
 			.leavePolicyListToLeavePolicyResponseDtoList(leavePolicyPage.getContent());
 
+		populateAssignedEmployeeCounts(leavePolicyResponseDtos);
```
akilarootcode: you can get the assigned employee count directly from findLeavePolicies(), and in many cases that can be cleaner because you avoid the separate countByPolicyIdsAndStatus() query.

akilarootcode (changes requested):

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/LeavePolicyServiceImpl.java
```diff
@@ -267,6 +281,64 @@ public void setDefaultLeavePolicyConfig() {
...
+	}
+
+	private int cancelPendingPolicyLeaveRequests(Long policyId) {
+		return voidPolicyLeaveRequests(
+				policyLeaveRequestDao.findByPolicy_IdAndStatus(policyId, LeaveRequestStatus.PENDING),
+				LeaveRequestStatus.CANCELLED);
+	}
+
+	private int revokeFutureApprovedPolicyLeaveRequests(Long policyId) {
+		return voidPolicyLeaveRequests(policyLeaveRequestDao.findByPolicy_IdAndStatusAndStartDateAfter(policyId,
```
akilarootcode: voidPolicyLeaveRequests use directly

akilarootcode (changes requested):

backend/src/main/java/com/skapp/community/leaveplanner/repository/impl/EmployeeLeavePolicyRepositoryImpl.java
hasalarootcode: Fixed

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/LeavePolicyServiceImpl.java
hasalarootcode: Fixed

backend/src/main/java/com/skapp/community/leaveplanner/service/impl/LeavePolicyServiceImpl.java
hasalarootcode: Fixed

akilarootcode (dismissed):

ThinuwanW (dismissed):

akilarootcode (approved):

ThinuwanW (approved):

PR #1840, opened 2026-08-27
feat: add CrmIndustry model, repository, and default industry values

YasinduRC:
### Summary

Adds the `crm_industry` table and wires it into the default CRM setup for new organizations — the first step toward letting industries be data instead of a hardcoded enum.

- New `crm_industry` table (id, audit columns, `name`), added as its own Liquibase changeset appended to `common-ddl-script-v1-create-tables.sql`
- New `CrmIndustry` entity + `CrmIndustryDao`
- `DefaultCrmIndustryValues` / `DefaultCrmIndustryTemplate` seed the same 20 industries currently in `CrmIndustry` (enum), mirroring `DefaultCrmDealStageValues`/`DefaultCrmTaskTypeValues`
- `CrmConfigServiceImpl.setDefaultCrmConfig()` now also seeds industries, so new orgs get the 20 rows automatically

**Out of scope for this PR:** no CRUD/search API, no changes to `crm_company.industry` or its enum — those are unaffected. This is schema + default-data only.

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds a `crm_industry` lookup table, entity, DAO and default-seeding logic. The Java layer is consistent with the existing deal-stage/task-type templates, but the database side has a serious deployment problem: the new table and the `crm_company.industry_id` FK were added *inside* the already-applied `common-ddl-script-v1-create-crm-tables` changeset rather than as a new changeset file, which will break startup on every existing database (Liquibase checksum mismatch, and `ddl-auto: validate` failing on the unmapped `crm_industry` table). Additionally the migration is half-finished — `industry_id` has no entity mapping and no backfill, and existing organizations will never get the default industries seeded since seeding only runs at org creation.

Found **14** new issue(s): 🔴 4 important, 🟡 7 suggestion(s), 🟣 3 nit(s)

backend/src/main/resources/community/db/changelog/ddl-script/common-ddl-script-v1-create-tables.sql
```diff
@@ -598,6 +598,18 @@ CREATE TABLE IF NOT EXISTS `crm_task_type`
     PRIMARY KEY (`id`)
 ) ENGINE = InnoDB;
 
+CREATE TABLE IF NOT EXISTS `crm_industry`
```
ThinuwanW, line 602: 🤖 **Claude** · 🔴 **Important**

The new `crm_industry` table (and the `crm_company` changes below) were inserted into the existing `-- changeset AkilaSachintha:common-ddl-script-v1-create-crm-tables` block. That changeset has already been applied on every existing deployment, so on upgrade Liquibase will recompute the MD5 of the changeset, find it different from DATABASECHANGELOG, and abort startup with a ValidationFailedException (no `runOnChange`, `validCheckSum` or `clearCheckSums` is configured anywhere — see db.changelog.yml and application-prd.yml/application-non-prd.yml, both of which have `liquibase.enabled: true`). Even if checksums were forced through, the changeset is marked EXECUTED and will not re-run, so `crm_industry` and `crm_company.industry_id` would never be created — and because `spring.jpa.hibernate.ddl-auto: validate` is set in prd/non-prd, Hibernate would then fail startup on the new `CrmIndustry` entity with a missing-table error. Note also that CI cannot catch this: tests run with `liquibase.enabled: false` and `ddl-auto: create` on H2. Fix: revert all edits to this baseline file and add a new script (e.g. `crm-ddl-script-v1-create-table-crm-industry.sql`) containing `CREATE TABLE crm_industry ...` plus `ALTER TABLE crm_company ADD COLUMN industry_id bigint DEFAULT NULL, ADD CONSTRAINT FK_crm_company_crm_industry_industry_id ...`, and append an `include` for it at the end of `community/db/changelog/db.changelog.yml` — this is exactly how every other post-baseline change in this repo is done (e.g. `people-ddl-script-v1-alter-table-com-work-location-add-is-deleted.sql`).

backend/src/main/resources/community/db/changelog/ddl-script/common-ddl-script-v1-create-tables.sql
```diff
@@ -607,11 +619,13 @@ CREATE TABLE IF NOT EXISTS `crm_company`
     `last_modified_date` datetime(6)      DEFAULT NULL,
     `name`               text    NOT NULL,
     `industry`           text             DEFAULT NULL,
+    `industry_id`        bigint           DEFAULT NULL,
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`industry_id` and its FK are added to `crm_company`, but `CrmCompany` (model/CrmCompany.java) has no corresponding field — no `@ManyToOne CrmIndustry industry` and no `industryId`. Nothing in the codebase ever reads or writes this column, and the old `industry` text column (mapped as `@Enumerated(EnumType.STRING) CrmIndustry`) is still the sole source of truth, so the schema now carries two competing representations of the same fact with no backfill from one to the other. Either finish the migration in this PR (add the `@ManyToOne` mapping to `CrmCompany`, update `CrmMapper`/`CrmMapperV2` and the create/edit flows in `CrmCompanyServiceImpl`, and add a DML changeset that populates `industry_id` by joining `crm_company.industry` to `crm_industry.name`, then plan the drop of the `industry` column), or drop `industry_id` and the FK from this PR and introduce them together with the entity change.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmConfigServiceImpl.java
```diff
@@ -44,4 +49,12 @@ private void setDefaultCrmTaskTypes() {
 		log.info("setDefaultCrmTaskTypes: execution ended");
 	}
 
+	private void setDefaultCrmIndustries() {
+		log.info("setDefaultCrmIndustries: execution started");
+
+		crmIndustryDao.saveAll(DefaultCrmIndustryTemplate.getDefaultIndustries());
```
ThinuwanW, line 55: 🤖 **Claude** · 🔴 **Important**

`setDefaultCrmConfig()` is only reachable from `OrganizationServiceImpl.setDefaultOrganizationConfigs()` (and the enterprise equivalent), which runs once at organization creation. Every already-provisioned deployment is past that point, so `crm_industry` will remain permanently empty for them — the feature will have no data for existing customers. Add a DML changeset that seeds the 20 default industries for upgrades (following the existing `common-dml-script-v1-update-employee-role-set-crm-role.sql` pattern), or make this method idempotent (e.g. insert only the names not already present) and invoke it on application startup as well.

backend/src/main/java/com/skapp/community/crmplanner/model/CrmIndustry.java
```diff
@@ -0,0 +1,29 @@
...
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+
+@Entity
+@Getter
+@Setter
+@NoArgsConstructor
+@Table(name = "crm_industry")
+public class CrmIndustry extends Auditable<String> {
```
ThinuwanW, line 19: 🤖 **Claude** · 🟡 **Suggestion**

The new entity `com.skapp.community.crmplanner.model.CrmIndustry` has exactly the same simple name as the existing enum `com.skapp.community.crmplanner.type.CrmIndustry`, which is used throughout the module (`CrmCompany.industry`, `CrmCompanyCreateDto`, `CrmCompanyResponseDto`, `CrmValidations.validateIndustry`, and many tests). `DefaultCrmIndustryTemplate` already hits this: it imports the model type while `value.getName()` returns the enum type, so a reader cannot tell which `CrmIndustry` is meant without checking imports, and any future class needing both (a mapper, a service, the eventual `CrmCompany.industry` relation) will have to fully-qualify one of them. Rename the entity to something unambiguous, e.g. `CrmIndustryOption` / `CrmIndustryLookup` (table name can stay `crm_industry`).

backend/src/main/java/com/skapp/community/crmplanner/model/CrmIndustry.java
```diff
@@ -0,0 +1,29 @@
...
+
+	@Id
+	@GeneratedValue(strategy = GenerationType.IDENTITY)
+	@Column(name = "id", nullable = false, updatable = false)
+	private Long id;
+
+	@Column(name = "name", nullable = false)
+	private String name;
```
ThinuwanW, line 27: 🤖 **Claude** · 🟡 **Suggestion**

The DDL declares `is_deleted boolean NOT NULL DEFAULT FALSE` on `crm_industry`, but the entity does not map it. Sibling lookup entities do (`CrmDealStage` and `CrmCompany` both declare `@Column(name = "is_deleted", nullable = false) private Boolean isDeleted = false;`), and `CrmDealStageDao` builds all its finders around `...AndIsDeletedFalse`. As written the column is unreachable from Java, so soft-delete can never be honoured and `CrmIndustryDao.findAll()` would return soft-deleted rows if the column is ever populated. Either add the `isDeleted` field (plus `findAllByIsDeletedFalse...` finders when a read path is added) or remove the column from the DDL.

backend/src/main/java/com/skapp/community/crmplanner/type/DefaultCrmIndustryValues.java
```diff
@@ -0,0 +1,35 @@
...
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+
+@Getter
+@RequiredArgsConstructor
+public class DefaultCrmIndustryValues {
+
+	private final CrmIndustry name;
+
+	public static final List<DefaultCrmIndustryValues> DEFAULT_INDUSTRIES = List.of(
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

This class is a single-field wrapper around the `CrmIndustry` enum and adds no data (unlike `DefaultCrmDealStageValues`/`DefaultCrmTaskTypeValues`, which carry colour/order/type). It duplicates the enum's contents by hand, so adding a constant to `CrmIndustry` will silently not be seeded — a drift bug with no compile-time guard. Prefer deriving the list directly: `Arrays.stream(CrmIndustry.values()).filter(v -> v != CrmIndustry.NONE).toList()`, dropping this class and the wrapper indirection entirely. Also please confirm the omission of `CrmIndustry.NONE` is intentional and that companies persisted with `industry = NONE` are handled once the FK relation is wired up.

backend/src/main/resources/community/db/changelog/ddl-script/common-ddl-script-v1-create-tables.sql
```diff
@@ -598,6 +598,18 @@ CREATE TABLE IF NOT EXISTS `crm_task_type`
...
 ) ENGINE = InnoDB;
 
+CREATE TABLE IF NOT EXISTS `crm_industry`
+(
+    `id`                 bigint  NOT NULL AUTO_INCREMENT,
+    `created_by`         text             DEFAULT NULL,
+    `created_date`       datetime(6)      DEFAULT NULL,
+    `last_modified_by`   text             DEFAULT NULL,
+    `last_modified_date` datetime(6)      DEFAULT NULL,
+    `name`               text    NOT NULL,
```
ThinuwanW, line 609: 🤖 **Claude** · 🟡 **Suggestion**

`crm_industry.name` has no uniqueness guarantee and `setDefaultCrmIndustries()` calls `saveAll(...)` unconditionally, so any second invocation of `setDefaultCrmConfig()` (a re-run of org setup, a retried provisioning request, or the startup-seeding suggested above) silently duplicates all 20 rows, which would then show up twice in any industry picker. Guard the seed (e.g. `if (crmIndustryDao.count() == 0)` or insert only missing names) and consider a unique index on `name`. Note the column is declared `text`, which MySQL cannot index uniquely without a prefix length — `varchar(255)` would be the cleaner declaration for a lookup key.

backend/src/main/java/com/skapp/community/crmplanner/constant/DefaultCrmIndustryTemplate.java
```diff
@@ -0,0 +1,28 @@
...
+import com.skapp.community.crmplanner.model.CrmIndustry;
+import com.skapp.community.crmplanner.type.DefaultCrmIndustryValues;
+
+import lombok.experimental.UtilityClass;
+
+@UtilityClass
+public class DefaultCrmIndustryTemplate {
+
+	public static List<CrmIndustry> getDefaultIndustries() {
+		List<CrmIndustry> industries = new ArrayList<>();
```
ThinuwanW, line 15: 🤖 **Claude** · 🟣 **Nit**

This template builds an `ArrayList` in a for-loop while the two sibling templates in the same package (`DefaultCrmDealStageTemplate.getDefaultStages()` and `DefaultCrmTaskTypeTemplate.getDefaultTaskTypes()`) both use `VALUES.stream().map(Template::toEntity).toList()`. The preceding commit deliberately moved away from that shared idiom; unless there was a specific reason, align with the siblings so all three read identically.

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmConfigServiceImpl.java
```diff
@@ -44,4 +49,12 @@ private void setDefaultCrmTaskTypes() {
 		log.info("setDefaultCrmTaskTypes: execution ended");
 	}
 
+	private void setDefaultCrmIndustries() {
```
ThinuwanW, line 52: 🤖 **Claude** · 🟣 **Nit**

No test accompanies the new seeding path — there is no `CrmConfigServiceImplTest` at all, and the integration tests run with `liquibase.enabled: false` and `ddl-auto: create` on H2, so neither the `CrmIndustry` entity/DDL agreement nor the seeded row set is exercised anywhere. A small unit test asserting that `setDefaultCrmConfig()` passes 20 industries to `crmIndustryDao.saveAll(...)` (and that the names match `CrmIndustry` minus `NONE`) would catch the enum-drift risk noted on `DefaultCrmIndustryValues`.

backend/src/main/resources/community/db/changelog/ddl-script/common-ddl-script-v1-create-tables.sql
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

No data migration accompanies the new `industry_id` FK, and the seed set cannot represent every existing value. `CrmCompany.industry` is an `@Enumerated(EnumType.STRING) CrmIndustry` persisted into the old `industry` text column, and every write path (`CrmCompanyServiceImpl.createCompany` / `updateCompany`, the MapStruct mapper, and all Criteria projections in `CrmCompanyRepositoryImpl`, `CrmContactRepositoryImpl`, `CrmTaskRepositoryImpl` which select `CrmCompany_.industry`) still uses it. So `industry_id` will be NULL for 100% of rows, old and new — the schema now has two competing sources of truth for the same attribute with no plan to reconcile them. Worse, `DefaultCrmIndustryValues.DEFAULT_INDUSTRIES` omits `CrmIndustry.NONE`, but `CrmValidations.validateIndustry` only rejects `null`, so `NONE` is an accepted and persistable company industry today with no corresponding `crm_industry` row — such companies could never be backfilled even if a migration were written. Fix: add a DML changeset that seeds `crm_industry` (including a `NONE` row, or explicitly decide `NONE` maps to `industry_id IS NULL`) and backfills `crm_company.industry_id` by joining on `industry`, and land the entity relation + write-path switch in the same PR. If the intent is to ship the column ahead of the code, say so in the changeset comment and drop the FK until the relation exists, rather than leaving a permanently-NULL FK column in production.

backend/src/main/java/com/skapp/community/crmplanner/repository/CrmIndustryDao.java
```diff
@@ -0,0 +1,10 @@
+package com.skapp.community.crmplanner.repository;
```
ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

The DAO is write-only — `crmIndustryDao.saveAll(...)` in `CrmConfigServiceImpl.setDefaultCrmIndustries()` is the only usage in the entire codebase. There is no `CrmIndustryService`, no controller, no response DTO, and no mapper; grepping the repo shows nothing ever queries this table. Meanwhile the frontend still sources its industry list from a hardcoded `CrmIndustryEnum` in `frontend/src/community/crm/v2/enums/common.ts` (and the v1 copy), so the seeded rows are consumed by nobody. Compare the sibling lookup tables, which both ship with a read path (`CrmTaskTypeController` / `CrmTaskTypeDao.findAllByOrderByOrderIndexAscIdAsc`, `CrmDealStageController` / `CrmDealStageDao`). As it stands this ships dead data plus a schema change with no observable behavior. Either land the read path (service + `/v1/crm/industries` endpoint returning a `*ResponseDto`, with the frontend switched to consume it) in this PR, or hold the schema/seed change until the consumer is ready.

backend/src/main/resources/community/db/changelog/ddl-script/common-ddl-script-v1-create-tables.sql
```diff
@@ -598,6 +598,18 @@ CREATE TABLE IF NOT EXISTS `crm_task_type`
...
+    `is_deleted`         boolean NOT NULL DEFAULT FALSE,
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`crm_industry` diverges from both sibling lookup tables by omitting `order_index`. `crm_deal_stage` has `order_index int NOT NULL` and `crm_task_type` has `order_index int DEFAULT NULL`, and both DAOs expose ordered finders (`findAllByIsDeletedFalseOrderByOrderIndexAsc()`, `findAllByOrderByOrderIndexAscIdAsc()`). `CrmIndustryDao` is empty, so any future list endpoint falls back to `findAll()`, which returns rows in unspecified order. Note that ordering by `name` will not help either: the stored values are enum constants (`TECHNOLOGY_INFORMATION_AND_MEDIA`), while the UI renders translated labels (`"Technology, information and media"`), so alphabetical order in the DB will not match alphabetical order on screen. Add `order_index int NOT NULL` to the table, an `orderIndex` field on the entity, populate it in `DefaultCrmIndustryValues`/`DefaultCrmIndustryTemplate` the way the sibling templates do, and add `findAllByIsDeletedFalseOrderByOrderIndexAsc()` to `CrmIndustryDao` (which also gives the table's unmapped `is_deleted` column a consumer).

backend/src/main/java/com/skapp/community/crmplanner/constant/DefaultCrmIndustryTemplate.java
```diff
@@ -0,0 +1,28 @@
...
+		for (DefaultCrmIndustryValues value : DefaultCrmIndustryValues.DEFAULT_INDUSTRIES) {
+			industries.add(toCrmIndustryEntity(value));
+		}
+		return industries;
+	}
+
+	private static CrmIndustry toCrmIndustryEntity(DefaultCrmIndustryValues value) {
+		CrmIndustry industry = new CrmIndustry();
+		industry.setName(value.getName().name());
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`industry.setName(value.getName().name())` uses `Enum.name()` directly, whereas both sibling templates go through an explicit display accessor: `DefaultCrmTaskTypeTemplate` calls `value.getName().getDisplayName()` and `DefaultCrmDealStageTemplate` calls `value.getStage().getName()` (their enums carry a `String` field precisely so the persisted label is decoupled from the Java constant). Because `CrmIndustry` has no such field, this silently couples the persisted `crm_industry.name` values to the Java identifiers — renaming or reordering an enum constant later changes what is written for new tenants while already-seeded tenants keep the old string, with no compiler signal. Give `CrmIndustry` a `displayName` field following the `CrmTaskTypeName`/`CrmDealStageName` pattern and call the accessor here, so the storage contract is explicit.

backend/src/main/resources/community/db/changelog/ddl-script/common-ddl-script-v1-create-tables.sql
ThinuwanW, line 609: 🤖 **Claude** · 🟣 **Nit**

`name text NOT NULL` is a poor type for a lookup key. In MySQL a `TEXT` column cannot take a `UNIQUE` (or any) index without an explicit prefix length, which means the natural fix for duplicate seeding — a unique constraint on `crm_industry.name` — cannot be added later without first altering the column type. It also prevents an efficient lookup-by-name join when backfilling `crm_company.industry_id` from the existing `crm_company.industry` text values. Declaring this as `varchar(255) NOT NULL` now (while the table is brand new and empty) costs nothing and keeps both options open; the same applies to `crm_company.industry` if it is ever used as a join key.

backend/src/main/resources/community/db/changelog/ddl-script/common-ddl-script-v1-create-tables.sql
YasinduRC: This is Intentional, the entity mapping and write path land in the follow-up PR, so this one stays schema + seed data only

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmConfigServiceImpl.java
YasinduRC: Correct that this only covers new tenents, and that's intentional. Existing tenants get seeded by an migration endpoint

backend/src/main/java/com/skapp/community/crmplanner/model/CrmIndustry.java
YasinduRC: Fixed, renamed the enum to CrmIndustryName

backend/src/main/java/com/skapp/community/crmplanner/model/CrmIndustry.java
YasinduRC: right now the column can't be read or written from Java at all. There's no delete endpoint planned for industries this was added there can be requirement in future to delete them, since that it was added to the table but no mentioned in the model

backend/src/main/resources/community/db/changelog/ddl-script/common-ddl-script-v1-create-tables.sql
YasinduRC: setDefaultCrmDealStages() and setDefaultCrmTaskTypes() right above this both call saveAll unguarded too, so I'd rather keep the three consistent

backend/src/main/java/com/skapp/community/crmplanner/constant/DefaultCrmIndustryTemplate.java
YasinduRC: I switched to for loop since we are limiting the usage for stream in out code as a practice

backend/src/main/java/com/skapp/community/crmplanner/service/impl/CrmConfigServiceImpl.java
YasinduRC: There's no CrmConfigServiceImplTest at all today, deal-stage and task-type seeding aren't covered either

backend/src/main/java/com/skapp/community/crmplanner/repository/CrmIndustryDao.java
YasinduRC: Intentional, this PR is schema + seed only so it can be reviewed and rolled back independently. The service, endpoint and frontend switch come in follow-up PRs

backend/src/main/resources/community/db/changelog/ddl-script/common-ddl-script-v1-create-tables.sql
YasinduRC: Left out on purpose,  there's no reorder feature and no update endpoint for industries, so nothing would write the column.

backend/src/main/resources/community/db/changelog/ddl-script/common-ddl-script-v1-create-tables.sql
YasinduRC: This was intentional, we use TEXT instead of VARCHAR

backend/src/main/resources/community/db/changelog/ddl-script/common-ddl-script-v1-create-tables.sql
YasinduRC: For new table creations we do not add new scripts to the end of the existing ddl file

backend/src/main/resources/community/db/changelog/ddl-script/common-ddl-script-v1-create-tables.sql
YasinduRC: The backfill and write-path switch are coming in a follow-up PR

backend/src/main/java/com/skapp/community/crmplanner/constant/DefaultCrmIndustryTemplate.java
YasinduRC: Fixed, followed the convection and used value.getStage().getName()

backend/src/main/java/com/skapp/community/crmplanner/type/DefaultCrmIndustryValues.java
YasinduRC: Removed the wrapper and now it directly mapped using the enum since it has only the name

anusham-rc (dismissed):

backend/src/main/resources/community/db/changelog/ddl-script/common-ddl-script-v1-create-tables.sql
```diff
@@ -598,6 +598,18 @@ CREATE TABLE IF NOT EXISTS `crm_task_type`
     PRIMARY KEY (`id`)
 ) ENGINE = InnoDB;
 
+CREATE TABLE IF NOT EXISTS `crm_industry`
+(
+    `id`                 bigint  NOT NULL AUTO_INCREMENT,
+    `created_by`         text             DEFAULT NULL,
+    `created_date`       datetime(6)      DEFAULT NULL,
+    `last_modified_by`   text             DEFAULT NULL,
+    `last_modified_date` datetime(6)      DEFAULT NULL,
```
akilarootcode: do same as migration BE

akilarootcode (changes requested):

backend/src/main/resources/community/db/changelog/ddl-script/common-ddl-script-v1-create-tables.sql
YasinduRC: Done remove default null and default false

anusham-rc (approved):

ThinuwanW (approved):

PR #1841, opened 2026-08-28
fix: use skapp-ui checkbox in other supervisors dropdown

Deshan1100:
### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-

### DIFF
--- a/frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
+++ b/frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
@@ -1,8 +1,7 @@
 import { Box, Stack, Typography } from "@mui/material";
-import { ButtonV2, SmallModal } from "@rootcodelabs/skapp-ui";
+import { ButtonV2, Checkbox, SmallModal } from "@rootcodelabs/skapp-ui";
 import { useFormik } from "formik";
 
-import Checkbox from "~community/common/components/atoms/Checkbox/Checkbox";
 import Icon from "~community/common/components/atoms/Icon/Icon";
 import Tooltip from "~community/common/components/atoms/Tooltip/Tooltip";
 import { Modules, RoleLevel } from "~community/common/enums/CommonEnums";
@@ -146,7 +145,6 @@ const RestrictedUserRolesModal = ({ initialData }: Props) => {
               <Checkbox
                 key={roleLevel}
                 label={translateText([ROLE_LEVEL_LABEL_KEYS[roleLevel]])}
-                name={roleLevel}
                 checked={values.selected.includes(roleLevel)}
                 onChange={() => onRoleLevelChange(roleLevel)}
               />
--- a/frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/styles.ts
+++ b/frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/styles.ts
@@ -13,7 +13,8 @@ const styles = () => ({
     marginX: "0.5rem"
   },
   fieldWrapper: {
-    paddingX: "0.875rem"
+    paddingX: "0.1875rem",
+    gap: "0.875rem"
   }
 });
 
--- a/frontend/src/community/people/components/molecules/MultiSelectManagerSearch/MultiSelectManagerSearch.tsx
+++ b/frontend/src/community/people/components/molecules/MultiSelectManagerSearch/MultiSelectManagerSearch.tsx
@@ -1,5 +1,6 @@
 import { CircularProgress, Stack } from "@mui/material";
 import { Box, type Theme, useTheme } from "@mui/system";
+import { Checkbox } from "@rootcodelabs/skapp-ui";
 import {
   ChangeEvent,
   Dispatch,
@@ -12,7 +13,6 @@ import {
   useState
 } from "react";
 
-import Checkbox from "~community/common/components/atoms/Checkbox/Checkbox";
 import AvatarChip from "~community/common/components/molecules/AvatarChip/AvatarChip";
 import SearchBox from "~community/common/components/molecules/SearchBox/SearchBox";
 import { KeyboardKeys } from "~community/common/enums/KeyboardEnums";
@@ -284,6 +284,8 @@ const MultiSelectManagerSearch = ({
                 sx={{
                   width: "100%",
                   px: "0.75rem",
+                  alignItems: "center",
+                  gap: "0.75rem",
                   backgroundColor: isFocused
                     ? theme.palette.action.hover
                     : !isSelected
@@ -299,17 +301,9 @@ const MultiSelectManagerSearch = ({
                 onMouseEnter={() => setFocusedIndex(index)}
               >
                 <Checkbox
-                  label={""}
-                  name={""}
                   checked={isSelected}
+                  ariaLabel={`${employee.firstName ?? ""} ${employee.lastName ?? ""}`}
                   onChange={() => toggleManagerSelection(employee)}
-                  customStyles={{
-                    color: theme.palette.primary.dark,
-                    "&.Mui-checked": {
-                      color: theme.palette.primary.dark
-                    }
-                  }}
-                  tabIndex={-1}
                 />
                 <Box
                   sx={{


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR migrates two components from the local `~community/common/components/atoms/Checkbox` atom to the design-system `Checkbox` from `@rootcodelabs/skapp-ui`, with accompanying spacing adjustments. The prop usage (`checked`/`label`/`ariaLabel`/`onChange(checked)`) matches the conventions already established in other DS `Checkbox` call sites (e.g. `PeopleSyncSectionTable.tsx`, `AddTeamSelectMembers.tsx`), and dropping `name` is safe since neither Formik in `RestrictedUserRolesModal` nor `MultiSelectManagerSearch` bound anything to it. I found no blocking correctness bugs, but several accessibility/behavioural details in `MultiSelectManagerSearch` were silently dropped along with the old atom's props and are worth confirming before merge.

Found **11** new issue(s): 🟡 6 suggestion(s), 🟣 5 nit(s)

frontend/src/community/people/components/molecules/MultiSelectManagerSearch/MultiSelectManagerSearch.tsx
```diff
@@ -299,17 +301,9 @@ const MultiSelectManagerSearch = ({
                 onMouseEnter={() => setFocusedIndex(index)}
               >
                 <Checkbox
```
ThinuwanW, line 303: 🤖 **Claude** · 🟡 **Suggestion**

The explicit `tabIndex={-1}` that was passed to the old atom is gone, and the DS `Checkbox` call here provides no replacement. This listbox is a virtual-focus widget: the search input is the `role="combobox"` with `aria-activedescendant`, and everything inside the popup was deliberately kept out of the tab order (`Stack` option at line 282 and `AvatarChip` at line 335 both set `tabIndex={-1}`). The DS Checkbox almost certainly renders a native `<input type="checkbox">`, which is tabbable by default, so every visible option now adds a tab stop inside the popper. `handleContainerKeyDown` does close the popper on Tab, but the `Popper` has `timeout={300}`, so focus can still land on a checkbox in a fading-out popup. Please verify against the `@rootcodelabs/skapp-ui@1.1.1` Checkbox API whether it forwards `tabIndex` (or `inputProps`) to the underlying input, and restore `tabIndex={-1}`; if the DS component does not support it, wrap the checkbox in a container that removes it from the tab order. Related: ARIA specifies `option` as having presentational children, so a focusable/labelled checkbox nested inside `role="option"` is not a valid pattern to begin with.

frontend/src/community/people/components/molecules/MultiSelectManagerSearch/MultiSelectManagerSearch.tsx
```diff
@@ -299,17 +301,9 @@ const MultiSelectManagerSearch = ({
...
-                  label={""}
-                  name={""}
                   checked={isSelected}
+                  ariaLabel={`${employee.firstName ?? ""} ${employee.lastName ?? ""}`}
                   onChange={() => toggleManagerSelection(employee)}
```
ThinuwanW, line 306: 🤖 **Claude** · 🟡 **Suggestion**

Clicking the checkbox fires both `Checkbox.onChange` -> `toggleManagerSelection(employee)` and, via bubbling, the parent `Stack`'s `onClick` -> `handleItemClick` -> `toggleManagerSelection(employee)`. Today this is self-cancelling only by accident: both calls read the same stale `selectedManagers` closure, compute the same `newSelectedManagers`, and call `setSelectedManagers` with an identical value (not an updater), so the net effect is a single toggle. That invariant breaks the moment anyone converts `toggleManagerSelection` to a functional `setSelectedManagers(prev => ...)` update, or adds any non-idempotent side effect next to `setEmploymentDetails`. The PR is already rewriting this exact block, so it is a good time to make it explicit: either drop the checkbox's own `onChange` (a no-op handler) and let the row click be the single source of truth, or wrap the checkbox in `<div onClick={(e) => e.stopPropagation()}>` — the pattern already used around DS checkboxes in `PeopleSyncSectionTable.tsx` and `PeopleSyncImportReviewTable.tsx`.

frontend/src/community/people/components/molecules/MultiSelectManagerSearch/MultiSelectManagerSearch.tsx
```diff
@@ -299,17 +301,9 @@ const MultiSelectManagerSearch = ({
                 onMouseEnter={() => setFocusedIndex(index)}
               >
                 <Checkbox
-                  label={""}
-                  name={""}
                   checked={isSelected}
+                  ariaLabel={`${employee.firstName ?? ""} ${employee.lastName ?? ""}`}
```
ThinuwanW, line 305: 🤖 **Claude** · 🟣 **Nit**

The new `ariaLabel` duplicates the accessible name already set on the parent `role="option"` element at line 281 (`${employee.firstName} ${employee.lastName}` plus the translated ", selected" suffix). Because `option` has presentational children per ARIA, this label will most likely not be exposed to assistive tech at all, and where it is, the employee name gets announced twice. Also note the inconsistency between the two: line 281 interpolates `employee.firstName`/`lastName` raw, while line 305 uses `?? ""` — since both are optional on `L4ManagerType`, line 281 can render the literal string "undefined undefined". Consider dropping `ariaLabel` here (the row already carries the name and `aria-selected`) and applying the `?? ""` guard to line 281 instead.

frontend/src/community/people/components/molecules/MultiSelectManagerSearch/MultiSelectManagerSearch.tsx
```diff
@@ -299,17 +301,9 @@ const MultiSelectManagerSearch = ({
                 onMouseEnter={() => setFocusedIndex(index)}
               >
                 <Checkbox
-                  label={""}
-                  name={""}
                   checked={isSelected}
```
ThinuwanW, line 304: 🤖 **Claude** · 🟣 **Nit**

The removed `customStyles` forced the checkbox to `theme.palette.primary.dark` in both the unchecked and `Mui-checked` states, which looks deliberate given selected rows render on a `theme.palette.secondary.main` background (line 293) and focused rows on `theme.palette.action.hover` (line 290). The DS Checkbox now supplies its own colours. Please eyeball the selected and keyboard-focused states to confirm the checkbox still has sufficient contrast against those two backgrounds.

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/styles.ts
```diff
@@ -13,7 +13,8 @@ const styles = () => ({
     marginX: "0.5rem"
   },
   fieldWrapper: {
-    paddingX: "0.875rem"
+    paddingX: "0.1875rem",
```
ThinuwanW, line 16: 🤖 **Claude** · 🟣 **Nit**

`0.1875rem` is exactly `0.875rem - 0.6875rem`, i.e. it is compensating for the `marginLeft: -11px` that MUI's `FormControlLabel` applied inside the old Checkbox atom — a magic number tied to an implementation detail of a component that is no longer used here. Two things worth tightening: (1) add a short comment explaining what the value aligns to, or express it as a theme/spacing token; (2) `paddingX` shrinks the right padding too (0.875rem -> 0.1875rem), which was never affected by the old negative margin — if only left alignment needed fixing, use `paddingLeft` and keep the original right padding.

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
```diff
@@ -1,8 +1,7 @@
 import { Box, Stack, Typography } from "@mui/material";
-import { ButtonV2, SmallModal } from "@rootcodelabs/skapp-ui";
+import { ButtonV2, Checkbox, SmallModal } from "@rootcodelabs/skapp-ui";
```
ThinuwanW, line 2: 🤖 **Claude** · 🟣 **Nit**

This migration is partial — `~community/common/components/atoms/Checkbox/Checkbox` is still imported by 4 files (`LeaveCarryForwardTypeContent.tsx`, `MarkOutOfOfficeModal.tsx`, `PeopleSyncImportGroupFilterList.tsx`, `ExternalUserContent.tsx`). Two checkbox implementations with divergent APIs (`onChange: () => void` + required `name`/`label` vs `onChange: (checked: boolean) => void` + `ariaLabel`) will keep coexisting and are an easy source of prop-shape mistakes. Worth a follow-up ticket to convert the remaining 4 call sites and delete the atom.

frontend/src/community/people/components/molecules/MultiSelectManagerSearch/MultiSelectManagerSearch.tsx
ThinuwanW, line 305: 🤖 **Claude** · 🟡 **Suggestion**

The new `ariaLabel` can degenerate to a single space, producing a checkbox with an effectively empty accessible name. `displayItems` is a union of `EmployeeDataType | L4ManagerType`, and when `managerSearchTerm` is empty the list is built from `selectedManagers`, i.e. `L4ManagerType` — where `firstName`, `lastName`, `employeeId` and `authPic` are ALL declared optional (`community/people/types/PeopleTypes.ts:207-212`). So `firstName ?? ""` and `lastName ?? ""` can both resolve to `""`, leaving `ariaLabel=" "`. Screen readers then announce an unnamed checkbox. Secondly, every other accessibility string in this component is routed through `translateAria` (`searchManagers`, `managerOptions`, `selected`), but this new label is raw interpolated data with no localized context — there is no verb, so it reads as a bare name rather than "Select <name>". If the label is kept at all, add a key such as `selectManager` to `commonAria.components.multiSelectManagerSearch` in `community/common/assets/languages/english/aria/common.json` and compose it, with a non-empty fallback for unnamed records, e.g. build a `displayName` variable once (`` `${employee.firstName ?? ""} ${employee.lastName ?? ""}`.trim() ``) and use `translateAria(["selectManager"], { name: displayName || translateAria(["unnamedEmployee"]) })`. Reusing that same `displayName` variable would also fix the adjacent `aria-label` on the `role="option"` Stack, which interpolates the raw fields with no `??` guard and will render the literal string "undefined undefined" for the same records — this PR added null-handling on one line while leaving the identical expression two lines above unguarded.

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
```diff
@@ -146,7 +145,6 @@ const RestrictedUserRolesModal = ({ initialData }: Props) => {
               <Checkbox
```
ThinuwanW, line 145: 🤖 **Claude** · 🟡 **Suggestion**

The migration drops to `label` only, with no `ariaLabel`, which may leave these checkboxes without an accessible name. The old atom rendered a MUI `FormControlLabel` wrapping the input, which guarantees a programmatic label/input association, so `label` alone was sufficient for a screen reader. The design-system component makes no such guarantee from this side of the boundary, and the established convention in this codebase is to pass both when a visible label exists — see `CareerProgressDetailsSection.tsx:407-415`, which passes `label={translateText(["currentEmployment"])}` AND `ariaLabel={translateAria(["selectCurrentEmploymentChecked"])}` together. `PeopleSyncSectionTable.tsx:66-70` likewise relies on `ariaLabel` for its name. The fact that the DS component exposes a separate `ariaLabel` prop at all, and that the one usage with a visible label still sets it, suggests `label` is not wired as the input's accessible name. Safest fix here is to also pass `ariaLabel={translateText([ROLE_LEVEL_LABEL_KEYS[roleLevel]])}` so the name is guaranteed regardless of how the DS renders `label`. Worth confirming against the `@rootcodelabs/skapp-ui@1.1.1` implementation before merging, since node_modules is not vendored and this cannot be verified from the repo alone.

frontend/src/community/people/components/molecules/MultiSelectManagerSearch/MultiSelectManagerSearch.tsx
```diff
@@ -284,6 +284,8 @@ const MultiSelectManagerSearch = ({
                 sx={{
                   width: "100%",
                   px: "0.75rem",
+                  alignItems: "center",
+                  gap: "0.75rem",
```
ThinuwanW, line 288: 🤖 **Claude** · 🟡 **Suggestion**

Adding `alignItems: "center"` + `gap: "0.75rem"` here makes this row markup a near-exact third copy of a pattern that already exists twice, which is the duplication case the guidelines call out (2+ components with near-identical markup should be extracted). Compare: `WorkLocationEmployeeSelector.tsx:368-389` renders `role="option"` + `className="flex items-center gap-3 ..."` + `<Checkbox checked={...} />` + `<AvatarChip ... />` — `gap-3` is exactly the 0.75rem being hardcoded here; and `AddTeamSelectMembers.tsx:104-119` renders `flex flex-row items-center` + `<Checkbox checked ariaLabel={...} onChange={...} />` + `<AvatarChip ... />`. All three are diverging in exactly the ways that matter: only `WorkLocationEmployeeSelector` omits `onChange` (letting the row's `onClick` own the toggle), only `AddTeamSelectMembers` and this file set `ariaLabel`, and each expresses the same 0.75rem gap differently (Tailwind class vs. `sx`). Consider extracting a shared `EmployeeOptionRow` molecule taking `{ employee, isSelected, onToggle }` that owns the `role="option"`/`aria-selected`/gap/checkbox composition, so the accessibility and click-ownership decisions are made once instead of drifting across three call sites.

frontend/src/community/people/components/molecules/MultiSelectManagerSearch/MultiSelectManagerSearch.tsx
```diff
@@ -1,5 +1,6 @@
 import { CircularProgress, Stack } from "@mui/material";
 import { Box, type Theme, useTheme } from "@mui/system";
+import { Checkbox } from "@rootcodelabs/skapp-ui";
```
ThinuwanW, line 3: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Pre-existing but directly in the interaction this PR is fixing: the search input cannot accept a space character, so multi-word searches are impossible. `handleKeyDown` is wired to `SearchBox`'s `onKeyDown`, and its `case KeyboardKeys.SPACE` branch (`KeyboardEnums.ts` defines `SPACE = " "`) calls `event.preventDefault()` unconditionally, before the `focusedIndex >= 0` guard. Because `preventDefault` on keydown suppresses character insertion, typing "John Smith" into the manager search silently drops the space and produces "JohnSmith", which will not match. Note the ENTER case is fine — Enter has no default text-insertion behavior. Fix by moving the `preventDefault()` inside the guard so it only fires when a list item is actually focused, and by splitting SPACE out of the shared ENTER case: for SPACE, only `preventDefault()` and toggle when `focusedIndex >= 0 && displayItems[focusedIndex]`, otherwise fall through and let the space be typed. This matters more now that the checkbox is the DS component whose `onChange` signature is `(checked: boolean) => void` with no event object, so keyboard toggling has to be handled entirely by this container.

ThinuwanW, line 3: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

Dead prop in the component being modified: `hasAllSelector?: boolean` is declared on the `Props` interface (line 29) but is never destructured in the component signature (lines 40-49) and never referenced in the body. I checked the only consumer, `SupervisorSelector.tsx:290-299`, and it does not pass it either, so nothing depends on it. Since the file is already being touched for the checkbox migration, remove the prop from the interface — otherwise it reads as a supported 'select all' capability that does not exist, and TypeScript will happily accept a caller passing it while nothing happens.

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
Deshan1100: only need to change this place

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/RestrictedUserRolesModal.tsx
Deshan1100: no need name here

frontend/src/community/configurations/components/organisms/RestrictedUserRolesModal/styles.ts
Deshan1100: checked

frontend/src/community/people/components/molecules/MultiSelectManagerSearch/MultiSelectManagerSearch.tsx
Deshan1100: no need checked

Deshan1100: Dead prop changed

frontend/src/community/people/components/molecules/MultiSelectManagerSearch/MultiSelectManagerSearch.tsx
Deshan1100: kept as previous and checked

frontend/src/community/people/components/molecules/MultiSelectManagerSearch/MultiSelectManagerSearch.tsx
Deshan1100: kept only skapp ui props

frontend/src/community/people/components/molecules/MultiSelectManagerSearch/MultiSelectManagerSearch.tsx
Deshan1100: focus handled in skapp ui

frontend/src/community/people/components/molecules/MultiSelectManagerSearch/MultiSelectManagerSearch.tsx
Deshan1100: need this

Deshan1100: need this

frontend/src/community/people/components/molecules/MultiSelectManagerSearch/MultiSelectManagerSearch.tsx
Deshan1100: checked

hasalarootcode (approved):

shakila-kamalasena (approved):

ThinuwanW (approved):

PR #1842, opened 2026-08-28
fix: stop clock-in timer from running past midnight after auto clock-out

SanithuGunasena07:
The clock-in timer used to keep counting past midnight even after the backend auto-clocked the user out, because the frontend never checked if a new day had started, and the one existing midnight timer was unreliable in background tabs. This PR adds a simple check to the timer's existing per-second tick: if the date has rolled over, it stops the timer and triggers the clock-out immediately, giving a more reliable backup trigger. No backend changes.


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds a client-side midnight-rollover auto clock-out to the Timer molecule, duplicating logic that already exists in TimeWidgetPopupController (the 23:59:58 setTimeout). The core idea is sound (covering the case where the scheduled timeout never fired, e.g. suspended tab), but the implementation has three production-impacting problems: it fires spuriously when `slotStartTime` is still `null` right after an optimistic clock-in, it swallows mutation failures while already flipping the UI to "clocked out", and it submits the END record after the day boundary has passed, which the backend's `recordClockInAndClockOut` (lookup by UTC date of the submitted time) will reject in any non-positive UTC offset. It also puts a mutation into a presentational molecule with no accompanying test.

Found **14** new issue(s): 🔴 7 important, 🟡 5 suggestion(s), 🟣 2 nit(s)

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
```diff
@@ -47,7 +54,18 @@ const Timer = ({ disabled }: TimerProps): JSX.Element => {
         status === AttendanceSlotType.START) &&
       !isAttendanceModalOpen
     ) {
+      const clockInDay = new Date(
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`attendanceParams.slotStartTime` is typed `string | null` and is `null` for the whole window between the optimistic clock-in and the employee-status refetch landing. `ClockWidget.handleClockIn` calls `setSlotType(AttendanceSlotType.START)` synchronously (ClockWidget.tsx:119) before `recordAttendance`, so Timer mounts with `slotType === START` and `slotStartTime === null`. `new Date(null + "Z")` is `new Date("nullZ")` → Invalid Date → `toDateString()` returns the literal string "Invalid Date", which never equals today, so one second later the interval fires the auto clock-out: it POSTs END (possibly racing/preceding the START POST), flips the store to END, and pops the AutoClockOutMidnightModal — which then renders "NaNh NaNm" because `calculateWorkedDuration` also produces NaN from a null start. This is a full second of exposure on every normal clock-in, and much longer on the enterprise geofencing path where `getCurrentLocation()` is awaited before the mutation is even sent. Fix: bail out unless the start time is usable, e.g. `const slotStart = attendanceParams.slotStartTime ? new Date(attendanceParams.slotStartTime + "Z") : null;` and only start the rollover check when `slotStart && !Number.isNaN(slotStart.getTime())`; otherwise fall back to the plain increment tick.

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
```diff
@@ -47,7 +54,18 @@ const Timer = ({ disabled }: TimerProps): JSX.Element => {
...
+        attendanceParams.slotStartTime + "Z"
+      ).toDateString();
+
       interval = setInterval(() => {
+        if (new Date().toDateString() !== clockInDay) {
+          interval && clearInterval(interval);
+          setTimer(0);
+          void updateEmployeeStatus(setSlotType(AttendanceSlotType.END));
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

The mutation has no error handling and no user feedback. `void updateEmployeeStatus(...)` uses `mutateAsync`, so a rejected request becomes an unhandled promise rejection (the `void` operator discards the promise, it does not catch it), and `useUpdateEmployeeStatus` defines no `onError`. Meanwhile `setSlotType(AttendanceSlotType.END)` has already mutated the store synchronously and `setIsAutoClockOutMidnightModalOpen(true)` on the next line shows the "you were clocked out" modal unconditionally — so on failure the user sees a confirmed clock-out while the backend still has them clocked in, and the discrepancy only surfaces on the next status refetch. Given finding #3 this failure path is not hypothetical. Fix: use `mutate(...)` with an `onError` that shows a `ToastType.ERROR` toast (the pattern used in `ClockWidget` via `useRecordAttendance`) and reverts/does not apply the local END state, and move `setIsAutoClockOutMidnightModalOpen(true)` into `onSuccess` so the modal only appears once the server accepted the clock-out.

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
```diff
@@ -47,7 +54,18 @@ const Timer = ({ disabled }: TimerProps): JSX.Element => {
         status === AttendanceSlotType.START) &&
       !isAttendanceModalOpen
     ) {
+      const clockInDay = new Date(
+        attendanceParams.slotStartTime + "Z"
+      ).toDateString();
+
       interval = setInterval(() => {
+        if (new Date().toDateString() !== clockInDay) {
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

Day-boundary semantics disagree with the backend, so this request can be rejected outright. The check uses the browser's *local* calendar day (`toDateString()`), but `updateStatus` sends `time = convertDateToUTC(new Date().toISOString())` and `TimeServiceImpl.recordClockInAndClockOut` resolves the record via `timeRecordDao.findByEmployeeAndDate(employee, DateTimeUtils.epochMillisToUtcLocalDate(timeInMillis))` — i.e. keyed on the *UTC* date of the submitted time. For any timezone with a UTC offset <= 0, local midnight is already past UTC midnight, so the END is looked up against a date with no TimeRecord and the backend throws `TIME_ERROR_CLOCK_IN_NOT_EXISTS_FOR_CURRENT_DATE`. Note also that, unlike the existing controller which deliberately fires at 23:59:58 (before the boundary), this branch by construction fires *after* the boundary, so the recorded clock-out timestamp belongs to the following day; and because `setInterval` is throttled to ~1/min in hidden tabs and suspended entirely across device sleep, the detected time can be minutes or hours late, producing a bogus record spanning midnight. Fix: compare UTC days to match the backend (e.g. `new Date(slotStartTime + "Z").toISOString().slice(0, 10)` vs `new Date().toISOString().slice(0, 10)`), and either extend `updateStatus` to accept an explicit `time` so the END can be clamped to the end of the clock-in day, or route this case through the existing incomplete-clock-out endpoint (`useUpdateIncompleteClockOuts`) which does accept an explicit `clockOutTime`.

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
```diff
@@ -2,6 +2,7 @@ import { Box, Stack, Typography, useTheme } from "@mui/material";
 import { JSX, useCallback, useEffect, useState } from "react";
 
 import { useGetAttendanceConfiguration } from "~community/attendance/api/AttendanceAdminApi";
+import { useUpdateEmployeeStatus } from "~community/attendance/api/AttendanceApi";
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

This duplicates auto clock-out ownership across two components. `TimeWidgetPopupController.tsx:83-105` already schedules the same `updateEmployeeStatus(setSlotType(END))` + `setIsAutoClockOutMidnightModalOpen(true)` at 23:59:58, and now the Timer molecule does it again on rollover — with a different day-boundary rule and, unlike the controller, without re-checking that `status` is still START/RESUME at fire time. In practice the controller's synchronous `setSlotType(END)` tears down the Timer's interval before it can double-fire, but that safety depends on React's batching/commit ordering rather than on anything explicit, and there is no guard if both fire in the same wake-up after a suspend. Two clock-out POSTs would make the second one fail. Also, a presentational molecule now owns a server mutation and global modal state, which is the controller organism's job. Suggest extracting a single `useAutoClockOutAtMidnight()` hook that owns both the scheduled-timeout and the rollover-detection paths (with an `isSubmitting` guard), and calling it only from `TimeWidgetPopupController`; the Timer then goes back to just rendering elapsed time.

ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

No test accompanies this change and `Timer.tsx` has no `Timer.test.tsx` at all, even though the sibling attendance modals (`AutoClockOutMidnightModal.test.tsx`, `ClockOutModal.test.tsx`) are covered. This is time-sensitive logic that silently mutates attendance records, so it is exactly the kind of code that needs a test. Add `Timer.test.tsx` using `jest.useFakeTimers()` + `jest.setSystemTime()` with a `MockTheme` wrapper, covering at minimum: (a) crossing midnight while START triggers exactly one `updateEmployeeStatus` call and opens the modal, (b) `slotStartTime === null` does NOT trigger a clock-out, (c) a rejected mutation surfaces an error and does not leave the UI in the clocked-out state.

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
```diff
@@ -47,7 +54,18 @@ const Timer = ({ disabled }: TimerProps): JSX.Element => {
         status === AttendanceSlotType.START) &&
       !isAttendanceModalOpen
     ) {
+      const clockInDay = new Date(
+        attendanceParams.slotStartTime + "Z"
+      ).toDateString();
+
       interval = setInterval(() => {
```
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Two smaller points on this block. (1) `clockInDay` is a misleading name: `slotStartTime` is populated from `ActiveTimeSlotResponseDto.starTime`, which is the *active time slot's* start (i.e. the last RESUME time after a break), not the original clock-in time — `slotStartDay` would be accurate. (2) Detecting the rollover by polling a 1-second interval is both imprecise and throttled in background tabs; computing the exact delay to the next boundary and using a single `setTimeout`, plus a re-check on `visibilitychange`/mount, would detect the rollover promptly after a suspend rather than up to a minute late, and keeps the per-second tick doing only what it needs to do.

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
```diff
@@ -32,6 +38,7 @@ const Timer = ({ disabled }: TimerProps): JSX.Element => {
   const translateText = useTranslator("attendanceModule", "timeWidget");
 
   const { data: attendanceConfig } = useGetAttendanceConfiguration();
+  const { mutateAsync: updateEmployeeStatus } = useUpdateEmployeeStatus();
```
ThinuwanW, line 48: 🤖 **Claude** · 🔴 **Important**

This bypasses `useRecordAttendance`, the established hook that every user-initiated clock-in/clock-out goes through (`ClockWidget.tsx:52`, `ClockOutModal.tsx:40`). That hook does two things this raw `useUpdateEmployeeStatus()` call does not: (1) in enterprise mode with `attendanceConfig.isGeoFencingEnabled` and a configured geofence, it routes through `useUpdateEmployeeStatusWithLocation` with lat/long instead of the plain endpoint, and (2) it accepts an `onError` callback so failures surface as a toast. As written, the midnight auto clock-out on a geofenced enterprise tenant posts with no location, so it will either be rejected by geofence validation or persist a clock-out with no location while every other clock-out on that tenant has one. Fix: use `const { recordAttendance } = useRecordAttendance(() => setToastMessage({...}))` and call `recordAttendance(AttendanceSlotType.END)` instead of the raw mutation. Note `TimeWidgetPopupController.tsx:32` has the same gap — fixing both in one place (a shared `useAutoClockOutAtMidnight` hook) would resolve it consistently.

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

There is no once-per-day or in-flight guard, so this can re-fire in a loop the user cannot escape. Concrete path: the auto clock-out opens `AutoClockOutMidnightModal`, whose "Clock in again" button (`AutoClockOutMidnightModal.tsx:39`) calls `mutate(setSlotType(AttendanceSlotType.START))`. `setSlotType` only flips `slotType` — `slotStartTime` still holds the previous day's value until the async `employeeStatus` refetch lands. So `ClockWidget.showTimer` immediately becomes true, Timer remounts, `clockInDay` is derived from yesterday's `slotStartTime`, and the very next 1s tick auto-clocks-out again and re-opens the modal. The same loop reopens later via `refetchOnWindowFocus` (not disabled for `useGetEmployeeStatus`): if the POST failed, the store keeps the un-rolled-back `END` but the refetch restores `START`/`RESUME` with a stale `slotStartTime`, and the tick fires again. Fix: hold the date already auto-clocked-out in a `useRef` and bail if it matches, and additionally skip the check while the mutation is pending (`isPending`) or while `attendanceParams.slotStartTime` has not yet been refreshed for the current session.

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
```diff
@@ -47,7 +54,18 @@ const Timer = ({ disabled }: TimerProps): JSX.Element => {
...
+      const clockInDay = new Date(
+        attendanceParams.slotStartTime + "Z"
+      ).toDateString();
+
       interval = setInterval(() => {
+        if (new Date().toDateString() !== clockInDay) {
+          interval && clearInterval(interval);
+          setTimer(0);
+          void updateEmployeeStatus(setSlotType(AttendanceSlotType.END));
+          setIsAutoClockOutMidnightModalOpen(true);
```
ThinuwanW: 🤖 **Claude** · 🔴 **Important**

`setIsPreMidnightClockOutAlertOpen(false)` is never called, so the two midnight modals can be open simultaneously. `TimeWidgetPopupController` opens the pre-midnight alert at 23:50 (`TimeWidgetPopupController.tsx:76`) and nothing closes it automatically. When this line then sets `isAutoClockOutMidnightModalOpen`, `modalContent()` (`TimeWidgetPopupController.tsx:144-152`) renders both `PreMidnightClockOutAlertModal` and `AutoClockOutMidnightModal` stacked inside the single `SmallModal`, and `getModalTitle()` returns `clockOutAlert` because `isPreMidnightClockOutAlertOpen` is checked first — so the dialog is titled "clock out alert" while showing the "you were auto clocked out" body. Worse, `handleCloseModal` only clears the first truthy flag, so dismissing once leaves the auto clock-out modal on screen under the wrong title. Fix: call `setIsPreMidnightClockOutAlertOpen(false)` immediately before `setIsAutoClockOutMidnightModalOpen(true)`, and make the controller's popup selection mutually exclusive (a single enum-typed popup state rather than three independent booleans).

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
```diff
@@ -47,7 +54,18 @@ const Timer = ({ disabled }: TimerProps): JSX.Element => {
         status === AttendanceSlotType.START) &&
```
ThinuwanW, line 86: 🤖 **Claude** · 🔴 **Important**

`AttendanceSlotType.PAUSE` is not covered, leaving the exact gap this PR is meant to close still open for employees on a break. `ClockWidget.showTimer` (`ClockWidget.tsx:91-97`) renders `<Timer/>` for START, RESUME **and** PAUSE, but this condition only starts the interval for START/RESUME, so a paused employee at midnight is never auto clocked out — and `TimeWidgetPopupController`'s 23:59:58 timeout (`TimeWidgetPopupController.tsx:96`) has the same START/RESUME-only check, so nothing else catches it. The record just stays open overnight. Either add PAUSE to the condition (the backend clock-out at `TimeServiceImpl.recordClockInAndClockOut` closes the active BREAK slot via `inactiveLastActivateTimeSlot`, so it is safe), or state explicitly why PAUSE is intentionally excluded.

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

A 1-second `setInterval` is an unreliable trigger for a wall-clock deadline. Browsers throttle timers in backgrounded tabs to at most once per minute (and suspend them entirely while the device is asleep), and the tab is almost certainly backgrounded at midnight. Since `updateStatus` (`AttendanceApi.ts:42`) sends `time: convertDateToUTC(new Date().toISOString())` — i.e. whenever the tick actually happens, not midnight — the persisted `clockOutTime` can be minutes or hours past midnight, silently inflating the employee's worked hours. Prefer computing the exact ms until the next local midnight and using a single `setTimeout` for the rollover (keeping the 1s interval purely for the display), and re-check the rollover on a `visibilitychange`/`focus` listener so a tab woken from sleep catches up immediately.

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
```diff
@@ -47,7 +54,18 @@ const Timer = ({ disabled }: TimerProps): JSX.Element => {
...
+        attendanceParams.slotStartTime + "Z"
+      ).toDateString();
+
       interval = setInterval(() => {
+        if (new Date().toDateString() !== clockInDay) {
+          interval && clearInterval(interval);
+          setTimer(0);
+          void updateEmployeeStatus(setSlotType(AttendanceSlotType.END));
+          setIsAutoClockOutMidnightModalOpen(true);
+          return;
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The confirmation modal will show the wrong worked hours moments after it opens. `AutoClockOutMidnightModal` recomputes `calculateWorkedDuration(attendanceParams)` from live store state on every render. The mutation's `onSuccess` runs `invalidateAttendanceTimeRecordQueries`, which refetches `employeeStatus`; the backend's completed-record branch (`TimeServiceImpl.getActiveTimeSlot`, the `clockOutTime != null` path) returns `starTime = clockOutTime`, `periodType = END`, and leaves `workHours`/`breakHours` unset. `useGetEmployeeStatus` writes those straight into the store, so `calculateWorkedDuration` then evaluates to `now - clockOutTime + 0` ≈ 0 and the chip flips from the real duration to "0h 0m" while the modal is still open. The displayed `clockOutTime` has the same problem — it is `new Date()` at render, not the timestamp actually sent. Fix: snapshot the worked duration and the clock-out timestamp at the moment of the auto clock-out and pass them into the modal as props instead of recomputing from live store state.

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
```diff
@@ -47,7 +54,18 @@ const Timer = ({ disabled }: TimerProps): JSX.Element => {
...
       !isAttendanceModalOpen
     ) {
+      const clockInDay = new Date(
+        attendanceParams.slotStartTime + "Z"
+      ).toDateString();
+
       interval = setInterval(() => {
+        if (new Date().toDateString() !== clockInDay) {
+          interval && clearInterval(interval);
+          setTimer(0);
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`setTimer(0)` here is dead code. `setSlotType(AttendanceSlotType.END)` on the next line makes `ClockWidget.showTimer` (`ClockWidget.tsx:91-97`) false, so `Timer` unmounts on the very next render pass — the zeroed timer never paints. The `interval && clearInterval(interval)` above it is likewise redundant, since changing `status` re-runs this effect and the cleanup clears the handle. Keeping the `clearInterval` as a guard against a second tick is defensible, but `setTimer(0)` should be dropped.

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
ThinuwanW: 🤖 **Claude** · 🟣 **Nit**

Nesting a store mutation inside an argument expression hides the ordering. `setSlotType` both writes to the Zustand store and returns its argument (`attendanceSlice.ts:33-42`), so this single line commits the local status to `END` *before* the request is even sent, and nothing rolls it back on failure. Split it for readability and to make the optimistic write explicit: assign `const slotType = AttendanceSlotType.END;`, call the mutation with it, and move `setSlotType(slotType)` into the success path (or keep it optimistic but pair it with a rollback in `onError`). This mirrors the copied pattern at `TimeWidgetPopupController.tsx:99`, which is worth cleaning up at the same time.

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
SanithuGunasena07: Added a guard so the rollover check only runs once slotStartTime is non-null, falling back to the normal +1 tick otherwise.

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
SanithuGunasena07: Aligned with the existing pattern used in TimeWidgetPopupController.tsx added await before showing the modal, so it now only appears once the mutation actually resolves, instead of firing unconditionally.

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
SanithuGunasena07: No longer applies as we now just call refetchEmployeeStatus() instead of submitting our own clock-out, so no date/time is sent from this path anymore.

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
SanithuGunasena07: No longer applies Timer only does a read-only refetch now, no mutation, so no duplicate-POST risk.

SanithuGunasena07: not required

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
SanithuGunasena07: Renamed clockInDay to slotStartDay

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
SanithuGunasena07: No longer applies as Timer only refetches now, no mutation is posted from this path.

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
SanithuGunasena07: Not required

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
SanithuGunasena07: Pre Existing out of scope for PR

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
SanithuGunasena07: Pre Existing out of scope for PR

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
SanithuGunasena07: No longer applies as Timer doesn't submit any timestamp now, only reads status via refetch, so late detection can't skew persisted data.

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
SanithuGunasena07: Pre Existing out of scope for PR

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
SanithuGunasena07: removed setTimer(0)

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
SanithuGunasena07: No longer applies we no longer call setSlotType/updateEmployeeStatus from this path at all; the code now only does a read-only refetchEmployeeStatus().

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
```diff
@@ -47,7 +53,22 @@ const Timer = ({ disabled }: TimerProps): JSX.Element => {
         status === AttendanceSlotType.START) &&
       !isAttendanceModalOpen
     ) {
+      const slotStart = attendanceParams.slotStartTime
+        ? new Date(attendanceParams.slotStartTime + "Z")
```
shakila-kamalasena: Use an util for this.

shakila-kamalasena: Why the "Z" is not contained in attendanceParams.slotStartTime

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
```diff
@@ -47,7 +53,22 @@ const Timer = ({ disabled }: TimerProps): JSX.Element => {
...
+
       interval = setInterval(() => {
+        if (slotStartDay && new Date().toDateString() !== slotStartDay) {
+          interval && clearInterval(interval);
+          void refetchEmployeeStatus().then((result) => {
+            const latestStatus = result.data?.data.results[0]?.periodType;
+            if (latestStatus === AttendanceSlotType.END) {
+              setIsAutoClockOutMidnightModalOpen(true);
+            }
+          });
```
shakila-kamalasena: Move the then block to handler

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
```diff
@@ -47,7 +53,22 @@ const Timer = ({ disabled }: TimerProps): JSX.Element => {
...
     ) {
+      const slotStart = attendanceParams.slotStartTime
+        ? new Date(attendanceParams.slotStartTime + "Z")
+        : null;
+      const slotStartDay = slotStart?.toDateString();
+
       interval = setInterval(() => {
+        if (slotStartDay && new Date().toDateString() !== slotStartDay) {
+          interval && clearInterval(interval);
+          void refetchEmployeeStatus().then((result) => {
```
shakila-kamalasena: What happened if there is an error?

shakila-kamalasena (changes requested):

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
SanithuGunasena07: moved

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
SanithuGunasena07: On refetch failure, nothing breaks it just falls back to the same pre-existing recovery path that was already there before this fix (route-change refetch, refetchOnWindowFocus), which will pick up the correct status once triggered. No special handling needed here since we're not introducing a new failure mode, just relying on what already exists.

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
SanithuGunasena07: done

SanithuGunasena07: made it handle from BE

frontend/src/community/attendance/utils/CalculateWorkedDuration.ts
```diff
@@ -3,6 +3,10 @@ import {
   attendanceStatusTypes
 } from "~community/attendance/types/attendanceTypes";
 
+export const parseUtcTimestamp = (timestamp: string | null): Date | null => {
```
akilarootcode: remove null on utils

akilarootcode (changes requested):

frontend/src/community/attendance/utils/CalculateWorkedDuration.ts
SanithuGunasena07: removed null handled in parent level

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
```diff
@@ -32,11 +44,28 @@ const Timer = ({ disabled }: TimerProps): JSX.Element => {
...
   const { data: attendanceConfig } = useGetAttendanceConfiguration();
+  const { refetch: refetchEmployeeStatus } = useGetEmployeeStatus();
+  const { mutateAsync: updateEmployeeStatus } = useUpdateEmployeeStatus();
 
   const handleClockOut = useCallback(() => {
     setIsAttendanceModalOpen(true);
   }, [setIsAttendanceModalOpen]);
 
+  const handleEmployeeStatusRefetched = useCallback(
+    (result: Awaited<ReturnType<typeof refetchEmployeeStatus>>) => {
```
hasalarootcode: refetchEmployeeStatus use the already defined return type directly here rather than using typeof

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
```diff
@@ -32,11 +44,28 @@ const Timer = ({ disabled }: TimerProps): JSX.Element => {
...
+        setIsAutoClockOutMidnightModalOpen(true);
+      } else {
+        void updateEmployeeStatus(setSlotType(AttendanceSlotType.END)).then(
+          () => {
+            setIsAutoClockOutMidnightModalOpen(true);
+          }
+        );
+      }
+    },
+    [setIsAutoClockOutMidnightModalOpen, updateEmployeeStatus, setSlotType]
```
hasalarootcode: remove setters

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
```diff
@@ -47,7 +76,17 @@ const Timer = ({ disabled }: TimerProps): JSX.Element => {
         status === AttendanceSlotType.START) &&
       !isAttendanceModalOpen
     ) {
+      const slotStart = attendanceParams.slotStartTime
+        ? parseUtcTimestamp(attendanceParams.slotStartTime)
+        : null;
+      const slotStartDay = slotStart?.toDateString();
+
       interval = setInterval(() => {
+        if (slotStartDay && new Date().toDateString() !== slotStartDay) {
```
hasalarootcode: simplify the condition by moving this slotStartDay && outside from this condition

hasalarootcode: Date().toDateString() move this a variable like slotStartDay

frontend/src/community/attendance/utils/CalculateWorkedDuration.ts
```diff
@@ -3,6 +3,10 @@ import {
   attendanceStatusTypes
 } from "~community/attendance/types/attendanceTypes";
 
+export const parseUtcTimestamp = (timestamp: string): Date => {
```
hasalarootcode: move to common time utils with a proper name

hasalarootcode (changes requested):

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
SanithuGunasena07: removed

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
SanithuGunasena07: done

SanithuGunasena07: Simplified

 slotStartDay is now checked once before the interval starts

frontend/src/community/attendance/utils/CalculateWorkedDuration.ts
SanithuGunasena07: Moved parseUtcTimestamp to dateTimeUtils.ts as parseTimestampToDate

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
```diff
@@ -47,9 +76,26 @@ const Timer = ({ disabled }: TimerProps): JSX.Element => {
...
+      if (slotStartDay) {
+        interval = setInterval(() => {
+          const currentDay = new Date().toDateString();
+          if (currentDay !== slotStartDay) {
+            interval && clearInterval(interval);
+            void refetchEmployeeStatus().then(handleEmployeeStatusRefetched);
+            return;
+          }
+          setTimer((prevTimer) => prevTimer + 1);
+        }, 1000);
```
akilarootcode: move to enum

hasalarootcode (dismissed):

frontend/src/community/attendance/utils/CalculateWorkedDuration.ts
```diff
@@ -13,7 +13,7 @@ export const calculateWorkedDuration = (
     slotType === AttendanceSlotType.RESUME ||
     slotType === AttendanceSlotType.END
   ) {
-    const startTime = new Date(slotStartTime + "Z");
+    const startTime = new Date(slotStartTime as string);
```
akilarootcode: as string remove if not needed

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
```diff
@@ -32,11 +44,28 @@ const Timer = ({ disabled }: TimerProps): JSX.Element => {
   const translateText = useTranslator("attendanceModule", "timeWidget");
 
   const { data: attendanceConfig } = useGetAttendanceConfiguration();
+  const { refetch: refetchEmployeeStatus } = useGetEmployeeStatus();
+  const { mutateAsync: updateEmployeeStatus } = useUpdateEmployeeStatus();
```
akilarootcode, line 48: why we need mutate async

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
```diff
@@ -32,11 +44,28 @@ const Timer = ({ disabled }: TimerProps): JSX.Element => {
...
 
   const handleClockOut = useCallback(() => {
     setIsAttendanceModalOpen(true);
   }, [setIsAttendanceModalOpen]);
 
+  const handleEmployeeStatusRefetched = useCallback(
```
akilarootcode: try with this approach again. 

1. log the results first and check the mismatch

Image

akilarootcode (changes requested):

frontend/src/community/attendance/utils/CalculateWorkedDuration.ts
SanithuGunasena07: removed

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
SanithuGunasena07: the timer component gets removed before onsuccess hence modal wont appear mutate keeps the thread running due to promise

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
SanithuGunasena07: changed approach as discussed

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
SanithuGunasena07: Replaced the magic 1000 with the existing unitConversion.MILLISECONDS_PER_SECOND enum, already used elsewhere in the codebase.

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
SanithuGunasena07: implemented discussed approach

hasalarootcode (dismissed):

shakila-kamalasena (dismissed):

shakila-kamalasena (dismissed):

hasalarootcode (dismissed):

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
```diff
@@ -32,11 +43,39 @@ const Timer = ({ disabled }: TimerProps): JSX.Element => {
...
+    if (
+      employeeStatusData.data.results[0]?.periodType === AttendanceSlotType.END
+    ) {
+      setIsAutoClockOutMidnightModalOpen(true);
+    } else {
+      updateEmployeeStatus(setSlotType(AttendanceSlotType.END), {
+        onSuccess: handleClockOutSuccess
+      });
+    }
+  }, [employeeStatusData, slotStartDay, updateEmployeeStatus, handleClockOutSuccess]);
```
akilarootcode: remove updateEmployeeStatus, handleClockOutSuccess

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
```diff
@@ -59,7 +110,13 @@ const Timer = ({ disabled }: TimerProps): JSX.Element => {
     return () => {
       interval && clearInterval(interval);
     };
-  }, [status, attendanceParams, isAttendanceModalOpen]);
+  }, [
+    status,
+    attendanceParams,
+    isAttendanceModalOpen,
+    refetchEmployeeStatus,
```
akilarootcode: remove refetchEmployeeStatus

akilarootcode (changes requested):

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
SanithuGunasena07: removed

frontend/src/community/attendance/components/molecules/Timer/Timer.tsx
SanithuGunasena07: removed

shakila-kamalasena (approved):

akilarootcode (approved):

ThinuwanW (approved):

PR #1843, opened 2026-08-28
refactor: add deal stage utils, validations and store fields for v2

dulsara-manakal:
Part 1 of 6 of the CRM deal-stage configuration refactor.

Adds the pure foundation layer only - no components, no wiring:

- stageUtil: getSelectedStage, toStageIds, getChangedStageFields, updateStage, removeStage, toStageReorderPayload, isStageNameTaken
- stageValidations: Yup schema for the stage add/edit form
- stageConstants: DEAL_STAGE_COLORS options for the colour selector
- crmRegexPatterns: isDealStageNameValid
- useStageNameMapper: translates seeded default stage names for display
- crmUiSlice + SliceTypes: three deal-stage modal fields

The v2 configurations page stays behind isCrmConfigurationsV2 = false until part 6, so nothing here has a runtime entry point yet.

### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR lands v2 scaffolding for deal-stage management: a `stageUtil` helper module with tests, a Yup validation schema, a stage-name display hook, a `ColorOption` list, and new Zustand UI-slice fields. The Zustand slice and type additions are clean and consistent with the existing modal/selection patterns, and `toStageReorderPayload` correctly matches the backend reorder contract (relative ordering, terminal stages excluded, unique indices). The main concerns are a Yup test that can be invoked with `undefined` and crash inside `isStageNameTaken`, a missing description character-validation that the backend enforces, an ineffective `useCallback`, and the fact that all new code is currently unused while duplicating existing v1 equivalents.

Found **25** new issue(s): 🔴 5 important, 🟡 13 suggestion(s), 🟣 7 nit(s)

frontend/src/community/crm/v2/utils/stageValidations.ts
```diff
@@ -0,0 +1,42 @@
...
+        translator(["dealStageModal", "validations", "nameLength"])
+      )
+      .matches(
+        isDealStageNameValid(),
+        translator(["dealStageModal", "validations", "nameInvalid"])
+      )
+      .test(
+        "is-deal-stage-name-unique",
+        translator(["dealStageModal", "validations", "nameExists"]),
+        (value) => !isStageNameTaken(stages, value, currentStageId)
```
ThinuwanW, line 33: 🤖 **Claude** · 🔴 **Important**

`isStageNameTaken(stages, value, currentStageId)` passes Yup's `value`, which is typed `string | undefined`, into a parameter declared `stageName: string`. Yup v1 runs custom `.test()` functions on absent values by default (that is exactly why `skipAbsent` exists), and `.trim()` on the schema does not coerce `undefined` to a string — so when `name` is missing from `initialValues` (or is cleared to `undefined`), `toStandardStageName` executes `undefined.trim()` and throws a TypeError inside validation, which Formik surfaces as a broken/never-resolving submit rather than a field error. TypeScript would normally catch this, but `next.config.mjs` sets `typescript.ignoreBuildErrors: true`, so it ships silently. Fix with either `(value) => !isStageNameTaken(stages, value ?? "", currentStageId)` or the object form `.test({ name: "is-deal-stage-name-unique", message: ..., test: (value) => !isStageNameTaken(stages, value, currentStageId), skipAbsent: true })`.

frontend/src/community/crm/v2/utils/stageValidations.ts
```diff
@@ -0,0 +1,42 @@
...
+      ),
```
ThinuwanW, line 34: 🤖 **Claude** · 🔴 **Important**

The `description` field validates length only, but the backend also enforces a character regex: `CrmConstants.DEAL_STAGE_DESCRIPTION_REGEX = "^[\\p{L}\\p{N} \\-.,']+$"` (checked in `CrmValidations.validateDealStageDescription`, throwing `CRM_ERROR_DEAL_STAGE_DESCRIPTION_INVALID_CHARS`). Common input such as `!`, `(`, `)`, `/`, `:`, `&`, `%`, or a newline from a multiline textarea passes client validation and then fails server-side, so the user gets a generic error toast instead of an inline field message. Since `name` already mirrors its backend regex via `isDealStageNameValid()`, add a matching `isDealStageDescriptionValid()` to `crmRegexPatterns.ts` returning `/^[\p{L}\p{N} \-.,']+$/u`, chain `.matches(isDealStageDescriptionValid(), translator(["dealStageModal","validations","descriptionInvalid"]), { excludeEmptyString: true })`, and add the `descriptionInvalid` key to `configurations.json` (the `dealStageModal.validations` block currently has no such key).

frontend/src/community/crm/v2/hooks/useStageNameMapper.ts
```diff
@@ -0,0 +1,27 @@
...
+import { DefaultStageNameEnum } from "~community/crm/v2/enums/common";
+
+const useStageNameMapper = () => {
+  const translateText = useTranslator(
+    "crmModule",
+    "deals",
+    "defaultStageNames"
+  );
+
+  const getStageDisplayName = useCallback(
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

The `useCallback` provides no memoization. `useTranslator` (`src/community/common/hooks/useTranslator.tsx:7`) returns a freshly-created arrow function on every render — it is not wrapped in `useCallback`/`useMemo` — so `translateText` has a new identity each render and `getStageDisplayName` is recreated every render anyway. This is worse than it looks: any consumer that puts `getStageDisplayName` in a `useMemo`/`useEffect` dependency array (the natural usage for mapping a stage list) will re-run on every parent render. Either memoize the returned function inside `useTranslator` (benefits all callers) or drop the `useCallback` here so the instability is not disguised.

frontend/src/community/crm/v2/hooks/useStageNameMapper.ts
```diff
@@ -0,0 +1,27 @@
...
+    (stageName?: string): string | undefined => {
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

`getStageDisplayName` is declared `(stageName?: string): string | undefined` and falls off the end of the function with no explicit `return` when `stageName` is undefined. This diverges from the v1 hook (`src/community/crm/hooks/useStageNameMapper.ts:13`), which returns `string`, and forces every call site to guard before rendering or `undefined` will be passed into a Typography/label. Prefer a total signature, e.g. `(stageName = ""): string => Object.values<string>(DefaultStageNameEnum).includes(stageName) ? translateText([stageName]) : stageName`, or at minimum make the `return undefined` explicit so the intent is reviewable.

frontend/src/community/crm/v2/constants/stageConstants.ts
```diff
@@ -14,3 +16,12 @@ export const STAGE_COLOR_MAP: Record<CrmDealStageColorsEnum, string> = {
   ROSEWOOD: "#b55253",
   INDIGO: "#9366fd"
 };
+
+export const DEAL_STAGE_COLORS: ColorOption[] = Object.entries(
```
ThinuwanW, line 20: 🤖 **Claude** · 🟡 **Suggestion**

`Object.entries(STAGE_COLOR_MAP)` widens the key to `string`, so `DEAL_STAGE_COLORS[n].value` and `.id` are typed `string` rather than `CrmDealStageColorsEnum`. Because `CrmStageEntity.color` is `CrmDealStageColorsEnum | undefined`, every consumer that feeds the picked color back into the form/payload will need an `as CrmDealStageColorsEnum` assertion — exactly the pattern the review guidelines call out. Build the list from the enum instead so the type survives: `Object.values(CrmDealStageColorsEnum).map((key) => ({ id: key, name: key, value: key, color: STAGE_COLOR_MAP[key] }))`, and narrow the exported type to `(ColorOption & { value: CrmDealStageColorsEnum })[]`.

frontend/src/community/crm/v2/constants/stageConstants.ts
```diff
@@ -1,3 +1,5 @@
+import { ColorOption } from "@rootcodelabs/skapp-ui";
```
ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Three of the new/changed artifacts are byte-for-byte duplicates of existing v1 code: `DEAL_STAGE_COLORS` is identical to `dealStageColors` in `src/community/crm/utils/crmUtil.ts:133-140`; `getStageValidationSchema` is near-identical to `dealStageValidations` in `src/community/crm/utils/dealStageValidations.ts` (only the uniqueness test differs); and `isDealStageNameValid()` is identical to `src/community/crm/regex/crmRegexPatterns.ts:13`. The v1/v2 fork is intentional in this codebase, but pure constants and regex patterns carry no v1/v2 semantics and will now drift independently (the color hex values in particular are duplicated in two places). Consider having the v2 files re-export the shared constant/regex rather than re-declaring them.

frontend/src/community/crm/v2/hooks/useStageNameMapper.ts
```diff
@@ -0,0 +1,27 @@
+import { useCallback } from "react";
```
ThinuwanW: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

This is the third copy of the hook and it currently has zero consumers — every v2 component still imports the v1 hook: `DealsTableV2.tsx:15`, `DealStageLaneV2.tsx:16`, `DealSidePanelV2/DealPropertiesSidebar.tsx:7`, and `AddDealSidePanelV2/DealNameStageSection.tsx:7` all pull from `~community/crm/hooks/useStageNameMapper`. Also true of `stageUtil.ts`, `stageValidations.ts` and `DEAL_STAGE_COLORS` — nothing in the repo imports them. If this is deliberate scaffolding for a follow-up PR that's fine, but note it means none of this code has been exercised against real call sites yet; migrating the four v2 components to the new hook in this PR would both validate the new signature and remove the ambiguity about which copy is canonical.

frontend/src/community/crm/v2/utils/stageUtil.ts
```diff
@@ -0,0 +1,102 @@
...
+export const isStageNameTaken = (
+  stages: CrmStageEntity[],
+  stageName: string,
+  currentStageId?: number
+): boolean => {
+  const standardName = toStandardStageName(stageName);
+
+  return stages.some(
+    (stage) =>
+      stage.id !== currentStageId &&
```
ThinuwanW, line 98: 🤖 **Claude** · 🟡 **Suggestion**

`stage.id !== currentStageId` silently skips a stage whenever both sides are `undefined`. `CrmStageEntity.id` is optional, and `currentStageId` is `undefined` in add mode — so in add mode any stage lacking an `id` is excluded from the duplicate check rather than compared. The exclusion is meant to apply only to the stage being edited. Make that explicit: `(currentStageId === undefined || stage.id !== currentStageId) && stage.name !== undefined && ...`.

frontend/src/community/crm/v2/utils/stageUtil.ts
```diff
@@ -0,0 +1,102 @@
...
+  return stageIds;
+};
+
+export const getChangedStageFields = (
+  currentValues: CrmStageEntity,
+  initialValues: CrmStageEntity
+): CrmStageEntity => {
+  const changedFields: CrmStageEntity = {};
+
+  if (currentValues.name?.trim() !== initialValues.name?.trim()) {
```
ThinuwanW, line 33: 🤖 **Claude** · 🟡 **Suggestion**

`getChangedStageFields` can produce a key whose value is `undefined`: if `currentValues.name` is undefined while `initialValues.name` is set, the inequality holds and `changedFields.name = undefined` is assigned. Two consequences: (1) a caller checking `Object.keys(changedFields).length === 0` to decide whether to skip the PATCH sees length 1 and fires a no-op request; (2) `updateStage` (line 54) spreads `updatedFields` over the stored stage, so `{ ...stages[id], name: undefined }` wipes the cached name — the optimistic store update would blank the stage label. Guard the assignment, e.g. only set the field when the trimmed current value is defined, or build the object with an explicit `!== undefined` check per field.

frontend/src/community/crm/v2/utils/stageUtil.ts
```diff
@@ -0,0 +1,102 @@
...
+  orderedStages: CrmStageEntity[]
+): CrmDealStageReorderItem[] => {
+  const payload: CrmDealStageReorderItem[] = [];
+
+  for (const stage of orderedStages) {
+    const isPinned =
+      stage.stageType === CrmDealStageEnum.WON ||
+      stage.stageType === CrmDealStageEnum.LOST;
+
+    if (stage.id !== undefined && !isPinned) {
```
ThinuwanW, line 78: 🤖 **Claude** · 🟣 **Nit**

`toStageReorderPayload` silently drops any stage with `id === undefined`. The backend rejects the whole request when the payload length does not match the number of non-terminal stages (`CrmDealStageServiceImpl.java:180` → `CRM_ERROR_DEAL_STAGE_REORDER_INVALID_REQUEST`), so a single id-less stage turns into an opaque server error rather than anything diagnosable. The rest of the function matches the backend contract correctly (1-based indices are fine since the server only uses relative order via `sorted(comparing(orderIndex))` and reassigns from `availableOrderIndexes`, and `WON`/`LOST` match `CrmConstants.TERMINAL_STAGES`). Consider asserting/logging when a stage is skipped.

frontend/src/community/crm/v2/utils/stageValidations.ts
```diff
@@ -0,0 +1,42 @@
...
+      .required(translator(["dealStageModal", "validations", "nameRequired"]))
+      .min(
+        characterLengths.DEAL_STAGE_NAME_MIN_LENGTH,
+        translator(["dealStageModal", "validations", "nameInvalid"])
+      )
+      .max(
+        characterLengths.DEAL_STAGE_NAME_LENGTH,
+        translator(["dealStageModal", "validations", "nameLength"])
+      )
+      .matches(
```
ThinuwanW, line 26: 🤖 **Claude** · 🟣 **Nit**

`.matches()` defaults to `excludeEmptyString: false`, so an empty `name` fails both `.required()` and `.matches()`. Yup runs the tests in a batch and returns whichever rejection lands first under `abortEarly`, so an empty field can display "Please enter a valid stage name" instead of "Stage name is required". Pass `{ excludeEmptyString: true }` as the third argument to `.matches()` to make `required` the only error on an empty field.

frontend/src/community/crm/v2/utils/stageValidations.ts
```diff
@@ -0,0 +1,42 @@
...
+export const getStageValidationSchema = (
+  translator: TranslatorFunctionType,
+  stages: CrmStageEntity[],
+  currentStageId?: number
+) =>
+  Yup.object().shape({
+    name: Yup.string()
+      .trim()
+      .required(translator(["dealStageModal", "validations", "nameRequired"]))
+      .min(
```
ThinuwanW, line 18: 🤖 **Claude** · 🟣 **Nit**

`.min(DEAL_STAGE_NAME_MIN_LENGTH)` reuses the `nameInvalid` message ("Please enter a valid stage name") while `.max()` uses the length-specific `nameLength`. A one-character name therefore reports a character-validity error rather than a length error, which is misleading. Consider adding a `nameTooShort` key to `dealStageModal.validations` in `configurations.json` (e.g. "Stage name must be at least 2 characters").

frontend/src/community/crm/v2/utils/stageValidations.ts
```diff
@@ -0,0 +1,42 @@
...
+        isDealStageNameValid(),
+        translator(["dealStageModal", "validations", "nameInvalid"])
+      )
+      .test(
+        "is-deal-stage-name-unique",
+        translator(["dealStageModal", "validations", "nameExists"]),
+        (value) => !isStageNameTaken(stages, value, currentStageId)
+      ),
+    description: Yup.string()
+      .trim()
```
ThinuwanW, line 36: 🤖 **Claude** · 🟣 **Nit**

`.optional()` is a no-op on `Yup.string()` — string schemas are already optional unless `.required()` is chained. It can be dropped.

frontend/src/community/crm/v2/utils/__test__/stageUtil.test.ts
```diff
@@ -0,0 +1,38 @@
+import { CrmStageEntity } from "~community/crm/v2/types/CrmCommonTypes";
```
ThinuwanW, line 2: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Only `isStageNameTaken` is covered; the other six exports in the new module have no tests. `toStageReorderPayload` is the highest-value gap — it encodes a backend contract (skip `WON`/`LOST`, 1-based contiguous `orderIndex`, payload length must equal the non-terminal stage count) that will break silently if someone reorders the loop or changes the pinned check. `getChangedStageFields` (trim-based diffing) and `updateStage`/`removeStage` (immutability) are also cheap to pin down. Worth adding a test for `getStageValidationSchema` too, particularly `validate({})` with no `name` — that is the `undefined` path called out above.

frontend/src/community/crm/v2/utils/stageUtil.ts
```diff
@@ -0,0 +1,102 @@
...
+      payload.push({ id: stage.id, orderIndex: payload.length + 1 });
+    }
+  }
+
+  return payload;
+};
+
+const toStandardStageName = (stageName: string): string =>
```
ThinuwanW, line 86: 🤖 **Claude** · 🟡 **Suggestion**

`toStandardStageName` normalizes whitespace to `_` and uppercases, which correctly makes the client reject "Demo Scheduled" against the seeded stage stored literally as "DEMO_SCHEDULED" (confirmed in `CrmDealStageName.java:10` and `DefaultCrmDealStageValues.java`). However the server-side guard is `existsByNameIgnoreCaseAndIsDeletedFalse` (`CrmDealStageDao.java:20`), a plain case-insensitive equality that does *not* apply this normalization — so "Demo Scheduled" is still creatable via the API and would render identically to the seeded stage once `useStageNameMapper` translates the latter. This PR's client guard is the only protection. Worth a backend follow-up to normalize the same way, and worth a comment here noting the FE rule is intentionally stricter than the API so the next reader does not "simplify" it back to `toLowerCase()`.

frontend/src/community/crm/v2/utils/stageValidations.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export const getStageValidationSchema = (
+  translator: TranslatorFunctionType,
+  stages: CrmStageEntity[],
+  currentStageId?: number
+) =>
+  Yup.object().shape({
+    name: Yup.string()
+      .trim()
+      .required(translator(["dealStageModal", "validations", "nameRequired"]))
```
ThinuwanW, line 17: 🤖 **Claude** · 🔴 **Important**

The translation keys this schema requests (`dealStageModal.validations.nameRequired` / `nameExists` / `nameLength` / `nameInvalid` / `descriptionLength`) do not exist anywhere in `crmModule.json`. They exist only in `configurations.json` under `crm.dealStageModal.validations`, which is why the v1 caller (`DealStageModalForm.tsx:44`) builds its translator as `useTranslator("configurations", "crm")`. This file lives in `~community/crm/v2/`, and its sibling `getCompanyValidationSchema` (`companyValidations.ts:14`) uses crm-namespace-relative keys like `["validations", "name"]`. Whoever wires this schema up in a v2 CRM component will naturally pass `useTranslator("crmModule", ...)`, and since `useTranslator` returns `t(key) || ""` and i18next returns the key itself when missing, every validation message will render as the literal dotted key string in the UI. Fix: either add a `dealStageModal.validations` block under `crmModule.deals` in `crmModule.json` and keep these key paths, or change the key paths to be relative to the crm deals namespace and document in a JSDoc comment which translator scope the caller must supply.

frontend/src/community/crm/v2/utils/stageUtil.ts
```diff
@@ -0,0 +1,102 @@
...
+  const stageIds: number[] = [];
+  for (const stage of stages) {
+    if (stage.id !== undefined) {
+      stageIds.push(stage.id);
+    }
+  }
+  return stageIds;
+};
+
+export const getChangedStageFields = (
```
ThinuwanW, line 27: 🤖 **Claude** · 🔴 **Important**

Argument order is inverted relative to its direct sibling in the same folder: `getChangedCompanyFields` (`companyUtil.ts:103`) is declared as `(initialValues, currentValues)`, while this one is `(currentValues, initialValues)`. Both parameters are the same type (`CrmStageEntity`, all-optional fields), so a caller that follows the company convention will compile cleanly and silently produce an inverted diff — the PATCH would carry the stage's *old* name/description/color, overwriting the user's edit with the pre-edit values. There is no compile-time protection and no caller yet to catch it. Fix: swap the parameters to `(initialValues, currentValues)` to match `getChangedCompanyFields`, or take a single named-object argument so the two cannot be confused.

frontend/src/community/crm/v2/utils/stageUtil.ts
```diff
@@ -0,0 +1,102 @@
...
+): CrmStageEntity | undefined => {
+  if (stageId !== null) {
+    return stages[stageId];
+  }
+};
+
+export const toStageIds = (stages: CrmStageEntity[]): number[] => {
+  const stageIds: number[] = [];
+  for (const stage of stages) {
+    if (stage.id !== undefined) {
```
ThinuwanW, line 20: 🤖 **Claude** · 🔴 **Important**

`toStageIds` and `toStageReorderPayload` guard with `stage.id !== undefined`, but every other id normalizer in this module uses `!= null`: `toDealIds` (`dealUtil.ts:16`), `toDealsRecord` (`dealUtil.ts:6`), `toStagesRecord` and `getBoardStageIds` (`commonUtil.ts:47,62`), `toCompaniesRecord` (`companyUtil.ts:137`). `CrmStageEntity.id` is typed `number | undefined`, but the value comes straight from a JSON API response, and a JSON `null` passes `!== undefined`. The result is a `null` pushed into a declared `number[]`, and for the reorder path a payload item with `id: null` — the backend's `CrmValidations.validateDealStageReorderRequest` rejects the entire request with `CRM_ERROR_DEAL_STAGE_REORDER_INVALID_REQUEST`, so one null id silently kills the whole reorder. Fix: use `stage.id != null` in both places.

frontend/src/community/crm/v2/utils/stageUtil.ts
```diff
@@ -0,0 +1,102 @@
...
+  stages: CrmStageRecord,
+  stageId: number
+): CrmStageRecord => {
+  const remainingStages = { ...stages };
+  delete remainingStages[stageId];
+
+  return remainingStages;
+};
+
+/** WON and LOST are pinned to the end, so the backend reorders only the rest. */
```
ThinuwanW, line 67: 🤖 **Claude** · 🟡 **Suggestion**

The doc comment states that only WON and LOST are pinned, but the backend enforces a second ordering invariant this helper does not encode: `CrmDealStageServiceImpl.reorderDealStages` sorts the request by `orderIndex` and then throws `CRM_ERROR_DEAL_STAGE_REORDER_INVALID_REQUEST` unless the first item is a stage whose current `stageType` is `INITIAL`. v1 handles this in the UI layer — `DealStagesDraggableContent.tsx:74` splits the INITIAL stage out of the draggable set entirely and re-prepends it. If the v2 board lets a user drag any other stage into first position, this helper will happily emit that order and the mutation will always fail. Fix: either drop `stageType === INITIAL` stages from the payload the way WON/LOST are dropped and re-prepend them, or add the constraint to the doc comment so the consuming drag component enforces it.

frontend/src/community/crm/v2/utils/stageUtil.ts
```diff
@@ -0,0 +1,102 @@
...
+    }
+  }
+
+  return payload;
+};
+
+const toStandardStageName = (stageName: string): string =>
+  stageName.trim().replace(/\s+/g, "_").toUpperCase();
+
+export const isStageNameTaken = (
```
ThinuwanW, line 89: 🤖 **Claude** · 🟡 **Suggestion**

`isStageNameTaken` only sees the stages the caller passes in, which will be the store's stages — and in enterprise free tier the backend hides two seeded stages from every read path. `EpCrmDealStageServiceImpl.filterVisibleDealStages` drops `DEMO_SCHEDULED` and `PROPOSAL_SENT` (`EpCrmConstants.FREE_TIER_HIDDEN_DEFAULT_DEAL_STAGE_NAMES`) for non-core/pro tenants, and both the deal-stage list and the board init payload go through that filter. So the whole point of `toStandardStageName` — blocking a user from typing "Demo Scheduled" against the stored `DEMO_SCHEDULED` — silently does not apply on free tier. The backend won't catch it either: `existsByNameIgnoreCaseAndIsDeletedFalse("Demo Scheduled")` does not match the stored `DEMO_SCHEDULED` (underscore vs space), so creation succeeds and the tenant ends up with two stages that both render as "Demo Scheduled" the moment they upgrade. Worth at least a comment noting the guard is best-effort, or having the create/edit flow surface the backend duplicate error rather than relying on this client check.

frontend/src/community/crm/v2/utils/stageUtil.ts
```diff
@@ -0,0 +1,102 @@
...
+  currentValues: CrmStageEntity,
+  initialValues: CrmStageEntity
+): CrmStageEntity => {
+  const changedFields: CrmStageEntity = {};
+
+  if (currentValues.name?.trim() !== initialValues.name?.trim()) {
+    changedFields.name = currentValues.name?.trim();
+  }
+
+  if (currentValues.description?.trim() !== initialValues.description?.trim()) {
```
ThinuwanW, line 37: 🤖 **Claude** · 🟡 **Suggestion**

Description-change detection drifts from v1. `getChangedDealStageFields` (`crmUtil.ts:63-64`) normalizes both sides with `.trim() || null` before comparing, so `undefined`, `""` and whitespace-only all compare equal. This version compares the raw trimmed values, so if a caller passes the store entity as `initialValues` (where `description` is `undefined` for a stage that never had one) and Formik values as `currentValues` (where an untouched textarea is `""`), `undefined !== ""` and the payload gains `description: ""` even though the user changed nothing. The backend then takes that branch — `editDealStage` only skips a field when it is `null`, and `validateDealStageDescription` returns early for blank — and writes an empty string. Fix: normalize both sides to `trim() || undefined` (or `|| null`) before comparing, matching v1.

frontend/src/community/crm/v2/utils/stageUtil.ts
```diff
@@ -0,0 +1,102 @@
...
+export const getSelectedStage = (
+  stages: CrmStageRecord,
+  stageId: number | null
+): CrmStageEntity | undefined => {
+  if (stageId !== null) {
+    return stages[stageId];
+  }
+};
+
+export const toStageIds = (stages: CrmStageEntity[]): number[] => {
```
ThinuwanW, line 17: 🤖 **Claude** · 🟡 **Suggestion**

`toStageIds` is a third copy of logic that already exists twice in this same utils folder: `getBoardStageIds` (`commonUtil.ts:59`) produces exactly this list from the store's `CrmStageRecord`, and `toDealIds` (`dealUtil.ts:13`) is the same loop for deals. It also has no consumer — the only store field it could feed, `stageIds`/`setStageIds` on `crmUiSlice`, is written nowhere and read nowhere; `DealsKanbanBoardV2.tsx:68` derives its ids via `getBoardStageIds(stages)` instead. Recommend deleting `toStageIds` and using `getBoardStageIds`, or (if the array-input shape is genuinely needed) refactoring `getBoardStageIds` to delegate to it so there is one implementation.

frontend/src/community/crm/v2/utils/stageUtil.ts
```diff
@@ -0,0 +1,102 @@
...
+export const updateStage = (
+  stages: CrmStageRecord,
+  stageId: number,
+  updatedFields: CrmStageEntity
+): CrmStageRecord => ({
+  ...stages,
+  [stageId]: { ...stages[stageId], ...updatedFields }
+});
+
+export const removeStage = (
```
ThinuwanW, line 57: 🤖 **Claude** · 🟣 **Nit**

`removeStage` clones the record unconditionally, so calling it with an id that is not present still returns a new object reference and re-renders every component subscribed to `stages`. The equivalent helper for deals, `removeDealFromRecord` (`dealUtil.ts:41`), guards with an `id in deals` check and returns the original reference when there is nothing to remove. Worth adopting the same guard for consistency and to avoid needless board re-renders. Separately, note that `removeCompany` (`companyUtil.ts:69`) returns both the pruned record *and* the filtered id array while `removeStage` returns only the record — fine today since nothing reads `stageIds`, but the asymmetry is a trap if that field is ever wired up.

frontend/src/community/crm/v2/constants/stageConstants.ts
ThinuwanW, line 1: 🤖 **Claude** · 🟣 **Nit**

`ColorOption` is used only as a type, so this should be `import type { ColorOption } from "@rootcodelabs/skapp-ui"`. `isolatedModules` is enabled in `tsconfig.json` and the crm module already uses `import type` elsewhere (e.g. `kanbanUtil.ts:1`, `useKanbanDrag.ts:1`). This matters more than usual here because `stageConstants.ts` is imported by leaf presentational components that only want `STAGE_COLOR_MAP` — `StageLabel.tsx:3` and `DealStageLaneV2.tsx:19` — and a value import couples them to the design-system package. If the intent is to keep the UI-shaped adapter out of the way, consider moving `DEAL_STAGE_COLORS` next to the color-selector consumer and leaving this file as pure data.

frontend/src/community/crm/v2/store/slices/crmUiSlice.ts
```diff
@@ -25,11 +25,14 @@ const CrmUiSlice: StateCreator<
...
   isTaskModalOpen: false,
   taskModalType: CrmModalTypes.ADD_TASK_MODAL,
+  isDealStageModalOpen: false,
+  dealStageModalType: CrmModalTypes.ADD_DEAL_STAGE_MODAL,
 
   selectedCompanyId: null,
   selectedContactId: null,
   selectedDealId: null,
   selectedTaskId: null,
+  selectedDealStageId: null,
```
ThinuwanW, line 35: 🤖 **Claude** · 🟣 **Nit**

`selectedDealStageId` reads ambiguously next to the existing `selectedDealId` — it scans as "the selected deal's stage id" rather than "the selected stage id". Everything else in this module that refers to the same entity calls it a stage: `stageIds`, `setStageIds`, `stages`, `preselectedStageId`, and the new `getSelectedStage(stages, stageId)` helper. `selectedStageId` (and `setSelectedStageId`) would be unambiguous. The counter-argument is that this mirrors the v1 configurations store field name, so it is a judgement call — but if the v2 CRM store is meant to be the single owner of this state, note that `isDealStageModalOpen` / `dealStageModalType` / `selectedDealStageId` now exist in two Zustand stores (`configurations/stores/slices/dealStageSlice.ts` holds the identical trio), and it should be decided which one owns the deal-stage modal before both are wired to UI.

frontend/src/community/crm/v2/constants/stageConstants.ts
dulsara-manakal: Known issue

frontend/src/community/crm/v2/utils/stageValidations.ts
dulsara-manakal: Tested against yup 1.7.1 , .test() callbacks are not invoked for absent values. With name: undefined or null the schema returns a clean required ValidationError and our test never runs, so undefined.trim() can't happen.

frontend/src/community/crm/v2/utils/stageValidations.ts
dulsara-manakal: Same as V1 not going to change validation

frontend/src/community/crm/v2/hooks/useStageNameMapper.ts
dulsara-manakal: Droped

frontend/src/community/crm/v2/constants/stageConstants.ts
dulsara-manakal: Not doing this one. .value is typed string because ColorOption in @rootcodelabs/skapp-ui declares value: string

frontend/src/community/crm/v2/hooks/useStageNameMapper.ts
dulsara-manakal: Dropped the annotation

frontend/src/community/crm/v2/utils/stageUtil.ts
dulsara-manakal: unreachable, toStagesRecord only admits stages where stage.id != null and keys the record by that id

frontend/src/community/crm/v2/utils/stageUtil.ts
dulsara-manakal: unreachable , toStagesRecord only admits stages where stage.id != null and keys the record by that id

frontend/src/community/crm/v2/utils/stageUtil.ts
dulsara-manakal: the input can't occur, initialValues sets name and description to "" and color to SKY, and Formik's handleChange on a text field writes strings only

frontend/src/community/crm/v2/hooks/useStageNameMapper.ts
dulsara-manakal: This is not true new V2 hook use V2 files.

frontend/src/community/crm/v2/utils/stageValidations.ts
dulsara-manakal: For name: "" the errors come back ordered required, min, matches, and Formik's yupToFormErrors takes the first per field

frontend/src/community/crm/v2/utils/stageValidations.ts
dulsara-manakal: Keep validation as V1 not going to change this.

frontend/src/community/crm/v2/utils/stageUtil.ts
dulsara-manakal: BE check it too API validation is there .

frontend/src/community/crm/v2/utils/__test__/stageUtil.test.ts
dulsara-manakal: Added test for validate them.

frontend/src/community/crm/v2/utils/stageValidations.ts
dulsara-manakal: The translation key refactoring part will be done as a separate task.

frontend/src/community/crm/v2/utils/stageUtil.ts
dulsara-manakal: That validation check is suitable to new types in V2 so  this is correct.

frontend/src/community/crm/v2/utils/stageUtil.ts
dulsara-manakal: Fixed

frontend/src/community/crm/v2/utils/stageUtil.ts
dulsara-manakal: Already handled, same way v1 does it.  Not dragable cards we take out from dragable section.

frontend/src/community/crm/v2/utils/stageUtil.ts
dulsara-manakal: Confirmed. `EpCrmDealStageServiceImpl.filterVisibleDealStages` hides `DEMO_SCHEDULED` and `PROPOSAL_SENT` from free-tier reads

frontend/src/community/crm/v2/utils/stageUtil.ts
dulsara-manakal: `toStageIds` has a consumer

frontend/src/community/crm/v2/utils/stageUtil.ts
dulsara-manakal: Correct about v1's normalisation, but unreachable here

frontend/src/community/crm/v2/utils/stageUtil.ts
dulsara-manakal: The case is unreachable though, the only caller passes selectedDealStage.id, and that stage came out of the record via getSelectedStage, so the id is always present.

frontend/src/community/crm/v2/store/slices/crmUiSlice.ts
dulsara-manakal: Know issue

frontend/src/community/crm/v2/constants/stageConstants.ts
dulsara-manakal: Leaving as-is  for consistancy.

frontend/src/community/crm/v2/utils/stageValidations.ts
dulsara-manakal: Drop it

PR #1844, opened 2026-08-28
refactor: add deal stage card and draggable list components for v2

dulsara-manakal:
Part 2 of 6 of the CRM deal-stage configuration refactor.

Presentational components only - no store access, no API calls:

- DraggableDealStageCard: one stage row with drag handle, colour dot, edit and delete actions
- DraggableDealStageCardSkeleton: loading placeholder
- DealStagesDraggableContent: dnd-kit context that splits stages into draggable and terminal groups and reports a new order upwards

Builds on part 1 for stageConstants, useStageNameMapper and toStageIds. Still behind isCrmConfigurationsV2 = false until part 6.

### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR ports three deal-stage components (draggable list, card, skeleton) from `~community/configurations/components/molecules/` into `~community/crm/v2/`, adapting them to the v2 `CrmStageEntity` type and adding default-stage-name translation via `useStageNameMapper`. The logic is faithful to the v1 originals, but the port carries over pre-existing gaps (unnamed drag handle, no PointerSensor activation constraint, no memoization) into code that is meant to be the corrected v2 baseline, and the three files currently have no importers so nothing is exercised at runtime. The main substantive risks are an accessibility defect on the drag handle, stale `orderIndex` values in the reordered array handed to the parent, and duplication of the terminal-stage predicate that already lives in `stageUtil.ts`.

Found **23** new issue(s): 🔴 3 important, 🟡 13 suggestion(s), 🟣 7 nit(s)

frontend/src/community/crm/v2/components/molecules/DraggableDealStageCard/DraggableDealStageCard.tsx
```diff
@@ -0,0 +1,108 @@
...
+  return (
+    <li
+      ref={setNodeRef}
+      style={style}
+      className="w-full flex flex-row justify-between"
+    >
+      {!isDraggable ? (
+        <div className="w-6 h-6 flex-shrink-0" aria-hidden="true" />
+      ) : (
+        <button
```
ThinuwanW, line 62: 🤖 **Claude** · 🔴 **Important**

The drag-handle button has no accessible name. `useSortable`'s `attributes` only supplies `role`, `tabIndex`, `aria-disabled`, `aria-roledescription="sortable"` and `aria-describedby` — never a label — and the button's only child is `<DragIcon />`. Screen readers announce an unnamed "sortable button" and axe will flag `button-name` (WCAG 4.1.2). This is especially inconsistent given the same component carefully builds `aria-label`s for the edit and delete icon buttons with the stage name interpolated, and the rest of crm/v2 (DealTitleSection, CompanyModalForm, DealsHeaderV2) labels every interactive control. Add a key such as `configurations.crm.aria.reorderStage` ("Reorder stage {{stageName}}") to `configurations.json` and pass `aria-label={translateText(["aria", "reorderStage"], { stageName })}` on the button.

frontend/src/community/crm/v2/components/molecules/DealStagesDraggableContent/DealStagesDraggableContent.tsx
```diff
@@ -0,0 +1,133 @@
...
+    const oldIndex = draggableStages.findIndex(
+      (stage) => stage.id === active.id
+    );
+    const newIndex = draggableStages.findIndex((stage) => stage.id === over.id);
+
+    if (oldIndex === -1 || newIndex === -1) return;
+
+    const reordered = arrayMove(draggableStages, oldIndex, newIndex);
+
+    onStagesReorder([...initialStages, ...reordered, ...terminalStages]);
```
ThinuwanW, line 75: 🤖 **Claude** · 🟡 **Suggestion**

`handleDragEnd` reorders array positions but leaves each stage's `orderIndex` field at its old value, so the array passed to `onStagesReorder` is internally inconsistent. The v1 consumer got away with this because it held the list in a plain `useState` array, but crm/v2 stores stages as `CrmStageRecord` (keyed by id) and every other v2 read path derives order via `getOrderedStages` / `getBoardStageIds`, which sort on `orderIndex`. If a v2 consumer merges this output back into the store (e.g. via `toStagesRecord`), the visual order will silently snap back to the pre-drag order. Either reassign `orderIndex` on the returned stages (mirroring what `toStageReorderPayload` computes: 1-based over the non-pinned stages) or document on the prop that callers must keep the array as the source of order and not merge it into the record.

frontend/src/community/crm/v2/components/molecules/DealStagesDraggableContent/DealStagesDraggableContent.tsx
```diff
@@ -0,0 +1,133 @@
...
+  onDelete
+}: DealStagesDraggableContentProps) => {
+  const translateText = useTranslator("configurations", "crm");
+
+  const sensors = useSensors(
+    useSensor(PointerSensor),
+    useSensor(KeyboardSensor, { coordinateGetter: sortableKeyboardCoordinates })
+  );
+
+  const isTerminalStage = (stageType?: CrmDealStageEnum) =>
```
ThinuwanW, line 43: 🤖 **Claude** · 🟡 **Suggestion**

The WON/LOST "terminal stage" rule is duplicated here as a local `isTerminalStage` helper, but `stageUtil.ts` already encodes exactly the same rule inside `toStageReorderPayload` (`isPinned`). Two copies of the same business rule in the same module will drift the moment a new terminal stage type is added. Export a single `isTerminalStage(stageType?: CrmDealStageEnum)` from `~community/crm/v2/utils/stageUtil` and use it in both places — it also gets test coverage for free, since `stageUtil.test.ts` already exists while these components have none.

frontend/src/community/crm/v2/components/molecules/DealStagesDraggableContent/DealStagesDraggableContent.tsx
```diff
@@ -0,0 +1,133 @@
...
+const DealStagesDraggableContent = ({
+  stagesData,
+  onStagesReorder,
+  onEdit,
+  onDelete
+}: DealStagesDraggableContentProps) => {
+  const translateText = useTranslator("configurations", "crm");
+
+  const sensors = useSensors(
+    useSensor(PointerSensor),
```
ThinuwanW, line 39: 🤖 **Claude** · 🟡 **Suggestion**

`useSensor(PointerSensor)` is registered with no `activationConstraint`, so a drag begins on pointerdown. dnd-kit calls `preventDefault()` on that event, which suppresses focus on the handle button — a user who clicks the handle and then tries to use the keyboard sensor (space/arrows) will not have focus on it. The sibling v2 board (`DealsKanbanBoardV2.tsx:40`) already uses `activationConstraint: { distance: DRAG_ACTIVATION_DISTANCE }` from `~community/crm/constants/boardConstants`. Reuse that constant here for consistency and to avoid accidental one-pixel reorders.

frontend/src/community/crm/v2/components/molecules/DealStagesDraggableContent/DealStagesDraggableContent.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+  return (
+    <div className="w-full flex flex-col gap-8">
+      <DndContext
+        sensors={sensors}
+        collisionDetection={closestCenter}
+        onDragEnd={handleDragEnd}
+      >
+        <SortableContext
+          items={toStageIds(draggableStages)}
```
ThinuwanW, line 86: 🤖 **Claude** · 🟡 **Suggestion**

Nothing in this component is memoized: `initialStages`, `draggableStages`, `terminalStages` and `toStageIds(draggableStages)` are all recomputed on every render. The `items` array in particular gets a fresh identity each render, which changes the `SortableContext` context value and re-renders every `DraggableDealStageCard`. Other v2 components follow the memoized pattern (`useMemo(() => getBoardStageIds(stages), [stages])` in DealsKanbanBoardV2, `useMemo(() => getOrderedStages(stagesRecord), [stagesRecord])` in DealPropertiesSidebar). Wrap the three filtered arrays and the id list in `useMemo` keyed on `stagesData`, and `handleDragEnd` in `useCallback`.

frontend/src/community/crm/v2/components/molecules/DealStagesDraggableContent/DealStagesDraggableContent.tsx
```diff
@@ -0,0 +1,133 @@
+import {
```
ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

All three new files are near-verbatim copies of existing components — `~community/configurations/components/molecules/DealStagesDraggableContent/DealStagesDraggableContent.tsx`, `.../DealStageCard/DraggableDealStageCard.tsx` and `.../DealStageCard/DraggableDealStageCardSkeleton.tsx` (the skeleton is byte-identical with zero v2-specific change). Meanwhile nothing in the repo imports the new crm/v2 versions — `CrmConfigurations.tsx` still imports the configurations copies — so this PR ships dead code alongside a live duplicate. Two copies of the same drag/reorder logic will drift. Please confirm the follow-up that wires the v2 components into the configurations screen and deletes the v1 copies is in flight; if the skeleton has no v2 delta at all, import the existing one rather than duplicating it.

frontend/src/community/crm/v2/components/molecules/DraggableDealStageCard/DraggableDealStageCard.tsx
```diff
@@ -0,0 +1,108 @@
...
+import { useTranslator } from "~community/common/hooks/useTranslator";
+import { IconName } from "~community/common/types/IconTypes";
+import { STAGE_COLOR_MAP } from "~community/crm/v2/constants/stageConstants";
+import { CrmDealStageColorsEnum } from "~community/crm/v2/enums/common";
+import useStageNameMapper from "~community/crm/v2/hooks/useStageNameMapper";
+import { CrmStageEntity } from "~community/crm/v2/types/CrmCommonTypes";
+
+interface DraggableDealStageCardProps {
+  stage: CrmStageEntity;
+  stageId: number;
```
ThinuwanW, line 15: 🤖 **Claude** · 🟡 **Suggestion**

`stageId` duplicates `stage.id` — two sources of truth for the same value with nothing enforcing they agree, and both are used (`useSortable({ id: stageId })`, `String(stageId)` for the StatusCard id, but `onEdit(stage)` / `onDelete(stage)` pass the entity). Drop the `stageId` prop and narrow inside the component instead (`if (stage.id === undefined) return null;`), which also removes the `stage.id !== undefined &&` guard duplication at the two call sites in DealStagesDraggableContent.

frontend/src/community/crm/v2/components/molecules/DealStagesDraggableContent/DealStagesDraggableContent.tsx
```diff
@@ -0,0 +1,133 @@
...
+        onDragEnd={handleDragEnd}
+      >
+        <SortableContext
+          items={toStageIds(draggableStages)}
+          strategy={verticalListSortingStrategy}
+        >
+          <ul className="flex flex-col gap-4">
+            {activeStages.map(
+              (stage) =>
+                stage.id !== undefined && (
```
ThinuwanW, line 92: 🤖 **Claude** · 🟣 **Nit**

`stage.id !== undefined && (<DraggableDealStageCard .../>)` inside `.map()` renders `false` children and silently drops stages from the UI. Two issues: (a) prefer filtering the array before mapping so the JSX stays declarative; (b) the null-check style is inconsistent with the rest of the module — `toStagesRecord` and `getBoardStageIds` in `commonUtil.ts` use `!= null`, while `toStageIds`/this file use `!== undefined`. If the API ever returns `id: null`, this guard passes and `useSortable({ id: null })` breaks, whereas `toStageIds` would have excluded it from `items`. Align on `!= null` everywhere.

frontend/src/community/crm/v2/components/molecules/DealStagesDraggableContent/DealStagesDraggableContent.tsx
```diff
@@ -0,0 +1,133 @@
...
+                    isTerminalStage={isTerminalStage(stage.stageType)}
+                    isDeletable={stage.stageType === CrmDealStageEnum.OPEN}
+                    isDraggable={stage.stageType === CrmDealStageEnum.OPEN}
+                  />
+                )
+            )}
+          </ul>
+        </SortableContext>
+      </DndContext>
+      <hr className="w-full border-t border-secondary-accent" />
```
ThinuwanW, line 108: 🤖 **Claude** · 🟣 **Nit**

The `<hr>` divider and the `dealCompleteStateNote` paragraph render unconditionally, even when `terminalStages` is empty (empty `stagesData`, or data that hasn't loaded and is passed as `[]`). That leaves a dangling rule plus an explanatory sentence with nothing beneath it. Guard the whole block with `terminalStages.length > 0 && (...)`.

frontend/src/community/crm/v2/components/molecules/DealStagesDraggableContent/DealStagesDraggableContent.tsx
```diff
@@ -0,0 +1,133 @@
...
+  onDelete: (stage: CrmStageEntity) => void;
+}
+
+const DealStagesDraggableContent = ({
+  stagesData,
+  onStagesReorder,
+  onEdit,
+  onDelete
+}: DealStagesDraggableContentProps) => {
+  const translateText = useTranslator("configurations", "crm");
```
ThinuwanW, line 36: 🤖 **Claude** · 🟣 **Nit**

Both new components pull strings from the `configurations.crm.*` i18n namespace (`useTranslator("configurations", "crm")`) while living in the crm module, whereas sibling crm/v2 components use `useTranslator("crmModule", ...)`. It works today because the keys exist in `configurations.json`, but it couples the crm module's rendering to the configurations translation file. If these components are meant to be crm-owned, consider moving `dealsSection.dealCompleteStateNote` and the `aria.editStage` / `aria.deleteStage` keys under `crmModule` (see the memory note about `~enterprise` aria namespaces silently falling back and rendering raw key strings — cross-namespace reads are exactly how that class of bug appears).

frontend/src/community/crm/v2/components/molecules/DraggableDealStageCard/DraggableDealStageCardSkeleton.tsx
```diff
@@ -0,0 +1,17 @@
+const DraggableDealStageCardSkeleton = () => {
+  return (
```
ThinuwanW, line 2: 🤖 **Claude** · 🟣 **Nit**

The skeleton conveys nothing to assistive tech during the loading window. Add `role="status"` with an `aria-label` (or `aria-hidden="true"` plus a visually-hidden live-region message elsewhere) so screen reader users are told content is loading. Also note the file lives in a folder named `DraggableDealStageCard/` rather than its own directory — minor, but it diverges from the one-folder-per-component layout used elsewhere under `crm/v2/components/molecules/`.

frontend/src/community/crm/v2/components/molecules/DraggableDealStageCard/DraggableDealStageCard.tsx
```diff
@@ -0,0 +1,108 @@
+import { useSortable } from "@dnd-kit/sortable";
+import { CSS } from "@dnd-kit/utilities";
+import { DragIcon, StatusCard } from "@rootcodelabs/skapp-ui";
+
+import Icon from "~community/common/components/atoms/Icon/Icon";
```
ThinuwanW, line 5: 🤖 **Claude** · 🔴 **Important**

This file imports the monolithic `~community/common/components/atoms/Icon/Icon` registry just to render two icons. That `Icon.tsx` is ~21KB of source that eagerly imports several hundred icon components at module scope, so importing it pulls the entire icon set into whatever route chunk renders this card. The rest of the crm/v2 module deliberately imports individual icons from the design system — this very file already does `import { DragIcon, StatusCard } from "@rootcodelabs/skapp-ui"` — and both icons needed here exist there (`EditIcon`, used in `~community/leave/.../LeaveTypesTable.tsx`; `DeleteButtonIcon`, used in `~community/crm/v2/components/molecules/CrmDeleteModalContent/CrmDeleteModalContent.tsx`). Fix: drop the `Icon` / `IconName` imports and use `import { DeleteButtonIcon, DragIcon, EditIcon, StatusCard } from "@rootcodelabs/skapp-ui"`, rendering `<EditIcon />` and `<DeleteButtonIcon />` directly.

frontend/src/community/crm/v2/components/molecules/DealStagesDraggableContent/DealStagesDraggableContent.tsx
```diff
@@ -0,0 +1,133 @@
...
+  );
+
+  const isTerminalStage = (stageType?: CrmDealStageEnum) =>
+    stageType === CrmDealStageEnum.WON || stageType === CrmDealStageEnum.LOST;
+
+  const initialStages = stagesData.filter(
+    (stage) => stage.stageType === CrmDealStageEnum.INITIAL
+  );
+
+  const draggableStages = stagesData.filter(
```
ThinuwanW, line 50: 🤖 **Claude** · 🔴 **Important**

Behavioral drift from the v1 component this was ported from: v1's `CrmDealStageType.stageType` was a required `CrmDealStageEnum`, but v2's `CrmStageEntity.stageType` is optional (`stageType?`, see CrmCommonTypes.ts:88 — the type is fully-optional because it doubles as a PATCH payload shape in `getChangedStageFields`/`updateDealStage`). A stage whose `stageType` is missing at runtime passes both filters here and lands in `draggableStages`, so its id goes into `SortableContext items`, but the card is then rendered with `isDraggable={stage.stageType === OPEN}` → false and `isDeletable` → false. Result: a row that is announced/registered as part of the sortable list but has no drag handle, cannot be deleted, and still gets carried into the array handed to `onStagesReorder` (and therefore into the reorder payload). Fix: either narrow the prop type so the component only accepts fully-resolved stages (e.g. `stagesData: (CrmStageEntity & Required<Pick<CrmStageEntity, "id" | "stageType">>)[]`, which also removes the need for the `stage.id !== undefined` guards and the separate `stageId` prop), or make the bucketing explicit and total: put only `stageType === OPEN` into `draggableStages` rather than defining it by exclusion.

frontend/src/community/crm/v2/components/molecules/DraggableDealStageCard/DraggableDealStageCard.tsx
```diff
@@ -0,0 +1,108 @@
...
+        }}
+        iconButtons={{
+          edit: {
+            icon: <Icon name={IconName.EDIT_ICON} />,
+            onClick: () => onEdit(stage),
+            "aria-label": translateText(["aria", "editStage"], {
+              stageName: stageName
+            })
+          },
+          ...(!isTerminalStage &&
```
ThinuwanW, line 91: 🤖 **Claude** · 🟡 **Suggestion**

The `!isTerminalStage` half of this guard is dead. The only caller sets `isDeletable={stage.stageType === CrmDealStageEnum.OPEN}`, and OPEN is mutually exclusive with WON/LOST, so `isDeletable === true` already implies `isTerminalStage === false`; the terminal list additionally never passes `onDelete` at all. That makes `isTerminalStage` a prop that exists purely to re-assert something the parent has already encoded in `isDeletable` — a third redundant source of truth in this props interface alongside `stageId`/`stage.id`. Fix: drop the `isTerminalStage` prop and gate the delete button on `isDeletable && onDelete` only; if a card really does need to know it is terminal, derive it from `stage.stageType` inside the component via the shared helper instead of accepting it as a prop.

frontend/src/community/crm/v2/components/molecules/DealStagesDraggableContent/DealStagesDraggableContent.tsx
```diff
@@ -0,0 +1,133 @@
...
+      <hr className="w-full border-t border-secondary-accent" />
+      <div className="flex flex-col gap-2">
+        <p className="body1 text-secondary-text mb-3">
+          {translateText(["dealsSection", "dealCompleteStateNote"])}
+        </p>
+        <ul className="flex flex-col gap-4">
+          {terminalStages.map(
+            (stage) =>
+              stage.id !== undefined && (
+                <DraggableDealStageCard
```
ThinuwanW, line 117: 🤖 **Claude** · 🟡 **Suggestion**

The terminal-stage `<ul>` is rendered outside the `<DndContext>` (which closes at line 107), yet each `DraggableDealStageCard` unconditionally calls `useSortable`. Outside a provider, dnd-kit falls back to its default internal context, so the `useDroppable` dispatch is a no-op and `useDraggable` writes into a module-level shared node map — no crash, but every WON/LOST card pays for hooks whose `attributes`/`listeners`/`transform` are then thrown away (the component renders the spacer `<div>` instead). This is dead machinery and it makes the component's name a lie for half its call sites. Fix: extract a presentational `DealStageCard` that renders the `<li>` + `StatusCard`, and a thin `DraggableDealStageCard` wrapper that calls `useSortable` and renders the handle — then the terminal list uses the plain card and never touches dnd-kit.

frontend/src/community/crm/v2/components/molecules/DraggableDealStageCard/DraggableDealStageCard.tsx
ThinuwanW, line 62: 🤖 **Claude** · 🟡 **Suggestion**

The drag handle has no `touch-action: none`. dnd-kit's `PointerSensor` relies on Pointer Events, and on touch devices the browser fires `pointercancel` the moment it decides the gesture is a scroll — which aborts the drag before it starts. dnd-kit's own docs require setting `touch-action: none` on the activator element when using `PointerSensor`. This is a separate concern from the sensor's activation constraint: without it, reordering stages is simply unusable on tablets/phones, and the configurations page is scrollable so the gesture will be claimed by the scroller. Fix: add `touch-none` to the button's className (or `style={{ touchAction: "none" }}`).

frontend/src/community/crm/v2/components/molecules/DraggableDealStageCard/DraggableDealStageCard.tsx
```diff
@@ -0,0 +1,108 @@
...
+    attributes,
+    listeners,
+    setNodeRef,
+    transform,
+    transition,
+    isDragging
+  } = useSortable({ id: stageId, disabled: !isDraggable });
+  const translateText = useTranslator("configurations", "crm");
+
+  const stageColor = stage.color ?? CrmDealStageColorsEnum.SKY;
```
ThinuwanW, line 42: 🤖 **Claude** · 🟡 **Suggestion**

Null-safety on the API shape is handled by silent fallbacks that degrade the UI rather than surfacing the problem. (a) `stage.color ?? CrmDealStageColorsEnum.SKY` renders a stage in the wrong colour indistinguishable from a genuinely sky-coloured stage. (b) `STAGE_COLOR_MAP[stageColor]` is typed `Record<CrmDealStageColorsEnum, string>`, so if the backend ever returns a colour token not in the frontend enum, TypeScript is satisfied but `code` is `undefined` at runtime and gets passed straight to `StatusCard`. (c) `getStageDisplayName(stage.name) ?? ""` means a nameless stage renders an empty title and, worse, produces the aria-labels "Edit stage " / "Delete stage " with a dangling interpolation — the a11y label silently loses its only distinguishing content. Fix: guard the map lookup (`STAGE_COLOR_MAP[stageColor] ?? STAGE_COLOR_MAP[CrmDealStageColorsEnum.SKY]`) and fall back the display name to a translated placeholder rather than `""` so the aria-label stays meaningful.

frontend/src/community/crm/v2/components/molecules/DraggableDealStageCard/DraggableDealStageCardSkeleton.tsx
```diff
@@ -0,0 +1,17 @@
...
+    <div className="animate-pulse flex flex-col gap-[18px] w-full">
```
ThinuwanW, line 3: 🤖 **Claude** · 🟡 **Suggestion**

This skeleton hand-rolls `animate-pulse bg-secondary-accent rounded-[8px]` blocks while the crm/v2 module already has exactly that atom: `~community/crm/v2/components/atoms/SkeletonShape/SkeletonShape.tsx`, used by every other v2 skeleton (SidePanelHeaderSkeleton, SidePanelTabsSkeleton, CompanySidePanelSkeleton, DealsKanbanBoardSkeletonV2). Two further points: the arbitrary values (`gap-[18px]`, `h-[60px]`, `h-[24px]`, `rounded-[8px]`) sidestep the Tailwind scale that the rest of the module uses (`h-9`, `w-75`, `rounded`), and the row count is hardcoded to 3 even though a pipeline always has at least INITIAL + WON + LOST plus custom stages, so the list visibly jumps on load. Fix: compose from `<SkeletonShape />` and accept a `count` prop with a sensible default, mirroring `DealCardSkeleton`'s `count` API.

frontend/src/community/crm/v2/components/molecules/DealStagesDraggableContent/DealStagesDraggableContent.tsx
ThinuwanW, line 75: 🤖 **Claude** · 🟡 **Suggestion**

The output contract fits the v1 architecture, not v2's. In v2, stages live in the Zustand store as `CrmStageRecord` (an id-keyed map — see crmDataSlice.ts:28/43), and display order is re-derived from `orderIndex` by `getOrderedStages` (commonUtil.ts:54). A positional `CrmStageEntity[]` therefore cannot round-trip through v2 state: whatever consumer receives it has to hold a parallel local array or immediately convert. Note also that `stageUtil.ts` — the very module this file already imports `toStageIds` from — exports `toStageReorderPayload`, which builds the `{ id, orderIndex }[]` request body from array position and correctly skips WON/LOST; it is currently unused by anything. Fix: have `onStagesReorder` emit `CrmDealStageReorderItem[]` via `toStageReorderPayload(...)` (or emit stages with recomputed `orderIndex`), so the consumer can feed `useReorderDealStages` directly instead of re-deriving the same rule a third time.

frontend/src/community/crm/v2/components/molecules/DealStagesDraggableContent/DealStagesDraggableContent.tsx
ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

None of the three new files are imported anywhere — grepping `DealStagesDraggableContent|DraggableDealStageCard` across `frontend/src` returns only the v1 `~community/configurations/...` copies and these new definitions. There is no v2 equivalent of `DealStagesContent`/`CrmConfigurations` wiring them to `useReorderDealStages` (which already exists at `~community/crm/v2/api/DealApi.ts:237`), and no v2 `DealStageModalController` behind the `onEdit`/`onDelete` callbacks. As shipped this is unreachable code that ESLint's unused-export rules won't catch and that will silently drift from the v1 original it was copied from. Either land the consumer in this PR, or state in the PR description which follow-up wires it so it doesn't sit unreferenced across releases.

frontend/src/community/crm/v2/components/molecules/DealStagesDraggableContent/DealStagesDraggableContent.tsx
```diff
@@ -0,0 +1,133 @@
...
+import { toStageIds } from "~community/crm/v2/utils/stageUtil";
+
+interface DealStagesDraggableContentProps {
+  stagesData: CrmStageEntity[];
+  onStagesReorder: (stages: CrmStageEntity[]) => void;
+  onEdit: (stage: CrmStageEntity) => void;
+  onDelete: (stage: CrmStageEntity) => void;
+}
+
+const DealStagesDraggableContent = ({
```
ThinuwanW, line 30: 🤖 **Claude** · 🟣 **Nit**

Neither new component annotates its return with `FC<Props>`, which is the dominant convention in this module (38 of 50 component files under `src/community/crm/v2/components` use `const X: FC<XProps> = ...`, including every sibling skeleton). Fix: `const DealStagesDraggableContent: FC<DealStagesDraggableContentProps> = ({ ... })` and likewise for `DraggableDealStageCard` and `DraggableDealStageCardSkeleton`, so the props type is checked at the declaration rather than only at the destructure.

frontend/src/community/crm/v2/components/molecules/DealStagesDraggableContent/DealStagesDraggableContent.tsx
ThinuwanW, line 1: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

No `.test.tsx` accompanies either component, despite `handleDragEnd` containing real branching logic (no-`over` bail, same-id bail, `-1` index bail, three-way concat of initial/reordered/terminal) that is easy to regress. The module already has Jest set up for logic (`src/community/crm/v2/utils/__test__/stageUtil.test.ts`, `hooks/useInitializeCrmData.test.ts`), and the repo has an established component-test pattern with `MockTheme` + `screen.getByRole`. At minimum, cover the bucketing (INITIAL pinned first, WON/LOST pinned to the terminal list) and the shape of the array handed to `onStagesReorder`, which is pure and testable without simulating a drag.

frontend/src/community/crm/v2/components/molecules/DraggableDealStageCard/DraggableDealStageCard.tsx
```diff
@@ -0,0 +1,108 @@
...
+          {...listeners}
+          {...attributes}
+          className="cursor-grab"
+          type="button"
+        >
+          <DragIcon />
+        </button>
+      )}
+      <StatusCard
+        id={String(stageId)}
```
ThinuwanW, line 72: 🤖 **Claude** · 🟣 **Nit**

`id={String(stageId)}` hands `StatusCard` a bare numeric string ("3", "7"). If that prop lands on a DOM `id` — which the name suggests — these are page-global identifiers that collide trivially with any other component on the configurations page that also keys off a raw entity id, and a numeric-leading id cannot be targeted by a plain CSS `#3` selector. Low cost to make it unambiguous: `id={`deal-stage-${stageId}`}`. Flagging as a nit since `@rootcodelabs/skapp-ui` isn't installed in this checkout, so I could not confirm how `StatusCard` consumes the prop.

PR #1846, opened 2026-08-28
refactor: add deal stage add and edit form for v2

dulsara-manakal:
Part 3 of 6 of the CRM deal-stage configuration refactor.

DealStageModalForm handles both add and edit through an isEdit prop: Formik form over name, description and colour, wired to useCreateDealStage and useUpdateDealStage, writing the edited stage back to the v2 store and calling onStageCreated so the caller can refetch after a create.

Builds on part 1 for stageValidations, getChangedStageFields, updateStage, DEAL_STAGE_COLORS and useStageNameMapper. Still behind isCrmConfigurationsV2 = false until part 6.

### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR adds a v2 `DealStageModalForm` that mirrors the existing `configurations` module version but sources data from the CRM v2 Zustand store instead of React Query. The i18n keys, Yup schema, regex and character limits all check out against `configurations.json` and the backend `CrmValidations`/`CrmDealStageController` contract. However, the port drops v1's `isEdit` guard around `selectedDealStage`, and the create path never writes the new stage back into the v2 store — which has no refetch path because the init query is disabled after first load — so a created stage will not appear until a full page reload.

Found **20** new issue(s): 🔴 5 important, 🟡 7 suggestion(s), 🟣 8 nit(s)

frontend/src/community/crm/v2/components/molecules/DealStageModalForm/DealStageModalForm.tsx
```diff
@@ -0,0 +1,232 @@
...
+    );
+
+  const orderedStages = getOrderedStages(stages);
+
+  const selectedDealStage = getSelectedStage(stages, selectedDealStageId);
+
+  const initialValues: CrmStageEntity = {
+    name: isEdit ? (getStageDisplayName(selectedDealStage?.name) ?? "") : "",
+    description: isEdit ? (selectedDealStage?.description ?? "") : "",
+    color: selectedDealStage?.color ?? CrmDealStageColorsEnum.SKY
```
ThinuwanW, line 62: 🤖 **Claude** · 🔴 **Important**

`color` is read from `selectedDealStage` without the `isEdit` guard that `name` (line 60) and `description` (line 61) have — the v1 component this is ported from used `color: isEdit ? selectedDealStage!.color : CrmDealStageColorsEnum.SKY`. The same omission appears at line 153, where `selectedDealStage?.id` is passed as `currentStageId` to `getStageValidationSchema` in both modes. Callers are expected to null out `selectedDealStageId` before opening the add modal (that is exactly what `CrmConfigurations.handleAddStage` does for v1), but nothing in the component enforces it. If a caller opens the add modal without resetting the id, the add form pre-fills the previously edited stage's colour AND `isStageNameTaken` excludes that stage from the duplicate check, letting a duplicate name through the client-side guard. Fix: `color: isEdit ? (selectedDealStage?.color ?? CrmDealStageColorsEnum.SKY) : CrmDealStageColorsEnum.SKY` and pass `isEdit ? selectedDealStage?.id : undefined` to the validation schema.

frontend/src/community/crm/v2/components/molecules/DealStageModalForm/DealStageModalForm.tsx
```diff
@@ -0,0 +1,232 @@
...
+  };
+
+  const handleCreateSuccess = () => {
```
ThinuwanW, line 65: 🤖 **Claude** · 🔴 **Important**

`handleCreateSuccess` discards the created stage and never updates the store, unlike `handleEditSuccess` (line 70) which calls `setStages(updateStage(...))`. `useCreateDealStage`'s `onSuccess` does receive `createdStage`, and `onStageCreated` is typed `() => void` so the parent cannot obtain it either. Stages are only ever populated by `useInitializeCrmData`, whose `useGetBoardInitData(!isCrmDataInitialized)` query is disabled once `isCrmDataInitialized` is true, so there is no refetch path — after creating a stage the store (and therefore the kanban lanes and the duplicate-name check at line 152) will not contain it until a hard reload. Note also that the backend `resolveOrderIndexForNewStage` shifts the `orderIndex` of the WON/LOST stages, so a plain insert of the created stage still leaves cached order indexes stale. Either accept the created stage and upsert it (`handleCreateSuccess = (createdStage: CrmStageEntity) => { if (createdStage.id !== undefined) setStages({ ...stages, [createdStage.id]: createdStage }); ... }`) plus fix up the terminal stages' order, or change `onStageCreated` to `(createdStage: CrmStageEntity) => void` and have the parent re-run initialisation (e.g. reset `isCrmDataInitialized`).

frontend/src/community/crm/v2/components/molecules/DealStageModalForm/DealStageModalForm.tsx
```diff
@@ -0,0 +1,232 @@
...
+    useCreateDealStage(handleCreateSuccess, handleError);
+
+  const { mutate: updateDealStage, isPending: isUpdatePending } =
+    useUpdateDealStage(handleEditSuccess, handleError);
+
+  const handleEdit = (values: CrmStageEntity) => {
+    const changedFields = getChangedStageFields(values, initialValues);
+    if (
+      Object.keys(changedFields).length === 0 ||
+      selectedDealStage?.id === undefined
```
ThinuwanW, line 130: 🤖 **Claude** · 🟡 **Suggestion**

When `selectedDealStage?.id === undefined` in edit mode the form silently closes the modal, discarding whatever the user typed, with no error toast. This is not just theoretical: nothing in the v2 codebase ever calls `useCrmStoreV2`'s `setSelectedDealStageId`, so `selectedDealStageId` is always `null` today — an edit modal would render with blank name/description and then throw the user's input away on save. Split the two conditions: keep the quiet close for the genuine "no fields changed" case, and surface `handleError()` (or don't render the form at all) when the selected stage cannot be resolved.

frontend/src/community/crm/v2/components/molecules/DealStageModalForm/DealStageModalForm.tsx
```diff
@@ -0,0 +1,232 @@
...
+      return;
+    }
+
+    updateDealStage({ id: selectedDealStage.id, ...changedFields });
+  };
+
+  const handleCreate = (values: CrmStageEntity) => {
+    createDealStage({
+      name: values.name?.trim(),
+      description: values.description?.trim(),
```
ThinuwanW, line 142: 🤖 **Claude** · 🟡 **Suggestion**

Behaviour change vs. the v1 form, which sent `description: values.description.trim() || null` on create. Here an untouched description is posted as `""`. The backend `CrmValidations.validateDealStageDescription` short-circuits on `isBlank`, so `CrmDealStage.description` is persisted as an empty string rather than NULL, giving a mix of `""` and `null` in the column depending on which form created the row. Use `values.description?.trim() || undefined` so the key is omitted from the JSON body and the backend receives `null`.

frontend/src/community/crm/v2/components/molecules/DealStageModalForm/DealStageModalForm.tsx
```diff
@@ -0,0 +1,232 @@
...
+        setIsDealStageModalOpen: store.setIsDealStageModalOpen,
+        selectedDealStageId: store.selectedDealStageId
+      }))
+    );
+
+  const orderedStages = getOrderedStages(stages);
+
+  const selectedDealStage = getSelectedStage(stages, selectedDealStageId);
+
+  const initialValues: CrmStageEntity = {
```
ThinuwanW, line 59: 🤖 **Claude** · 🟡 **Suggestion**

Using the API entity type `CrmStageEntity` as the Formik values type is too loose — every field on it is optional, so `values.name` and `values.color` are `string | undefined` where the backend requires both (`validateDealStageName` / `validateDealStageColor` throw on null). That leaks into `value={values.name}` (line 178) and `selectedColorId={values.color}` (line 201), and `createDealStage({ name: values.name?.trim(), color: values.color })` (lines 141-143) type-checks even though it can produce an invalid payload. Related: `setFieldValue("color", color.value)` on line 202 writes an untyped `string` into a field declared as `CrmDealStageColorsEnum`, so nothing verifies the ColorSelector option actually maps to an enum member. v1 had a dedicated `CrmDealStageFormTypes`; define the equivalent here (`{ name: string; description: string; color: CrmDealStageColorsEnum }`) and narrow `color.value` when setting it.

frontend/src/community/crm/v2/components/molecules/DealStageModalForm/DealStageModalForm.tsx
```diff
@@ -0,0 +1,232 @@
...
+
+  const handleEditSuccess = (updatedStage: CrmStageEntity) => {
+    if (updatedStage.id !== undefined) {
+      setStages(updateStage(stages, updatedStage.id, updatedStage));
+    }
+    handleSuccess();
+  };
+
+  const handleSuccess = () => {
+    setSubmitting(false);
```
ThinuwanW, line 78: 🤖 **Claude** · 🟣 **Nit**

`setSubmitting(false)` here and on line 98 is dead code: `isSubmitting` is never read — both buttons are driven by `isPending` from the mutations (line 170) — and the early-return path at line 132 never resets it anyway. Drop `setSubmitting` from the destructure and both call sites, or switch the buttons to `isPending || isSubmitting` if the intent was to also block while Yup validation runs.

frontend/src/community/crm/v2/components/molecules/DealStageModalForm/DealStageModalForm.tsx
```diff
@@ -0,0 +1,232 @@
...
+    handleBlur,
+    setSubmitting,
+    setFieldValue,
+    submitForm
+  } = formik;
+
+  const isPending = isCreatePending || isUpdatePending;
+
+  return (
+    <div className="flex flex-col h-full justify-between gap-[0.625rem]">
```
ThinuwanW, line 173: 🤖 **Claude** · 🟣 **Nit**

The root element is a `<div>` and the save button is `type="button"` with `onClick={submitForm}`, so pressing Enter in the name field does nothing and the fields have no owning form for assistive tech. Wrap the content in `<form onSubmit={formik.handleSubmit}>` and change the save button to `type="submit"` (the cancel button correctly stays `type="button"`).

frontend/src/community/crm/v2/components/molecules/DealStageModalForm/DealStageModalForm.tsx
```diff
@@ -0,0 +1,232 @@
...
+      name: values.name?.trim(),
+      description: values.description?.trim(),
+      color: values.color
+    });
+  };
+
+  const formik = useFormik({
+    initialValues,
+    onSubmit: isEdit ? handleEdit : handleCreate,
+    validationSchema: getStageValidationSchema(
```
ThinuwanW, line 150: 🤖 **Claude** · 🟣 **Nit**

`getStageValidationSchema(...)` builds a fresh Yup schema on every render, and `getOrderedStages(stages)` on line 55 re-sorts the whole stage record on every render. Both are cheap in isolation but run on each keystroke since `handleChange` re-renders the form. Wrap them in `useMemo` (`[stages]` for the ordered list, `[translateText, orderedStages, selectedDealStage?.id]` for the schema).

frontend/src/community/crm/v2/components/molecules/DealStageModalForm/DealStageModalForm.tsx
```diff
@@ -0,0 +1,232 @@
+import {
```
ThinuwanW, line 1: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

No `DealStageModalForm.test.tsx` accompanies the new component. The repo does have component tests (e.g. `src/community/attendance/components/molecules/AttendanceCard/AttendanceCard.test.tsx`) even though crm/v2 currently only tests utils and hooks. Worth covering: add vs. edit initial values, the duplicate-name Yup rule against default stages such as `DEMO_SCHEDULED`, and the "no fields changed" close path — these are exactly the branches flagged above.

frontend/src/community/crm/v2/components/molecules/DealStageModalForm/DealStageModalForm.tsx
```diff
@@ -0,0 +1,232 @@
...
+    handleChange,
+    handleBlur,
+    setSubmitting,
+    setFieldValue,
+    submitForm
+  } = formik;
+
+  const isPending = isCreatePending || isUpdatePending;
+
+  return (
```
ThinuwanW, line 172: 🤖 **Claude** · 🟣 **Nit**

The entire JSX block (lines 172-229) is byte-for-byte identical to `src/community/configurations/components/molecules/DealStageModalForm/DealStageModalForm.tsx` — only the data sources differ. If both versions will coexist for a while, consider extracting the presentational part into a shared molecule that takes formik state plus the colour list as props, so a label/layout fix does not have to be made twice.

frontend/src/community/crm/v2/components/molecules/DealStageModalForm/DealStageModalForm.tsx
```diff
@@ -0,0 +1,232 @@
...
+  isEdit?: boolean;
+  onStageCreated?: () => void;
+}
+
+const DealStageModalForm: FC<DealStageModalFormProps> = ({
+  isEdit = false,
+  onStageCreated
+}) => {
+  const { setToastMessage } = useToast();
+  const translateText = useTranslator("configurations", "crm");
```
ThinuwanW, line 42: 🤖 **Claude** · 🟣 **Nit**

A component under `~community/crm/v2` reads its strings from the `configurations` i18n namespace while `useStageNameMapper` (line 43) reads from `crmModule`. Both key sets resolve today, but it means CRM v2 stage-modal copy lives in `configurations.json` and is easy to miss when the CRM strings are updated. Consider mirroring the `dealStageModal`/`addDealStageModal`/`editDealStageModal` blocks under `crmModule` for the v2 component, or leave a comment noting the deliberate cross-module namespace.

frontend/src/community/crm/v2/components/molecules/DealStageModalForm/DealStageModalForm.tsx
ThinuwanW, line 1: 🤖 **Claude** · 🔴 **Important**

📁 **File-level observation**

This component is never rendered and its paired wiring is missing. A repo-wide grep for `DealStageModalForm` finds only the v1 import in `configurations/components/organisms/DealStageModalController`; there is no v2 `DealStageModalController` in `crm/v2/components/organisms/` (only AddDealSidePanelV2, CompanyModalController, CompanySidePanel, CompanyTable, DealSidePanelV2, DealsKanbanBoardV2, DealsSectionV2, DealsTableV2). Worse, nothing anywhere in `src/community/crm/v2` ever calls `setSelectedDealStageId` or `setDealStageModalType` — both setters exist in `crmUiSlice.ts` but have zero call sites. Consequence: `selectedDealStageId` is permanently `null`, so `getSelectedStage` returns `undefined`, edit mode seeds a blank form and `handleEdit` hits the `selectedDealStage?.id === undefined` branch and silently closes every time. Also missing: the v1 add-stage entry point wraps the modal open in `guardCrmCreate(CrmLimitResource.DEAL_STAGES, () => { setSelectedDealStageId(null); ... })` (CrmConfigurations.tsx:130-137); without an equivalent v2 trigger the plan stage limit is only enforced by the server. Ship the v2 controller + a limit-guarded trigger that sets `selectedDealStageId`/`dealStageModalType` in this PR, otherwise this file is dead code.

frontend/src/community/crm/v2/components/molecules/DealStageModalForm/DealStageModalForm.tsx
```diff
@@ -0,0 +1,232 @@
...
+    });
+  };
+
+  const formik = useFormik({
+    initialValues,
+    onSubmit: isEdit ? handleEdit : handleCreate,
+    validationSchema: getStageValidationSchema(
+      translateText,
+      orderedStages,
+      selectedDealStage?.id
```
ThinuwanW, line 153: 🤖 **Claude** · 🔴 **Important**

`currentStageId` is passed to `getStageValidationSchema` as `selectedDealStage?.id` without an `isEdit` guard, so in ADD mode a leftover `selectedDealStageId` in the store excludes that stage from the `is-deal-stage-name-unique` test (`isStageNameTaken` skips `stage.id === currentStageId`, stageUtil.ts:98) and the user can create a second stage with a name identical to the last-edited one. v1 only avoided this because the caller explicitly resets it — `CrmConfigurations.handleAddStage` calls `setSelectedDealStageId(null)` before opening the add modal (CrmConfigurations.tsx:132-134). No v2 caller does that today, and relying on the caller to null a shared store field is fragile. Fix inside the component: pass `isEdit ? selectedDealStage?.id : undefined`. This is the same root cause as the unguarded `color` on line 62 but a separate, security/data-integrity-relevant consequence.

frontend/src/community/crm/v2/components/molecules/DealStageModalForm/DealStageModalForm.tsx
```diff
@@ -0,0 +1,232 @@
...
+      ])
+    });
+  };
+
+  const handleCloseModal = (): void => {
+    setIsDealStageModalOpen(false);
+  };
+
+  const { mutate: createDealStage, isPending: isCreatePending } =
+    useCreateDealStage(handleCreateSuccess, handleError);
```
ThinuwanW, line 121: 🤖 **Claude** · 🔴 **Important**

Neither mutation invalidates any stage cache. In v1, `useCreateDealStage` and `useUpdateDealStage` both call `queryClient.invalidateQueries({ queryKey: crmDealQueryKeys.DEAL_STAGES })` (crmDealApi.ts:240 and :269). The v2 equivalents in `crm/v2/api/DealApi.ts` invalidate only `crmLimitationQueryKeys.GET_CRM_LIMITATION` on create (line 197) and nothing at all on update (lines 217-225). In v2, stage data is hydrated once from `crmBoardQueryKeys.BOARD_INIT_DATA` into Zustand via `useInitializeCrmData` (gated by `isCrmDataInitialized`), so after a create or an edit both `BOARD_INIT_DATA` and `crmDealQueryKeys.DEAL_STAGES` hold stale stage rows and the component compensates with a hand-written store patch on the edit path only. The review guidelines call for `invalidateQueries` after mutations that affect cached data and prefer invalidation over manual cache updates. Add `queryClient.invalidateQueries({ queryKey: crmBoardQueryKeys.BOARD_INIT_DATA })` (and `crmDealQueryKeys.DEAL_STAGES`) to both v2 stage mutations so every consumer converges.

frontend/src/community/crm/v2/components/molecules/DealStageModalForm/DealStageModalForm.tsx
ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

This file diverges from the container/presenter split that the rest of `crm/v2` already uses. The established pattern is a pair of container molecules — `AddCompanyModalContent.tsx` and `EditCompanyModalContent.tsx` — that each own `useFormik`, the store selectors and the mutation, plus a purely presentational `CompanyModalForm.tsx` that takes `{ formik, isPending, translateText, onCancel }` as props and renders only markup. This component is named `*ModalForm` (matching the presentational sibling) but is actually a container, and it collapses add + edit into a single `isEdit` boolean with a create-only `onStageCreated` prop that edit mode silently ignores. Restructure as `AddDealStageModalContent` / `EditDealStageModalContent` + a props-driven `DealStageModalForm`, which also removes the need for the `isEdit ? … : …` ternaries scattered through `initialValues`, `handleSuccess`, `handleError` and `onSubmit`, and lets each variant use its own translator sub-namespace as `AddCompanyModalContent` does.

frontend/src/community/crm/v2/components/molecules/DealStageModalForm/DealStageModalForm.tsx
```diff
@@ -0,0 +1,232 @@
...
+    setSubmitting,
+    setFieldValue,
+    submitForm
+  } = formik;
+
+  const isPending = isCreatePending || isUpdatePending;
+
+  return (
+    <div className="flex flex-col h-full justify-between gap-[0.625rem]">
+      <InputField
```
ThinuwanW, line 174: 🤖 **Claude** · 🟡 **Suggestion**

Missing accessibility labels and input length caps that the sibling v2 form applies. `CompanyModalForm.tsx` puts `aria-label={translateText(["ariaLabels", …])}` on every `InputField`/`Dropdown` and on both `ButtonV2`s (lines 94, 109, 122, 135, 146, 158, 168) and caps input with `maxLength={characterLengths.COMPANY_NAME_LENGTH}` (line 95). Here none of the three controls and neither button carries an aria-label, and there is no `maxLength` on the name field (Yup caps it at `characterLengths.DEAL_STAGE_NAME_LENGTH` = 50) or the description (`DEAL_STAGE_DESCRIPTION_LENGTH` = 255) — with `validateOnChange: false` the user only discovers the cap on blur, after typing well past it. Add `maxLength` to both fields and aria-labels sourced from new `configurations.crm.dealStageModal.ariaLabels.*` keys. Note the existing `configurations.crm.aria` block only has `editStage`/`deleteStage`, so new keys are needed.

frontend/src/community/crm/v2/components/molecules/DealStageModalForm/DealStageModalForm.tsx
```diff
@@ -0,0 +1,232 @@
...
+        <ButtonV2
+          variant="primary"
+          type="button"
+          onClick={submitForm}
+          disabled={isPending}
+          isLoading={isPending}
+        >
+          {translateText(["dealStageModal", "buttons", "save"])}
+        </ButtonV2>
+      </div>
```
ThinuwanW, line 227: 🤖 **Claude** · 🟡 **Suggestion**

The save button is never disabled for an untouched form. `CompanyModalForm.tsx:166` uses `disabled={isPending || isSubmitting || isAlreadyNameExists || !dirty}`. Here `dirty` is not even destructured from formik, so in edit mode clicking Save with no changes runs `getChangedStageFields`, finds nothing and just closes the modal — visually indistinguishable from a successful save, and no success toast fires either. In add mode it submits an empty form purely to trigger validation errors. Destructure `dirty` from `formik` and add `|| !dirty` to the save button's `disabled` prop; that also makes the `Object.keys(changedFields).length === 0` early-return on line 129 effectively unreachable rather than a silent no-op path.

frontend/src/community/crm/v2/components/molecules/DealStageModalForm/DealStageModalForm.tsx
```diff
@@ -0,0 +1,232 @@
...
+    useCrmStoreV2(
+      useShallow((store) => ({
+        stages: store.stages,
+        setStages: store.setStages,
+        setIsDealStageModalOpen: store.setIsDealStageModalOpen,
+        selectedDealStageId: store.selectedDealStageId
+      }))
+    );
+
+  const orderedStages = getOrderedStages(stages);
```
ThinuwanW, line 55: 🤖 **Claude** · 🟡 **Suggestion**

`orderedStages` is fed to the uniqueness validator with RAW backend stage names, while the input field holds DISPLAY names. Line 60 seeds `values.name` with `getStageDisplayName(...)` (which maps `DefaultStageNameEnum` members through `crmModule.deals.defaultStageNames`), but `getOrderedStages(stages)` returns the untranslated store entities, so `isStageNameTaken` compares e.g. the typed "Demo Scheduled" against the stored "DEMO_SCHEDULED". It happens to work today only because `toStandardStageName` uppercases and underscore-joins (stageUtil.ts:86-87) and the English translations are exact title-case echoes of the enum keys — that equivalence breaks the moment a second locale is added (currently only `assets/languages/english` exists) or a default stage's English label is reworded, at which point a user can create a stage whose displayed name is identical to a default stage. v1 avoided this by passing display-mapped stages: `useGetMappedDealStages` maps `name: getStageByName(stage.name)` before handing them to `dealStageValidations`. Map `orderedStages` through `getStageDisplayName` before passing it in.

frontend/src/community/crm/v2/components/molecules/DealStageModalForm/DealStageModalForm.tsx
```diff
@@ -0,0 +1,232 @@
...
+  };
+
+  const { mutate: createDealStage, isPending: isCreatePending } =
+    useCreateDealStage(handleCreateSuccess, handleError);
+
+  const { mutate: updateDealStage, isPending: isUpdatePending } =
+    useUpdateDealStage(handleEditSuccess, handleError);
+
+  const handleEdit = (values: CrmStageEntity) => {
+    const changedFields = getChangedStageFields(values, initialValues);
```
ThinuwanW, line 127: 🤖 **Claude** · 🟣 **Nit**

Edit-path counterpart to the create-path empty-string issue: `getChangedStageFields` (stageUtil.ts:37-39) assigns `changedFields.description = currentValues.description?.trim()`, so clearing an existing description PATCHes `description: ""`. v1's `getChangedDealStageFields` normalised empty to `null` on both sides of the comparison (`crmUtil.ts:63-64`, `newDescription = newValues.description.trim() || null`). Two consequences: the cleared field is persisted as an empty string rather than nulled, and the comparison itself differs — going from `null`/absent to `""` is now treated as no-change on one side and a change on the other depending on what the backend returned. Normalise with `|| null` in `getChangedStageFields` to match v1, or confirm the backend treats `""` and `null` identically for this column.

frontend/src/community/crm/v2/components/molecules/DealStageModalForm/DealStageModalForm.tsx
```diff
@@ -0,0 +1,232 @@
...
+        errorMessage={errors?.description}
+        onChange={handleChange}
+        onBlur={handleBlur}
+      />
+
+      <ColorSelector
+        id="status-color-selector"
+        label={translateText(["dealStageModal", "colorInputLabel"])}
+        selectedColorId={values.color}
+        onColorChange={(color) => setFieldValue("color", color.value)}
```
ThinuwanW, line 202: 🤖 **Claude** · 🟣 **Nit**

`color` has no Yup validation at all (`getStageValidationSchema` only defines `name` and `description`), and `onColorChange` writes `color.value` — a plain `string` from `ColorOption` — into a field typed `CrmDealStageColorsEnum`, with formik's untyped `setFieldValue` silently accepting it. It is correct at runtime only because `DEAL_STAGE_COLORS` is derived from `STAGE_COLOR_MAP`, which is keyed by `Record<CrmDealStageColorsEnum, string>`, so nothing enforces that contract if the colour list is ever sourced differently. Add `color: Yup.mixed<CrmDealStageColorsEnum>().oneOf(Object.values(CrmDealStageColorsEnum)).required()` to the schema, and type the callback parameter so `color.value` is narrowed to the enum rather than `string`. The returned promise from `setFieldValue` is also floating — `void setFieldValue(...)` if the lint config allows it.

PR #1847, opened 2026-08-28
refactor: add deal stage delete modal and modal controller for v2

dulsara-manakal:
Part 4 of 6 of the CRM deal-stage configuration refactor.

- DeleteDealStageModalContent: confirmation body wired to useDeleteDealStage, removing the stage from the v2 store on success
- DealStageModalController: single SmallModal that picks its title and body from dealStageModalType - add, edit or delete

The controller needs both the form from part 3 and the delete body above, so they land together.
Still behind isCrmConfigurationsV2 = false until part 6.

### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This PR ports the v1 configurations deal-stage delete modal and modal controller into the CRM v2 module, backed by the `useCrmStoreV2` Zustand store instead of a per-id query. The port is faithful, but it bypasses the module's existing shared `CrmDeleteModalContent` component and consequently drops two behaviours that component provides (spinner on confirm, cancel disabled while pending) — the latter lets the user dismiss the modal mid-delete and silently lose the store update and toast. The store cleanup on success is also incomplete: only `stages` is pruned, leaving an orphaned `board` column for the deleted stage that the Kanban drag logic scans.

Found **20** new issue(s): 🔴 6 important, 🟡 5 suggestion(s), 🟣 9 nit(s)

frontend/src/community/crm/v2/components/molecules/DeleteDealStageModalContent/DeleteDealStageModalContent.tsx
```diff
@@ -0,0 +1,135 @@
...
+        })}
+      </div>
+      <div className="flex flex-row justify-end py-[0.85rem] gap-[1rem]">
+        <ButtonV2
+          variant="tertiary"
+          type="button"
+          onClick={handleCloseModal}
+          icon={<CloseIcon />}
+          iconPosition="end"
+          aria-label={translateText([
```
ThinuwanW, line 101: 🤖 **Claude** · 🔴 **Important**

Pending state is only half-handled. The confirm button gets `disabled={isPending}` but no `isLoading={isPending}` (so there is no spinner), and the Cancel button is not disabled at all while the delete is in flight. The modal's own close (SmallModal `onClose` in DealStageModalController) is likewise ungated. If the user clicks Cancel (or the modal X) after firing the delete, this component unmounts, the React Query mutation observer is removed, and `handleSuccess` never runs — the stage is deleted on the server but remains in the `stages` store and no success toast is shown, so the config list keeps showing a stage that no longer exists until a full reload. Compare `CrmDeleteModalContent` (frontend/src/community/crm/v2/components/molecules/CrmDeleteModalContent/CrmDeleteModalContent.tsx), which sets `disabled={isPending}` on both buttons and `isLoading={isPending}` on confirm. Fix: add `disabled={isPending}` to the cancel button and `isLoading={isPending}` to the confirm button (or reuse `CrmDeleteModalContent`, see the separate duplication comment).

frontend/src/community/crm/v2/components/molecules/DeleteDealStageModalContent/DeleteDealStageModalContent.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+  const handleSuccess = () => {
+    if (selectedDealStage?.id !== undefined) {
+      setStages(removeStage(stages, selectedDealStage.id));
+    }
+
+    setToastMessage({
+      open: true,
+      toastType: ToastType.SUCCESS,
+      title: translateText([
```
ThinuwanW, line 45: 🤖 **Claude** · 🔴 **Important**

On success only the `stages` record is pruned; the `board` record (`CrmBoardRecord`, keyed by stage id) still holds a column for the deleted stage, and the deals it listed stay in `deals`. Because the store is module-level and `isCrmDataInitialized` stays true across client-side navigation, that orphan column survives: `ingestBoardStageDeals` spreads `...current.board` even when `append: false`, so it is never dropped. This is reachable — delete a stage on the configurations page (backend reassigns its deals), navigate back to the board, then drag one of those reassigned deals: `handleDragEnd` calls `findStageIdByDealId(snapshot, activeId)` (useKanbanDragV2.ts:105), which iterates `Object.entries(board)` in ascending numeric-key order and will return the *deleted* stage id if it is lower than the deal's real stage. The move then computes `fromStageId` = deleted stage, so the card is never removed from its real column and ends up rendered twice with a wrong `totalCount`. Fix: mirror the `removeDeal` pattern from DeleteDealModalV2 — extend the success handler (or add a `removeStageFromBoard` helper in `stageUtil`) to also `setBoardColumn` without the deleted stage key and drop its deal ids from `deals`/`dealIds`.

frontend/src/community/crm/v2/components/molecules/DeleteDealStageModalContent/DeleteDealStageModalContent.tsx
```diff
@@ -0,0 +1,135 @@
...
+    }
+  };
+
+  return (
+    <div className="flex flex-col">
+      <div>
+        {translateText(["deleteDealStageModal", "description"], {
+          stageName: getStageDisplayName(selectedDealStage?.name)
+        })}
+      </div>
```
ThinuwanW, line 93: 🤖 **Claude** · 🟡 **Suggestion**

This JSX is a byte-for-byte duplicate of the shared `CrmDeleteModalContent` molecule that already exists in this exact directory (`~community/crm/v2/components/molecules/CrmDeleteModalContent/CrmDeleteModalContent`) and is the established v2 pattern — `DeleteCompanyModalContent` and `DeleteDealModalV2` both delegate to it. Duplicating it is what caused the missing `isLoading`/cancel-disabled behaviour flagged above, and it will drift again. Replace the inline markup with `<CrmDeleteModalContent description={...} isPending={isPending} confirmLabel={...} cancelLabel={...} confirmAriaLabel={...} cancelAriaLabel={...} onConfirm={handleDeleteStage} onClose={handleCloseModal} />`.

frontend/src/community/crm/v2/components/molecules/DeleteDealStageModalContent/DeleteDealStageModalContent.tsx
```diff
@@ -0,0 +1,135 @@
...
+        "successDescription"
+      ])
+    });
+    handleCloseModal();
+  };
+
+  const handleError = () => {
+    setToastMessage({
+      open: true,
+      toastType: ToastType.ERROR,
```
ThinuwanW, line 62: 🤖 **Claude** · 🟡 **Suggestion**

`selectedDealStageId` is left pointing at the just-deleted stage after a successful delete. `DeleteCompanyModalContent` explicitly calls `setSelectedCompanyId(null)` in its success handler, and `DeleteDealModalV2` calls `setSelectedDealId(null)`, for exactly this reason. The stale id makes `getSelectedStage(stages, selectedDealStageId)` return `undefined` on the next render (so the description interpolates an empty stage name during the close transition) and leaves a dangling reference for any later consumer. Add `setSelectedDealStageId(null)` after `handleCloseModal()` (pull the setter from the `useShallow` selector). Also worth checking `preselectedStageId` for the same dangling-reference problem.

frontend/src/community/crm/v2/components/molecules/DeleteDealStageModalContent/DeleteDealStageModalContent.tsx
```diff
@@ -0,0 +1,135 @@
...
+          onClick={handleCloseModal}
+          icon={<CloseIcon />}
+          iconPosition="end"
+          aria-label={translateText([
+            "deleteDealStageModal",
+            "buttons",
+            "cancel"
+          ])}
+        >
+          {translateText(["deleteDealStageModal", "buttons", "cancel"])}
```
ThinuwanW, line 107: 🤖 **Claude** · 🟣 **Nit**

Both `aria-label` values are the same strings as the visible button text (`deleteDealStageModal.buttons.cancel` / `.confirm`), which makes them a no-op — the accessible name was already correct. The v2 convention elsewhere uses dedicated keys: `DeleteCompanyModalContent` passes `translateText(["ariaLabels", "confirm"])` / `["ariaLabels", "cancel"]`. Either drop the `aria-label` props entirely, or add a `deleteDealStageModal.ariaLabels` block to configurations.json with genuinely descriptive labels (e.g. "Delete stage {{stageName}}", matching the existing `crm.aria.deleteStage` key).

frontend/src/community/crm/v2/components/molecules/DeleteDealStageModalContent/DeleteDealStageModalContent.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+  return (
+    <div className="flex flex-col">
+      <div>
+        {translateText(["deleteDealStageModal", "description"], {
+          stageName: getStageDisplayName(selectedDealStage?.name)
+        })}
+      </div>
+      <div className="flex flex-row justify-end py-[0.85rem] gap-[1rem]">
+        <ButtonV2
```
ThinuwanW, line 95: 🤖 **Claude** · 🟣 **Nit**

The description wrapper is a bare `<div>` with no typography class, while every other CRM delete modal renders it as `<div className="body1">` (see CrmDeleteModalContent.tsx:27). This will render at the default font size instead of the design-system body style. Add `className="body1"` (or reuse the shared component).

frontend/src/community/crm/v2/components/molecules/DeleteDealStageModalContent/DeleteDealStageModalContent.tsx
```diff
@@ -0,0 +1,135 @@
...
+    handleError
+  );
+
+  const handleDeleteStage = () => {
+    if (selectedDealStage?.id !== undefined) {
+      deleteStage(selectedDealStage.id);
+    }
+  };
+
+  return (
```
ThinuwanW, line 87: 🤖 **Claude** · 🟣 **Nit**

`handleDeleteStage` silently does nothing when `selectedDealStage?.id` is `undefined` — the user clicks Delete and gets no spinner, no toast, no close. Similarly `getStageDisplayName(selectedDealStage?.name)` returns `string | undefined`, which i18next interpolates as an empty string, so the confirmation reads "delete the '' stage". Consider rendering an error toast (or not opening the modal at all) when the selected stage cannot be resolved, so the failure is visible rather than a dead button.

frontend/src/community/crm/v2/components/organisms/DealStageModalController/DealStageModalController.tsx
```diff
@@ -0,0 +1,69 @@
...
+import { useShallow } from "zustand/react/shallow";
+
+import { useTranslator } from "~community/common/hooks/useTranslator";
+import DealStageModalForm from "~community/crm/v2/components/molecules/DealStageModalForm/DealStageModalForm";
+import DeleteDealStageModalContent from "~community/crm/v2/components/molecules/DeleteDealStageModalContent/DeleteDealStageModalContent";
+import { useCrmStoreV2 } from "~community/crm/v2/store/store";
+import { CrmModalTypes } from "~community/crm/v2/types/CrmTypes";
+
+interface DealStageModalControllerProps {
+  onStageCreated: () => void;
```
ThinuwanW, line 12: 🤖 **Claude** · 🟣 **Nit**

`onStageCreated` is declared as a required prop but is only consumed by the ADD branch; `DealStageModalForm` itself declares it as optional (`onStageCreated?: () => void`). Any future caller that only needs edit/delete will be forced to pass a no-op. Make it optional (`onStageCreated?: () => void`) to match the form's contract.

frontend/src/community/crm/v2/components/organisms/DealStageModalController/DealStageModalController.tsx
```diff
@@ -0,0 +1,69 @@
+import { SmallModal } from "@rootcodelabs/skapp-ui";
```
ThinuwanW, line 1: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

Neither `DealStageModalController` nor `DeleteDealStageModalContent` is imported anywhere yet — the only live consumer of the stage-config flow is still `~community/configurations/components/organisms/CrmConfigurations/CrmConfigurations.tsx`, which renders the v1 controller. Both new files are unreachable dead code until the v2 configurations page is wired up. Fine if that's a follow-up PR, but worth confirming it is tracked so this doesn't sit unexercised; note that the `isLoading`/board-cleanup issues above will only surface once it is wired in.

frontend/src/community/crm/v2/components/molecules/DeleteDealStageModalContent/DeleteDealStageModalContent.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+  const selectedDealStage = getSelectedStage(stages, selectedDealStageId);
+
+  const handleCloseModal = () => {
+    setIsDealStageModalOpen(false);
+  };
+
+  const handleSuccess = () => {
+    if (selectedDealStage?.id !== undefined) {
+      setStages(removeStage(stages, selectedDealStage.id));
```
ThinuwanW, line 39: 🤖 **Claude** · 🟣 **Nit**

Return types are inconsistent between the two new files: `DealStageModalController` annotates its helpers (`handleCloseModal: void`, `getModalTitle: string`, `getModalContent: ReactNode`), while `handleCloseModal`, `handleSuccess`, `handleError` and `handleDeleteStage` here have none. Strict mode is on and the checklist asks for explicit return values — add `: void` to these four for consistency with the sibling file and with `DeleteDealModalV2`.

frontend/src/community/crm/v2/components/molecules/DeleteDealStageModalContent/DeleteDealStageModalContent.tsx
```diff
@@ -0,0 +1,135 @@
+import { ButtonV2, CloseIcon, DeleteButtonIcon } from "@rootcodelabs/skapp-ui";
```
ThinuwanW, line 1: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

No `.test.tsx` accompanies either new component. To be fair, the entire `src/community/crm` tree currently has zero component tests (only util/hook tests), so this is a module-wide gap rather than a regression — but since this modal performs a destructive mutation with non-trivial store side effects, a small `@testing-library/react` test (wrapped in `MockTheme`) covering confirm → mutate called with the selected id, success → stage removed from the store + success toast, and error → error toast would be cheap insurance.

frontend/src/community/crm/v2/components/molecules/DeleteDealStageModalContent/DeleteDealStageModalContent.tsx
```diff
@@ -0,0 +1,135 @@
...
+    useCrmStoreV2(
+      useShallow((store) => ({
+        stages: store.stages,
+        setStages: store.setStages,
+        selectedDealStageId: store.selectedDealStageId,
+        setIsDealStageModalOpen: store.setIsDealStageModalOpen
+      }))
+    );
+
+  const selectedDealStage = getSelectedStage(stages, selectedDealStageId);
```
ThinuwanW, line 31: 🤖 **Claude** · 🔴 **Important**

The component resolves the stage from `useCrmStoreV2`'s `stages` record, but that record is only ever populated by `useInitializeCrmData` (it calls `setStages(toStagesRecord(data.stages))` from the board init query). `useInitializeCrmData` is mounted in exactly two places: `pages/community/crm/deals.tsx` and `pages/community/crm/companies.tsx`. The only screen that opens deal-stage modals is the CRM tab of `pages/community/configurations/index.tsx` -> `CrmConfigurations`, which never initializes CRM v2 data. So on the page where this modal will actually be used, `stages` is `{}`, `getSelectedStage` returns `undefined`, the description renders with an unresolved `{{stageName}}` placeholder, and the Delete button silently does nothing. Fix by choosing a real data source for this screen: either mount `useInitializeCrmData()` (or a dedicated `useGetDealStages()` fetch) on the configurations page and seed the store before rendering the controller, or pass the selected stage in as a prop from the parent that already holds the stage list. Compare `DeleteCompanyModalContent`, which is used on `companies.tsx` where the store is genuinely initialized.

frontend/src/community/crm/v2/components/organisms/DealStageModalController/DealStageModalController.tsx
```diff
@@ -0,0 +1,69 @@
...
+interface DealStageModalControllerProps {
+  onStageCreated: () => void;
+}
+
+const DealStageModalController: FC<DealStageModalControllerProps> = ({
+  onStageCreated
+}) => {
+  const translateText = useTranslator("configurations", "crm");
+
+  const { isDealStageModalOpen, dealStageModalType, setIsDealStageModalOpen } =
```
ThinuwanW, line 20: 🤖 **Claude** · 🔴 **Important**

The controller reads `isDealStageModalOpen`, `dealStageModalType` and `setIsDealStageModalOpen` from `useCrmStoreV2`, but the only producer of that state is `CrmConfigurations.tsx` (lines 108-137), which writes to `useConfigurationStore` via the legacy `dealStageSlice`. Grepping `crm/v2` shows `setSelectedDealStageId` is defined in `crmUiSlice` and never called anywhere in the v2 tree. The result is two independent Zustand stores holding the same modal UI state, and the v2 controller can never open or receive a selected stage id. The paired change is missing: `CrmConfigurations` must be switched to `useCrmStoreV2` (`setSelectedDealStageId` / `setDealStageModalType` / `setIsDealStageModalOpen`) at the same time the new controller is swapped in, and the now-duplicated `isDealStageModalOpen` / `dealStageModalType` / `selectedDealStageId` fields plus `configurations/stores/slices/dealStageSlice.ts` and `configurations/components/organisms/DealStageModalController` should be removed so there is a single source of truth.

frontend/src/community/crm/v2/components/molecules/DeleteDealStageModalContent/DeleteDealStageModalContent.tsx
```diff
@@ -0,0 +1,135 @@
...
+    handleCloseModal();
+  };
+
+  const handleError = () => {
+    setToastMessage({
+      open: true,
+      toastType: ToastType.ERROR,
+      title: translateText([
+        "deleteDealStageModal",
+        "toastMessages",
```
ThinuwanW, line 65: 🤖 **Claude** · 🔴 **Important**

`handleError` takes no arguments and always renders the generic `errorDescription` ("Failed to delete the deal stage. Please try again"). The backend `CrmDealStageServiceImpl.deleteDealStage` rejects deletes with two specific, non-retryable business errors: `CRM_ERROR_CANNOT_DELETE_TERMINAL_STAGE` and, most commonly, `CRM_ERROR_CANNOT_DELETE_STAGE_WITH_DEALS` (thrown whenever the stage still has non-deleted deals). Telling the user to "please try again" for a delete that can never succeed until they move the deals is actively misleading. Fix by accepting the error and branching on the message key, following the existing CRM pattern in `SidePanelAddDeal.tsx:89-98`: read `error?.response?.data?.results?.[0]?.messageKey`, add `CANNOT_DELETE_STAGE_WITH_DEALS` / `CANNOT_DELETE_TERMINAL_STAGE` members to `CrmErrorMessageKeyEnum` in `crm/v2/enums/common.ts` (it currently only has `DEAL_EXISTS`), and add matching copy under `configurations.crm.deleteDealStageModal.toastMessages` so the toast explains that the stage's deals must be reassigned first.

frontend/src/community/crm/v2/components/molecules/DeleteDealStageModalContent/DeleteDealStageModalContent.tsx
```diff
@@ -0,0 +1,135 @@
...
+      ]),
+      description: translateText([
+        "deleteDealStageModal",
+        "toastMessages",
+        "errorDescription"
+      ])
+    });
+  };
+
+  const { mutate: deleteStage, isPending } = useDeleteDealStage(
```
ThinuwanW, line 76: 🤖 **Claude** · 🔴 **Important**

No React Query invalidation covers the query that actually supplies the stage list. `useDeleteDealStage` in `crm/v2/api/DealApi.ts` only invalidates `crmLimitationQueryKeys.GET_CRM_LIMITATION`; it never touches `crmBoardQueryKeys.BOARD_INIT_DATA` (the query `useInitializeCrmData` reads `data.stages` from) or `crmDealQueryKeys.DEAL_STAGES`. Because this component patches the Zustand `stages` record by hand instead, the React Query cache keeps serving the deleted stage: any remount that resets `isCrmDataInitialized` (route change that clears the store, hard reload within the cache window) rehydrates `stages` from the stale init payload and the deleted stage reappears. Add `queryClient.invalidateQueries({ queryKey: crmBoardQueryKeys.BOARD_INIT_DATA })` and `{ queryKey: crmDealQueryKeys.DEAL_STAGES }` to `useDeleteDealStage`'s `onSuccess` alongside the existing limitation invalidation (the same gap exists in `useCreateDealStage` / `useUpdateDealStage` and is worth fixing together).

frontend/src/community/crm/v2/components/organisms/DealStageModalController/DealStageModalController.tsx
```diff
@@ -0,0 +1,69 @@
...
+        return translateText(["deleteDealStageModal", "title"]);
+      default:
+        return "";
+    }
+  };
+
+  const getModalContent = (): ReactNode => {
+    switch (dealStageModalType) {
+      case CrmModalTypes.ADD_DEAL_STAGE_MODAL:
+        return <DealStageModalForm onStageCreated={onStageCreated} />;
```
ThinuwanW, line 49: 🤖 **Claude** · 🟡 **Suggestion**

The three branches use three different post-mutation refresh strategies. ADD delegates entirely to the `onStageCreated` callback — `DealStageModalForm.handleCreateSuccess` calls `onStageCreated?.()` and never writes the created stage into the `stages` record, even though `useCreateDealStage` hands back the created entity. EDIT writes directly to the store via `setStages(updateStage(...))` with no callback. DELETE writes directly to the store via `setStages(removeStage(...))` with no callback. So after an add the store is stale unless the parent independently refetches, and there is no `onStageUpdated`/`onStageDeleted` for a parent that needs to react (e.g. `CrmConfigurations` resetting its unsaved-reorder `hasChanges` flag after a stage disappears). Pick one strategy for all three — the cleanest is to invalidate the stage query in the API hooks (see the invalidation finding) and drop the manual store patching and the `onStageCreated` prop entirely.

frontend/src/community/crm/v2/components/molecules/DeleteDealStageModalContent/DeleteDealStageModalContent.tsx
```diff
@@ -0,0 +1,135 @@
...
+    if (selectedDealStage?.id !== undefined) {
+      deleteStage(selectedDealStage.id);
+    }
+  };
+
+  return (
+    <div className="flex flex-col">
+      <div>
+        {translateText(["deleteDealStageModal", "description"], {
+          stageName: getStageDisplayName(selectedDealStage?.name)
```
ThinuwanW, line 91: 🤖 **Claude** · 🟡 **Suggestion**

Null-safety gap in the i18n interpolation. `getStageDisplayName` is typed `(stageName?: string) => string | undefined` and returns `undefined` whenever `stageName` is undefined, and `CrmStageEntity.name` is optional, so `stageName` can be `undefined` here. `useTranslator` forwards the interpolation object straight to i18next's `t()`, and i18next leaves an unresolved placeholder in place when the value is `undefined` — the user sees the literal `Are you sure you want to delete the '{{stageName}}' stage?`. Guard it: `getStageDisplayName(selectedDealStage?.name) ?? ""`, or better, do not render the modal body at all when `selectedDealStage` is undefined. Note `DealStageModalForm.tsx:60` already applies the `?? ""` fallback to the same call.

frontend/src/community/crm/v2/components/molecules/DeleteDealStageModalContent/DeleteDealStageModalContent.tsx
ThinuwanW, line 39: 🤖 **Claude** · 🟡 **Suggestion**

`removeStage(stages, ...)` uses the `stages` snapshot captured when the component rendered, and `setStages` replaces the whole record (`set({ stages })`), so this is a blind overwrite rather than a functional update. On the CRM configurations screen a stage reorder save and other stage mutations can complete while a delete is in flight; whichever resolves second wipes the other's changes. Read the current value inside the success handler instead — `setStages(removeStage(useCrmStoreV2.getState().stages, selectedDealStage.id))` — or add a functional-update overload to `setStages` in `crmDataSlice`. `DealStageModalForm.handleEditSuccess` has the same shape and should be fixed alongside.

frontend/src/community/crm/v2/components/organisms/DealStageModalController/DealStageModalController.tsx
```diff
@@ -0,0 +1,69 @@
...
+        dealStageModalType: store.dealStageModalType,
+        setIsDealStageModalOpen: store.setIsDealStageModalOpen
+      }))
+    );
+
+  const handleCloseModal = (): void => {
+    setIsDealStageModalOpen(false);
+  };
+
+  const getModalTitle = (modalType: CrmModalTypes): string => {
```
ThinuwanW, line 33: 🤖 **Claude** · 🟣 **Nit**

Inconsistent helper signatures within the same file: `getModalTitle(modalType: CrmModalTypes)` takes the modal type as a parameter and is then called with the very store value it could have closed over, while `getModalContent()` closes over `dealStageModalType` directly. Make both parameterless (they read the same store field) so the two switches obviously stay in sync.

frontend/src/community/crm/v2/components/molecules/DeleteDealStageModalContent/DeleteDealStageModalContent.tsx
```diff
@@ -0,0 +1,135 @@
...
+import { useDeleteDealStage } from "~community/crm/v2/api/DealApi";
+import useStageNameMapper from "~community/crm/v2/hooks/useStageNameMapper";
+import { useCrmStoreV2 } from "~community/crm/v2/store/store";
+import {
+  getSelectedStage,
+  removeStage
+} from "~community/crm/v2/utils/stageUtil";
+
+const DeleteDealStageModalContent: FC = () => {
+  const translateText = useTranslator("configurations", "crm");
```
ThinuwanW, line 17: 🤖 **Claude** · 🟣 **Nit**

Both new files call `useTranslator("configurations", "crm")`, so components living under `~community/crm/v2` now depend on the `configurations` translation namespace, while the rest of the CRM v2 tree uses `crmModule.*` (e.g. `useStageNameMapper` reads `crmModule.deals.defaultStageNames`, and `DeleteCompanyModalContent` reads `crmModule.companies.deleteCompanyModal`). This cross-module i18n coupling is inherited from the legacy component being ported, but since this PR is moving the code into the CRM module it is the right moment to move the `addDealStageModal` / `editDealStageModal` / `deleteDealStageModal` blocks from `configurations.crm` into `crmModule` and switch the namespace. It also explains the aria-label gap: `configurations.crm` has only `aria.editStage` / `aria.deleteStage` and no `ariaLabels.confirm` / `ariaLabels.cancel` equivalent to what `crmModule.companies.deleteCompanyModal` provides.

PR #1848, opened 2026-08-28
refactor: add deal stages section organism for v2

dulsara-manakal:
Part 5 of 6 of the CRM deal-stage configuration refactor.

DealStagesSection owns the data flow for the stages list:

- reads stages from useGetDealStages into the v2 store
- keeps a draft order while the user drags, with save and cancel
- reorders through useReorderDealStages and writes the persisted stages from the response back into the store
- opens the add, edit and delete modals through the part 4 controller
- guards stage creation with useCrmLimitGuard

Builds on parts 1, 2 and 4.
Still behind isCrmConfigurationsV2 = false until part 6.

### Summary

-

### How to test

1.

### Project Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

#### Other

- [ ] New molecules added
- [ ] New pages(routes) added

### PR Checklist

### Additional Information
-


ThinuwanW (commented):
## 🤖 Claude Code Review

This adds a v2 `DealStagesSection` organism that is a near port of the deal-stage half of `CrmConfigurations.tsx`, moving stage state from local `useState` into the shared `useCrmStoreV2` Zustand record and adding a draft/`hasChanges` layer on top. The port is faithful in layout and i18n, but it drops the `<CrmLimitModalController />` that v1 renders (silently breaking the enterprise limit-reached path), and the new `hasChanges` guard on the draft-sync effect lets `draftStages` desync from the store when a stage is created or deleted mid-reorder, which produces an invalid reorder payload the backend rejects. There is also no React Query invalidation for the `DEAL_STAGES` key after reorder/edit/delete, and the component is not yet referenced anywhere.

Found **23** new issue(s): 🔴 4 important, 🟡 10 suggestion(s), 🟣 9 nit(s)

frontend/src/community/crm/v2/components/organisms/DealStagesSection/DealStagesSection.tsx
```diff
@@ -0,0 +1,210 @@
...
+          onClick={handleSaveReorder}
+          icon={<SaveIcon />}
+          iconPosition="end"
+          disabled={!hasChanges || isReordering}
+        >
+          {translateText(["buttons", "save"])}
+        </ButtonV2>
+      </div>
+
+      <DealStageModalController onStageCreated={refetchStages} />
```
ThinuwanW, line 205: 🤖 **Claude** · 🔴 **Important**

`<CrmLimitModalController />` is missing. `useCrmLimitGuard.guardCrmCreate` (enterprise impl) calls `setLimitedResource(resource)` and returns *without* invoking the callback when the deal-stage limit is reached — the modal controller is what turns that store flag into user-visible feedback. The v1 counterpart renders it explicitly (`CrmConfigurations.tsx:238`), and `ContentLayout` only renders it when `module === Modules.CRM` (`ContentLayout.tsx:398`), which the configurations page (`pages/community/configurations/index.tsx`) does not set. As written, an enterprise tenant at the deal-stage limit clicks "Add" and absolutely nothing happens — no modal, no toast, no error. Add `import CrmLimitModalController from "~enterprise/crm/components/organisms/CrmLimitModalController/CrmLimitModalController";` and render `<CrmLimitModalController />` next to `<DealStageModalController />` (the community fallback renders `null`, so this is safe in community mode).

frontend/src/community/crm/v2/components/organisms/DealStagesSection/DealStagesSection.tsx
```diff
@@ -0,0 +1,210 @@
...
+  const [hasChanges, setHasChanges] = useState(false);
+
+  useEffect(() => {
+    if (!fetchedStages) return;
+
+    setStages(toStagesRecord(fetchedStages));
+  }, [fetchedStages]);
+
+  useEffect(() => {
+    if (hasChanges) return;
```
ThinuwanW, line 71: 🤖 **Claude** · 🔴 **Important**

The `if (hasChanges) return;` guard makes `draftStages` desync from the store whenever the *set* of stages changes during an in-progress reorder, not just their order. Concrete failure: user drags a stage (`hasChanges` becomes true) → clicks "Add" (nothing disables it) → creates a stage → `refetchStages` updates `fetchedStages` → `setStages` updates the store → but this effect bails, so the new stage is invisible in the list. The user then clicks "Save": `toStageReorderPayload(draftStages)` emits N non-terminal items while the backend sees N+1 (`CrmDealStageServiceImpl.reorderDealStages` throws `CRM_ERROR_DEAL_STAGE_REORDER_INVALID_REQUEST` on `existingStages.size() != changedStages.size()`), so the save fails with the generic reorder error toast. The same happens in reverse after a delete (`DeleteDealStageModalContent` removes the stage from the store) — the deleted stage stays rendered and is still sent in the payload. Fix by resyncing whenever the membership changes, not just when `hasChanges` is false — e.g. compute a stable signature `const stageIdSignature = useMemo(() => toStageIds(orderedStages).join(","), [orderedStages])` and resync (plus `setHasChanges(false)`) when that signature changes, or pass `handleCancelReorder` down as an `onStageCreated`/`onStageDeleted` callback so create/delete always discards the stale draft.

frontend/src/community/crm/v2/components/organisms/DealStagesSection/DealStagesSection.tsx
```diff
@@ -0,0 +1,210 @@
...
+      ]),
+      description: translateText([
+        "dealsSection",
+        "toastMessages",
+        "reorderErrorDescription"
+      ])
+    });
+  };
+
+  const { mutate: reorderStages, isPending: isReordering } =
```
ThinuwanW, line 112: 🤖 **Claude** · 🔴 **Important**

No React Query invalidation for `crmDealQueryKeys.DEAL_STAGES` after the reorder. `useReorderDealStages` (`DealApi.ts:237`) only forwards `onSuccess`/`onError`; `useUpdateDealStage` likewise, and `useDeleteDealStage` invalidates only `crmLimitationQueryKeys.GET_CRM_LIMITATION`. So after reorder/edit/delete the Zustand record is patched locally while the `DEAL_STAGES` cache entry still holds the pre-mutation list. On the next remount (no `staleTime` is configured, so cached data is served immediately while refetching), the `useEffect` at line 64 writes that stale cache back into the store via `setStages`, briefly restoring the old order or resurrecting a deleted stage. `refetchStages` is wired only into `onStageCreated` (line 205), which covers create but none of the other three paths. Either add `queryClient.invalidateQueries({ queryKey: crmDealQueryKeys.DEAL_STAGES })` inside the reorder/update/delete mutation hooks in `DealApi.ts`, or call `refetchStages()` from `handleSuccess` here and pass it to the edit/delete flows too.

frontend/src/community/crm/v2/components/organisms/DealStagesSection/DealStagesSection.tsx
```diff
@@ -0,0 +1,210 @@
...
+      setStages: store.setStages,
+      setIsDealStageModalOpen: store.setIsDealStageModalOpen,
+      setDealStageModalType: store.setDealStageModalType,
+      setSelectedDealStageId: store.setSelectedDealStageId
+    }))
+  );
+
+  const {
+    data: fetchedStages,
+    isLoading,
```
ThinuwanW, line 55: 🤖 **Claude** · 🟡 **Suggestion**

`useGetDealStages()` destructures only `data` and `isLoading` — `isError` is dropped, so a failed stage fetch renders `DealStagesDraggableContent` with an empty `draftStages` array and the "complete state" note, indistinguishable from a tenant with no stages. Per the review guidelines every async surface needs a rendered error state. Pull `isError` out of the hook and render an error message (or fire an error toast, consistent with `useInitializeCrmData.ts:47-55`) instead of the empty list.

frontend/src/community/crm/v2/components/organisms/DealStagesSection/DealStagesSection.tsx
```diff
@@ -0,0 +1,210 @@
...
+  }, [fetchedStages]);
+
+  useEffect(() => {
+    if (hasChanges) return;
+
+    setDraftStages(orderedStages);
+  }, [orderedStages, hasChanges]);
+
+  const handleSuccess = (reorderedStages: CrmStageEntity[]) => {
+    setStages({ ...stages, ...toStagesRecord(reorderedStages) });
```
ThinuwanW, line 77: 🤖 **Claude** · 🟡 **Suggestion**

`handleSuccess` closes over `stages` from the render in which `mutate` was invoked. React Query builds the mutation with the observer options captured at `mutate()` time, so if any other mounted consumer of `useCrmStoreV2` mutates `stages` while the reorder request is in flight (e.g. the delete modal's `setStages(removeStage(...))`), this spread will resurrect the stale snapshot and discard that change. Read the live value instead: `setStages({ ...useCrmStoreV2.getState().stages, ...toStagesRecord(reorderedStages) })`.

frontend/src/community/crm/v2/components/organisms/DealStagesSection/DealStagesSection.tsx
```diff
@@ -0,0 +1,210 @@
+import {
```
ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

`DealStagesSection` is not imported anywhere in the repo — grepping `src/` and `pages/` returns only its own definition and default export. It ships as dead code that no bundle entry reaches, so none of it is exercised in the running app. If this is an intentional incremental step toward replacing the deal-stage block in `CrmConfigurations.tsx`, that's fine, but please note it in the PR description; otherwise wire it into `configurationTabsUtil.tsx` / the v2 configurations tab in this PR so it actually gets reviewed against real data.

ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

This duplicates roughly 100 lines of `src/community/configurations/components/organisms/CrmConfigurations/CrmConfigurations.tsx` — the header + add button, the skeleton/draggable-content switch, both toast handlers with identical translation keys, and the cancel/save button row are byte-for-byte equivalent modulo the v1/v2 type and store swap. Two copies of the same deal-stage reorder UI will drift (they already differ on the limit modal and the draft-reset semantics). Please state the migration plan: either delete the deal-stage block from `CrmConfigurations` and have it render `<DealStagesSection />`, or add a tracking ticket so the v1 copy is removed once the v2 configurations tab lands.

frontend/src/community/crm/v2/components/organisms/DealStagesSection/DealStagesSection.tsx
```diff
@@ -0,0 +1,210 @@
...
+  const handleAddStage = () => {
+    guardCrmCreate(CrmLimitResource.DEAL_STAGES, () => {
+      setSelectedDealStageId(null);
+      setDealStageModalType(CrmModalTypes.ADD_DEAL_STAGE_MODAL);
+      setIsDealStageModalOpen(true);
+    });
+  };
+
+  return (
+    <>
```
ThinuwanW, line 154: 🤖 **Claude** · 🟣 **Nit**

The component returns a bare fragment with two sibling top-level `div`s, whereas the v1 equivalent wraps them in `<div className="flex flex-col gap-6 w-[49rem]">`. There is no spacing between the stage list and the cancel/save button row unless the (currently nonexistent) parent happens to be a `flex flex-col` with a gap. Since this is an organism meant to be dropped into a tab, either wrap the two blocks in a `flex flex-col gap-6` container here or document the layout contract the parent must satisfy.

frontend/src/community/crm/v2/components/organisms/DealStagesSection/DealStagesSection.tsx
```diff
@@ -0,0 +1,210 @@
...
+        <ButtonV2
+          variant="tertiary"
+          onClick={handleCancelReorder}
+          icon={<CloseIcon />}
+          iconPosition="end"
+          disabled={!hasChanges || isReordering}
+        >
+          {translateText(["buttons", "cancel"])}
+        </ButtonV2>
+        <ButtonV2
```
ThinuwanW, line 194: 🤖 **Claude** · 🟣 **Nit**

The Save button uses `isReordering` only to disable itself; it never shows progress. `ButtonV2` supports `isLoading` (used a few lines up at line 166 for `isCheckingCrmLimit`, and in `DealStageModalForm.tsx:223`). Add `isLoading={isReordering}` so the user gets feedback while the reorder POST is in flight.

frontend/src/community/crm/v2/components/organisms/DealStagesSection/DealStagesSection.tsx
ThinuwanW, line 1: 🤖 **Claude** · 🟣 **Nit**

📁 **File-level observation**

No `DealStagesSection.test.tsx` accompanies the new organism. The draft/`hasChanges` state machine here is the non-trivial part of the component (sync-on-fetch, cancel-resets-draft, success-clears-changes, payload construction via `toStageReorderPayload`) and is exactly the kind of logic worth pinning down. A test rendering with the `MockTheme` wrapper plus a mocked `useGetDealStages` that asserts the reorder payload and the cancel behaviour would catch the desync described above.

frontend/src/community/crm/v2/components/organisms/DealStagesSection/DealStagesSection.tsx
```diff
@@ -0,0 +1,210 @@
...
+      setIsDealStageModalOpen: store.setIsDealStageModalOpen,
+      setDealStageModalType: store.setDealStageModalType,
+      setSelectedDealStageId: store.setSelectedDealStageId
+    }))
+  );
+
+  const {
+    data: fetchedStages,
+    isLoading,
+    refetch: refetchStages
```
ThinuwanW, line 56: 🤖 **Claude** · 🔴 **Important**

The DEAL_STAGES query cache is never invalidated after a stage is edited or deleted, yet this component makes that cache the authoritative hydration source for the shared Zustand `stages` slice (the effect blindly calls `setStages(toStagesRecord(fetchedStages))` on every identity change of the fetched array). `useUpdateDealStage` and `useDeleteDealStage` in `~community/crm/v2/api/DealApi.ts` only invalidate `crmLimitationQueryKeys.GET_CRM_LIMITATION`; the modal contents patch the store locally via `updateStage`/`removeStage` and leave the query cache holding the pre-mutation list. Concretely: delete a stage, switch to another configurations tab and back — the component remounts, React Query hands back the stale cached array synchronously, the effect runs, and the deleted stage reappears in the store (and in the kanban/table, which read the same slice) until the background refetch lands; if that refetch fails the resurrection is permanent. Same for an edited name/colour reverting. The `onStageCreated` refetch prop only patches the create path, which is exactly the asymmetry that gives this away. Fix: add `queryClient.invalidateQueries({ queryKey: crmDealQueryKeys.DEAL_STAGES })` to the `onSuccess` of `useUpdateDealStage` and `useDeleteDealStage` (and `useCreateDealStage`), then drop the manual `onStageCreated`/`refetch` prop entirely so all four mutations resync through one mechanism.

frontend/src/community/crm/v2/components/organisms/DealStagesSection/DealStagesSection.tsx
```diff
@@ -0,0 +1,210 @@
...
+      description: translateText([
+        "dealsSection",
+        "toastMessages",
+        "reorderSuccessDescription"
+      ])
+    });
+  };
+
+  const handleError = () => {
+    setToastMessage({
```
ThinuwanW, line 96: 🤖 **Claude** · 🟡 **Suggestion**

The reorder `onError` handler shows a toast but performs no rollback and no resync: `draftStages` keeps the unsaved order and `hasChanges` stays true, so the UI keeps rendering an order that was never persisted. This matters more than it looks because the backend (`CrmDealStageServiceImpl.reorderDealStages`) throws `CRM_ERROR_DEAL_STAGE_REORDER_INVALID_REQUEST` whenever the payload's item count differs from the server's non-terminal stage count — i.e. precisely when the local store has drifted from the server (a stage added or deleted in another tab/session). In that state every retry fails identically and the user is stuck with a permanently unsavable draft and a generic 'try again' message. Also, the handler ignores its `AxiosError` argument, so no server-supplied message is surfaced. Suggested fix: in `handleError`, accept the error, reset `hasChanges` to false and call `refetchStages()` (or invalidate `crmDealQueryKeys.DEAL_STAGES`) so the draft resyncs to server truth, and surface the API message where available.

frontend/src/community/crm/v2/components/organisms/DealStagesSection/DealStagesSection.tsx
```diff
@@ -0,0 +1,210 @@
...
+    useShallow((store) => ({
+      stages: store.stages,
+      setStages: store.setStages,
+      setIsDealStageModalOpen: store.setIsDealStageModalOpen,
+      setDealStageModalType: store.setDealStageModalType,
+      setSelectedDealStageId: store.setSelectedDealStageId
+    }))
+  );
+
+  const {
```
ThinuwanW, line 53: 🤖 **Claude** · 🟡 **Suggestion**

This introduces a second, uncoordinated writer for the same `stages` slice of the v2 CRM store. `useInitializeCrmData` already loads stages (from the board init-data endpoint) and writes them with `setStages(toStagesRecord(...))`, guarded by `isCrmDataInitialized` — a flag that is never reset back to false anywhere in the codebase. This component fetches the same data again from `/deal/stage` and overwrites the slice wholesale. The result is an extra network round-trip for data already in memory, plus a last-writer-wins race whenever both hooks are alive during a client-side navigation. Consider either gating this query on the store being empty (`useGetDealStages(Object.keys(stages).length === 0)` — the hook already accepts an `enabled` argument that is currently not passed), or having this section read from the store populated by `useInitializeCrmData` and only invalidate/refetch on mutation.

frontend/src/community/crm/v2/components/organisms/DealStagesSection/DealStagesSection.tsx
```diff
@@ -0,0 +1,210 @@
...
+    data: fetchedStages,
+    isLoading,
+    refetch: refetchStages
+  } = useGetDealStages();
```
ThinuwanW, line 57: 🤖 **Claude** · 🟡 **Suggestion**

This is the only consumer of `useGetDealStages`, and unlike every other query in `DealApi.ts` (`useGetDealsInfinite`, `useGetDealById`, `useGetDealsGroupedByStages`) that hook sets neither `refetchOnWindowFocus: false` nor a `staleTime`. Combined with the effect that unconditionally pushes the result into the shared Zustand slice, every window refocus and every remount triggers a refetch that overwrites global CRM state — including state that other mounted v2 components (deals table, kanban lanes, side panels) read. Add `refetchOnWindowFocus: false` and a `staleTime` to `useGetDealStages` in `DealApi.ts` to match the module's established query configuration.

frontend/src/community/crm/v2/components/organisms/DealStagesSection/DealStagesSection.tsx
```diff
@@ -0,0 +1,210 @@
...
+    setSelectedDealStageId(stage.id);
+    setDealStageModalType(CrmModalTypes.DELETE_DEAL_STAGE_MODAL);
+    setIsDealStageModalOpen(true);
+  };
+
+  const handleSaveReorder = () => {
+    reorderStages(toStageReorderPayload(draftStages));
+  };
+
+  const handleAddStage = () => {
```
ThinuwanW, line 145: 🤖 **Claude** · 🟡 **Suggestion**

The CRM limit guard fails open while its own query is in flight. `handleAddStage` calls `guardCrmCreate` unconditionally; the enterprise implementation reads `limitationData?.[resource]?.limitReached`, which is `undefined` while the limitation query is pending or has errored, so the falsy check falls straight through to `onAllowed()` and the add-stage modal opens even for a tenant that is already at its deal-stage limit. `isCheckingCrmLimit` is wired only to the button's `isLoading` prop, which may or may not imply `disabled` in `ButtonV2` — I could not verify since `@rootcodelabs/skapp-ui` is not installed in this checkout. Make it explicit: add `disabled={isCheckingCrmLimit}` to the add button, and/or early-return from `handleAddStage` when `isCheckingCrmLimit` is true. (v1 `CrmConfigurations` has the same shape, so worth fixing in both.)

frontend/src/community/crm/v2/components/organisms/DealStagesSection/DealStagesSection.tsx
```diff
@@ -0,0 +1,210 @@
...
+        "toastMessages",
+        "reorderErrorDescription"
+      ])
+    });
+  };
+
+  const { mutate: reorderStages, isPending: isReordering } =
+    useReorderDealStages(handleSuccess, handleError);
+
+  const handleStagesReorder = (reordered: CrmStageEntity[]) => {
```
ThinuwanW, line 115: 🤖 **Claude** · 🟡 **Suggestion**

`handleStagesReorder`, `handleEdit` and `handleDelete` are recreated on every render and passed as props into `DealStagesDraggableContent`, which in turn passes `onEdit`/`onDelete` down to every `DraggableDealStageCard`. Each card runs `useSortable` and a dnd-kit sortable subtree, so the whole drag context re-renders on any parent state change (including each keystroke of an unrelated re-render or each `isReordering` toggle). Wrap the three handlers in `useCallback` (deps: the relevant store setters, which are stable) and consider wrapping `DraggableDealStageCard` in `React.memo` — the review guidelines call this out explicitly for callbacks passed to children.

frontend/src/community/crm/v2/components/organisms/DealStagesSection/DealStagesSection.tsx
```diff
@@ -0,0 +1,210 @@
...
+            icon={<PlusIcon />}
+            iconPosition="end"
+            size="md"
+            isLoading={isCheckingCrmLimit}
+          >
+            {translateText(["dealsSection", "buttons", "add"])}
+          </ButtonV2>
+        </div>
+
+        {isLoading ? (
```
ThinuwanW, line 172: 🤖 **Claude** · 🟣 **Nit**

The skeleton is gated purely on the query's `isLoading`, ignoring that the store may already be hydrated. If the user arrives from the CRM deals board, `useInitializeCrmData` has already populated `stages`, so `draftStages` is non-empty and renderable, yet this still shows `DraggableDealStageCardSkeleton` for the duration of a redundant fetch. Gate on both, e.g. show the skeleton only when `isLoading && draftStages.length === 0`, so cached/store-backed data renders immediately.

frontend/src/community/crm/v2/components/organisms/DealStagesSection/DealStagesSection.tsx
ThinuwanW, line 205: 🤖 **Claude** · 🟣 **Nit**

React Query's `refetch` is passed directly where the prop type is `() => void`. Two small problems: the returned promise is floating (an unexpected rejection surfaces as an unhandled rejection rather than a toast), and the real signature takes a `RefetchOptions` argument, so any future caller that forwards a value — an event object, a created-stage payload — would have it silently reinterpreted as refetch options. Wrap it: pass an arrow that calls `void refetchStages()`. Better still, remove the prop entirely and invalidate `crmDealQueryKeys.DEAL_STAGES` inside `useCreateDealStage` (see the cache-invalidation finding).

frontend/src/community/crm/v2/components/organisms/DealStagesSection/DealStagesSection.tsx
```diff
@@ -0,0 +1,210 @@
...
+  const orderedStages = useMemo(() => getOrderedStages(stages), [stages]);
+
+  const [draftStages, setDraftStages] = useState<CrmStageEntity[]>([]);
+  const [hasChanges, setHasChanges] = useState(false);
+
+  useEffect(() => {
+    if (!fetchedStages) return;
+
+    setStages(toStagesRecord(fetchedStages));
+  }, [fetchedStages]);
```
ThinuwanW, line 68: 🤖 **Claude** · 🟣 **Nit**

The effect that syncs fetched stages into the store omits `setStages` from its dependency array. `react-hooks/exhaustive-deps` is configured as `warn` in `.eslintrc.json`, so this will produce a lint warning rather than a build failure. Zustand setters are referentially stable so there is no runtime bug, but adding `setStages` (or pulling it via `useCrmStoreV2.getState()` inside the effect, as `DealsKanbanBoardV2` does) keeps the file lint-clean and consistent with the module.

frontend/src/community/crm/v2/components/organisms/DealStagesSection/DealStagesSection.tsx
```diff
@@ -0,0 +1,210 @@
...
+    setHasChanges(true);
+  };
+
+  const handleCancelReorder = () => {
+    setDraftStages(orderedStages);
+    setHasChanges(false);
+  };
+
+  const handleEdit = (stage: CrmStageEntity) => {
+    if (stage.id === undefined) return;
```
ThinuwanW, line 126: 🤖 **Claude** · 🟣 **Nit**

The `stage.id === undefined` early returns in `handleEdit` and `handleDelete` are unreachable dead code. Every stage rendered by `DealStagesDraggableContent` originates from `toStagesRecord`, which drops entries whose id is null/undefined, and the child additionally renders a card only when `stage.id !== undefined`. Beyond being dead, the guard is a silent no-op: if it ever did fire the user would click Edit/Delete and get nothing, with no toast and no log. Either drop the guards and tighten the callback prop types to require a defined id, or replace the silent return with real feedback.

frontend/src/community/crm/v2/components/organisms/DealStagesSection/DealStagesSection.tsx
ThinuwanW, line 77: 🤖 **Claude** · 🟣 **Nit**

Inconsistent null handling between the two consumers of the same response shape. The fetch effect defensively bails when the stages array is falsy, but `handleSuccess` feeds the mutation result straight into `toStagesRecord`, which does a `for...of` over it. Both values come from an `response?.data?.results` access whose optional chaining implies the envelope is not guaranteed; if `results` were ever absent this throws a TypeError inside the mutation's `onSuccess`, which React Query does not route to `onError`, so the user sees neither a success nor an error toast. Add the same falsy guard (or default to an empty array) before calling `toStagesRecord`.

frontend/src/community/crm/v2/components/organisms/DealStagesSection/DealStagesSection.tsx
```diff
@@ -0,0 +1,210 @@
...
+          iconPosition="end"
+          disabled={!hasChanges || isReordering}
+        >
+          {translateText(["buttons", "cancel"])}
+        </ButtonV2>
+        <ButtonV2
+          variant="primary"
+          onClick={handleSaveReorder}
+          icon={<SaveIcon />}
+          iconPosition="end"
```
ThinuwanW, line 198: 🤖 **Claude** · 🟣 **Nit**

i18n key namespacing is inconsistent within a single component: the add button resolves under `crm.dealsSection.buttons.add`, while cancel and save reach up to the tab-level `crm.buttons.cancel` / `crm.buttons.save`. Meanwhile `configurations.json` already defines `crm.dealsSection.buttons.cancel` and `crm.dealsSection.buttons.saveChanges` with identical copy, and those keys are now dead (as is `crm.dealsSection.addButtonText`). Pick one namespace for this section's buttons — `dealsSection.buttons.*` is the more specific and matches the guideline's `module.section.key` convention — and delete the orphaned keys.

frontend/src/community/crm/v2/components/organisms/DealStagesSection/DealStagesSection.tsx
ThinuwanW, line 1: 🤖 **Claude** · 🟡 **Suggestion**

📁 **File-level observation**

Module placement and the role-gating that must accompany wiring this up. The component sits under `~community/crm/v2/components/organisms/` but is a configurations-tab section: it translates against the `configurations.crm` namespace, and its v1 counterpart lives in `~community/configurations/components/organisms/CrmConfigurations/`. Consider moving it to the configurations module (or at minimum note why the v2 CRM module owns it) so the directory matches the feature. More importantly for the wiring: today the CRM configurations surface is gated solely by `getConfigurationTabs`, which restricts the `crm` tab to `AdminTypes.SUPER_ADMIN` and `AdminTypes.CRM_ADMIN` — there is no role check inside the component itself. Whoever mounts this must register it through the tabs util (community and the enterprise `getEnterpriseConfigurationTabs`) with the same `requiredRoles`, not on a bare page, or every stage add/edit/delete/reorder control becomes reachable by any authenticated user. Note also that the configurations page renders `ContentLayout` without a `module` prop, so the layout-level `CrmLimitModalController` is not mounted there.

